PR #32, opened 2016-07-26
Updating instrument docs

cjermain:
Many of the instruments are not included in the current docs. This PR adds those instruments and ensures a basic level of information is available.

### DIFF
--- a/docs/api/adapters.rst
+++ b/docs/api/adapters.rst
@@ -2,40 +2,54 @@
 pymeasure.adapters
 ##################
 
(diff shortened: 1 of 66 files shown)


cjermain:
The previous `add_control` method has been replaced with `Instrument.control`, which returns a property instead of setting a class variable with the property on construction. This allows properties to be defined on importing the class, before it has been constructed, so that the docs are readily available for these properties. Also, `add_measurement` has been replaced with `Instrument.measurement`. The new functions should be called above the class `__init__` and set to their respective class variables.

PR #33, opened 2016-07-27
Validator implementation for Instrument.control

cjermain:
The new `Instrument.control` method implementations are limited by their ability to validate the values being passed into the SCPI/GPIB commands. This PR adds support for validator functions that act to ensure the value is appropriate, and allows the user to specific custom functions. A group of stock validators are provided in `pymeasure.instruments.validators`. Before merging, this will be integrated into the existing instruments and include working tests.


cjermain:
The version number increment to 0.4 is included here.

PR #34, opened 2016-07-28
Replacing pip in continous-integration with conda

cjermain:
This PR replaces pip in the CI scripts with conda. The RTD docs have already been switched to conda, where the easy installation of the Qt libraries makes conda superior. Before merging, this branch will successfully demonstrate CI with conda on both TravisCI and Appveyor, as well as for macOS.

This follows a few good examples: 
https://github.com/glue-viz/conda-package-builder-macosx
https://github.com/rmcgibbo/python-appveyor-conda-example


cjermain:
TravisCI is building correctly for Python 3.4 and 3.5 on Linux and MacOSX.

cjermain:
Conda is working in both TravisCI and Appveyor.

bilderbuchi:
@cjermain  Question: why are the requirements specified separately for the different python versions (conda-3.*.yml), and different package versions specified in them to boot? This leads to [strange build failures](https://travis-ci.org/ralph-group/pymeasure/builds/354710347) as the 3 tests run 3 different setuptools versions.

cjermain:
@bilderbuchi, great question. The conda files are typically very specific about the build numbers, which can differ between Python versions and operating systems. That being said, I expect there is a better way to do this. What do you think?

bilderbuchi:
Well, this is probably one of those cases that just has no "nice" solution. Even if you find a set of package versions that are compatible with all your python versions, you would still have to supply the correct python version elsewhere (that is currently given in the files). We should have a look around how other big packages manage this.

PR #41, opened 2016-08-03
Add instructions on how to install with conda

melund:
I  have added instructions on how to install with conda. 

I think the install instructions would be simpler if you didn't suggest doing the install through a notebook. I haven't changed that though...

### DIFF
--- a/docs/quick_start.rst
+++ b/docs/quick_start.rst
@@ -19,12 +19,31 @@ Download and install the appropriate Python 3.5 version of `Anaconda`_ for your 
 Installing PyMeasure
 ====================
 
-Now that you have Python installed, open up a "Jupyter Notebook". We will use this notebook to install the PyMeasure package with :code:`pip`, which you can alternatively do in the command-line. Run the following code in the notebook (here the "!" tells the notebook to run a terminal command instead of regular Python):
+Install with conda
+------------------
+
+Now that you have Python installed, open up a "Jupyter Notebook". We will use this notebook to install the PyMeasure package with :code:`conda`, which you can alternatively do in the command-line. Run the following code in the notebook (here the "!" tells the notebook to run a terminal command instead of regular Python):
+
+.. code-block:: none
+
+   > !conda config --add channels conda-forge
+   > !conda install xonsh
+
+This will install PyMeasure and all the required dependencies. 
+
+Install with ``pip``
+--------------------
+
+If you don't have the conda package manager (i.e. you don't use the `Anaconda distribution`_ ) you can also install PyMeasure using :code:`pip`.
 
 .. code-block:: none
     
     !pip install pymeasure
 
+    
+Check that it works
+-------------------
+   
 In another cell of the notebook, we can check the version number of PyMeasure by executing the following.
 
 .. code-block:: python
--- a/docs/quick_start.rst
+++ b/docs/quick_start.rst
@@ -26,8 +26,8 @@ Now that you have Python installed, open up a "Jupyter Notebook". We will use th
 
 .. code-block:: none
 
-   > !conda config --add channels conda-forge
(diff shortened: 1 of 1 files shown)


cjermain:
Please correct the package name and remove the ">" characters from the commands. 

I wrote the instructions for the notebook to allow beginners to install and test the package in the same environment, and introduce the "!" command that I think is very handy for Windows users. I assume more advanced users can get it installed on their own. I'm open to suggestions for improving the instructions though.

melund:
Done. 

Finding the best way to target beginner is extremely tricky. That is also my experience. So maybe you are onto something with the notebook approach. Although I guess most users trying PyMeasure would know what a command prompt/terminal is. 

I have had some luck creating a tutorial as a Jupyter Notebook for one of my first projects. https://github.com/AnyBody-Research-Group/AnyPyTools/blob/master/Tutorial/00_AnyPyTools_tutorial.ipynb

The cool thing about that approach is that I could bundle the tutorial with the conda package. So users would get a live version of the tutorial after installing my package with conda.

cjermain:
I've had a chance to run the instructions in the notebook, and found it requires the `--yes` argument.
```
!conda config --add channels conda-forge
!conda install pymeasure --yes
```
The print out is pretty badly formatted, so I don't think that conda in the notebook is worth pointing people towards as @melund mentioned. Since its preferred that people move to the [in-place installation](http://pymeasure.readthedocs.io/en/latest/dev/contribute.html#using-the-development-version) (`pip install -e .`) version for contributing features, I want to keep this consistent. For now, lets keep the pip instruction, and add a sub-section of "Installing PyMeasure" that briefly describes the alternative installation using conda-forge.

@ddale, do you have any thoughts on the best way to instruct beginners to install the package?

melund:
You are right. IPython and the notebook, in particular, is not well suited for working with CLI's like pip or conda. And in general, it is not the best practice to introduce to new users. Let them install PyMeasure from the command line. 

I am not a big fan of `pip`.  And suggesting people to use `pip` with the anaconda python distribution is a terrible idea. `Conda` will not know of the packages that were pip-installed and you can easily end in a situation where the dependencies of PyMeasure are suddenly broken when users install something else. 

I guess `pip install -e` is the new `python setup.py develop`. That is the only use of pip that makes sense when you are using the Anaconda python distribution. 

I would really suggest that you just make the instructions for different scenarios. The first one could be targeted, beginners. I.e. Anaconda python distribution and the conda commands.  The second one could be installing with pip for people who don't want to use Anaconda.

ddale:
I agree with all of this.

ddale:
> 
> @ddale, do you have any thoughts on the best way to instruct beginners to install the package?
> 

I think the easiest way for beginners is to use Anaconda Navigator: https://docs.continuum.io/anaconda/navigator. I'll look into supporting it on Monday.

melund:
@ddale. That is a good option for non-tech users. It is also very easy to do. All that is needed is to add an `app` section to the recipe. See how we do it for xonsh. https://github.com/conda-forge/xonsh-feedstock/blob/master/recipe/meta.yaml#L41

The only downside is that the navigator is still a little unfinished. Icon and the detailed descriptions don't work but should be fixed soon. See this conversation: https://github.com/ContinuumIO/anaconda-issues/issues/937#issuecomment-234980373

cjermain:
Thats a good idea @ddale, but I agree with @melund that Navigator is not that polished. It may be a good solution in the future when it comes out of beta. 

@melund, can you update the docs to use the "Anaconda prompt" and conda instructions? Please make sure to correct the package to `pymeasure`.

melund:
@cjermain I have updated the instruction. Hope this is better

cjermain:
@melund, thanks!

PR #42, opened 2016-08-09
Re-introducing the Newport ESP300

cjermain:
From the older commit 1c4153e7d2a9a896fcfbb1964fdcfd47f31aaf32, this PR re-introduces the Newport ESP300 Motion Controller that Graham wrote. [Previous code for the ESP300](https://github.com/ralph-group/pymeasure/blob/1c4153e7d2a9a896fcfbb1964fdcfd47f31aaf32/control/instruments/esp.py) needs to be modified to be up to date with the new `Instrument.control` and `Instrument.measurement` properties.

PR #46, opened 2016-08-12
Adding GMW 5201 and 5403 electromagnets

cjermain:
Following the addition of the Newport ESP 300 Motor Controller in #42, the GMW 5201 Projected Field Electromagnet can be defined. This PR also includes the GMW 5403 C-Frame Electromagnet, which share common attributes and methods. Code is based partially on Graham's previous work in the previous commit 1c4153e7d2a9a896fcfbb1964fdcfd47f31aaf32, specifically [magnet.py](https://github.com/ralph-group/pymeasure/blob/1c4153e7d2a9a896fcfbb1964fdcfd47f31aaf32/control/instruments/magnet.py).

### DIFF
--- a/docs/api/instruments/gmw/gmw5201.rst
+++ b/docs/api/instruments/gmw/gmw5201.rst
@@ -1,0 +1,7 @@
+######################################
+GMW 5201 Projected Field Electromagnet
+######################################
+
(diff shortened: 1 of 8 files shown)


cjermain:
Since we are transitioning instruments to InstrumentKit (#53, #94), new instruments will not be accepted.

PR #51, opened 2016-08-29
Remove support for old PyVisa versions

minhhaiphys:
for a couple of reasons:
1. In the requirements for using PyMeasure, we ask users to use PyVisa 1.8 or later.

2. There are some discrepancies between PyVisa 1.4 and 1.7 (or later) in the way they interact with instruments, especially return values, which has been the source of many reading errors.

3. We don't want to test every new (and current) functionality on so many version of PyVisa.

### DIFF
--- a/docs/about/authors.rst
+++ b/docs/about/authors.rst
@@ -1,7 +1,7 @@
 Authors
 =======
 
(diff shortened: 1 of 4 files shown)


cjermain:
Thanks @minhhaiphys! This looks good. Can you remove line 34? `long = int # Python 3 fix` is no longer needed. Go ahead and merge when ready.

PR #58, opened 2017-03-01
Implement USBTMC Protocol for Adapters and Instruments

TvBMcMaster:
USBTMC is a new protocol for communicating with GPIB enabled instruments over a USB port.  

The core `Adapter` and `Instrument` classes have been extended to allow for the `USBTMCAdapter` and `USBTMCInstrument` classses, which both allow for specialized instantiation parameters specific to USB devices.  

I use this code on a current experiment with a USBTMC instrument.  

Relevant documentation is included as well for quick integration for new users.

### DIFF
--- a/pymeasure/adapters/usbtmc.py
+++ b/pymeasure/adapters/usbtmc.py
@@ -25,50 +25,78 @@
 import logging
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
+from time import sleep
 from functools import wraps
-from .adapter import Adapter
-try:
-    from usbtmc import Instrument as USBTMCDriver
-    from usb.core import USBError
-except ImportError:
-    log.warning("Missing required package, python-usbtmc")
+from .adapter import Adapter, AdapterError
+from usb.core import USBError
+from usbtmc.usbtmc import Instrument as USBTMCDriver, UsbtmcException, struct
+
+RECONNECTION_ATTEMPTS = 3
 
 
-def reconnect_attempt(fn):
-    """Method wrapper to reconnect if method times out."""
-
+def attempt_reconnect(fn):
+    """Method wrapper to reconnect if wrapped method times out."""
+    @wraps(fn)
     def wrapped(self, *args, **kwargs):
-        n_attempts = 3
+        n_attempts = RECONNECTION_ATTEMPTS
         while n_attempts > 0:
-            log.info("Reconnect Attempts: {}".format(n_attempts))
             try:
                 return fn(self, *args, **kwargs)
             except USBError as err:
-                log.error(dir(err))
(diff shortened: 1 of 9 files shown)


pymeasure/adapters/usbtmc.py
```diff
@@ -0,0 +1,170 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2016 PyMeasure Developers
+#
+# Python USBTMC driver, lifted from python-usbtmc.
```
cjermain: What is the license for `python-usbtmc`? Can this code be included under the MIT License?

cjermain:
@TvBMcMaster, what do you think about adding a `__del__` method to `USBTMCAdapter`, to ensure `close()` is called?

Thanks also for removing all of the trailing whitespace.

pymeasure/adapters/usbtmc.py
TvBMcMaster: https://github.com/python-ivi/python-usbtmc/blob/master/COPYING

TvBMcMaster:
@cjermain That's a good idea.  I'll give it a try and see if it will fail gracefully without the Instrument hanging up.

requirements/dev.txt
```diff
@@ -0,0 +1,3 @@
+sphinx
+python-usbtmc
```
bilderbuchi: don't you also have to add the python-usbtmc dependency to setup.py?

pymeasure/adapters/usbtmc.py
bilderbuchi: What does "lifted" mean? If copied directly, maybe it's better to import-and-extend instead of copying source code (as you won't get upstream fixes/features that way). If not, maybe worth it to clarify that, and the actual license that was under, in the source.

bilderbuchi:
Keep in mind that `__del__` is not the opposite of `__init__`. It is only called when the object is garbage collected, not necessarily when you delete it (as there might be some other references left). I have run into this problem in the past with serial connections.
For resource closing you should use a context manager, this seems to be the only clean way to do that in python. Sadly. :-(

Further reading:
https://docs.python.org/3/reference/datamodel.html#object.__del__
https://stackoverflow.com/questions/24611529/pythonic-way-to-close-connection-like-objects-in-del
https://stackoverflow.com/questions/865115/how-do-i-correctly-clean-up-a-python-object
https://stackoverflow.com/questions/6104535/i-dont-understand-this-python-del-behaviour
https://stackoverflow.com/questions/10767905/db-connections-with-closing-vs-del

cjermain:
It seems that this PR is stale. I will close it for now, and am happy to re-open.

PR #62, opened 2017-03-17
Enable Read/Write delay on Prologix adpaters.

TvBMcMaster:
Prologix will return an empty string from an `ask` request if the instrument is not able to
process the write command fast enough before the read command is issued.  This can cause sync issues with subsequent read commands.

This commit allows for a simple keyword parameter, `rw_delay` to the prologix adapter instantiation,
which is a value in seconds to wait between the read and write of any `ask` call.


cjermain:
Looks good. Thanks @TvBMcMaster!

PR #67, opened 2017-03-22
Adding LCRmeter Agilent E4980

dvspirito:
This is to add LCRmeter Agilent/Keysight E4980.


cjermain:
Looks good. Thanks @dvspirito!

PR #69, opened 2017-03-23
Adding resource manager

dvspirito:
This is to add a wrapper around pyvisa ResourceManager. It can be useful to quickly check instruments.
Is it ok to put the code in the pymeasure root?

In the future, I'd like to add a function that automatically chooses the Instrument class for a resource based on idn string. This would require a list of all the idn string.


cjermain:
@dvspirito, this is an interesting idea. It would be great to be able to automatically generate instrument objects for those attached. 

I suggest that you put the `resources.py` file in `pymeasure.instruments`, and import the `list_resources` function in the `__init__.py` file. That way we can import the function as:

```python
from pymeasure.instruments import list_resources
```
I agree with your suggestion for using the idn to classify each instrument. There are a few ways to do this, but I suggest a centralized dictionary in `resources.py` for all VISA compatible instruments. This has the advantage that the code does not need to iterate over all of the instruments to find the match, which would be the case if the idn strings are stored in a class variable in each object. The disadvantage of a centralized dictionary is that it has to be kept up to date in addition to the instrument objects. What do you think of this design? Do you want to start a new PR for this?

dvspirito:
Thanks for the suggestions @cjermain.

I agree that a centralized dictionary is probably easier and more convenient; moreover, it would not be a big issue if the dictionary is not updated, as the resource manager is only for interactive use.

I will start with this idea, and then open another PR. For now, `list_resources` should be enogh.

pymeasure/instruments/__init__.py
```diff
@@ -51,4 +51,6 @@ class RangeException(Exception): pass
 from . import srs
 from . import tektronix
 from . import thorlabs
-from . import yokogawa
\ No newline at end of file
+from . import yokogawa
+from resources import list_resources
```
cjermain: The CI is failing because this line needs to a relative import (`from .resources import list_resources`).

dvspirito: Thanks, I've seen it.

cjermain:
Thanks @dvspirito!

Issue #71, opened 2017-04-03
pip install pymeasure generates error


cbrown1:
I get the following error when trying to install via pip:

> [cbrown@cbrown2` ~]$ sudo pip2 install pymeasure
> Collecting pymeasure
>   Using cached PyMeasure-0.4.3.tar.gz
>     Complete output from command python setup.py egg_info:
>     Traceback (most recent call last):
>       File "<string>", line 1, in <module>
>       File "/tmp/pip-build-9TN0h5/pymeasure/setup.py", line 57, in <module>
>         long_description=open('README.rst').read() + "\n\n" + open('CHANGES.txt').read(),
>     IOError: [Errno 2] No such file or directory: 'CHANGES.txt'
>     
>     ----------------------------------------
> Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-9TN0h5/pymeasure/
> [cbrown@cbrown2 ~]$ sudo pip2 install pymeasure==0.4.2
> Collecting pymeasure==0.4.2
>   Using cached PyMeasure-0.4.2.tar.gz
>     Complete output from command python setup.py egg_info:
>     Traceback (most recent call last):
>       File "<string>", line 1, in <module>
>       File "/tmp/pip-build-62toxu/pymeasure/setup.py", line 55, in <module>
>         long_description=open('README.rst').read() + "\n\n" + open('CHANGES.txt').read(),
>     IOError: [Errno 2] No such file or directory: 'CHANGES.txt'
>     
>     ----------------------------------------
> Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-62toxu/pymeasure/
> [cbrown@cbrown2 ~]$

cjermain:
@cbrown1, thanks for reporting this. It looks like `CHANGES.txt` is not included in `MANIFEST.in`, so its not being rolled in with the package. I will fix this shortly.

I'm curious if this effects the wheel version. Do you have wheel installed?

cbrown1:
I can confirm that the wheel does install properly on python3, and also that the github version now installs on python2.

Thanks!

cjermain:
Great. PyMeasure only supports Python 3 to help [phase out Python 2.7](http://python3statement.github.io/). Are the pip issues specific to Python 2.7?

cjermain:
[...]

PR #72, opened 2017-04-04
Fix pip install by including CHANGES.txt in MANIFEST

cjermain:
This PR addresses #71, where the `CHANGES.txt` file is not found during basic `pip` install.


cjermain:
This issue only affects the non-wheel package.

PR #73, opened 2017-05-16
Update to AgilentE4980

dvspirito:
This is an update for the AgilentE4980 class:

1. Fix and change in the `freq_sweep` method:
	a. add a `try...except` as a workaround for a Timeout error occurring with long frequency lists.
	b. add an option to read and return the frequencies from the instrument (useful for debug)
	c. at the end, the trigger returns to Manual mode instead of Internal
2. Add a `trigger_source` control


cjermain:
@dvspirito, thanks for the update!

PR #81, opened 2017-05-30
Install the tests alongside the package

### LINKED ISSUE #80: Consider installing the tests
Since you decided to embed the test suite inside the package instead of leaving it outside in a common `tests` folder (both of which are totally acceptable and covered by the `pytest` guidelines), why not install the tests with the package? It would give the following benefits:

- Simplify the `setuptools` metadata, by replacing the explicit listing of the modules with a single call to `find_package`,
- Easy testing of the installed package via `python -m pytest --pyargs pymeasure`.

ghisvail:
Closes #80

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,6 @@ install:
   - conda env create -f $ENVIRONMENT_FILE
   - source activate pymeasure
   - conda info -a
-  - python setup.py install
 
 # command to run tests
 script: py.test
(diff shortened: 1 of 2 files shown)


ghisvail:
Should be ready to merge now

cjermain:
Thanks @ghisvail!

ghisvail:
Could you push a new release to PyPI with the latest changes in? Thanks.

cjermain:
Yes. I will do a minor release before the end of the weekend.

PR #82, opened 2017-06-04
Include docs to sdist

ghisvail:


cjermain:
@ghisvail, thanks for this addition. Can you include the __pycache__ folders to be excluded?

ghisvail:
> Can you include the __pycache__ folders to be excluded?

That's not the topic of this PR though. Feel free to merge this and follow it up with a commit for filtering the cache files.

PR #85, opened 2017-06-09
Bug fixes and enhancements

feinsteinben:
## General Fixes
1. Create new file with base errors.
2. Make sure imports are always at the top of the file, before any other code (also clears IDE from yellow warnings).
3. Fix pep8 compatibility (spaces, blank lines, indentations).
4. Replace mutable default values (`values=[]`) with immutable alternative (`values=()`).
6. Replace `super(...)` with the new `super()`. (see: https://github.com/ralph-group/pymeasure/pull/87#issuecomment-311655493)
7. Replace three single-quotes with three double-quotes for documentation (for consistency, and also to clear yellow warnings in IDE).
8. Replace blank `except:` statements with `except Exception:` or with more specific exceptions.
9. Replace `b <= c and b >= a` kind of conditions with `a <= b <= c`.
10. Prefer relative import for modules from the same package.
11. Fix `!= None` to `is not None`. Same for `True`, `False`.
12. Use `isinstance(...)` instead of `type(...) is ...`.
13. Remove the `force-True` argument from `set_start_method('spawn')` in pymeasure/process.py. It doesn't seem to be such an argument.

## pymeasure.adapters
1. Change the way `PrologixAdapter` and 'SerialAdapter' handle `serial.Serial` object as its adapter.
2. Inherit `__del__` from `SerialAdapter` to `PrologixAdapter` instead of rewriting it.

## pymeasure.experiments
1. Support recording results to multiple files.
3. Change `Recorder` to inherit from `logging.handlers.QueueListener` that does about the same thing. Same thing for `Scribe`. Actually, it doesn't inherit directly from `logging.handlers.QueueListener` but from `pymeasure.log.QueueListener` that includes the additional `is_alive()` method.
4. Add to `pymeasure.experiment.listeners` a class called `Monitor`, which is like `Recorder` but for console logging.
5. Rename the `data`  to `record` in `Worker.emit()` arguments, for consistency with the other modules (where it's called `record`).
6. Modify the way `Worker` handle exceptions and aborts:
   - Use specific methods `handle_exception()`, `handle_abort()`.
   - Log the event using `log.exception()` that saves the traceback automatically.
   - Handle failure in `startup()` like failure in `execute()`. The motivation for this is that `startup()` may turn on some instruments that need to be shutdown safely if the procedure fails (even if it fails in the `startup()` phase). The alternative would be to include the startup phase in the `execute()` method. However, I thought it'd be better to separate startup from execute.
7. Add `**kwargs` to `file_log()`. It's useful when you wish to use different mode (i.e. `a` instead of `w`) in the log file.
8. Use the `prepare()` method in `TopicQueueHandler` to prepare the record. This method was made exactly for this use.

## pymeasure.display
1. It's seems that `QtCore.QObject._continuous` was renamed to `QtCore.QObject._is_continuous`, so I also changed it in the relevant code.
2. Add `matplotlib` as an optional requirement.

## TCP
1. Use `cloudpickle` for serialization. It make the serialization process more straight-forward.
2. Add `zmq`, `cloudpickle` as optional requirements.

## Package and Requirements
1. Add Python3.6 to CI, and add conda-3.6.yml.
2. Add `matplotlib` and `zmq, cloudpickle` as extra requirements in the setup file (`extra_require`). TO install it with `matplotlib`, use `pip install pymeasure[matplotlib]`. To install it with TCP-support, use `pip install pymeasure[tcp]`. To do both, `pip install pymeasure[tcp,matplotlib]`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,17 @@ matrix:
     - os: linux
       env: PYTHON=3.5  CONDA_PY=35
       sudo: true
+    - os: linux
+      env: PYTHON=3.6  CONDA_PY=36
+      sudo: true
     - os: osx
       env: PYTHON=3.4  CONDA_PY=34
       sudo: false
     - os: osx
       env: PYTHON=3.5  CONDA_PY=35
+      sudo: false
+    - os: osx
+      env: PYTHON=3.6  CONDA_PY=36
       sudo: false
 
 # command to install dependencies
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,6 +25,13 @@ environment:
       #PYTHON_ARCH: "64"
       PLATFORM: x64
 
+    - PYTHON_VERSION: "3.6"
+      MINICONDA: C:\Miniconda36-x64
+      #PYTHON: "C:\\Python36-x64"
+      #DISTUTILS_USE_SDK: "1"
+      #PYTHON_ARCH: "64"
+      PLATFORM: x64
+
     # 32-bit
     - PYTHON_VERSION: "3.4"
       MINICONDA: C:\Miniconda3
@@ -34,6 +41,11 @@ environment:
     - PYTHON_VERSION: "3.5"
       MINICONDA: C:\Miniconda35
       #PYTHON: "C:\\Python35"
+      PLATFORM: x86
+
+    - PYTHON_VERSION: "3.6"
+      MINICONDA: C:\Miniconda36
+      #PYTHON: "C:\\Python36"
       PLATFORM: x86
 
 skip_tags: true
--- a/conftest.py
+++ b/conftest.py
@@ -27,4 +27,4 @@ import pytest
 
 def pytest_addoption(parser):
     parser.addoption("--runslow", action="store_true",
-        help="run slow tests")
+                     help="run slow tests")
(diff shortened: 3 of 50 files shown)


feinsteinben:
Hi @cjermain ,

I think that appveyor fails because I didn't added the conda-3.6 tests correctly (for the windows version). I'd appreciate your help in doing it.

Thanks,
Ben.

cjermain:
@feinsteinben, this is failing on Appveyor because the `setuptools 27.2.0 py36_0` package is not found. If you look at the Anaconda Cloud page for this package you will see that they don't have a Windows version that is consistent (https://anaconda.org/anaconda/setuptools/files). I recommend finding a version that is consistent and setting that in `requirements/conda-3.6.yml`. You can certainly omit the `py36_0` distinction in the `conda-3.6.yml` file, since this is directing to a specific build (#0) and may not be consistent between OSs. Overall the version of `setuptools` should not be a major issue here, so you can probably use an older version for now.

cjermain:
@feinsteinben, thanks again for all your work!

bilderbuchi:
Btw, Point 8. should also be reverted in this PR.

cjermain:
@feinsteinben, overall everything looks good! 

For future PRs, please break up each specific point into a separate PR. That will make it much faster for code-review, so we can get the changes merged quickly. It will also make it easier for people to chime in and discuss that specific point.

I like the `QueueListener` changes that take better advantage of existing logging handler code!

PR #86, opened 2017-06-09
Fix issues in pymeasure.instruments

feinsteinben:
1. Create new file with base errors.
2. Make sure imports are always at the top of the file, before any other code (also clears IDE from yellow warnings).
3. Fix pep8 compatibility (spaces, blank lines, indentations).
4. Replace mutable default values (`values=[]`) with immutable alternative (`values=()`).
5. Replace `type(<var>) is <type>` with `isinstance(<var>, <type>)`.
6. Replace `super(...)` with the new `super()`. (see: https://github.com/ralph-group/pymeasure/pull/87#issuecomment-311655493)  [previously: Replace `super(...)` with direct reference to the super-class. This is in par with the way [CPython](https://github.com/python/cpython) is written.]
7. Replace three single-quotes with three double-quotes for documentation (for consistency, and also to clear yellow warnings in IDE).
8. Replace blank `except:` statements with `except Exception:` or with more specific exceptions.
9. Replace `b <= c and b >= a` kind of conditions with `a <= b <= c`.

### DIFF
--- a/pymeasure/errors.py
+++ b/pymeasure/errors.py
@@ -22,39 +22,14 @@
 # THE SOFTWARE.
 #
 
-import visa
+
+class Error(Exception):
+    pass
 
 
-def list_resources():
-    """
-    Prints the available resources, and returns a list of VISA resource names
-    
-    .. code-block:: python
+class RangeError(Error):
+    pass
 
-        resources = list_resources()
-        #prints (e.g.)
-            #0 : GPIB0::22::INSTR : Agilent Technologies,34410A,******
-            #1 : GPIB0::26::INSTR : Keithley Instruments Inc., Model 2612, *****
-        dmm = Agilent34410(resources[0])
-    
-    """
-    rm = visa.ResourceManager()
-    instrs = rm.list_resources()
-    for n, instr in enumerate(instrs):
-        # trying to catch errors in comunication
-        try:
-            res = rm.open_resource(instr)
-            # try to avoid errors from *idn?
-            try:
-                # noinspection PyUnresolvedReferences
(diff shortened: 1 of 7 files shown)


cjermain:
@feinsteinben, please add a PR description to outline your changes and use a descriptive title that is specific to what you are accomplishing here. This is necessary for code review. Same goes for #87.

feinsteinben:
Hi @cjermain ,

Sorry for the lack of description. It's my bad.

I updated the PR description. I thought it would be good to squash all these changes into a single PR, but maybe I got the wrong decision.

Anyway, I tried to find a more specific and descriptive name for this PR, but since it includes several types of changes - it couldn't find a good name. Do you have any suggestions?

Thanks,
Ben.

PS. Thanks for this excellent package, and all the hard work you've put in writing it and maintaining it!

cjermain:
@feinsteinben, thanks for adding the descriptions. In the future, don't worry about squashing commits, since we typically do that on the merge. 

With regards to the `super` calls, I agree with @bilderbuchi in #87 that we should stick with the `super` calls instead of direct parent class calls. This is important for dealing with multi-class inheritance and mixin classes, and as a convention `super` is the preferred method to my knowledge. Please revert those changes.

Otherwise this PR looks great! Thanks for getting PyMeasure up to standard!

bilderbuchi:
Great work spotting nr. 4, btw! Mutable default values is a frequent and potentially dangerous mistake.

feinsteinben:
Thanks for the review. Squashed this change into PR  #85.
Think it can be closed.

PR #87, opened 2017-06-09
Fix issues in pymeasure.adapters

feinsteinben:
1. Make sure imports are always at the top of the file, before any other code (also clears IDE from yellow warnings).
2. Fix pep8 compatibility (spaces, blank lines, indentations).
3. Replace blank `except:` statements with `except Exception:` or with more specific exceptions.
4. Replace `super(...)` with the new `super()`. (see: https://github.com/ralph-group/pymeasure/pull/87#issuecomment-311655493)  [previously: Replace `super(...)` with direct reference to the super-class. This is in par with the way [CPython](https://github.com/python/cpython) is written.]
5. Change the way `PrologixAdapter` and 'SerialAdapter' handle `serial.Serial` object as its adapter.
6. Inherit `__del__` from `SerialAdapter` to `PrologixAdapter` instead of rewriting it.

### DIFF
--- a/pymeasure/adapters/__init__.py
+++ b/pymeasure/adapters/__init__.py
@@ -22,10 +22,11 @@
 # THE SOFTWARE.
 #
 import logging
-log = logging.getLogger(__name__)
-log.addHandler(logging.NullHandler())
 
 from .adapter import Adapter, FakeAdapter
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
 
 try:
     from pymeasure.adapters.visa import VISAAdapter
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -73,8 +73,8 @@ class Adapter(object):
         for i, result in enumerate(results):
             try:
                 results[i] = cast(result)
-            except:
-                pass # Keep as string
+            except Exception:
+                pass  # Keep as string
         return results
 
     def binary_values(self, command, header_bytes=0, dtype=np.float32):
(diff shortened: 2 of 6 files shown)


bilderbuchi:
> Replace super(...) with direct reference to the super-class. This is in par with the way CPython is written.

TBH, I have never seen the recommendation to got back from `super()` to naming the specific superclass. I wonder where you got that from?

feinsteinben:
@bilderbuchi, I read some of the source of the python standard library lately (mainly the [logging module](https://github.com/python/cpython/tree/master/Lib/logging)). I found that the direct superclass approach is used in the vast majority of the cases. The `super(...)` syntax was used only once. That gave me the impression that it's the _standard_ way of doing it. However, I might got the wrong impression. If you think the `super(...)` syntax is better, I may revert back to it.

bilderbuchi:
That's because (if you check the git blame) those portions were written 10-15 years ago, when today's incarnation of `super()` did not (I think) exist yet. "New-style" super, i.e. just using `super().somecall()` is the way to go, see e.g. https://rhettinger.wordpress.com/2011/05/26/super-considered-super/ or some other resources on google or in a current Python book.

cjermain:
@feinsteinben, same comments go  this PR along with #86. Please revert the `super` calls and I will squash and merge.

Thanks @bilderbuchi for weighing in on the `super` distinction! I agree with your thoughts.

feinsteinben:
Thanks for the review. Squashed this change into PR  #85.
Think it can be closed.

PR #95, opened 2017-07-19
Fix result logging

feinsteinben:
1. Pass a `logging.Formatter` to `logger.setFormatter` instead of a `format()` method.
2. Don't decode `topic` when it's passed to `monitor_queue`.

### DIFF
--- a/pymeasure/experiment/listeners.py
+++ b/pymeasure/experiment/listeners.py
@@ -43,7 +43,7 @@ except ImportError:
 class Monitor(QueueListener):
     def __init__(self, results, queue):
         console = StreamHandler()
-        console.setFormatter(results.format)
+        console.setFormatter(results.formatter)
 
         super().__init__(queue, console)
 
@@ -103,7 +103,7 @@ class Recorder(QueueListener):
         handlers = []
         for filename in results.data_filenames:
             fh = FileHandler(filename=filename, **kwargs)
-            fh.setFormatter(results.format)
+            fh.setFormatter(results.formatter)
             fh.setLevel(logging.NOTSET)
             handlers.append(fh)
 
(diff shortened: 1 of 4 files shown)


bilderbuchi:
I don't know the policy about this here, but I think it would make sense to add tests to the PR to make sure that the behaviour doesn't inadvertantly change later. Is that feasible?

feinsteinben:
good idea. maybe we should run the simple example as a test, and compare the output file to some ground truth. what do you think?

bilderbuchi:
yeah, that would be a rather high-level regression test, but probably a bit slow. I guess a more granular test level (e.g. here if CSVFormatter produces the expected output) would be more desirable, but I'd say whatever @cjermain wants to see is preferable. :-)

cjermain:
Thanks @feinsteinben for putting together a fix! 

@bilderbuchi I agree with you. As a new strategy let's have passing tests for all new features, aiming for full coverage of the non-instrument code for a PyMeasure 0.5 release.

I suggest the following tests for this PR:
- [ ] Check that CSVFormatter.format produces a correct row (`pymeasure/experiment/tests/test_listeners.py`)
- [ ] Check that CSVFormatter.format produces a correct header (`pymeasure/experiment/tests/test_listeners.py`)
- [ ] Check that a Worker writes 10 (or more) results to a temp file correctly through the Monitor/CSVFormatter (`pymeasure/experiment/tests/test_workers.py`)

For now put the tests in a test directory in the appropriate subpackage.

feinsteinben:
1. Changed the mechanism behind `CSVFormatter`. It should be more robust now.
2. Added tests for `CSVFormatter`. (in `pymeasure/experiment/tests/test_results.py`)

bilderbuchi:
Nice work!

cjermain:
Looks good @feinsteinben! I will merge and set up the Worker test so that we can make a minor release.

Issue #90, opened 2017-07-12
Using the plotter example does not work


bilderbuchi:
The [example code](http://pymeasure.readthedocs.io/en/latest/tutorial/graphical.html#using-the-plotter) has an oversight - SimpleProcedure should be replaced by RandomProcedure.
Even when that's corrected, I don't get any window popping up.

Similar situation for the managed window example (which btw is missing an `import sys`)

This is on pymeasure 0.4.5 installed via pip, pyqtgraph 0.10.0, pyqt 5.6.0, and Python 3.6.1 on Windows.

feinsteinben:
I think I fixed the `ValueError: Parameter value is not set` error in the managed window example.
However, now I get another error after putting an experiment in the queue:
```
Process Worker-1:
Traceback (most recent call last):
  File "/home/benfei/.local/share/miniconda3/envs/pymeasure/lib/python3.6/site-packages/zmq/sugar/socket.py", line 356, in send_multipart
    _buffer_type(msg)
TypeError: memoryview: a bytes-like object is required, not 'int'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/benfei/.local/share/miniconda3/envs/pymeasure/lib/python3.6/multiprocessing/process.py", line 249, in _bootstrap
    self.run()
  File "/home/benfei/Documents/Github/pymeasure/pymeasure/experiment/workers.py", line 160, in run
    self.update_status(Procedure.RUNNING)
  File "/home/benfei/Documents/Github/pymeasure/pymeasure/experiment/workers.py", line 116, in update_status
    self.emit('status', status)
  File "/home/benfei/Documents/Github/pymeasure/pymeasure/experiment/workers.py", line 96, in emit
    self.publisher.send_multipart(data)
  File "/home/benfei/.local/share/miniconda3/envs/pymeasure/lib/python3.6/site-packages/zmq/sugar/socket.py", line 363, in send_multipart
    i, rmsg,
TypeError: Frame 0 (128) does not support the buffer interface.
```

bilderbuchi:
When merging #95 #96 #97, the randomprocedure example works as expected I think. The managed window example works with some caveats, see https://github.com/ralph-group/pymeasure/pull/96#issuecomment-317748308

robmarkcole:
[...]

PR #96, opened 2017-07-19
Fix inputs initialization

feinsteinben:
Try to set `value` to `parameter.value`. Skip if `parameter` was not initialized yet.

Should fix the managed window example in issue #90.

### DIFF
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -41,25 +41,26 @@ class Input(QtCore.QObject):
         super().__init__()
         self._parameter = None
         self.set_parameter(parameter)
-        self.set_value(parameter)
 
     def set_parameter(self, parameter):
+        """Connects a parameter to the input box, and initializes the box value.
+
+        :param parameter: parameter to connect.
+        :return:
+        """
         self._parameter = parameter
 
-    def set_value(self, parameter):
-        if hasattr(self, 'setValue'):
-            try:
-                value = parameter.value  # if parameter is not set yet, this will raise ValueError
-            except ValueError:
-                pass  # parameter is not set yet
-            else:
-                self.setValue(value)
+        if parameter.default:
+            self.setValue(parameter.default)
+
+        if hasattr(parameter, 'units') and parameter.units:
+            self.setSuffix(" %s" % parameter.units)
 
     def update_parameter(self):
         """ Mutates the self._parameter variable to update
         its value
         """
-        pass
+        self._parameter.value = self.value()
 
     @property
     def parameter(self):
@@ -71,11 +72,17 @@ class StringInput(QtGui.QLineEdit, Input):
     def __init__(self, parameter, parent=None):
         QtGui.QLineEdit.__init__(self, parent)
         Input.__init__(self, parameter)
-        if self._parameter.default:
-            self.setText(self._parameter.default)
 
-    def update_parameter(self):
-        self._parameter.value = self.text()
+    def setValue(self, value):
+        # QtGui.QLineEdit has a setText() method instead of setValue()
+        return super().setText(value)
+
+    def setSuffix(self, value):
+        pass
+
+    def value(self):
+        # QtGui.QLineEdit has a text() method instead of value()
+        return super().text()
 
 
 class FloatInput(QtGui.QDoubleSpinBox, Input):
@@ -85,13 +92,6 @@ class FloatInput(QtGui.QDoubleSpinBox, Input):
         self.setMinimum(self._parameter.minimum)
         self.setMaximum(self._parameter.maximum)
         self.setButtonSymbols(QtGui.QAbstractSpinBox.NoButtons)
-        if self._parameter.default:
-            self.setValue(self._parameter.default)
-        if self._parameter.units:
-            self.setSuffix(" %s" % self._parameter.units)
-
-    def update_parameter(self):
-        self._parameter.value = self.value()
 
 
 class IntegerInput(QtGui.QSpinBox, Input):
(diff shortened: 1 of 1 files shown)


pymeasure/display/inputs.py
```diff
@@ -41,10 +41,20 @@ def __init__(self, parameter):
...
     def set_parameter(self, parameter):
         self._parameter = parameter
 
+    def set_value(self, parameter):
+        if hasattr(self, 'setValue'):
+            try:
+                value = parameter.value  # if parameter is not set yet, this will raise ValueError
+            except ValueError:
+                pass  # parameter is not set yet
+            else:
```
feinsteinben: 1. The `StringInput` that inherits from `QtGui.QLineEdit` doesn't have a `setValue()` method.
2. Maybe a better name for that method would be `init_value()`.
3. I didn't want to catch an `ValueError` inside the `setValue()` call. That's why it was separated. I'll try to do it less obscure.

pymeasure/display/inputs.py
```diff
@@ -41,10 +41,20 @@ def __init__(self, parameter):
...
 
     def set_parameter(self, parameter):
         self._parameter = parameter
 
+    def set_value(self, parameter):
+        if hasattr(self, 'setValue'):
+            try:
+                value = parameter.value  # if parameter is not set yet, this will raise ValueError
+            except ValueError:
+                pass  # parameter is not set yet
```
feinsteinben: Changed the way this is done. Now it uses `parameter.default` for initialization.

bilderbuchi: great :-)

pymeasure/display/inputs.py
bilderbuchi: OK.
ad 3) that's actually a good reason to do that with an else. I didn't know if that was a relevant scenario, though, that's why I asked.

bilderbuchi:
When I am testing this PR (and nr 95), I now get a managed window, but 
* the queued procedures don't do anything (just stay "queued").
* Hitting Open results in

```
  File "<snip>\pymeasure\pymeasure\display\widgets.py", line 381, in _setup_ui
    self.layout().addWidget(preview_tab, 0, 5, 4, 1)

AttributeError: 'NoneType' object has no attribute 'addWidget'
```
* Also, to run at all the managed window example misses an `import sys` and an `import tempfile`.

feinsteinben:
did you run it after patching #95, #96 and #97? they are all needed for the
managed window example.

on what platform did you run the example? I'll try to run the latest
version on my pc (Linux, x64, python 3.6.1 via anaconda).

bilderbuchi:
Yes, Windows 10, Python 3.6. via anaconda.

feinsteinben:
Did the following:
```
git checkout master
git pull upstream master
git checkout -b test_959697
git merge fix_log
git merge fix_inputs
git merge fix_serialization
```
Then run this:
```python
import logging
import random
import sys
import tempfile
from time import sleep

from pymeasure.display.Qt import QtGui

    iterations = IntegerParameter('Loop Iterations')

    DATA_COLUMNS = ['Iteration', 'Random Number']

    def startup(self):
        random.seed(self.seed)

    def execute(self):
            data = {
                'Iteration': i,
                'Random Number': random.random()
            }
            self.emit('results', data)
            sleep(self.delay)
            if self.should_stop():
                break


    def __init__(self):
        super(MainWindow, self).__init__(
            inputs=['iterations', 'delay', 'seed'],
            displays=['iterations', 'delay', 'seed'],
            x_axis='Iteration',
            y_axis='Random Number'
        )

    def queue(self):
        filename = tempfile.mktemp()

if __name__ == "__main__":
    app = QtGui.QApplication(sys.argv)
    window = MainWindow()
    window.show()
    sys.exit(app.exec_())
```
It seems to work:
pymeasure_test_959697

The exact packages are
[freeze_959697.yml.txt](https://github.com/ralph-group/pymeasure/files/1174416/freeze_959697.yml.txt)

feinsteinben:
Can you try to install the environment from the attached file, and check whether it works?

bilderbuchi:
so, what happens if you hit "Open"? Also, in Spyder, your kernel does not die?

bilderbuchi:
Yes, I can try this, but probably only tomorrow.

bilderbuchi:
OK after adapting your instructions to work correctly, I ran the code in spyder, now on my Ubuntu 16.04 home machine. 
On first run the window opens, I can queue procedures, but they never start. Open button works correctly here, though (at least opens a file dialog). Running a second time after closing the GUI gives the same "kernel died" message.

feinsteinben:
very strange.
may you please include the result of `conda env export`?

bilderbuchi:
OK, your env file is not even installable on the Windows machine because of lots of missing packages, `NoPackagesFoundError: Package missing in current win-64 channels` for 
```
- cairo 1.14.8 0
- curl=7.52.1=0
- dbus=1.10.10=0
- fontconfig 2.12.1 3
etc...
```
I'll send the result for conda env export on the linux machine later.

bilderbuchi:
OK, so now, on the Linux machine, after starting your example a second time in spyder, I get
```
/home/bilderbuchi/anaconda3/envs/pymeasure/lib/python3.6/multiprocessing/semaphore_tracker.py:129: UserWarning: semaphore_tracker: There appear to be 28 leaked semaphores to clean up at shutdown
len(cache))
```
I'm officially thoroughly confused on the repro for these issues, I suspect some pyqt or whatever dark fairies spitting into the soup. 
If you can't repro any of these issues I'm seeing at all, I just have to conclude that it's on my end (although it would be good if someone else could give this a spin, too... @cjermain?). Thanks in any case for the follow-up!

Just for completeness, here's my `conda env export` output: 
[my_conda_export.txt](https://github.com/ralph-group/pymeasure/files/1177679/my_conda_export.txt)

feinsteinben:
1. Clicking on **Open** opens a file dialog. Everything continue to work as expected. (didn't try to open a file). 
pymeasure-managed_window-open

2. After installing an environment based on your yaml file, it still works as expected.
3. +1 for the dark [fairies](https://arxiv.org/abs/1204.0492) hypothesis.

bilderbuchi:
did you try running the example twice? i.e. run, close it, run again (F5 in spyder)?

feinsteinben:
I use pycharm to run it.
I ran it in many ways: one-after-another, two instances at the same time,
etc. Always worked as expected.

bilderbuchi:
Hm, yeah, I can try with pycharm, too. Thanks.

bilderbuchi:
Agreed, see #105

cjermain:
Nice work @feinsteinben and @bilderbuchi! 

I am planning to work on this and merge things during the weekend, and ship a new minor release.

PR #97, opened 2017-07-19
Fix zmq's send/receive operations

feinsteinben:
Replace `send_multipart` with `send_serialized`, and `recv_multipart` with `recv_serialized`.

It solves the error encountered in the [managed window example](https://github.com/ralph-group/pymeasure/issues/90#issuecomment-316546569) in issue #90.

### DIFF
--- a/pymeasure/display/listeners.py
+++ b/pymeasure/display/listeners.py
@@ -71,8 +71,7 @@ class QListener(StoppableQThread):
         self.timeout = timeout
 
     def receive(self, flags=0):
-        data = self.subscriber.recv_multipart(flags=flags)
-        topic, record = cloudpickle.loads(data)
+        topic, record = self.subscriber.recv_serialized(deserialize=cloudpickle.loads, flags=flags)
         return topic, record
 
     def message_waiting(self):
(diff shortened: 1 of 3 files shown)


bilderbuchi:
I don't know the policy about this here, but I think it would make sense to add tests to the PR to make sure we don't run into the same bug later. Is that feasible?

feinsteinben:
@bilderbuchi it's a good idea, however I don't know how to test such a complex mechanism (some kind of integration tests). maybe with mocks?

bilderbuchi:
Yeah, that would probably be necessary. Maybe you can "just" mock out file I/O (which slows things down). As soon as you touch GUI-related things, I guess it becomes complex, so I'd be hesitant to start with that. Maybe you can isolate the detail that broke in #90, and test for the fact that it's not broken anymore. This is more or less what test-driven development is about: encounter a bug, write a test that shows the bug, fix the bug, run the new test to confirm the fix, run all the tests to confirm you haven't broken anything else (tested) in the process.

feinsteinben:
maybe it's possible to write down a simple unittest that checks only the `emit()` and `receive()` methods.
however, I won't be able to do it before august.

bilderbuchi:
This seems to mostly work with #95 #96 #97 merged, see https://github.com/ralph-group/pymeasure/pull/96#issuecomment-317748308

Issue #92, opened 2017-07-13
It's not possible to test locally installed pymeasure


bilderbuchi:
Today I ran `py.test` on my locally installed pymeasure. This breaks with 
```
______________ ERROR collecting experiment/tests/test_workers.py ______________
..\_pytest\config.py:1213: in getoption
    val = getattr(self.option, name)
E   AttributeError: 'CmdOptions' object has no attribute '--runslow'

During handling of the above exception, another exception occurred:
experiment\tests\test_workers.py:39: in <module>
    not pytest.config.getoption("--runslow"),
..\_pytest\config.py:1223: in getoption
    raise ValueError("no option named %r" % (name,))
E   ValueError: no option named '--runslow'
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!
```
because the `conf.py` file is not packaged along with the tests. This makes the `getoption` bail because the option has not been added in the conf.py.

This is not visible on CI because you run the tests on the local pymeasure repo directory, instead of an installed version, which might be a better idea because it reflects your users' situation better. That should probably be changed.

The easiest fix for the present issue that I could see, since you apparently decided to use the test structure which [distributes tests with your application](https://docs.pytest.org/en/latest/goodpractices.html#tests-as-part-of-application-code), is to move the conftest.py into the `pymeasure` directory to distribute it with the rest of the tests. 
It's still possible to test a locally available, not installed, repo clone by calling pytest from the `./pymeasure` directory, not the repo root dir.

cjermain:
@bilderbuchi, I can reproduce this. I am working on a branch to overhaul the tests and will address this.

bilderbuchi:
Thanks!

bilderbuchi:
I know that finding the proper test structure do cover all usecases can be really tricky.

Issue #105, opened 2017-07-27
Leaked semaphores?


bilderbuchi:
When running the managed window example in Spyder, closing the window, and running it again, I get a crash of the consol kernel. This does not happen in pycharm. @feinsteinben can also reproduce this.
Sometimes I noticed this error message after closing for the first time:
```
/home/bilderbuchi/anaconda3/envs/pymeasure/lib/python3.6/multiprocessing/semaphore_tracker.py:129: UserWarning: semaphore_tracker: There appear to be 28 leaked semaphores to clean up at shutdown
```
and some userwarning about how the application should be shutdown with ctrl-q and some other key combos.

cjermain:
I am also getting this after merging #95, #96, #97.

```
Traceback (most recent call last):
  File "/home/colin/.miniconda3/envs/pymeasure/lib/python3.5/multiprocessing/util.py", line 253, in _run_finalizers
    finalizer()
  File "/home/colin/.miniconda3/envs/pymeasure/lib/python3.5/multiprocessing/util.py", line 185, in __call__
    res = self._callback(*self._args, **self._kwargs)
  File "/home/colin/.miniconda3/envs/pymeasure/lib/python3.5/multiprocessing/synchronize.py", line 87, in _cleanup
    from .semaphore_tracker import unregister
SystemError: Parent module 'multiprocessing' not loaded, cannot perform relative import
```

cjermain:
This error goes away using `fork` instead of `spawn`, but it's not clear yet whether this is a Python or PyMeasure bug.

cjermain:
This looks like a CPython bug, but have not gotten to the bottom of it completely. It only occurs on Unix, and only with `spawn`.

bilderbuchi:
I can't reproduce this anymore, on Ubuntu, python 3.9, Spyder 4.2.3, so I'll close this. Feel free to re-open with a precise repro procedure should you encounter this again.

Issue #98, opened 2017-07-20
Add tests for all non-instrument features


bilderbuchi:
Migrated from TODO

cjermain:
This is currently a top priority for the next major release, PyMeasure 0.5. I plan to do some reorganization of the test, but would welcome any contributed PRs to improve the test coverage.

bilderbuchi:
There is a connection to #102 here, as the doctest facilities of pytest could be used to test some of the examples in the docs. Also, doctests in the pymeasure code docstrings could be a good way to get some coverage increases going, while at the same time improving documentation.
Pytest can [consume](https://docs.pytest.org/en/latest/doctest.html) doctests, both in .py and .rst files. Unfortunately, some of the features of Sphinx doctests [are not supported yet](https://github.com/pytest-dev/pytest/issues/758) by pytest, so more complex examples will probably not be reasonably testable.
Sphinx doctest functionality [is quite featureful](http://www.sphinx-doc.org/en/stable/ext/doctest.html), but to keep compatibility with pytest as much as possible, [the simple doctest format](http://www.sphinx-doc.org/en/stable/ext/doctest.html#confval-doctest_test_doctest_blocks), i.e.  separate rst pararagraphs starting with `>>>`, are preferable.

These are more or less research notes, I'll try to poke at seeing if we can integrate what's in the docs already if I find the sparetime during the summer.

bilderbuchi:
doctests for the documentation have been added in #111.

PR #106, opened 2017-07-31
Enhancing unit tests

cjermain:
This PR enhances the unit tests available to PyMeasure on the non-instruments code. It seeks to solve #92 and ensure that example code is tested in CI.


cjermain:
This is a work in progress, with more needed on the `pymeasure.experiment` tests and tests to ensure that the examples run without errors. I plan to release a new minor release when this is finished.

bilderbuchi:
:+1:

bilderbuchi:
One thing that could be useful in increasing test coverage would be to add `sphinx doctest` to the CI scripts, and adapt code in the docs to run as doctests. This does not always make sense, due to this package's close connection with hardware. However, in the [adding instruments](http://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html) section, which is nearly testable already, it could be feasible to redefine the "Extreme 5000" as a `FakeInstrument` under the hood (non-user-visible) using a [test-setup](http://www.sphinx-doc.org/en/stable/ext/doctest.html#directive-testsetup) directive. This way, conceivably the whole section could be run as a doctest.
Also, see https://github.com/ralph-group/pymeasure/issues/98#issuecomment-317273099.

cjermain:
@bilderbuchi, I like the Sphinx doctest idea. Let's hold off on it for another PR and continue discussion in #98.

bilderbuchi:
Sure :-)

cjermain:
I've removed setting `spawn`, to allow Unix platforms to use `fork`. This solves #105, which were being caused by an issue in the [semaphore cleanup method](https://github.com/python/cpython/blob/48350412b70c76fa51f488cfc736c80d59b5e8eb/Lib/multiprocessing/synchronize.py#L87). This branch needs to confirm that results are being written to the temp file in `test_worker_finish` (looks like this is failing on Windows and needs debugging). I would also like to add better coverage to the `Results` class before merging. Still a work in progress...

PR #110, opened 2017-08-27
Documentation fixes

bilderbuchi:
This cleans up some of the warnings that `make html` throws when generating the docs.

Note that many warnings remain where Sphinx detects identically-named attributes of different classes exist (e.g. `current`), 
```
pymeasure/pymeasure/instruments/keithley/keithley2000.py:docstring of pymeasure.instruments.keithley.Keithley2000.acquire_reference:: WARNING: more than one target found for cross-reference 'mode': pymeasure.instruments.agilent.AgilentE4980.mode, pymeasure.instruments.keithley.Keithley2000.mode
```
and sometimes picks the wrong one, e.g. `mode` [here](http://pymeasure.readthedocs.io/en/latest/api/instruments/keithley/keithley2000.html#pymeasure.instruments.keithley.Keithley2000.acquire_reference).
This is connected to sphinx-doc/sphinx#2549, I think.

Note that I put an import guard around the pycomedi code, as this otherwise throws exceptions when autodoc crawls through the code.

PR #111, opened 2017-08-27
Add doctests

bilderbuchi:
I added the ability to run Sphinx doctests to the documentation. This also necessarily include the changes in #110, so it would be best to merge that one first, to get a smaller diff here.

To check this out. go to `doc` and run `make doctest`. Currently this executes and checks the code in the "Adding instruments" section of the docs, which lent themselves most to a first shot at this.

In the process, I extended ` FakeInstrument` a bit. Pytest tests still all pass.

@cjermain please advise how you want this integrated into the CI runners. I would just run this in addition to the pytest tests, but then we need to install sphinx in all the builds, too. Alternatively, you could have one extra builder for generating docs. This might be useful anyway elsewhere - how do you update the deployed documentation?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ matrix:
     - os: linux
       env: PYTHON=3.6  CONDA_PY=36
       sudo: true
+    - os: linux
+      env: PYTHON=3.6 DOCS=true
+      sudo: true
     - os: osx
       env: PYTHON=3.4  CONDA_PY=34
       sudo: false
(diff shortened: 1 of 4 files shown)


cjermain:
@bilderbuchi, this is a great step forward!

The docs are automatically updated on release to ReadTheDocs. Is it worth setting up CircleCI to run the doctests as a separate CI test? Or should we just modify the TravisCI tests to include this as well?

bilderbuchi:
I see. I wouldn't bother with another ci service, but modify the Travis setup.

bilderbuchi:
@cjermain I rebased for a better diff and added a docs travis job, please review the changes. :-)

bilderbuchi:
Btw, are the various requirements.txt and complex travis.yml really necessary? If you look at InstrumentKit, [those](https://github.com/Galvant/InstrumentKit/blob/master/requirements.txt) are [both](https://github.com/Galvant/InstrumentKit/blob/master/.travis.yml) a good deal simpler.

cjermain:
@bilderbuchi, nice work!

With regards to the requirements.txt file and complex travis.yml, we are using the Conda environment. This is the easiest way for people to get Qt up and running, and in the future we can build in Qt unit tests. If we use pip the Qt packages would more difficult to install.

bilderbuchi:
Sure,  I'm behind continuing to use conda, I'm just wondering if we could get away with one file instead of one for every python version we test.

Issue #114, opened 2017-09-19
Tests for 32 bit builds


cjermain:
Currently we are running unit tests for:

- Windows 32 bit: Python 3.4, 3.5, and 3.6
- Windows 64 bit: Python 3.4, 3.5, and 3.6
- Linux 64 bit: Python 3.4, 3.5, and 3.6
- macOS 64 bit: Python 3.4, 3.5, and 3.6

I suggest that we drop support for 32 bit Windows to reduce the test run time. 

Any thoughts?

bilderbuchi:
yeah why not. Are the windows builds running that much longer than the Travis builds?

cjermain:
@bilderbuchi, great question. The travis builds are doing both Linux and macOS. I was thinking to break them up, and use CircleCI to run the Linux builds.

bilderbuchi:
With the amount of activity this report is generally seeing, is it worth the hassle of supporting 3 different CI configurations to shave a single-digit number (?) of minutes from the testing time? Or do you expect the amount of testing, both in length and frequency, to increase substantially in the future?

bilderbuchi:
Report -> repo, thank you phone...

cjermain:
@bilderbuchi, that is a good point. Overall, I want to make sure that the CI does not get in the way, and that unit testing is quick. For now, lets leave it as is and keep this issue open if people have other thoughts.

cjermain:
It looks like the main delay for the CI tests are actually the macOS builds. TravisCI has a fairly consistent backlog of macOS builds in their status (https://www.traviscistatus.com/).

PR #119, opened 2017-11-05
pymeasure.display.inputs fixes and additions

### LINKED ISSUE #115: ManagedWindow GUI: FloatParameter doesn't accept negative Values
Hi,
another issue, I found using the `ManagedWindow`:
Every time a `FloatParameter `is to be set to a negative value, it is set to "0" instead. IntegerParameters can be set to negative Values but the default value on start-up is also reset to "0".
I suspect this has something to do with QtValidators ?!

To clarify:
`
Voltage = FloatParameter('Voltage', units='V', default=-1)
`

leads to `value=-1` before `setup_gui` and `value=0` after `setup_gui`. Changeing it in the InputBox of `ManagedWindow` also resets all negative values to 0.

`Voltage = IntegerParameter('Voltage', units='V', default=-1)`

leads to the same behaviour as above, except now `Parameter.value` sets correctly when the value in the InputBox is changed.

I hope this is understandable - let me know if you need further clarification.

### LINKED ISSUE #118: ScientificInput, used for FloatParameter GUI, does not correctly set minimum/maximum
A ScientificInput GUI element used for a FloatParameter in a Procedure does not correctly set the minimum and maximum attributes on the GUI element. Instead, they are set to the QtDoubleSpinBox default of 0.00 and 99.99.

For example, in the code example below, phase will only be settable in the range 0 to 99.99 inclusively, despite the `minimum`/`maximum` attributes set on the Parameter. The GUI default will be set to 99.99.

Using PyMeasure ce406284e0d0fc69bf90c0897a5a04784ddd275f.

```python
import logging
import sys

from time import sleep, perf_counter

from pymeasure.display.Qt import QtGui
from pymeasure.experiment import IntegerParameter, FloatParameter, \
                                 BooleanParameter, Parameter

class AcFreqProcedure(Procedure):
    # ...
    phase = FloatParameter(
        'Phase', units='°', minimum=-180.0, maximum=+180.0, default=165.0)

    DATA_COLUMNS = ['Time (s)', 'Magnitude (VRMS)', 'Phase (deg)']
    # ...

    def __init__(self):
        super(MainWindow, self).__init__(
            procedure_class=AcFreqProcedure,
            inputs=['phase'],
            displays=['phase'],
            x_axis='Time (s)',
            y_axis='Magnitude (VRMS)'
        )
        self.setWindowTitle('AC Single-Frequency Measurement')

    def queue(self):
        filename = "acfreq.csv"


if __name__ == "__main__":
    app = QtGui.QApplication(sys.argv)
    window = MainWindow()
    window.show()
    sys.exit(app.exec_())
```

Laogeodritt:
I've made a few changes to the inputs module as follows:

* Fixed #118 in ScientificInput (may also fix #115)
* Implemented BooleanInput and ListInput, which were TODOs in the code (and added functionality to Parameters as needed to support it)
* Inputs classes documentation improvements
* Misc: Fixed missing imports in the GUI example in documentation, fixed a misc typo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,15 @@ install:
     fi
   - conda info -a
 
+# test environment setup
+before_script:
+  # GUI testing on Linux
+  - 'if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        export DISPLAY=:99.0;
+        sh -e /etc/init.d/xvfb start;
+        sleep 3;
+    fi'
+
 # command to run tests
 script: 
   - python setup.py test
--- a/docs/tutorial/graphical.rst
+++ b/docs/tutorial/graphical.rst
@@ -98,6 +98,8 @@ Below we adapt our previous example to use a ManagedWindow. ::
     log = logging.getLogger(__name__)
     log.addHandler(logging.NullHandler())
 
+    import sys
+    import tempfile
     import random
     from time import sleep
     from pymeasure.log import console_log
--- a/pymeasure/display/Qt.py
+++ b/pymeasure/display/Qt.py
@@ -48,3 +48,13 @@ def fromUi(*args, **kwargs):
         if isinstance(element, QtGui.QWidget):
             setattr(widget, name, element)
     return widget
+
+def qt_min_version(major, minor=0):
+    """
+    Check for a minimum Qt version. For example, to check for version 4.11
+    or later, call ``check_qt_version(4, 11)``.
+
+    :return bool: True if PyQt version >= min_version
+    """
+    return (QtCore.QT_VERSION >= ((major << 16) + (minor << 8)))
+
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -55,7 +55,7 @@ class Input(QtCore.QObject):
         """
         self._parameter = parameter
 
-        if parameter.default:
+        if parameter.default is not None:
             self.setValue(parameter.default)
 
         if hasattr(parameter, 'units') and parameter.units:
@@ -168,7 +168,7 @@ class ListInput(QtGui.QComboBox, Input):
     """
     def __init__(self, parameter, parent=None):
         self._stringChoices = None
-        QtGui.QCheckBox.__init__(self, parent)
+        QtGui.QComboBox.__init__(self, parent)
         Input.__init__(self, parameter)
         self.setEditable(False)
 
@@ -180,7 +180,12 @@ class ListInput(QtGui.QComboBox, Input):
             self._stringChoices = tuple()
         super().set_parameter(parameter)
         self.clear()
-        self.insertItems(0, self._stringChoices)
+        self.addItems(self._stringChoices)
+
+        # can't be set in super().set_parameter: addItems not yet called there
+        if parameter.default is not None:
+            self.setValue(parameter.default)
+
 
     def setValue(self, value):
         try:
@@ -188,7 +193,7 @@ class ListInput(QtGui.QComboBox, Input):
             self.setCurrentIndex(index)
         except (TypeError, ValueError) as e: # no choices or choice invalid
             raise ValueError("Invalid choice for parameter. "
-                             "Must be one of %s" % str(self._choices)) from e
+                             "Must be one of %s" % str(self._parameter.choices)) from e
 
     def setSuffix(self, value):
         self._stringChoices = tuple(choice + str(value) for choice in self._stringChoices)
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -33,8 +33,8 @@ log.addHandler(logging.NullHandler())
 
 
 class Input(QtCore.QObject):
-    """ Takes a Parameter object in the constructor and has a
-    parameter method
+    """
+    Takes a Parameter object in the constructor and has a parameter method
     """
 
     def __init__(self, parameter):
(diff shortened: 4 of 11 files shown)


Laogeodritt:
Hmm, not sure what to make of the AppVeyor build failure, seems to be a dependency problem? Not sure what `vc` is, but I don't think anything I've done should've affected dependencies or AppVeyor config... (I don't know AppVeyor at all though>)

```
Fetching package metadata: ....
Solving package specifications: 
Error: Could not find some dependencies for conda: vc 9.*, vc 10.*, vc 14.*
You can search for this package on anaconda.org with
    anaconda search -t conda vc 14.*
 (and similarly for the other packages)
You may need to install the anaconda-client command line client with
    conda install anaconda-client
Command exited with code 1
```

cjermain:
@Laogeodritt, thanks for implementing these features!

With regards to the Appveyor issue, it looks like we need to drop support for Python 3.4. I'm testing this in #121. Once that is fixed, please merge from the master branch.

Please add unit tests to ensure that your new features work as expected (and never break in the future):
- ScientificInput, ensuring #118 is solved
- Basic tests for ScientificInput, BooleanInput, and ListInput to ensure values are set properly, etc.

Laogeodritt:
@cjermain Merged your latest changes and added unit tests as requested. Also fixed a few bugs with ListInput I discovered while implementing the tests.

These unit tests add `pytest-qt` as a dependency in order to be able to test/instantiate Qt widgets. This appears to be the first test of GUI classes in PyMeasure. (I don't do much user input testing/simulation of the widgets as most of the functionality is part of the widget parent class, but some Qt set-up that `pytest-qt` handles is needed to instantiate the Qt widgets anyway.)

Laogeodritt:
Ah, lovely. Tests fail/Qt segfaults on CI tests that pass with flying colours on my system.

Laogeodritt:
Fixed the consistent Linux failures (xvfb needed to be running since it's GUI-based).

Still getting some inconsistent segfaults on Travis (earlier, hasn't happened for 4-5 runs - possibly adding pytest-qt version explicitly fixed it?) and AppVeyor (see last results, haven't run it for myself yet). Not sure why Qt is segfaulting.

cjermain:
Awesome work so far @Laogeodritt! The errors are appearing on Windows 32 bit -- I will look into this further.

cjermain:
Running `python setup.py test` on this branch in Windows 32 bit, I find Python crashes on `test_inputs.py`.

```
Problem signature:
  Problem Event Name:	APPCRASH
  Application Name:	python.exe
  Application Version:	3.5.2150.1013
  Application Timestamp:	577be458
  Fault Module Name:	sip.pyd
  Fault Module Version:	0.0.0.0
  Fault Module Timestamp:	563e838b
  Exception Code:	c0000005
  Exception Offset:	0000220f
  OS Version:	6.1.7600.2.0.0.256.48
  Locale ID:	1033
  Additional Information 1:	0a9e
  Additional Information 2:	0a9e372d3b4ad19135b953a78882e789
  Additional Information 3:	0a9e
  Additional Information 4:	0a9e372d3b4ad19135b953a78882e789
```

I have thought about dropping Windows 32 #114, which has the least [downloads on conda-forge](https://anaconda.org/conda-forge/pymeasure/files). However, it would be nice to determine the actual solution to this issue.

cjermain:
@ddale, do you have any experience working with PyQt on Windows 32 bit?

cjermain:
I boiled down one of the tests and am still getting the crash, with all other tests removed on Windows 32 bit. The error message is pointing to `sip.pyd`, which is used in the binding of Qt with Python.

```python
def test_qt(qtbot):
    bool_param = BooleanParameter('potato', default=True)
    bool_input = BooleanInput(bool_param)
    qtbot.addWidget(bool_input)

    # test
    assert bool_input.text() == bool_param.name
    assert bool_input.value()
```

cjermain:
@Laogeodritt, I've narrowed down the errors to the mixin-class `Input`. Apparently the Windows 32-bit `sip.pyd` does not support mixin-classes properly. Since the mixin-classes have worked on Windows 32-bit before, I suspect this issue is with `pytest-qt` (where `qtbot.addWidget` causes the failure).

I suggest we skip the `pytest-qt` tests on Windows 32-bit. Perhaps we can add a decorator to the pytest functions to ignore them in this case.

What do you think?

Laogeodritt:
@cjermain After you mentioned the issues coming down to the mixin class, I noticed that Input inherits from QObject but doesn't do anything Qt-related (e.g. signals/slots), but only has pure Python functionality.

In consequence, I've refactored Input to be a  pure Python mixin class (3caef7b), which appears to resolve the issue on all platforms. I've also manually checked that the RandomProcedure example still runs in Qt4 and Qt5 on my system, given lack of extensive GUI tests currently.

Along with fixing this issue, I believe this would ultimately help simplify the `inputs` module, by reducing the complexity of the relationship to Qt.

However, this does introduce one bit of complexity, in that due to the support for [cooperative multi-inheritance in PyQt5](http://pyqt.sourceforge.net/Docs/PyQt5/multiinheritance.html) which doesn't exist in PyQt4, the way you call superclass init has to be different in PyQt4 and PyQt5 when you're using multiple inheritance with classes that take required `__init__` parameters. Basically, the problem is that PyQt5 will call `super().__init__(**kwargs)` in its own widgets' constructors, but PyQt4 does not. (As it was, I think there was already an unnoticed bug in that with PyQt5, `Input.__init__(self, parameter)` was being called twice - once by PyQt5 via the widget's init, once by the Input subclass).

Please see 3caef7b for my solution to this, too.

What do you think? Can you think of any problems to changing Input like this? I'm not very familiar with PyQt or Qt, so I could very well be missing something.

cjermain:
@Laogeodritt, nice work! I think that the solution you developed is reasonable for supporting both PyQt4 and PyQt5. I agree that our previous implementation was probably calling the Input constructor twice in PyQt5 -- this is why it's great to start getting some unit tests on the Qt side.

PR #124, opened 2017-11-17
Added Support for Keithley2600 

JuMaD:
**Not ready to merge**

Added support for Keithley 26XX - Specifically for the 2635A and 2635B. Basic functions are implemented and tested with the Device. However, a function wait_for_buffer - to get data from buffer AFTER measurement is complete is still missing. The workaround in Keithley/Buffer.py is not working for factory scripts since the instrument is not responding to SCPI during execution of factory / TSP scripts. Therefore, the continuous promting is timed out. 

I want to try to implement this function using SRQ (e.g. set ServiceRegisterEnable bit for buffer_ready and then use serial poll to wait for SRQ) - If anyone has experience with that or has even done this before - help is much appreciated.


cjermain:
It seems that this PR is stale. I will close it for now, and am happy to re-open.

PR #126, opened 2017-11-20
Exposing pyqtgraph.PlotItem API for plot display customisation

### LINKED ISSUE #120: Logarithmic and semilog plotting using ManagedWindow
Is there presently any way of showing log-log or semi-log plots using ManagedWindow? Looks like [pyqtgraph supports it](http://www.pyqtgraph.org/documentation/graphicsItems/plotitem.html#pyqtgraph.PlotItem.setLogMode) but I'm not seeing those facilities called in PyMeasure at the moment.

Main use case for this would be for multi-decade frequency sweeps (x-axis) and plotting power data logarithmically when we're obtaining linear units from an instrument.

Laogeodritt:
Fixes #120. These changes expose convenient, documented means for a `pymeasure` user to customise the plot display, e.g. using logarithmic axes, setting axis ranges, etc.

This pull request introduces new APIs, and does not introduce breaking changes.

* `ManagedWindow.plot` and `PlotterWindow.plot` attributes are now documented, public APIs.
* Added `ManagedWindow.setup_plot()` and `Plotter.setup_plot()`: these methods are NOOP and can be overridden in sub-classes or monkey-patched to customise the plot display. It's necessary for the Plotter (as the window is created in a separate process), and provided in ManagedWindow as a convenience method to better organise code (as otherwise it would be possible to set it up via `ManagedWindow.plot` after object construction).
* Added tests for the above.
* Added feature documentation in the tutorial and API docs.

### DIFF
--- a/docs/tutorial/graphical.rst
+++ b/docs/tutorial/graphical.rst
@@ -4,6 +4,8 @@ Using a graphical interface
 
 In the previous tutorial we measured the IV characteristic of a sample to show how we can set up a simple experiment in PyMeasure. The real power of PyMeasure comes when we also use the graphical tools that are included to turn our simple example into a full-flegged user interface.
 
+.. _tutorial-plotterwindow:
+
 Using the Plotter
 ~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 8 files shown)


cjermain:
Looks good! Thanks @Laogeodritt!

Issue #109, opened 2017-08-25
Forking/Pickling issue when worker.start() is called?!


JuMaD:
Using the ManagedWindow class, I can add experiments to the queue. However, once they should "start" (i.e. the next() method is used) the whole .py file gets interpreted again instead.

In the debugger I traced it down to a new thread being started when the worker is started. When Manager.next._worker.start() is called). Due to the error message below my best guess is, that it has something to do with the forking/pickling issue(s). Since I have no clue about multi-threading, I hope someone can help me fix this. 

![image](https://user-images.githubusercontent.com/13823675/29722901-df3fba50-89c2-11e7-9f97-4c32b699a961.png)

cjermain:
@JuMaD, thanks for reporting this issue!

Can you provide more context? Which OS are you using? Are you running one of the examples?

JuMaD:
I'm working on Windows 7/10. The problem quite literally popped-up in an own program, but I tested it with the ManagedWindow example (which has a typo in the docs btw ;) ) where it is also an issue.

cjermain:
Which version of PyMeasure are you running? Version 0.4.5 had some regressions concerning the Worker behavior that we didn't catch before the release.

JuMaD:
I'm running the latest version with some minor adjustments (just expanded the GUI of ManagedWindow with a list to pick an instrument and starting the queue only on Button click [Disabled start_on_add and connected the button to "resume"]).

Everything else is running just fine. But as I said: When I debug, I can see that "starting" the worker leads to a new thread instead.

bilderbuchi:
It's probably helpful for the maintainers if you paste the code in question, or a minimum reproducible example and the output you get, here instead of making them search the changes you could mean. ;-)

JuMaD:
Ok then. Here's the (adjusted) example from the docs

`
print('Running')
class RandomProcedure(Procedure):
    iterations = IntegerParameter('Loop Iterations')
    delay = FloatParameter('Delay Time', units='s', default=0.2)
    seed = Parameter('Random Seed', default='12345')

    DATA_COLUMNS = ['Iteration', 'Random Number']

    def startup(self):
        log.info("Setting the seed of the random number generator")
        random.seed(self.seed)

    def execute(self):
        log.info("Starting the loop of %d iterations" % self.iterations)
        for i in range(self.iterations):
            data = {
                'Iteration': i,
                'Random Number': random.random()
            }
            self.emit('results', data)
[...]

PR #127, opened 2017-11-22
Update to Threads in Plotter

dvspirito:
As discussed in  #109 , this fixes an issue with handling of multiprocesses in Plotter class.


cjermain:
Thanks @dvspirito!

PR #131, opened 2018-03-17
 Add set_process to Instrument.setting. 

### LINKED ISSUE #128: Instrument.setting does not have set_process
`Instrument.control` has `get_process` and `set_process`. `Instrument.measurement` has `get_process`. Why does `Instrument.setting` not have `set_process`? 

I think it would be useful, to e.g. cast truthy/falsy values to `1`or `0`, respectively. I know that I can (mis)use `validator` by passing an appropriate lambda, but it feels kludgy.

bilderbuchi:
This fixes #128. Finally gotten around to do this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+.pytest_cache
 
 # Translations
 *.mo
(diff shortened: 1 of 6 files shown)


bilderbuchi:
I'm pretty sure the build failures are because we are using too old setuptools versions in py34 and py35, see #34 and https://github.com/pytest-dev/pytest-runner/issues/39

cjermain:
Thanks @bilderbuchi!

PR #133, opened 2018-04-01
typo: "for a its" => "for its"

ederag:


cjermain:
@ederag, thanks for the typo corrections!

PR #144, opened 2018-05-30
Added Ametek 7270 Lockin

jmittelstaedt:
Hi,

Changelist:
- New instrument driver for the Ametek 7270 DSP Lockin amplifier. 
- New validator function called `modular_range`, useful for inputting degrees.
    - Related new validator called `modular_range_bidirectional` which accounts for possible negative values, i.e. -1 mod 365 is usually 364 but you may want -1returned instead.
- Added validators on all of the Signal Recovery DSP 7265 Lockin amplifier's properties.

### DIFF
--- a/pymeasure/instruments/ametek/ametek7270.py
+++ b/pymeasure/instruments/ametek/ametek7270.py
@@ -1,0 +1,179 @@
+import logging
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import modular_range, truncated_discrete_set, truncated_range
+
+class Ametek7270(Instrument):
+    """This is the class for the Ametek DSP 7270 lockin amplifier"""
+
+
+    sensitivity = Instrument.control( # NOTE: only for IMODE = 1 for now. Not sure if validator is used correctly for now...
+        "SEN.", "SEN %d",
+        """ A floating point property that controls the sensitivity
+        range in Volts, which can take discrete values from 2 nV to
+        1 V. This property can be set. """,
+        validator=truncated_discrete_set,
+        values=[
+            0.0, 2.0e-9, 5.0e-9, 10.0e-9, 20.0e-9, 50.0e-9, 100.0e-9,
+            200.0e-9, 500.0e-9, 1.0e-6, 2.0e-6, 5.0e-6, 10.0e-6,
+            20.0e-6, 50.0e-6, 100.0e-6, 200.0e-6, 500.0e-6, 1.0e-3,
+            2.0e-3, 5.0e-3, 10.0e-3, 20.0e-3, 50.0e-3, 100.0e-3,
+            200.0e-3, 500.0e-3, 1.0
+        ],
+        map_values=True
+    )
+    slope = Instrument.control(
+        "SLOPE", "SLOPE %d",
+        """ A integer property that controls the filter slope in
+        dB/octave, which can take the values 6, 12, 18, or 24 dB/octave.
+        This property can be set. """,
+        validator=truncated_discrete_set,
+        values=[6, 12, 18, 24],
+        map_values=True
+    )
+    time_constant = Instrument.control( # NOTE: only for NOISEMODE = 0
+        "TC.", "TC %d",
+        """ A floating point property that controls the time constant
+        in seconds, which takes values from 10 microseconds to 100,000
+        seconds. This property can be set. """,
+        validator=truncated_discrete_set,
+        values=[
+            10.0e-6, 20.0e-6, 50.0e-6, 100.0e-6, 200.0e-6, 500.0e-6,
+            1.0e-3, 2.0e-3, 5.0e-3, 10.0e-3, 20.0e-3, 50.0e-3, 100.0e-3,
+            200.0e-3, 500.0e-3, 1.0, 2.0, 5.0, 10.0, 20.0, 50.0,
+            100.0, 200.0, 500.0, 1.0e3, 2.0e3, 5.0e3, 10.0e3,
+            20.0e3, 50.0e3, 100.0e3
+        ],
+        map_values=True
+    )
+    x = Instrument.measurement("X.",
+        """ Reads the X value in Volts """
+    )
+    y = Instrument.measurement("Y.",
+        """ Reads the Y value in Volts """
+    )
+    x1 = Instrument.measurement("X1.",
+        """ Reads the first harmonic X value in Volts """
+    )
+    y1 = Instrument.measurement("Y1.",
+        """ Reads the Y value in Volts """
+    )
+    x2 = Instrument.measurement("X2.",
+        """ Reads the second harmonic X value in Volts """
+    )
+    y2 = Instrument.measurement("Y2.",
+        """ Reads the second harmonic Y value in Volts """
+    )
+    xy = Instrument.measurement("XY.",
+        """ Reads both the X and Y values in Volts """
+    )
+    mag = Instrument.measurement("MAG.",
+        """ Reads the magnitude in Volts """
+    )
+    harmonic = Instrument.control(
+        "REFN", "REFN %d",
+        """ An integer property that represents the reference
+        harmonic mode control, taking values from 1 to 127.
+        This property can be set. """,
+        validator=truncated_discrete_set,
+        values=list(range(1,128))
+    )
+    phase = Instrument.control(
+        "REFP.", "REFP. %g",
+        """ A floating point property that represents the reference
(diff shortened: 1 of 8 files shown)


pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -35,60 +35,76 @@ class DSP7265(Instrument):
...
+        validator=truncated_range,
+        values=[0,2.5e5]
+    )
     dac1 = Instrument.control(
         "DAC. 1", "DAC. 1 %g",
         """ A floating point property that represents the output
-        value on DAC1 in Volts. This property can be set. """
+        value on DAC1 in Volts. This property can be set. """,
+        validator=truncated_range,
+        values=[-12,12]
```
cjermain: Looking at the manual, I am not clear on whether these values should be over the range -12 to 12 V or -12000 to 12000 (fixed point mode). Can you confirm that the command "DAC. 1" is the normal floating point number? The same comments go for the remainder of the added ranges.

pymeasure/instruments/ni/daqmx.py
```diff
@@ -77,7 +77,7 @@ def setup_analog_voltage_in(self, channelList, numSamples, sampleRate=10000, sca
                                 DAQmx_Val_Rising,DAQmx_Val_FiniteSamps,
                                 uInt64(self.numSamples)));
 
-    def setupAnalogVoltageOut(self, channel=0):
+    def setup_analog_voltage_out(self, channel=0):
```
cjermain: I understand the rational for modifying the style on these to match the rest of the package. The reason these differ is because the underlying DAQmx software uses the camel case format. I think it is reasonable to change the names as you did, since the instrument classes are only included for convenience.

pymeasure/instruments/instrument.py
```diff
@@ -35,8 +35,8 @@
 
 
 class Instrument(object):
-    """ This provides the base class for all Instruments, which is 
-    independent of the particular Adapter used to connect for 
+    """ This provides the base class for all Instruments, which is
```
cjermain: It looks like your editor is automatically removing spaces. I would be careful with this behavior, but it looks like there are no material changes.

pymeasure/instruments/validators.py
```diff
@@ -72,6 +72,31 @@ def truncated_range(value, values):
         return min(values)
 
 
+def modular_range(value, values):
```
cjermain: I think `modular_range` and `modular_range_bidirectional` are great additions! I think they deserve tests to ensure that they continue to work as planned. These should happen for both classes in https://github.com/ralph-group/pymeasure/blob/master/tests/instruments/test_validators.py, following the other tests as examples.

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -0,0 +1,203 @@
...
+#
+
+import logging
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import modular_range, truncated_discrete_set, truncated_range
+
+class Ametek7270(Instrument):
```
cjermain: Looks good!

pymeasure/instruments/instrument.py
jmittelstaedt: I think stripping ending white space is one of my own bad habits when looking through files, I'll try to keep from doing that in the future.

pymeasure/instruments/ni/daqmx.py
jmittelstaedt: I had changed this before actually trying to use the DAQ drivers as written. Unfortunately, I couldn't get the daqmx file included with PyMeasure to control my instruments. I ended up using PyDAQmx instead, which seemed to work. I hadn't included those changes with this pull request. After working with DAQ a bit more, I'm actually of the opinion that DAQ should be treated as an adapter of some sort and instruments made which use DAQ as their only adapter, but this is probably a larger discussion which should happen somewhere else.

pymeasure/instruments/signalrecovery/dsp7265.py
jmittelstaedt: Giving the command as "DAC. 1 %g" (with the period included) is floating point mode, so the lockin should expect a normal floating point number as written.

pymeasure/instruments/ametek/ametek7270.py
jmittelstaedt: All of the code as written should work using a VISA GPIB adapter. I'm currently using this machine over ethernet, so I first tried using VISA since it should be able to connect to things over TCP/IP. I ran into trouble with this, and I think it is because of the strange termination characters that the machine returns (for the IP interface, the 7270 returns three termination characters and under normal operation two of them will be ASCII 0 characters, which I could not get to work in VISA. You can set custom termination characters, but VISA gave me an error which seemed to say that there could not be two of the same character in a terminaton sequence). To get around this, I wrote a custom adapter using the Python built-in `socket` class, but since it seems to be pretty specific I'm not sure if you would want it included in "official PyMeasure."

pymeasure/instruments/validators.py
jmittelstaedt: Done. I included `isclose` in the validator testing file, which should be in the standard math library in Python version 3.5+ (https://docs.python.org/3/whatsnew/3.5.html#pep-485-a-function-for-testing-approximate-equality). This seemed necesary since I was getting errors due to floating point imprecision using Python's built-in modulo operator.

bilderbuchi: Note that pytest nowadays has an assert_almost_equal statement (or is it also called is close? I'm on the phone so can't easily check)

pymeasure/instruments/signalrecovery/dsp7265.py
cjermain: Got it, thanks!

pymeasure/instruments/ni/daqmx.py
cjermain: I agree this is a topic for further discussion. I will include these changes for now.

pymeasure/instruments/instrument.py
cjermain: No worries. It just required more review to ensure the files did not change.

pymeasure/instruments/validators.py
cjermain: @bilderbuchi raised a great suggestion to use pytest directly. @jmittelstaedt, check out the `approx` function from pytest (https://docs.pytest.org/en/latest/reference.html#pytest-approx). I would suggest using that instead, because that should satisfy tests on Python 3.4 as well.

pymeasure/instruments/ametek/ametek7270.py
cjermain: Interesting. Your socket adapter could be a good addition in the future. I will keep it in mind if another user raises an issue on the subject.

pymeasure/instruments/validators.py
jmittelstaedt: Thanks for pointing that out! I've changed the tests to use approx instead.

jmittelstaedt: Apparently approx is also only available in a more recent version of python (or pytest at least). I'm now just checking what should be the simplest way by using `abs(validator(x) - x) < tol`. One of the Travis tests still failed, but I couldn't understand what the error is trying to tell me (or, I couldn't understand how I could have caused an error like that with what I've done)

pymeasure/instruments/ametek/ametek7270.py
bilderbuchi: yeah, changing used/expected terminators (for all adapters) would be a great addition for pymeasure, tbh. I 'm currently fighting with two instruments that insist on terminators that seem to be non-default, so it's difficult to change them without replacing pymeasure internals.

pymeasure/instruments/validators.py
bilderbuchi: approx is available from pytest 3.0 (so you have to edit the entry in setup.py to reflect that). 3.0 is available for py3.4+, so that won't be a problem. (Note the API [changed a little in pytest 3.2](https://docs.pytest.org/en/latest/reference.html#pytest-approx) so you can't use  >, >=, < and <= anymore)

bilderbuchi: I don't know for sure, but I don't think the CI error is related to this PR, but @cjermain will be better able to judge that.

cjermain: It looks like a Qt error, which is not related to these changes. I will merge and we can revisit.

cjermain:
Thanks @jmittelstaedt!

PR #146, opened 2018-06-27
Added Keithley 2000 control to enable/disable system beeper.

watersjason:
New control for system beeper is tested. Atom editor also automatically cleaned up extra white space.

PR #147, opened 2018-07-04
Reverting Qt to 4.11 instead of 5.6

cjermain:
The Python 3.6 requirement file is not consistent with the Python 3.5 tests in terms of the Qt version. This PR reverts the version of Qt to 4.11 from 5.6 for those tests. This is intended to solve an issue found during the tests of #146.

### DIFF
--- a/requirements/conda-3.6.yml
+++ b/requirements/conda-3.6.yml
@@ -4,7 +4,7 @@ dependencies:
 - pandas=0.19.2=np112py36_1
 - pip=9.0.1=py36_1
 - py=1.4.34=py36_0
-- pyqt=5.6.0=py36_2
+- pyqt=4.11
 - pyqtgraph=0.10.0=py36_0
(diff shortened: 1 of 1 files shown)


bilderbuchi:
Shouldn't you rather pull the 3.5 tests into the future instead of dragging the 3.6 test into the past (version), for future development and testing on "current" versions?

cjermain:
@bilderbuchi, I agree that updating to Qt 5.6 is the next logical step. However, this PR is intended as a short term fix to keep the master branch working, and to keep all environments consistent (which are otherwise on Qt 4.11).

cjermain:
It looks like the reason Qt 5.6 was used is because conda does not publish 4.11 for Python 3.6. It looks like the best option is indeed to migrate to Qt 5.6. I'm closing this PR to start a new one for that purpose.

Issue #151, opened 2018-07-31
Setting time-constant DSP7265 not working


CasperSchippers:
Whenever I try to set the time-constant of the signal-recovery DSP7265, I end up setting the wrong time-constant: it will revert to the minimal time-constant (10e-6 s) while i want to set e.g. 200e-3 s.
Here is a MWE:
```
from pymeasure.instruments.signalrecovery import DSP7265

lockin = DSP7265("GPIB::12")
lockin.time_constant = 200e-3
```

However, I found that when i replace the property time_constant of the DSP7265 with the exact definition from the DSP7265 instrument class, it seems to be working perfectly.

```
from pymeasure.instruments.signalrecovery import DSP7265
from pymeasure.instruments import Instrument
from pymeasure.instruments.validators import truncated_discrete_set

DSP7265.time_constant = Instrument.control(
    "TC.", "TC %d",
    """ A floating point property that controls the time constant
        in seconds, which takes values from 10 microseconds to 50,000
        seconds. This property can be set. """,
    validator=truncated_discrete_set,
    values=[
        10.0e-6, 20.0e-6, 40.0e-6, 80.0e-6, 160.0e-6, 320.0e-6,
        640.0e-6, 5.0e-3, 10.0e-3, 20.0e-3, 50.0e-3, 100.0e-3,
        200.0e-3, 500.0e-3, 1.0, 2.0, 5.0, 10.0, 20.0, 50.0,
        100.0, 200.0, 500.0, 1.0e3, 2.0e3, 5.0e3, 10.0e3,
        20.0e3, 50.0e3
    ],
    map_values=True
)

lockin = DSP7265("GPIB::12")

lockin.time_constant = 200e-3
```
Is this a bug or something in my setup?

cjermain:
@CasperSchippers, thanks for reporting this. This is certainly not expected. Without overwriting `time_constant`, are you able to write other values (e.g. `20.0`)? Which version of Python and PyMeasure are you using?

CasperSchippers:
I am able to set different values for the time constant, but not as intended (for example trying to set 1s results in a time constant of 20e-6, 2 gives 40e-6, and 3 gives 320e-6, so it also doesn't follow the order of the values in the definition of the parameter). Other properties of the lock-in seem to be working, so I can set e.g. the voltage
I use python 3.5.2 with pymeasure 0.5.1

cjermain:
[...]

PR #155, opened 2018-08-15
Modifying Instruments

jmittelstaedt:
Hello,

I improved the instrument drivers for the ametek7270, newport esp300 and signalrecovery dsp7265. For the dsp7265, I also expanded the functionality of the reference signal specification, but it is *not backwards compatible.* what was previously called `external` only referred to the front-panel external reference. This has been renamed to `external front` and the option `external rear` has been added.
I also added documentation for the ametek7270. I followed the format of the other instruments so I'm assuming this was done correctly.

### DIFF
--- a/docs/api/instruments/ametek/ametek7270.rst
+++ b/docs/api/instruments/ametek/ametek7270.rst
@@ -2,6 +2,6 @@
 Ametek 7270 DSP Lockin Amplifier
 ################################
 
-.. autoclass:: pymeasure.instruments.danfysik.Danfysik8500
+.. autoclass:: pymeasure.instruments.ametek.Ametek7270
     :members:
     :show-inheritance:
--- a/docs/api/instruments/ametek/ametek7270.rst
+++ b/docs/api/instruments/ametek/ametek7270.rst
@@ -1,0 +1,7 @@
+################################
+Ametek 7270 DSP Lockin Amplifier
+################################
+
+.. autoclass:: pymeasure.instruments.danfysik.Danfysik8500
+    :members:
+    :show-inheritance:
--- a/pymeasure/instruments/ametek/ametek7270.py
+++ b/pymeasure/instruments/ametek/ametek7270.py
@@ -84,16 +84,16 @@ class Ametek7270(Instrument):
         """ Reads the Y value in Volts """
     )
     x1 = Instrument.measurement("X1.",
-        """ Reads the X value in Volts """
+        """ Reads the first harmonic X value in Volts """
     )
     y1 = Instrument.measurement("Y1.",
-        """ Reads the Y value in Volts """
+        """ Reads the first harmonic Y value in Volts """
     )
     x2 = Instrument.measurement("X2.",
-        """ Reads the X value in Volts """
+        """ Reads the second harmonic X value in Volts """
     )
     y2 = Instrument.measurement("Y2.",
-        """ Reads the Y value in Volts """
+        """ Reads the second harmonic Y value in Volts """
     )
     xy = Instrument.measurement("XY.",
         """ Reads both the X and Y values in Volts """
--- a/pymeasure/instruments/ametek/ametek7270.py
+++ b/pymeasure/instruments/ametek/ametek7270.py
@@ -1,27 +1,3 @@
-#
-# This file is part of the PyMeasure package.
-#
-# Copyright (c) 2013-2017 PyMeasure Developers
-#
-# Permission is hereby granted, free of charge, to any person obtaining a copy
-# of this software and associated documentation files (the "Software"), to deal
-# in the Software without restriction, including without limitation the rights
-# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-# copies of the Software, and to permit persons to whom the Software is
-# furnished to do so, subject to the following conditions:
-#
-# The above copyright notice and this permission notice shall be included in
-# all copies or substantial portions of the Software.
-#
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-# THE SOFTWARE.
-#
-
 import logging
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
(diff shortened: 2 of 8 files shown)


docs/api/instruments/ametek/ametek7270.rst
```diff
@@ -0,0 +1,7 @@
+################################
+Ametek 7270 DSP Lockin Amplifier
+################################
+
+.. autoclass:: pymeasure.instruments.danfysik.Danfysik8500
```
cjermain: This should be referencing `pymeasure.instruments.ametek.Ametek7270`.

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -32,20 +32,29 @@
 class Ametek7270(Instrument):
     """This is the class for the Ametek DSP 7270 lockin amplifier"""
 
+    SENSITIVITIES = [
```
cjermain: This looks good! Having them as class variables makes them more accessible to other uses.

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -200,4 +205,8 @@ def auto_gain(self, setval):
         else:
             self.write("AUTOMATIC 0")
 
-    # TODO: shutdown function. Not sure what defaults should be, if any needed for it.
+    def shutdown(self):
+        """ Ensures the instrument in a safe state """
+        self.voltage = 0.
+        self.isShutdown = True
```
cjermain: This is the correct variable set in the `Instrument` class. However, it would be great to get the code in line with Python standards and use snake case -- preferring `is_shutdown`. I think that is worth doing in a future PR.

pymeasure/instruments/newport/esp300.py
```diff
@@ -176,7 +175,7 @@ class Axis(object):
         "MD?",
         """ Returns a boolean that is True if the motion is finished.
         """,
-        cast=bool
```
cjermain: The old version uses the casting in the adapter class. I would prefer to keep this behavior.
https://github.com/ralph-group/pymeasure/blob/29ce412f058607075d8d5ad8b3119c40c6f1c819/pymeasure/adapters/adapter.py#L62

pymeasure/instruments/newport/esp300.py
```diff
@@ -203,11 +202,11 @@ def disable(self):
         """ Disables motion for the axis. """
         self.write("MF")
 
-    def home(self, type=1):
```
cjermain: It looks like the previous version of `home` was not well tested. Can you add `htype` to the docs, so that it is clear what this does?

pymeasure/instruments/newport/esp300.py
```diff
@@ -220,10 +219,10 @@ def zero(self):
         self.write("DH")
 
     def wait_for_stop(self, delay=0, interval=0.05):
-        """ Blocks the program until the motion is completed. A further 
+        """ Blocks the program until the motion is completed. A further
         delay can be specified in seconds.
         """
-        self.write("WS%g" % delay*1e3)
+        self.write("WS%d" % (delay*1e3))
```
cjermain: Can you confirm that this should be an integer?

pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -110,19 +132,21 @@ class DSP7265(Instrument):
     id = Instrument.measurement("ID",
         """ Reads the instrument identification """
     )
+    reference = Instrument.control(
```
cjermain: This is a nice clean up of the old property. Awesome!

pymeasure/instruments/newport/esp300.py
jmittelstaedt: Yes, the documentation for the instrument specifies this should be an integer. I have also tried it with the instrument itself and it has worked.

pymeasure/instruments/newport/esp300.py
jmittelstaedt: Hmm, I changed this long enough ago that I can't remember exactly why I did it, but presumably it wasn't returning what I expected. I have a feeling that it would always return `True` before, but I'll check it out again.

docs/api/instruments/ametek/ametek7270.rst
jmittelstaedt: Whoops! that's what I get for copy/pasting. It should be fixed now.

pymeasure/instruments/newport/esp300.py
jmittelstaedt: Honestly, I'm not completely clear on what the command does, so if I did add it to the documentation I would just be copying the manual. The change here just came from me realizing while looking through the code that an int wasn't passed where it clearly should be. I changed `type` to `htype` since `type` is a Python keyword, and although I'm sure Python can figure it out it seems like bad practice.

pymeasure/instruments/newport/esp300.py
cjermain: Interesting. Please verify with the instrument. It would be nice to fix it in the adapter class, since we use this `cast` argument in other contexts.

pymeasure/instruments/newport/esp300.py
cjermain: Fair enough. Adding which values `htype` can take would be great. I think using `type` as a variable is acceptable, as long as you don't intend to request the type of a variable in the same local context (in this case the `home` method).

pymeasure/instruments/newport/esp300.py
cjermain: Cool, thanks!

pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -135,18 +159,12 @@ class DSP7265(Instrument):
         map_values=True
     )
     time_constant = Instrument.control(
-        "TC.", "TC %d",
+        "TC", "TC %d",
```
cjermain: For future reference, this is the change requested in #151.

pymeasure/instruments/newport/esp300.py
jmittelstaedt: I'm not sure where my other comment went, but I have tested putting `cast = lambda x: bool(float(x))` in thand it does work with the instrument. 
How were you thinking of fixing this in the adapter class? I would think we would need to assume that we would need to always cast to float or something first, but is it reasonable to assume that we are getting numbers in strings from all instruments?

cjermain: I realize the issue now. By default, all returned values are cast as floats. So when `cast=bool` is used, it directly casts as a boolean -- this will return True for strings as long as they are not blank.
```python
>>> s = '1'
>>> bool(s)
True
>>> bool(float(s))
True
>>> s = '0'
>>> bool(s)
True
>>> bool(float(s))
False
>>> 
```
So, we can either:
1. Use `bool(float(x))` in the adapter if `bool` is passed as a cast function.
2. Use `get_process=bool`, since `cast=float` by default.

What do you think?

jmittelstaedt: Using `bool(float(x))` in the adapter whenever `bool` is passed as a cast would probably be fine, and would probably be easier to work with for any future instrument development.  Off the top of my head, I can't think of any cases where this would give an unexpected result. If the instrument is returning something strange which should be converted to a boolean (I'm thinking like a string `"true"` or `"T"` or something) then anyone creating a new instrument would probably realize they would need to handle that by passing a `get_process`

jmittelstaedt: I've made the change to the base `adapter` class. I tested using `cast == bool` by making a function `def afunc(x): return x == bool` in an interactive environment, and it behaved as expected.

cjermain:
Thanks @jmittelstaedt! Awesome work.

bilderbuchi:
@cjermain did you check the bit I remarked on above before merging?

cjermain:
@bilderbuchi, I am not seeing your comments. Can you post them here?

bilderbuchi:
I started a PR review, currently "pending"(you should have been presented with that information when merging I think), in agilent8257D.py:

Isn't this missing the quotes? Right now, this is the python built-in function input. Same above....

cjermain:
@bilderbuchi, I appreciate you reviewing this PR! As far as I know, [pending reviews are only visible to their authors until submitted](https://help.github.com/articles/reviewing-proposed-changes-in-a-pull-request/). In the future, feel free to submit a review, requesting changes if needed. Let me know if you need to be added to a team for this to happen.

I'm not sure what line number you are mentioning in agilent8257D.py. Can you post a [permanent link to the code section](https://help.github.com/articles/creating-a-permanent-link-to-a-code-snippet/)? That should display the code inline in the preview.

bilderbuchi:
ha, I was not aware that it worked this way, my bad! 
Meantime, I have checked my pending comments, and they actually weren't relevant. Sorry for the noise!

Issue #54, opened 2016-11-18
Autoscaling error when a trace color changed and then  that trace is hidden


ndr37:
In a somewhat hilarious turn of events, changing the color of a trace and then hiding it causes the autoscaling to fail spectacularly. The scaling will zoom out to huge nonsense values for the axes and then loop back to correct values then scale out again without end. Turning off autoscaling stops this cyclical runaway scaling. Removing the trace from trace list does not fix the problem. Only closing the program seems to fix it.

guenp:
I can verify that this is indeed hilarious and spectacular.

bilderbuchi:
@CasperSchippers can you take a look if this still occurs?

PR #157, opened 2018-08-24
Adding Live Image Plotting Functionality

jmittelstaedt:
Hi,

I've been working on adding live image plotting to PyMeasure. I now have a *functional* version that I think is ready to open up for other people's suggestions about and/or help in making this into something which is ready for a wider audience. I'm expecting that there is still a decent amount of work to be done before this should be merged with master. An example of the image plotting in action is shown at the bottom of this post.

I haven't added any documentation for the new changes (besides comments in the files I've changed), but I have included an example in the examples directory (in examples/Basic_Image_GUI/test_image_gui.py)  which showcases it's basic functionality. The list of added classes and their functions is described below. All of them have direct counterparts of the usual plotting classes (sorry, I don't know how to reference code files in a PR).

Added Classes:
- `display.curves.ResultsImage`: inherits from PyQtGraph's `ImageItem` and takes the place of the `ResultsCurve` in the `PlotWidget`. This handles creating and updating the image displayed. Also, added a dependency to `matplotlib.colormap` here.
- `display.widgets.ImageFrame`: Direct counterpart of `PlotFrame` which adds crosshairs to the image plot.
- `display.widgets.ImageWidget`: Direct counterpart of `PlotWidget` which adds a z axis needed for the image plotting, and removes the option to change the X and Y axes.
- `display.windows.ManagedImageWindow`: Counterpart of `ManagedWindow` adding a new tab for the results image plot and makes sure UI elements interact with the image properly.
- `display.manager.ImageExperiment`: added storing the `ResultsImage`
- `display.manager.ImageExperimentQueue`: modified to work with `ImageExperiment`
- `display.manager.ImageManager`: modified to handle working with images

Now for the fun part: known bugs!
- ~~Images do not automatically display when queued. They need to be deselected and reselected in the browser to actually appear. I know that this has to do with the `ResultsImage` only being added to the `ImageWidget` in the `ManagedImageWindow.browser_item_changed` method. I looked around for where the regular `ResultsCurve` gets added to the `PlotWidget`, but couldn't find anywhere else. Hopefully this fix will be obvious to someone more familiar with PyMeasure's display stuff.~~
- ~~Haven't been able to reproduce this, but for a while deselecting the browser item of the first experiment queued would make the axes zoom out uncontrollably until it was re-checked.~~
- ~~After deselecting an image, the axes on the image revert to "pixel coordinates" instead of "data coordinates"~~

Needed Improvements:
- In contrast to plots, where showing several makes sense, only one image should be shown at a time. Maybe adding radio buttons in the browser items to select which experiment has its image displayed would be nice. I'm not sure how that would deal with several queued experiments, though.
- ~~Currently, the axes on the image do not show the "data coordinates" well. This has to do with how they are set which currently happens in `ManagedImageWindow.browser_item_changed`. Part of the problem is that these axes are from a PyQtGraph `PlotWidget` which we are putting an image into. There may not be a great way of setting the axes in the `PlotWidget`~~
- Options for changing the colormap, it is currently hard-coded to use viridis to appease @ndr37 
- Usually, there is an empty row of pixels at the end of the image. The user must decide which data columns of the procedure will end up being the X and Y axes of the image and create parameters to describe the start, stop and step expected values for these data columns (as shown in the example). If you specify the start, stop, step as `(-2,2,1)` respectively, usually something like `range` will be used to create the values which will end up being `[-2,-1,0,1]`, but perhaps the user ensures that the *actual* range of values in the respective data column will be `[-2,-1,0,1,2]`. To account for this, we create a blank image to be filled with 5 pixels on that side, where there is a good possibility only the left or bottom four will actually be used, which results in an empty row/column of pixels.

Final Thoughts:
It would probably be more appropriate to have the image stored inside of a PyQtGraph `ImageView` or something similar. I believe these have much of the functionality needed with regards to converting between data and pixel coordinates for the image. I did not use them since they also came with a lot of functionality which I thought was unnecessary (like ROI selection, finding averages in the ROI etc.) but these can probably be stripped off. It also seems less straightforward to integrate these with the current GUI, and doing that integration seems like something that is above my head at the moment.

I look forward to all suggestions and improvements from others in the PyMeasure community. 
Also, let me know if there is a better way to do this pull request, e.g. not requesting to pull to master since this may take a while to actually develop, don't request to pull a branch other than main of my fork, etc.

Example of the image plotting in action:

pm_img_example


cjermain:
@jmittelstaedt, this is awesome! I especially liked your screenshot. Working with images has been a missing functionality that I am glad to see being worked on. As far as logistics, I think basing this branch on master is fine. I am interested to get a number of reviewers on this PR, and make sure that we have decent test coverage where possible. I will start going through these changes, and look forward to converging on a final solution.

pymeasure/display/curves.py
```diff
@@ -26,6 +26,7 @@
 
 import pyqtgraph as pg
 import numpy as np
+from matplotlib.cm import viridis  # JM: To make Neal happy ;)
```
cjermain: Nice! @ndr37, what do you think?

Currently matplotlib is considered an extra dependency, i.e. it is not required (see [setup.py#47](https://github.com/ralph-group/pymeasure/blob/60a2720486397dff8f6d5d85bcb211207545afbe/setup.py#L47)). We need to either promote it to a requirement or use it conditionally (using a try-catch on import). This is certainly open for debate. My first thought is to use it conditionally, so that we don't have to introduce more dependencies. Any thoughts?

pymeasure/display/curves.py
```diff
@@ -75,6 +76,75 @@ def update(self):
...
+    def __init__(self, results, x, y, z, force_reload=False):
+
+        self.results = results
+        self.x = x
+        self.y = y
+        self.z = z
+        # NOTE: Below we're making an assumption on the form of *any* procedure which we would use
+        # with an image. Seems reasonable, we need to make some compromise so that
+        # we know how to construct the resolution and so that we know what the heck the range is
+        self.xstart = getattr(self.results.procedure,self.x + '_start')
```
cjermain: The `getattr` calls should have a space between the arguments to be PEP8.

pymeasure/display/curves.py
```diff
@@ -75,6 +76,75 @@ def update(self):
...
+        self.xend = getattr(self.results.procedure,self.x + '_end')
+        self.xstep = getattr(self.results.procedure,self.x + '_step')
+        self.xsize = int(np.ceil((self.xend - self.xstart) / self.xstep)) + 1 # JM: add one since arange doesn't include bound, but maybe experiment will give it. Just defaults to empty pixels, not the end of the world.
+        self.ystart = getattr(self.results.procedure,self.y + '_start')
+        self.yend = getattr(self.results.procedure,self.y + '_end')
+        self.ystep = getattr(self.results.procedure,self.y + '_step')
+        self.ysize = int(np.ceil((self.yend - self.ystart) / self.ystep)) + 1 # JM: add one since arange doesn't include bound, but maybe experiment will give it. Just defaults to empty pixels, not the end of the world.
+        self.img_data = np.zeros((self.ysize,self.xsize,4))
+        self.force_reload = force_reload
+        self.colormap = viridis # JM: Hard coded for Neal
```
cjermain: As above, it may be wise to swap this in conditionally on the successful import of matplotlib, so that it doesn't become a requirement.

pymeasure/display/curves.py
```diff
@@ -75,6 +76,75 @@ def update(self):
...
+        super().__init__(image = self.img_data)
```
cjermain: Key-word arguments should not have spaces around the equals.

pymeasure/display/curves.py
```diff
@@ -75,6 +76,75 @@ def update(self):
 
 # TODO: Add method for changing x and y
 
+#############################################################################################
```
cjermain: This comment should be removed before merge.

pymeasure/display/curves.py
```diff
@@ -75,6 +76,75 @@ def update(self):
...
+        """Convenience function since numpy rounds to even"""
+        if x%1 >= 0.5:
+            return int(x) + 1
+        else:
+            return int(x)
+
+    # TODO: colormap selection
+
+
+# JM: ###############################################################################################
```
cjermain: This comment should be removed before merge.

pymeasure/display/manager.py
```diff
@@ -279,3 +278,78 @@ def abort(self):
             self._worker.stop()
 
             self.aborted.emit(self._running_experiment)
+
+# JM: ####################################################################
```
cjermain: This comment should be removed before merge.

pymeasure/instruments/newport/esp300.py
```diff
@@ -175,7 +175,7 @@ class Axis(object):
         "MD?",
         """ Returns a boolean that is True if the motion is finished.
         """,
-        cast=bool
+        get_process=lambda x: bool(x)
```
cjermain: Did you merge in the latest `master` branch?

pymeasure/display/manager.py
```diff
@@ -279,3 +278,78 @@ def abort(self):
...
+        if isinstance(value, ExperimentImg):
+            return value in self.queue
+        if isinstance(value, str):
+            for experiment in self.queue:
+                if basename(experiment.data_filename) == basename(value):
+                    return True
+            return False
+        return False
+
+# JM: QUESTION: Can this just inherit from Manager and add im_plot and override
```
cjermain: That seems reasonable.

pymeasure/display/manager.py
```diff
@@ -279,3 +278,78 @@ def abort(self):
...
+    """ 
+    Overwrites needed features from :class:`.ExperimentQueue` to make image 
+    features work
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def __contains__(self, value):
+        if isinstance(value, ExperimentImg):
```
cjermain: Is `ExperimentImg` -> `ImageExperiment`?

pymeasure/display/widgets.py
```diff
@@ -229,6 +229,241 @@ def update_y_column(self, index):
         axis = self.columns_y.itemText(index)
         self.plot_frame.change_y_axis(axis)
 
+# JM: #########################################################################
```
cjermain: This comment should be removed before merge.

pymeasure/display/widgets.py
```diff
@@ -229,6 +229,241 @@ def update_y_column(self, index):
...
+    # def update_y_column(self, index):
+    #     axis = self.columns_y.itemText(index)
+    #     self.plot_frame.change_y_axis(axis)
+
+
+    def update_z_column(self, index):
+        axis = self.columns_z.itemText(index)
+        self.image_frame.change_z_axis(axis)
+
+#JM: ########################################################################
```
cjermain: This comment should be removed before merge.

pymeasure/display/widgets.py
```diff
@@ -229,6 +229,241 @@ def update_y_column(self, index):
...
+            match = None
+            
+        if match:
+            if 'units' in match.groupdict():
+                label = re.sub(units_pattern, '', axis)
+                return label, match.groupdict()['units']
+        else:
+            return axis, None
+
+    # def change_x_axis(self, axis):
```
cjermain: Please remove the commented code in this file.

pymeasure/display/curves.py
bilderbuchi: if this import is just for the colormap definition, it might be worth it to just hand-define viridis for pg use. There's a [PR open](https://github.com/pyqtgraph/pyqtgraph/pull/446) for including that into pyqtgraph itself, and a [related issue](https://github.com/pyqtgraph/pyqtgraph/issues/561), maybe you can pick a good way to do that from those.

cjermain: Thanks @bilderbuchi for finding the PR on pyqtgraph. It would be great to solve this upstream, but worst case, we can use can hand define it in the short term.

pymeasure/display/widgets.py
jmittelstaedt: Yup. As a rule, any comment with a JM at the beginning I was using to make development easier and is intended to be removed in a final version. I'll begin removing some of them now.

pymeasure/display/curves.py
jmittelstaedt: Thanks, it's been fixed. There are probably more PEP8 violations in what I've written, so let me know if you find any more.

examples/Basic_Image_GUI/test_image_gui.py
```diff
@@ -0,0 +1,93 @@
+import sys
```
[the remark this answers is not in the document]
jmittelstaedt: Ah that's a good idea, I've added one.

pymeasure/display/curves.py
jmittelstaedt: I've made it optional and included a greyscale colormap to fall back on. I know that pyqtgraph has some builtin colormaps, but from what I read in the documentation it wasn't evident how to  use them without using an `ImageView` or `ImageWidget` which is why I am "hand-making" the image in `curves.ResulsImage`. Ultimately, switching to containing the image inside of one of the proper pyqtgraph image widgets would probably be best.

pymeasure/display/curves.py
jmittelstaedt: Done.

pymeasure/display/curves.py
jmittelstaedt: Thanks

pymeasure/instruments/newport/esp300.py
jmittelstaedt: Oh, I thought I had. Since this seems to be the only thing missing from `master`, I've changed it back by hand.

pymeasure/display/widgets.py
jmittelstaedt: Done.

pymeasure/display/curves.py
bilderbuchi: ok, but if you just need the 2D array of colors (as evidenced by the greyscale fallback), why drag the mpl dependency in at all, if you can hardcode the values (which should not change) until this gets upstreamed, e.g. by `np.array(mp.cm.viridis.colors)` or somesuch?

pymeasure/display/curves.py
```diff
@@ -23,15 +23,23 @@
...
 
 import logging
+import sys
 
 import pyqtgraph as pg
 import numpy as np
+try:
+    from matplotlib.cm import viridis
+except ImportError:
+    log.warning("Matplotlib not found. Images will be greyscale")
```
cjermain: The `log` variable is called before construction. That section (currently lines 37-38) needs to be above this one.

cjermain:
I ran the `image_gui.py` example. I see the data plotted in the "Results Graph", but none of the data is shown in the "Results Image". I am using the [conda-3.6.yml](https://github.com/ralph-group/pymeasure/blob/master/requirements/conda-3.6.yml) environment, with matplotlib installed as well.

I have not had time to do a debug.  Any ideas?

jmittelstaedt:
@cjermain one current "quirk" is that the image is only displayed if you un-check then re-check the item in the browser window. I mentioned this in the initial message in the pull request, but to reiterate here the only place where I could find that the `plotItem` is put into the `plotWidget` (or in this case where the `ImageItem` is put into the `plotWidget`) is when the browser was checked. If you know where this should be happening, I'd love to know.

cjermain:
@jmittelstaedt, Ok, I missed that originally. I am able to see the live image plotting now -- cool! This suggests that the initial construction is not working properly. Have you tried calling `update` or replicating the behavior in `browser_item_changed`?

I had a few ideas when running through the example:
1. How do you envision handling displaying images from multiple experiments (overlay or only allow one image at a time)?
2. Is this a feature to add to the Open file dialog ([ResultsDialog](https://github.com/ralph-group/pymeasure/blob/master/pymeasure/display/widgets.py#L359))?

pymeasure/display/curves.py
```diff
@@ -75,6 +83,76 @@ def update(self):
...
+        self.img_data = np.zeros((self.ysize,self.xsize,4))
+        self.force_reload = force_reload
+        if 'matplotlib.cm' in sys.modules:
+            self.colormap = viridis
+        else:
+            self.colormap = _greyscale_colormap
+
+        super().__init__(image=self.img_data)
+
+    def update_img(self):
```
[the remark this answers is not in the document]
jmittelstaedt: Sounds like a good idea, I've changed it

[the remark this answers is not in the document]
jmittelstaedt: Well, apparently just doing a find and replace for `update_img` through all of the display files broke something, I'll try to figure it out later.

jmittelstaedt:
@cjermain With fresh eyes, I found what I had missed before. I needed to add the image item to the plot in `display.manager.ImageManager.load` which is eventually called when the Queue button is hit. I've tested the change and the image should display right when queued, without needing to de-select and re-select the browser item.

1. The best way to do this would probably be to have radio buttons on the browser items separate from the current check boxes which would allow the user to select which experiment has its image currently displayed. This would also fix some problems with how we set the "data coordinates" on the axes of the image. Currently, if you queue an experiment with e.g. x limits from -1 to 1 and then queue another with limits of -1 to 2, the x coordinate on the plot will go from -1 to 2 *while the first experiment is still running* meaning the "data coordinates" for that experiment's image are completely incorrect. This can be fixed by deselecting and then re-selecting the browser item, since doing that also resets the coordinates displayed on the axes. There are some other problems with the coordinates on the axes as currently implemented as well.

2. Ah, that is tied to the Open button on the browser widget correct? Whenever I've clicked that, PyMeasure immediately crashes and I end up with an `AttributeError: 'NoneType' object has no attribute 'addWidget'` error. What is the intended functionality of that button?
The full traceback for the error is:
`Traceback (most recent call last):
  File "C:\Users\Joseph\Anaconda3\lib\site-packages\pymeasure-0.5.1-py3.6.egg\pymeasure\display\windows.py", line 709, in open_experiment`
`dialog = ResultsDialog(self.procedure_class.DATA_COLUMNS, self.x_axis, self.y_axis)`
  `File "C:\Users\Joseph\Anaconda3\lib\site-packages\pymeasure-0.5.1-py3.6.egg\pymeasure\display\widgets.py", line 545, in __init__`
    `self._setup_ui()`
  `File "C:\Users\Joseph\Anaconda3\lib\site-packages\pymeasure-0.5.1-py3.6.egg\pymeasure\display\widgets.py", line 568, in _setup_ui`
    `self.layout().addWidget(preview_tab, 0, 5, 4, 1)`
AttributeError: 'NoneType' object has no attribute 'addWidget'`

cjermain:
@jmittelstaedt, thanks for the update. It looks like this feature is getting really close. I agree with using radio buttons as a solution for only displaying a single image at a time. I think if we can resolve that, this PR will be ready to merge.

As for the "Open" button, it should provide an open file window with a plot preview. It sounds like there are additional issues. Can you raise this as a separate issue? While it would be nice, I don't think the open file preview is necessary for this initial PR.

I'm interested to hear reviews and comments from other PyMeasure users. @bilderbuchi, any further review comments?

bilderbuchi:
No comments, but I don't have the time for reviewing this unfortunately. :-\

cjermain:
@jmittelstaedt, I would love to get this merged in for PyMeasure 0.6. What is the status?

jmittelstaedt:
I've been pretty busy with other stuff recently so I haven't had much chance to work on it. After this weekend I should have more time. Just adding radio buttons will *hopefully* not be too difficult and won't take much time after that.

cjermain:
@jmittelstaedt, sorry for the slow response -- I have not had much bandwidth for PyMeasure recently.

I have seen the zooming issue in the past, and it was reported in #54. I'm not sure if this is an implementation issue in PyMeasure or it is upstream in pyqtgraph. I appreciate your suggestions on where to diagnose. I will make some time to review the issue.

cjermain:
I've done some profiling on the zooming issue. I can reproduce it, and am working on a solution.

cjermain:
I've done some profiling on the zooming issue. I can reproduce it, and am working on a solution.

cjermain:
I think this feature would be great to get across the finish line. I don't have the bandwidth to diagnose the zooming issue. @jmittelstaedt, any luck on your end?

jmittelstaedt:
@cjermain I have not been able to figure this out yet. However, this bug was only introduced after changing how the images were displayed, after I made it possible to only show one image at a time. 

What I've been doing locally is using commit 02904fc where, if there is more than one image, they are superimposed on top of one another. I think there are actually cases where this behavior could be considered desirable. The pixels rendered have the right "units" so if you do a coarse scan from X=0-100, Y=0-100 to find some feature and then a finer scan from X=20-50, Y=20-50, the finer scan will be superimposed on top of the coarser scan. Users still have the option to hide any superfluous images, and there are no zooming issues when images are hidden/shown. What do you think about using this commit instead?

cjermain:
@jmittelstaedt, I think that it is a reasonable approach to revert the PR to 02904fc. I agree that multiple regions on the same plot can be an advantage. Can you create a separate issue to track the findings, and provide a place for discussing future improvements after this is merged?

jmittelstaedt:
Hi @cjermain I believe I have made the necessary rollbacks in the code and fixed the one merge issue which needed to be taken care of. I've also created issue #261 to track bugs and improvements in the future.

pymeasure/display/manager.py
cjermain: It looks like a number of `# JM:` comments are still in the PR. Can you remove them?
```
$ grep -rnI "# JM"
pymeasure/display/widgets.py:263:        self.x_axis_changed.emit(x_axis) # JM: Not sure if this does anything now...
pymeasure/display/widgets.py:267:        self.y_axis_changed.emit(y_axis) # JM: Not sure if this does anything now...
pymeasure/display/widgets.py:279:        self.plot_widget = pg.PlotWidget(self, background='#ffffff') # JM: NOTE: This is just pulling a base pg PlotWidget, so no infinite regress!
pymeasure/display/widgets.py:403:    # JM: Maybe should have a change image function, so that we know there is 
pymeasure/display/manager.py:282:# JM: ####################################################################
pymeasure/display/curves.py:94:        # JM: NOTE: Below we're making an assumption on the form of *any* procedure 
pymeasure/display/curves.py:101:        self.xsize = int(np.ceil((self.xend - self.xstart) / self.xstep)) + 1 # JM: add one since arange doesn't include bound, but maybe experiment will give it. Just defaults to empty pixels, not the end of the world.
pymeasure/display/curves.py:105:        self.ysize = int(np.ceil((self.yend - self.ystart) / self.ystep)) + 1 # JM: add one since arange doesn't include bound, but maybe experiment will give it. Just defaults to empty pixels, not the end of the world.
pymeasure/display/windows.py:31:from .curves import ResultsCurve # JM: only need to import another if we want a PlotterWindow equivalent
pymeasure/display/windows.py:501:# JM: ############################################################################
pymeasure/display/windows.py:645:    def browser_item_changed(self, item, column): # JM: adds or removes item based on checkbox in browser item. Need to destroy old and add new.
pymeasure/display/windows.py:680:            action_change_color = QtGui.QAction(menu) # JM: probably remove this??
pymeasure/display/windows.py:772:        # if color is None: # JM: I guess we could extend this to colormaps?
pymeasure/display/windows.py:825:    def setup_plot(self, plot): # JM: Will need to change this probably.
pymeasure/display/windows.py:839:    def setup_im_plot(self, im_plot): # JM: Will need to change this probably.

```

pymeasure/display/curves.py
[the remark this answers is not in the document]
cjermain: Please rename this to `update`. Here are the places the it is used in this branch:
```
$ grep -rnI update_img
pymeasure/display/widgets.py:259:                item.update_img()
pymeasure/display/widgets.py:311:                        item.update_img()
pymeasure/display/widgets.py:335:                item.update_img()
pymeasure/display/manager.py:353:        experiment.image.update_img()
pymeasure/display/curves.py:121:    def update_img(self):
pymeasure/display/windows.py:661:                experiment.image.update_img()
pymeasure/display/windows.py:741:                    experiment.image.update_img()
```

pymeasure/display/windows.py
```diff
@@ -497,3 +497,396 @@ def finished(self, experiment):
...
             self.browser_widget.clear_button.setEnabled(True)
+
+# JM: ############################################################################
+
+# QUESTION: Can we just inherit from ManagedWindow?
+class ManagedImageWindow(QtGui.QMainWindow):
+    """
+    Abstract base class.
+
+    The ManagedWindow provides an interface for inputting experiment
```
cjermain: This should read `ManagedImageWindow`. Please do a search/replace in this section.

pymeasure/display/windows.py
```diff
@@ -497,3 +497,396 @@ def finished(self, experiment):
         if not self.manager.experiments.has_next():
             self.abort_button.setEnabled(False)
             self.browser_widget.clear_button.setEnabled(True)
+
+# JM: ############################################################################
+
+# QUESTION: Can we just inherit from ManagedWindow?
```
cjermain: Yes, I think it would be better to inherit from `ManagedWindow` to avoid re-stating code.

cjermain: I think this is worth putting off for another PR.

cjermain:
Great work @jmittelstaedt! This is a major new feature, and I think people will find it very useful.

PYmeas:
i am using live plotting with managed window. i wonder where are the input parameters updated when a new experiment is queued? I am trying to use one of the input to name a file. when saved in the data file i do see the inputs updated but when i call it in queue for filename i only get the initial default value.

pymeasure/display/curves.py
bilderbuchi: @jmittelstaedt @cjermain the mpl 2.0 colormaps are included in pyqtgraph 0.11 now.

PR #165, opened 2018-09-24
Updating to PyQt 5.6

cjermain:
This PR upgrades the requirements to PyQt 5.6 to ensure that all CI builds are using the same environment.


cjermain:
The unit tests `test_plotter` and `test_windows` cause CI checks to intermittently fail. They are being disabled for now, until they can be better implemented. They are of limited value given the extensive use of mocking needed to get them to work. It will be worth revisiting them in the future.

Issue #170, opened 2018-12-13
Few bugs to fix


thosou:
Hi,

I have gathered a few bugs:

* In  pymeasure/pymeasure/log.py, I think there is a mistake.
      
https://github.com/ralph-group/pymeasure/blob/e2a3c7ecae3e2c533df32e59490fb553a786e616/pymeasure/log.py#L102

It should be something like:
```
if filename is not None:
````

* If we call the status property it will return the property object. Properties cannot be added dynamically to the instance but should be added to the class:

https://github.com/ralph-group/pymeasure/blob/e2a3c7ecae3e2c533df32e59490fb553a786e616/pymeasure/instruments/instrument.py#L67-L72

Changing self to Instrument will fix this:

```
        if includeSCPI:
            # Basic SCPI commands
            Instrument.status = self.measurement("*STB?",
                                           """ Returns the status of the instrument """)
            Instrument.complete = self.measurement("*OPC?",
                                             """ TODO: Add this doc """)
````

* write_termination is missing as a safe keyword and term_chars is not supported anymore.

https://github.com/ralph-group/pymeasure/blob/e2a3c7ecae3e2c533df32e59490fb553a786e616/pymeasure/adapters/visa.py#L56-L58

* Finally I have found a typo in the IV curve tutorial:
https://pymeasure.readthedocs.io/en/latest/tutorial/procedure.html#modifying-our-script

The unit in parameter should be "units". When I copy paste this code I get an error.

I don't know how you guys want to proceed but maybe I can make a or several pull requests. I could also add more things to the SCPI check for instance.

cjermain:
@thosou, thanks for calling out these bugs! Please make PRs as you see fit, and then request a review.

The second issue you mention with the properties being set on `self` is a good catch. The challenge here is that the Instrument class is shared among a number of other instruments, so I don't think we want to set it at that level. I would suggest using `__new__(cls, ...)` to set the properties.

bilderbuchi:
I think you can create a custom class on the fly, see e.g. [here](https://stackoverflow.com/a/2954373/599884) or [this question](https://stackoverflow.com/questions/48448074/adding-a-property-to-an-existing-object-instance) (by @vascotenner I heavily suspect).

vascotenner:
Properties can be dynamically added to an instance. For an example see the stackoverflow questions above, or this camera driver:
https://github.com/vascotenner/lantz/blob/drivers_developv/lantz/drivers/basler/basler.py

[...]

PR #171, opened 2018-12-17
Fix/visa adapter

thosou:
Hi,

Here is the fix of the visa adapter so that write_termination is now a safe keyword (#170) .
Looking at the adapter, could we also add an argument to configure the backend?
something like:

`self.manager = visa.ResourceManager(backend)`

Right now my way to change the backend is to set the environment variable

`os.environ['PYVISA_LIBRARY'] = '@py`

### DIFF
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -55,7 +55,7 @@ class VISAAdapter(Adapter):
         self.manager = visa.ResourceManager()
         safeKeywords = ['resource_name', 'timeout',
                         'chunk_size', 'lock', 'delay', 'send_end',
-                        'values_format', 'read_termination']
+                        'values_format', 'read_termination', 'write_termination']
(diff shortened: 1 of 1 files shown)


cjermain:
@thosou, thanks for the PR. It looks like the `term_chars` field was [dropped in PyVISA 1.6](https://pyvisa.readthedocs.io/en/1.8/migrating.html?highlight=write_termination#deprecated-term-chars-and-automatic-removal-of-cr-lf). I think a `backend` argument is a good idea -- feel free to add to this PR since it is within the scope.

thosou:
I added the selection of the backend with the visa_library parameter. This is the one that is passed to the ressource manager in [pyvisa](https://pyvisa.readthedocs.io/en/master/api/resourcemanager.html#pyvisa.highlevel.ResourceManager). The default value is empty string.

cjermain:
@thosou, nice work! Thanks for the fixes.

PR #175, opened 2019-01-17
Added support for Agilent 33220A Function Generator

CasperSchippers:
**Finnished and ready for implementation**

Added support for the agilent 33220A Function / Arbitrary Waveform Generator, 20 MHz


cjermain:
Looks like a good start. Ping me when you are ready for review.

CasperSchippers:
Tested it (finally) and seems to work like a charm, so I think it is ready for merging

PR #179, opened 2019-02-08
Tinix84

tinix84:
'i would discuss about the new instruments added and the lecroy adapter driver implemented
thanks

### DIFF
--- a/docs/api/instruments/index.rst
+++ b/docs/api/instruments/index.rst
@@ -28,6 +28,7 @@ Instruments by manufacturer:
    hp/index
    keithley/index
    lakeshore/index
+   lecroy/index
    newport/index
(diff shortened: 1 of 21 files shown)


cjermain:
@tinix84 thanks for submitting this PR! I would recommend that you split this PR up for specific instruments or concepts. So for example you can have different PRs for each:
- ActiveDSOAdapter
- LeCroy WaveRunner606ZI
- Agilent34461A
- Agilent34465A
- Chroma62024p6008

That will help with the review and feedback process to get them accepted into the codebase. It will also provide an opportunity to fix the merge conflict with `pymeasure/adapters/__init__.py`.

tinix84:
OK I will try

PR #181, opened 2019-02-14
Added support for Aux inputs and outputs on the SRS SR830

CasperSchippers:
**Finished & Ready for implementation**

Added support for the auxiliary inputs and outputs on the rear panel of the Standford Research Systems - SR830 Lock-In Amplifier.

For all aux attributes, I added a 2nd attribute to be consistent with other lock-in modules (that often have some DAC & ADC ports, which is the identical to the Aux ports, but named differently)


cjermain:
Looks good! Thanks for adding these commands.

Issue #184, opened 2019-03-14
Keithley 2400 - check_errors() does not cover delayed readings


hlgirard:
When using a Keithley 2400 sourcemeter with a Prologix adapter, it regularly happens that a reading returns an empty string. In these cases, it seems like the reading is delayed by the adapter because it will usually come concatenated with the next read as in the example below:

```
In [17]: sourcemeter.current                                                                          
Out[17]: 3.690513e-09

In [18]: sourcemeter.current                                                                          
Out[18]: ''

In [19]: sourcemeter.current                                                                          
Out[19]: '-2.182382E-09\n\n-1.803804E-09'
```

This is a problem of its own and will be the object of another issue but the current problem is that if one runs a command that invokes `check_errors()`, the string that is returned by the read function includes the previous measurement and therefore cannot be parsed correctly by the `check_errors()` function.

```
In [21]: sourcemeter.current                                                                          
Out[21]: -4.206825e-09

In [22]: sourcemeter.current                                                                          
Out[22]: ''

In [23]: sourcemeter.check_errors()                                                                   
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-23-56f515ac65b7> in <module>
----> 1 sourcemeter.check_errors()

/anaconda3/lib/python3.7/site-packages/pymeasure/instruments/keithley/keithley2400.py in check_errors(self)
    398         while code != 0:
    399             t = time.time()
--> 400             log.info("Keithley 2400 reported error: %d, %s" % (code, message))
    401             code, message = self.error
    402             if (time.time()-t)>10:

TypeError: %d format: a number is required, not str
```

It would make sense to render `check_errors()` more versatile by allowing it to catch such exceptions and clear the read buffer until it gets a correctly formatted error code.

hlgirard:
This issue was caused by the Prologix not returning upon receiving a termination character which caused concatenated answers to be returned over Serial. The cause was fixed in PR #185 by changing the read method to `++read eoi` such that the prologix never reads more than one line at a time.

PR #185, opened 2019-03-14
Prologix / Keithley2400 bug fixes

hlgirard:
While debugging the connection to my Keithley2400 through a Prologix Adapter, I found it useful to be able to select the serial timeout manually. This would imply a very simple change in the `__init__` of PrologixAdapter as such:

```
    def __init__(self, port, address=None, rw_delay=None, serial_timeout = 0.5, **kwargs):
        super().__init__(port, timeout = serial_timeout, **kwargs)
```

instead of the current code that has a hardcoded value of 0.5s
```
    def __init__(self, port, address=None, rw_delay=None, **kwargs):
        super().__init__(port, timeout = 0.5, **kwargs)
```

### DIFF
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -104,7 +104,7 @@ class PrologixAdapter(SerialAdapter):
 
         :returns: String ASCII response of the instrument
         """
-        self.write("++read")
+        self.write("++read eoi")
         return b"\n".join(self.connection.readlines()).decode()
 
     def gpib(self, address, rw_delay=None):
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -61,8 +61,8 @@ class PrologixAdapter(SerialAdapter):
 
     """
 
-    def __init__(self, port, address=None, rw_delay=None, **kwargs):
-        super().__init__(port, timeout=0.5, **kwargs)
+    def __init__(self, port, address=None, rw_delay=None, serial_timeout = 0.5, **kwargs):
+        super().__init__(port, timeout = serial_timeout, **kwargs)
         self.address = address
         self.rw_delay = rw_delay
         if not isinstance(port, serial.Serial):
--- a/pymeasure/instruments/keithley/buffer.py
+++ b/pymeasure/instruments/keithley/buffer.py
@@ -87,7 +87,7 @@ class KeithleyBuffer(object):
             sleep(interval)
             if should_stop():
                 return
-            if (time()-t)>10:
+            if (time()-t)>timeout:
                 raise Exception("Timed out waiting for Keithley buffer to fill.")
 
     @property


hlgirard:
Also found a bug in the `wait_for_buffer` function where the timeout value was hardcoded at 10 seconds despite the presence of an argument `timeout` which was ignored.

bilderbuchi:
Not at a pc to check right now, but what happens if you supply a timeout kwarg in the current implementation? Doesn't the timeout passed on to the super constructor via **kwargs override the default value?

hlgirard:
If one supplies a value for timeout in the current implementation, the following error occurs:

```
> adap = PrologixAdapter('/dev/tty.xyz', address = 24, timeout = 14)
TypeError                                 Traceback (most recent call last)
<ipython-input-2-6c67a67baf2b> in <module>
----> 1 adap = PrologixAdapter('/dev/tty.xyz', address = 24, timeout = 14)

/anaconda3/lib/python3.7/site-packages/pymeasure/adapters/prologix.py in __init__(self, port, address, rw_delay, **kwargs)
     63 
     64     def __init__(self, port, address=None, rw_delay=None, **kwargs):
---> 65         super().__init__(port, timeout=0.5, **kwargs)
     66         self.address = address
     67         self.rw_delay = rw_delay

TypeError: __init__() got multiple values for keyword argument 'timeout'
```

It looks like the timeout passed via **kwards doesn't override the default one.

bilderbuchi:
Maybe the timeout default could be dropped instead? I mean, why specifically 0.5 seconds? Failing that, I would pass it through with the same name to keep the API simpler. After all, port is also not renamed to serial_port. But maybe let the maintainer chime in now.

hlgirard:
@bilderbuchi Thank you for the input! 
The timeout property must be set because `SerialAdapter.read()` uses the `readlines()` method from the IO library that relies on the timeout to stop reading so I would advocate to keep a default. The 0.5s is debatable, I left it like I found it and it seems to work for me but if anyone has a better idea I am listening.

As for the naming, I wanted to avoid possible confusion with the Prologix adapter's own timeout property (set with `++read_tmo_ms <time>` and not currently implemented in pymeasure) which we might want to implement at some point too.

hlgirard:
While trying to speed up the minimum time between consecutive measurements, I noticed a couple of things:
- Using the Prologix's `++read eoi` command is potentially faster than `++read` because it will return when it receives a termination character while `++read` waits for `++read_tmo_ms` to expire. According to the documentation, it will timeout if it doesn't get a termination character so the change is safe as far as I can see. It also means that the Prologix shouldn't return concatenated values anymore which fixes #184 
- Then the timeout of the serial connection becomes the bottleneck between consecutive measurements (because of the use of `readlines()` which wait until timeout) hence the importance of allowing the user to adjust it. With my setup, I can reliably use a `serial_timeout` of 0.1 seconds but that might vary based on the instrument.

cjermain:
@hlgirard, nice work! I agree that the `++read eoi` command is an improvement. Thanks for fixing the timeout variables.

PR #187, opened 2019-04-04
Delta elektronica power supply sm70-45D

CasperSchippers:
**Ready for merging**

Added support for the Delta Elektronika SM70-45D Power supply (serves also as an example for other power supplies of Delta Elektronika as they use the same communication)

### DIFF
--- a/pymeasure/instruments/deltaelektronika/sm7045d.py
+++ b/pymeasure/instruments/deltaelektronika/sm7045d.py
@@ -30,7 +30,17 @@ import numpy as np
 
 
 class SM7045D(Instrument):
-    """ This is the class for the SM 70-45 D power supply """
+    """ This is the class for the SM 70-45 D power supply.
+
+    .. code-block:: python
+
+        source = SM7045D("GPIB::8")
+
+        source.ramp_to_zero(1)               # Set output to 0 before enabling
+        source.enable()                      # Enables the output
+        source.current = 1                   # Sets a current of 1 Amps
+
+    """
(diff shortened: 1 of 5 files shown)


pymeasure/instruments/deltaelektronika/sm7045d.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def enable(self):
+        # Disable remote shutdown, hence output will be enabled.
+        self.write("SO:FU:RSD 0")
+
+    def disable(self):
+        # Enables remote shutdown, hence input will be disabled.
+        self.write("SO:FU:RSD 1")
+
+    def ramp_to_current(self, target_current, current_step):
```
cjermain: I think it is worth setting a default for `current_step`.

CasperSchippers: That's a good suggestion, i implemented it

cjermain:
Looks good. Thanks for improving the methods.

mjgardes:
I think there is a typo in sm7045d.py at line 72

```
    max_current = Instrument.control(
        "SO:CU:MA?", "SO:VO:MA %g",
```

should probably be
```
    max_current = Instrument.control(
        "SO:CU:MA?", "SO:CU:MA %g",
```

CasperSchippers:
I think you're right, @mjgardes. Thanks for the remark. I will include it in an open PR that (by now) includes a number of instrument fixes (#418)

PR #192, opened 2019-04-22
Fix DeprecationWarnings

bilderbuchi:
This fixes various [DeprecationWarnings](https://travis-ci.org/ralph-group/pymeasure/jobs/523174766#L768-L810) caused by using regex with non-raw strings.

I also propose (for you to decide) to remove the unused --runslow argument to pytest, as it uses a deprecated pytest API:

>  /home/travis/build/ralph-group/pymeasure/tests/experiment/test_workers.py:41: PytestDeprecationWarning: the `pytest.config` global is deprecated.  Please use `request.config` or `pytest_configure` (if you're a pytest plugin) instead.
    not pytest.config.getoption("--runslow"),

If needed can easily be put back [according to the pytest docs](https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option).

Note that to have a passing CI, this PR also includes the changes from #191.

### DIFF
--- a/requirements/conda-3.4.yml
+++ b/requirements/conda-3.4.yml
@@ -20,4 +20,4 @@ dependencies:
 - wheel=0.29.0=py34_0
 - pip:
   - pyvisa==1.8
-  - pytest-qt==2.2.1
+  - pytest-qt==2.4.0
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,8 +23,3 @@
 #
 
 import pytest
-
-
-def pytest_addoption(parser):
-    parser.addoption("--runslow", action="store_true",
-                     help="run slow tests")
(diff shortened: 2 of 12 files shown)


requirements/conda-3.4.yml
```diff
@@ -20,4 +20,4 @@ dependencies:
 - wheel=0.29.0=py34_0
 - pip:
   - pyvisa==1.8
-  - pytest-qt==2.2.1
```
cjermain: I'm surprised that these changes continue to show as part of the diff after merging #191.

tests/conftest.py
```diff
@@ -23,8 +23,3 @@
 #
 
 import pytest
-
-
-def pytest_addoption(parser):
```
cjermain: I agree it's worth cleaning up this option.

requirements/conda-3.4.yml
bilderbuchi: Possibly this only gets refreshed if a new commit comes in?

PR #196, opened 2019-05-20
Branch psu viewer

sumatrae:
Added a keysight5767n GUI control example


cjermain:
@sumatrae, thanks for putting up this PR. Can you provide more context around this example? How do you envision it fitting in to the existing documentation?

bilderbuchi:
@CasperSchippers do you think it reasonable to update this stale PR for inclusion, or should we close it?

CasperSchippers:
I'm a bit reluctant to put effort in updating a PR that's been stale for 3 years (seeing how sparse our time often is and seeing that it might be a lot of work to update it/difficult to make it work correctly without having the required instrument), so in that sense I think I'd vote for closing, unless the original author has any interest in putting in effort

bilderbuchi:
Ok, then lets close it. If there's unexpected interest we can reopen it.

PR #200, opened 2019-05-28
Keithley 2750 Instrument

dennisfeng2:
Added the beginning of the new instrument: Keithley 2750 Instrument


pymeasure/instruments/keithley/keithley2750.py
```diff
@@ -0,0 +1,95 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Instrument
+
+
+def clean_closed_channels(output):
```
dennisfeng2: I wrote this function to handle the output from the command `self.ask(":ROUTe:CLOSe?")`, and clean it up into a list of integers.
Here are some example outputs from `self.ask(":ROUTe:CLOSe?")`:

"(@)"                                                         --> if no channels are open. Note that it's a string and not a list
"(@101)"                                                    --> if only 1 channel is open. Note that it's a string and not a list
["(@101", "105)"] .                                    --> if only 2 channels are open
["(@101", 102.0, 103.0, 104.0, "105)"]     --> if more than 2 channels are open

If you think the implementation is a bit too clunky, could you suggest a different way to implement it? Thanks.

pymeasure/instruments/keithley/keithley2750.py
```diff
@@ -0,0 +1,95 @@
...
+        # Note: if `MULTiple` is omitted, then the specified channel will close, but all other channels will open.
+        self.write(":ROUTe:MULTiple:CLOSe (@{channel})".format(channel))
+
+    def open_all(self):
+        """ Opens (disconnects) all the channels on the switch matrix.
+
+        :return: None
+        """
+        self.write(":ROUTe:OPEN:ALL")
+        # Channel 133 is the switch between Mux 1 and Mux 2; we want it to be open. From testing the instrument,
```
dennisfeng2: This is what I found from testing my instrument. Let me know what you think about this, e.g. if it's better to leave it out for the general case

dennisfeng2:
I've noticed that there are some failed tests/checks with this pull request, but after taking a look at the error messages, it's still unclear to me whether these issues are coming from bugs in the pull request itself. For example, some of the errors involve import errors of VXI11Adapter and HP34410A, which my code doesn't seem to touch at all. Could someone point out to me why these errors are occurring, or anything I can fix from my end, so we can make progress on this PR?

pymeasure/instruments/keithley/keithley2750.py
cjermain: These examples should be added to unit tests for this function, so that we can ensure that the function continues to perform the same way in the future. See the [docs section on unit testing](https://pymeasure.readthedocs.io/en/latest/dev/contribute.html#unit-testing).

pymeasure/instruments/keithley/keithley2750.py
cjermain: I think we should leave it out for the general case.

pymeasure/instruments/keithley/keithley2750.py
```diff
@@ -0,0 +1,95 @@
+#
```
cjermain: All instruments are required to have accompanying documentation. See the [docs for adding documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#adding-documentation), and feel free to based it off of the other Keithley instruments..

cjermain:

pymeasure/instruments/keithley/keithley2750.py
dennisfeng2: okay, done.

dennisfeng2:
Okay, addressed the comments, and also merged with the latest master. Could you take a look?

cjermain:
@dennisfeng2, sorry for the slow response. This looks good. Will merge after running through the CI tests.

cjermain:
Thanks @dennisfeng2!

PR #203, opened 2019-05-30
Keithley 2400 instrument property modifications

dennisfeng2:
Here, I modify `source_enabled` to also be "settable" and I add bounds to `source_current`. I also tested these on the actual instrument.


dennisfeng2:
I've noticed that there are some failed tests/checks with this pull request, but after taking a look at the error messages, it's still unclear to me whether these issues are coming from bugs in the pull request itself. For example, some of the errors involve import errors of VXI11Adapter and HP34410A, which my code doesn't seem to touch at all. Could someone point out to me why these errors are occurring, or anything I can fix from my end, so we can make progress on this PR?

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -73,9 +73,15 @@ class Keithley2400(Instrument, KeithleyBuffer):
         values={'current':'CURR', 'voltage':'VOLT'},
         map_values=True
     )
-    source_enabled = Instrument.measurement("OUTPUT?",
-        """ Reads a boolean value that is True if the source is enabled. """,
-        cast=bool
+
+    source_enabled = Instrument.control(
```
cjermain: I generally prefer to have one way to do something, but I think this is a useful addition.

cjermain:

cjermain:
Please merge the latest master branch to fix the tests.

dennisfeng2:
Apologies for the late response. Finished merging the latest master branch!

cjermain:
@dennisfeng2, thanks for updating these PRs. It looks like the builds are having problems for Python 3.7 (which is not related to your work). I will take a look soon, and may proceed to merge these without that build.

PR #205, opened 2019-06-14
Prevent line breaks in CSV header

feph:
Add escaping of the string representation of Parameter objects in CSV headers.

---
If the `__str__ `representation of a `Parameter `object (I was trying to pass numpy arrays to a procedure) contains line breaks (like `numpy.ndarray` from a certain length) these breaks will show up in the resulting CSV file and subsequently break file parsing. By using `str(parameter).encode("unicode_escape").decode("utf-8")` special chars will be escaped and preserved in the file.


cjermain:
@feph, thanks for posting this PR! Can you add a unit test in `tests/experiment/test_results.py`? This will help to document the solution, as well as ensure it does not regress.

feph:
@cjermain I added `test_regression_param_str_should_not_include_newlines()` to `test_results.py`, which fails on my machine without the bugfix and passes afterwards. I'm new to pytest and mocking so maybe this can be implemented more elegantly.

bilderbuchi:
2 suggestions: 
* I think you should use [`tempfile.TemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile) instead of `mktemp`. 
* Maybe even `tempfile.SpooledTemporaryFile`, as this would keep the file in memory instead of disk, which probably makes the test faster (but check this first).
* Pytest offers a [tmpdir](https://docs.pytest.org/en/latest/tmpdir.html) fixture, which gives you an easy way to generate temporary places for file storage which get cleaned up after testing.

cjermain:
@feph, thanks for adding the test. I agree with @bilderbuchi that it is nice to use `tempfile` or `tmpdir`. I generally prefer the pytest fixture approach.

cjermain:
@feph, I am interested to include your bugfix in the upcoming release. Are you able to make the test updates?

feph:
Hi,
sorry for the delay. I modified the test case to use `tmpdir`.

cjermain:
Great, thanks @feph for adding this bug fix and unit test!

PR #207, opened 2019-06-30
Bug fixes for keithley2400

Fattotora:


cjermain:
Good catch. Thanks @Fattotora!

PR #217, opened 2019-07-31
Dev/afg3152 c

StePhanino:
Hi, thanks a lot for pymeasure package. I'm adding a new instrument: Tektronix AFG3152C (2-channels Arbitrary Function Generator). This is a first test with some basic functions. I have a major question: how do I deal with the 2 channels using the Instrument.Control property?

- First approach: two series of commands, one for each channel with validators to check inputs. But I'm not able to set properly the validator for the amplitude since limits depends on the selected channel unit (voltage or dbm)

- Second approach: normal self.write based function that allows a complete control on inputs ranges.

What do you suggest?

The class is working with AFG3152C instrument connected through a NI-GPIB-USB cable to a Win7 notebook.

By doing 'py -3 setup.py test' I have an error:  _[WinError 87] The parameter is incorrect_ . Complete output in the attached file. It's a windows 

Thanks
stefano

[Output_pytest.txt](https://github.com/ralph-group/pymeasure/files/3453021/Output_pytest.txt)


cjermain:
@StePhanino, thanks for putting up this new instrument. I think the most ideal pattern for the multi-channel problem is to use object composition. This means having individual objects for each channel.
```python
afg.ch1.offset = 0.4
```
I don't think this has been done yet in the package though. Here is an example of how to get started.
```python
class Channel(object):

    amplitude = Instrument.measurement(
        "voltage:amplitude?",
        ...
    )

    offset = Instrument.control(
        "voltage:offset?", "voltage:offset %e",
        ...
    )

    def __init__(self, instrument, number):
        self.instrument = instrument
        self.number = number

    def write(self, command):
        self.instrument.write("source%d:%s" % (self.number, command))

    def ask(self, command):
        self.instrument.ask("source%d:%s" % (self.number, command))

    def read...

    def enable(self):
        self.instrument.write("output%d:state on" % self.number)

    def disable...


class AFG3152C(Instrument):

        super(AFG3152C, self).__init__(
            adapter,
            "Tektronix AFG3152C arbitrary function generator",
            **kwargs
        )
        self.ch1 = Channel(self, 1)
        self.ch2 = Channel(self, 2)

```

StePhanino:
Hi, I followed your suggestion in this new version. 
Checked on our AFG3152C: it works in both setting and reading.

pymeasure/instruments/tektronix/afg3152c.py
```diff
@@ -23,66 +23,103 @@
 #
 from pymeasure.instruments import Instrument, RangeException
 from pymeasure.instruments.validators import strict_range, strict_discrete_set
-from math import sqrt,log10
 
 class Channel(object):
-
-        SOURCE_VALUES= [1,2]
+        SHAPES={
```
cjermain: Please follow PEP8 style (putting whitespaces around operators). Note that this does not apply to keyword arguments. https://www.python.org/dev/peps/pep-0008/#other-recommendations

pymeasure/instruments/tektronix/afg3152c.py
```diff
@@ -23,66 +23,103 @@
...
             This property can be set.""",
             validator=strict_range,
             values=FREQ_LIMIT
         )
+        duty=Instrument.control(
+            "pulse:dcycle?","pulse:dcycle %d",
+            """ A floating point property that represents the duty cycle of pulse.
+            This property can be set.""",
+            validator=strict_range,
+            values=[0.001,99.999]
```
cjermain: Following PEP8, there should be a whitespace between numbers in a list: 
```
values=[0.001, 99.999]
````

**Comment by StePhanino** [comment]
Hi, thanks. I'm upgrading the code following PEP8. At the same time I was thinking on the limits in the amplitude function
```python
amplitude = Instrument.control(
            "voltage:amplitude?","voltage:amplitude %e",
            """ A floating point property that controls the output power
            in Vpp, Vrms or dBm. This property can be set.""",
            validator =ch_validator,
            values = AMPLITUDE_LIMIT,
        )
```
where the limit is dependent on the amplitude unit of the channel. I was trying to create a new validator in the class Channel, something like this one
```python
def ch_validator(self, value):
            unit=self.ask("voltage:unit?") #function to retrieve the amplitude unit of channel
            values=AMPLITUDE_LIMIT[unit]
            if min(values) <= value <= max(values):
                return value
            else:
                raise ValueError('Value of {:g} is not in range [{:g},{:g}]'.format(
                    value, min(values), max(values)
                ))
```
but I'm unable to have something working. Sorry but I'm still learning python!
Thanks

cjermain:
It sounds like for the amplitude you want to use the `strict_range` validator ([see the source code](https://github.com/ralph-group/pymeasure/blob/master/pymeasure/instruments/validators.py#L26)). That would be used as follows (allowing any value between -1.0 and 1.0 in the example).
```python
amplitude = Instrument.control(
    "voltage:amplitude?", "voltage:amplitude %e",
    """ A floating point property that controls the output power
    in Vpp, Vrms or dBm. This property can be set.""",
    validator=strict_range,
    values=[-1.0, 1.0],
)
```

bilderbuchi:
I think this is more about having dynamic (i.e., changeable at runtime) `values` for the limits of the validator. I could have also used this previously, for a PSU with adjustable range/mode, but did not find a good way to do it. this is because of the way the control etc  properties are constructed as class, not instance attributes, during import time.

StePhanino:
Yes, it's that @bilderbuchi: `values` should dinamically change following the amplitude unit. Maybe the easiest way is to have 3 different `Instrument.control`, one for each amplitude unit: `amp_vpp`, `amp_dbm` and `amp_vrms`. The final code for the experiment would be more readable...what do you think?

cjermain:
That is an interesting point. It seems like we need validators that are aware of the underlying units. I think we can postpone solving this for a future PR.

Please add yourself to the authors files: `docs/about/authors.rst` and `AUTHORS.rst`, and I will merge.

bilderbuchi:
More generally, validators that are aware of dynamically changing properties, which I found not a good way to implement last time I looked. Python properties being attached to a class, not an instance, seem to make this kind of thing tricky to realise.

StePhanino:
Sorry, I've forgotten to upgrade the branch to the master...now I'm upgrading and repush.

cjermain:
Thanks @StePhanino!

PR #226, opened 2019-09-17
add read_bytes method for VISA communication

moritzj29:
For some instruments it is handy to read a specified number of bytes from the data output buffer. This just passes the corresponding method of the underlying visa module all the way to the instrument level.

This is my first ever pull request, I hope to not not miss out on something...

### DIFF
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -87,11 +87,20 @@ class VISAAdapter(Adapter):
 
     def read(self):
         """ Reads until the buffer is empty and returns the resulting
-        ASCII respone
+        ASCII response
 
         :returns: String ASCII response of the instrument.
         """
         return self.connection.read()
+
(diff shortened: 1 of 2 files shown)


bilderbuchi:
One thing that comes to mind: what of other adapters (that don't have read_bytes)? Wouldn't they raise/fail if they used the new function defined in Instrument?

moritzj29:
This is probably true... I didn't think of that.

To circumvent that, one could leave Instrument untouched and just add the visa method. Then one could access `read_bytes` via `self.adapter.read_bytes()` instead of `self.read_bytes()`. In my specific usecase, this would be possible since the instrument can only be interfaced with visa.

What do you think?

moritzj29:
I will close this PR for now and open a new one (since choosing my master branch for the PR was a bad idea...) with the suggested change in visa only. I hope this is fine

PR #227, opened 2019-09-18
add read_bytes method for VISA communication

moritzj29:
As discussed in previous #226, adding `read_bytes` to `Instrument` as well would break `Instrument` 's functionality when using adapters that do not support `read_bytes` method. Therefore just adding the `read_bytes` methods in the `visa` library.

If needed `read_bytes` can be accessed from a particular instrument via `self.adapter.read_bytes()` instead of the shorter `self.write()` etc.

### DIFF
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -87,11 +87,20 @@ class VISAAdapter(Adapter):
 
     def read(self):
         """ Reads until the buffer is empty and returns the resulting
-        ASCII respone
+        ASCII response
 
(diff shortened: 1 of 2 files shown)


pymeasure/adapters/visa.py
```diff
@@ -87,12 +87,21 @@ def write(self, command):
 
     def read(self):
         """ Reads until the buffer is empty and returns the resulting
-        ASCII respone
+        ASCII response
```
cjermain: Nice spelling catch!

PR #228, opened 2019-09-20
Measurement sequencer

CasperSchippers:
**Ready to be merged**

This PR proposes to add a new widget to pymeasure that allows the user to generate a sequence of measurements by varying one (or more) parameters.

The module by default uses the available input parameters to generate a sequence but overrules one (or more) of the input parameters with a defined sequence.
The sequence is entered in the form of a python (or numpy) array, where you can use a number of expressions to generate simple or more complex array structures.
Moreover, you have the option to load a sequencer from a simple text file.
Finally, also an option to add a scroll-bar to the inputs area is implemented (for longer input lists or when there is less space for the inputs list due tot the sequencer)

This PR includes an example of how to include the sequencer in a managed window and a example text file with a pre-defined sequence that can be loaded into the widget.

I'm very much open for ideas / comments / remarks, so please let me know what you think.

**To do:**
- [x] Streamline code
- [x] Document code
- [x] Write tutorial
- [x] Add a optional scroll area around the inputs


pymeasure/display/widgets.py
```diff
@@ -418,3 +422,352 @@ def update_plot(self, filename):
...
+        if inputs is not None:
+            self._inputs = inputs
+        else:
+            self._inputs = self._parent.displays
+
+        self._get_properties()
+        self._layout()
+        self._add_to_interface()
+
+    def _get_properties(self):
```
cjermain: Please add docstrings to each method for clarity.

pymeasure/display/widgets.py
```diff
@@ -418,3 +422,352 @@ def update_plot(self, filename):
                 new_item = QtGui.QTreeWidgetItem([param.name, str(param)])
                 self.preview_param.addTopLevelItem(new_item)
             self.preview_param.sortItems(0, QtCore.Qt.AscendingOrder)
+
+
+SAFE_FUNCTIONS = {
```
cjermain: It would be great to give a clear description of how to use these in the documentation.

moritzj29:
I just tried the example example code because this feature looks very interesting to me. Unfortunately, upon queuing a measurement the execution breaks with 

> ManagedWindow can not make a Procedure without a InputsWidget type

Did I miss out on something? The example should run without any modifications, right?
Thanks for your work, I'm really looking forward to this!

cjermain:
@CasperSchippers, let me know when you are ready for a review. I think this is going to be a great feature.

CasperSchippers:
> @CasperSchippers, let me know when you are ready for a review. I think this is going to be a great feature.

I think it is ready!

pymeasure/display/widgets.py
```diff
@@ -418,3 +422,403 @@ def update_plot(self, filename):
...
+    """
+    Widget that allows to generate a sequence of measurements with varying
+    parameters. Moreover, one can write a simple text file to easily load a
+    sequence.
+
+    Currently requires a queue function of the ManagedWindow to have a
+    "procedure" argument.
+    """
+
+    MAXDEPTH = 10
```
cjermain: This is a good limit. :)

pymeasure/display/widgets.py
```diff
@@ -418,3 +422,403 @@ def update_plot(self, filename):
...
+        :param name: Name of the to-be-interpreted string, only used for
+            error messages.
+        :param depth: Depth of the to-be-interpreted string, only used
+            for error messages.
+        """
+
+        evaluated_string = None
+        if len(string) > 0:
+            try:
+                evaluated_string = eval(
```
cjermain: I think this is a good use of `eval`. I think this is a safe usage because you are blocking globals and setting the locals to the numpy functions.

pymeasure/display/widgets.py
```diff
@@ -418,3 +422,403 @@ def update_plot(self, filename):
...
+                    string, {"__builtins__": None}, SAFE_FUNCTIONS
+                )
+            except TypeError:
+                log.error("TypeError, likely a typo in one of the " +
+                          "functions for parameter {}, depth {}".format(
```
cjermain: I think it is worth quoting the parameter here: `"functions for parameter '{}'...` to keep it separated. The same goes for the below error messages.

docs/tutorial/gui_sequencer_example_sequence.txt
```diff
@@ -0,0 +1,3 @@
+- "Delay Time", "arange(0.25, 1, 0.25)"
+-- "Random Seed", "[1, 4, 8]"
+-- "Random Seed", "arange(10, 100, 10)"
```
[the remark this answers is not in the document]
CasperSchippers: Good suggestion, fixed this in both the example and the tutorial

pymeasure/display/widgets.py
CasperSchippers: Fixed this

PR #230, opened 2019-10-30
Adding AgilentN9320A spectrum analyser

StePhanino:
Hola, here a new isntrument of the agilent family. There is one point i'm not totally sure... it's the use of OPC. In any case the delay time can be set by using the `D_FACTOR` factor: this allows to wait until average done. But it would be nice to have a check on `*OPC?`

```python 
while True:
            if self.opc() == 1:
                break
            sleep(2)
```
I've tested the class and it's working.
Thanks


pymeasure/instruments/agilent/agilentN9320A.py
```diff
@@ -0,0 +1,262 @@
...
+                               #peak function
+    DELAY = 0.2                #A delay in second can be useful
+
+
+    start_frequency = Instrument.control(
+        ":SENS:FREQ:STAR?", ":SENS:FREQ:STAR %e Hz",
+        """ A floating point property that represents the start frequency
+        in Hz. This property can be set.
+        """,
+		validator=strict_range,
```
cjermain: Please correct the indentation here to be consistent. I would suggest that the docstring ends on the previous line:
```
    in Hz. This property can be set. """,
    validator=strict_range,
    values=FREQ_LIMIT,
)
```

pymeasure/instruments/agilent/agilentN9320A.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+    res_bw = Instrument.control(
+        ":SENS:BAND:RES?", ":SENS:BAND:RES %e",
+        """ A floating point property that represents the resolution
+        bandwidth in Hz. This property can be set.
+        """,
+        validator=strict_discrete_set,
+        values=RES_LIMIT
+    )
+    vid_bw = Instrument.control(
```
cjermain: I think `video_bandwidth` would be a clearer name for this control.

pymeasure/instruments/agilent/agilentN9320A.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+    att = Instrument.control(
+        ":SENS:POW:RF:ATT?", ":SENS:POW:RF:ATT %d",
+        """ A floating point property that represents the input attenuator
+        in dB. This property can be set.
+        """,
+        validator=strict_range,
+        values=ATT_LIMIT
+    )
+    res_bw = Instrument.control(
```
cjermain: I think `resolution_bandwidth` would be a clearer name for this control.

pymeasure/instruments/agilent/agilentN9320A.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+
+    def __init__(self, resourceName, **kwargs):
+        super(AgilentN9320A, self).__init__(
+            resourceName,
+            "Agilent AgilentN9320A Spectrum Analyzer",
+            **kwargs
+        )
+    def init_imm(self):
+        self.write("INIT:IMM")
```
cjermain: Please add docstrings to each method. Also consider making the names more human readable, along the same lines as the comments above.

pymeasure/instruments/agilent/agilentN9320A.py
```diff
@@ -0,0 +1,262 @@
...
+        It can also look the first peaks at both left and right
+        around the highest one.
+        """
+        self.avg('ON')
+        self.avg_set(avg)
+        sleep(self.DELAY)
+        self.init_single()
+        self.init_imm()
+
+        print('Averaging for %.1f seconds' % (avg*self.sweep_time))
```
cjermain: Print statements should not be used in the PyMeasure library. Instead, use a Python logger. Here is an example from another instrument:
https://github.com/ralph-group/pymeasure/blob/2dccdb4091a6a66f9d254379866bc6bd13d7a14a/pymeasure/instruments/anritsu/anritsuMS9710C.py#L36-L37
https://github.com/ralph-group/pymeasure/blob/2dccdb4091a6a66f9d254379866bc6bd13d7a14a/pymeasure/instruments/anritsu/anritsuMS9710C.py#L297-L303
Notice that `log.debug` is used for high-volume messages that typically won't be displayed.

pymeasure/instruments/agilent/agilentN9320A.py
```diff
@@ -0,0 +1,262 @@
...
+        sleep(avg*self.D_FACTOR*self.sweep_time)
+
+        #I'm not sure about this part
```
cjermain: This seems like it should be broken into a separate method, and limited in time. I would suggest following the work already done in the Keithley buffer code:
https://github.com/ralph-group/pymeasure/blob/2dccdb4091a6a66f9d254379866bc6bd13d7a14a/pymeasure/instruments/keithley/buffer.py#L73-L91

pymeasure/instruments/agilent/agilentN9320A.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def max_x(self, number=1):
+        """Returns the frequency in Hz at marker position"""
+        return float(self.ask("CALC:MARK%s:X?" % number))
+
+    def max_y(self, number=1):
+        """Returns the amplitude in dBm at marker position"""
+        return format(float(self.ask("CALC:MARK%s:Y?" % number)), '.4f')
+
+    def peak(self, number=1, avg=1, center=False, lr=False):
```
cjermain: This method should be broken into smaller more manageable pieces (other methods), so that it can be read and understood more easily.

cjermain:
@StePhanino, any luck addressing the review comments? It would be great to include this in the next release in a few weeks.

StePhanino:
@cjermain sorry but it has been a strong/difficult period. I'm working on the upgrades.

pymeasure/instruments/agilent/agilentN9320A.py
StePhanino: Unfortunately, N9320A is not as sophisticated as keithley. I tried to by-pass this with a redefinition of the timeout time for OPC? command. It should be the same.

bilderbuchi:
don't worry about delays, work on this at a pace you are comfortable with!

StePhanino:
I added a few methods for triggering, averaging, error check, peak and marker utilities.

StePhanino:
Hola @bilderbuchi. Do you know why tests failed? I just pushed to this branch the newest commits on the main branch of pymeasure..thanks

bilderbuchi:
This is a failure of the docs CI job, I [fixed this quite recently](https://github.com/ralph-group/pymeasure/commit/de39dd1bef9059e5f9dea8552032923c26cf51fa) - merging/current master should fix that.

StePhanino:
@bilderbuchi thanks. It's working now.

StePhanino:
@cjermain and @bilderbuchi : what do you think about the updates? thanks

pymeasure/instruments/agilent/agilentN9320A.py
```diff
@@ -0,0 +1,296 @@
...
+        around the highest one. """
+
+        self.init_single()
+        self.average_on()
+        self.average_number(avg)
+
+        sleep(self.DELAY)
+
+        if self.adapter.connection.timeout < (avg*self.D_FACTOR*self.sweep_time)*1000:
+            self.adapter.connection.timeout = (avg*self.D_FACTOR*self.sweep_time)*1000
```
bilderbuchi, line 251: Is this timeout modification purely for the opc call? If yes, it should be reset to the original value after you are finished, otherwise you permanently set the timeout.

pymeasure/instruments/agilent/agilentN9320A.py
```diff
@@ -0,0 +1,296 @@
...
+    def set_opc_sqr(self):
+        """ Set the instrument to generate SRQ when operation is complete:
+        remember to run a *CLS before the command that you want
+        to sync/wait for. Timeout in seconds. """
+        self.write("*ESE 1;*SRE 32")
+
+    def opc(self):
+        """ It returns 1 when operation complete. You may need to change the
+        visa timeout (in ms) for long lasting operations. """
+        self.ask("*OPC?")
```
bilderbuchi, line 164: This function always returns `None`, irrespective of the result of the ask() mentioned in the docstring. How do you tell the difference between a successful and an aborted wait?
I don't know how the device handles long operations, but it might be useful to include custom timeout handling here instead of below in peak_output. you can look how other instruments in pymeasure deal with `OPC` instructions.

bilderbuchi (changes requested):

bilderbuchi:
Hey! There is one request from Colin still pending about indentation fixes, and I just looked at the peak and opc methods that you remarked on and left a couple of comments.

bmoneke:
Hey @StePhanino do you intend to finish this instrument or not?
In the latter case, I'd close the pull request (the code remains in this PR for future works).

bmoneke:

PR #232, opened 2019-11-01
Correct spelling

mivade:


cjermain:
Nice catch @mivade! Thanks.

PR #236, opened 2019-11-22
Add validator for discrete range of values

moritzj29:
Add validator to check if value is within a range of values and is a multiple of a given stepsize. This is useful for values which have a resolution limit. 
E.g. setting a sampling interval to 0.015s is allowed for a range of 0.01s to 2s. But if the maximum resolution of the instrument is 0.1s, it makes no sense to "pretend" higher accuracy in setting the value. This validator checks if the given value is a multiple of the resolution value or stepsize given.

### DIFF
--- a/pymeasure/instruments/validators.py
+++ b/pymeasure/instruments/validators.py
@@ -39,6 +39,29 @@ def strict_range(value, values):
             value, min(values), max(values)
         ))
 
+def strict_discrete_range(value, values, step):
+    """ Provides a validator function that returns the value
+    if its value is less than the maximum and greater than the
+    minimum of the range and is a multiple of step. 
+    Otherwise it raises a ValueError.
+
+    :param value: A value to test
+    :param values: A range of values (range, list, etc.)
+    :param step: Minimum stepsize (resolution limit)
+    :raises: ValueError if the value is out of the range
+    """
+    if (min(values) <= value <= max(values)): 
+        if (value/step).is_integer():
+            return value
+        else:
+            raise ValueError('Value of {:g} is not a multiple of {:g}'.format(
+                value, step
+            ))
+    else:
+        raise ValueError('Value of {:g} is not in range [{:g},{:g}]'.format(
+            value, min(values), max(values)
+        ))
+
 
 def strict_discrete_set(value, values):
     """ Provides a validator function that returns the value
--- a/pymeasure/instruments/validators.py
+++ b/pymeasure/instruments/validators.py
@@ -22,6 +22,8 @@
 # THE SOFTWARE.
 #
 
+from decimal import Decimal
+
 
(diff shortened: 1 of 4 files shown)


pymeasure/instruments/validators.py
```diff
@@ -40,6 +40,30 @@ def strict_range(value, values):
...
+    minimum of the range and is a multiple of step.
+    Otherwise it raises a ValueError.
+
+    :param value: A value to test
+    :param values: A range of values (range, list, etc.)
+    :param step: Minimum stepsize (resolution limit)
+    :raises: ValueError if the value is out of the range
+    """
+    if (min(values) <= value <= max(values)):
+        if (value/step).is_integer():
```
cjermain: I think it is preferable to use the modulus to determine if it is a multiple: `value % step == 0`

pymeasure/instruments/validators.py
```diff
@@ -40,6 +40,30 @@ def strict_range(value, values):
         ))
 
 
+def strict_discrete_range(value, values, step):
```
cjermain: Please add tests in [test_validators.py](https://github.com/ralph-group/pymeasure/blob/master/tests/instruments/test_validators.py) to cover all use cases and edge cases.

cjermain:
Please also add yourself to the Authors files:
https://github.com/ralph-group/pymeasure/blob/master/AUTHORS.txt
https://github.com/ralph-group/pymeasure/blob/master/docs/about/authors.rst

pymeasure/instruments/validators.py
moritzj29: This was my first intention as well, but I recognized that the modulus does not work very reliably with decimals:
`0.1 % 0.02 == 0` returns `False`
whereas the proposed
`(0.1/0.02).is_integer()` gives `True`
Do you have any idea how to fix this? It is probably due to some accuracy limit in number representation, since the modulus gives a very small (E-18) close to zero value.
The chosen variant is not as elegant as using the modulus but it works :D

bilderbuchi: I thinks this is mainly down to floating point arithmetic. `math.isclose()` with appropriate tolerances is probably the safest/cleanest way to handle this.

moritzj29: thanks for your comment! Upon your suggestion I tried `math.isclose()` and found a counterexample: `math.isclose(0,0.6%0.1,rel_tol=1e-9,abs_tol=1e-2)` evaluates to `False` despite the "huge" absolute tolerance. Are there any obejctions against the `Decimal` type solution? Computing time? As far as I understand this datatype is intended especially for such scenarios.

cjermain: I think `Decimal` is a good solution for this use case. Thanks!

Issue #99, opened 2017-07-20
Add tests for each instrument


bilderbuchi:
Migrated from TODO

cjermain:
As we transition instruments to InstrumentKit (#53, #94), I think this will no longer be critical. For now, let's hold off on developing further instrument tests.

bilderbuchi:
Well, for IK you will also need tests if you want to integrate new Instruments, but you have a point.

PR #238, opened 2019-12-16
driver and test for chroma PSU Chroma62024p6008

tinix84:
this is a driver for chroma power supply
can be extended to many different model in the future adapting the methods used in python-ivi https://github.com/python-ivi/python-ivi/tree/master/ivi/chroma


pymeasure/instruments/chroma/chroma62000p.py
```diff
@@ -0,0 +1,156 @@
...
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import (
+    truncated_range
+)
+from time import sleep
+import numpy as np
+
+
+
+class Chroma62024p6008(Instrument):
```
cjermain: We should be consistent with the naming, and go with `Chroma62024P6008` here and in the file names.

pymeasure/instruments/chroma/chroma62000p.py
```diff
@@ -0,0 +1,156 @@
...
+        """,
+        validator=truncated_range,
+        values=[0, 8]
+    )
+
+    get_output_status = Instrument.measurement("CONF:OUTP?", "This function queries the output state of the instrument.")
+
+    output_voltage_level = Instrument.control(
+        "SOUR:VOLT?", "SOUR:VOLT %g",
+        """This function control the output voltage limit of the instrument.
```
cjermain: Minor readability improvement here and below: "This function controls the".

pymeasure/instruments/chroma/chroma62000p.py
```diff
@@ -0,0 +1,156 @@
...
+    output_voltage = Instrument.measurement("FETC:VOLT?", "DC output voltage, in Volts")
+
+    output_current = Instrument.measurement("FETC:CURR?", "DC output voltage, in Amps")
+
+    output_slew_rate = Instrument.control(
+        "SOUR:VOLT:SLEW?", "SOUR:VOLT:SLEW %g",
+        """This function control the output voltage slew rate.
+         """,
+        validator=truncated_range,
+        values=[0.001, 10] # 62012P-600-8: 0.001V/ms - 10V/ms
```
cjermain: Thanks for the specific documentation here, this is a useful comment.

pymeasure/instruments/chroma/chroma62000p.py
```diff
@@ -0,0 +1,156 @@
...
+    )
+
+    def output_enable(self):
```
cjermain: I think it would be more consistent to use `enable_output` and `disable_output` for these function names.

pymeasure/instruments/chroma/chroma62000p.py
```diff
@@ -0,0 +1,156 @@
...
+        self.write("CONF:OUTP ON")
+
+    def output_disable(self):
+        self.write("CONF:OUTP OFF")
+
+    def ramp_to_current(self, current, steps=25, duration=0.5):
```
cjermain: This code is similar to that in the Keithley2400 instrument. https://github.com/ralph-group/pymeasure/blob/f9c4012397e5ea426732d8198f3a8115335de64c/pymeasure/instruments/keithley/keithley2400.py#L417-L432

It would be nice to keep the arguments the same. I think we chose to set the pause to avoid it being set too low. I'm certainly open to using a direct duration, but we should cover that case. It would be nice to refactor this out so all instruments can use a common function, but that can happen in a future PR.

tests/instruments/test_chroma62024p6008.py
```diff
@@ -0,0 +1,12 @@
+from pymeasure.instruments.chroma.chroma62000p import Chroma62024p6008
```
cjermain: Please add the license to this file.

cjermain: This file should go in a mirror directory to the source: `tests/instruments/chroma/test_chroma62000p.py`

So far we don't have good tests around instruments, since we don't have access to the instruments. We could do mock tests, but there is little value in my opinion to writing mocks without actually testing them at the instrument level. I would suggest that you put this test in a function, and behind a flag -- which we can set to ignore (e.g. `@pytest.mark.skip`). Happy to discuss this further, but I think this is beyond the scope of this PR -- I'm open to accepting it without the test file for now.
http://doc.pytest.org/en/latest/skipping.html#skipping-test-functions

tinix84: should I provide an example how to use it instead of test?

bilderbuchi: > So far we don't have good tests around instruments, since we don't have access to the instruments. We could do mock tests, but there is little value in my opinion to writing mocks without actually testing them at the instrument level.

@cjermain I always found the instrumentkit way of [checking for the expected protocol](https://instrumentkit.readthedocs.io/en/latest/devguide/testing.html#expected-protocols) pretty neat - you can define sends and expected responses (as are often found in examples in the device manual), add the instrument code that executes the operation in question, and confirm that your defined instrument emits the correct commands.
What do you think? should that be added to pymeasure, too? maybe cf #99.

tests/instruments/chroma/test_chroma62000p.py
```diff
@@ -0,0 +1,39 @@
...
+
+@pytest.mark.skip(reason="no way of currently testing this just in SW")
+def test_Chroma_current_limit_turn_on_and_off():
+    psu=Chroma62024P6008('USB0::0x1698::0x0837::007000000446::0::INSTR')
+    print(psu.id)
+    psu.output_current_limit = 1
+    psu.output_voltage_level = 5
+    psu.output_enable()
+    sleep(2)
+    print(psu.output_voltage)
```
bilderbuchi, line 37: In addition to printing, this test should assert that the voltage level is reached (when a device is connected, of course): 
```
assert psu.output_voltage == pytest.approx(5)

**Review by bilderbuchi — CHANGES_REQUESTED** [review]

**Comment by bmoneke** [comment]
Hey @tinix84, do you want to pick up your instruments or give some word regarding the state of your contributions (I assume they are working)?

**Comment by tinix84** [comment]
Hi sorry I'm pretty busy, can you say what you need exactly?

File: pymeasure/instruments/__init__.py [hunk 8]
```diff
@@ -34,6 +34,8 @@
 from . import ami
 from . import anapico
 from . import anritsu
+from . import chroma
+from . import deltaelektronika
```
bmoneke, line 38: That line is already included two lines below(new line 40):
```suggestion
```

bmoneke (changes requested):
I added a few comments (most of them have already suggestions, which you might commit as they are).

Then we need to add some documentation about the API of these devices, see the [documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#adding-documentation).

Finally you can merge the current master into your branch, in order to run all the tests via CI.

All together, it should not take too long (adding tests would be nice, but I understand, if you do not have the time).

PR #240, opened 2019-12-16
lecroy driver based on ActiveDSO dll

tinix84:
this is a driver for lecroy oscilloscope, in this case ActiveDSO.dll is necessary to install. Does anyone have other methods to connect to them?


bmoneke:

PR #242, opened 2019-12-25
added custom terminator string and improved serial connection

tecamenz:
The serial connection (in Windows) with our Keithley 2000 DMM did not work  so I changed the serial adapter in the following ways:
1.) Added the ability to define custom terminator string 
````
adapter = SerialAdapter("COM4", terminator="\n", baudrate=19200, timeout=5)
````
2.) replaced ``serial.readlines`` with ``serial.readline`` because after all data was read, the adapter would block till a timeout is reached

### DIFF
--- a/pymeasure/adapters/serial.py
+++ b/pymeasure/adapters/serial.py
@@ -23,7 +23,7 @@
 #
 
 import logging
-
+import time
 import serial
 import numpy as np
 
@@ -41,7 +41,8 @@ class SerialAdapter(Adapter):
     :param kwargs: Any valid key-word argument for serial.Serial
     """
 
-    def __init__(self, port, **kwargs):
+    def __init__(self, port, terminator="\n", **kwargs):
+        self.terminator = terminator
         if isinstance(port, serial.Serial):
             self.connection = port
         else:
@@ -51,13 +52,23 @@ class SerialAdapter(Adapter):
         """ Ensures the connection is closed upon deletion
         """
         self.connection.close()
+        
+    def reset(self):
+        """ Flush input and output buffer to prevent side effects for example
+        when the user interacts locally with the device while the serial connection 
+        is kept open
+        """
+        self.connection.reset_input_buffer()
+        self.connection.reset_output_buffer()
 
     def write(self, command):
         """ Writes a command to the instrument
 
         :param command: SCPI command string to be sent to the instrument
         """
+        command = command + self.terminator
         self.connection.write(command.encode())  # encode added for Python 3
+        time.sleep(0.1) # give the device time to process the commands
 
     def read(self):
         """ Reads until the buffer is empty and returns the resulting
@@ -65,7 +76,15 @@ class SerialAdapter(Adapter):
 
         :returns: String ASCII response of the instrument.
         """
-        return b"\n".join(self.connection.readlines()).decode()
+        buff = []
+        while self.connection.in_waiting > 0:
+            tmp = self.connection.readline().decode() # readlines replaces with readline
+            buff.append(tmp)
+        if len(buff) > 1:
+            return buff
+        if len(buff) == 1:
+            return buff[0]
+
 
     def binary_values(self, command, header_bytes=0, dtype=np.float32):
         """ Returns a numpy array from a query for binary data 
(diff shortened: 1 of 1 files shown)


bilderbuchi:
I would welcome the possibility to set the termination for serial connections. Even better would be to set separate read and write termination, as some few devices need that (there is an issue on the tracker about this).

I don't agree with introducing an arbitrary time limit, especially non-configurable, as this would hard-limit your query rate. Also why is this tacked to the end of the write? Should such a delay not be part of user code, or at least in the query/ask method, where a read immediately follows a write?

Also, is it intentionally that you change the API of the read method to return 3 different data types depending on device behaviour (bytestring, list or None), instead of only bytestring?

tecamenz:
The delay at the end of the write was added to not break the interface of keithley2000.py and buffer.py where for example ``config_buffer(...)``  (in pymeasure/instruments/keithley/buffer.py) would execute multiple writes to the the device. I found that without the delay, the device would sometimes return "None" on a subsequent read. A second read would then return the requested data. 
Another idea would be to handle "None" returns in read and try again?
Because the device can return multiple lines / messages, I changed the Interface to a list. 

Example-Code:
````
adapter = SerialAdapter("COM4", terminator="\n", baudrate=19200, timeout=5)
adapter.reset()
dmm = Keithley2000(adapter)
dmm.reset()
dmm.clear()
dmm.measure_voltage(1)
dmm.voltage_nplc = 0.01
dmm.reset_buffer()
dmm.config_buffer(600)
dmm.start_buffer()
dmm.wait_for_buffer(timeout=10)
x = dmm.buffer_data
````

Error:
````
ValueError                                Traceback (most recent call last)
<ipython-input-3-0b87ae62a90f> in <module>
      7 dmm.voltage_nplc = 0.01
      8 dmm.reset_buffer()
----> 9 dmm.config_buffer(600)
     10 dmm.start_buffer()
     11 dmm.wait_for_buffer(timeout=10)

~\Documents\Projects\Keithley_DMM\pymeasure\pymeasure\instruments\keithley\buffer.py in config_buffer(self, points, delay)
     64         self.trigger_delay = delay
     65         self.write(":TRAC:FEED SENSE;:TRAC:FEED:CONT NEXT;")
---> 66         self.check_errors()
     67 
     68     def is_buffer_full(self):

~\Documents\Projects\Keithley_DMM\pymeasure\pymeasure\instruments\keithley\keithley2000.py in check_errors(self)
    449         while True:
    450             err = self.values(":SYST:ERR?")
--> 451             if int(err[0]) != 0:
    452                 errmsg = "Keithley 2000: %s: %s" % (err[0],err[1])
    453                 log.error(errmsg + '\n')

ValueError: invalid literal for int() with base 10: 'None'

````

bilderbuchi:
@tecamenz have you tried this out with a VISAAdapter instead of the SerialAdapter? That should have most of the features you need already -- you can pass `read_termination`  and `write_termination` to the Instrument constructor, and as soon as #272 is resolved you can also use `query_delay` to introduce a delay between write and read as required by your hardware.

bilderbuchi:
One possible limitation is that only some VISA backends support giving the port string as `COM9`, then you have to say `ASRL9::INSTR` instead. If you have the NI or Keysight VISA libraries installed you should be fine.

Applepi:
Any chances of getting this fixed? My serial connection to GRBL just hangs and will not proceed no matter what I've tried.

bilderbuchi:
> @tecamenz have you tried this out with a VISAAdapter instead of the SerialAdapter?

Applepi:
> > @tecamenz have you tried this out with a VISAAdapter instead of the SerialAdapter?
> 
> Or @Applepi, too - using a VISAAdapter should probably help depending on your specific problem.

Yes I've tried it but the usage of ask has been deprecated by pyvisa, I've managed to get it to work with write and reads but I have to do a bunch of excessive reads to clear the buffer to get the data I require.

bilderbuchi:
ask is replaced by query, the fix is already in pymeasure master, but not released yet.

bilderbuchi:
In the meantime, we have now updated the connection configuration of VISAAdapter significantly, please try with that one, instead of SerialAdapter, if you still have this problem. 

Please see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#defining-default-connection-settings for instructions on how to implement an instrument driver with custom connection settings, and https://pymeasure.readthedocs.io/en/latest/tutorial/connecting.html#modifying-connection-settings for how to override those settings as a user.

I think this change has been superseded, so I'll close this PR.

PR #243, opened 2020-01-10
Added support for Keithley 2700 & Keithley 6621 and added functionality for Keithley 2400

CasperSchippers:
**Ready to be merged**

This PR adds support for the Keithley 2700 MultiMeter/Switch System and the Keithley 6221 DC and AC current source.
Moreover, it adds functionality to the already existing Keithley 2400 and fixes the missing doc-links for the Keithley 2450.


bilderbuchi:
Where are those changes related to the 2400? I can't find them in the pr changes.

Furthermore, there's a header in one file "This file is part of the PyMeasure package.", that seems out of place.

bilderbuchi:
> Furthermore, there's a header in one file "This file is part of the PyMeasure package.", that seems out of place.
huh, seems I was confused and thought I'm in the instrumentkit repository, sorry bout the noise.

bilderbuchi:
@CasperSchippers how are you using your 2400? I remember that I had to implement as special adapter to append a \n to every command, otherwise the commands defined in pymeasure were not correctly recognized. Did you see something like that?

Also, it seems really easy to get into a wrong operation or somesuch, without a timeout or other escape hatch bailing you out. E.g. using mykeithley.voltage without having set up voltage measurement correctly seemed to make the 2400 hang indefinitely. I am not 100% sure that that was not a weird error on my part, as it was a hectic debugging session and I didn't have time to dig down deeper.

CasperSchippers:
> @CasperSchippers how are you using your 2400? I remember that I had to implement as special adapter to append a \n to every command, otherwise the commands defined in pymeasure were not correctly recognized. Did you see something like that?
> 
> Also, it seems really easy to get into a wrong operation or somesuch, without a timeout or other escape hatch bailing you out. E.g. using mykeithley.voltage without having set up voltage measurement correctly seemed to make the 2400 hang indefinitely. I am not 100% sure that that was not a weird error on my part, as it was a hectic debugging session and I didn't have time to dig down deeper.

I've been using the Keithley 2400 for pulsed 4-probe resistance measurements.
I agree that sometimes (but for me really not that often) the system can hang when bad commands are sent.
I didn't need to send a additional \n at the end of each command, but that might depend on how the Keithley is connected (I used GPIB with a National Instruments GPIB-USB adapter); I know that when using serial (i.e. rs232) these things are way more tedious to get right, but you can play around with a number of key-word arguments (timeout, delay, send_end, read_termination, write_termination) when making an instance of the Keithley2400 class to get the connection working; I doubt that you need to make a custom adapter for this purpose.

bilderbuchi:
Thanks for the reply!

> When I use it, I didn't need to send a additional \n at the end of each command, but that might depend on how the Keithley is connected (I used GPIB with a National Instruments GPIB-USB adapter)

Thanks, that is an interesting data point! @cjermain is this the way this works? Need for setting termination depends on which kind of connection you use (gpib vs. serial)?

Yeah, the custom adapter was mainly meant so that our other users don't have to mess around with configuring termination, baud rate, whatever, and can just say `k = Keithley2400('COM4')` and everything works. A custom adapter seemed to me the easiest way to achieve that (and also the only way I found to deal with the read/write termination).

I think next time the device is free I will have to play around to get to the bottom of what is going on.

pymeasure/instruments/keithley/keithley2700.py
```diff
@@ -0,0 +1,362 @@
...
+        map_values=True,
+    )
+    display_text = Instrument.control(
+        "DISP:TEXT:DATA?", "DISP:TEXT:DATA '%s'",
+        """ A string property that controls the text shown on the display of
+        the Keithley 2700. Text can be up to 12 ASCII characters and must be
+        enabled to show.
+        """,
+        validator=text_length_validator,
+        values=12,
```
cjermain: This is an interesting use of the validator pattern. As more cases like this occur, it will be worth revisiting the syntax to see if there are better ways to generalize. In the meantime, I think this is still clear.

pymeasure/instruments/oxfordinstruments/itc503.py
```diff
@@ -119,7 +119,7 @@ class ITC503(Instrument):
         get_process=lambda v: float(v[1:]),
     )
 
-    temperature_2 = Instrument.measurement(
+    temperature_3 = Instrument.measurement(
```
cjermain: Good catch!

PR #244, opened 2020-01-17
Adding National Instruments Virtual Bench Instrument

moritzj29:
Hi,
this is an Instrument driver for the National Instruments Virtual Bench device. This instrument driver just wraps the methods of the armstrap-pyvirtualbench Python driver interface (https://github.com/armstrap/armstrap-pyvirtualbench) to a pymeasure compatible format.
- Mixed-Signal-Oscilloscope (MSO)
- Digital Input Output (DIO)
- Function Generator (FGEN)
- Power Supply (PS)
are implemented and should work. Not every method is tested to the bones yet, but I've been running the driver for a while.
The i2c and spi functionalities are not implemented so far and I don't plan to do so in the near future, since they are quite special and I don't need them. (Maybe someone else gets inspired? ;))
Also the calibration methods are not implemented, since they won't be required regularly.

I'm happy to incorporate your feedback!

### DIFF
--- a/docs/api/instruments/ni/index.rst
+++ b/docs/api/instruments/ni/index.rst
@@ -1,0 +1,12 @@
+.. module:: pymeasure.instruments.ni
+
+####################
+National Instruments
+####################
+
+This section contains specific documentation on the National Instruments instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments>`.
+
+.. toctree::
+   :maxdepth: 2
+
+   virtualbench
--- a/docs/api/instruments/ni/virtualbench.rst
+++ b/docs/api/instruments/ni/virtualbench.rst
@@ -1,23 +1,23 @@
-##################################
+################
 NI Virtual Bench
-##################################
+################
 
-**********************************************
+*******************
 General Information
-**********************************************
+*******************
 
 The `armstrap/pyvirtualbench <https://github.com/armstrap/armstrap-pyvirtualbench>`_ 
 Python wrapper for the VirtualBench C-API is required.
 This Instrument driver only interfaces the pyvirtualbench Python wrapper.
 
-**********************************************
+********
 Examples
-**********************************************
-to be documented. Check the examples in the pyvirtualbench repository to get an idea.
+********
+To be documented. Check the examples in the pyvirtualbench repository to get an idea.
 
-**********************************************
+****************
 Instrument Class
-**********************************************
+****************
 
 .. autoclass:: pymeasure.instruments.ni.virtualbench.VirtualBench
     :members:
--- a/docs/api/instruments/ni/virtualbench.rst
+++ b/docs/api/instruments/ni/virtualbench.rst
@@ -1,0 +1,15 @@
+##################################
+National Instruments Virtual Bench
+##################################
+
+.. autoclass:: pymeasure.instruments.ni.VirtualBench
+    :members:
+    :show-inheritance:
+    :inherited-members:
+    :exclude-members:
+
+.. autoclass:: pymeasure.instruments.ni.VirtualBench_Direct
+    :members:
+    :show-inheritance:
+    :inherited-members:
+    :exclude-members:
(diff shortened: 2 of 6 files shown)


pymeasure/instruments/ni/virtualbench.py
```diff
@@ -0,0 +1,1713 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2019 PyMeasure Developers
+# pyvirtualbench library: Copyright (c) 2015 Charles Armstrap <charles@armstrap.org>
```
moritzj29: I included the copyright of the underlying library, since I copied some docstrings

pymeasure/instruments/ni/virtualbench.py
```diff
@@ -0,0 +1,1713 @@
...
+import logging
+import re
+# ctypes only required for VirtualBench_Direct class
+from ctypes import (c_bool, c_size_t, c_double, c_uint8, c_int32, c_uint32,
+                    c_int64, c_uint64, c_wchar, c_wchar_p, Structure, c_int,
+                    cdll, byref)
+from datetime import datetime, timezone, timedelta
+import numpy as np
+import pandas as pd
+import pyvirtualbench as pyvb
```
cjermain: This will need to be included as an optional dependency. It's worth doing a try-except around this import and giving a sensible warning using the log.

pymeasure/instruments/ni/__init__.py
```diff
@@ -22,4 +22,7 @@
 # THE SOFTWARE.
 #
 
-from .daqmx import DAQmx
```
cjermain: I see that they both have different optional dependencies. If we wrap the imports in try-exceptcatch, we can keep them both in here, and just raise a warning.

docs/api/instruments/ni/virtualbench.rst
```diff
@@ -0,0 +1,32 @@
+##################################
```
cjermain: Line lengths for the title markers (i.e. `#` and `*`) should match the length of the content.

pymeasure/instruments/ni/virtualbench.py
cjermain: Is the underlying library licensed with the MIT License?

docs/api/instruments/ni/index.rst
```diff
@@ -0,0 +1,12 @@
...
+####################
+National Instruments
+####################
+
+This section contains specific documentation on the National Instruments instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments>`.
+
+.. toctree::
+   :maxdepth: 2
+
+   virtualbench
```
cjermain: We should also include the DAQmx docs here...but this can be considered out of scope for this PR.

pymeasure/instruments/ni/virtualbench.py
moritzj29: yes, it's also MIT License: https://github.com/armstrap/armstrap-pyvirtualbench/blob/master/LICENSE

docs/api/instruments/ni/index.rst
moritzj29: the DAQmx code does merely contain docstrings so far, so I don't add it to the index file now. As I don't have such an instrument I can't help on the documentation. A separate PR is good idea!

docs/api/instruments/ni/virtualbench.rst
moritzj29: done

pymeasure/instruments/ni/virtualbench.py
moritzj29: I implemented a try-except statement which emits a warning to the logger. So far I only catched the ModuleNotFoundError. I'm unsure whether it makes sense to catch more generously...

pymeasure/instruments/ni/__init__.py
moritzj29: I also added try-except for the DLL import in DAQmx. In this case I only catch the OSError so far...
On a system where neither the DAQmx nor VirtualBench drivers are installed, importing the NI instruments throws no errors now (only log entries).

cjermain:
Thanks @moritzj29! I think this is a great addition, and helps to improve some of the documentation around the NI features we support.

Issue #235, opened 2019-11-09
Issue with file access in the "using the plotter " example


badarirao:
Hi,
I have just begun to try the graphical interface part of the package. I see that when I run the [1st example](https://pymeasure.readthedocs.io/en/latest/tutorial/graphical.html) in the "using the plotter" tutorial, the program runs fine in spyder (windows 10 machine). 
However, after the program has finished running, when I try to open the "random.csv" file, I get the following notice.
![scrnshot](https://user-images.githubusercontent.com/47620203/68530560-7a271180-034c-11ea-9c5c-4c01478bf473.PNG)

I have just started with this package, and not sure where exactly to look for the source of the problem. It feels like the file has not been properly closed by the code. I even tried to reduce the timeout in worker.join to 60 seconds, but the problem persists.
The only workaround I have found is that I have to restart the kernel in sypder, or close it, only then the file can be accessed properly.
Is this a bug or am I supposed to do something?

Also, this is the first example I have checked. Similar problem could be occurring in other examples as well!

cjermain:
@badarirao, thanks for raising this issue. When you are running the example are you doing so in an interactive terminal within Spyder?

For background, a `Recorder` object keeps a handle on the file as the process is running -- this is spawned by the `Worker` object. It runs in the background and appends data to the relevant files for the `Results` object. It's likely that this thread is still going in the background if you are using a interactive interpreter.
https://pymeasure.readthedocs.io/en/latest/tutorial/procedure.html#running-procedures

badarirao:
@cjermain, Yes I am running the example within Spyder.
Is there a way to stop the thread in Spyder without having to restart the kernel?

cjermain:
@badarirao, I've made a change in #245 that I suspect will fix the issue. Can you please test this change to see if it resolves the problem?

badarirao:
@cjermain, Hi, I tried the example after the change, however the problem persists.
After running the program, worker.is_alive() returns False, but opening the file still gives the same message.
I also tried invoking worker.recorder.stop() separately, however, it gave an error: "AttributeError: 'NoneType' object has no attribute 'join'", pointing to error in logging\handlers.py, line 1503, in stop self._thread.join()

Also, I verified that running the program from command prompt does not give any problem. However, python gave a warning: "QObject::~QObject: Timers cannot be stopped from another thread" Does this have anything to do with the above problem?

cjermain:
[...]

PR #245, opened 2020-01-18
Ensuring that the Recorder thread is joined properly on Worker.shutdown

cjermain:
This PR ensures that the `Recorder` thread is joined properly during `Worker.shutdown`. In #235 on Windows, this appears to cause the file handle to be left open.

Previously, we used the `enqueue_sentinel` method to tell the `Recorder` to stop. This is called within the `stop` method, with the addition of `join` -- so we have a stronger guarantee that the thread is finished.


cjermain:
This does not appear to solve the issue in #235, but I think it is still worth merging.

PR #247, opened 2020-02-14
Keithley6517B added

wehlgrundspitze:
Instrument Keithley6517B (Electrometer/High Resistance Meter) added. At the moment the class provides only the functionalities for single measurements. Procedures concerning filling and reading the buffer are not implemented so far.


pymeasure/instruments/keithley/keithley6517b.py
```diff
@@ -0,0 +1,343 @@
...
+import time
+import re
+
+import numpy as np
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import truncated_range, strict_discrete_set
+from .buffer import KeithleyBuffer
+
+# Setup logging
```
cjermain: I think this comment can be removed.

pymeasure/instruments/keithley/keithley6517b.py
```diff
@@ -0,0 +1,343 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2019 PyMeasure Developers
```
cjermain: Please add yourself to the authors.
https://github.com/ralph-group/pymeasure/blob/master/AUTHORS.txt
https://github.com/ralph-group/pymeasure/blob/master/docs/about/authors.rst

pymeasure/instruments/keithley/keithley6517b.py
```diff
@@ -0,0 +1,343 @@
+#
```
cjermain: Please add docs for this instrument by following the instructions in the Contributing guide:
https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html

pymeasure/instruments/keithley/keithley6517b.py
```diff
@@ -0,0 +1,343 @@
...
+
+    """
+
+    source_enabled = Instrument.measurement("OUTPUT?",
+        """ Reads a boolean value that is True if the source is enabled. """,
+        cast=bool
+    )
+
+
+    def extract_value(result):
```
cjermain: This function seems that it could be generalized out of this class. Are there other places it can be reused? I would suggest also providing more details in the docs.

wehlgrundspitze:
Concerning the generalization I totally agree, that some functions e.g. "extract_value" can be taken out of this class. Unfortunately, I'm not so familiar with this project so far, to know where to put it instead. Please let me know, if there is already an hierarchy similar to my quickly drawn UML diagram (this case this exists, the functionality would be an issue for the KeithleyInstrument or the - not drawn - KeithleyDCElectricMeter).
pymeasure-interface-hierarchy

pymeasure/instruments/keithley/keithley6517b.py
```diff
@@ -0,0 +1,343 @@
...
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import truncated_range
+from .buffer import KeithleyBuffer
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class Keithley6517B(Instrument, KeithleyBuffer):
```
bilderbuchi, line 39: Note that you still need to import this in the `pymeasure/instruments/keithley/__init__.py` file.

pymeasure/instruments/keithley/keithley6517b.py
bilderbuchi: docs have been added.

pymeasure/instruments/keithley/keithley6517b.py
```diff
@@ -0,0 +1,343 @@
...
+        """ A floating point property that controls the source voltage
+        range in Volts, which can take values from -1000 to 1000 V.
+        Auto-range is disabled when this property is set. """,
+        validator=truncated_range,
+        values=[-1000, 1000]
+    )
+
+    source_voltage_limit = Instrument.control(
+        ":SOUR:CURR:RLIM?", ":SOUR:CURR:RLIM %g",
+        """ A floating point property that controls the compliance resistance
```
bilderbuchi, line 149: That docstring does not seem to match the property name?

pymeasure/instruments/keithley/keithley6517b.py
```diff
@@ -0,0 +1,343 @@
...
+        else:
+            self.current_range = current
+        self.check_errors()
+
+    def auto_range_source(self):
+        """ Configures the source to use an automatic range.
+        """
+        self.write(":SOUR:VOLT:RANG:AUTO 1")
+
+    def apply_voltage(self, voltage_range=None,
```
bilderbuchi, line 268: Is this methods incompletely migrated from another instrument? Both `self.source_mode` and `self.compliance_current` are not defined in the present implementation, and are only defined in here.

pymeasure/instruments/keithley/keithley6517b.py
```diff
@@ -0,0 +1,343 @@
...
+        message = err[1].replace('"', '')
+        return (code, message)
+
+    def check_errors(self):
+        """ Logs any system errors reported by the instrument.
+        """
+        code, message = self.error
+        while code != 0:
+            t = time.time()
+            log.info("Keithley 2450 reported error: %d, %s", code, message)
```
bilderbuchi, line 305: Overlooked old instrument name. 
If we factor that out, we could use `self.__class__.__name__` or somesuch to make this dynamically correct.

pymeasure/instruments/keithley/keithley6517b.py
bilderbuchi: has been done

pymeasure/instruments/keithley/keithley6517b.py
bilderbuchi: done

bilderbuchi:
There is no existing inheritance hierarchy for the Keithleys (with the exception of the KeithleyBuffer). From reading this PR I see a lot of similarity with the Keithley2400 sourcemeter and the Keithley2450 - I have a feeling some refactoring would be possible.
Maybe it would be best to merge this when it is complete as-is, and work on the refactoring in a later PR? This way the instrument driver gets in as soon as possible, and the refactoring effort can'T delay that.

One thing that I don't know right now how to implement is using Pymeasure properties (`control`, `measurement`, `setting`) with validators that are different by subclass. E.g. 3 different models have the same API/command set, but different voltage/current limits. @cjermain do you have a best practice here?
Further on, how can you handle limits that can be adjusted dynamically with validators (as the limits get baked in during class definition).

wehlgrundspitze:
Thanks @bilderbuchi for your detailed analysis of my code. I have revised all mentioned segments.

Please let me know if there is anything new concerning the inheritance hierarchy. My main motivation for this is the exchangeability of instruments. In my special case, I wanted to use an Agilent frequency generator to replace the previously used internal voltage source of an SR830 lockin amplifier - both would implement "ACSource".

pymeasure/instruments/keithley/keithley6517b.py
bilderbuchi: I think we have achieved what we can here, now.

bilderbuchi:
I recognize your motivation. I think in practice it will be hard to pull this off consistently. I think [IVI](https://www.ivifoundation.org/about/instrument_classes.aspx) aims at basically this problem, but for example python-ivi also [does not really implement](https://github.com/python-ivi/python-ivi/issues/50) those ideas.

bilderbuchi:
I think we should merge this as-is, and in follow-up work see what we can do to reduce duplication. There are some merge conflicts, can you try to resolve that, please?

bilderbuchi:
I fixed the merge conflict. Thank you for your contribution!

PR #252, opened 2020-03-01
ReadtheDocs update

cjermain:
This PR updates the ReadtheDocs configuration to version 2.


cjermain:
I'm using `autodoc_mock_imports` to avoid the necessity to provide optional requirements. This helps to avoid a new memory limit that has been causing the docs-builds to fail.
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports

PR #254, opened 2020-03-03
Implemented some abstractions and Rigol HAL

KOLANICH:
Improved VISA adapter: now it parses headers.

@vik-s, @dvspirito, @minhhaiphys, @watersjason, @jmittelstaedt, @dennisfeng2, @nowacklab-user, @ronan-sensome, @troylf, @dvase

### DIFF
--- a/pymeasure/instruments/instrument.py
+++ b/pymeasure/instruments/instrument.py
@@ -70,6 +70,9 @@ class Instrument(object):
                                            """ Returns the status of the instrument """)
             self.complete = self.measurement("*OPC?",
                                              """ TODO: Add this doc """)
+        if self.name is None:
+            id_dict = self.id_dict
+            self.name = " ".join(id_dict[k] for k in ("vendor", "model", "serial"))
 
         self.isShutdown = False
         log.info("Initializing %s." % self.name)
@@ -81,15 +84,21 @@ class Instrument(object):
             return self.adapter.ask("*IDN?").strip()
         else:
             return "Warning: Property not implemented."
+    
+    @property
+    def id_dict(self):
+        return dict(zip(("vendor", "model", "serial", "firmware"), self.id.split(",")))
 
     # Wrapper functions for the Adapter object
-    def ask(self, command):
+    def query(self, command):
         """ Writes the command to the instrument through the adapter
         and returns the read response.
 
         :param command: command string to be sent to the instrument
         """
         return self.adapter.ask(command)
+    
+    ask = query # ask is deprecated
 
     def write(self, command):
         """ Writes the command to the instrument through the adapter.
--- a/pymeasure/instruments/rigol/__init__.py
+++ b/pymeasure/instruments/rigol/__init__.py
@@ -1,0 +1,31 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2019 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from ..oscilloscope import *
+from .enums import *
+
+from .products.generic import RigolHAL
+from .products import products
+
+RigolHALDispatcher = ModelDictHALDispatcher(products, RigolHAL)
--- a/pymeasure/instruments/oscilloscope.py
+++ b/pymeasure/instruments/oscilloscope.py
@@ -1,0 +1,309 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2019 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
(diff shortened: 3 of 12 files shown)


KOLANICH:
@scottprahl, DS1204B

bilderbuchi:
Hi!
Thank for your contributing this! This looks like a rather large PR, and as such it will probably take long to review. Do you think it feasible to separate the work, e.g. into the generic changes and the RIGOL-specific parts, or are you adapting both as you go?

pymeasure/instruments/instrument.py
```diff
@@ -81,15 +84,21 @@ def id(self):
...
-    def ask(self, command):
+    def query(self, command):
         """ Writes the command to the instrument through the adapter
         and returns the read response.
 
         :param command: command string to be sent to the instrument
         """
         return self.adapter.ask(command)
+    
+    ask = query # ask is deprecated
```
bilderbuchi: This is deprecated by whom? can you clarify this? also, note that in the line above you still do adapter.ask...

KOLANICH: by pyvisa, if I remember right.

KOLANICH:
>Do you think it feasible to separate the work, e.g. into the generic changes and the RIGOL-specific parts, or are you adapting both as you go?

Not really, it is pretty unfinished, I am completely unsure that I won't change the architecture drastically. That's why I have posted it as a draft - to get feedback from the community and the most importantly from  maintainers of other instruments.

bilderbuchi:
I see, thanks for the context! 
As you expect architecture changes, I think it would be great to coordinate/discuss with @cjermain to make sure your changes will be compatible with the project direction, and to avoid spending effort on work that can't be merged later.

pymeasure/instruments/instrument.py
bilderbuchi: I checked, this was deprecated in [pyvisa 1.9](https://github.com/pyvisa/pyvisa/blob/0147607bccfb1c7de9d44d36d66ec543d93e8b3e/CHANGES#L93) and will be [removed in 1.10](https://github.com/pyvisa/pyvisa/blob/47cc2b1ae3318b7b437cc7a49be8778c713377b8/pyvisa/resources/messagebased.py#L615) (although at 1.10.1, I still have the warning), so thank you for spotting this!
However, this has already been fixed in the [visa adapter](https://github.com/ralph-group/pymeasure/blob/master/pymeasure/adapters/visa.py#L112), so there is no need to change our own public API.
(As an aside, if you end up deprecating something, it's better to notify users about this, e.g. like [this](https://github.com/pyvisa/pyvisa/blob/47cc2b1ae3318b7b437cc7a49be8778c713377b8/pyvisa/resources/messagebased.py#L615-L618))

cjermain: The purpose of the `Adapter` classes is to abstract away the details of the underlying communication protocols/libraries. We originally based the API on an earlier version of pyvisa, but we are not tied the pyvisa API. The `Instrument` API is stable, and is decoupled from the specific adapters. As @bilderbuchi points out, changes were made specifically to the VISAAdapter to handle changes based on the version of pyvisa.

pymeasure/instruments/instrument.py
```diff
@@ -81,15 +84,21 @@ def id(self):
             return self.adapter.ask("*IDN?").strip()
         else:
             return "Warning: Property not implemented."
+    
+    @property
+    def id_dict(self):
+        return dict(zip(("vendor", "model", "serial", "firmware"), self.id.split(",")))
```
cjermain: This assumes a specific format from the `id`, which does not generalize for different adapters. Those cases need to be handled in this class, or it should be removed.

pymeasure/adapters/visa.py
```diff
@@ -121,18 +121,20 @@ def ask_values(self, command):
         """
         return self.connection.query_values(command)
 
-    def binary_values(self, command, header_bytes=0, dtype=np.float32):
+    def binary_values(self, command, header_bytes=None, dtype=np.float32):
```
cjermain: I would suggest breaking this change out into a separate PR. It will make it much easier to get merged.

pymeasure/adapters/visa.py
```diff
@@ -121,18 +121,20 @@ def ask_values(self, command):
...
         """ Returns a numpy array from a query for binary data
-
         :param command: SCPI command to be sent to the instrument
         :param header_bytes: Integer number of bytes to ignore in header
```
cjermain: The default case needs to be updated. There are now multiple types being allowed for the same input, which I think will be confusing. I suggest adding a different argument here.

pymeasure/instruments/rigol/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from ..oscilloscope import *
```
cjermain: Please remove all `*` imports. This leads to polluted namespaces and makes it very hard to trace imports.

pymeasure/instruments/instrument.py
KOLANICH: Yes, but when porting code from using pyvisa to using pymeasure it'd be nice to have methods (that are in fact aliases) mimicking current pyvisa API. Since it doesn't harm, I have added an alias.

pymeasure/instruments/instrument.py
KOLANICH: Isn't the id string standardized?

pymeasure/adapters/visa.py
KOLANICH: A different arg IMHO will be more confising. The docs is surely needs to be updated. BTW, how about dropping python 2 and applying type annotations everywhere?

pymeasure/instruments/rigol/__init__.py
KOLANICH: IMHO it should be done in the end, when we decide what our interfaces finally are (we likely have a lot of stuff to do to unify all the devices). For now it is too early to do this IMHO.

KOLANICH:
> You raise an interesting question in this PR: should we have abstract classes for instrument types (e.g. oscilliscope, multimeter, etc.)? Each instrument is very specialized in what it provides.

Each instrument is very specialized in the API and in the additional features it uses. But they cannot be different in general. For example each ADC and DAC device (so, almost every measurement device satisfies this) must have channels. Every oscilloscope must be able to acquire waveforms (signals pairs of time and value channels). Every signal generator must be able to generate waveforms. So we can really put most of devices behind unified interfaces.


>I've found it easier to build specific files for each instrument and abstract code within vendors (e.g. [KeithleyBuffer](https://github.com/ralph-group/pymeasure/blob/master/pymeasure/instruments/keithley/buffer.py) as a mixin class).

This ease is just work that hasn't be done and shifted to the devs of the software using this lib. Layering is a good idea (and I have a half-finished lib wrapping one the lowest layers - SCPI commands layer, whic can be merged into HALs introduced here in some future), but `pymeasure` pretends to be a high-level lib, and to be a high-level lib one needs abstract interfaces allowing to plug almost any device to my software without changing the code of my software, only configs.

>In the short term I expect being instrument specific is the most effective strategy. I would certainly encourage a discussion in the Issues to identify commonalities across instruments that could be better abstracted.

Imho it makes more sense to discuss in this PR itself. It has some interfaces supporting different paradigms different devices can use. I want the instruments maintainers evaluate how the interfaces fit their devices, and if they don't, suggest what is missing.

pymeasure/adapters/visa.py
KOLANICH: #259

pymeasure/adapters/visa.py
KOLANICH: #259

bmoneke:

PR #255, opened 2020-03-03
Moved the metadata into setup.cfg

KOLANICH:


bilderbuchi:
This PR breaks all the tests. 
@cjermain what's your opinion about moving package metadata into setup.cfg?

cjermain:
@KOLANICH, thanks for raising this. I am not convinced there is a clear advantage to using `setup.cfg` or `pyproject.yml` at this point given the considerable uncertainly about these formats (pypa/setuptools#1160 and pypa/setuptools#1688 -- I see you've been a participant in these conversations). I don't expect `setup.py` to go away anytime soon, and I think that it serves the current needs of the project well. Happy to reconsider if there are specific use cases that are relevant to PyMeasure.

PR #256, opened 2020-03-03
Added .editorconfig

KOLANICH:


bilderbuchi:
I don't understand to be honest - these look like default editor settings (at least I can tell for python files) - why does this need to be handled per-repo?.

KOLANICH:
>these look like default editor settings

These are not default editor settings for me.

I guess if there was such a thing as "default editor settings fitting everybody's needs" there were no such things as `.editorconfig` spec and modules for editors implementing it.

cjermain:
This is an interesting addition that could help developers to be on the same page. However, I would prefer to avoid adding editor configuration files (even if they are broadly accessible) unless there is a clear consensus among PyMeasure developers that this is valuable. There are many other tools that we could choose (e.g. [psf/black](https://github.com/psf/black)) to handle format consistency. I think it would be worth raising as an Issue first to gather feedback. We can re-open and merge if there is broad support.

KOLANICH:
>There are many other tools that we could choose (e.g. psf/black) to handle format consistency.

IMHO `black` is not good. Because
0. It doesn't support auto-wrap in editors. To workaround it I set `-l` to arbitrary large-enough value (i.e. `-l  100500`).
1. It doesn't handle multiline stuff properly, i.e. especially `dict`s, in presence of `-l  100500`
2. It doesn't handle space-indent formatted stuff properly in presence of `-l  100500`
3. It doesn't handle the similar stuff that is aligned using spaces to fit into invisible table columns
4. It doesn't handle commented-out code properly (it shouldn't be considered as an ordinary comment and shouldn't be formatted and transformed as such)
5. It has the use of 4 spaces hardcoded instead of making the indentation configurable. I workaround it by changing a single string constant (I use tabs in my projects) in `black`.
5. it is `an uncompromising Python code formatter`, meaning that the issues mentioned are considered by its devs not as issues but as already implemented features.

So I vote against `black`. It spoils the code that sometimes I (and some other authors) have to carefully format by hand (IMHO it can be fully automated, but it's just noone has written a tool for that) in order to keep it understandable in violation of some python community sacred cows like PEP 8. So I usually have to do a manual pass after blackening (and without an `.editorconfig` file configuring usage of spaces for the source file in the repo my editor uses tabs).

PR #259, opened 2020-03-10
Improved VISA adapter: now it parses headers.

KOLANICH:


pymeasure/adapters/visa.py
```diff
@@ -21,7 +21,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 #
-
+import typing
```
cjermain: There are no plans to support static typing in PyMeasure. This adds a lot of additional code to manage, without a clear advantage for this project. Type documentation should be obvious from the variable name or clearly stated in the docstring.

pymeasure/adapters/visa.py
```diff
@@ -121,18 +121,20 @@ def ask_values(self, command):
         """
         return self.connection.query_values(command)
 
-    def binary_values(self, command, header_bytes=0, dtype=np.float32):
+    def binary_values(self, command: str, header_bytes: typing.Union[str, int], dtype=np.float32):
```
cjermain: I still think that multiple arguments are clearer to support these two behaviors. We need to support the existing behavior by default.
```
    def binary_values(self, command, header_bytes=0, header_fmt=None, dtype=np.float32):
        """ Returns a numpy array from a query for binary data. By default this is implemented using the PyVISA
         `read_raw` method. If `header_fmt` is provided, the `query_binary_values` method is used instead.
 
        ...
        :param header_bytes: Integer number of bytes to ignore in header (ignored if header_format is set)
        :param header_fmt: String format of the header to parse, based on PyVISA definition (default: ignored)
        ...
        """
        if header_fmt is not None:
            return self.connection.query_binary_values(
                command,
                datatype=dtype.char,
                container=np.array,
                header_fmt=header_fmt
            )
        else:
            ...
```

pymeasure/adapters/visa.py
KOLANICH: >or clearly stated in the docstring.

docstrings mustn't contain types. It's much more better to store them as annotations.

>There are no plans to support static typing in PyMeasure.

There are no static typing in python. Anything can be stored in annotations, it's just a convention to store types there.

pymeasure/adapters/visa.py
KOLANICH: IMHO having 2 distinct mutually exclusive parameters for the same thing is semantically incorrect.

>to support the existing behavior by default.

I can set the default value to 0.

pymeasure/adapters/visa.py
bilderbuchi: "incorrect" is a pretty strong word for what is basically a preference, don't you think?

cjermain: I should have been clear to say "type hinting" instead of "static typing", but the point is the same.

@KOLANICH, your comments on this project so far have taken a very condescending tone. If you have technical issues, you are welcome to raise them in a respectful way -- but you should recognize that your assessment may not align with the project goals. Please don't use "IMHO" as a substitute for being humble and considerate.

KOLANICH: Are there any real reasons not to use type hints instead of annotations in comments given that we don't support python 2? Type hints seem to be strictly better than annotations in comments.

bmoneke:
We reworked the instruments and adapters, therefore this PR is outdated (and quite some time stale) and I'll close this PR.
If you @KOLANICH want to submit changes according to the new structure, you can reopen this PR or open a new one.

PR #274, opened 2020-05-15
Adding Agilent/Keysight B1500 Instrument

moritzj29:
Hello,
some time ago I started implementing the Agilent/Keysight B1500 instrument. The instrument is similar to the Agilent 4155/4156 in the way, that it consists of individual SMUs (or even other devices). It is a mainframe which can be equipped with different measurement modules. Since we only have SMU modules in our lab, I only implented those. So far it is possible to source/measure voltage/current, sample over time and perform staircase sweep measurements.

For every SMU a SMU class instance is created within the B1500 instrument class, numbered smu1, smu2 etc., with own properties and settings. I tried documenting all the features as good as possible, but in the end it is a complex instrument 😄 The Programming Guide available as a download from Keysight is mandatory to understand the operation. I included some basic examples in the doc file.

I've been running the code for some weeks now and it works. But I can't guarantee that it is 100% bug free, since I've not tested every method with all possible settings.

Anyhow, I would be happy if this code is helpful for someone else and if it gets merged!
I'm happy for all comments/suggestions on this code!
Moritz

### DIFF
--- a/docs/api/instruments/agilent/agilentB1500.rst
+++ b/docs/api/instruments/agilent/agilentB1500.rst
@@ -203,9 +203,20 @@ Classes to communicate with the instrument:
 * :class:`AgilentB1500`: Main instrument class
 * :class:`SMU`: Instantiated by main instrument class for every SMU
 
-.. automodule:: pymeasure.instruments.agilent.agilentB1500
-    :members: AgilentB1500, SMU
+All `query` commands return a human readable dict of settings. These are intended for debugging/logging/file headers, not for passing to the accompanying setting commands.
+
+.. autoclass:: AgilentB1500
+    :members:
     :show-inheritance:
+    :member-order: bysource
+.. autoclass:: SMU
+    :members:
+    :show-inheritance:
+    :member-order: bysource
+
+.. .. automodule:: pymeasure.instruments.agilent.agilentB1500
+..     :members: AgilentB1500, SMU
+..     :show-inheritance:
 
 **********************************************
 Supporting Classes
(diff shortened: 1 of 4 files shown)


bilderbuchi:
Thank you for this contribution! Obviously this has been a lot of work. I have read/skimmed through the code, I think it will be impossible to review such a large diff in more depth I think, especially without having the device. It's good that you have checked the implementation by using it with your device. :D How much of the implementation do think you covered in your use?

The code looks well-structured to me. For such a complex device I don't think there's much that could be reused elsewhere (and thus should be factored out), what do you think?

@cjermain should we merge this as-is, as I don't see how we can do an in-depth review? It would be fine with me; docs links are there, Moritz is already in the authors file.

moritzj29:
Thank you very much for taking time to read through this code! I think I use about 80-90% of the implemented methods, of course not with all possible combinations of settings. I only implemented methods I use/plan to use, otherwise I would still be coding...

By factoring out you mean keeping this code in a separate repo? This is also an option, of course.

bilderbuchi:
I know the feeling about implementing only what you need :D
With factoring out I meant [changing the structure of the code](https://en.wikipedia.org/wiki/Code_refactoring), in the same repo, to improve quality. For example you would want to re-use common functionality, e.g. a class method that was basically copy-pasted from one instrument to the other. Not a must right now, and sometimes challenging to pull off.

moritzj29:
I see... at the moment I don't see much code that might easily be reusable by other instruments. The most simular instrument is probably the Agilent 4156 Semiconductor Analyzer. I mimiced the substructure of the SMUs (.smu1,...) from that instrument, but the command structure is completely different. Also the B1500 is much more modular, which I tried to account for in the code. (Dynamic set up of SMUs instead of hard coded, different SMU/CMU/... types, etc.)

The only thing which might be reusable, if desired, is the formatting of instrument parameters into dicts (`to_dict`). But this is only a few lines of code. From a more general point of view, I like the idea of reading a lot of settings from the instrument and writing them into the header of the measurement file. That's what I use all the query methods for. It's a way to verify that the instrument was set correctly. Even after a while when analyzing the data. But this is not (yet) common practice with pymeasure, I guess. So I don't see a need yet to factor this out.

bilderbuchi:
agreed on all parts. 
This is fine to merge for me; considering the size I'd like to get some quick GO/NOGO feedback from @cjermain before the merge.

bilderbuchi:
I'll merge this now. Thank for for this huge effort!

PR #275, opened 2020-05-18
Fixes for NI VirtualBench + Fix Documentation

moritzj29:
Hi everyone,

The autodoc feature does not import the VirtualBench documentation, since the pyvirtualbench dependency is not available. I didn't think of that in the original PR. I added the package to the mock_imports to enable compilation. 

Also I added a short example of the DIO module to the docs.


moritzj29:
I also fixed an error in the Mixed Signal Oscilloscope class (channel validation).

pymeasure/instruments/ni/virtualbench.py
```diff
@@ -1274,15 +1250,15 @@ def configure_analog_pulse_width_trigger(self, trigger_source,
...
-        def configure_digital_pulse_width_trigger(self, trigger_source,
-                                                  trigger_polarity,
-                                                  comparison_mode,
-                                                  lower_limit, upper_limit,
-                                                  trigger_instance):
-            ''' Configures a trigger to activate on the specified source when
-            the digital edge reaches the specified levels within a specified
-            window of time.
-            '''
+        # def configure_digital_pulse_width_trigger(self, trigger_source,
```
bilderbuchi, line 1253: Please delete this instead of only commenting out.

moritzj29: you are right. removed the methods which were commented-out. added some information on the missing methods in the classes docstring instead. hope this is fine.

PR #276, opened 2020-05-22
Fix ListParameter for numbers

moritzj29:
I stumbled upon an issue where it is not possible to load a measurement file containing a numeric `ListParameter`. For Example:
`ch1_vertical_range = ListParameter(
        'CH1 Vertical Range', units='V', default=20,
        choices=[20, 10, 5, 2, 1, 0.5, 0.2, 0.1, 0.05]
    )`
The parameter is stored in the CSV file correctly, but upon reading it cannot be set because it is read as string. But `choices` does not contain that string.

In this PR I propose to check the types of all values that are passed as `choices`. They have to be of the same type or all numeric. The type is then stored as a class variable and upon setting a value, the value is converted into this type. If the choices are a mix of int and float, all set values are converted to float.

Are there other cases where different types of values are supposed to be within one list? I couldn't think of any. Even my usecase with numeric values is probably quite rare, since `FloatParameter` or `IntegerParameter` are often more appropriate.

I'm happy for your thoughts on this!


moritzj29:
ok, as I see now from the failed build checks, there is a test for the ListInput which assumes a multi-type list input: `choices = [123, 'abc', 0]`
So this appears to be desired... but breaks the import of measurement data. I think it doesn't make sense to allow multi-type lists, since it is too complicated to convert to string and back. One could write the type of the list item to the file as well and restore on read?

moritzj29:
I mark this as ready for review now to get some feedback from the maintainers regarding the questions on top. Otherwise I don't know what to change

bilderbuchi:
what you proposed sounds reasonable to me, but I don't have good confidence yet about that part of pymeasure. 
I think restricting a listparameter to have only one type of entries seems like a resonable restriction, currently I can't come up with a scenario where I would use several types in choices; 
It might just be that the test suite was not coded for that intentionally, maybe @Laogeodritt who apparently wrote [test_setValue_should_update_parameter](https://github.com/ralph-group/pymeasure/blame/07cf309b8d113931525c8cc216cc704f2b7ff413/tests/display/test_inputs.py#L105) knows the intent behind that.

Laogeodritt:
It's been a long while, so I don't recall my reasoning for that test all too well. I suspect I wrote that test purely to test different types in a single test, not necessarily to test multiple types in a single ListParameter.

Restricting the choices of the ListParameter to a list of a single type seems non-pythonic to me. In general strict type constraints isn't very pythonic, and this could break the ability to duck-type: e.g. if measurement code expects a certain type or object API for the ListParameter value, one can usually provide a duck-typed equivalent in the list. I could see this affecting a context where, e.g., you want to use a Strategy design pattern to allow selecting algorithm, or you want to pass a complex set of parameters in an object based on the choice. In the case of this specific PR, I would also add that this wouldn't be able to handle class inheritance, so it would break the traditional OOP way of customising behaviour from a base type too. Although this wouldn't be the most common way of using ListParameters, I would favour keeping the flexibility.

I might propose this counter-solution: Enforce uniqueness of the str or repr of the choices in a given ListParameter. Displaying the ListParameter dropdown and writing to the CSV should use the str or repr of each choice object explicitly, if not a string. Reading from the CSV should compare the stored data to the str or repr of the choices.

moritzj29:
Thanks @Laogeodritt for your extensive reply! I see that there may be use cases where limitation to a single type may break functionality. Even though these may be very rare, the code should be able to deal with that.

> Reading from the CSV should compare the stored data to the str or repr of the choices.

I think this is a good way to solve the problem loading a ListParameter from file. Thanks for this suggestion! I will continue working on the code...

moritzj29:
Finally found some time to continue this... I went with the suggestion of @Laogeodritt and just compare the `str` representation of the choices. Of course, those must be unique then. But this should be the case anyway, otherwise the list entries are not distinguishable.

Eventually one could drop the `self._choices` entirely and just rely on the dict. But this may change the order of choices, since dict does not maintain order. Or go for `OrderedDict` which requires an additional import. What is the preferred way here?

bilderbuchi:
> dict does not maintain order. 

From python 3.6, dicts retain insertion order as an implementation detail of cpython, from 3.7 this is guaranteed. Our minimum supported version is 3.6, and I am not aware of any usage outside cpython, so i think it's fine to rely on the ordering already.

pymeasure/experiment/parameters.py
```diff
@@ -287,8 +296,8 @@ def value(self):
 
     @value.setter
     def value(self, value):
-        if self._choices is not None and value in self._choices:
-            self._value = value
+        if self._choices is not None and str(value) in self._choices_dict.keys():
```
CasperSchippers, line 299: Think the error raise in case `self._choices` is `None` could be clearer; now if choices is not defined, the error is not very clear on that.

CasperSchippers (commented):
Look solid to me. I've only got a single remark.

Concerning the leaving out of `self._choices`: I agree that it should be fine to rely fully on the dict and think that it would be an easy change (I think replacing `self._choices` by either `choices` in `__init__` or by `self._choices_dicit`, maybe with `.values`, should suffice at most places). In that case, it would be good I think to also add a test to ensure that the order is indeed maintained, in case someone alters the code in the future.

pymeasure/experiment/parameters.py
moritzj29: I added a separate check for `None` resulting in a more specific error message.

moritzj29:
Ok, thanks for all the input and remarks! `self._choices` is a `dict` now and the property `self.choices=tuple(self._choices.values())` for the UI etc. I hope this is fine. Also I added tests.

pymeasure/experiment/parameters.py
```diff
@@ -293,6 +293,7 @@ def __repr__(self):
 
 class ListParameter(Parameter):
     """ :class:`.Parameter` sub-class that stores the value as a list.
+    String representation of value must be unique.
```
CasperSchippers, line 296: One minor question: would it be possible to enforce/error-check this? I think that could help users a lot. (If so, a small test that check whether an error is raised might also be nice)

(maybe change `value` to `values` or `choices`)

CasperSchippers (commented):
This implementation seems to work very well, I've only got a minor remark

PS My apologies for the (very) long delay.

pymeasure/experiment/parameters.py
msmttchr: @moritzj29 this is very useful fix for `ListParameter`. If you find time, could please address @CasperSchippers comment, so that we can proceed to merge?

moritzj29: added as suggested, thanks for the input!

CasperSchippers (approved):
Looks good to me as well; thanks for the nice work

Issue #136, opened 2018-04-17
Supporting alternative output formats


mivade:
CSV is useful as a human-readable, easily-transportable file format and serves as a good default. Other formats (HDF5, SQL, numpy's format when using `np.save`, etc) also have their advantages and may be more appropriate for certain use cases.

One way to do this is to create a new abstract base `Formatter` class from which various formatters can inherit. Then the `Results` class can take an optional keyword argument `formatter` which tells it what formatter to use. This approach requires a bit of rework, though, since currently `Results` makes a lot of assumptions about using a plaintext CSV format (these assumptions make interacting with e.g. a SQL database more difficult).

Does anyone have any thoughts on this? Are there alternative approaches to supporting different output formats to consider?

mivade:
This would be good as a way of including many supported formats by default, but I think refactoring things a little to allow for custom solutions would be helpful. These ideas are not mutually exclusive, of course.

cjermain:
@mivade, great suggestion! I agree that CSV files are not ideal for every case. The main advantage is that experiment parameters can easily be stored alongside the data in the same file -- this is used extensively though PyMeasure so that old experiments can be opened and re-run. I would consider it a requirement for other storage formats to support storing the experiment parameters. HDF5 and SQL can certainly do this as well, but it will require some work to have consistent behavior.

tinix84:
im saving everything in mat file with this package https://pypi.org/project/mat4py/, for simple data is ok, if you need complex numbers or struct in matlab is better the scipy

carez:
Making pymeasure handle arbitrary data structures would be a huge improvement.

Can somebody with experience in the pymeasure codebase list the required changes to migrate to pandas/xarray, so we can maybe make a plan for a new Results/Formatter beyond csv?

bilderbuchi:
Ping @CasperSchippers, also see #277. 
I'd be careful to ask for "arbitrary datastructures", to be honest. Aside from that, a Pandas formatter/center point (i.e., put stuff into a DataFrame) that then can use Pandas' various export capabilities to output whatever formats Pandas supports (not only csv) sounds best/easiest, also since we already have pandas as a dependency.

CasperSchippers:
I agree that a Pandas center point might be nice (just to see if I understand you correctly, that would mean that the records would be cast into a DataFrame and handled from there; i.e. basically replacing the 'old-style'-formatter with a Pandas-based formatter like the one in PR #277?) to use it's export capabilities.

[...]

Issue #93, opened 2017-07-13
Not possible to save array of data (like a scope trace for example) on every iteration


benmetcalf42:
I have a procedure object which on every iteration captures a full oscilloscope trace containing 2452 separate data points. Additionally, I capture this trace 11 separate times on each iteration for averaging purposes - this gives me a 2D (11*2452) numpy array that I wish to save under a single column ('Scope Trace') on each iteration of my procedure

However - the results object does not seem capable of handling anything other than a single value for each data_column specified

When I pass my 2D numpy array (11*2452) to the data variable in my procedure definition  - I see this gets transformed to a literal string in the results.format method which literally writes out what is displayed in the python console when a large array is printed - i.e. just the first and last 3 values from the array separated by ellipses - this in turn is written to the .csv file which means only 6 out of the 11*2452 values are saved each time

Is there any way to modify the results class to be able to handle arbitrary sized numpy arrays contained within each Data Column? Even better would be to have the option of saving data to a pandas dataframe

I have only just started exploring PyMeasure though so I may have missed something obvious

cjermain:
@benmetcalf42, the Results object is currently limited to a single value per data column. This limitation comes from the way the Results are written to a file, since this is done to the CSV format. If we introduce additional methods/formats for saving the data, then this limitation is not necessary.

I think this would be a great feature. Do you want to spec out what this feature would look like?

Check out `Results._data` and the `Results.data` property. This is a pandas DataFrame where the data resides.

bilderbuchi:
Maybe it would make sense to more fully embrace pandas when processing results/data? The dataframe is well-suited for the time-based aquisition of columns of data, and you could leverage pandas' considerable [I/O capabilties](https://pandas.pydata.org/pandas-docs/stable/api.html#id12) without reinventing the wheel.

cjermain:
@bilderbuchi, I agree. However, we need to retain the header information that provides the Procedure information. So perhaps we want to look at HDF5, which has key value storage that could serve this purpose. Does anyone want to champion this feature of expanding the I/O capabilities of the Results object?

bilderbuchi:
I hope that some day pandas will finally get metadata storage in a data frame. There's a much-discussed issue on their tracker, but it won't go far without outside contributions, apparently.

jmittelstaedt:
[...]

PR #277, opened 2020-05-22
Results as Pandas Dataframe to CSV

moritzj29:
Okay, so this is PR number 4 this week. I'm feeling sorry for all the reviewing that has to be done. Please feel free to postpone/close this, it's just a suggestions and may also remain in my fork only.

As discussed in #93 it is not possible to emit more than one data point per column due to the way the data is written to the CSV files. In this issue (and #136) it was suggested to move towards pandas dataframes for storing and exporting data. With this PR I try to make step in this direction.

An additional formatter is added which takes a pandas dataframe and converts it to CSV. This formatter can be used the same way as the original `CSVFormatter`, only that it takes a whole dataframe instead of a single row of data.
To use this formatter, initialize the `Result` object with the `output_format="CSV_PANDAS` argument and emit a pandas dataframe to the result object.
So far, there is no check whether the dataframe matches the `DATA_COLUMNS` in size or order. I will try to add this.

Especially in #136 a complete transition from CSV to pandas is discussed to take advantage of the I/O capabilities of pandas. This PR does not do that since the header and all the file handling remains unchanged. This may be a next step, as far as I understand pandas does support handling of metadata now?

Anyhow, for those who just want to save some scope traces or spectral data, this will do the trick. Be aware though, that large amounts of data may freeze the plotting window. I tried so far with 31MB scope traces (960000 points) which are saved nicely to CSV but are not processable anymore by the managed window.

### DIFF
--- a/pymeasure/experiment/results.py
+++ b/pymeasure/experiment/results.py
@@ -92,6 +92,42 @@ class CSVFormatter(logging.Formatter):
         return self.delimiter.join(self.columns)
 
 
+class CSVFormatter_Pandas(logging.Formatter):
+    """ Formatter of data results """
+
+    def __init__(self, columns, delimiter=',', line_break='\n'):
+        """Creates a csv formatter for a given list of columns (=header).
+
+        :param columns: list of column names.
+        :type columns: list
+        :param delimiter: delimiter between columns.
+        :type delimiter: str
+        """
+        super().__init__()
+        self.columns = columns
+        self.delimiter = delimiter
+        self.line_break = line_break
+
+    def format(self, record):
+        """Formats a record as csv.
+
+        :param record: record to format.
+        :type record: pandas.DataFrame
+        :return: a string
+        """
+        return record.to_csv(
+            sep=self.delimiter,
+            header=False,
+            index=False,
+            # explicit line_terminator required, otherwise Windows
+            # uses \r\n which results in double blank lines
+            line_terminator=self.line_break
+        )
+
+    def format_header(self):
+        return self.delimiter.join(self.columns)
+
+
 class Results(object):
     """ The Results class provides a convenient interface to reading and
     writing data in connection with a :class:`.Procedure` object.
@@ -104,6 +140,8 @@ class Results(object):
     :param procedure: Procedure object
     :param data_filename: The data filename where the data is or should be
                           stored
+    :param output_format: Formatter which converts the emitted result data
+                          so it can be written to file, defaults to CSV
     """
 
     COMMENT = '#'
@@ -111,7 +149,7 @@ class Results(object):
     LINE_BREAK = "\n"
     CHUNK_SIZE = 1000
 
-    def __init__(self, procedure, data_filename):
+    def __init__(self, procedure, data_filename, output_format='CSV'):
         if not isinstance(procedure, Procedure):
             raise ValueError("Results require a Procedure object")
         self.procedure = procedure
@@ -119,7 +157,14 @@ class Results(object):
         self.parameters = procedure.parameter_objects()
         self._header_count = -1
 
-        self.formatter = CSVFormatter(columns=self.procedure.DATA_COLUMNS)
+        if output_format == 'CSV_PANDAS':
+            self.formatter = CSVFormatter_Pandas(
+                columns=self.procedure.DATA_COLUMNS,
+                delimiter=self.DELIMITER,
+                line_break=self.LINE_BREAK
+            )
+        else:  # default to CSV
+            self.formatter = CSVFormatter(columns=self.procedure.DATA_COLUMNS)
 
         if isinstance(data_filename, (list, tuple)):
             data_filenames, data_filename = data_filename, data_filename[0]


bilderbuchi:
> as far as I understand pandas does support handling of metadata now?

Indeed, with 1.0, pandas [learned](https://pandas.pydata.org/pandas-docs/stable/reference/frame.html#metadata) `Dataframe.attrs` to store metadata (still marked experimental).

PYmeas:
Hello

I am measuring electrical data through IV sweep over a few cycle , for each cycle I save the data into a dataframe then save to a csv. i do get a csv with a dataframe saved for each cycle but it is just a replicate. i ve try the append on dataframe but doesnt seem to make it. withtout the cycle loop and manually appending it works but when try to automate the cycle i cannot iterate the dataframe. i am trying with this example :

```ruby
from pymeasure.instruments.keithley import Keithley2400
import numpy as np
import pandas as pd
import matplotlib.pyplot as plt
import time
from time import sleep
from datetime import datetime
import random

n = random.randint(1, 30)
df = []
for cycle in range(1,4):
        for i in range(3):
                rows = [[i, i + n] for i in range(3)]
                df = pd.DataFrame(rows, columns=["A", "B"])
                '\n'
        data=pd.DataFrame(df)
        data.append(df)
        data.to_csv('testarray.csv', mode="a", index=False)
```

Also, i cannot pass an empty line in between cycle in the csv. thanks for any input.

Best Regards

bilderbuchi:
Your logic seems messed up a bit:

```
for cycle in range(1,4):
        for i in range(3):
                rows = [[i, i + n] for i in range(3)]
                df = pd.DataFrame(rows, columns=["A", "B"])  # This overwrites df with a new dataframe on every i, that means only the last i loop will be stored
                '\n'  # this just says "a string with a newline", and does not do anything with that string
        data=pd.DataFrame(df)  # after running through the i loop, you copy df to a new dataframe, ok.
        data.append(df)  # then you try to append df to a new copy again? so that would yield 2 copies in data?
        data.to_csv('testarray.csv', mode="a", index=False)
```

PYmeas:
for each cycle i need to save each dataframe and save it into a csv file. i dont get the way to iterate the name of the dataframe save it for each cycle so i avoid overwriting it. I ve found some suggestions using of dat._topickle but i don't get the syntax in a loop.

```ruby
n = random.randint(1, 30)
df = []
for cycle in range(3):  # would correspond to 3 measurement cycle of the same measurement
        rows = [[i, i + n] for i in range(3)]
        df = pd.DataFrame(rows, columns=["A", "B"]) #random DF for test  - would like to save each DF for a cycle
        data=pd.DataFrame(df)                   # copy measurement to new DF to save in csv file
        data.to_csv('testarray.csv', mode="a", index=False)
```

bilderbuchi:
I see. This should give you 3 csv files named `testarray_0.csv`, `testarray_1.csv`, `testarray_2.csv`, is that what you need?
```
n = random.randint(1, 30)
df = []
for cycle in range(3):  # would correspond to 3 measurement cycle of the same measurement
        rows = [[i, i + n] for i in range(3)]
        df = pd.DataFrame(rows, columns=["A", "B"]) #random DF for test  - would like to save each DF for a cycle
        df.to_csv('testarray_' + str(cycle) + '.csv', mode="a", index=False)
```

PYmeas:
I tried that it but it saves the same dataframe in 3 different .csv. Ideally i would like to save each DF in the same .csv iteratively each DF would be separated by an empty line .  statistics will be run on all these dataframes ultimately reason why i d prefer saving in one . csv. something like what i describe in the table below. there is 2 ways [https://stackoverflow.com/questions/53756240/how-to-assign-new-variable-in-for-loop-pandas-dataframe-variable-name-assignme](url)  and [https://stackoverflow.com/questions/58013449/how-to-save-a-list-of-dataframes-while-producing-them-iteratively](url) but i cant reproduce.
```
time or date     A  B
1pm              1  2     #this would be DF 1  from first cycle iteration - the DF is the result of iteration through i (my IV sweep)
                                   # an empty line
1pm 1 min        12 26   #this would be DF2 for second cycle iteration

....                                  # and so on
```

bilderbuchi:
Well you are creating the same fake data here in every cycle, so I would expect the file contents to be identical.
I think if you want to run statistics/further processing on the individual cycles, it will be easier if you kept separate files, also because I think you will have a hard time loading a single CSV dataset correctly if you rely on an empty line as a separator - I don't know if that is foreseen in the CSV format or pandas' csv loader.

PYmeas:
Sorry, my mistake. Here i generate a random number twice within a loop of 2 cycles. i get all the dataframes into 1 .csv including the empty line spacing but while cycle 0 is composed of 2 random sets as expected, cycle 1 gives me again the 2 first plus the 2 new ones. saving in multiple .csv as suggested earlier gives the same output but then the last generated file has the complete set of expected data. 
``` ruby
df = []
data = []
for cycle in range(2):
        for i in range(2):
            n = random.randint(1, 50)
            data.append(n)
            data.append(n/2)
            data.append(n*2)
            data.append('\r')
        df = pd.DataFrame({
            'cycle is: ' + str(cycle): '',
            'Data': data
        })
        df.to_csv("testsample.csv", mode="a", index=False)
```
sample

bilderbuchi:
you need to reset `data = []` for every cycle, right now you just keep appending to `data`.

PYmeas:
Thanks for all the feedback. below a running code. Indeed better save each data (measurement i.e will be IV from Keithley 2420) cycle in a different .csv. There is a possibility to save each dataframe in a different sheet of one .xls file using xlsxwriter but i prefer to stick to .csv as suggested therefore handle individual .csv files. Each dataframe (cycle) is saved as depicted in the figure.
``` ruby
df = pd.DataFrame()
for cycle in range(3):
    data = []
    for i in range(1):
        n = random.randint(1, 30)
        now = datetime.datetime.now()
        data = np.append(data, n)
        data = np.append(data, n/2)
        data = np.append(data, n*2)
        df = pd.DataFrame({'cycle:'+ str(cycle): now, 'Data': data})
    print(df)
    df.to_csv('testarray_' + str(cycle) + '.csv', mode="a", index=False)
```
Capture

PYmeas:
Hello,

I am trying to put previous codes in a pymeasure procedure, using this test below to plot a random number as a function of iteration or time in real time with the Plotter. this code works without the time parameter in the definition and i get the time in the .csv file but cannot plot it only iteration vs number is plotted. I d like to see the plot of random number versus time when defining the time variable i get an error "ValueError: FloatParameter value is above the maximum".   

```ruby
import logging

import random
import time
from datetime import datetime
from time import sleep
from pymeasure.display import Plotter
from pymeasure.experiment import Procedure, Results, Worker


    iterations = IntegerParameter('Loop Iterations')
    now = FloatParameter('Time', units='s', default=0)

    DATA_COLUMNS = ['Iteration', 'Time', 'Random Number']

    def startup(self):
          log.info("Setting the time")
          random.now(self.now)

    def execute(self):
                data = {
                    'Iteration': i,
                    'Time': now.strftime("%H:%M:%S"),
                    #'Time': now,
                    'Random Number': random.randint(1, 100)
                }
                self.emit('results', data)
                sleep(self.delay)
                if self.should_stop():
                    break

if __name__ == "__main__":
    console_log(log)

    log.info("Constructing a RandomProcedure")
    procedure = RandomProcedure()
    procedure.iterations = 50
    procedure.now = time.time()

    data_filename = 'random.csv'
    log.info("Constructing the Results with a data file: %s" % data_filename)
    results = Results(procedure, data_filename)

    log.info("Constructing the Plotter")
    plotter = Plotter(results)
    plotter.start()
    log.info("Started the Plotter")

    log.info("Constructing the Worker")
    worker = Worker(results)
    worker.start()
    log.info("Started the Worker")

    log.info("Joining with the worker in at most 1 hr")
    worker.join(timeout=3600) # wait at most 1 hr (3600 sec)
    log.info("Finished the measurement")
```

bilderbuchi:
FloatParameter has a [default maximum](https://pymeasure.readthedocs.io/en/latest/api/experiment/parameters.html#pymeasure.experiment.parameters.FloatParameter) value that you probably have to override.
Aside of that, I'd ask you to open a separate issue if you need more help, to keep this pull request's discussion focused and not distracted by unrelated issues.

AdityaOP:
@moritzj29 I try to use this class as below:
        wavelengths = np.array(wavelength)
        wavelengths = wavelengths * 1e9
        wavelengths = pd.DataFrame(wavelengths)
        powers = np.array(power)
        powers = 10 * np.log10(powers / 1e-3)
        powers = pd.DataFrame(powers)

        for i in range(len(wavelengths)):
            data = {'Wavelength(nm)': wavelengths, 'Power(dBm)': powers}
            self.emit('results', data)
            time.sleep(0.0001)
            if self.should_stop():
                break

and run procedure like this:
data_filename = 'example10.csv'
results1 = Results(procedure1, data_filename, output_format='CSV_PANDAS')
worker1 = Worker(results1)
worker1.start()

I am getting this error,
   msg = self.format(record)
  File "C:\Users\Aditya\Anaconda3\lib\logging\__init__.py", line 869, in format
    return fmt.format(record)
  File "C:\Users\Aditya\Documents\Pymeasure\pymeasure-master\pymeasure\experiment\results.py", line 115, in format
    return record.to_csv(
AttributeError: 'dict' object has no attribute 'to_csv'
Can you please help me out?

moritzj29:
Hello @AdityaOP , thanks for interest in this piece of code! Unfortunately this PR is still not finished and especially missing out on documentation...

I think the problem is, that you need to emit a Pandas DataFrame to the results object, not a dict as it is usually the case. The `to_csv` method is a built-in method for Pandas DataFrames which handles the CSV export.

You set up two dataframes and iterate over the wavelengths, but the emitted data is the same in every iteration (no use of index `i`)? I guess you want one dataframe whith two columns: wavelength and power. This dataframe is then emitted as result.

Of course **this only makes sense if you already get the data in a dataframe/array like form**. If you iterate over the values anyway, you can easily use the standard pymeasure way of emitting a dict of values per point.

So far I only used this to save one dataframe to csv per run. For example one scope trace or one spectrum. If you need to iterate you will probably want to add a column to the dataframe containing the iteration index (or some changing parameter) to separate the CSV data afterwards.

Here is an example is used:

```
def execute(self):
        self.vb.mso.run(autoTrigger=False)

        data = self.vb.mso.read_analog_digital_dataframe()
       # this is a special instrument which already returns data as dataframe


        self.vb.mso.stop()

        self.emit('results', data)
        self.emit('progress', 1)
```
As you can see, I do not use a loop. I write one file per measurement.

I hope this helps!

AdityaOP:
@moritzj29. Before I was emitting single values. And I forget to change the for loop range while emitting data frames. Anyways with your code i am able to emit the dataframes to csv. Thank you very much for your help!!!

Regards,
Aditya

carez:
Hey @moritzj29, what is missing to merge this PR?

moritzj29:
Hey @carez , thanks for your interest! It's been some time since I wrote this, but I recently started to use this code again. It works fine, I use it to save oscilloscope traces.
From my initial post I recall:

> So far, there is no check whether the dataframe matches the DATA_COLUMNS in size or order. I will try to add this.

This would be beneficial I guess... Additionally, there was no feedback from the maintainers so far (probably because it is still marked as draft...)

The PR in some way breaks with the current result processing in pymeasure since it requires a `pd.dataframe` to be sent to the results object. But I think `pandas` is the way to go, since it offers a lot of flexibitlity, as mentioned in #136 .

I invite you to give this implementation a shot! Just pull this dev branch. If you are interested in different export formats based on `pd.dataframe`, it should be quite easy to write an appropriate formatting class (take `CSV_Formatter_Pandas` as a reference).

pymeasure/experiment/results.py
```diff
@@ -104,22 +140,31 @@ class Results(object):
...
 
-        self.formatter = CSVFormatter(columns=self.procedure.DATA_COLUMNS)
+        if output_format == 'CSV_PANDAS':
+            self.formatter = CSVFormatter_Pandas(
+                columns=self.procedure.DATA_COLUMNS,
+                delimiter=self.DELIMITER,
+                line_break=self.LINE_BREAK
+            )
+        else:  # default to CSV
+            self.formatter = CSVFormatter(columns=self.procedure.DATA_COLUMNS)
```
CasperSchippers, line 167: maybe it would be nicer if instantiating the formatter could be done in a more uniform way (to prevent clutter, especially _if_ in the future new formatters pop up); maybe something like this:
```python
def __init__(.... output_format='CSV'):
    ...
    ...
    formatter_class = {
        'CSV': CSVFormatter,
        'CSV_PANDAS': CSVFormatter_Pandas,
    }[output_format]
    self.formatter = formatter_class(.....)
```
Here the arguments for the formatter-classes need to be uniform; an abstract `Formatter` class might help for that, see the general remark for that.

pymeasure/experiment/results.py
```diff
@@ -104,22 +140,31 @@ class Results(object):
     :param procedure: Procedure object
     :param data_filename: The data filename where the data is or should be
                           stored
+    :param output_format: Formatter which converts the emitted result data
+                          so it can be written to file, defaults to CSV
```
CasperSchippers, line 144: Please add the possible options here.

pymeasure/experiment/results.py
```diff
@@ -92,6 +92,42 @@ def format_header(self):
...
+    """ Formatter of data results """
+
+    def __init__(self, columns, delimiter=',', line_break='\n'):
+        """Creates a csv formatter for a given list of columns (=header).
+
+        :param columns: list of column names.
+        :type columns: list
+        :param delimiter: delimiter between columns.
+        :type delimiter: str
+        """
```
CasperSchippers, line 105: This docstring needs updating

CasperSchippers (changes requested):
Thanks for the contribution, and I'm very sorry you had to wait this long for a review. I think this is a very good improvement and seems to work very well. I left a couple of small changes/suggestions at the code.

One thing I would suggest to add a few lines to the `format`-method such that it can also handle "old-style" records (i.e. a dict of values, rather than a DataFrame), especially in the implementation as this is now (i.e. with an additional `CSVFormatter_Pandas` alongside the old `CSVFormatter`). If the old formatter would be replaced by this new version (which I would argue for, see #136), this converting between dict and DataFrame could be either done here or at another place; I'm not yet sure where the best place for this would be.

A final thing to discuss (see also #136) is whether we want to have an abstract Formatter class and to base this (and for now the old-style formatter) on.

Apart from these discussion, there is need for documentation/tutorials for how to use this.

PS I'd also suggest to update this branch from master.

pymeasure/experiment/results.py
moritzj29: done

pymeasure/experiment/results.py
moritzj29: done

pymeasure/experiment/results.py
```diff
@@ -92,6 +92,42 @@ def format_header(self):
...
+        :return: a string
+        """
+        return record.to_csv(
+            sep=self.delimiter,
+            header=False,
+            index=False,
+            # explicit line_terminator required, otherwise Windows
+            # uses \r\n which results in double blank lines
+            line_terminator=self.line_break
+        )
```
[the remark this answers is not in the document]
moritzj29: done

pymeasure/experiment/results.py
```diff
@@ -92,6 +92,42 @@ def format_header(self):
...
+
+    def format_header(self):
+        return self.delimiter.join(self.columns)
```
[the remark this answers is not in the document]
moritzj29: I'm not sure either... but since in the future eventually all file handling is supposed to be done by pandas, let's have it take care of the header column as well. I agree.

pymeasure/experiment/results.py
moritzj29: I followed your suggestion. If/When new formatters come up, one probably needs to adapt this anyway...

moritzj29:
Hi everyone, I addressed the very valid comments of @CasperSchippers and I think the code is ready for some review now. Unfortunately, I will not be able to extend this any further. My time at the university ends this year, so I won't have access to instruments anymore. Thanks for this wonderful project, it was a great help!

pymeasure/experiment/results.py
```diff
@@ -138,12 +149,54 @@ def format(self, record):
...
+        :param record: record to format.
+        :type record: pandas.DataFrame or dict
+        :return: str
+        """
+        if isinstance(record, pd.DataFrame):
+            pass
+        elif isinstance(record, dict):
+            # ensure correct order of columns
+            record = (record[c] for c in self.columns)
+            record = pd.DataFrame(record)
```
CasperSchippers, line 176: This doesn't seem to work for me, but changing it to the following does work and keep the correct order:
```suggestion
            record = pd.DataFrame([record], columns=self.columns)
```

pymeasure/experiment/results.py
```diff
@@ -138,12 +149,54 @@ def format(self, record):
...
+        else:
+            log.error('Formatting of data failed. ' +
+                      'Pandas dataframe or dict required.')
+            return ''
```
CasperSchippers, line 180: This may be a too gentle way to handle a wrong type of data provided; this way, the measurement will continue without storing any data if it is provided as a wrong type.

I suggest changing it to the following:
```suggestion
            raise TypeError('Formatting of data failed. '
                            'Pandas dataframe or dict required.')
```

pymeasure/experiment/results.py
```diff
@@ -138,12 +149,54 @@ def format(self, record):
...
+        """Formats a record as csv using pandas built-in ``to_csv`` method.
+        Accepts a pandas dataframe or a dict of values matching
+        the given list of columns.
+
+        :param record: record to format.
+        :type record: pandas.DataFrame or dict
+        :return: str
+        """
+        if isinstance(record, pd.DataFrame):
+            pass
```
CasperSchippers, line 172: I'd suggest to ensure the correct order and columns are printed using a line like this:
```suggestion
            record = record.reindex(columns=self.columns)
```

CasperSchippers (changes requested):
Thanks for the improvements!!!
I just have some remarks/suggestions and then I think this seems finished to me. 

A final suggestion is to add a remark about this in the tutorial/documentation, to make the feature discoverable.

Issue #233, opened 2019-11-08
FloatParameter default<0.01 changes to 0


moritzj29:
When setting the default value of a `FloatParameter` to values smaller than `0.01`, the initial value in the GUI just states `0`. This happens also for exponential notations, e.g. `1e-6`. When setting the value in the GUI everything works as expected. So it has to be related to the initial setting of the value.

I tried tracing down the error but I had no success so far. The change happens in the `set_parameter` method of the `Input` class in `inputs.py`. The value passed to `self.setValue(parameter.default)` is fine. Reading the set value with `self.value()` directly afterwards gives `0`.

relevant part of the Input class code:
```
class Input(object):
    """
    Mix-in class that connects a :mod:`Parameter <.parameters>` object to a GUI
    input box.
    :param parameter: The parameter to connect to this input box.
    :attr parameter: Read-only property to access the associated parameter.
    """

    def __init__(self, parameter, **kwargs):
        super().__init__(**kwargs)
        self._parameter = None
        self.set_parameter(parameter)

    def set_parameter(self, parameter):
        """
        Connects a new parameter to the input box, and initializes the box
        value.
        :param parameter: parameter to connect.
        """
        self._parameter = parameter

        if parameter.default is not None:
            # here the parameter.default value is correct
            self.setValue(parameter.default)
            # here self.value() gives 0

        if hasattr(parameter, 'units') and parameter.units:
            self.setSuffix(" %s" % parameter.units)
```

Going further down into QtGUI's setValue method exceeds my python/programming skills... I have no idea how this works.

Its interesting that the number of decimals of the `QDoubleSpinBox` is set to 2. (by default?) This would explain the rounding to 0. But on the other hand, this value remains 2 when values with higher precision are entered in the GUI and these are displayed and passed correctly.

It's actually not a huge problem, since the values can still be set in the GUI. But it would be handy...

cjermain:
Thanks for reporting this @moritzj29. This appears to be related to #225. I will look at this problem as time allows, but certainly welcome a PR from others.

moritzj29:
should be fixed in #278, so closing this issue

Issue #225, opened 2019-09-11
Error creating QT GUI when FloatParameter minimum >=0.005


jeffmincey:
pymeasure==[0.7.0](https://github.com/ralph-group/pymeasure/tree/v0.7)

if a FloatParameter param "minimum" is set to a value greater than 0.005 as the following: 
```
delay = FloatParameter('Delay Time', units='s', minimum=.1, maximum=5, default=0.2)
```

then QT gui will fail with the following error:
```
Traceback (most recent call last):
  File "...\lib\site-packages\pymeasure\display\inputs.py", line 250, in validate
    if self._parameter.units:
AttributeError: 'NoneType' object has no attribute 'units'
```

---

Also, likely related, if minimum is set >0 and <0.005 by adding the following test case to TestScientificInput parametrize'd inputs
```
[0.004, 5.5, 3.3],
```
then this error is thrown in pytest, however the Qt GUI behaves correctly.
```
>       assert sci_input.minimum() == min_
E       assert 0.0 == 0.004
E        +  where 0.0 = <built-in method minimum of ScientificInput object at 0x0000022AA25C5288>()
E        +    where <built-in method minimum of ScientificInput object at 0x0000022AA25C5288> = <pymeasure.display.inputs.ScientificInput object at 0x0000022AA25C5288>.minimum

tests\display\test_inputs.py:155: AssertionError
```

related to:  #118, #115

jeffmincey:
just in case, here's the version of Qt I'm using:
```
In [2]: pyqtgraph.Qt.VERSION_INFO
Out[2]: 'PyQt5 5.9.2 Qt 5.9.6'
```

cjermain:
Thanks for reporting this @jeffmincey. I'll take a look at adding these test as soon as time allows. If anyone else is interested, PRs are welcome.

bilderbuchi:
I think this is also fixed with #278.

PR #278, opened 2020-05-26
Fix ScientificInput/FloatParameter precision and default value

moritzj29:
Hi,
this should fix the issues described in #233 and #225 which are related to the default precisison of 2 of `QDoubleSpinBox`. I introduced the new keyword argument `decimals` for the `FloatParameter` to adjust the precision of the input box (and the value passed to the instrument!) manually. I set the default to 15, as I think femto-something should be sufficient for most parameters.

I also changed to set the `self._parameter` attribute in the derived `set_parameter` method already, since this is required by the validator. The validator may already get called when setting minimum, maximum and decimals and requires `self._parameter.units`. (also seen in #225)

As always, happy for remarks and improvements on the code.

### DIFF
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -237,14 +237,18 @@ class ScientificInput(QtGui.QDoubleSpinBox, Input):
 
     def set_parameter(self, parameter):
         # Override from :class:`Input`
-        self.setMinimum(parameter.minimum)
-        self.setMaximum(parameter.maximum)
+        self._parameter = parameter  # required before super().set_parameter
+        # for self.validate which is called when setting self.decimals()
         self.validator = QtGui.QDoubleValidator(
             parameter.minimum,
             parameter.maximum,
-            10, self)
+            parameter.decimals,
+            self)
+        self.setDecimals(parameter.decimals)
+        self.setMinimum(parameter.minimum)
+        self.setMaximum(parameter.maximum)
         self.validator.setNotation(QtGui.QDoubleValidator.ScientificNotation)
-        super().set_parameter(parameter) # default gets set here, after min/max
+        super().set_parameter(parameter)  # default gets set here, after min/max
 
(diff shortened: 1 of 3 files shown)


bilderbuchi:
Since we already depend on PyQtGraph, check out if its custom [SpinBox](https://pyqtgraph.readthedocs.io/en/latest/widgets/spinbox.html) for scientific use does not make more sense to use here instead of the vanilla QDoubleSpinbox. I have migrated some other code to it just recently (I only found it this week) and found it much better to use!

moritzj29:
thanks for the comment! It looks like this SpinBox would be perfect for the use cases in pymeasure. I tried using SpinBox from PyQtGraph as the base class but it throws some exceptions upon initialization. Probably it's simple, but I'm not able to figure out. I have no experience with this Qt stuff...
Anyhow, even with the new SpinBox the number of decimals needs to be set correctly upon initialization of the SpinBox. This PR solves the reported issues. Moving to pyqtgraph's SpinBox might be worth a separate PR.

bilderbuchi:
Granted; the question is if it makes sense to introduce some code only to immediately throw it out again. 
The implementation I have found working nicely in my usecase is this:
```
            value_widget = pg.SpinBox(
                parent=self,
                siPrefix=True,
                step=0.1,  # 10% change with every step change
                dec=True,  # step change relative to current value
                minStep=1 if intdisplay else 1e-9,  # can do either int or float display, intdisplay is a bool
                int=intdisplay,
                decimals=3,
            )
```
which even gives nice functionality for arrowup/arrowdown to tweak values. Killer feature imo is the ability to enter "5n" and get a value of 5e-9, etc, that is soo useful (I have everything from kilo- to micro order in one context)

CasperSchippers (approved):
Thanks for this PR. I have no further remarks.

bilderbuchi:
Would you open an issue about that so we don't lose track, please?

PR #281, opened 2020-06-06
Correct pyVISA kwarg query_delay.

### LINKED ISSUE #272: PyVISA keyword `delay` replaced by `query_delay`
So, since pyVISA 1.6 (quite some time ago), it seems that the `delay` kwarg needs to be replaced by `query_delay` - I think we just missed that.
https://pyvisa.readthedocs.io/en/latest/faq/migrating.html#removal-of-instrument-delay-and-added-instrument-query-delay
As we [still have](https://github.com/ralph-group/pymeasure/blob/e48f9d679d6ee970e2e875d2fc9a5679378b07aa/pymeasure/adapters/visa.py#L59) the `delay` defined as safe, we should change that to `query_delay`.
The only other place that I found (with Github) where we used that is [in an Agilent device](https://github.com/ralph-group/pymeasure/blob/07cf309b8d113931525c8cc216cc704f2b7ff413/pymeasure/instruments/agilent/agilent8257D.py#L248).
@cjermain is this something where we need to keep API consistency with other adapters beside the VISAAdapter, or do we just change those two spots and be done? I think the latter, just want to make sure.

bilderbuchi:
Correct VISAAdapter keyword argument "query_delay" instead of the removed "delay".

This was overlooked in the migration to PyVISA 1.6. Also remove instances where the old "delay" kwarg was defined but unused.

Add the instrument test only if pyvisa-sim is installed. pyvisa-sim is not available via conda-forge, so don't place it in the list of dependencies just for this one test.

Closes #272

### DIFF
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -56,7 +56,7 @@ class VISAAdapter(Adapter):
         self.resource_name = resourceName
         self.manager = visa.ResourceManager(visa_library)
         safeKeywords = ['resource_name', 'timeout',
-                        'chunk_size', 'lock', 'delay', 'send_end',
+                        'chunk_size', 'lock', 'query_delay', 'send_end',
                         'values_format', 'read_termination', 'write_termination']
         kwargsCopy = copy.deepcopy(kwargs)
         for key in kwargsCopy:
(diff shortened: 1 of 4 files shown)


bilderbuchi:
@cjermain I notice that the Travis CI job did not run for this PR - do you know what's up there?

bilderbuchi:
Huh, in fact it [did run](https://travis-ci.org/github/ralph-group/pymeasure/builds/695361848), just not phone back. Maybe this disappears by itself.

PR #287, opened 2020-06-12
Update agilent8257D.py

axs5489:
Fixed copy-pasta. second start_power changed to stop_power


axs5489:
https://github.com/ralph-group/pymeasure/issues/286

PR #295, opened 2020-06-29
Adding Apex AP1000

Mike-HubGit:


pymeasure/instruments/apex/apex1000.py
```diff
@@ -0,0 +1,176 @@
...
+from pymeasure.instruments.validators import strict_discrete_set, \
+    truncated_discrete_set, truncated_range
+
+import numpy as np
+import time
+import re
+
+
+class ApexAP1000(Instrument):
+    """Represents Apex's AP1000 Multi-test platform"""
```
CasperSchippers, line 35: Seeing that there is a need for specifying a slot first after initialization, it might be nice to add an example (with explanation) in the docstring.

pymeasure/instruments/apex/apex1000.py
```diff
@@ -0,0 +1,176 @@
...
+        return ApexAP1000.DEV_KINDS[kind]+"[%02d]:"%self.slot
+
+    @property
+    def slot(self):
+        """ Get current slot number
+        """
+        if self.__slot in ApexAP1000.SLOTS:
+            return self.__slot
+        else:
+            raise ValueError('Bad slot number !')
```
CasperSchippers, line 61: For debugging purposes, I would suggest adding an `elif self.__slot is None` case to check whether the slot has been set at al.

pymeasure/instruments/apex/apex1000.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # Powermeter module related functions
+    @property
+    def PM_averages(self):
+        """ PowerMeter module.
+        Get wanted number of averages for current slot.
+        """
+        return int(self.ask(self.headStr('PM')+'AVG?'))
```
CasperSchippers, line 80: I would suggest using the `cast=int` keyword argument for the `self.ask` method rather than casting to int afterwards.

pymeasure/instruments/apex/apex1000.py
```diff
@@ -0,0 +1,176 @@
...
+
+import numpy as np
+import time
+import re
+
+
+class ApexAP1000(Instrument):
+    """Represents Apex's AP1000 Multi-test platform"""
+    SLOTS = range(1,9)
+    DEV_KINDS = {'PM':'POW','TF':'FLT','TL':'TLS'}
```
CasperSchippers, line 37: Could you comment on the usage of this dict? It seems it is only used in the `headStr` method, which itself is only called with hardcoded arguments; so why not directly use the 'correct' strings when calling `headStr` rather than using a dict for lookup?

pymeasure/instruments/apex/apex1000.py
```diff
@@ -0,0 +1,176 @@
...
+    def PM_averages(self,num):
+        """ PowerMeter module.
+        Sets wanted number of averages for current slot.
+        """
+        self.write(self.headStr('PM')+'AVG %d'%num)
+
+    def PM_getPower(self,channel,unit='W'):
+        """ PowerMeter module.
+        Get actual power on given channel with given unit.
+        - channel can be {}.
```
CasperSchippers, line 92: I think you forgot to format this string to fill in the valid values.
Also, please keep in mind that the newlines are not always preserved in the documentation.

pymeasure/instruments/apex/apex1000.py
```diff
@@ -0,0 +1,176 @@
...
+            self.__slot=num
+        else:
+            raise ValueError('Bad slot number !')
+
+    # Powermeter module related functions
+    @property
+    def PM_averages(self):
+        """ PowerMeter module.
+        Get wanted number of averages for current slot.
+        """
```
CasperSchippers, line 79: Could you make a single docstring for the properties (i.e. including the docs for both the set and get function) in the get function, as this is the only docstring that is displayed in the documentation. (Holds also for other properties).

pymeasure/instruments/apex/apex1000.py
```diff
@@ -0,0 +1,176 @@
...
+        """ PowerMeter module.
+        Get actual power on given channel with given unit.
+        - channel can be {}.
+        - unit can be {}.
+        """
+        if unit not in ApexAP1000.PM_UNIT:
+            raise ValueError('Unknow physical unit during power measurement')
+        if channel not in ApexAP1000.PM_CHANNELS:
+            raise ValueError('Unknow channel during power measurement')
+        str = {'W':'MW','mW':'MW','dBm':'DBM'}
```
CasperSchippers, line 99: It can be dangerous / confusing to use `str` as this is a standard type of python and hence is already defined. I would suggest changing this to something like `unit_str`.

pymeasure/instruments/apex/apex1000.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def PM_setWavelength(self,channel,wavelength):
+        """ PowerMeter module.
+        Sets wavelength for given channel for calibration
+        - channel can be {}.
+        """
+        if channel not in ApexAP1000.PM_CHANNELS:
+            raise ValueError('Unknow channel during power measurement')
+        sentStr = self.headStr('PM')+'SETWAVELENGTH[%d] %g'%(channel,wavelength)
+        return self.write(sentStr)
```
CasperSchippers, line 113: I think the return is not required over here.

pymeasure/instruments/apex/apex1000.py
```diff
@@ -0,0 +1,176 @@
...
+    def TF_wavelength(self):
+        """ Tunable Filter module.
+        Gets tunable filter wavelength.
+        """
+        return int(self.ask(self.headStr('TF')+'TWL?'))
+
+    @TF_wavelength.setter
+    def TF_wavelength(self,wavelength):
+        """ Tunable Filter  module.
+        Sets tunable filter wavelength.
```
CasperSchippers, line 126: It would be nice if you can comment on the valid values and units for setting (and getting) the wavelength. Maybe also consider using a validator. (This comment also holds for other functions/properties)

CasperSchippers (changes requested):
Thanks for the contribution. I've had a look at the PR (sorry it took so long) and added a couple of remarks.

When performing some changes, it could also be useful (for merging and testing purposes) to update this branch from the master branch.

mikamyara:
@bmoneke : I am not sure I understand what you are asking. I am not really skilled at github. Can you explain ?

mikamyara:
I am ok with your proposal, but I need help about how to do. I do not use github very often ...

bmoneke:
Do you know git (which is the foundation of Github) or do you use it rarely as well?

When you do changes, you can commit them. When you "push" these changes, this pull request here, will be updated to reflect these changes.

bmoneke:

PR #308, opened 2020-08-19
Itc503 update

CasperSchippers:
In this pull-request a few modifications and addons to the ITC503 package are proposed:
- a changeable temperature range (as optional parameters to the class)
- an improved wait for temperature (often a communication error arises in the communication with the ITC; this update allows for a number of there errors to arise before terminating the wait)
- properties to control the heater and the gass-flow

### DIFF
--- a/pymeasure/instruments/oxfordinstruments/itc503.py
+++ b/pymeasure/instruments/oxfordinstruments/itc503.py
@@ -47,6 +47,7 @@ class ITC503(Instrument):
         print(itc.temperature_1)        # Print the temperature at sensor 1
 
     """
+    _T_RANGE = [0, 301]
 
     control_mode = Instrument.control(
         "X", "$C%d",
@@ -124,7 +125,7 @@ class ITC503(Instrument):
         the ITC in kelvin. """,
         get_process=lambda v: float(v[1:]),
         validator=truncated_range,
-        values=[0, 301]
+        values=_T_RANGE
     )
 
     temperature_1 = Instrument.measurement(
@@ -183,7 +184,8 @@ class ITC503(Instrument):
         get_process=lambda v: float(v[1:]),
     )
 
-    def __init__(self, resourceName, clear_buffer=True, **kwargs):
+    def __init__(self, resourceName, clear_buffer=True,
+                 max_temperature=301, min_temperature=0, **kwargs):
         super(ITC503, self).__init__(
             resourceName,
             "Oxford ITC503",
@@ -196,6 +198,9 @@ class ITC503(Instrument):
         # Clear the buffer in order to prevent communication problems
         if clear_buffer:
             self.adapter.connection.clear()
+
+        self._T_RANGE[0] = min_temperature
+        self._T_RANGE[1] = max_temperature
 
     def wait_for_temperature(self, error=0.01, timeout=3600,
                              check_interval=0.5, stability_interval=10,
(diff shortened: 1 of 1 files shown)


pymeasure/instruments/oxfordinstruments/itc503.py
```diff
@@ -196,30 +229,47 @@ def wait_for_temperature(self, error=0.01, timeout=3600,
...
                 break
 
-            if timeout >= 0 and (time() - t0) > timeout:
+            if timeout > 0 and (time() - t0) > timeout:
                 raise TimeoutError(
                     "Timeout expired while waiting for the Oxford ITC305 to \
                     reach the set-point temperature"
                 )
 
+            if max_comm_errors > 0 and comm_errors > max_comm_errors:
```
CasperSchippers, line 268: Now max_comm_errors == 0 means that an infinite amount errors is allowed, but now it is not possible to disallow any errors

bilderbuchi: Maybe a better approach: use as kwarg default `max_comm_errors=None`, check with `if max_comm_errors is not None ...`. then it becomes possible to have max_comm_errors=0.

pymeasure/instruments/oxfordinstruments/itc503.py
```diff
@@ -177,10 +206,14 @@ def __init__(self, resourceName, clear_buffer=True, **kwargs):
...
 
+        self._T_RANGE[0] = min_temperature
+        self._T_RANGE[1] = max_temperature
+
     def wait_for_temperature(self, error=0.01, timeout=3600,
                              check_interval=0.5, stability_interval=10,
                              thermalize_interval=300,
-                             should_stop=lambda: False):
+                             should_stop=lambda: False,
+                             max_comm_errors=0):
```
bilderbuchi, line 216: I find a value of "0" meaning "infinity" confusing. Maybe use `None` instead -- see comment below.

pymeasure/instruments/oxfordinstruments/itc503.py
```diff
@@ -177,10 +206,14 @@ def __init__(self, resourceName, clear_buffer=True, **kwargs):
         if clear_buffer:
             self.adapter.connection.clear()
 
+        self._T_RANGE[0] = min_temperature
+        self._T_RANGE[1] = max_temperature
```
bilderbuchi, line 210: Have you confirmed that this works correctly? Keep in mind that `Instrument.control` is a property *factory*, which iiuc creates the property once at instantiation time. I am not sure if later changes (e.g. during `__init__` or even later) to `self._T_RANGE` have an effect on the property.

bilderbuchi (changes requested):

pymeasure/instruments/oxfordinstruments/itc503.py
CasperSchippers: I agree; fixed this issue.

pymeasure/instruments/oxfordinstruments/itc503.py
CasperSchippers: This seems to work so far; the fact that a mutable (list) is used seems to ensure that the property can be used like this. However, I would agree that this is not the cleanest solution, but didn't see another way. If you have a suggestion, I'd be happy to change it.

bilderbuchi: No other suggestion, I was just surprised that this works at all! But indeed, the mutability seems to be the saving grace here, this will come in handy in other instruments. I think this is quite elegant.

bilderbuchi (approved):

PR #313, opened 2020-08-31
Dev/scpi instrument

thosou:
Hi,

This a pull request to address the SCPI part of #170.

Instead of implementing the property dynamically, I propose to implement them in the class. I believe that will avoid problem down the road with inheritance or things like that.
There is already a test in the id property to check whether the instrument is SCPI type:

    @property
    def id(self):
        if self.SCPI:
            return self.adapter.ask("*IDN?").strip()
        else:

I think it is consistent to use the same pattern for the other scpi specific attributes. An exception will be raised if the instrument is not SCPI.
I also added a close method and read_stb.


pymeasure/instruments/instrument.py
```diff
@@ -80,7 +72,38 @@ def id(self):
...
             return self.adapter.ask("*IDN?").strip()
         else:
-            return "Warning: Property not implemented."
+            raise NotImplementedError("Only implemented for SCPI instruments. Must be re-implemented by the subclass")
+
+    @property
+    def status(self):
+        """ Returns the status of the instrument"""
+        if self.SCPI:
+            return int(self.adapter.ask("*STB?"))
```
CasperSchippers, line 81: Is there a reason you use the `ask` method of `self.adapter` rather than simply `self.ask`? In the later case you can also use the `cast=int` optional argument to cast the value to an int, rather than doing it afterwards. (The same holds for line 89)

CasperSchippers (commented):
Thanks for this contribution. I've had a look and added just a singular question.

msmttchr (commented):
This PR addressed some important problems in `Instrument` class, butI think that changes included in this PR are addressed by merged PR #453.
@thosou, do you agree to close this pull request then ?

bilderbuchi:
Yes, this has been implemented in #453 already, so we can close this.

PR #317, opened 2020-09-02
Add directory input

paulgoulain:
Add a line edit for the directory where the result will be saved.
The content of this line edit can be retrieve with the method self.directory_line.text().
A switch is available to enable it or not.


paulgoulain:
Any thoughts, inputs, feedback ? Do you want me to elaborate on tee purpose of the PR ?

paulgoulain:
Did a wrong operation on the git. I'll recreate it.

PR #322, opened 2020-09-24
Adding a directory input 

paulgoulain:
Allow to add a filepath entry in a ManagedWindow. The content of this entry can be retrieve in the queue function for example.
A switch is available in the ManagedWindow init to enable  it or not.

Here is an example of this directory input in a managed window.


CasperSchippers:
Thank you for this suggestion. I really like the idea!
Before I get into a detailed review: maybe it would be nice to implement this as an additional type of parameter (DirectoryParameter) in line with other inputs (like FloatParameter), with an associated input field, such that users can add it a bit more flexibly and use the directory from within the measurement Procedure. What do you think about that? (if you want, I would be happy to help implement it in such a way)

paulgoulain:
I'm not against the idea of creating this DirectoryParameter but what would be the benefit ?

CasperSchippers:
For example, in one of the measurement scripts in my group, an extensive configuration file is loaded; in that case you need access to the selected directory from within the measurement procedure, and could benefit from being able to add multiple DirectoryParameters.

paulgoulain:
I see. For this particular case (loading a file that is needed as an input parameter), a DirectoryParameter or a PathParameter makes sense, because the parameter is used inside the Procedure class. 
But the directory path, at least in my case, is used in the ManagedWindow class only, in the queue function.

```python
    def queue(self):

        directory = self.directory_line.text() #Ask the directory line to return the path 
        filename = unique_filename(directory, prefix='IV') # Then use this path for generating a unique filename

        results = Results(procedure, filename) # Use the filename for the Results class

```

I don't know how people are handling the choice of where their result files will be saved.

CasperSchippers:
That is a good question; I'm not sure how other people do this, but I now use an ordinary Parameter, which allows both accessing from within the Procedure and within the queue function, and allows for easily setting a default value.

paulgoulain:
Also concerning the directory input, I planned to implement some special sequence of characters that would be interpreted when you start your measurement. I made a little program to test the concept. Here the sequence of characters that is interpreted is <date> which is changed into the time currently is (I know it should have been <time> I got lazy and didn't change it). Many other could be implemented like actual date, or also parameters that are currently used for the measurement.

image

I think this concept could be handy. But I don't know if it is possible to implement that if we chose to use a parameter for the directory.

CasperSchippers:
These placeholders could be a very useful addition indeed, especially for filenames. It might be useful if it would be possible to have placeholders for other inputs (though I'm not sure it this will be difficult to implement). I also understand your concern for how to implement this when using a parameter.

Regarding the way of implementing: as both ways of implementing (i.e. the current PR or using a parameter) could serve different use-cases (I also discussed this with @cjermain), I would say that it might indeed be best to leave the implementation of the PR as it is.

Especially with the placeholders, I think this feature will also need a section in the tutorial.

paulgoulain:
Ok. Shoud I add to this PR the filename input too ? 
About the placehoders (thanks for the name, I didn't know how to name this), I think the best way to handle them would be to add a function, that would take the raw string and returns the processed string. But that's something for another PR.

CasperSchippers:
I think it's better to do that in a separate PR.

The handling of the placeholder could indeed be a separate PR. Could you make an issue or PR to keep track of this suggestion?

CasperSchippers (changes requested):
I just had a look at the changes; just a few small remarks.
Apart from changes indicated in the files:

- A suggestion for ease of use: it could be useful to implement a getter-property to the ManagedWindow to obtain the directory value
- Please write a section in the tutorial (can be fairly brief I think) on how to use this widget
- Please add yourself to the lists of authors (AUTHORS.TXT and docs/about/authors.rst)

pymeasure/display/widgets.py
```diff
@@ -1006,3 +1006,28 @@ def eval_string(string, name=None, depth=None):
 
         evaluated_string = numpy.array(evaluated_string)
         return evaluated_string
+
+class DirectoryLineEdit(QtGui.QLineEdit):
+
+    def __init__(self, parent=None):
+        super(DirectoryLineEdit, self).__init__(parent=parent)
```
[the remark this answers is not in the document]
paulgoulain: No problem, I'll do the change.

pymeasure/display/widgets.py
```diff
@@ -1006,3 +1006,28 @@ def eval_string(string, name=None, depth=None):
...
+        browse_action = QtGui.QAction(self)
+        browse_action.setIcon(self.style().standardIcon(getattr(QtGui.QStyle, 'SP_DialogOpenButton')))
+        browse_action.triggered.connect(self.browse_triggered)
+
+        self.addAction(browse_action, QtGui.QLineEdit.TrailingPosition)
+
+    def browse_triggered(self):
+        path = QtGui.QFileDialog.getExistingDirectory(self, 'Directory', '/')
+        if path is not '':
+            self.setText(path)
```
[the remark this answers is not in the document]
paulgoulain: Ok I add a line.

pymeasure/display/windows.py
```diff
@@ -522,6 +522,11 @@ def finished(self, experiment):
             self.abort_button.setEnabled(False)
             self.browser_widget.clear_button.setEnabled(True)
 
+    @property
+    def directory(self):
+        if not self.directory_input:
+            raise Exception("No directory input in the ManagedWindow")
```
bilderbuchi, line 528: Judging purely from the exception message, a `ValueError` seems an appropriate exception type here.

paulgoulain: I have to say, I just copied the exception of the function make procedure. Do I change to this exception ?

bilderbuchi: I see. In principle it's desirable to make the exceptions one throws _as specific as possible_, choosing basically from [this list](https://docs.python.org/3.7/library/exceptions.html?highlight=exception#concrete-exceptions). This has the benefit that whoever consumes your code and deals with these exceptions has an easier time filtering for the correct errors. You can also define your own exception hierarchy in code, but I don't think that's necessary here.
Yes, you would replace `Exception` with `ValueError` (or another more fitting one if you find one in the list, this  is just the one that came to my mind).

paulgoulain: Done.

CasperSchippers (approved):
Thank you very much for this addition! It seems a useful addition and I really like the completer.

Issue #249, opened 2020-02-25
Unit test extension: protocol tests


bilderbuchi:
InstrumentKit has this very neat way of testing if the instrument implementation follows an expected protocol (`expected_protocol`) and, given a defined query-response set, behaves correctly: https://instrumentkit.readthedocs.io/en/latest/devguide/testing.html#expected-protocols
@cjermain I think this could help reduce PR review effort and improve the test suite if such tests can be used to confirm correct instrument behaviour. What do you think? 

@scasagrande is it ok for you if we implement an equivalent functionality here? Do you have any input/pointers what to watch out for or improve?
I'm also hoping if IK and pymeasure gain more API/usage alignment (see also #248) a future interoperability between the projects will be easier to accomplish.

cjermain:
@bilderbuchi, I think this a great idea. I've been wanting to move our `Instruments` to an automated testing suite, but have not had the bandwidth to innovate on it. This would be a great addition on a course toward a stable 1.0 release.

scasagrande:
There's lots of better ways one can test the drivers if I were to do it from scratch. I'd still follow the same general style as what I have, but with the following:

- include a captured copy of input/output in an accompanying structured format file, and automatically test against it if the file is present with no/minimal boilerplate. Sometimes when developing the tests I was making guesses on the behaviour of the instrument based on experimentation or just the documentation. Sometimes these don't capture the full picture or are incorrect. This would ideally use a tool to iterate through the driver, dynamically call all the defined members with a real copy of the instrument connected and record all the traffic.
- rely on `hypothesis` for basic fuzzing coverage to help catch edge cases without having to implement each case manually and individually
- make a companion tool to generate test boilerplate / stubs

bilderbuchi:
Thanks for the insight, those points sound good!

That query-response record would be useful. I think iterating through all available functions is not too complicated, but the devil will be in the detail, for example 
* concerning functions that are not applicable in the present situation, for example because it needs some specific setup, and error handling in that situation (we don't want to record the error response as "known-good" :D)
* how to deal with noisy floating point responses
* how to deal with updating that file meaningfully when others update/regenerate the file because they extend/fix/modify the driver (especially connected to the previous point).

How do you see Hypothesis coming in here? Are you thinking mainly about checking the input/output validation (that's where I would see some benefit of Hypothesis throwing weird values at IK)? Or something running with the instrument connected?

[...]

PR #324, opened 2020-10-02
Dev/agilent34450a

theMashUp:
Adds instrument Agilent34450A Digital Multimeter.


pymeasure/instruments/agilent/agilent34450A.py
```diff
@@ -0,0 +1,571 @@
...
+
+        # Eliminate empty string elements
+        list_without_empty_elements = list(filter(lambda v: v != '', list_of_elements))
+
+        # Convert numbers from str to float, where applicable
+        for i, v in enumerate(list_without_empty_elements):
+            try:
+                list_without_empty_elements[i] = float(v)
+            except ValueError as e:
+                print(e)
```
bilderbuchi, line 565: that should probably use the `log` you define above, instead of `print`

bilderbuchi (changes requested):
I gave this a cursory look, noticed nothing critical (only one minor thing). Did you test (all) the functionality with a real device?

pymeasure/instruments/agilent/agilent34450A.py
theMashUp: Agreed, modified in new commit e0e1a7a.

theMashUp:
All functions were tested with the instrument, I took out the tests with commit 5539936 to not mess up with the normal testing routine when the device is not connected.

bilderbuchi:
Nice work on the large number of tests! I'm hesitant to just throw this away - alternatively, could you mark the tests as skipped instead? Using `pytest.skip('Only work with connected hardware', allow_module_level=True)` at the top of the file should skip all the tests, but would keep them around.
In case we implement a better test suite even for non-connected hardware (see #249), we could then enable/adapt these tests. 
(You'll need to fix the merge conflict in the authors file first so that CI can run correctly)

theMashUp:
Re-added the tests with the skip. I corrected the AUTHORS.txt conflict, but it still shows up somehow, I am not sure what is up with that.

tests/instruments/agilent/test_agilent34450A.py
```diff
@@ -0,0 +1,488 @@
...
+    # TESTS #
+    #########
+
+    def test_dmm_initialization_bad(self):
+        bad_resource = "USB0::10893::45848::MY12345678::0::INSTR"
+        with pytest.raises(VisaIOError):
+            dmm = Agilent34450A(bad_resource)
+
+    def test_good_address(self):
+        dmm = Agilent34450A(self.resource)
```
bilderbuchi, line 89: can you add an assert here, e.g. query a status or `.id` or something to confirm communicaiton with the device actually works, or is this considered successful if it doesn't raise an exception? If the latter, maybe add a comment about this.

tests/instruments/agilent/test_agilent34450A.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def test_reset(self):
+        dmm = Agilent34450A(self.resource)
+        dmm.reset()
```
bilderbuchi, line 93: Is there some query that you can use to assert the successful reset here? A boot cycle counter? Prior to the reset setting a quantity that gets reset by the device `reset()`, then checking if that happened?

bilderbuchi:
Don't worry, I'll sofrt out the authors.txt for you.

bilderbuchi:
Oh, you used f-strings, which are incompatible with python 3.5. I think it is time to remove support for EOL'd python3.5, see #338, so this should be fine for now.

There is a warning in the test log:
```
pymeasure/instruments/agilent/agilent34450A.py:555

  /home/travis/build/ralph-group/pymeasure/pymeasure/instruments/agilent/agilent34450A.py:555: DeprecationWarning: invalid escape sequence \s

    list_of_elements = re.split('["\s,]', one_long_string)
```

Pytest does not show the skipped tests, but that is apparently a [known issue](https://github.com/pytest-dev/pytest/issues/5551).

bilderbuchi:
I removed Python 3.5 support from master, so with your next modifications, do a merge/rebase from pymeasure master and the CI should start to work again. :-)

tests/instruments/agilent/test_agilent34450A.py
```diff
@@ -76,28 +70,35 @@ def make_reseted_dmm(self):
...
+        return self.DMM
+
     #########
     # TESTS #
     #########
 
     def test_dmm_initialization_bad(self):
         bad_resource = "USB0::10893::45848::MY12345678::0::INSTR"
-        with pytest.raises(VisaIOError):
+        # The pure python VISA library (pyvisa-py) raises a ValueError while the PyVISA library raises a VisaIOError.
```
bilderbuchi, line 89: huh, weird, that inconsistency might be worth to report a bug against pyvisa-py...

bilderbuchi:
Thank you for your effort! Merging this now.

PR #325, opened 2020-10-02
Dev/keysightdsox1102g

theMashUp:
Adds Instrument KeysightDSOX1102G


pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,647 @@
...
+    def _ch(self, chNb):
+        """
+        Reads setup data from channel and converts it to a more convenient dict of values.
+        """
+        if chNb == 1:
+            ch_prefix = ":channel1"
+        elif chNb == 2:
+            ch_prefix = ":channel2"
+        else:
+            raise ValueError("Bad channel nb.")
```
bilderbuchi, line 557: Please avoid unnecessarily terse/abbreviated formulations. Same in general for variables like `chNb` above - in Python in my experience, `channel_number` would be totally ok and even preferred because of increased clarity. :-)

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,647 @@
...
+    def autoscale(self):
+        """ Autoscale displayed channels. """
+        self.write(":autoscale")
+
+    #############
+    # Channel 1 #
+    #############
+
+    @property
+    def ch1(self):
```
bilderbuchi, line 72: There seems to be a lot of duplicated code for the two channels. Can you take a look at how another instrument has [implemented channels](https://github.com/ralph-group/pymeasure/blob/07cf309b8d113931525c8cc216cc704f2b7ff413/pymeasure/instruments/tektronix/afg3152c.py#L28) and [their use](https://github.com/ralph-group/pymeasure/blob/07cf309b8d113931525c8cc216cc704f2b7ff413/pymeasure/instruments/tektronix/afg3152c.py#L186-L187) for one pattern how to approach this. Would that be more efficient/elegant in your case, and reduce duplication?

bilderbuchi (changes requested):

theMashUp:
Implemented suggested modifications, re-added the tests (skipped by default). Sorry for the delay, I didn't have the instrument to run the tests by myself these past weeks. New authors are colleagues who contributed to testing and final touches.

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,559 @@
...
+            - "yincrement": voltage difference between data points (float)
+            - "yorigin": voltage at center of screen (float)
+            - "yreference": data point associated with yorigin (int)"""
+        return self._waveform_preamble()
+
+    @property
+    def waveform_data(self):
+        """ Get the binary block of sampled data points transmitted using the IEEE 488.2 arbitrary
+        block data format."""
+        # Other waveform formats raise UnicodeDecodeError
```
bilderbuchi, line 396: Hm, you can try to use [Instrument.binary_values](https://github.com/ralph-group/pymeasure/blob/c0e3ff0b4ac2bce58639e5410da4af4c6147dddb/pymeasure/instruments/instrument.py#L113) to get the reply correctly (it also can strip headers and convert datatype)? (I have to find out why this is not in the documentation)

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,559 @@
...
+    def system_setup(self):
+        """ A string parameter that sets up the oscilloscope. Must be in IEEE 488.2 format.
+        It is recommended to only set a string previously obtained from this command."""
+        return self.ask(":system:setup?")
+
+    @system_setup.setter
+    def system_setup(self, setup_string):
+        self.write(":system:setup " + setup_string)
+
+    def __init__(self, adapter, **kwargs):
```
bilderbuchi, line 418: can you please move the init to the top of the class? It's maybe personal preference, but I find it a little weird to bury this in the middle somewhere. I know others in the codebase are also like this, I should formalize this one day.

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,559 @@
...
+
+    def download_image(self, format_: str = "png", color_palette: str = "color"):
+        """ Get image of oscilloscope screen in bytearray of specified file format.
+
+        :param format_: "bmp", "bmp8bit", or "png"
+        :param color_palette: "color" or "grayscale"
+        """
+        query = f":DISPlay:DATA? {format_}, {color_palette}"
+        # Using binary_values query because default interface does not support binary transfer
+        img = self.adapter.connection.query_binary_values(query, datatype="s")
```
bilderbuchi, line 481: you should be able to use Instrument.binary_values here, I think.

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,559 @@
...
+
+        # Create dict of setup parameters
+        tb_setup = dict(map(lambda v: v.split(" "), tb_setup_splitted))
+
+        # Clean up badly named key
+        key_to_pop = None
+        for key in tb_setup:
+            if "TIM" in key:
+                key_to_pop = key
+        tb_setup["MODE"] = tb_setup.pop(key_to_pop)
```
bilderbuchi, line 523: same logic question as above - this only uses the last key containing "TIM".

tests/instruments/keysight/test_keysightDSOX1102G_expectedSetup.txt
```diff
@@ -0,0 +1,3717 @@
+#800087683<setup prod="InfiniiVision" ver="01.01.2016092800" srver="2.0">
```
bilderbuchi, line 1: Where is this file used in the test? I could not find this filename elsewhere...

bilderbuchi (changes requested):
I went through the PR, it was a lot to digest :D
I left a number of comments/requests/questions.

As you have a testsuite for the instrument, maybe it makes sense for you to practice test-driven development: add tests for issues where this is appropriate, check that they fail, then implement the fix.

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,559 @@
...
+        :param offset: Value represented at center of screen, must be inside the legal range.
+        :param probe_attenuation: Probe attenuation values from 0.1 to 1000.
+        :param vertical_range: Full-scale vertical axis of the selected channel. When using 1:1 probe
+                                attenuation, legal values for the range are  from 8mV to 40 V. If the
+                                probe attenuation is changed, the range value is multiplied by the
+                                probe attenuation factor.
+        :param scale: Units per division. """
+
+        if vertical_range is not None and scale is not None:
+            raise Warning('Both "vertical_range" and "scale" are specified. Specified "scale" has priority.')
```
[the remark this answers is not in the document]
theMashUp: Any thoughts over the use of warnings.warn() vs log.warning()?

Here's the logging documentation (https://docs.python.org/3/howto/logging.html):
_warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning

logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted_

Since this case is more of a "good to know" that a library-breaking issue, I'd favor your suggestion and replace the raise Warning() by log.warning().

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,559 @@
...
+
+        # Add "CHAN" key
+        ch_setup_dict["CHAN"] = self.number
+
+        # Clean up badly named key
+        key_to_pop = None
+        for key in ch_setup_dict:
+            if "RANG" in key:
+                key_to_pop = key
+        ch_setup_dict["RANG"] = ch_setup_dict.pop(key_to_pop)
```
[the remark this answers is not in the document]
theMashUp: The end result is correct, the process is a bit convoluted - clarified in upcoming commit.

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,559 @@
...
+        """ Acquire waveforms according to the settings of the :ACQuire commands and specified source,
+         as a string parameter that can take the following values: "channel1", "channel2", "function",
+         "math", "fft", "abus", or "ext". """,
+        validator=strict_discrete_set,
+        values={"channel1": "CHAN1", "channel2": "CHAN2", "function": "FUNC", "math": "MATH",
+                "fft": "FFT", "abus": "ABUS", "ext": "EXT"},
+        map_values=True
+    )
+
+    def digitize(self, source: str):
```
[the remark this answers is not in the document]
theMashUp: Digitize is not actually a property - it is an operation. I thought it made more sense from a usability point-of-view to call it like a method, with digitize(param), as opposed to using digitize = param. Furthermore, digitize cannot be read (i.e. this would not work: var = digitize).

I define _digitize as Instrument.setting simply to facilitate the verification of correct values.

pymeasure/instruments/keysight/keysightDSOX1102G.py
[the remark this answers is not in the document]
bilderbuchi: Hm, tanks, I was not aware of that distinction. I was more going after usage in this codebase - it always uses log.warning(), never Warning().

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,559 @@
...
+        "fft", "wmemory1", "wmemory2", or "ext".""",
+        validator=strict_discrete_set,
+        values={"channel1": "CHAN1", "channel2": "CHAN2", "function": "FUNC", "fft": "FFT",
+                "wmemory1": "WMEM1", "wmemory2": "WMEM2", "ext": "EXT"},
+        map_values=True
+    )
+    waveform_format = Instrument.control(
+        ":waveform:format?", ":waveform:format %s",
+        """ A string parameter that controls how the data is formatted when sent from the 
+        oscilloscope. Can be "ascii".""",
```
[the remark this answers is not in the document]
theMashUp: 'WORD' and 'BYTE' formats are also available. I do believe I had not included them here because of the issue with querying binary values (as per your other comments).

I'll try to work with instrument.binary_values and might be able to allow 'WORD' and 'BYTE' as waveform types here.

tests/instruments/keysight/test_keysightDSOX1102G_expectedSetup.txt
theMashUp: Good catched - used in previous iteration, not anymore. Will delete it.

pymeasure/instruments/keysight/keysightDSOX1102G.py
```diff
@@ -0,0 +1,551 @@
...
+        scope.autoscale()
+        ch1_data_array, ch1_preamble = scope.download_data(source="channel1", points=2000)
+        # ...
+        scope.shutdown()
+
+     Known issues:
+        - The digitize command will be completed before the operation is. May lead to
+            VI_ERROR_TMO (timeout) occuring when sending commands immediately after digitize.
+            Current fix: if deemed necessary, add delay between digitize and follow-up command
+            to scope.
```
bilderbuchi, line 228: This should fix the doc generation warning reported above.

```suggestion
    .. code-block:: python
    
        scope = KeysightDSOX1102G(resource)
        scope.autoscale()
        ch1_data_array, ch1_preamble = scope.download_data(source="channel1", points=2000)
        # ...
        scope.shutdown()

    Known issues:
    
    - The digitize command will be completed before the operation is. May lead to
      VI_ERROR_TMO (timeout) occuring when sending commands immediately after digitize.
      Current fix: if deemed necessary, add delay between digitize and follow-up command
      to scope.
```

pymeasure/instruments/keysight/keysightDSOX1102G.py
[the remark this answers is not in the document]
bilderbuchi: Fair enough, thanks.

pymeasure/instruments/keysight/keysightDSOX1102G.py
bilderbuchi: Was that a suggestion you considered?

bilderbuchi (changes requested):
I went through the latest changes, this is shaping up nicely, I think we are as good as there.

I am very sorry that I did not notice this earlier, but [we prefer](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#python-style-guides) type hinting in the docstrings, not in the function signature, in this codebase - can you please change this in the couple of places where you used it?

pymeasure/instruments/keysight/keysightDSOX1102G.py
theMashUp: I've decided to keep the format of data requested by this method to "ascii", which means that we do not received binary values, so no need to use binary_values.

theMashUp:
Updated the docstrings per suggestion. The Keysight-specific unit tests are all passing on our side.

bilderbuchi (approved):

bilderbuchi:
I think we are done here. Thank you for your contribution, and your patience during iteration on this PR!

theMashUp:
Thanks Christoph for your great comments and review! Glad to see these changes merged.

Issue #319, opened 2020-09-18
Keithley 2400


jskaggs99:
Can we include high impedance settings in the Keithley 2400 function library? 

```
def set_high_impedance(self):
        self.inst.write(":OUTP:SMOD HIMP")
        print('Set high impedance mode')
```

Something like this does not exist in the documentation currently, and would help out a lot of people. Thank you!

paulgoulain:
Hi, I'm preparing a PR for some functionalities that are currently not implemented in the Keithley 2400 driver. I can add this function.

Paul

jskaggs99:
That's awesome. Thank you so much

paulgoulain:
Looking at the manual, I think I would implement it using the Instrument.control function. It would look like that :

```python
    output_off_state = Instrument.control(
        ":OUTP:SMOD?", ":OUTP:SMOD %s",
        """ Some explanations """,
        validator=strict_discrete_set,
        values=['HIMP', 'NORM', 'ZERO', 'GUAR'],
        map_values=False)
```

Does it look ok for you ?

Paul

jskaggs99:
That looks great Paul

paulgoulain:
Great, I'll add it to my futur PR.

bilderbuchi:
Closed by #326, thank you!

PR #326, opened 2020-10-05
Added new functions for Keithley2400 

paulgoulain:
Added two missing functions for the Keithley2400 driver : 
- Filters
- Output state

### DIFF
--- a/pymeasure/instruments/keithley/keithley2400.py
+++ b/pymeasure/instruments/keithley/keithley2400.py
@@ -303,6 +303,83 @@ class Keithley2400(Instrument, KeithleyBuffer):
         values=[0, 999.9999]
     )
 
+    ###########
+    # Filters #
+    ###########
+
+    current_filter_type = Instrument.control(
+        ":SENS:CURR:AVER:TCON?", ":SENS:CURR:AVER:TCON %s",
+        """ A String property that controls the filter's type for the current.
+        REP : Repeating filter
+        MOV : Moving filter""",
+        validator=strict_discrete_set,
+        values=['REP', 'MOV'],
+        map_values=False)
+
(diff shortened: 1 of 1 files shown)


pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -303,6 +303,83 @@ class Keithley2400(Instrument, KeithleyBuffer):
...
+
+    #####################
+    # Output subsystem #
+    #####################
+
+    current_output_off_state = Instrument.control(
+        ":OUTP:CURR:SMOD?", ":OUTP:CURR:SMOD %s",
+        """ Select the output-off state of the SourceMeter.
+        HIMP : output relay is open, disconnects external circuitry.
+        NORM : V-Source is selected and set to 0V, Compliance is set to 0.5%
```
bilderbuchi, line 359: Is the meaning of the various parameters really identical between current_ and voltage_ versions?

bilderbuchi (changes requested):
Thanks for this, I had a quick look.

pymeasure/instruments/keithley/keithley2400.py
paulgoulain: The documentation is quite shallow on this function. I added it after someone requested it (see issue #319). From what I understand, depending on which output is currently selected, the Keithley will switch to right kind of off state mode. I don't know how to test it though to confirm the behavior.

bilderbuchi (approved):

pymeasure/instruments/keithley/keithley2400.py
bilderbuchi: OK then.

PR #327, opened 2020-10-07
implement the Attocube ANC300 Piezo controller

dkriegner:
this device needs a custom adapter which is derived from the Python standard module telnetlib. Since a general TelnetAdapter could be useful also for other devices I included a general TelnetAdapter from which the specialized Attocube Adapter is derived.

### DIFF
--- a/pymeasure/adapters/telnet.py
+++ b/pymeasure/adapters/telnet.py
@@ -39,7 +39,7 @@ class TelnetAdapter(Adapter):
     :param kwargs: Any valid key-word argument for telnetlib.Telnet
     """
 
-    def __init__(self, host, port, query_delay=0, **kwargs):
+    def __init__(self, host, port=0, query_delay=0, **kwargs):
         self.query_delay = query_delay
         self.connection = telnetlib.Telnet(host, port, **kwargs)
 
--- a/pymeasure/adapters/telnet.py
+++ b/pymeasure/adapters/telnet.py
@@ -46,8 +46,8 @@ class TelnetAdapter(Adapter):
         for kw in kwargs:
             if kw not in safe_keywords:
                 raise TypeError(
-                    "TelnetAdapter: unexpected keyword argument '%s', "
-                    "allowed are: %s" % (kw, str(safe_keywords)))
+                    f"TelnetAdapter: unexpected keyword argument '{kw}', "
+                    f"allowed are: {str(safe_keywords)}")
         self.connection = telnetlib.Telnet(host, port, **kwargs)
 
     def __del__(self):
--- a/pymeasure/adapters/telnet.py
+++ b/pymeasure/adapters/telnet.py
@@ -1,0 +1,80 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2020 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import telnetlib
+import time
+
+from .adapter import Adapter
+
+
+class TelnetAdapter(Adapter):
+    """ Adapter class for using the Python telnetlib package to allow
+    communication to instruments
+
+    :param host: host address of the instrument
+    :param port: TCPIP port
+    :param query_delay: delay in seconds between write and read in the ask
+                        method
+    :param kwargs: Any valid key-word argument for telnetlib.Telnet
+    """
+
+    def __init__(self, host, port, query_delay=0, **kwargs):
+        self.query_delay = query_delay
+        self.connection = telnetlib.Telnet(host, port, **kwargs)
+
+    def __del__(self):
+        """ Ensures the connection is closed upon deletion
+        """
+        self.connection.close()
+
+    def write(self, command):
+        """ Writes a command to the instrument
+
+        :param command: command string to be sent to the instrument
+        """
+        self.connection.write(command.encode())
+
+    def read(self):
+        """ Read something even with blocking the I/O. After something is
+        received check again to obtain a full reply.
+
+        :returns: String ASCII response of the instrument.
+        """
+        return self.connection.read_some().decode() + \
+                self.connection.read_very_eager().decode()
+
+    def ask(self, command):
+        """ Writes a command to the instrument and returns the resulting ASCII
+        response
+
+        :param command: command string to be sent to the instrument
+        :returns: String ASCII response of the instrument
+        """
+        self.write(command)
+        time.sleep(self.query_delay)
+        return self.read()
+
+    def __repr__(self):
+        return "<TelnetAdapter(host=%s, port=%d)>" % (self.connection.host, self.connection.port)
+
--- a/pymeasure/adapters/telnet.py
+++ b/pymeasure/adapters/telnet.py
@@ -46,8 +46,8 @@ class TelnetAdapter(Adapter):
         for kw in kwargs:
             if kw not in safe_keywords:
                 raise TypeError(
-                    f"TelnetAdapter: unexpected keyword argument '{kw}', "
-                    f"allowed are: {str(safe_keywords)}")
+                    "TelnetAdapter: unexpected keyword argument '%s', "
+                    "allowed are: %s" % (kw, str(safe_keywords)))
         self.connection = telnetlib.Telnet(host, port, **kwargs)
 
     def __del__(self):
--- a/pymeasure/adapters/telnet.py
+++ b/pymeasure/adapters/telnet.py
@@ -36,11 +36,18 @@ class TelnetAdapter(Adapter):
     :param port: TCPIP port
     :param query_delay: delay in seconds between write and read in the ask
                         method
-    :param kwargs: Any valid key-word argument for telnetlib.Telnet
+    :param kwargs: Valid keyword arguments for telnetlib.Telnet, currently
+    this is only 'timeout'
     """
 
     def __init__(self, host, port=0, query_delay=0, **kwargs):
         self.query_delay = query_delay
+        safe_keywords = ['timeout']
+        for kw in kwargs:
+            if kw not in safe_keywords:
+                raise TypeError(
+                    f"TelnetAdapter: unexpected keyword argument '{kw}', "
+                    f"allowed are: {str(safe_keywords)}")
         self.connection = telnetlib.Telnet(host, port, **kwargs)
 
     def __del__(self):
--- a/pymeasure/instruments/attocube/adapters.py
+++ b/pymeasure/instruments/attocube/adapters.py
@@ -76,9 +76,11 @@ class AttocubeConsoleAdapter(TelnetAdapter):
         :returns: String ASCII response of the instrument.
         """
         raw = super().read().strip(self.read_termination)
-        lines = raw.split('\n')  # line endings inconsistent '\n', or '\r\n'
-        ret = '\n'.join(line.strip() for line in lines[:-1])
-        self.check_acknowledgement(lines[-1], ret)
+        # one would want to use self.read_termination as 'sep' below, but this
+        # is not possible because of a firmware bug resulting in inconsistent
+        # line endings
+        ret, ack = raw.rsplit(sep='\n', maxsplit=1)
+        self.check_acknowledgement(ack, ret)
         return ret
 
     def write(self, command, checkAck=True):
--- a/pymeasure/instruments/attocube/adapters.py
+++ b/pymeasure/instruments/attocube/adapters.py
@@ -1,0 +1,104 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2020 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import time
+
+from pymeasure.adapters import TelnetAdapter
+
+
+class AttocubeConsoleAdapter(TelnetAdapter):
+    """ Adapter class for connecting to the Attocube Standard Console. This
+    console is a Telnet prompt with password authentication.
+
+    :param host: host address of the instrument
+    :param port: TCPIP port
+    :param passwd: password required to open the connection
+    :param kwargs: Any valid key-word argument for TelnetAdapter
+    """
+    def __init__(self, host, port, passwd, **kwargs):
+        self.terminator = '\r\n'
+        super(AttocubeConsoleAdapter, self).__init__(host, port, **kwargs)
+        # clear the initial messages from the controller
+        time.sleep(self.query_delay)
+        super(AttocubeConsoleAdapter, self).read()
+        # send password and check authorization
+        self.write(passwd, checkAck=False)
+        time.sleep(self.query_delay)
+        ret = super(AttocubeConsoleAdapter, self).read()
+        authmsg = ret.split(self.terminator)[1]
+        if authmsg != 'Authorization success':
+            raise Exception("Attocube authorization failed ('%s')" % authmsg)
+        # switch console echo off
+        self.ask('echo off')
+
+    def check_acknowledgement(self, reply, msg=""):
+        """ Reads the reply of the instrument until the termination sequence.
+        The returned string is check to be 'OK' and otherwise a ValueError is
+        raised.
+
+        :param msg: optional message for the eventual error
+        :returns: String ASCII response of the instrument.
+        """
+        if reply.strip() != 'OK':
+            if msg == "":  # clear buffer
+                msg = reply.strip()
+                super(AttocubeConsoleAdapter, self).read()
+            raise ValueError("AttocubeConsoleAdapter: Error after command '%s'"
+                             " with message '%s'" % (self.lastcommand, msg))
+
+    def read(self):
+        """ Reads a reply of the instrument which consists of two or more
+        lines. The first ones are the reply to the command while the last one
+        is 'OK' or 'ERROR' to indicate any problem. In case the reply is not OK
+        a ValueError is raised.
+
+        :returns: String ASCII response of the instrument.
+        """
+        raw = super(AttocubeConsoleAdapter, self).read().strip()
+        lines = raw.split('\n')  # line endings inconsistent '\n', or '\r\n'
+        ret = '\n'.join(line.strip() for line in lines[:-1])
+        self.check_acknowledgement(lines[-1], ret)
+        return ret
+
+    def write(self, command, checkAck=True):
+        """ Writes a command to the instrument
+
+        :param command: command string to be sent to the instrument
+        """
+        self.lastcommand = command
+        super(AttocubeConsoleAdapter, self).write(command + self.terminator)
+        if checkAck:
+            reply = self.connection.read_until(self.terminator.encode())
+            self.check_acknowledgement(reply.decode())
+
+    def ask(self, command):
+        """ Writes a command to the instrument and returns the resulting ASCII
+        response
+
+        :param command: command string to be sent to the instrument
+        :returns: String ASCII response of the instrument
+        """
+        self.write(command, checkAck=False)
+        time.sleep(self.query_delay)
+        return self.read()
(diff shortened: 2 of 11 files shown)


pymeasure/instruments/attocube/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+    """ Adapter class for connecting to the Attocube Standard Console. This
+    console is a Telnet prompt with password authentication.
+
+    :param host: host address of the instrument
+    :param port: TCPIP port
+    :param passwd: password required to open the connection
+    :param kwargs: Any valid key-word argument for TelnetAdapter
+    """
+    def __init__(self, host, port, passwd, **kwargs):
+        self.terminator = '\r\n'
```
bilderbuchi, line 40: Please use `self.read_termination` and `self.write_termination` to mirror the [pyvisa nomenclature](https://pyvisa.readthedocs.io/en/latest/introduction/resources.html?highlight=query_delay#termination-characters)

pymeasure/instruments/attocube/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+        super(AttocubeConsoleAdapter, self).__init__(host, port, **kwargs)
+        # clear the initial messages from the controller
+        time.sleep(self.query_delay)
+        super(AttocubeConsoleAdapter, self).read()
```
bilderbuchi, line 44: this will block until some global timeout, right? or do you expect a response here?

bilderbuchi, line 44: Can't you just say `super()` here nowadays?

pymeasure/instruments/attocube/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+        # send password and check authorization
+        self.write(passwd, checkAck=False)
+        time.sleep(self.query_delay)
+        ret = super(AttocubeConsoleAdapter, self).read()
+        authmsg = ret.split(self.terminator)[1]
+        if authmsg != 'Authorization success':
+            raise Exception("Attocube authorization failed ('%s')" % authmsg)
+        # switch console echo off
+        self.ask('echo off')
```
bilderbuchi, line 53: why not `write()`, as you don't do anything with the return?

pymeasure/instruments/attocube/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def check_acknowledgement(self, reply, msg=""):
```
bilderbuchi, line 55: This is usually implemented on the instrument level by writing a [`check_errors`](https://pymeasure.readthedocs.io/en/latest/api/instruments/instruments.html#pymeasure.instruments.Instrument.check_errors) method, so I suggest moving it into the ANC300 class.

pymeasure/instruments/attocube/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        if reply.strip() != 'OK':
+            if msg == "":  # clear buffer
+                msg = reply.strip()
+                super(AttocubeConsoleAdapter, self).read()
+            raise ValueError("AttocubeConsoleAdapter: Error after command '%s'"
+                             " with message '%s'" % (self.lastcommand, msg))
+
+    def read(self):
+        """ Reads a reply of the instrument which consists of two or more
```
bilderbuchi, line 71: not sure if this would not be better handled by using `check_errors=True` in the property definitions?

pymeasure/instruments/attocube/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+        lines. The first ones are the reply to the command while the last one
+        is 'OK' or 'ERROR' to indicate any problem. In case the reply is not OK
+        a ValueError is raised.
+
+        :returns: String ASCII response of the instrument.
+        """
+        raw = super(AttocubeConsoleAdapter, self).read().strip()
```
bilderbuchi, line 78: strip should use `self.read_termination` as proposed above

pymeasure/instruments/attocube/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+        return ret
+
+    def write(self, command, checkAck=True):
+        """ Writes a command to the instrument
+
+        :param command: command string to be sent to the instrument
+        """
+        self.lastcommand = command
+        super(AttocubeConsoleAdapter, self).write(command + self.terminator)
+        if checkAck:
```
bilderbuchi, line 91: watch out here with API structure, your `write` now effectively is an `ask` (write-read cycle), which feels weird/inconsistent with the other adapters.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -0,0 +1,227 @@
...
+
+    offset_voltage = Instrument.control(
+            "geta", "seta %.3f",
+            """ Offset voltage in Volts from 0 to 150 V.
+            This property can be set. """,
+            validator=strict_range, values=[0, 150],
+            get_process=extract_float)
+
+    output_voltage = Instrument.measurement("geto",
+                                            """ Output voltage in volts.""",
```
bilderbuchi, line 102: maybe keep indentation style consistent with above (also below).

pymeasure/instruments/attocube/anc300.py
```diff
@@ -0,0 +1,227 @@
...
+        completion of the movement
+        """
+        self.mode = 'stp'
+        # perform the movement
+        if steps > 0:
+            self.stepu = steps
+        elif steps < 0:
+            self.stepd = abs(steps)
+        else:
+            pass  # do not set stepu/d to 0 since it triggers a continous move
```
bilderbuchi, line 168: What happens if you say `move(0)`? would that move ever finish below?

pymeasure/instruments/attocube/anc300.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def measure_capacity(self):
+        """ Obtains a new measurement of the capacity. The mode of the axis
+        returns to 'gnd' after the measurement.
+
+        :returns capacity: the freshly measured capacity in nF.
+        """
+        self.mode = 'cap'
+        # wait for the measurement to finish
+        self.ask('capw')
```
bilderbuchi, line 182: do you want to return the mode to the previously set one after finishing?

pymeasure/instruments/attocube/anc300.py
```diff
@@ -0,0 +1,227 @@
...
+            positive.""",
+            validator=strict_range, values=[0, inf])
+
+    stepd = Instrument.setting(
+            "stepd %d",
+            """ Step downwards for N steps. Mode must be 'stp' and N must be
+            positive.""",
+            validator=strict_range, values=[0, inf])
+
+    def __init__(self, axis, controller):
```
bilderbuchi, line 122: please put the "parent/instrument/controller" first, then the name of the axis.

bilderbuchi (changes requested):
I went through this, looks promising already with some changes.

bilderbuchi:
@cjermain as this implements an additional adapter, could you take a look through this PR?

pymeasure/adapters/telnet.py
```diff
@@ -0,0 +1,80 @@
...
+    communication to instruments
+
+    :param host: host address of the instrument
+    :param port: TCPIP port
+    :param query_delay: delay in seconds between write and read in the ask
+                        method
+    :param kwargs: Any valid key-word argument for telnetlib.Telnet
+    """
+
+    def __init__(self, host, port, query_delay=0, **kwargs):
```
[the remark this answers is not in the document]
dkriegner: agreed, will make according change

pymeasure/instruments/attocube/adapters.py
dkriegner: will make according change

pymeasure/instruments/attocube/adapters.py
dkriegner: Actually TelnetAdapter.read() is non-blocking. This is because typically on a telnet connection one expects a prompt to be sent to indicate that the console is able to receive further commands. So reading until a termination character sequence is received is not so trivial. So this call here is basically to flush the input buffer, which as far as I know has no dedicated command for telnetlib. if you insist on a blocking read call I would need to redesign a bit?

pymeasure/instruments/attocube/adapters.py
dkriegner: Ok, I missed this option of properties. will convert to check_errors.

pymeasure/instruments/attocube/adapters.py
dkriegner: Ok, here I see a limit of check_errors: It is not clear to the read method how many lines the reply is. And its the last line which contains the 'OK' or 'ERROR' which would need to be checked by the check_errors function. So my approach was to read everything from the interface, check the last line for the error and declare the rest as the reply. Can this be mimicked with check_error?

Also every set command returns 'OK' or 'ERROR' which in this case could of course be easily checked by check_errors. But when the read method can not I guess it would be better to keep this in the adapter. Please advise!

pymeasure/instruments/attocube/adapters.py
dkriegner: The device responds to every command (also a set command) by replying 'OK' or 'ERROR'. Thats why in the write I also read back the reply of the instrument. I understand this could be handled by the check_errors method discussed above. And this is the reason why ask method uses `checkAck = False`. For a command with a reply value the 'OK' is received after the return value. The return can, however, consist of multiple lines which makes the read complicated since one does not know how far to read to not read the 'OK' which would need to be again handled by  `check_error`. advise is welcome

pymeasure/instruments/attocube/anc300.py
dkriegner: agreed

pymeasure/instruments/attocube/anc300.py
dkriegner: it does "finish", because 'stepw' waits for stepping to finish but will return immediately if no stepping is in progress.

pymeasure/instruments/attocube/anc300.py
dkriegner: The controller returns the mode automatically after a single measurement

pymeasure/instruments/attocube/anc300.py
dkriegner: ok

pymeasure/instruments/attocube/adapters.py
dkriegner: I now used the super() style calls (see 30f9678). This could likely be done across all your repository.

bilderbuchi: No, I don't insist on anything as I am not familiar with telnet. Are you sure that TelnetAdapter.read() is non-blocking? I ask because the docstring says otherwise, and as the first step it uses telnet.read_some()

bilderbuchi: Thanks. Well, this repo has some history as you can imagine :D, and nobody got around to harmonizing this, as it's not that important and we have enough work with the existing prs and issues. props for keeping with existing style, though!

pymeasure/instruments/attocube/anc300.py
```diff
@@ -0,0 +1,228 @@
...
+           "ver", """ Version number and instrument identification """
+           )
+
+    controllerBoardVersion = Instrument.measurement(
+           "getcser", """ Serial number of the controller board """
+           )
+
+    def __init__(self, host, axisnames, passwd, **kwargs):
+        if 'query_delay' not in kwargs:
+            kwargs['query_delay'] = 0.05
```
bilderbuchi, line 206: you could also have that as a kwarg with default argument in __init__ that then gets passed on to AttocubeConsoleAdapter. would you think that clearer (as the default value ends up in the init signature)?

bilderbuchi:
Since your device returns strings that don't fit into the simple pattern that Adapter.values can deal with, I don't see a much more elegant way than what you are already doing - this part could probably use some improvement. One alternative way to handle this would be to use a modified `extract_value` as the `cast` callable and make it return a float - but this only replaces your use of `get_process=` with `cast=`, so it's not simpler per se.

If one wanted to apply the same processing/parsing to all gets of an Instrument without specifying this for every control/measurement individually, I could see optionally implementing some generic method at the Adapter level. If that is defined (e.g. via kwarg to the Instrument constructor that gets passed to the Adapter), it gets called within Adapter.values right after receiving the string. This can be used to prepare the reply for the usual separator/cast application, via regex or whatever you want to use, and you only have to define/use this once for the whole Instrument. Of course this depends on the instrument's replies all having the same structure.
I think you should leave what you did here in place, and we take this discussion into a separate issue/feature request - that needs some thinking and will just get buried here. Can you open an issue, please?

pymeasure/instruments/attocube/adapters.py
dkriegner: you are of course right that its blocking, sorry for that (I changed that at some point). But this call here is still not blocking because upon opening a connection the device sends some welcome message and asks for the authentication code.

bilderbuchi: I see. maybe add a comment explaining this additional context?

pymeasure/instruments/attocube/adapters.py
bilderbuchi: Do you want to clarify the intent here by saying `_ = self.ask('echo off')` instead?

pymeasure/instruments/attocube/anc300.py
dkriegner: I guess your proposed way is slightly more elegant. will change

dkriegner:
I will report an issue with my next pull request.

I think most comments are resolved in this one. I have one last question: One option not discussed here (or I misunderstood something) was to totally integrate AttocubeConsoleAdapter into the ANC300Controller. So basically this means it would be directly be based on TelnetAdapter and handle all the device communication protocol directly in the instrument. I guess this largely would mean to copy paste the methods from the Adapter into the Instrument. If you prefer this I can do it. Functionally I guess it will almost remain the same. The small advantage would be that a user has easier access to the raw communication by just calling Instrument.adapter.read/write. Which now is a bit masked by the AttocubeConsoleAdapter

pymeasure/instruments/attocube/adapters.py
```diff
@@ -0,0 +1,108 @@
...
+        """ Reads a reply of the instrument which consists of two or more
+        lines. The first ones are the reply to the command while the last one
+        is 'OK' or 'ERROR' to indicate any problem. In case the reply is not OK
+        a ValueError is raised.
+
+        :returns: String ASCII response of the instrument.
+        """
+        raw = super().read().strip(self.read_termination)
+        lines = raw.split('\n')  # line endings inconsistent '\n', or '\r\n'
+        ret = '\n'.join(line.strip() for line in lines[:-1])
```
bilderbuchi, line 80: These last two lines can actually be done more elegantly (as you put it) with `rsplit`:
```
ack, ret = raw.rsplit(sep='\n', maxsplit=1)
```
if I'm not deceived.

bilderbuchi:
If the adapter is the same/very similar for most Attocube console uses, then by all means keep it separate as it is now.

pymeasure/instruments/attocube/adapters.py
dkriegner: done

pymeasure/instruments/attocube/adapters.py
dkriegner: true, except return value needs to be reversed. I changed it, but left there a verbose comment why sep can not be self.read_termination

dkriegner:
Based on the discussion here I propose to leave the basic structure of the code as is. If there are further suggestions please let me know. From my side this is ready to be merged. 

I will report the issue with the extract_* functions

pymeasure/instruments/attocube/adapters.py
bilderbuchi: > the return can, however, consist of multiple lines which makes the read complicated since one does not know how far to read to not read the 'OK' which would need to be again handled by check_error. advise is welcome

OK in that case check_error would indeed be weird to use.

bilderbuchi:
Don't worry about it, I just made #331.

bilderbuchi:
Yeah, this already looks good! As this adds a new adapter, I would like to wait for @cjermain and @CasperSchippers input if this fits in well. 
I'd also like another opinion if there is a better way to solve the unusual behaviour of the write() in the AttocubeConsoleAdapter.

bilderbuchi (approved):

pymeasure/instruments/attocube/adapters.py
```diff
@@ -0,0 +1,109 @@
...
+        lines. The first ones are the reply to the command while the last one
+        is 'OK' or 'ERROR' to indicate any problem. In case the reply is not OK
+        a ValueError is raised.
+
+        :returns: String ASCII response of the instrument.
+        """
+        raw = super().read().strip(self.read_termination)
+        # one would want to use self.read_termination as 'sep' below, but this
+        # is not possible because of a firmware bug resulting in inconsistent
+        # line endings
```
CasperSchippers, line 80: Do you mean that using '\n' as sep results in one of the strings (i think ret) will often have a carriage return ('\r')? In that case you could consider adding a `ret = ret.strip()` to remove this for a more consistent output.

pymeasure/instruments/attocube/adapters.py
CasperSchippers: Maybe you could mimic this kind of error checking by (temporarily) storing the last response in the adapter such that it can be checked by the check_errors function? (And it would be good to remove the temporarily stored response is cleared after the error checking).

If this is, however, an adapter-wide behavior for all commands, i'm inclined to say that it would be best to keep it in the adapter.

CasperSchippers (commented):
Nice addition! Could be very useful indeed to have the TelnetAdapter!

pymeasure/adapters/__init__.py
```diff
@@ -43,3 +43,5 @@
     from pymeasure.adapters.vxi11 import VXI11Adapter
 except ImportError:
     log.warning("VXI-11 library could not be loaded")
+
+from pymeasure.adapters.telnet import TelnetAdapter
```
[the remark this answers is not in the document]
dkriegner: telnetlib is in the python standard library thats why i left out the try and except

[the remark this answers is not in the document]
CasperSchippers: Ah, yeah, that's a good point.

pymeasure/adapters/telnet.py
```diff
@@ -0,0 +1,80 @@
...
+    :param host: host address of the instrument
+    :param port: TCPIP port
+    :param query_delay: delay in seconds between write and read in the ask
+                        method
+    :param kwargs: Any valid key-word argument for telnetlib.Telnet
+    """
+
+    def __init__(self, host, port=0, query_delay=0, **kwargs):
+        self.query_delay = query_delay
+        self.connection = telnetlib.Telnet(host, port, **kwargs)
```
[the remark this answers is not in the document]
dkriegner: I can do that, but do consider this more a limitation than a feature. I guess this is not the right place to criticise VISAAdapter, but many more keyword arguments than in your safelist are possible depending on the connection type of VISA. One such limitation I encountered recently was when using a serial device where one can not specify the `baud_rate` as keyword argument. But adding `baud_rate` to the safe keyword list would be likely problematic since not every connection type will allow such a keyword... So in the end only `get_instrument` knows which keywords are actually allowed

For the case of [`Telnet`](https://docs.python.org/3.9/library/telnetlib.html#telnetlib.Telnet) in fact only one additional parameter (`timeout`) seems allowed.

pymeasure/instruments/attocube/adapters.py
dkriegner: exactly, will change that. The extra '\r' is not harmful but better to remove it

dkriegner:
I was thinking about buffering the last response for the check_error function but the code is not getting easier. The problem is that the buffering can in my opinion not only be limited to the `read` function. Also after a `write` (in case of a set command) the device replies with the acknowledgement. If the `write` function will not read it back a later `ask` query will have some stray responses in the input buffer. Or upon every `write` the input buffer would need to be flushed, but this also seems like a dirty hack. The `write` used in the `ask` method, however, can not read the reply because the acknowledgement comes back after the return value (see below).

Here example communication with the instrument for illustration of the situation (lines starting in '>' are sent to the instrument, lines starting in '<' are received:

  \> 'setv 1 33.000\r\n' # here one would typically use a write, but still get a response
  < 'OK\r\n'

or a query of a value
  \> 'getv 1\r\n'
  < 'voltage = 33.000000 V\r\nOK\r\n'

a query with a multiline reply:
  \> 'ver\r\n'
  < 'attocube ANC300 controller version 1.2.2-3 2019-01-25 11:02\r\nANC150 compatibillity console\r\nOK\r\n'

If I mimick the behavior with check_error the buffering will once have to be done in write (for example for check_set_errors=True of a `control` or `setting` property and once in read when check_get_errors=True.
If I missed something please let me know.

cjermain (approved):
This is a great addition @dkriegner! Thanks also to @bilderbuchi and @CasperSchippers for the good review feedback. I only have a few minor comments. I think the `check_acknowledgement` function is useful as implemented. I think the discussion around error and acknowledgement checking is very useful to have, and can be addressed as a broader look across the codebase outside of this work.

pymeasure/instruments/attocube/adapters.py
```diff
@@ -22,10 +22,48 @@
 # THE SOFTWARE.
 #
 
+import re
 import time
 
 from pymeasure.adapters import TelnetAdapter
 
+# compiled regular expression for finding numerical values in reply strings
+reg_value = re.compile("\w+\s+=\s+([-+]?[0-9]*\.?[0-9]+)")
```
bilderbuchi, line 31: You should use raw strings, prefixed with `r`, when feeding regular expressions to avoid intepreting the backslashes - `r'blabla'`

bilderbuchi: Also, maybe make this variable _reg_value to signal that it should not be used by others.

pymeasure/adapters/telnet.py
```diff
@@ -46,8 +46,8 @@ def __init__(self, host, port=0, query_delay=0, **kwargs):
         for kw in kwargs:
             if kw not in safe_keywords:
                 raise TypeError(
-                    f"TelnetAdapter: unexpected keyword argument '{kw}', "
```
bilderbuchi: If you want you can bring the f-strings back, we just removed Python3.5 support yesterday.

dkriegner:
thanks for all the constructive comments. I hope I have addressed all of them in the latest commits. 
Sorry for the late adding of one new property (an array property) which also made it necessary to implement some new validators. At least I do not know how to treat this otherwise. Feedback is welcome and I am happy to make more changes if necessary.

@cjermain : thanks for agreeing that `check_ack` as implemented here is ok. I actually with a different device with similar communication protocol tried to use the `check_errors` approach but found it more cumbersome than putting it into the Adapter where it seems to belong 

I will definitely introduce some simplifications in this instrument once issue #331 is fixed. But propose to accept it as is (or after additional fixes) and I would open a new pull request when the time comes.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -29,7 +29,45 @@
...
+    return value
+
+
+def truncated_int_array(value, values):
+    """ validator function to check if the supplied value is an iterable-object
+    with appropriate values.
+    """
+    ret = []
+    for i, v in enumerate(value):
+        if v < values[0]:
```
bilderbuchi, line 50: Do you consider a value of -1.7 as OK? I.e. is the integer-ness only strict within the correct range?

pymeasure/instruments/attocube/anc300.py
```diff
@@ -29,7 +29,45 @@
...
+            value, values))
+    return value
+
+
+def truncated_int_array(value, values):
+    """ validator function to check if the supplied value is an iterable-object
+    with appropriate values.
+    """
+    ret = []
+    for i, v in enumerate(value):
```
bilderbuchi, line 49: is this check going to happen often/frequently? If yes, it might be good to check if it would not run faster to have the "happy case" return early, a la
```
if lower <= v <= upper:
    if float(v).is_integer
       return v
    else:
       raise...
else:
    return max(min(upper, v), lower)

pymeasure/instruments/attocube/anc300.py
dkriegner: The device needs integer values from 0 to 255 (I guess one byte). I will change the logic to raise an error for out of range floats. good catch!

pymeasure/instruments/attocube/anc300.py
dkriegner: This will happen very seldom. Its a basic configuration setting which most people will never touch. I will quickly check the run-time and make according changes

pymeasure/instruments/attocube/anc300.py
bilderbuchi: yeah, your code already did that, it would clamp -1.7 to 0. 
The question was more, if a user supplied -1.7, do you consider this a "harmless" out of range (as you are ok with clamping to 0-255), or consider this a "grave" user error that should bail instead of clamping.

bilderbuchi:
Thank you for your contribution and your patience with our comments. I think this is ready to merge now. :partying_face:

Issue #168, opened 2018-11-02
ZMQ error


jmittelstaedt:
Hi,

While running PyMeasure on a newer version of anaconda, I got an error which looks like it's coming from an update ZMQ had. In particular, the error occurs when a `worker` tries to use its `emit` method. The error is in the `self.publisher.send_serialized` function, which I'm guessing has changed in a newer version of ZMQ or something. A fix for now is to add a `TypeError` to the following `except` statement, but some functionality (which it seems I don't use) is probably being lost along the way. The full error message is at the bottom of this post, and the versions of everything that I'm using are below.

ZMQ version: 17.1.2
Python version: 3.6.6
Conda version: 4.5.11

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "C:\Users\Joseph\Anaconda3\lib\site-packages\zmq\sugar\socket.py", line 427, in send_multipart
    memoryview(msg)
TypeError: memoryview: a bytes-like object is required, not 'int'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Joseph\Anaconda3\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\Users\Joseph\Anaconda3\lib\site-packages\pymeasure-0.5.1-py3.6.egg\pymeasure\experiment\workers.py", line 161, in run
    self.update_status(Procedure.RUNNING)
  File "C:\Users\Joseph\Anaconda3\lib\site-packages\pymeasure-0.5.1-py3.6.egg\pymeasure\experiment\workers.py", line 115, in update_status
    self.emit('status', status)
  File "C:\Users\Joseph\Anaconda3\lib\site-packages\pymeasure-0.5.1-py3.6.egg\pymeasure\experiment\workers.py", line 95, in emit
    self.publisher.send_serialized((topic, record), serialize=cloudpickle.dumps)
  File "C:\Users\Joseph\Anaconda3\lib\site-packages\zmq\sugar\socket.py", line 512, in send_serialized
    return self.send_multipart(frames, flags=flags, copy=copy, **kwargs)
  File "C:\Users\Joseph\Anaconda3\lib\site-packages\zmq\sugar\socket.py", line 434, in send_multipart
    i, rmsg,
TypeError: Frame 0 (128) does not support the buffer interface.
```

lawa42:
Same problem here

zmq: 18.0.1
zeromq: 4.3.1
python: 3.7.3
[...]

PR #329, opened 2020-10-09
Add time-estimator widget

CasperSchippers:
A PR to add a widget to the ManagedWindow to allow displaying of upfront estimates of a measurement. The widget requires the user to implement a get_time_estimate method that can return a list of estimates.

- [x] Auto-add widget if get_time_estimates function is found (or implemented) in the Procedure class
- [x] Solve todo's in the code
- [x] Write documentation & tutorial
- [x] Implement option to return a duration in seconds


CasperSchippers:
I think it is largely finished, could (one of) you have a look and let me know what you think, @cjermain @bilderbuchi.

bilderbuchi:
It will take quite some time, I hardly find the time to chip away on #168/#354, which I am trying to focus on since I suspect some other issues could also be triggered by this. I haven't made any headway yet, multi-threaded debugging is unfun, and I seemingly have to fight to get pymeasure to spit out enough debug logging. :-/

CasperSchippers:
No of course, please don't hurry with this; it is definitely not a priority.

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,66 @@ This adds the input line above the Queue and Abort buttons.
 
 A completer is implemented allowing to quickly select an existing folder, and a button on the right side of the input widget opens a browse dialog.
 
+Using the estimator widget
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to provide estimates of the measurement procedure, an EstimatorWidget is provided that allows the user to define and calculate estimates.
+The widget is automatically activated when the :code:`get_estimates` method is added in the :code:`Procedure`.
```
msmttchr, line 345: I would replace EstimatorWidget with ```:class:\`~pymeasure.display.widgets.EstimatorWidget` ``` to have a cross reference to the class documentation.
Incidentally, I have discovered that none of the classes derived from Qt stuff get document by Sphinx. Is this is a know issue ?

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,66 @@ This adds the input line above the Queue and Abort buttons.
...
+
+The quickest and most simple implementation of the :code:`get_estimates` function simple returns a the estimated duration of the measurement in seconds (as an :code:`int` or a :code:`float`).
+As an example, in the example provided in the ManagedWindow section, the :code:`Procedure` is changed to:
+.. code-block:: python
```
msmttchr, line 349: An empty line is needed to have the code-block statement understood

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,66 @@ This adds the input line above the Queue and Abort buttons.
...
+               ("Number of lines", "%d" % int(self.iterations)),
+               ("Sequence length", str(sequence_length)),
+               ('Measurement finished at', str(datetime.now() + timedelta(seconds=duration))),
+           ]
+
+           return estimates
+
+
+This will add the estimator widget at the dock on the left.
+.. image:: pymeasure-estimator.png
```
msmttchr, line 389: Empty line is needed before ..image

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,66 @@ This adds the input line above the Queue and Abort buttons.
...
+
+This will add the estimator widget at the dock on the left.
+The duration and finishing-time of a single measurement is always displayed in this case.
+Depending on whether the SequencerWidget is also used, the length, duration and finishing-time of the full sequence is also shown.
+
+For maximum flexibility (e.g. for showing multiple and other types of estimates, such as the duration, filesize, finishing-time, etc.) it is also possible that the :code:`get_esimates` returns a list of tuples.
+Each of these tuple consists of two strings: the first is the name (label) of the estimate, the second is the estimate itself.
+
+As an example, in the example provided in the ManagedWindow section, the :code:`Procedure` is changed to:
+
```
msmttchr, line 367: Is it possible to change ManagedWindow such that you have a cross reference link to the mentioned section ?

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,66 @@ This adds the input line above the Queue and Abort buttons.
...
 A completer is implemented allowing to quickly select an existing folder, and a button on the right side of the input widget opens a browse dialog.
 
+Using the estimator widget
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to provide estimates of the measurement procedure, an EstimatorWidget is provided that allows the user to define and calculate estimates.
+The widget is automatically activated when the :code:`get_estimates` method is added in the :code:`Procedure`.
+
+The quickest and most simple implementation of the :code:`get_estimates` function simple returns a the estimated duration of the measurement in seconds (as an :code:`int` or a :code:`float`).
+As an example, in the example provided in the ManagedWindow section, the :code:`Procedure` is changed to:
```
msmttchr, line 348: "function simple returns a the estimated duration", possibly there is a typo here. Did you mean "function simply returns the estimated duration" ?

pymeasure/display/widgets.py
```diff
@@ -1036,3 +1039,218 @@ def browse_triggered(self):
...
+        self.update_thread = EstimatorThread(self.get_estimates)
+        self.update_thread.new_estimates.connect(self.display_estimates)
+
+        self._setup_ui()
+        self._layout()
+
+        self.update_estimates()
+
+        self.update_box.setCheckState(1)
+
```
msmttchr, line 1101: `PySide2` does not like this :(, could You please replace with: `self.update_box.setCheckState(QtCore.Qt.CheckState.Checked)`

examples/Basic/gui_estimator.py
```diff
@@ -0,0 +1,135 @@
...
+            displays=['iterations', 'delay', 'seed'],
+            x_axis='Iteration',
+            y_axis='Random Number',
+            sequencer=True,
+            sequence_file="gui_sequencer_example_sequence.txt"
+        )
+        self.setWindowTitle('GUI Example')
+
+    def queue(self, procedure=None):
+        filename = tempfile.mktemp()
```
msmttchr, line 120: `def queue(self, procedure=None):` should be replaced with `def queue(self, *, procedure=None):` as in the `gui_sequencer.py` example, otherwise when clicking to "Queue" button it will fail. Reason for this is that the method `queue` is always called with a single parameter with value `False` when button "Queue" is clicked (`checked` parameter in the `QAbstractButton` `clicked` signal) and with a keyword parameter `procedure` when "Queue sequence" button is clicked. Possibly this mechanism needs to be updated in the future as it is error-prone.

docs/tutorial/graphical.rst
CasperSchippers: Indeed, thanks

docs/tutorial/graphical.rst
CasperSchippers: Indeed, thanks

docs/tutorial/graphical.rst
CasperSchippers: Yes, indeed that's what i meant

docs/tutorial/graphical.rst
CasperSchippers: Thanks, good idea.

I'll have a look at the absence of these classes in the documentation; I didn't notice this before at any rate.

pymeasure/display/widgets.py
CasperSchippers: Fixed it; Also changed the comparison of the `self.update_box.checkState()` with integers to a comparison with these enumerators, doesn't seem to be required but is more readable.

examples/Basic/gui_estimator.py
CasperSchippers: Ahh yeah, that's a good point indeed

msmttchr: I meant that the `queue` method is defined in the code and documentation without parameters, while it is called with a positional parameter set to `False` when activated by "Queue" button click and with a `procedure` keyword parameter when called by a sequencer.
So, if you initially write `queue` without parameters and later you add a sequencer, code will not work. On the other hand, if you add the `procedure` parameter, then you have to use the '*' syntax to avoid the positional parameter.  A bit weird, in my opinion.

CasperSchippers: Ah I see what you mean. A possible would be to have a very simple wrapper which gets called by the queue button, which simply calls the queue method without any positional arguments (i.e. the False) to fix that issue.
For the sequencer, the only thing I could think of at this moment is to insert an inspection of the queue method upon initializing the sequencer and raise an error if the method doesn't accept the keyword argument. We could also consider adding the procedure keyword argument to the default implementation; in the basic gui example is also possible, though I'm a bit reluctant to add it there as well because I think it'd be best to keep that example as basic as possible.

What do you think? Or do m you have any other suggestions?

msmttchr: Using the simple wrapper to discard the unwanted parameter is, in my opinion, the best approach.

bilderbuchi:
I'll have to pass on reviewing this, I'll let you two handle this, you're in a much better position anyway. :-)

docs/tutorial/graphical.rst
CasperSchippers: Yes it is; fixed it.

examples/Basic/gui_estimator.py
CasperSchippers: Implemented both the wrapper and the signature check

docs/tutorial/graphical.rst
CasperSchippers: Hmmm, this is a bit weird; the docs pages of the last stable version does in fact show these classes, but the latest version doesn't. However, when I compile the docs on my own PC (both in the master branch and this branch), the classes are also there.

msmttchr: I wonder whether this is linked to the changes to support PySide2.

CasperSchippers: When I compile the docs it works in an environment with PyQT as well as in an environment with PySide2

msmttchr: I think this is related to the version of Sphinx:
- v3.2.1 (Windows): problem present
- v3.5.2 (Windows): problem not present

In our CI file, we use v3.2.1, so that's why the problem is present in readthedocs latest. On the hand, it is not clear why stable version is OK.

Do you know who is looking after the CI part of the project ? Is @bilderbuchi ?

CasperSchippers: Ahh yes it indeed is: I pinned it down to a bugfix in v3.4.2 of Sphinx:
```
#8164: autodoc: Classes that inherit mocked class are not documented
```

@bilderbuchi and @cjermain made the CI files; I'm inclined to think that we can safely update the Sphinx version; the latest version is 4.0.2, so I'd suggest to update it to that version.

pymeasure/display/windows.py
```diff
@@ -172,6 +173,19 @@ def __init__(self, procedure_class, inputs=(), displays=(), x_axis=None, y_axis=
...
+            proc.get_estimates()
+        except NotImplementedError:
+            self.use_estimator = False
+        except TypeError:
+            # Raised if arguments are asked by the get_estimates method.
+            self.use_estimator = True
+        else:
+            self.use_estimator = True
+
         self._setup_ui()
```
msmttchr, line 189: An alternative implementation could be:
```
        self.use_estimator = True
        try:
            proc.get_estimates()
        except NotImplementedError:
            self.use_estimator = False
```
It is more compact, but I am not sure whether it is better.

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,68 @@ This adds the input line above the Queue and Abort buttons.
...
+       def get_estimates(self, sequence_length=None, sequence=None):
+
+           return self.iterations * self.delay
+
+This will add the estimator widget at the dock on the left.
+The duration and finishing-time of a single measurement is always displayed in this case.
+Depending on whether the SequencerWidget is also used, the length, duration and finishing-time of the full sequence is also shown.
+
+For maximum flexibility (e.g. for showing multiple and other types of estimates, such as the duration, filesize, finishing-time, etc.) it is also possible that the :code:`get_esimates` returns a list of tuples.
+Each of these tuple consists of two strings: the first is the name (label) of the estimate, the second is the estimate itself.
```
msmttchr, line 365: Small typo get_esimates instead get_estimates

docs/tutorial/graphical.rst
CasperSchippers: I've opened a PR to fix this issue (#460)

docs/tutorial/graphical.rst
CasperSchippers: Thanks, fixed

pymeasure/display/windows.py
CasperSchippers: That should also work (though the TypeError should then also be caught with as pass), but I now have another approach by comparing the `get_estimates` methods:

> self.use_estimator = not self.procedure_class.get_estimates == Procedure.get_estimates

This way, there is no need to really instantiate the procedure class

msmttchr: Much better than my proposal 😉

msmttchr (approved):

examples/Basic/gui_estimator.py
```diff
@@ -0,0 +1,135 @@
...
+            if self.should_stop():
+                log.warning("Catch stop command in procedure")
+                break
+
+    def get_estimates(self, sequence_length=None, sequence=None):
+        """ Function that returns estimates for the EstimatorWidget. If this function
+        is implemented (and does not return a NotImplementedError) the widget is
+        automatically activated.
+
+        The function is expected to an int or float, or a list of tuples. If an int or
```
CasperSchippers, line 63: missing "return"

examples/Basic/gui_estimator.py
```diff
@@ -0,0 +1,135 @@
...
+        float is returned, it should represent the duration in seconds.If a list of
+        tuples is returned, each tuple containing two strings, a label and the estimate
+        itself:
+        estimates = [
+            ("label 1", "estimate 1"),
+            ("label 2", "estimate 2"),
+        ]
+        The length of the number of estimates is not limited but has to remain unchanged after
+        initialisation. Not that also the label can be altered after initialisation.
```
CasperSchippers, line 72: typo

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,68 @@ This adds the input line above the Queue and Abort buttons.
...
+
+Note that after the initialisation of the widget both the label of the estimate as of course the estimate itself can be modified, but the amount of estimates is fixed.
+
+The keyword arguments are not required in the implementation of the function, but are passed if asked for (i.e. :code:`def get_estimates(self)` does also works).
+Keyword arguments that are accepted are :code:`sequence`, which contains the full sequence of the sequencer (if present), and :code:`sequence_length`, which gives the length of the sequence as integer (if present).
+If the sequencer is not present or the sequence cannot be parsed, both :code:`sequence` and :code:`sequence_length` will contain :code:`None`.
+
+The estimates are automatically updated every 2 seconds.
+Changing this update interval is possible using the "Update continuously"-checkbox, which can be toggled between three states: off (i.e. no updating), auto-update every two seconds (default) or auto-update every 100 milliseconds.
+Manually updating the estimates (useful whenver continuous updating is turned of) is also possible using the "update"-button.
```
CasperSchippers, line 402: typo in 'of'

CasperSchippers:
Thanks @msmttchr for the reviewing!

PR #330, opened 2020-10-11
Fix for use these values

CasperSchippers:
### Work in progress

When using the option "Use these values" on an entry in the BrowserWidget, the default values for an input are inserted into the input, rather than the actual values used in that entry.

- [x] Fix the general behavior
- [ ] check if all types of inputs work as intended (at this moment a string seems not correctly escaped whenever it is filled back into the input widget)

### DIFF
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -55,7 +55,9 @@ class Input(object):
         """
         self._parameter = parameter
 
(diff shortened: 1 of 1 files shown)


CasperSchippers:
PR #358 fixes this same issue in a shorter way. Therefore this PR is closed in favor of #358.

PR #336, opened 2020-10-18
Add widget for controlling and reading instruments

CasperSchippers:
### Work in progress

An addition that allows for easily adding widgets that can display values of instruments and control its settings from within the software.

- [ ] Finish basic working of widget
- [ ] Add possibility for custom instrument-widgets
- [ ] Move to the standard inputs (i.e. ScientificInput, BooleanInput, etc.) in stead of only spinbox and try to get default, limits etc from the property itself (i.e. not requiring to manually define the limits)
- [ ] Look at best way of displaying the values

### DIFF
--- a/pymeasure/display/widgets.py
+++ b/pymeasure/display/widgets.py
@@ -1019,6 +1019,9 @@ class InstrumentWidget(QtGui.QWidget):
 
         self.instrument = instrument
 
+        self.update_timer = QtCore.QTimer(self)
+        self.update_timer.timeout.connect(self.update_values)
+
         if hasattr(self.instrument, 'name'):
             self.instrument_name = self.instrument.name
         else:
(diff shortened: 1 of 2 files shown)


pymeasure/display/widgets.py
```diff
@@ -39,11 +39,43 @@
 from .Qt import QtCore, QtGui
 from ..experiment import parameters, Procedure
 from ..experiment.results import Results
+from functools import partial
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 
+def input_from_parameter(parameter):
```
CasperSchippers, line 48: I defined a general function to get the type of Input that corresponds to a given type of Parameter. The code is taken from the InputsWidget and this function is also used in that class.

pymeasure/display/widgets.py
```diff
@@ -1129,6 +1129,10 @@ def _layout(self):
         update_hbox.addWidget(self.update_button)
         f_layout.addRow("Auto-update (off / slow / fast)", update_hbox)
 
+    def change_size_for_floating(self):
+        print(self.parent().isFloating())
```
bilderbuchi, line 1133: please use log.warning()|error|etc instead of print. :-)

CasperSchippers: Yes, I will indeed, it is now there just for (temporary) debugging purpose.

bilderbuchi: You could do it with `log.debug()`, use `logging.basicConfig(level=logging.DEBUG)` in your testing script/console. Then debug to your heart's content, maybe adapt the debug output for maximum usefulness, and then...just leave the log in the code, when you need to debug again in the future. :-)

PR #344, opened 2020-10-22
Implement Github actions

### LINKED ISSUE #342: Consider using GitHub Actions over Appveyor/TravisCI.
Travis jobs take a long time to complete, and we use 2 different services (appveyor and travis) for CI. Github now offers Actions, which could unify and simplify our CI setup.

Current GH action usage limits seem very reasonable: https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#usage-limits

bilderbuchi:
Closes #342


bilderbuchi:
@cjermain I don't know why the appveyor build is still running here, Travis seems to have stopped after removing the config file. Do you know how this is configured on the Appveyor side?

bilderbuchi:
@CasperSchippers @cjermain I have completed the transition to Github actions here, please check it out. Should we switch to this? 

One possible improvement that i can still do is to use [Problem matchers](https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md) to tease out warnings from sphinx and pytest to make them available in the [annotations view](https://github.com/ralph-group/pymeasure/actions/runs/322901274), so that people can easily discover them without going through logs. 
I think they could even appear in the PR view, which would be an awesome feature that the other CI solutions don't offer, I think.
Note: the high number of warnings is due to some bugs in the setup-miniconda action integration that should mostly disappear with the next release.

bilderbuchi:
Note to self: There is an action for flake8 checking available, should we want that in the future. (we need to define and clean up linting for the repo first): https://github.com/marketplace/actions/flake8-with-annotations

CasperSchippers:
It looks nice and promising indeed. It seems convenient to have everything in one place.

bilderbuchi:
OK, let's do this. See #346 for linting modifications in the future.

Issue #331, opened 2020-10-14
Parse returned strings on Adapter level


bilderbuchi:
If a device returns strings that don't fit into the simple pattern that Adapter.values can deal with, one has to parse after values() by defining get_process for every property. If the reply structure is consistent for an Instrument, this could probably use some improvement. 

One alternative way to handle this would be to use a modified `extract_value` as the `cast` callable and make it return a float - but this only replaces the use of `get_process=` with `cast=`, so it's not simpler per se.

If one wanted to apply the same processing/parsing to all gets of an Instrument without specifying this for every control/measurement individually, I could see optionally implementing some generic method at the Adapter level. If that is defined for a given Instrument (e.g. via kwarg to the Instrument constructor that gets passed to the Adapter), it gets called within Adapter.values right after receiving the string. This can be used to prepare the reply for the usual separator/cast application, via regex or whatever you want to use, and you only have to define/use this once for the whole Instrument. Of course this depends on the instrument's replies all having the same structure.

_Originally posted by @bilderbuchi in https://github.com/ralph-group/pymeasure/pull/327#issuecomment-707998416_

dkriegner:
Although its of course possible that a specific device has a syntax which allows for the use of a general get_process function I think this would be a very academic case. In the case of #327 this get_process functions are only applied to certain properties. Namely those which return numerical values which requires different functions for every data type. 

One significant simplification would be possible if `get_process` would get `**kwargs` forwarded as its the case for the `values` method. In this case one could use a `cast` (or similar) argument also in `get_process`. Otherwise one can of course write somehow lengthy lambda expressions in get_process which use a general get_process and perform the type cast. But this is also not very readable.

bilderbuchi:
Hm, I was under the impression that your device always replies with something like `Name = somevalue`, and you wanted to avoid defining a `get_process` (or other kwarg) for _every_ control/measurement just to get rid of the `Name = ` part and get numbers back. 
If you handled that at Adapter level, you could avoid adding something to every control/measurement, and the default `values()` behaviour makes nice floats out of `somevalue`.

Or are you arguing for a possibility for more fine-grained control of the `values()` behaviour, for individual controls/measurements, beyond the `call` and `separator` kwargs that don't catch many reply structures? 
I also have a device where basically an appended unit in the reply string that I don't use makes it impossible that `values()` spits out numbers, but I handle this with individual lambdas for `get_process=`.

bilderbuchi:
[...]

PR #345, opened 2020-10-22
add a preprocess_reply function on the adapter value

dkriegner:
this optional callable is used to perform processing of the return string in Adapter.values before any other processing occurs. It was discussed in issue #331 

A typical use case are instruments which need to use very similar `get_process` function. In your code base a prime example is [`oxfordinstruments/ITC503`](https://github.com/ralph-group/pymeasure/blob/4d2a46ef59cc8c25d192f7f52782a18f02e64ab6/pymeasure/instruments/oxfordinstruments/itc503.py#L40). Oxford instruments devices typically reply with one letter before the numerical value which is removed by almost all `Instrument.control/measurement` properties by individual get_process functions. Using this adapter level functionality many such properties will not need an extra get_process function anymore.

Looking forward for your feedback!


pymeasure/adapters/adapter.py
```diff
@@ -32,7 +32,12 @@ class Adapter(object):
     techniques.
 
     This class should only be inhereted from.
+
+    :param preprocess_reply: optional callable used to preprocess strings
+    received from the instrument
```
bilderbuchi, line 37:
```suggestion
    received from the instrument. The callable returns the processed string.
```

pymeasure/adapters/adapter.py
```diff
@@ -59,16 +64,23 @@ def read(self):
...
+    def values(self, command, separator=',', cast=float, preprocess_reply=None):
         """ Writes a command to the instrument and returns a list of formatted
         values from the result 
 
         :param command: SCPI command to be sent to the instrument
         :param separator: A separator character to split the string into a list
         :param cast: A type to cast the result
+        :param preprocess_reply: optional callable used to preprocess values
+        received from the instrument. If not specified the Adapter default is
+        used.
```
bilderbuchi, line 76:
```suggestion
        received from the instrument. The callable returns the processed string.
        If not specified, the Adapter default is used if available, otherwise no
        preprocessing is done.
```

tests/adapters/test_adapter.py
```diff
@@ -37,3 +37,9 @@ def test_adapter_values():
     assert a.values("X,Y,Z") == ['X', 'Y', 'Z']
     assert a.values("X,Y,Z", cast=str) == ['X', 'Y', 'Z']
     assert a.values("X.Y.Z", separator='.') == ['X', 'Y', 'Z']
+
+
+def test_adapter_preprocess_reply():
```
bilderbuchi, line 42: Thanks for the test. Can you please also add tests for the other possible constellations to confirm they work correctly? The following come to my mind, but maybe I am missing some
* adapter-level only (done)
* adapter-level, overridden by values kwarg
* values-kwarg only
* values-kwarg None ensures adapter-level only does not run.

Can you please also add one showing how an Instrument `control` uses this and confirm correct operation?

pymeasure/adapters/adapter.py
```diff
@@ -32,7 +32,12 @@ class Adapter(object):
     techniques.
```
bilderbuchi, line 32: Can you please add some documentation about how this feature can be used, e.g. as an additional section in https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#writing-properties?

bilderbuchi (changes requested):
This is looking great already, thank you. I left a couple suggestions.

pymeasure/adapters/adapter.py
bilderbuchi: I can hopefully assist with the rst stuff if you need help.

dkriegner: I will write some further tests and a first draft for documentation.

tests/adapters/test_adapter.py
dkriegner: > values-kwarg None ensures adapter-level only does not run.
This is against the behavior of the code where None is the default value of the `preprocess_reply` values-kwarg. To deactivate the adapter level one currently has to specify a real noop-function like `lambda v: v`. (Maybe one disadvantage when one explicitly includes the kwarg in the call signature.) Is this behavior sufficient, or you have ideas?

bilderbuchi: Ah, didn't think of that, of course that doesn't work. Yes, a noop lambda seems like the best approach, we just need to document this so folks know how to achieve disabling the preprocessor where needed.

pymeasure/adapters/adapter.py
bilderbuchi: One pattern for flexible processors that might not be obvious to end-users is to use regexes for matching (can even be several ones), and if none match to pass the string through unmodified. That would be good to include in the docs.

dkriegner:
I added a small fix in `FakeInstrument`, some more tests and some documentation. I am not sure if I missed something, but the file your pointed me to did not at all mention `setting` properties? 
Also I found it useful to introduce `set_process` and `get_process` along with `preprocess_reply`. So I ended up writing quite a bit of text and code.

I am not familiar with the doctest code. So I expect there will be many fixes needed. If you point me in the right directions I can investigate this myself.

But I think you should check if what I wrote is generally in the right direction.

docs/dev/adding_instruments.rst
```diff
@@ -327,3 +329,134 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+
+The same can be also achieved by the `preprocess_reply` keyword argument to :func:`Instrument.control <pymeasure.instruments.Instrument.control>` or :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>`. This function is forwarded to :func:`Adapter.values <pymeasure.adapters.values>` and runs directly after receiving the reply from the device. One can therefore take advantage of the built in casting abilities and simplify the code accordingly:
+
+.. testcode::
+
+    Extreme5000.capacity = Instrument.measurement(
+        ":CAP?",
+        """ A measurement returning a capacity in nF in the format '<cap> nF'
+        """,
+        preprocess_reply=lambda v: v.replace('nF', '')
```
bilderbuchi, line 398:
```suggestion
        preprocess_reply=lambda v: v.replace('nF', '')
        # notice how we don't need to cast to float anymore
```

docs/dev/adding_instruments.rst
```diff
@@ -327,3 +329,134 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+    )
+
+The real purpose `preprocess_reply` is, however, for instruments where many/all properties need similar reply processing. `preprocess_reply` can be applied to all :func:`Instrument.control <pymeasure.instruments.Instrument.control>` or :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` properties. For example if all quantities are returned with a unit as in the example above. To avoid running into troubles for other properties this `preprocess_reply` should be clever enough to skip the processing in case for example some identification string is replied. Typically this can be achieved by regular expression matching. In case of no match the reply is returned unchanged:
```
bilderbuchi, line 401:
```suggestion
The real purpose of `preprocess_reply` is, however, for instruments where many/all properties need similar reply processing. `preprocess_reply` can be applied to all :func:`Instrument.control <pymeasure.instruments.Instrument.control>` or :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` properties, for example if all quantities are returned with a unit as in the example above. To avoid running into troubles for other properties this `preprocess_reply` should be clever enough to skip the processing in case it is not appropriate, for example if some identification string is returned. Typically this can be achieved by regular expression matching. In case of no match the reply is returned unchanged:
```

docs/dev/adding_instruments.rst
```diff
@@ -327,3 +329,134 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+.. testcode::
+
+    Extreme5001.channel = Instrument.control(
+        ":CHAN?", ":CHAN %d",
+        """ A string property that controls the measurement channel,
+        which can take the values X, Y, or Z.
+        """,
+        validator=strict_discrete_set,
+        values=[1,2,3],
+        preprocess_reply=lambda v: v
```
bilderbuchi, line 459:
```suggestion
        preprocess_reply=lambda v: v,
```

docs/dev/adding_instruments.rst
```diff
@@ -327,3 +329,134 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+
+.. testcode::
+
+    Extreme5000.current = Instrument.setting(
+        ":CURR %g",
+        """ A floating point property that takes the measurement current in A
+        """,
+        validator=strict_range,
+        values=[0, 10],
+        set_process=lambda v: 1e3*v  # convert current to mA
```
bilderbuchi, line 347:
```suggestion
        set_process=lambda v: 1e3*v,  # convert current to mA
```
Putting a trailing comma on the last kwarg like this keeps the diff smaller when modifying/adding kwargs, as this line does not need to be touched.

docs/dev/adding_instruments.rst
```diff
@@ -327,3 +329,134 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+.. testcode::
+
+    Extreme5000.current = Instrument.control(
+        ":CURR?", ":CURR %g",
+        """ A floating point property representing the measurement current in A
+        """,
+        validator=strict_range,
+        values=[0, 10],
+        set_process=lambda v: 1e3*v  # convert to mA
+        get_process=lambda v: 1e-3*v  # convert to A
```
bilderbuchi, line 369:
```suggestion
        set_process=lambda v: 1e3*v,  # convert to mA
        get_process=lambda v: 1e-3*v,  # convert to A
```

docs/dev/adding_instruments.rst
```diff
@@ -327,3 +329,134 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+        "*idn?",
+        """ The identification of the instrument.
+        """)
+
+        def __init__(self, resourceName, **kwargs):
+            super().__init__(
+                resourceName,
+                "Extreme 5000",
+                preprocess_reply=extract_value,
+                **kwargs
```
bilderbuchi, line 445:
```suggestion
                **kwargs,
```

docs/dev/adding_instruments.rst
```diff
@@ -327,3 +329,134 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+        """
+        r = _reg_value.search(reply)
+        if r:
+            return r.groups()[0]
+        else:
+            return reply
+
+    class Extreme5001(Instrument):
+        """ Represents the imaginary Extreme 5001 instrument. This instrument
+        sends numerical values including there units in an format "<value>
```
bilderbuchi, line 422:
```suggestion
        sends numerical values including their units in an format "<value>
```

tests/instruments/test_instrument.py
```diff
@@ -150,6 +151,41 @@ class Fake(FakeInstrument):
     assert fake.x == 5
 
 
+def test_control_preprocess_reply():
+    # test setting preprocess_reply at property-level
+    class Fake(FakeInstrument):
+        x = Instrument.control(
+            "", "JUNK%d", "",
+            preprocess_reply=lambda v: v.replace('JUNK', ''),
```
bilderbuchi, line 159:
```suggestion
            "", "JUNK%d",
            "",
            preprocess_reply=lambda v: v.replace('JUNK', ''),
```

tests/instruments/test_instrument.py
```diff
@@ -150,6 +151,41 @@ class Fake(FakeInstrument):
...
+            cast=int
+        )
+
+    fake = Fake()
+    fake.x = 5
+    assert fake.read() == 'JUNK5'
```
bilderbuchi, line 165: huh, I am confused why this works, i.e. why `read()` returns `JUNK5` - maybe add a comment explaining this?

tests/instruments/test_instrument.py
```diff
@@ -150,6 +151,41 @@ class Fake(FakeInstrument):
...
+    fake.x = 5
+    assert fake.x == 5
+    fake.x = 5
+    assert type(fake.x) == int
```
bilderbuchi, line 169: I think the second fake.x is not needed here?
```suggestion
    assert fake.x == 5
    assert type(fake.x) == int
```

tests/instruments/test_instrument.py
```diff
@@ -150,6 +151,41 @@ class Fake(FakeInstrument):
...
+
+    # test setting preprocess_reply at Adapter-level
```
bilderbuchi, line 171: can you put this in a separate test, please?

bilderbuchi (changes requested):

bilderbuchi:
This is great, including the additional documentation which improves that aspect (I think this was mainly/only in the API reference so far).
Yeah, let's see about the doctests, I have to re-familiarize myself. 
You should be able to run them locally if you install `sphinx` and `sphinx_rtd_theme` and then in `docs` do `make doctest`.

pymeasure/instruments/instrument.py
```diff
@@ -318,7 +318,7 @@ class FakeInstrument(Instrument):
 
     def __init__(self, adapter=None, name=None, includeSCPI=False, **kwargs):
         super().__init__(
-            FakeAdapter(),
+            adapter or FakeAdapter(),
```
bilderbuchi, line 321: Please revert this - otherwise if you pass a string as `adapter`, the Instrument init will go the usual route and try to make a VisaAdapter which will fail. This is why the doctests fail.

bilderbuchi (changes requested):
Found the cause why many doctests fail.

tests/instruments/test_instrument.py
dkriegner: I will add a comment. `read` is used inside `ask`, which is used by `values`. Only in `values` the `preprocess_reply` is applied.

tests/instruments/test_instrument.py
dkriegner: Its a `FakeInstrument` with `FakeAdapter`. `FakeAdapter.read` seems to clear the buffer and therefore the second set is needed.
Here the corresponding error (of `python setup.py test` otherwise:
```
_______________________________ test_control_validator ________________________________

    def test_control_validator():
        class Fake(FakeInstrument):
            x = Instrument.control(
                "", "%d", "",
                values=range(10),
            )
    
        fake = Fake()
        fake.x = 5
        assert fake.read() == '5'
    #    fake.x = 5
>       assert fake.x == 5
E       AssertionError: assert '' == 5
E        +  where '' = <test_instrument.test_control_validator.<locals>.Fake object at 0x7f13a9713b50>.x

tests/instruments/test_instrument.py:62: AssertionError
```

pymeasure/instruments/instrument.py
dkriegner: I can't just revert. The problem is that to test `preprocess_reply` at the adapter level one should be able to pass kwargs to the Adapter constructor. I understand you want to have your strings in the documentation but still get a `FakeAdapter`, right?

dkriegner: a90e4bd should fix this. Let me know if this is ok?

bilderbuchi: I think the FakeAdapter needs an init that is adapted to the new structure, e.g. try this:
```
    def __init__(self, preprocess_reply=None, **kwargs):
        self.preprocess_reply = preprocess_reply
        # kwargs are discarded as they are non-functional.
```
then here you should be able to say 
```
        super().__init__(
            FakeAdapter(**kwargs),
```

tests/instruments/test_instrument.py
dkriegner: done

tests/instruments/test_instrument.py
bilderbuchi: ah, right.

pymeasure/instruments/instrument.py
bilderbuchi: ^ to be honest, I would prefer the outlined approach (if it proves to work as intended) to a90e4bd .

dkriegner: `FakeAdapter` has this init already! inherited from `Adapter`. I would not like your proposal because you mix `kwargs` of `Adapter` and `Instrument`, but I guess for pyvisa instruments that is what you already do... If you dislike my fix let me know

dkriegner: sorry, again crossed comments. I now implemented your suggestion. Seems to work as intended.

bilderbuchi:
I just noticed: did you leave out the PrologixAdapter intentionally? Shouldn't that one also get its init signature adapted to be consistent with the others (although it works as is, I think).

pymeasure/adapters/adapter.py
```diff
@@ -59,16 +64,24 @@ def read(self):
...
         """ Writes a command to the instrument and returns a list of formatted
         values from the result 
 
         :param command: SCPI command to be sent to the instrument
         :param separator: A separator character to split the string into a list
         :param cast: A type to cast the result
+        :param preprocess_reply: optional callable used to preprocess values
+        received from the instrument. The callable returns the processed string.
+        If not specified, the Adapter default is used if available, otherwise no
+        preprocessing is done.
```
bilderbuchi, line 77:
```suggestion
        :param preprocess_reply: optional callable used to preprocess values
            received from the instrument. The callable returns the processed string.
            If not specified, the Adapter default is used if available, otherwise no
            preprocessing is done.
```

pymeasure/adapters/adapter.py
```diff
@@ -32,7 +32,12 @@ class Adapter(object):
     techniques.
 
     This class should only be inhereted from.
+
+    :param preprocess_reply: optional callable used to preprocess strings
+    received from the instrument. The callable returns the processed string.
```
bilderbuchi, line 37:
```suggestion
    :param preprocess_reply: optional callable used to preprocess strings
        received from the instrument. The callable returns the processed string.
```

dkriegner:
The PrologixAdapter is derived from SerialAdapter and therefore does IMHO not need a code change. `kwargs` are forwarded to the SerialAdapter constructor so it should just work as is

bilderbuchi (changes requested):
some docstring fixes found by CI.

bilderbuchi:
Yeah, I think we should adapt all of them (would also apply to the AttocubeConsoleAdapter, DanfysikAdapter and LakeshoreUSBAdapter). I know it's basically a no-op at this point, bu we can demonstrate "best practice" - I think many contributors start by looking at other already implemented instruments to see how to go about the one they implement.

dkriegner:
I understand your argument, but I tend to disagree. The people writing Adapters in the instrument folders already should know if the `preprocess_reply` functionality is relevant for their instrument. Adding an explicit change like this would also require the docstring to mention this argument.

```
--- a/pymeasure/instruments/lakeshore/adapters.py
+++ b/pymeasure/instruments/lakeshore/adapters.py
@@ -35,13 +35,14 @@ class LakeShoreUSBAdapter(SerialAdapter):
     :param port: A string representing the serial port
     """
 
-    def __init__(self, port):
+    def __init__(self, port, preprocess_reply=None):
         super(LakeShoreUSBAdapter, self).__init__(
             port,
             baudrate=57600,
             timeout=0.5,
             parity='O',
-            bytesize=7
+            bytesize=7,
+            preprocess_reply=preprocess_reply,
         )
 
     def write(self, command):
```

bilderbuchi:
OK, as you propose is fine by me. I don't have a strong preference. :-)

dkriegner:
What is maybe problematic with the Adapters in the instrument/vendor folders is that some do not even allow to forward keyword arguments to the underlying Adapters. But I would solve this by a generic `**kwarg` forwarding. If you want that give me a hint. (But this does in principle not interfere with this PR)

should I add the docstring of the `preprocess_reply` argument to all Adapters which explicitly mention it? I guess this would be appropriate. It currently reads:

 ```
:param preprocess_reply: optional callable used to preprocess strings
received from the instrument. The callable returns the processed string.  
```
Do you have proposals for changes? Should the second line be intended?

bilderbuchi:
OK, then leave out the vendor adapters, and only fix the prologix one.
Yes, add the parameter description to the docstrings. The docstring is fine, include the second line; but you need an indentation in the second line.

bilderbuchi:
Thank you very much for your patience, I think we are done here now. This is a very nice feature I think!

PR #347, opened 2020-10-24
Add CI problem matcher for Sphinx.

bilderbuchi:
Let's see if we can have CI show doc errors in the PR files view, which would be useful to diagnose/fix docs mistakes.

### DIFF
--- a/.github/sphinx.json
+++ b/.github/sphinx.json
@@ -1,0 +1,27 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "sphinx",
+            "pattern": [
+                {
+                    "regexp": "^(?<path>.+?)(?<location>:.*?|): (?<severity>WARNING|ERROR): (?<message>.+)$",
+                    "file": 1,
+                    "line": 2,
+                    "severity": 3,
+                    "message": 4
+                }
+            ]
+        },
+        {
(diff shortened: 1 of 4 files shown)


bilderbuchi:
The following issues present themselves:
Checking sphinx-doctest for errors picks them up, but since we are in the working directory /docs, the files are not resolved to the correct location, so the errors are annotated at the correct location, and are not displayed in the PR's "Files" section

Sphinx-doctest error messages consist of one line for the path, and then several lines for the error message, followed by a line o ****. Github actions problem matchers cannot deal with this correctly currently. See also https://github.com/actions/toolkit/issues/193 https://github.com/actions/toolkit/issues/319 
Interestingly, the VS code problem matchers this seems to originate from [supports](https://code.visualstudio.com/docs/editor/tasks#_defining-a-problem-matcher) a ` fileLocation` field where you can indicate how the path is supposed to be read.

bilderbuchi:
Reported a problem resolving the relative paths out of `make doctest` here: https://github.com/actions/runner/issues/765.

bilderbuchi:
I have squashed & pushed this directly to master, but wanted to avoid force-pushing here so that the logs presented in https://github.com/actions/runner/issues/765 don't disappear. Closing as everything relevant is in master now.

PR #351, opened 2020-10-27
use preprocess_reply functionality in Attocube ANC300

dkriegner:
using `preprocess_reply` the code of the `AttocubeConsoleAdapter` and ANC300 properties was simplified

* `extract_value` function was converted into a method of `AttocubeConsoleAdapter` and is now used as `preprocess_reply`
* `extract_float`, and `extract_int` are not needed anymore
* most properties now work without a dedicated `get_process`, also note that not a single property needed to deactivate the default `preprocess_reply`


bilderbuchi:
It's great to see how the new feature can simplify an instrument implementation!

PR #352, opened 2020-10-27
New Instrument: Toptica Smart Laser diode

dkriegner:
This pull request implements a [laser diode](https://www.toptica.com/products/single-mode-diode-lasers/ibeam-smart/) we use in our lab.

The device has a serial interface over which it provides some sort of Console. This seems to be more oriented towards interactive use which comes with some annoyances.

The communication protocol is as follows (sent messages >, received <):

asking the instrument version:
```
> 'ver\r\n'
< '\r\n'
< 'iBPs-001A01-05\r\n'
< '[OK]\r\n'
```

asking a numerical value (laser power):
```
> 'sh power\r\n'
< '\r\n'
< 'PIC  = 000001 uW  \r\n'
< '[OK]\r\n'
```

setting a value
```
> 'la off\r\n'
< '\r\n'
< '[OK]\r\n'
```

The first `\r\n` is received directly after the request. Although both `echo` and `prompt` are turned off (otherwise there is even more clutter) I believe this is some (intentional?) leftover of the echo the device in its default settings is returning.

So this device's protocol has some similarities with the Attocube/anc300 device merged in PR #327. Since I received some comments this could be implemented via the `check_errors` functionality I attempted this at first without actually making a dedicated adapter. It worked, but IMHO was not very elegant. It felt especially awkward that I had to reimplement `Instrument.values` because otherwise my `read/write/ask` methods would have been bypassed. Also every property had to use `check_set/get_errrors=True`. [The `check_errors` implementation can be seen here](https://github.com/dkriegner/pymeasure/blob/ibeam/pymeasure/instruments/toptica/ibeamsmart.py).

The code proposed here therefore includes its own adapter (similar to the AttocubeConsoleAdapter), but based on `VISAAdapter`. It implements the same awkward read back of the confirmation replies inside `write`. If you see a more elegant way to solve this or consider the `check_errors` approach better suited let me know.

sorry for the many pull requests, which I am aware cause a lot of work on your end (this will be my last in a while!). I appreciate the effort and all the work you already put into this project!


pymeasure/instruments/toptica/adapters.py
```diff
@@ -0,0 +1,132 @@
...
+
+        :param reply: reply string of the instrument which should be checked
+        """
+        if reply != '[OK]':
+            self.flush()
+            raise ValueError(
+                f"TopticaAdapter: Error after command '{self.lastcommand}' "
+                f"with message '{reply}'")
+
+    def _read_stripped(self):
```
bilderbuchi, line 83: are you sure this is necessary? According to the [docs](https://pyvisa.readthedocs.io/en/1.8/resources.html?highlight=read_termination#termination-characters) should already strip `read_termination` for you.

pymeasure/instruments/toptica/adapters.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    # compiled regular expression for finding numerical values in reply strings
+    _reg_value = re.compile(r"\w+\s+=\s+(\w+)")
+
+    def __init__(self, port, baud_rate, **kwargs):
+        kwargs.setdefault('preprocess_reply', self.extract_value)
+        super().__init__(port, **kwargs)
+        # hack to set baud_rate since VISAAdapter filters it out when spcified
+        # as keyword argument. issue #334
+        self.connection.baud_rate = baud_rate
```
bilderbuchi, line 47: agreed, this should be fixed.

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -0,0 +1,117 @@
...
+            "sh temp",
+            """ temperature of the laser diode in degree centigrade.""",
+    )
+
+    system_temp = Instrument.measurement(
+            "sh temp sys",
+            """ base plate (heatsink) temperature in degree centigrade.""",
+    )
+
+    laser = Instrument.control(
```
bilderbuchi, line 55: maybe clearer as "laser_enabled" with a boolean?

bilderbuchi (changes requested):
Did a really quick skim through, will get back to this sometime later.

pymeasure/instruments/toptica/adapters.py
dkriegner: I forgot to set read_termination and by default its empty. will fix this and remove this function

pymeasure/instruments/toptica/ibeamsmart.py
dkriegner: you are right. I changed this.

dkriegner:
commit 7675a60 is in principle unrelated to this PR, but seems to be fix a bug. One could also directly use `VISAAdapter.resource_name`... and be independent of the pyvisa stuff

dkriegner:
Are there more changes needed here? I guess the existing comments were somehow fixed but no thorough review was done

bilderbuchi:
We/I have not yet had the time for a in-depth review, sorry. I am currently trying to concentrate on fixing bugs, but time is always short. I can only ask for your patience.

pymeasure/instruments/toptica/adapters.py
```diff
@@ -0,0 +1,129 @@
...
+
+from pymeasure.adapters import VISAAdapter
+
+
+class TopticaAdapter(VISAAdapter):
+    """ Adapter class for connecting to Toptica Console via a serial
+    connection.
+
+    :param port: pyvisa resource name of the instrument
+    :param baud_rate: communication speed, defaults to 115200
```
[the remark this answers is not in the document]
bilderbuchi: Probably better to have those settings in the instrument, if my assumption is correct that the Toptica Console could be generic for more than 1 instrument (with different comms needs/settings)

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    laser_enabled = Instrument.control(
+            "sta la", "la %s",
+            """ Status of the laser diode driver.
+                This can be True if the laser is on or False otherwise""",
+            validator=strict_discrete_set,
+            values=[True, False],
+            get_process=lambda s: True if s == 'ON' else False,
+            set_process=lambda v: "on" if v else "off",
```
bilderbuchi, line 62: does capitalization really differ between send and receive? Otherwise you can use the pattern of 
```
        map_values=True,
        values={True: "on", False: "off"},
```
instead of [set|get]_process

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -0,0 +1,118 @@
...
+    power = Instrument.control(
+            "sh pow", "ch pow %f mic",
+            """ Actual output power in uW of the laser system. In pulse mode
+            this means that the set value might not correspond to the readback
+            one.""",
+            validator=strict_range,
+            values=[0, 200000],
+    )
+
+    def __init__(self, port, baud_rate=115200, **kwargs):
```
bilderbuchi, line 94: I think it would be good to check that port is a string, and otherwise raise, because all this does not work with anything else than a Toptica/VISAAdapter, and people could pass a SerialAdapter in.

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -0,0 +1,118 @@
...
+
+    power = Instrument.control(
+            "sh pow", "ch pow %f mic",
+            """ Actual output power in uW of the laser system. In pulse mode
+            this means that the set value might not correspond to the readback
+            one.""",
+            validator=strict_range,
+            values=[0, 200000],
+    )
+
```
bilderbuchi, line 93: Kindly move the init to the top of the class, please.

bilderbuchi (changes requested):
Looks good, thank you. There are a couple questions/requests, let's work this out. :-) I'll be faster with the follow-ups, now that the initial effort is done :-)

bilderbuchi:
> It felt especially awkward that I had to reimplement Instrument.values because otherwise my read/write/ask methods would have been bypassed.

This is because the VISAAdapter is the only one of ours which delegates to a specific (pyvisa's) `query` method instead of doing write-then-read as the others.

pymeasure/instruments/toptica/adapters.py
bilderbuchi: not gonna be fixed in here, and a link to the issue is present.

pymeasure/adapters/visa.py
```diff
@@ -154,5 +154,9 @@ def wait_for_srq(self, timeout=25, delay=0.1):
         """
         self.connection.wait_for_srq(timeout * 1000)
 
+    def flush(self):
+        """ Flush the input buffer """
+        self.connection.flush(pyvisa.constants.VI_READ_BUF)
```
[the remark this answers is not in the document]
dkriegner: I went for the flush_read_buffer name since this is what I want to do and it seems fitting.

pymeasure/instruments/toptica/adapters.py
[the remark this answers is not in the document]
dkriegner: I put it to the instrument only. The default speed can in priniciple be changed so it has to remain an argument of the instrument.

pymeasure/instruments/toptica/adapters.py
```diff
@@ -0,0 +1,129 @@
...
+        :returns: string with only the numerical value, or the original string
+        """
+        r = self._reg_value.search(reply)
+        if r:
+            return r.groups()[0]
+        else:
+            return reply
+
+    def check_acknowledgement(self, reply):
+        """ checks if reply is '[OK]', otherwise a ValueError is raised and the
```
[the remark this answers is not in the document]
dkriegner: `[OK]` is even returned after an error or invalid command. I think the `[OK]` has to be understood more as an indicator that the device is ready to receive more commands. In case an invalid command is sent an error code is returned before the `[OK]`.

e.g. ">" indicates send things, "<" received lines

```
> 'sh invalid\r\n'
< '\r\n'
< '%SYS-W-027, command unknown'
< '[OK]\r\n'
```

pymeasure/instruments/toptica/adapters.py
```diff
@@ -0,0 +1,129 @@
...
+                f"with message '{reply}'")
+
+    def read(self):
+        """ Reads a reply of the instrument which consists of two lines. The
+        first one is the reply to the command while the last one should be
+        '[OK]' which acknowledges that the device is ready to receive more
+        commands.
+
+        Note: This command only understands replies with one data line. More
+        complicated replies have to be parsed by using the underlying adapter
```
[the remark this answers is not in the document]
dkriegner: I changed the code to allow for multiline replies. please check if this seems ok for you.

pymeasure/instruments/toptica/adapters.py
```diff
@@ -0,0 +1,129 @@
...
+        self.lastcommand = command
+        super().write(command)
+        # read back the LF+CR which is always sent back.
+        reply = super().read()
+        if reply != '':
+            raise ValueError(
+                f"TopticaAdapter.write: Error after command '{command}' with "
+                f"message '{reply}'")
+        if check_ack:
+                self.check_acknowledgement(super().read())
```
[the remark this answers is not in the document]
dkriegner: I included a few lines to make it clear, also referring to this PR.

pymeasure/instruments/toptica/ibeamsmart.py
dkriegner: unfortunately it really differs and the device understands it only in the way implemented.

pymeasure/instruments/toptica/ibeamsmart.py
dkriegner: done

dkriegner:
I hopefully addressed all the comments, tested that the device works. thanks for the comments, I think the device/adapter implementation is better now.

pymeasure/instruments/toptica/adapters.py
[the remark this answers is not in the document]
bilderbuchi: Looking good, thanks!

pymeasure/instruments/toptica/ibeamsmart.py
bilderbuchi: :rofl: crazy

bilderbuchi:
Looks good, I just adapted one docstring a little. Thank you for your contribution, merging!

PR #354, opened 2020-10-27
Fix ZMQ bug

### LINKED ISSUE #168: ZMQ error

bilderbuchi:
Closes #168.

Let's first see that we can trigger the problem.

### DIFF
--- a/pymeasure/experiment/workers.py
+++ b/pymeasure/experiment/workers.py
@@ -156,9 +156,10 @@ class Worker(StoppableThread):
                 self.publisher = self.context.socket(zmq.PUB)
                 self.publisher.bind('tcp://*:%d' % self.port)
                 log.info("Worker connected to tcp://*:%d" % self.port)
-                time.sleep(0.01)
+                time.sleep(0.3)  # wait so that the socket will be ready before starting to emit messages
             except Exception:
                 log.exception("couldn't connect to ZMQ context")
+                # FIXME: this happily continues even if connecting to the context fails!
 
         log.info("Worker started running an instance of %r", self.procedure.__class__.__name__)
         self.update_status(Procedure.RUNNING)
--- a/pymeasure/experiment/workers.py
+++ b/pymeasure/experiment/workers.py
@@ -76,8 +76,26 @@ class Worker(StoppableThread):
         self.log_queue = log_queue
         self.log_level = log_level
 
+        global log
+        log = logging.getLogger()
+        log.setLevel(self.log_level)
+        # log.handlers = []  # Remove all other handlers
+        # log.addHandler(TopicQueueHandler(self.monitor_queue))
+        # log.addHandler(QueueHandler(self.log_queue))
+
         self.context = None
         self.publisher = None
+        if self.port is not None and zmq is not None:
+            try:
+                self.context = zmq.Context()
+                log.debug("Worker ZMQ Context: %r" % self.context)
+                self.publisher = self.context.socket(zmq.PUB)
+                self.publisher.bind('tcp://*:%d' % self.port)
+                log.info("Worker connected to tcp://*:%d" % self.port)
+                time.sleep(0.3)  # wait so that the socket will be ready before starting to emit messages
+            except Exception:
+                log.exception("couldn't connect to ZMQ context")
+                # FIXME: this happily continues even if connecting to the context fails!
 
     def join(self, timeout=0):
         try:
@@ -136,12 +154,6 @@ class Worker(StoppableThread):
             self.context.term()
 
     def run(self):
-        global log
-        log = logging.getLogger()
-        log.setLevel(self.log_level)
-        # log.handlers = []  # Remove all other handlers
-        # log.addHandler(TopicQueueHandler(self.monitor_queue))
-        # log.addHandler(QueueHandler(self.log_queue))
         log.info("Worker thread started")
 
         self.procedure = self.results.procedure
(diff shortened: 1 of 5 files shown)


bilderbuchi:
OK, I have started a proposed fix, but I am running into weird issues. Locally (py3.7), if I run the new `test_zmq_serialisation` test, the test passes but then pytest hangs until I press Ctrl-C:
```
(base) bilderbuchi@osprey:~/Python/pymeasure$ pytest -rsx -v -k zmq
========================================================================= test session starts =========================================================================
platform linux -- Python 3.7.9, pytest-6.1.1, py-1.9.0, pluggy-0.13.1 -- /home/bilderbuchi/anaconda3/bin/python
cachedir: .pytest_cache
PyQt5 5.9.2 -- Qt runtime 5.9.7 -- Qt compiled 5.9.6
rootdir: /home/bilderbuchi/Python/pymeasure
plugins: qt-3.3.0, datadir-1.3.2.dev1+gf4f6a5f
collected 73 items / 72 deselected / 1 skipped                                                                                                                        

tests/experiment/test_workers.py::test_zmq_serialisation PASSED                                                                                                 [100%]

^CException ignored in: <function Context.__del__ at 0x7fa6a408f4d0>
  File "/home/bilderbuchi/anaconda3/lib/python3.7/site-packages/zmq/sugar/context.py", line 54, in __del__
    self.term()
  File "/home/bilderbuchi/anaconda3/lib/python3.7/site-packages/zmq/sugar/context.py", line 154, in term
    return super(Context, self).term()
  File "zmq/backend/cython/context.pyx", line 91, in zmq.backend.cython.context.Context.term
  File "zmq/backend/cython/checkrc.pxd", line 13, in zmq.backend.cython.checkrc._check_rc
KeyboardInterrupt

======================================================================= short test summary info =======================================================================
SKIPPED [1] tests/instruments/agilent/test_agilent34450A.py:30: Only work with connected hardware
============================================================= 1 passed, 1 skipped, 72 deselected in 4.55s =============================================================
```
In CI, this fails even

bilderbuchi:
By now the indications are becoming more that this is a problem with not setting the ZQM Linger setting: https://stackoverflow.com/questions/44273941/python-script-not-terminating-after-timeout-in-zmq-recv https://stackoverflow.com/questions/7939977/zeromq-with-python-hangs-if-connecting-to-invalid-socket
I still have to check if this can be solved by a cleaner shutdown procedure, i.e. making sure to consume all messages before shutdown, or just setting a `linger` value and being done with it.

bilderbuchi:
hey @CasperSchippers @Kassikakk @jmittelstaedt: I have fixed the ZMQ problems reported in #168. Could you maybe test if this PR works in your real workflow? 
I have added some tests that are passing (on all platforms now), but want to make sure that this also covers a real-life scenario.

pymeasure/experiment/workers.py
```diff
@@ -76,8 +76,26 @@ def __init__(self, results, log_queue=None, log_level=logging.INFO, port=None):
...
+            try:
+                self.context = zmq.Context()
+                log.debug("Worker ZMQ Context: %r" % self.context)
+                self.publisher = self.context.socket(zmq.PUB)
+                self.publisher.bind('tcp://*:%d' % self.port)
+                log.info("Worker connected to tcp://*:%d" % self.port)
+                time.sleep(0.3)  # wait so that the socket will be ready before starting to emit messages
+            except Exception:
+                log.exception("couldn't connect to ZMQ context")
+                # FIXME: this happily continues even if connecting to the context fails!
```
CasperSchippers, line 98: What about this remark? Is this a problem (it's not completely clear to me what happens if connecting fails)

CasperSchippers (commented):
Awesome! Well done @bilderbuchi!
I tested it in various environments and as far as I can see it works like a charm.

tests/experiment/test_workers.py
```diff
@@ -82,3 +85,55 @@ def test_worker_closes_file_after_finishing():
 
     # Test if the file has been properly closed by removing the file
     os.remove(file)
+
+
+@pytest.mark.skipif(not importlib.util.find_spec('cloudpickle'),
```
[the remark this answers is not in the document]
bilderbuchi: Indeed!

pymeasure/experiment/workers.py
bilderbuchi: I wanted to postpone that, but indeed it's better to do it right away - I let this fall back to proper values (as if zmq were not installed)

bilderbuchi:
Thanks for your feedback, I'll merge this, then. :-)

Issue #356, opened 2020-11-05
Keithley2400 and Keithley2450 classes: incompatible functions


pyMatJ:
Hello,

Recently added functions in the Keithley2400 class ( [https://github.com/ralph-group/pymeasure/pull/326](https://github.com/ralph-group/pymeasure/pull/326) ) are actually written in the command syntax that match the 2450 series. The syntax for the 2400 class is very close. (code snippets at the end).

What is the preferred method to fix this, i.e. correct the syntax in the keithley2400 driver and add them in the keithley2450 one ?
I can. Should I prepare a PR with the modifications/corrections, or is there another preferred way ?

Thank you in advance,

Mathieu

----------------------------- code snippets ------------------
The functions in question are all from the `Filters` section. Basically, they include a ` :FUNC: ` (`:CURR:` or `:VOLT:` ) which acutally belongs to the 2450 series.
Current implementation in 2400 (which instead should go in 2450) example:
```python
 current_filter_count = Instrument.control(
        ":SENS:CURR:AVER:COUNT?", ":SENS:CURR:AVER:COUNT %d",
        """ A integer property that controls the number of readings that are 
        acquired and stored in the filter buffer for the averaging""",
        validator=truncated_range,
        values=[1, 2500],
        cast=int)
```
Correct implementation in 2400 example:
``` python
current_filter_count = Instrument.control(
        ":SENS:AVER:COUNT?", ":SENS:AVER:COUNT %d",
        """ A integer property that controls the number of readings that are 
        acquired and stored in the filter buffer for the averaging""",
        validator=truncated_range,
        values=[1, 2500],
        cast=int)
```

bilderbuchi:
Hi! Yes, a PR would be great! Can you also check the other properties added in that PR, maybe they also need to be modified. From what I can see, it seems these properties are generic in the 2400, while the 2450 has got separate voltage/current ones, correct? 
also, pinging @paulgoulain who contributed #326 - do you have any further insight here?

paulgoulain:
Hi. Sadly no, I apparently made a boo-boo here. I didn't double checked well enough the changes. I'll check with Mathieu if the other functions I implemented in this PR are also compatible with our 2401.

pyMatJ:
I checked with @paulgoulain and the other functions were redacted with the 2450 syntax as well. I fixed all of them and submitted the PR https://github.com/ralph-group/pymeasure/pull/357

PR #357, opened 2020-11-06
Fixing #356: incompatible functions btw keithley2400 and keithley2450

pyMatJ:
Fixing issue [356](https://github.com/ralph-group/pymeasure/issues/356): incompatible functions introduced by PR [326](https://github.com/ralph-group/pymeasure/pull/326) in keithley2400 driver that actually belong to keithley2450. 

Corrected the functions in keithley2400 and added the corresponding ones in keithley2450.

_EDIT:_
To give more context: I am writing an IV program that should run on both Keithley2400 and 2450 series. While comparing the two syntax, I realized some functions are incorrect and/or missing in either one of the two drivers so I am trying to get them to a similar state, despite the slightly different syntax.

### DIFF
--- a/pymeasure/instruments/keithley/keithley2400.py
+++ b/pymeasure/instruments/keithley/keithley2400.py
@@ -307,53 +307,37 @@ class Keithley2400(Instrument, KeithleyBuffer):
     # Filters #
     ###########
 
-    current_filter_type = Instrument.control(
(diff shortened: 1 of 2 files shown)


bilderbuchi:
Thanks, and thanks @paulgoulain for cross-checking!

PR #361, opened 2020-11-12
Fix for warning by pyvisa on using deprecated API (get_instrument)

msmttchr:
Minor fix on pyvisa deprecated API

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -28,3 +28,4 @@ Dominik Kriegner
 Jonathan Larochelle
 Dominic Caron
 Mathieu Plante
+Michele Sardo
--- a/docs/about/authors.rst
+++ b/docs/about/authors.rst
@@ -28,3 +28,4 @@ The following developers have contributed to the PyMeasure package:
 | Jonathan Larochelle
 | Dominic Caron
 | Mathieu Plante
+| Michele Sardo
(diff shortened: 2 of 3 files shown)


CasperSchippers (approved):
Thanks for this fix. 

Could you add yourself to the authors lists at AUTHORS.txt and docs/about/authors.rst.

bilderbuchi:
LGTM. I am wondering why we don't see the FutureWarning in our pytest runs, though? How did you become aware of the deprecation, @msmttchr ?

bilderbuchi:
> I am wondering why we don't see the FutureWarning in our pytest runs, though.

Oh I see, we don't cover that part of the code, yet - see #362

msmttchr:
> LGTM. I am wondering why we don't see the [FutureWarning](https://github.com/pyvisa/pyvisa/blob/master/pyvisa/highlevel.py#L3352) in our pytest runs, though. Is there a way to surface these warnings in CI? It's not the first pyvisa deprecation we address.
> How did you become aware of the deprecation, @msmttchr ?

I just used the module in my setup and the warning is shown in the DOS box.
I thought this could be related also to the version of pyvisa people are using.

bilderbuchi:
Yes it is related, I confirmed that the change is fine with out supported range of pyvisa versions.

Issue #362, opened 2020-11-12
config method in VISAAdapter class (adapter/visa.py) is not supported since pyvisa v1.10


msmttchr:
The config method probably needs to be updated to be compatible with latest version of pyvisa.
The values_format has been removed and it is no longer supported.
The change in pyvisa where support has been removed says "* pyvisa: drop Python 2 compat stuff" (28 aug 2018).

bilderbuchi:
Thanks for the heads-up, we should fix this before the next release!
A couple of other things we should fix at the same time:
* I found out that we don't see these deprecation warnings in time since we don't have test coverage of the VisaAdapter. We should implement a test that exercises that part of the codebase, maybe using pyvisa-sim that we already pull in for another test, or with some other faked instrument? That should be implemented first to check if it correctly surfaces the warnings in CI.
* We also need to fix the call to ` query_values(command)`, probably by using [`query_ascii_values(command)`](https://github.com/pyvisa/pyvisa/blob/master/CHANGES#L37-L48). 
* I'm not sure if our `.binary_values` should use `read_binary_values` instead of ` read_raw` -- we should take a look at that.

cjermain:
Thanks for raising these. I believe they are fully addressed in #391. Please let us know if not.

msmttchr:
Looks good to me. Thanks

PR #363, opened 2020-11-13
Added write_binary_values API

msmttchr:
File updates are  instrument, adapter and visa.
No implementation provided for serial adapter and no tests added.
Minor fix on wrong name for variable representing the resource name.

### DIFF
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -107,6 +107,16 @@ class Adapter(object):
         raise NameError("Adapter (sub)class has not implemented the "
                         "binary_values method")
 
+    def write_binary_values(self, command, values, format='B'):
+        """ Write binary data to the instrument, e.g. waveform for signal generators
+
+        :param command: SCPI command to be sent to the instrument
+        :param values: iterable representing the binary values
+        :format: "struct module" format character for each list item (see struct documentation)
+.       :returns: number of bytes written
+        """
+        raise NameError("Adapter (sub)class has not implemented the "
+                        "write_binary_values method")
 
 class FakeAdapter(Adapter):
     """Provides a fake adapter for debugging purposes,
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -78,7 +78,7 @@ class VISAAdapter(Adapter):
             return False
 
     def __repr__(self):
-        return "<VISAAdapter(resource='%s')>" % self.connection.resourceName
+        return "<VISAAdapter(resource='%s')>" % self.connection.resource_name
 
     def write(self, command):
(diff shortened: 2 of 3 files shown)


bilderbuchi:
Thanks! I think it makes sense to take a look to clean up this part of the adapter API while we are changing this. 
Thoughts, @cjermain @CasperSchippers?

* As a certain generality w.r.t. adapters is one of the features of pymeasure, we should try to add implementations for other adapters, at least the SerialAdapters, the others as appropriate.
* We should probably try to come up with some tests to make sure that the results match expectation, also across adapters.
* for consistency with the ascii read/write/ask nomenclature, rename `binary_values` -> `ask_binary_values` ?
* as we need some visa changes anyway (see #362), maybe we can get rid of the `VISAAdapter.config` use in `ask_values` and implement this in a way more aligned with the pyvisa API
 * VISAAdapter: `binary_values` could use `self.read_bytes`?
 * Check if our API language is still consistent w.r.t. read/write/ask, `binary` vs. nothing, what the `_values` suffix is supposed to mean/signify. 

@msmttchr this means that the scope of this PR will probably increase, but we can work on this together if needed. Or we can discuss and flesh this out in an issue, and then make a new PR.

msmttchr:
> Thanks! I think it makes sense to take a look to clean up this part of the adapter API while we are changing this.
> Thoughts, @cjermain @CasperSchippers?
> 
> * As a certain generality w.r.t. adapters is one of the features of pymeasure, we should try to add implementations for other adapters, at least the SerialAdapters, the others as appropriate.
> * We should probably try to come up with some tests to make sure that the results match expectation, also across adapters.
> * for consistency with the ascii read/write/ask nomenclature, rename `binary_values` -> `ask_binary_values` ?
> * as we need some visa changes anyway (see #362), maybe we can get rid of the `VISAAdapter.config` use in `ask_values` and implement this in a way more aligned with the pyvisa API
> * VISAAdapter: `binary_values` could use `self.read_bytes`?
> * Check if our API language is still consistent w.r.t. read/write/ask, `binary` vs. nothing, what the `_values` suffix is supposed to mean/signify.
> 
> @msmttchr this means that the scope of this PR will probably increase, but we can work on this together if needed. Or we can discuss and flesh this out in an issue, and then make a new PR.

@bilderbuchi for the SerialAdapters, I am not 100% sure what they are. In our lab we have SCPI devices with serial interface, but they are supported by VISA with appropriate address (e.g: ASRLboard[::INSTR]). Looking at SerialAdapter class, I see SCPI mentioned, so if they are SCPI Serial interface instruments, I cannot figure out why they are not supported via VISA adapter.

In any case, I will be happy to collaborate in this as we have some other changes to propose in the pipe.

PR #368, opened 2020-11-19
Dev/temptronic

mroeleke:
please add temptronic device


mroeleke:
please merge to master

pymeasure/instruments/temptronic/temptronic.py
```diff
@@ -0,0 +1,1301 @@
...
+        bit 1 – request control -- not used
+        bit 0 – operation complete -- not implemented
+        NOTE: The above bits are latched, and are automatically cleared
+        when the register is read.
+        """,
+        validator=truncated_range,
+        values={'operation complete' : 1,   # not implemented
+                'request control' : 2,      # not used
+                'query error (qye)' : 4,
+                'device dependent error (dde)' : 8,
```
CasperSchippers, line 683: Are these the most convenient values (with the abbreviation included in them), especially if you also need to use them to set the value.

pymeasure/instruments/temptronic/temptronic.py
```diff
@@ -0,0 +1,1301 @@
...
+                'execution error (exe)' : 16,
+                'command error (cme)' : 32,
+                'user request' : 64,        # not used
+                'power on' : 128},          # not used
+        map_values=True
+        )
+
+    standard_event_status_mask = Instrument.control(
```
CasperSchippers, line 664: Could you comment on what precisely is the purpose of having this as a control, rather than a measurement?

pymeasure/instruments/temptronic/temptronic.py
```diff
@@ -0,0 +1,1301 @@
...
+        validator=truncated_range,
+        values={'operation complete' : 1,   # not implemented
+                'request control' : 2,      # not used
+                'query error (qye)' : 4,
+                'device dependent error (dde)' : 8,
+                'execution error (exe)' : 16,
+                'command error (cme)' : 32,
+                'user request' : 64,        # not used
+                'power on' : 128},          # not used
+        map_values=True
```
CasperSchippers, line 688: I'm wondering, is the value that is returned by the instrument actually one of the values in the dict (i.e 2, 4, 8, 16 etc) or is it a value between 0 and 255?

CasperSchippers (changes requested):
Thank you for the contribution. I've had a look at the PR and have added some remarks.

docs/api/instruments/temptronic/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.temptronic
+
+#########
+Temptronic
```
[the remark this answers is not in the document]
mroeleke: updated it

pymeasure/instruments/temptronic/temptronic.py
```diff
@@ -0,0 +1,1301 @@
...
+   
+    Methods
+    -------
+
+    Examples
+    -------
+    Instantate a temptronic thermostream, define a temperature force
+    operation and force 100 degC.
+        
+    >>> from pymeasure.instruments import TemptronicATS525
```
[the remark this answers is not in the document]
mroeleke: updated file

pymeasure/instruments/temptronic/temptronic.py
```diff
@@ -0,0 +1,1301 @@
...
+Instrument.setting = wait_for_cmd_exec(Instrument.setting)
+Instrument.control = wait_for_cmd_exec(Instrument.control)
+Instrument.measurement = wait_for_cmd_exec(Instrument.measurement)
+
+
+class TemptronicBase(Instrument):    
+    """Represent the TemptronicATSXXX instruments.
+   
+    Methods
+    -------
```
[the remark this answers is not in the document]
mroeleke: updated file

pymeasure/instruments/temptronic/temptronic.py
```diff
@@ -0,0 +1,1301 @@
...
+        setpoint forever during cycling if a NEXT command or MCT
+        interface “end of test” pulse is not received.
+        """,
+        validator=truncated_range,
+        values=[0, 9999]
+        )
+
+    dut_mode = Instrument.control(
+        "DUTM?", "DUTM %g",
+        """Turn DUT mode on or off.
```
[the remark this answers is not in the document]
mroeleke: updated file

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,879 @@
...
+        
+        5 = on Operator screen (manual mode)
+        6 = on Cycle screen (program mode)
+        """,
+        values={'manual' : 5, 
+                'program'  : 6},
+        map_values=True
+        )
+
+    def __init__(self, adapter, delay=0.02, **kwargs):
```
bilderbuchi, line 665: you are not using the `delay` argument?

pymeasure/instruments/temptronic/temptronic.py
```diff
@@ -0,0 +1,1301 @@
...
+        
+    >>> from pymeasure.instruments import TemptronicATS615
+    >>> with TemptronicATS615("ASRL6::INSTR") as ts:
+    >>>     ts.configure(temp_window=5, dut_type='T', soak_window=30)
+    >>>     ts.set_temperature(100)
+    >>>     ts.start()
+    >>>     ts.wait_for_settling()
+    >>>     # <do your stuff>
+    """
+    pass
```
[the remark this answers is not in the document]
mroeleke: updated file

pymeasure/instruments/temptronic/temptronic.py
```diff
@@ -0,0 +1,1301 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+"""
```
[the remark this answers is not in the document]
mroeleke: I left only the not implemented features in the docstring

pymeasure/instruments/temptronic/temptronic.py
```diff
@@ -0,0 +1,1301 @@
...
+        validator=truncated_range,
+        values=[0, 9999]
+        )
+
+    dut_mode = Instrument.control(
+        "DUTM?", "DUTM %g",
+        """Turn DUT mode on or off.
+
+        DUTM 0 -- off (air control)
+        DUTM 1 -- on (DUT control)
```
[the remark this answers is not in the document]
mroeleke: yes, that is true. I have changed it in a more meaningful way.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,879 @@
...
+    def at_temperature(self):
+        """ check if at temperature.
+        """
+        return self.temperature_condition_status_code & TemperatureStatusCode.AT_TEMPERATURE
+    
+    def __enter__(self):
+        return self
+
+    def __exit__(self):
+        self.stop()
```
bilderbuchi, line 879: This is a neat idea! I would prefer if you used/override the `shutdown()` method [defined](https://github.com/pymeasure/pymeasure/blob/e340f1c43dc302badbea46f060967199d508ec95/pymeasure/instruments/instrument.py#L303-L306) in `Instrument`.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,879 @@
...
+    @functools.wraps(func)
+    def wait_after_cmd(*args, **kwargs):
+        val = func(*args, **kwargs)
+        time.sleep(GLOBAL_DELAY_FOR_CMD_EXEC)
+        return val
+    return wait_after_cmd
+
+Instrument.setting = wait_for_cmd_exec(Instrument.setting)
+Instrument.control = wait_for_cmd_exec(Instrument.control)
+Instrument.measurement = wait_for_cmd_exec(Instrument.measurement)
```
bilderbuchi, line 82: I'm not sure if this does what you think it does. Right now, this should add 50 ms between each call when defining the Temptronic *class* -- the .setting etc.  methods are property factories that create and return a property to be attached to an Instrument subclass.
Are you maybe looking for the `query_delay` kwarg of the VISAAdapter/pyvisa?

pymeasure/instruments/temptronic/temptronic.py
mroeleke: Because if you want to set one bit more and you dont know whats currently set you could simply write:
self.standard_event_status_mask = self.standard_event_status_mask | <bit xyz> maybe
self.standard_event_status_mask |= <bit xyz> works as well.
or if you just want to unset one bit without touching the rest you write:
self.standard_event_status_mask &= ~<bit xyz>

pymeasure/instruments/temptronic/temptronic_base.py
mroeleke: you are right, it should be removed there.

pymeasure/instruments/temptronic/temptronic.py
mroeleke: updated file

pymeasure/instruments/temptronic/temptronic.py
mroeleke: updated file

pymeasure/instruments/temptronic/temptronic_base.py
mroeleke: I removed it now

pymeasure/instruments/temptronic/temptronic_base.py
mroeleke: yes exactly, in the manual is written that there need to be a delay between cmds of 50 ms. How would you enforce this delay for all possible adapters, even future ones?

pymeasure/instruments/temptronic/temptronic_base.py
mroeleke: I updated the .stop() method

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+                      f"status = {self.temperature_event_status}")
+                
+            if t > time_limit:
+                print('no settling achieved')
+                break
+
+    def shutdown(self):
+        """Brings the instrument to a safe and stable state"""
+        self.isShutdown = True
+        log.info("Shutting down %s" % self.name)
```
CasperSchippers, line 798: As this is code from the `shutdown` function of the inherited class, I would suggest replacing this by `super().shutdown()`.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+    def at_temperature(self):
+        """ check if at temperature.
+        """
+        return self.temperature_condition_status_code & TemperatureStatusCode.AT_TEMPERATURE
+    
+    def __enter__(self):
+        return self
+
+    def __exit__(self):
+        self.shutdown()
```
CasperSchippers, line 841: Very nice idea! @bilderbuchi: would it be a good idea to add this to the standard implementation of the `Instruments` class?

CasperSchippers (changes requested):

pymeasure/instruments/temptronic/temptronic_base.py
bilderbuchi: Yes, in principle it's nice to enable using instruments with a `with` statement. 
However, I think many users don't use instruments like that, but instead create an instrument, run the whole experiment/scripted measurement/whatever, and destroy the instrument at the end. In my experience this would mean nest the whole code into a `with` statement for every used instrument, which makes the code structure not very nice. 

Additionally, for most instruments the `shutdown` method is basically a no-op (and what a "safe" shutdown means might be specific to most setups/experiments), so we would not gain much of the safety enabled by with statements.

mroeleke: In this particular case I use the with statement to safeguard the measurement setup. During the unsupervised tests I set the thermostream to 225°C for a few seconds. If my program crashes during that time and nobody is around it will melt down my PCB and might start to burn setup. For me it's not jut nice to have but a safety feature for this piece of equipment.

bilderbuchi: Sure, it's well suited for that. You could also achieve the same safety with a `try` - `finally`, without needing this specific class feature, though.

bilderbuchi: Let's keep this here for now, if this pattern gets more usage we can always roll it out to the generic instrument class.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,826 @@
...
+                         f"status = {self.temperature_event_status}")
+
+            if t > time_limit:
+                log.info('no settling achieved')
+                break
+
+    def shutdown(self):
+        """Brings the instrument to a safe and stable state"""
+        super().isShutdown = True
+        log.info("Shutting down %s" % self.name)
```
bilderbuchi, line 783: These lines duplicate what is currently in `Instrument.shutdown()`. Better to do stuff specific to your instrument (the next 3 lines) first, then call ` super().shutdown()`. That should do the trick.

bilderbuchi (changes requested):

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+        time.sleep(GLOBAL_DELAY_FOR_CMD_EXEC)
+        return val
+    return wait_after_cmd
+
+Instrument.setting = wait_for_cmd_exec(Instrument.setting)
+Instrument.control = wait_for_cmd_exec(Instrument.control)
+Instrument.measurement = wait_for_cmd_exec(Instrument.measurement)
+
+
+class StandardEventStatusCode(Enum):
```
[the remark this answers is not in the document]
bilderbuchi: You're correct I think, but I think it should be fine to leave around, in case it's needed in another submodel, or in the future.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+        bit 0 – not used (always 0)
+        NOTE: See <serial poll> for the meaning of each bit in the mask.
+        Read the service request enable (mask) register.
+        """,
+        validator=truncated_range,
+        values=[0,255]
+        )
+
+    local_lockout = Instrument.setting(
+        "%LL",
```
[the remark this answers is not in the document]
bilderbuchi: I agree, this should not work correctly.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,826 @@
...
+        )
+
+    current_cycle_count = Instrument.measurement(
+        "CYCL?",
+        """Read the system’s current draw (amperage).
+        """,
+        )
+
+    error_code = Instrument.measurement(
+        "EROR?",
```
bilderbuchi, line 490: Is there a typo here, or is the instrument command really EROR, with 1 R?

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,826 @@
...
+        bit 6 -- not used
+        bit 5 -- cycling stopped ("stop on fail" signal was received)
+        bit 4 -- end of all cycles
+        bit 3 -- end of one cycle
+        bit 2 -- end of test (test time has elapsed)
+        bit 1 -- not at temperature
+        bit 0 -- at temperature (soak time has elapsed)
+        """,
+        validator=truncated_range,
+        values=[0,255]
```
bilderbuchi, line 313: If you use an `Enum.IntFlag` to store these status bits, and return such a value, you could save all the methods below that query all the possibilities, and get a readable status mask back. You can use the `get_process` and  `set_process` kwargs to define the interaction.

bilderbuchi: E.g. (sketched without checking this)
``` 
class TempStatus(IntFlag):
    at_temperature = 1
    not_at_temperature = 2
    # ...
```
and here use ` get_process= lambda v: TempStatus(v),`, then querying the status should get you ` <TempStatus.not_at_temperature: 2>`, and this behaves like an int, prints nicely, and importantly does all the bit mask resolving automatically, so if multiple flags match you can easily see that.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,826 @@
...
+from pymeasure.instruments.validators import (
+        strict_discrete_set,
+        truncated_range,
+        strict_range
+        )
+from enum import Enum
+import time
+
+
+class StandardEventStatusCode(Enum):
```
bilderbuchi, line 70: All those look like they should be `Enum.IntFlag` children, which would yield some nice benefits (see my remarks below)

bilderbuchi:
@mroeleke I tried to clear up some questions myself, but I could not locate a manual of these devices online. Is the vendor that strict around their pdfs? ^^

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+    AIR_SENSOR_OPEN                = 32     # bit 5  – air sensor open
+    LOW_INPUT_AIR_PRESSURE         = 16     # bit 4  – low input air pressure
+    LOW_FLOW                       = 8      # bit 3  – low flow
+                                            # bit 2  – reserved
+    AIR_OPEN_LOOP                  = 2      # bit 1  – air open loop
+    OVERHEAT                       = 1      # bit 0  – overheat
+
+
+class TemptronicBase(Instrument):    
+    """Represent the bas class for TemptronicATSXXX instruments.
```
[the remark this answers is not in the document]
bilderbuchi: I fixed that.

pymeasure/instruments/temptronic/temptronic_base.py
fjuenger: yes, no typo, that's correct, I verified again and put in a comment.

pymeasure/instruments/temptronic/temptronic_base.py
fjuenger: Switched to IntFlag and also introduced those get_processes for most relevant properties.

pymeasure/instruments/temptronic/temptronic_base.py
fjuenger: Switched to IntFlag and also introduced those get_processes for most relevant properties (like stated in the other thread).

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+        -------
+        none
+        
+        """
+
+        time.sleep(1)
+        t = 0
+        while( not self.at_temperature()): #assert at temperature
+            time.sleep(1)
+            t += 1
```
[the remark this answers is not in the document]
fjuenger: done

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+
+    def not_at_temperature(self):
+        """ check if not (yet) at temperature.
+        """
+        return self.temperature_condition_status_code & TemperatureStatusCode.NOT_AT_TEMPERATURE
+
+    def at_temperature(self):
+        """ check if at temperature.
+        """
+        return self.temperature_condition_status_code & TemperatureStatusCode.AT_TEMPERATURE
```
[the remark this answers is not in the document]
fjuenger: This was indeed not working anymore. Now with the IntFlag it does.

pymeasure/instruments/temptronic/temptronic_base.py
fjuenger: Incorporated like suggested :)

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+                                 # bit 6 -- not used
+    CYCLING_STOPPED        = 32  # bit 5 -- cycling stopped("stop on fail" signal was received)
+    END_OF_ALL_CYCLES      = 16  # bit 4 -- end of all cycles
+    END_OF_ONE_CYCLE       = 8   # bit 3 -- end of one cycle
+    END_OF_TEST            = 4   # bit 2 -- end of test (test time has elapsed)
+    NOT_AT_TEMPERATURE     = 2   # bit 1 -- not at temperature
+    AT_TEMPERATURE         = 1   # bit 0 -- at temperature (soak time has elapsed)
+
+
+class ErrorCode(Enum):
```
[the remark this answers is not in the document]
fjuenger: now it is in the related ` error_code` measurement

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+        )
+
+    enable_air_flow = Instrument.setting(
+        "FLOW %g",
+        """Turn the main nozzle air flow on or off.
+        enable_air_flow = True  (air flow on)
+        enable_air_flow = False (air flow off)
+        """,
+        validator=strict_discrete_set,
+        values={0,1,True,False}
```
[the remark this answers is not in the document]
fjuenger: It works like this due to `True`and `False` are also being converted to `0` and `1` via `%g`.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+        5 = on Operator screen (manual mode)
+        6 = on Cycle screen (program mode)
+        """,
+        values={'manual' : 5, 
+                'program'  : 6},
+        map_values=True
+        )
+
+    def __init__(self, adapter, **kwargs):
+        super().__init__(adapter, "Temptronic ATS-525 Thermostream", **kwargs)
```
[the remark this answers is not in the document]
fjuenger: Done

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,841 @@
...
+    copy_active_setup_file = Instrument.setting(
+        "CFIL %g",
+        """Local lockout – the System touch screen controls are disabled, and
+        no “Return to local” button appears on the panel.
+        """,
+        validator=strict_range,
+        values=[1, 12]
+        )
+
+    clear_errors = Instrument.setting(
```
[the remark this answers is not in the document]
fjuenger: I overlooked this previously, now with [this commit](https://github.com/BoschSensortec/pymeasure/commit/9f92b364308bffa6e508d1e686aad81ea79f3dd6) it will also be fixed.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -22,90 +22,103 @@
...
+    # bit 8  – reserved
+    INTERNAL_ERROR = 128    # bit 7  – internal error
+    # bit 6  – reserved
+    AIR_SENSOR_OPEN = 32     # bit 5  – air sensor open
+    LOW_INPUT_AIR_PRESSURE = 16     # bit 4  – low input air pressure
+    LOW_FLOW = 8      # bit 3  – low flow
+    # bit 2  – reserved
+    AIR_OPEN_LOOP = 2      # bit 1  – air open loop
+    OVERHEAT = 1      # bit 0  – overheat
+    HEAD = 0  # ok state
```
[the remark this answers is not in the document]
fjuenger: fixed

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,820 @@
...
+    def at_temperature(self):
+        """:returns: ``True`` if at temperature.
+        """
+        return self.temperature_condition_status_code == TemperatureStatusCode.AT_TEMPERATURE
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self):
+        self.shutdown()
```
bilderbuchi, line 820: These methods have [landed in master](https://github.com/pymeasure/pymeasure/blob/master/pymeasure/instruments/instrument.py#L165-L169) by now, so should be removed.

pymeasure/instruments/temptronic/temptronic_base.py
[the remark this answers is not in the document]
bilderbuchi: In pymeasure we generally prefer "boolean" properties to expose Bools only (to have a cleaner API), in this pattern: 
```
        map_values=True,
        values={True: 1, False: 0},
```

pymeasure/instruments/temptronic/temptronic_base.py
bilderbuchi: This has been added to `Instrument` by now.

pymeasure/instruments/temptronic/temptronic_base.py
bilderbuchi: funny :D
thanks!

bilderbuchi (changes requested):
Thank you! Two small things are left I think.

pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -0,0 +1,820 @@
...
+        "FLOW %g",
+        """Set TS air flow.
+
+        ``True`` enables air flow, ``False`` disables it
+
+        :type: bool
+
+        """,
+        validator=strict_discrete_set,
+        values={0, 1, True, False}
```
bilderbuchi, line 222: In pymeasure we generally prefer "boolean" properties to expose Bools only (to have a cleaner API), in this pattern: 
```
        map_values=True,
        values={True: 1, False: 0},
```
(I put this here, instead of in the other previously resolved conversation because that one seems not nicely visible because it's attached to an outdated source line)

bilderbuchi (changes requested):

bilderbuchi:

fjuenger:
Thanks @bilderbuchi for reviewing again and for the merge :)

PR #370, opened 2020-11-20
Fix file access issue

CasperSchippers:
PR to solve the issue raised in #235.


CasperSchippers:
I proposed a test and a fix for this issue. Could (one of) you have a look at these, @bilderbuchi  @cjermain?

Interestingly, with this test, it seemed to show that the issue arises only on windows. I am not sure if that is actually the case or if the test isn't adequate on Linux and mac.

msmttchr:
@CasperSchippers , I have tried your fix with Windows and it works fine for me. I believe that the issue is not visible in Linux since this OS use a different file locking mechanism compared to Windows (https://en.wikipedia.org/wiki/File_locking).

CasperSchippers:
It seems you're right about the locking mechanism.

msmttchr:
> It seems you're right about the locking mechanism.

Are you going to merge your PR ?

bilderbuchi:
@msmttchr Casper requested a review/another set of eyes to check this PR before merging.. :-)
In any case, looks good to me. Thanks for adding a test for the issue, too!

Issue #250, opened 2020-02-25
Dynamic validator limits


bilderbuchi:
I have a power supply with voltage/current limits that are changeable depending on a mode setting. I want to use a strict_range validator to check against those limits. What would be the best way to do this?
I had previously tried to do this, but did not find a way to get around the fact that the properties are attributes of the class, and I did not see how to make the values that are passed upon construction dynamically changeable. 

Did I miss something? Do we need a `dynamic_validator` that takes a (class) function or somesuch to track the state of the limits?

bilderbuchi:
See also https://github.com/ralph-group/pymeasure/pull/243/files/f903411e8e99a02c193ea4bf41d28da8eb890b97#diff-638e256c0011ed60431a146ca1a44939R333-R334 for another interesting use of validators

msmttchr:
Hi, I have question that may be related to this issue and I would like your comments, if possible.
Let's say I am implementing a family of devices which differs for their supported frequency range and I have a command that set/read the frequency parameter. The most obvious way to implement it is to have a generic class implementing an Instrument.control for frequency property with dummy frequency parameters for validator range.
I would then inherit from generic class and have a subclass for each instrument in the family which only specifies the frequency limits.
Being the Instrument.control a property and being the property a class attribute, I found no easy way to change the validator limit, except redefining the property all together for each subclass which clearly violates the DRY principle and it is very tedious. Any easy way to achieve what I need ?

bilderbuchi:
I have a suspicion, which I have been meaning to try out for quite some time, that this is in fact quite easy if you use a mutable class attribute, e.g. a tuple with lower, upper limits. Then you can define the class attribute as you propose in the base class, and all subclasses just need to redefine the _class attribute_ in their implementations.
Then I think it even is possible to dynamically change that with a method, because the class attribute gets then overruled by an _instance_ attribute, and the property _should_ still find that. Still have to check/test that, but haven't found the time yet.

See [here](https://github.com/pymeasure/pymeasure/pull/243/files/f903411e8e99a02c193ea4bf41d28da8eb890b97#diff-ed0a3f46fc16d6535c06bbba333a60a47c8a1b1699297f8808733b5dcbef3f6fR101-R117) for the `CLIST_VALUES`.

bilderbuchi:
I agree that redefining the property should be avoided for the reasons you give.

msmttchr:
I had a look at the example of CLIST_VALUES, but I think the property does not get changed.
[...]

PR #371, opened 2020-11-20
Implemented dynamic validator limits

CasperSchippers:
PR to implement the suggestion from #250.

A test is yet to be implemented to test the behaviour.


msmttchr:
I have written a similar proposal following @bilderbuchi suggestions in #250. Do you mind if I send this as PR to evaluate pros and cons? To be honest this an interesting problem, but the solutions are really a workaround more than a proper fix which I don't know whether it exists or not in python.

CasperSchippers:
Of course, feel free to submit a PR, then we can see what is a better solution.

I do admit that it indeed feels a bit as a hack, but couldn't yet think of a better solution within the existing framework.

pymeasure/instruments/instrument.py
```diff
@@ -140,43 +141,53 @@ def control(get_command, set_command, docs,
         :param check_get_errors: Toggles checking errors after getting
         """
 
+        inverse = None
         if map_values and isinstance(values, dict):
             # Prepare the inverse values for performance
             inverse = {v: k for k, v in values.items()}
 
         def fget(self):
+            values_local = values
```
bilderbuchi, line 150: for what purpose do you introduce these additional two variables? isn't this redundant?

pymeasure/instruments/instrument.py
```diff
@@ -140,43 +141,53 @@ def control(get_command, set_command, docs,
...
                     return value
-                elif isinstance(values, (list, tuple, range)):
-                    return values[int(value)]
-                elif isinstance(values, dict):
-                    return inverse[value]
+                elif isinstance(values_local, (list, tuple, range)):
+                    return values_local[int(value)]
+                elif isinstance(values_local, dict):
+                    if inverse_local is None:
+                        inverse_local = {v: k for k, v in values_local.items()}
```
bilderbuchi, line 165: it is weird to do this inverse creation twice (here and above) - I would rather prefer we adapt the logic condition above, or only do it here if needed, but not in both places.

pymeasure/instruments/instrument.py
```diff
@@ -140,43 +141,53 @@ def control(get_command, set_command, docs,
...
+                    return inverse_local[value]
                 else:
                     raise ValueError(
                         'Values of type `{}` are not allowed '
-                        'for Instrument.control'.format(type(values))
+                        'for Instrument.control'.format(type(inverse_local))
```
bilderbuchi, line 170: why the switch to inverse instead of values? It's the type of `values_local_ that gets checked in the ifs above.

bilderbuchi:
Thanks for the proposed implementation! 
I think selecting on the type of the `values` argument can be considered "pythonic", and given the nature of properties in Python I'm not sure if we can do better (but would be delighted to be proven false!).
Yeah, tests would be a good idea to add (ideally before doing the implementation ;-) ); also ensure we don't change the behaviour of generated properties in the non-dynamic cases with this fix.

bilderbuchi:
Whatever approach/PR we choose in the end, we should then also add documentation about this, and an example or two how to use this when creating instrument properties (similar to the [`preprocess_reply`](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#processing-of-return-values) feature that was recently added.

pymeasure/instruments/instrument.py
CasperSchippers: I thought so as well, but it seems python doesn't allow using the global variable `values` if there is a possibility that it is assigned locally (it starts complaining that it is referenced before assignment).

pymeasure/instruments/instrument.py
CasperSchippers: I would agree, but the reason for this duplicate is performance: if an actual dict is provided as values it is possible to define te inverse beforehand, which means it is only defined once. In case of a string, it has to be calculated every single time, because the values can change.

pymeasure/instruments/instrument.py
CasperSchippers: Oh good catch, that was a mistake indeed.

pymeasure/instruments/instrument.py
bilderbuchi: I suspected that. Probably this could be avoided if the "dynamic" lookup is handled in an additional else-clause, before the ValueError fallthrough. We would have to check if the structure will be simpler in the end, though, and confirm that the end result/behaviour is the same.

pymeasure/instruments/instrument.py
bilderbuchi: Agreed, makes sense.

pymeasure/instruments/instrument.py
bilderbuchi: I thought further on this, I don't see a way how we end up with a structure that is nice than what is already implemented.

CasperSchippers:
Closing in favour of #376

PR #374, opened 2020-11-22
Fixed incorrect skipping of 'test_correct_visa_kwarg' test

CasperSchippers:
Came accross the test "test_correct_visa_kwarg", which was skipped even though the pyvisa-test was actually installed. It seems that, as pyvisa-sim provides a library for the pyvisa package rather than an importable module, it is not recognisable by `importlib.util.find_spec('pyvisa-sim')`.

### DIFF
--- a/tests/adapters/test_visa.py
+++ b/tests/adapters/test_visa.py
@@ -29,7 +29,7 @@ from pytest import approx
 from pymeasure.adapters import VISAAdapter
 from pymeasure.instruments import Instrument
 
-pyvisa_sim_installed = bool(importlib.util.find_spec('pyvisa-sim'))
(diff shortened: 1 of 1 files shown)


msmttchr:
I think you have to use pyvisa_sim instead of pyvisa-sim. Isn't it ?
In my system:
importlib.util.find_spec('pyvisa_sim')
works

bilderbuchi:
Hm, weird. I could have sworn that I checked if that find_spec did work correctly. I agree with @msmttchr, though, importlib.util.find_spec('pyvisa_sim') does also work correctly on my system.

CasperSchippers:
Ahh, thanks, that is indeed a better way to fix it.

PR #375, opened 2020-11-23
Fixes, updates for averaging and scan mechanics on Agilent8722ES

alexmcnabb:
This instrument was previously not working but seems to now. I also changed how the scan detects when it has finished, this will let longer scans happen without timing out.

### DIFF
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -90,6 +90,9 @@ class Adapter(object):
                     # Need to cast to float first since results are usually
                     # strings and bool of a non-empty string is always True
                     results[i] = bool(float(result))
+                elif cast == int:
+                    # int() can't handle strings with decimals or scientific notation
+                    results[i] = int(float(result))
                 else:
                     results[i] = cast(result)
             except Exception:
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -90,9 +90,6 @@ class Adapter(object):
                     # Need to cast to float first since results are usually
                     # strings and bool of a non-empty string is always True
                     results[i] = bool(float(result))
-                elif cast == int:
-                    # int() can't handle strings with decimals or scientific notation
-                    results[i] = int(float(result))
                 else:
                     results[i] = cast(result)
             except Exception:
--- a/pymeasure/instruments/agilent/agilent8722ES.py
+++ b/pymeasure/instruments/agilent/agilent8722ES.py
@@ -23,6 +23,7 @@
 #
 
 from pymeasure.instruments import Instrument, discreteTruncate, RangeException
+from pyvisa import VisaIOError
 
 import numpy as np
 import re
@@ -57,6 +58,13 @@ class Agilent8722ES(Instrument):
         in seconds. This property can be set.
         """
     )
+    averages = Instrument.control(
+        "AVERFACT?", "AVERFACT%d",
+        """ An integer representing the number of averages to take. Note that
+        averaging must be enabled for this to take effect. This property can be set.
+        """,
+        get_process=int
+    )
 
     def __init__(self, resourceName, **kwargs):
         super(Agilent8722ES, self).__init__(
@@ -73,6 +81,7 @@ class Agilent8722ES(Instrument):
 
     @property
     def parameter(self):
+        """The S parameter to measure, one of S11, S12, S21, or S22"""
         for parameter in Agilent8722ES.SCATTERING_PARAMETERS:
             if int(self.values("%s?" % parameter)) == 1:
                 return parameter
@@ -80,6 +89,7 @@ class Agilent8722ES(Instrument):
 
     @parameter.setter
     def parameter(self, value):
+        """The S parameter to measure, one of S11, S12, S21, or S22"""
         if value in Agilent8722ES.SCATTERING_PARAMETERS:
             self.write("%s" % value)
         else:
@@ -121,48 +131,51 @@ class Agilent8722ES(Instrument):
                                  "8722ES exceeded")
 
     def set_averaging(self, averages):
-        """ Turns on averaging of a specific number between 0 and 999
-        """
-        if int(averages) > 999 or int(averages) < 0:
-            raise RangeException("Averaging must be in the range 0 to 999")
+        """Sets the number of averages and enables/disables averaging"""
+        averages = int(averages)
+        if not 1 <= averages <= 999:
+            assert RangeException("Invalid number of averages -", averages)
+        self.averages = averages
+        if averages > 1:
+            self.enable_averaging()
         else:
-            self.write("AVERO1")
-            self.write("AVERFACT%d" % int(averages))
+            self.disable_averaging()
 
     def disable_averaging(self):
-        """ Disables averaging """
+        """Disables averaging"""
         self.write("AVERO0")
+
+    def enable_averaging(self):
+        """Enables averaging"""
+        self.write("AVERO1")
 
     def is_averaging(self):
         """ Returns True if averaging is enabled """
         return self.ask("AVERO?") == '1\n'
 
-    def restart_averaging(self, averages):
-        if int(averages) > 999 or int(averages) < 0:
-            raise RangeException("Averaging must be in the range 0 to 999")
-        else:
-            self.write("NUMG%d" % averages)
-
-    def scan(self, averages=1, blocking=True, timeout=25, delay=0.1):
+    def scan(self):
         """ Initiates a scan with the number of averages specified and
-        blocks until the operation is complete if blocking is True
+        blocks until the operation is complete
         """
-        if averages == 1:
-            self.disableAveraging()
-            self.setSingleSweep()
-        else:
-            self.setAveraging(averages)
-            self.write("*CLS;SRE 4;ESNB 1;")
-            self.restartAveraging(averages)
-            if blocking:
-                self.adapter.wait_for_srq(timeout, delay)
-
-    def is_scan_complete():
-        pass  # TODO: Implement method for determining if the scan is completed
+        self.write("*CLS")
+        self.scan_single()
+        # All queries will block until the scan is done, so use NOOP? to check.
+        # These queries will time out after several seconds though,
+        # so query repeatedly until the scan finishes.
+        while True:
+            try:
+                self.ask("NOOP?")
+                break
+            except VisaIOError as e:
+                if e.abbreviation != "VI_ERROR_TMO":
+                    raise e
 
     def scan_single(self):
         """ Initiates a single scan """
-        self.write("SING")
+        if self.is_averaging():
+            self.write("NUMG%d" % self.averages)
+        else:
+            self.write("SING")
 
     def scan_continuous(self):
         """ Initiates a continuous scan """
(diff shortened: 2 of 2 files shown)


pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -57,6 +58,13 @@ class Agilent8722ES(Instrument):
         in seconds. This property can be set.
         """
     )
+    averages = Instrument.control(
+        "AVERFACT?", "AVERFACT%d",
+        """ An integer representing the number of averages to take. Note that
+        averaging must be enabled for this to take effect. This property can be set.
+        """,
+        get_process=int
```
CasperSchippers, line 66: I think it is nice to use `cast=int`.

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -121,48 +131,51 @@ def set_IF_bandwidth(self, bandwidth):
                                  "8722ES exceeded")
 
     def set_averaging(self, averages):
-        """ Turns on averaging of a specific number between 0 and 999
-        """
-        if int(averages) > 999 or int(averages) < 0:
-            raise RangeException("Averaging must be in the range 0 to 999")
+        """Sets the number of averages and enables/disables averaging"""
```
CasperSchippers, line 134: It would be nice to indicate the range of valid values in the docstring.

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -121,48 +131,51 @@ def set_IF_bandwidth(self, bandwidth):
...
 
+    def enable_averaging(self):
+        """Enables averaging"""
+        self.write("AVERO1")
+
     def is_averaging(self):
         """ Returns True if averaging is enabled """
         return self.ask("AVERO?") == '1\n'
 
-    def restart_averaging(self, averages):
```
CasperSchippers, line 140: I would suggest leaving the existing functions and properties (I did not mark the other occasions this remark would hold for) in existence (albeit as a wrapper for a new, updated function), with the same outcome / functionality as before, such that people using this instrument can keep relying on the functionality as they used to do.

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -179,31 +192,32 @@ def frequencies(self):
...
 
-    def phase(self, real, imaginary):
-        """ Returns the phase in degrees from a real and imaginary
-        number or numpy arrays
+    @property
+    def data_phase(self):
+        """ Returns the phase in degrees from the last scan
         """
-        return np.arctan2(imaginary, real)*180/np.pi
+        return np.angle(self.data_complex())*180/np.pi
```
CasperSchippers, line 223: Although this is fine, you could also use `np.degrees()` to convert from radians to degrees.

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -121,48 +131,51 @@ def set_IF_bandwidth(self, bandwidth):
...
-        pass  # TODO: Implement method for determining if the scan is completed
+        self.write("*CLS")
+        self.scan_single()
+        # All queries will block until the scan is done, so use NOOP? to check.
+        # These queries will time out after several seconds though,
+        # so query repeatedly until the scan finishes.
+        while True:
+            try:
+                self.ask("NOOP?")
+                break
```
CasperSchippers, line 168: I would suggest placing the `break` in an `else` statement after the `except` statement.

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -179,31 +192,32 @@ def frequencies(self):
...
 
-    def log_magnitude(self, real, imaginary):
-        """ Returns the magnitude in dB from a real and imaginary
-        number or numpy arrays
+    @property
+    def data_log_magnitude(self):
+        """ Returns the magnitude in dB from the last scan
         """
-        return 20*np.log10(self.magnitude(real, imaginary))
+        return 20*np.log10(self.data_magnitude())
```
CasperSchippers, line 211: As `self.data_magnitude` is a property, I think you don't need to call it as `self.data_magnitude()` but can just use `self.data_magnitude`.

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -179,31 +192,32 @@ def frequencies(self):
...
 
-    def magnitude(self, real, imaginary):
-        """ Returns the magnitude from a real and imaginary
-        number or numpy arrays
+    @property
+    def data_magnitude(self):
+        """ Returns the absolute magnitude from the last scan
         """
-        return np.sqrt(real**2 + imaginary**2)
+        return np.sqrt(np.abs(self.data_complex()))
```
CasperSchippers, line 217: As `self.data_complex` is a property, I think you don't need to call it as `self.data_complex()` but can just use `self.data_complex`.

pymeasure/instruments/agilent/agilent8722ES.py
CasperSchippers, line 223: As `self.data_complex` is a property, I think you don't need to call it as `self.data_complex()` but can just use `self.data_complex`.

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -121,48 +131,51 @@ def set_IF_bandwidth(self, bandwidth):
...
+            assert RangeException("Invalid number of averages -", averages)
+        self.averages = averages
+        if averages > 1:
+            self.enable_averaging()
         else:
-            self.write("AVERO1")
-            self.write("AVERFACT%d" % int(averages))
+            self.disable_averaging()
 
     def disable_averaging(self):
```
CasperSchippers, line 144: You could consider making a `Instrument.control` for this, such that you could use `Agilent8722ES.averaging = True`, and have `disable_averaging` and `enable_averaging` as a wrapper for this control.

CasperSchippers (changes requested):
Thanks for this addition. I've had a look at the PR and added some remarks.

pymeasure/instruments/agilent/agilent8722ES.py
bilderbuchi: I would also like some tests to cover this change for all 4 quadrants/constellations of values, to make sure this does not change the behaviour.

pymeasure/instruments/agilent/agilent8722ES.py
bilderbuchi: I agree!

pymeasure/instruments/agilent/agilent8722ES.py
bilderbuchi: Agreed, we should not suddenly remove functionality. In methods, you can emit a warning about this with an appropriate message instructing the user how to avoid the warning: `warnings.warn(message, FutureWarning)`.

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -179,31 +192,32 @@ def frequencies(self):
         )
 
     @property
-    def data(self):
-        """ Returns the real and imaginary data from the last scan
+    def data_complex(self):
```
bilderbuchi, line 195: This rename seems it would just arbitrarily break existing code using this. What is the benefit, as the docstring already made clear the purpose of `data`?

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -179,31 +192,32 @@ def frequencies(self):
...
         """
-        return np.sqrt(real**2 + imaginary**2)
+        return np.sqrt(np.abs(self.data_complex()))
 
-    def phase(self, real, imaginary):
-        """ Returns the phase in degrees from a real and imaginary
-        number or numpy arrays
+    @property
+    def data_phase(self):
+        """ Returns the phase in degrees from the last scan
```
bilderbuchi, line 221: I'm not sure what aim these restructurings and renamings follow. While conversion to properties, where possible, is surely idiomatic for pymeasure, you should not break users' code by removing methods. A preferable approach would be to implement a new property, have the old method use/wrap around that, and also emit a Futurewarning (as sketched above) telling users about the new, improved way to use the instrument. :-)
After some waiting period, we can remove the deprecated methods.

bilderbuchi (changes requested):

pymeasure/instruments/agilent/agilent8722ES.py
alexmcnabb: I opted to refactor things a little more aggressively as it seems the module was already broken. There's some functions that were renamed back in April 2015 with commit 0c4597ba, and at least the scan function has been non-functional since then.

Should I go back and make sure things are as compatible as possible with the existing functions, or do you think it doesn't matter too much? I should get a chance to look at the other stuff soon.

alexmcnabb:
I opted to refactor things a little more aggressively as it seems the module was already broken. There's some functions that were renamed back in April 2015 with commit 0c4597b, and at least the scan function has been non-functional since then.

Should I go back and make sure things are as compatible as possible with the existing functions, or do you think it doesn't matter too much? I should get a chance to look at the other stuff soon.

CasperSchippers:
I see your point. Nevertheless I still think it would be nice to (for the time being at least) keep the existing functions alive; you never know if someone is using any of these functions as they are right now, even though some of them might be broken.

pymeasure/instruments/agilent/agilent8722ES.py
```diff
@@ -178,10 +205,42 @@ def frequencies(self):
...
+        """
+        return np.degrees(np.angle(self.data_complex))
+    
     @property
     def data(self):
         """ Returns the real and imaginary data from the last scan
         """
+        warnings.warn("Don't use this function, use data_complex instead", FutureWarning)
```
bilderbuchi, line 243: Thank you! Now, to round this out, change the deprecated properties/methods to use the *new* ones (if appropriate, i.e. the same result can be computed). 
This way, you avoid duplicating code (which later maybe only gets fixed in one place, not the other.
For example, here you could just say
``` 
res = self.data_complex
return res.real, res.complex
```
if I'm not deceived.

bilderbuchi (changes requested):

pymeasure/adapters/adapter.py
```diff
@@ -90,6 +90,9 @@ def values(self, command, separator=',', cast=float, preprocess_reply=None):
                     # Need to cast to float first since results are usually
                     # strings and bool of a non-empty string is always True
                     results[i] = bool(float(result))
+                elif cast == int:
+                    # int() can't handle strings with decimals or scientific notation
```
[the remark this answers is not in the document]
alexmcnabb: It seems like the device returns the numbers in scientific notation even though they should be integers, for example I'm seeing 9.990000000000000E+02 when I expect 999, and 4.000000000000000E+00 when I expect 4 for the averages parameter. I could solve this without making changes to the adapter class by making a custom cast function though, `cast=lambda x: int(float(x))` Should I do that?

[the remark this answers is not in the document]
bilderbuchi: Yes, that would be a better approach. :-)

bilderbuchi:
Thank you for your work!
@CasperSchippers, this looks good to me now. If you agree, feel free to merge, probably best to squash-and-merge to get rid of the bit of back-and-forth in the commits.

bilderbuchi (approved):

CasperSchippers:
Indeed thanks for the work! I'll go ahead and merge it.

CasperSchippers (approved):

PR #369, opened 2020-11-19
Fixes for Agilent B1500 instrument

moritzj29:
Hi,

these are two fixes for the Agilent B1500 instrument:

- fixing the ability to select fixed measurement ranges (negative indizes), addressing #366 
- (code cleanup and restructuring of ranging classes)
- fixing FMT11 formatting

PR #378, opened 2020-11-27
[Agilent B1500] Data Formats and Documentation

moritzj29:
Follow up on #369 

Fixing:

- allow to connect to instrument when not implemented data format is set
- setting data_format() afterwards is now mandatory instead of just recommended -> added to example code
- fix sampling example in documentation
- fix issue with FMT1 data format throwing an exception when set

I don't know why the ranging stuff is also shown again in this PR, this is already merged from this same branch.


CasperSchippers (approved):
I've had a look, but as @bilderbuchi mentioned last time it is nearly impossible to say something useful about this complex instrument.

In order for me to merge this, could you solve the existing conflict (I would think that either of the options in the conflict is fine, but it would be nice if you could confirm this).
Moreover, it could be useful (especially if you continue on this branch as you did since the previous PR) to again merge the master branch into this branch; then the PR will perform the tests again.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -427,8 +423,8 @@ def format_channel_check_status(self, status_string, channel_string):
     class _data_formatting_FMT1(_data_formatting_generic):
         """ Data formatting for FMT1 format
         """
-        def __init__(self, smu_names={}):
-            super().__init__("FMT1", smu_names)
+        def __init__(self, smu_names={}, output_format_string="FMT1"):
```
bilderbuchi, line 426: If nothing opposes this, maybe switch `smu_names`and `output_format_string` parameter order, so that it matches the superclass's?

moritzj29: good point, I switched the arguments of the super class and dropped the defaults there. For the derived classes it is desirable to have `smu_names` as first argument, since `output_format_string` is only used when deriving additional classes and can be omitted in most cases.

moritzj29:
> merge the master branch

done! conflicts resolved!

PR #379, opened 2020-11-29
Fix open file externally

msmttchr:
Added support for Windows in open_file_externally

This is done using start from a shell which is equivalent to 'double click' on the file.
Linux behavior (unchanged) is to unconditionally call gedit with file.

### DIFF
--- a/pymeasure/display/windows.py
+++ b/pymeasure/display/windows.py
@@ -389,6 +389,10 @@ class ManagedWindow(QtGui.QMainWindow):
             experiment.curve.setPen(pg.mkPen(color=color, width=2))
 
     def open_file_externally(self, filename):
+        """ Method to open the datafile using an external editor or viewer. Uses the default
+        application to open a datafile of this filetype, but can be overridden by the child
+        class in order to open the file in another application of choice.
+        """
         system = platform.system()
         if (system == 'Windows'):
             # The empty argument after the start is needed to be able to cope correctly with filenames with spaces
@@ -764,6 +768,10 @@ class ManagedImageWindow(QtGui.QMainWindow):
             experiment.curve.setPen(pg.mkPen(color=color, width=2))
 
     def open_file_externally(self, filename):
+        """ Method to open the datafile using an external editor or viewer. Uses the default
+        application to open a datafile of this filetype, but can be overridden by the child
+        class in order to open the file in another application of choice.
+        """
         system = platform.system()
         if (system == 'Windows'):
             # The empty argument after the start is needed to be able to cope correctly with filenames with spaces
--- a/pymeasure/display/windows.py
+++ b/pymeasure/display/windows.py
@@ -25,6 +25,9 @@
 import logging
 
 import os
+import subprocess, platform
+
+
 import pyqtgraph as pg
 
 from .browser import BrowserItem
@@ -148,7 +151,6 @@ class ManagedWindow(QtGui.QMainWindow):
 
 
     """
-    EDITOR = 'gedit'
 
     def __init__(self, procedure_class, inputs=(), displays=(), x_axis=None, y_axis=None,
                  log_channel='', log_level=logging.INFO, parent=None, sequencer=False,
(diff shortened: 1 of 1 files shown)


CasperSchippers (commented):
Thanks for this fix. Looks like a nice way to fix this. I have two questions:

- pyMeasure is also checked against python on mac, so it would be nice to include a case for mac. I think that `proc = subprocess.Popen(['open', '', filename], shell=True)` should work (based on what I could find on opening a file on mac), but I have no mac to test this.
- This makes the `EDITOR` attribute obsolete for windows. Would it be a useful idea to include a check whether `EDITOR` has been overwritten if people want to change their editor, or just remove the `EDITOR` attribute altogether?

ps. I did get an error when trying to use 'open file externally' stating that the file could not be opened because it was in use by another process (i.e. it was still opened by pymeasure); this seems to be solved by #370.

msmttchr:
Unfortunately, I don't have Mac available so I am not sure how to test it. Possibly, we need to leave the extension to support Mac to some people with that PC.
Regarding the `EDITOR `attribute for Windows, this is indeed not needed, at least for the solution I have proposed.
In fact my solution is equivalent to double click to a file and so, if Windows does not know the extension, it will ask user to select the application, but if it does, it will use the associated application without allowing the user to make a different choice. In the latter case, your proposal to check the `EDITOR `variable could be good.

Also, for some reason, in Windows, if I open the file with `emacs`, I do not have any issue, but if I open a `.csv` like in the `image_gui.py` example automatically trigger Excel in my PC and I will have the issue you mentioned (I didn't notice before).

A bit confused, I will try to make a proposal to use the `EDITOR `attribute as well

CasperSchippers:
Having `editor` as an argument for `__init__` would indeed be the nicest solution. (We might have to be careful not to break the current behaviour with the `EDITOR` attribute which people might set; probably raising a FutureWarning would be in place if people do overwrite the attribute)

You're right in that we have to leave it to someone with a Mac to implement the feature for Mac; might be nice to raise a NotImplementedError in that case.

bilderbuchi:
I think on Linux, `xdg-open` is the correct command to use
> xdg-open - opens a file or URL in the user's preferred application
Furthermore, if you are on a KDE system, `gedit` is not even installed, which would fail hard, then, so removing the self.EDITOR field seems like the way to go.

If we assume that people have defined the correct application to use for a given filetype, does it make sense to make this user-selectable in pymeasure, too?
It would also be great to reduce the amount of code duplication, but maybe that's a larger undertaking in this part of the code, right @CasperSchippers ?

On Mac, `open` [seems to be](https://apple.stackexchange.com/a/25855) the way to go, but I have no way to test. Maybe just do some research on if that is still correct, and then just go with it and wait for potential bug reports instead of breaking it with NotImplementedError.

On Windows, I agree with using `start`, I think this is the way to go (although I don't know why the extra `''` is there in the argument list)

Watch out, however, for spaces in the paths, and make sure this still works if your paths have spaces. Typically, you have to surround the path with quotes to make that work - maybe there is a Python Path utility to do that, but I don't know.

msmttchr:
@bilderbuchi, the empty string in argument list is needed to be able to cope with filename with spaces.
For some weird reason, the windows command `start` behaves differently whether the filename has spaces in it or not
(see https://superuser.com/questions/239565/can-i-use-the-start-command-with-spaces-in-the-path).
The solution is to pass an argument with empty string which is always interpreted as the title of the window.
I have checked the commands with filename with and without spaces and it looks good to me.

CasperSchippers:
> If we assume that people have defined the correct application to use for a given filetype, does it make sense to make this user-selectable in pymeasure, too?

Though I mostly agree with this reasoning, I'd still favour allowing people to define an editor in pymeasure, since I could imagine that people who use for example the .csv format for multiple purposes would want to open the pymeasure output in a text editor (or maybe something of an analysis tool they built themself), but would want most other .csv files to be opened in excel.

> On Mac, `open` [seems to be](https://apple.stackexchange.com/a/25855) the way to go, but I have no way to test. Maybe just do some research on if that is still correct, and then just go with it and wait for potential bug reports instead of breaking it with NotImplementedError.

I'm fine with this; as far as I can find, `open` is still the way to go indeed.

msmttchr:
> Though I mostly agree with this reasoning, I'd still favour allowing people to define an editor in pymeasure, since I could imagine that people who use for example the .csv format for multiple purposes would want to open the pymeasure output in a text editor (or maybe something of an analysis tool they built themself), but would want most other .csv files to be opened in excel.
What about considering to redefine `open_file_externally` when user are not happy with implementation provided ?

CasperSchippers:
> @CasperSchippers: what about considering to redefine `open_file_externally` method in subclasses when user is not happy with implementation provided ?

Yeah, that may indeed be is the best approach. Let's go for that.

bilderbuchi:
Another way to inject a custom editor command, without having to go through init, would be to define a module-level constant, e.g. `EDITOR` in windows.py, which is by default None, but users can easily override it, e.g. by 
```
pymeasure.display.windows.EDITOR = 'gedit'
# all the rest o the user code
```
Then, if defined with a value, it gets used instead of the default command when opening files. 
I'll let you evaluate if that makes more sense than letting users override the whole method, I am not that steeped in the proper usage of this part of pymeasure.

CasperSchippers (approved):
I don't think there is a sufficiently big advantage of defining a module level variable for this purpose, so I'd say let's keep it at the override-option. I'll add this as a remark to the docstring of the function and then I'll merge it.

PR #380, opened 2020-12-03
adding Tektronix TDS620B Oscilloscope and agilent 33250A AWG

neal-kepler:
I subclassed the agilent 33220A and just updated the frequency range. The TDS620B is a copy pasta from the TDS2000 with a commands to control the CURVE acquisition and transfer implemented now implemented.

### DIFF
--- a/docs/api/instruments/agilent/agilent33250A.rst
+++ b/docs/api/instruments/agilent/agilent33250A.rst
@@ -1,0 +1,7 @@
+###########################################
+Agilent 33250A Arbitrary Waveform Generator
+###########################################
+
+.. autoclass:: pymeasure.instruments.agilent.Agilent33250A
+    :members:
+    :show-inheritance:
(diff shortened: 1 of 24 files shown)


msmttchr (commented):
For AWG, I think using inheritance is good approach, especially if the instruments belong to same family. I see one minor problem that, in the inherited instrument, the name will remain "Agilent 33220A Arbitrary Waveform generator". It could possibly make sense to define a generic family class which defines all the common features and includes `name` parameter in `__init__`. The subclasses, one for each instrument, will define the correct name in `__init__`. In addition they will specify only the specific properties that are different.
A more structured approach would include a generic AWG instrument class from which an Agilent 322xxx generic family is inheriting from, which would finally be subclassed by specific Agilent AWG instruments. Here I would also highlight that the availability of dynamic property as proposed in #376 will avoid repetition in the inherited classes allowing to change, for instance, the frequency range by simply defining an attribute like `frequency_values=[1e-6,20e6]` or `frequency_values=[1e-6,80e6]` instead of redefining the complete property violating the DRY principle.

bilderbuchi:
Hm, something most have gone awry with your PR, there are lots of changes in there that are unrelated to what you describe - did something go wrong when branching off or did you merge some strange state?

neal-kepler:
Oh dang, yes, some crazy stuff went down, let me close this and get more organized

Issue #265, opened 2020-03-29
ManagedImageWindow inherit from ManagedWindow


cjermain:
The `ManagedImageWindow` in #157 does not inherit from the `ManagedWindow` class, which leads to some code redundancy and lacking support for the sequencer. This issue tracks moving both classes onto the same base class, or inheriting from the `ManagedWindow` class.

msmttchr:
Fixed with PR #383

PR #383, opened 2020-12-27
Fix managed window

msmttchr:
Address Issue #265, avoid code duplication and make the ManagedWindow class more flexible while retaining the original behaviour.
The changes are preliminary and intended to trigger feedback especially from the people more experienced with the implementation.


pymeasure/display/widgets.py
```diff
@@ -138,21 +138,45 @@ def change_y_axis(self, axis):
...
             if isinstance(item, ResultsCurve):
                 item.y = axis
-                item.update()
+                item.update_data()
         label, units = self.parse_axis(axis)
         self.plot.setLabel('left', label, units=units, **self.LABEL_STYLE)
         self.y_axis = axis
         self.y_axis_changed.emit(axis)
 
+class TabWidget(object):
```
CasperSchippers, line 147: I think it is a good solution to have this base-class for TabWidgets!

pymeasure/display/windows.py
```diff
@@ -389,19 +400,24 @@ def open_experiment(self):
                 else:
                     results = Results.load(filename)
                     experiment = self.new_experiment(results)
-                    experiment.curve.update()
+                    for index in range(len(self.widget_list)):
```
CasperSchippers, line 403: Why not just iterate over `experiment.curve`?

pymeasure/display/windows.py
```diff
@@ -414,16 +430,27 @@ def make_procedure(self):
...
         return self.inputs.get_procedure()
 
-    def new_curve(self, results, color=None, **kwargs):
+    def new_curve(self, wdg, results, color=None, **kwargs):
         if color is None:
             color = pg.intColor(self.browser.topLevelItemCount() % 8)
-        return self.plot_widget.new_curve(results, color=color, **kwargs)
+        return wdg.new_curve(results, color=color, **kwargs)
+
+    def new_experiment(self, results, curve=[]):
```
CasperSchippers, line 438: I think in principle it is safer to have `None` as a default argument rather than a mutable (like `[]`); if anything is added to this mutable, the default value for next calls of this function will be changed. (In this case I do not see this problem arise, but just as a precaution).

pymeasure/display/windows.py
```diff
@@ -557,360 +577,81 @@ class ManagedImageWindow(QtGui.QMainWindow):
...
-        self.plot_widget = PlotWidget(self.procedure_class.DATA_COLUMNS, self.x_axis, self.y_axis)
-        self.im_plot = self.image_widget.plot
+    def __init__(self, procedure_class, inputs=(), displays=(), x_axis=None, y_axis=None,
+                 log_channel='', log_level=logging.INFO, parent=None, sequencer=False,
+                 sequencer_inputs=None, sequence_file=None, inputs_in_scrollarea=False, directory_input=False, wdg_list=()):
+        self.x_axis = x_axis
+        self.y_axis = y_axis
+        self.log_widget = LogWidget("Experiment Log")
+        self.plot_widget = PlotWidget("Results Graph", procedure_class.DATA_COLUMNS, self.x_axis, self.y_axis)
         self.plot = self.plot_widget.plot
```
CasperSchippers, line 589: I think that this line is not necessary anymore (as you removed all usage of `.plot` with the `widget_list`.

pymeasure/display/windows.py
```diff
@@ -465,19 +492,19 @@ def queue(self):
         raise NotImplementedError(
             "Abstract method ManagedWindow.queue not implemented")
 
-    def setup_plot(self, plot):
-        """
-        This method does nothing by default, but can be overridden by the child
-        class in order to set up custom options for the plot
+    # def setup_plot(self, plot):
```
CasperSchippers, line 495: I can imagine that people use it to format their graph differently (e.g. set a log-scale). However, I can also imagine that with changing from a single `PlotWidget` to the `widget_list` another implementation is required anyway.

pymeasure/display/windows.py
```diff
@@ -389,19 +400,24 @@ def open_experiment(self):
...
+                        experiment.curve[index].update_data()
                     experiment.browser_item.progressbar.setValue(100.)
                     self.manager.load(experiment)
                     log.info('Opened data file %s' % filename)
 
     def change_color(self, experiment):
+        for index, wdg in enumerate(self.widget_list):
+            if isinstance(wdg, PlotWidget):
+                curve = experiment.curve[index]
+                break
```
CasperSchippers, line 413: Now only for a single `PlotWidget` the color is changed; maybe it would be nice to do this for any `PlotWidget` that is present in the `widget_list` (or maybe even nicer if it would be possible to do this for any tab-widget for which it is relevant, though this might be a bit more difficult as there are so far no other tab-widgets for which it is relevant).

CasperSchippers (commented):
Thanks for this (very nice) work (and sorry for the slow response).

I think it is a good idea to have this ManagedWindowsBase class; it makes for a nice standarized way of buidling ManagedWindows and also allows for more flexibility for both users and future additions.

The same basically holds for the TabWidget and the widget_list. Here I do have to remark that I am not yet certain whether to including the LogWidget into the widget_list and making it a child of TabWidget is the best way to go; you could argue that you in all cases need this LogWidget and that it is sufficiently different from the other widgets (in the sense that it has nothing to do with the data of the experiments whereas the PlotWidget and the ImageWidget, and any future additional Widgets I think, do depend on the data of the experiments) to make it separate from the other tab-widgets.

pymeasure/display/manager.py
```diff
@@ -252,7 +255,9 @@ def _finish(self):
         experiment = self._running_experiment
         self._clean_up()
         experiment.browser_item.setProgress(100.)
-        experiment.curve.update()
+        for curve in experiment.curve:
```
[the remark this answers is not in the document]
msmttchr: I agree

pymeasure/display/manager.py
```diff
@@ -170,7 +170,9 @@ def _update_log(self, record):
     def load(self, experiment):
         """ Load a previously executed Experiment
         """
-        self.plot.addItem(experiment.curve)
+        for wdg in self.widget_list:
+            wdg.load(experiment.curve)
```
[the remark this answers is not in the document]
msmttchr: Indeed you are right, thanks for the suggestion, I'll do that

pymeasure/display/windows.py
msmttchr: I agree

pymeasure/display/windows.py
msmttchr: Thanks for the head up, I did not realize this potential side effect which is very hard to debug. I will change to None

pymeasure/display/windows.py
msmttchr: If we allow specialization of `TabWidget` then I think the `setup_plot` does not make too much sense. We will break backward compatibility in the interface. In my view the `TabWidget` is something generic, a view of experiment results. It can a plot, it can be a table, it can be anything else, so again the `setup_plot`, if maintained, can be called something like `setup_widget` or similar.

pymeasure/display/windows.py
msmttchr: Ok, I think you are right. I will add a `set_color` method to the `TabWidget` and add the implementation for `PlotWidget`.

pymeasure/display/windows.py
msmttchr: I agree

pymeasure/display/windows.py
```diff
@@ -169,16 +180,12 @@ def __init__(self, procedure_class, inputs=(), displays=(), x_axis=None, y_axis=
...
         self.log.setLevel(log_level)
-        self.x_axis, self.y_axis = x_axis, y_axis
+        self.widget_list = widget_list
         self._setup_ui()
         self._layout()
-        self.setup_plot(self.plot)
+        #??? self.setup_plot(self.plot)
 
     def _setup_ui(self):
-        self.log_widget = LogWidget()
```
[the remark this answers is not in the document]
msmttchr: Well, here I see the fact that the `LogWidget` is something special, but I am not 100% sure it is necessary to make an exception in the code for it. If people will subclass `ManagedWindows` will always have the `LogWidget`. We could also imagine that when saving an experiment, all the widgets including `LogWidget` will be saved, so again another reason to make it as a `TabWidget` subclass (here my thought is to add a method, in the future, to TabWidget to save experiment logging for example).

pymeasure/display/widgets.py
```diff
@@ -146,9 +146,9 @@ def change_y_axis(self, axis):
 
 class TabWidget(object):
     """ Utility class to define default implementation for some basic methods """
-    def __init__(self, name, index):
+    def __init__(self, name, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
CasperSchippers, line 150: Could you comment on why you allow passing `*args` and `**kwargs` to the `__init__` of `object`; if I'm not mistaken the `object` class doesn't allow passing any arguments.

pymeasure/display/windows.py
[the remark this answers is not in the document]
CasperSchippers: Yeah, that logic makes sense to me.

pymeasure/display/widgets.py
```diff
@@ -233,6 +254,15 @@ def update_y_column(self, index):
...
+        self.plot.addItem(curve)
+
+    def remove(self, curve):
+        self.plot.removeItem(curve)
+
+    def set_color(self, curve, color):
+        """ Remove curve from widget """
+        curve.setPen(pg.mkPen(color=color, width=2))
 
 class ImageFrame(QtGui.QFrame):
```
CasperSchippers, line 267: It seems there is quite a bit of overlap with the `PlotFrame` class; would it be possible to inherit from that class?

pymeasure/display/widgets.py
```diff
@@ -138,21 +138,42 @@ def change_y_axis(self, axis):
...
+                item.update_data()
         label, units = self.parse_axis(axis)
         self.plot.setLabel('left', label, units=units, **self.LABEL_STYLE)
         self.y_axis = axis
         self.y_axis_changed.emit(axis)
 
+class TabWidget(object):
+    """ Utility class to define default implementation for some basic methods """
+    def __init__(self, name, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
CasperSchippers, line 150: Could you comment on why you allow passing `*args` and `**kwargs` to the `__init__` of `object`; if I'm not mistaken the `object` class doesn't allow passing any arguments.

pymeasure/display/windows.py
```diff
@@ -430,16 +441,27 @@ def make_procedure(self):
...
             color = pg.intColor(self.browser.topLevelItemCount() % 8)
-        return self.plot_widget.new_curve(results, color=color, **kwargs)
+        return wdg.new_curve(results, color=color, **kwargs)
 
     def new_experiment(self, results, curve=None):
-        if curve is None:
-            curve = self.new_curve(results)
-        browser_item = BrowserItem(results, curve)
-        return Experiment(results, curve, browser_item)
+        if curve == None:
```
CasperSchippers, line 450: To check whether `curve` equals `None`, a `curve is None` is a better comparison, as there always will be only a single `None` object (and for some specific, maybe very rare, cases the `==` comparison can result in unexpected results)

pymeasure/display/widgets.py
msmttchr: @CasperShippers, here I followed some suggestions from https://rhettinger.wordpress.com/2011/05/26/super-considered-super/ to cope with problem of multiple inheritance and different `__init__` signature. In particular I have used the approach of https://stackoverflow.com/a/25352819 comment and https://github.com/brentpayne/learning-python/blob/master/MixInMultipleInheritance/mixin_multiple_inheritance.py
In any case, I agree with you that `object` class should receive no parameters, but possibly when using `super` for multiple inheritance passing arguments maybe "allowed".

CasperSchippers: Ahh yes, that makes sense.

pymeasure/display/windows.py
msmttchr: OK, thanks for the tip. I will do that

pymeasure/display/widgets.py
msmttchr: Indeed you are right. I made the changes in my commit

pymeasure/display/windows.py
msmttchr: Done

CasperSchippers (commented):
Nice that you've also made ImageFrame inherit from PlotFrame

CasperSchippers (approved):
I've had another look and have no more remarks.
Thank you very much for this nice work!

PR #384, opened 2021-01-02
Fix for incorrect reading of parameters

CasperSchippers:
This is a fix for two bugs that occur when reading the results back into pymeasure:

- a `BooleanParameter` get incorrectly interpreted (always as `True`, independent of the actual value, since the bool-value of a non-empty string is `True`)
- a `Parameter` that contains a space gets cut-off at this space (since the space is also used to separate values and units).


tests/experiment/data/results_for_testing_parameters.csv
```diff
@@ -0,0 +1,212 @@
+#Procedure: <test_results.test_parameter_reading.<locals>.DummyProcedure>
+#Parameters:
+#	Directory string: //test directory with space/test_filename.csv
```
bilderbuchi, line 3: It's probably not material to this test, but this path starting with `//` looks really strange. :-)

pymeasure/experiment/results.py
```diff
@@ -269,7 +269,7 @@ def units_found(parameter, units):
             if parameter.name in parameters:
                 value, units = parameters[parameter.name]
 
-                if units_found(parameter, units):
+                if units_found(parameter, units) or (type(parameter) == Parameter and units is not None):
```
CasperSchippers, line 272: In my fix for the string with space bug I simply extended the if-statement that was already there. I wonder though what the purpose was for the original if-statement, since I couldn't find a case where this was True, nor could think of a case where this should be True. Do you have an idea why it was there @cjermain (a variant of this seems to have been there since your first commit)?

tests/experiment/data/results_for_testing_parameters.csv
CasperSchippers: Yeah, you're right, it's a bit strange but not material; I'll change it anyway.

pymeasure/experiment/results.py
bilderbuchi, line 272: `procedure.parameter_objects()` returns a dict of Parameters, so is it actually possible that `parameter` is not a `Parameter`? Also, typically it's preferable to use `isinstance`  to check object type - or are you intending to check for the base class only?

pymeasure/experiment/parameters.py
```diff
@@ -138,6 +138,12 @@ def value(self):
 
     @value.setter
     def value(self, value):
+        if isinstance(value, str):
+            if value.lower() in "true":
```
bilderbuchi, line 142: Should compare with `==`?

pymeasure/experiment/parameters.py
```diff
@@ -138,6 +138,12 @@ def value(self):
...
+                value = True
+            elif value.lower() == "false":
+                value = False
+
```
bilderbuchi, line 146: Shouldn't there be an `else` clause that errors out? What kind of `value` strings do you want to accept here? I think as implemented (case-insensitive "true"/"false") is reasonable. As it is, i think a value of `"0"` or `"blalblubb"` would be stored as `True`, which is probably not intended.

pymeasure/experiment/parameters.py
CasperSchippers: oh yeah, good one.

pymeasure/experiment/results.py
CasperSchippers: Yes, I intend to check for a base class only (hence it is indeed possible to have a `parameter` that is not a `Parameter` but rather one of the subclasses) since I think only in the `Parameter` base class it makes sense to have a space somewhere in the value.

pymeasure/experiment/parameters.py
CasperSchippers: I didn't implement the `else` clause because I thought to leave as much of the original behavior as possible; there is specifically a test (tests\experiment\test_parameters.py::test_boolean_value) that tests for a `BooleanParameter` accepting a string ('a' in this case) and expects it to result in `True`.
I do agree that it makes no sense for the `BooleanParameter` to accept other strings, so do you think I should modify this test-case?

pymeasure/experiment/results.py
bilderbuchi: >  parameter that is not a Parameter but rather one of the subclasses

Hold on, though, a `BooleanParameter`, being a subclass, *is a* `Parameter` (see behaviour of `isinstance()`)

A behaviour that is valid for a base class, but not a subclass, looks like a code smell to me. I realise this might be hard to change now, but we should try to not compound/rely on that too much if possible.

pymeasure/experiment/parameters.py
bilderbuchi: hmm, I lean towards changing the expected/tested behaviour. Can you find out who implemented that test? (probably @cjermain  :D)

bilderbuchi: > leave as much of the original behavior as possible

good call on that decision, by the way! In principle the right way to go. As we are still < 1.0 (but around for quite some time), maybe we should start collecting behaviour that we want to change/break before 1.0, and have one release later that implements all the breaks in one go, with good documentation.

pymeasure/experiment/results.py
CasperSchippers: I completely agree on that, but I couldn't think of a better approach.
An alternative would by to replace `type(parameter) == Parameter` by `not hasattr(parameter, 'units')` such that value and units are combined only when the `Parameter` has not units; in that case the full string is also passed to `BooleanParameter`, though you might expect that this never has any consequence as a string for the `BooleanParameter` should in principle always be `False` or `True`.

Another way would be to always pass the full string (so probably not even separate the value and unit from the string) and only perform the separation within the Parameter subclasses where it makes sense.

pymeasure/experiment/parameters.py
CasperSchippers: > Can you find out who implemented that test?

Indeed, as far is I can find it was implemented by @cjermain, so maybe he can comment on this behavior.

pymeasure/experiment/results.py
bilderbuchi: puh, no idea to be honest :-P

CasperSchippers:
I decided to move the interpretation of the strings (when read from a file) to the setter of the `value` in the respective `Parameter` classes, meaning that always the full string is passed to the `Parameter` and there the unit (if a unit is present and if applicable) is removed from the string.

Hendrick42:
Hi @CasperSchippers, is it possible to split the PR with no big effort?
In my oppinion the "Parameter fix" could be merged right now. 
The Boolean part can be left open for discussion? (For me, Boolean should be only t/True f/False

CasperSchippers:
> Hi @CasperSchippers, is it possible to split the PR with no big effort?
It's a bit annoying to split the PR, so I'll leave it as a single PR

> The Boolean part can be left open for discussion? (For me, Boolean should be only t/True f/False
Do you mean that it now also interprets numbers as a boolean? In that case I partly agree: I do not see the use of accepting values other than 0 or 1 as a boolean, but I do think that 0 and 1 should still be accepted.
I added a suggestion to incorporate this, though I do would like to hear from @cjermain or @bilderbuchi if they agree, since there was a test-case that checked whether another number would accepted (but I don't know the idea behind that test).

bilderbuchi:
The change to only accepting 0 and 1 is fine with me.

Hendrick42:
> Do you mean that it now also interprets numbers as a boolean? In that case I partly agree: I do not see the use of accepting values other than 0 or 1 as a boolean, but I do think that 0 and 1 should still be accepted.

t/True f/False and 0 1 are fine for backwards compatibility

cjermain:
Looks good. Thanks @CasperSchippers!

PR #388, opened 2021-01-05
Update instrument.py

AdityaOP:
To read bytes from the instrument read_bytes is necessary. So I added read_bytetes method in instrument.py file.
Thank You,
Aditya.

### DIFF
--- a/pymeasure/instruments/instrument.py
+++ b/pymeasure/instruments/instrument.py
@@ -103,6 +103,12 @@ class Instrument(object):
         response.
         """
         return self.adapter.read()
+    
+    def read_bytes(self):
+        """ Read bytes from the instrument through the adapter and returns the
+        response.
+        """
+        return self.adapter.read_bytes(size = 100)
 
     def values(self, command, **kwargs):
         """ Reads a set of values from the instrument through the adapter,


msmttchr:
@AdityaOP, is this only applicable to visa adapter ? What is the use of this method ?
Have you checked `values` and `binary_values` method to read data from instrument ?

AdityaOP:
As only visa adapter has that method, its only applicable to visa adapter. This method returns bytes which are stored in Instrument. I try to read bytes using values and binary_values but it did not work as values returns list and binary values return Numpy Array.

bilderbuchi:
Then as it is now this will not work because using other adapters this will throw an error when used.
Is it possible for you to use `instrument.adapter.read_bytes` in your instrument implementation, instead?

AdityaOP:
@bilderbuchi Yes, I try to use instrument.adapter.read_bytes in my program and I can directly use it. Thank you. But I think in Instrument.py only visa adapter is imported.

moritzj29:
for refrence: this was already discussed in #226 and #227 
`self.adapter.read_bytes()` works well e.g. for the Agilent B1500 instrument

bilderbuchi:
Yes, please let's go with `self.adapter.read_bytes()` for now. If we were to include this in `Instrument`, all the other adapters would need to gain the same method to keep a common API.

PR #391, opened 2021-01-17
Pre v0.9 patches and PyVISA updates

cjermain:
This PR introduces minor patches before the v0.9 release, and improves the VISA compatibility. It addresses some of the issues raised in #362, and leverages code from #363 (thanks @msmttchr!). Further work will be needed to wrap up the compatibility. The `pyvisa-sim` library is added as a test requirement, which allows tests to be written against the instruments. This has a lot of potential for improving test coverage of our instruments.


cjermain:
It looks like the `pyvisa-sim` tests are being skipped because the package is not installed. In the CI, we are no longer running the tests through `setup.py test`, so those `tests_require` packages are not being installed. Running that way would ensure all `test_requires` packages are installed.

https://github.com/pymeasure/pymeasure/blob/4e2119eb5263e9f8f2e423567b3e031724cc35a6/.github/workflows/pymeasure_CI.yml#L82-L89

bilderbuchi:
Great to see this! 
Yes, we are using Anaconda in CI. `python setup.py test` would install the requirements through pip, not conda, which we want to avoid.
There is a small snag in that pyvisa-sim is not available on conda-forge, but I think you can add
```
- pip:
  - pyvisa-sim == your-version
```
to the dependencies section of https://github.com/pymeasure/pymeasure/blob/master/requirements/pymeasure.yml to get it installed.

Huh, nice, I thought you need to create a (simulated) instrument definition for every instrument you use with pyvisa-sim.

msmttchr:
@cjermain, I am not very familiar with the process, but I guess it makes sense to close #363 in favour of this, right ?

cjermain:
Thanks @bilderbuchi for the suggestion. I've made some changes to the CI to use a base and dev requirements YAML file.

@msmttchr, I haven't included all of the changes you made in #363, so I think the broader changes are still worth discussing.

.github/workflows/pymeasure_CI.yml
```diff
@@ -64,26 +82,15 @@ jobs:
...
+        run: |
+          conda info
+          conda list
       - name: Run pytest with xvfb
-        shell: bash -l {0}
         if: runner.os == 'Linux'
-        run: xvfb-run -a pytest
-      - name: Run pytest 
-        shell: bash -l {0}
+        run: xvfb-run -a python setup.py test
```
bilderbuchi, line 93: Fine with me to change to this way of building & testing in one go.

.github/workflows/pymeasure_CI.yml
```diff
@@ -64,26 +82,15 @@ jobs:
...
-        shell: bash -l {0}
-        run: conda list
-      - name: Build pymeasure
-        shell: bash -l {0}
-        run: python setup.py sdist bdist_wheel
-      - name: Install pymeasure
-        shell: bash -l {0}
-        run: python -m pip install dist/*.whl
+      - name: Install dev requirements
+        run: conda env update -f requirements/pymeasure_dev.yml
```
bilderbuchi, line 86: Note that this now always pulls sphinx and flake8 dependencies into every testing CI job, even though we don't use them at all in those jobs.

tests/adapters/test_visa.py
```diff
@@ -29,15 +29,42 @@
...
-    instr = Instrument(adapter='ASRL1::INSTR', name='delayed', query_delay=0.5, visa_library='@sim')
+    instr = Instrument(adapter=SIM_RESOURCE, name='delayed', query_delay=0.5, visa_library='@sim')
     assert instr.adapter.connection.query_delay == approx(0.5)
+
+
+def test_visa_adapter():
+    adapter = make_visa_adapter()
+    assert repr(adapter) == "<VISAAdapter(resource='ASRL1::INSTR')>"
+
+    adapter.ask("?IDN") == "LSG Serial #1234\n"
```
bilderbuchi, line 57: This lacks an `assert` in the beginning I think.

tests/adapters/test_visa.py
```diff
@@ -29,15 +29,42 @@
...
+
+    adapter.write("?IDN")
```
bilderbuchi, line 59: Funny that this works correctly, I thought that the [proper SCPI command is `*IDN?`](http://na.support.keysight.com/vna/help/latest/Programming/GP-IB_Command_Finder/Common_Commands.htm#idn)?

bilderbuchi (commented):
I went over the recent CI config changes and left some comments/questions. 
I now realise that I probably should have left more comments in the CI files documenting my reasoning for the implement configuration. Sorry about that!

requirements/pymeasure_dev.yml
```diff
@@ -0,0 +1,11 @@
+channels:
+  - conda-forge
+dependencies:
+  - pytest-qt=3.3.0
+  - pytest-runner=5.2
+  - pytest=6.1.1
+  - flake8=3.8.4
+  - sphinx=3.2.1
+  - sphinx_rtd_theme=0.4.3
+  - pip:
```
bilderbuchi, line 10: [Apparently](https://github.com/pymeasure/pymeasure/pull/391/checks?check_run_id=1766645369#step:5:6) pip nowadays also has to be added as an explicit dependency.

tests/adapters/test_visa.py
msmttchr: I agree, this is completely confusing. For some reason, the first default emulated instrument in pyvisa-sim is a non SCPI fictional instrument with commands very similar to SCPI (see pyvisa-sim/pyvisa_sim/default.yaml). I guess the intent is to demonstrate that you can emulate different kind of instruments, not just SCPI, but the result is questionable in my opinion

bilderbuchi:
I'll try to fix some of the issues I raised as soon as I get an opportunity.

cjermain:
Thanks @bilderbuchi! I'm finally getting back to this now and see your updates. I'll go through the comments first.

.github/workflows/pymeasure_CI.yml
```diff
@@ -10,16 +10,32 @@ jobs:
   docs_lint:
     name: Docs and Linting
     runs-on: "ubuntu-latest"
+    defaults:
+      run:
+        shell: bash -l {0}
```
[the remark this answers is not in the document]
cjermain: Yea, I saw that by reading the miniconda action docs:
https://github.com/conda-incubator/setup-miniconda#example-4-conda-options

tests/adapters/test_visa.py
bilderbuchi: I've changed this to use another pyvisa-sim default instrument that uses scpi-compatible commands.

.github/workflows/pymeasure_CI.yml
```diff
@@ -10,16 +10,32 @@ jobs:
...
+      - name: Install pymeasure requirements
+        uses: conda-incubator/setup-miniconda@v2
         with:
           python-version: "3.8"
-      - name: Install docs dependencies
+          environment-file: requirements/pymeasure.yml
+          activate-environment: pymeasure
+          auto-update-conda: true
+          use-only-tar-bz2: true  # for caching to work properly
+      - name: Install dev requirements
```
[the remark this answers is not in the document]
cjermain: I was trying to avoid the duplication of requirements in multiple places. That allows PyMeasure users to only install the main environment to get the basic packages necessary. However, I agree that this requires additional time during install for the CI pieces. I'm open to combining them and just giving PyMeasure users everything if they want to use that environment. What do you think @bilderbuchi?

Good catch also on the pip installs over the conda ones. We should fix that either way.

.github/workflows/pymeasure_CI.yml
```diff
@@ -10,16 +10,32 @@ jobs:
...
+        run: conda env update -f requirements/pymeasure_dev.yml
+      - name: Conda info & list
         run: |
-          pip install wheel
-          pip install -r requirements/pymeasure_readthedocs.txt
```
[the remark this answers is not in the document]
cjermain: Good point -- we are using pip for RTD builds. Historically we used conda because it was difficult to pip install the Qt requirements. I wonder if that is still a limitation -- if not we could move away from conda. Otherwise, I think we should consider moving RTD to conda. If dependencies are missing, the doc pages can be dropped, so it is important to have all requirements in that job.

I like the mamba suggestion, and will look into it more.

.github/workflows/pymeasure_CI.yml
cjermain: I was thinking we can centralize on the cache, since those packages should also get wrapped up and then shared with the RTD test. I agree it is an extra install here, but I'm not sure how optimized we need to be.

tests/adapters/test_visa.py
cjermain: Good catch!

requirements/pymeasure_dev.yml
cjermain: Thanks for adding that. I saw the funny error message, but did not finish adding it.

tests/adapters/test_visa.py
```diff
@@ -29,15 +29,43 @@
...
+
+
+def test_visa_adapter_ask_values():
+    adapter = make_visa_adapter()
+    assert adapter.ask_values(":VOLT:IMM:AMPL?", separator=",") == [1.0]
+
+
+def test_visa_adapter_write_binary_values():
+    adapter = make_visa_adapter()
+    adapter.write_binary_values("OUTP", [1], datatype='B')
```
bilderbuchi, line 71: I think this test also needs to assert something. I played around with what I think you wanted to do, but could not get it to work - does pyvisa-sim work correctly with `write_binary_values`?

bilderbuchi:
Sure thing! I'm about done for today (as soon as I see that CI is green), so feel free to iterate now!

.github/workflows/pymeasure_CI.yml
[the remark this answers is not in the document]
bilderbuchi: I guess that people who install from git master (and thus using/needing the requirements.yml) will probably don't mind a couple of packages more, or will even be more inclined to run the tests because they are troubleshooting something on the bleeding edge?

My hope is that if we give all in one go/one env file, the resolution also happens in one go and it does not install pip packages over conda ones, but we'll have to try (locally?) and check the logs/output.

tests/adapters/test_visa.py
cjermain: Good question. I was looking for a way to see what is written. We could use `mocker` from `pytest-mock` to ensure that the right call is being made. Ideally `pyvisa-sim` has a place for that already, but I haven't found it yet.

.github/workflows/pymeasure_CI.yml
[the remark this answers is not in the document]
bilderbuchi: IDK how flexible RTD is regarding builds - never used it. If we can move that to conda, why not. I guess you can check a recent (or master-branch) CI log if there were any difficulties with pip & qt.

I would very much like to keep conda, I have virtually stopped using pip years ago, for scientific python + Windows, conda just makes so much sense.

.github/workflows/pymeasure_CI.yml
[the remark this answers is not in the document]
bilderbuchi: +1 on reducing duplication

.github/workflows/pymeasure_CI.yml
bilderbuchi: Good question what is more efficient - smaller/fewer cache or faster install. I also though in terms of complexity of moving parts, would be good to keep docs & linting separate from code testing, e.g. sphinx dependency hiccups or whatnot would not fail/error our pytest tests.

.github/workflows/pymeasure_CI.yml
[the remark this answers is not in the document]
cjermain: Yea, sticking with conda sounds good. I think mamba is something to keep an eye on and maybe try later this year.

.github/workflows/pymeasure_CI.yml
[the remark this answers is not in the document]
cjermain: Sounds good. I merged them into 1 environment file and moved RTD to use that file as well.

.github/workflows/pymeasure_CI.yml
bilderbuchi: Do you have any insight how we stand on using our CI quota?
 [AFAICT](https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions) we have 2000 "actions minutes" per month, where linux workers count 1x, Win 2x, Macos 10x. One CI run apparently takes ~15 min Linux time, 20 Macos, 24 Win.
If I am correct, and this could become a problem, from that point of view we should probably a) optimise for runtime and b) probably reduce our build matrix, e.g. only test with latest python on Win and Mac, and run tests on Linux wherever possible.

tests/adapters/test_visa.py
cjermain: I agree that we should improve this test, but am thinking to leave it as a future improvement. There is a lot we can do with `pyvisa-sim`, but I don't want to go too deep in this PR.

.github/workflows/pymeasure_CI.yml
bilderbuchi: The thing is, my gut says that we would have easily blown through our quota this month if the above numbers are correct, so something seems strange.

bilderbuchi: Ahhhh, "free for public repositories", so maybe there is no quota. Still, considering how Gitlab just slashed *their* free minutes from 2000 to 400 (or so), I think it would be good to be "efficient" citizens. 
I'm not sure how much we learn anyhow from the additional Win and Mac builds at lower python versions (as long as we test with the latest) - Python compat problems should be caught in the Linux builds, anyway.

cjermain:
I started a RTD run for this branch:
https://readthedocs.org/projects/pymeasure/builds/12902015/

.github/workflows/pymeasure_CI.yml
cjermain: I believe that is only on private repos. :) For the pymeasure Org it shows that we have not used any of the 2k actions minutes yet.

cjermain:
RTD looks good using the conda environment. I am going to go ahead with these changes to close this out.

.github/workflows/pymeasure_CI.yml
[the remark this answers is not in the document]
bilderbuchi: Mamba would be interesting, and I would be fine to switch. I meant conda as opposed to pip, mentally mamba counts as conda for me :D

Issue #382, opened 2020-12-22
Qthread destroyed error


neal-kepler:
I'm not sure if this has been raised before ( I seem to recall it but I can't find the issue). But, I not infrequently get `QThread: Destroyed while thread is still running` errors which cause the software to crash. I looked through the code and it looks like the only QThreads are the monitors. _EDIT The following statement is unrelated and affects an older version of pymeasure:_ In addition I frequently end up with a bunch of stale python processes and I think that is related (like filling up 2 Gb of RAM of stale processes) _End EDIT_. So to get around this I added a flag to Monitor, `self.stop` which can break out of the run loop if set to True.

```
class Monitor(QtCore.QThread):
    """ Monitor listens for status and progress messages
    from a Worker through a queue to ensure no messages
    are losts
    """

    status = QtCore.QSignal(int)
    progress = QtCore.QSignal(float)
    log = QtCore.QSignal(object)
    worker_running = QtCore.QSignal()
    worker_failed = QtCore.QSignal()
    worker_finished = QtCore.QSignal()  # Distinguished from QThread.finished
    worker_abort_returned = QtCore.QSignal()

    def __init__(self, queue):
        super().__init__()
        self.queue = queue
        self.stop = False

    def run(self):
        while not self.stop:
            self.data = self.queue.get()
            if self.data is None:
                break
            topic, data = self.data
            if topic == 'status':
                self.status.emit(data)
                if data == Procedure.RUNNING:
                    self.worker_running.emit()
                elif data == Procedure.FAILED:
                    self.worker_failed.emit()
                elif data == Procedure.FINISHED:
                    self.worker_finished.emit()
                elif data == Procedure.ABORTED:
                    self.worker_abort_returned.emit()
            elif topic == 'progress':
                self.progress.emit(data)
            elif topic == 'log':
                self.log.emit(data)

        log.info("Monitor caught stop command")
```

Then where the Manager instance cleans up the self._monitor instance, I set the flag, give it a chance to stop (`self.monitor.wait(100)`), record if it finishes properly in wait for posterity, and then just terminate.  (I know it's dangerous, but I don't know why the QThread isn't exiting and I can't have my code crashing) :

```
    def _clean_up(self):
        self._worker.join()
        self._monitor.stop = True
        success = self._monitor.wait(100)
        if not success:
            raise ValueError('Monitor did not properly exit')
        else:
            self._monitor.terminate()
        del self._worker
        del self._monitor
        self._worker = None
[...]

PR #392, opened 2021-01-22
Fix for issue #382

msmttchr:
Fix for #382  thanks to @neal-kepler who discovered the root cause of the problem.


bilderbuchi:
Thank you!

PR #394, opened 2021-01-28
Adapters: Adding __close__ magic where it is still missing (VISA, vxi11)

bklebel:
Some instruments allow only a limited number of connections at once - while some connections might be broken (without consciously closing them) in a way the instrument can notice, others might not be. 
Communication via VISA over a TCP-IP connection is such a case. For example, the sr860 LockIn only allows a certain number of active tcp connection - if this is exceeded because of repeatedly connecting and, without closing, breaking the connection, at some point a new connection cannot be established. 

This PR introduces the `__del__` method for those adapters where it was not yet included (i.e. visa, and vxi11), on the basis of the corresponding frameworks.  
Additionally, just for the looks, it groups the "magic" methods of the adapter classes to the bottom, as is most common across pymeasure (currently), while I could not find a mention of importance of such an order in the PEP8 style guide.

### DIFF
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -39,6 +39,11 @@ class Adapter(object):
     """
     def __init__(self, preprocess_reply=None, **kwargs):
         self.preprocess_reply = preprocess_reply
+        self.connection = None
+
+    def __del__(self):
+        """close connection upon garbage collection of the device"""
+        self.connection.close()
 
     def write(self, command):
         """ Writes a command to the instrument
--- a/pymeasure/adapters/serial.py
+++ b/pymeasure/adapters/serial.py
@@ -50,11 +50,6 @@ class SerialAdapter(Adapter):
         else:
             self.connection = serial.Serial(port, **kwargs)
 
-    def __del__(self):
-        """ Ensures the connection is closed upon deletion
-        """
-        self.connection.close()
-
     def write(self, command):
         """ Writes a command to the instrument
 
--- a/pymeasure/adapters/telnet.py
+++ b/pymeasure/adapters/telnet.py
@@ -54,11 +54,6 @@ class TelnetAdapter(Adapter):
                     f"allowed are: {str(safe_keywords)}")
         self.connection = telnetlib.Telnet(host, port, **kwargs)
 
-    def __del__(self):
-        """ Ensures the connection is closed upon deletion
-        """
-        self.connection.close()
-
     def write(self, command):
         """ Writes a command to the instrument
 
@@ -89,3 +84,7 @@ class TelnetAdapter(Adapter):
     def __repr__(self):
         return "<TelnetAdapter(host=%s, port=%d)>" % (self.connection.host, self.connection.port)
 
+    def __del__(self):
+        """ Ensures the connection is closed upon deletion
+        """
+        self.connection.close()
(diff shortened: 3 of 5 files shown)


bilderbuchi:
Thanks for this. In principle I agree with this change. However, have you tested that this actually makes a difference with your instrument and fixes your issue? I'm asking because both vxi11 and visa already have a `__del__` method in their respective resources objects ( https://github.com/python-ivi/python-vxi11/blob/cc4671da699f1f379137dc40ffc4a302d72e6f55/vxi11/vxi11.py#L584 and https://github.com/pyvisa/pyvisa/blob/bb8fd9d21b1efa1f311e12402e21292a656a0e6a/pyvisa/resources/resource.py#L197), so if you destroy the Instrument/adapter objects, these close() should already be called when the python garbage collector kicks in.

Also, as all adapter classes have this method if we go ahead with your change, I think it makes most sense to pull this up into the `Adapter` base class, to avoid repetition of these identical methods.

bklebel:
yes, I have tested it, actually I made the change precisely because of the issue I faced: the stanford sr860, which we recently introduced to pymeasure, only accepts 5 tcp socket connections at a time, and closing this connections does not seem to work by itself. In this case, when testing the device, and most of all testing my code, repeated restarts of my program (including crashing it with Ctrl+C) lead to the case that I could not connect to the device anymore, until I did a restart of the device itself by turning off and on its power. This change of explicitly closing the connection resolved this issue. 
To give complete information: We are running the sr860 over eth using visa with the NI drivers. 

Possibly when the `Adapter` class is removed, not all its attributes are explicitly removed too? I don't know how the python garbage collection handles this. Maybe it would work with sth like the following too:
```    
    def __del__(self):
        """close connection upon device deletion"""
        del self.connection
```
I did not test this with vxi11, since we do not run any devices through it, however the documentation does list this command, so I did not see any harm to "complete the set". 

I agree, if all the adapters have it, even if it is only for good measure, it would make sense to put it into the `Adapter` base class. 
If there is general agreement on that, I will do those changes.

bilderbuchi:
> repeated restarts of my program

just to be sure that I understand correctly, you are not running this in a console (that lives throughout the multiple restarts), e.g. via Spyder or Ipython? You are either using command line [i]python or an IDE setting that creates a fresh console/interpreter for every script run?

I'm not an expert on garbage collection, but would assume all objects get collected at the end, at least if the process shuts down cleanly. 
(As a side note, keep in mind that the `__del__` magic method is empathically *not* a C++-like "destructor", there are generally no guarantees when it runs)

Yes, put the method in the base class (and remove from all children). Keep the way of using `.close()`, not `del`. I confirmed that vxi11 has the relevant method, too. Use the wording "on garbage collection" in the docstring, though, it is a little clearer.

bklebel:
I run this from both from windows consoles (which live throughout multiple restarts), as well as quite clunky as windows services through nssm ("non-sucking service manager") - there, new consoles are started every time. This setup is designed so that all devices collect data continuously, and stream them to a central storage place. 
My main use of pymeasure is the instrument drivers, I do not use any other part of pymeasure - my setup is a rather distributed approach, you can look it up here [CryoGUI](https://github.com/Cryostat-GUI/Cryostat-GUI/tree/zmq_architecture) and here [measureSequences](https://github.com/bklebel/measureSequences). 

As I currently have a few problem with running the python scripts as windows services in way I can shut them down cleanly, from time to time, I have to outright kill the process which runs a python console. With this PR in place, I can re-establish connections to the sr860, for now however I closed it, although I did not test this behaviour thoroughly.

bilderbuchi:
LGTM, thanks!

PR #395, opened 2021-01-28
[new instrument] Keithley 2260B DC Power Supply (minimal implementation)

bklebel:
Introduce a minimal implementation of the Keithley 2260B dc power supply family. This instrument comes in different versions for different available powers, which leads to more individual possible settings not explored in this minimal PR. 

This PR implements setting current, voltage, output on/off, as well as reading those values, and reading the applied power. 
These functions are the same across the whole 2260B family, although ranges, and rated current output levels differ.


pymeasure/instruments/keithley/keithley2260B.py
```diff
@@ -0,0 +1,71 @@
...
+
+    applied = Instrument.measurement(
+        ":APPly?",
+        """ Reads the applied voltage (volts) and current (amps), returns as list."""
+    )
+
+    def __init__(self, adapter, **kwargs):
+        super().__init__(adapter, "Keithley 2260B DC Power Supply", read_termination="\n", **kwargs)
+
+    def apply(self, voltage, current):
```
bilderbuchi, line 69: have you tried making `applied` above a control, and feed it a tuple of (voltage, current) when setting, with set command `"apply %g,%g"`? I.e., `instr.applied = (15, 30)`
I think this could work, then you could drop this method and use the property instead to set. As the interface is basically equivalent, the API is a bit confusing as-is.

pymeasure/instruments/keithley/keithley2260B.py
```diff
@@ -0,0 +1,71 @@
...
+
+    voltage = Instrument.control(
+        ":SOUR:VOLT?", ":SOUR:VOLT %g",
+        """ A floating point property that controls the source voltage
+        in volts. The allowed range is not fixed. """,
+    )
+
+    power = Instrument.measurement(
+        ":MEAS:POW?",
+        """ Reads the power (in watts) the dc power supply is putting out.  """
```
bilderbuchi, line 58: Trailing whitespace in the docstring (also elsewhere).

pymeasure/instruments/keithley/keithley2260B.py
```diff
@@ -0,0 +1,71 @@
...
+        print(source.current)
+        print(source.power)
+        print(source.applied)
+
+    """
+
+    enabled = Instrument.control(
+        "OUTPut?", "OUTPut %d",
+        """A boolean property that controls whether the source is enabled, takes
+        values True or False (1 or 0). """,
```
bilderbuchi, line 38: This should only take True/False, not 1/0 as indicated in the docstring.

pymeasure/instruments/keithley/keithley2260B.py
```diff
@@ -0,0 +1,71 @@
...
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True
```
bilderbuchi, line 41: If you put a trailing `,` here, if you add another argument in the future you will not have to touch this line, too (to add it). This makes the git log/blame clearer.

pymeasure/instruments/keithley/keithley2260B.py
```diff
@@ -0,0 +1,71 @@
...
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class Keithley2260B(Instrument):
+    """ Represents the Keithley 2260B Power Supply (minimal implementation)
+    and provides a high-level interface for interacting with the instrument.
+
+    For a connection through tcpip, the socket is hardcoded to 2268,
```
bilderbuchi, line 20: I can't find this hard-coding in the source code. Do you maybe mean that the instrument only accepts connections at that port?

pymeasure/instruments/keithley/keithley2260B.py
```diff
@@ -0,0 +1,71 @@
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import (
+    # truncated_range,
```
bilderbuchi, line 3: Please remove unused imports.

pymeasure/instruments/keithley/keithley2260B.py
```diff
@@ -0,0 +1,71 @@
...
+        values True or False (1 or 0). """,
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True
+    )
+
+    current = Instrument.control(
+        ":SOUR:CURR?", ":SOUR:CURR %g",
+        """ A floating point property that controls the source current
+        in amps. The allowed range is not fixed. """,
```
bilderbuchi, line 47: Maybe better wording: "This is not checked against the allowed range"

bilderbuchi: When reading, does this return the actual current achieved, or the set current limit?

bilderbuchi (changes requested):
Thank you for your contribution, this is looking good! I left a couple of change requests for you.
Accommodating the different voltage/current limits is currently not really elegantly possible in pymeasure, this waits for the resolution of #250 (we can easily do this afterwards).

pymeasure/instruments/keithley/keithley2260B.py
bklebel: Yes, if connected via LAN and TCPIP, the instrument accepts connections only at this port (as far as I know), although a webserver can be reached at the usual http port 80. 
This is what I meant with hard-coded: it is hard-coded in the instrument. 
For me, it is usually a bit of a pain to find the correct port, at which an instrument will accept a connection, as (at least with Keithley) the manufacturers often neglect to mention it in the instrument documentation (reference manuals). I know of port scanning, but haven't had time to look up how it works, and many of my colleagues in experimental research fields are either not aware of it altogether, or don't immediately know how to do it either. 

Since this instrument can be connected via other methods too, I did not want to actually hard-code the adapter in pymeasure, but wanted to leave the hint for the socket in the documentation. I will change it to your wording.

pymeasure/instruments/keithley/keithley2260B.py
```diff
@@ -0,0 +1,71 @@
...
+        ":MEAS:POW?",
+        """ Reads the power (in watts) the dc power supply is putting out.  """
+    )
+
+    applied = Instrument.measurement(
+        ":APPly?",
+        """ Reads the applied voltage (volts) and current (amps), returns as list."""
+    )
+
+    def __init__(self, adapter, **kwargs):
```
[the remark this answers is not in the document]
bklebel: the `read_termination` is `\n` for the tcpip interface - I did not try for another. I will do the changes

pymeasure/instruments/keithley/keithley2260B.py
bklebel: I was not aware of this possibility, even though, now that I know of it, it seems obvious. I will make that change, and would at the same time suggest to improve pymeasure's documentation of "Adding instruments" to include this capability.

pymeasure/instruments/keithley/keithley2260B.py
bklebel: I think, this reads the set current limit - however, the device has the capability to operate both in constant voltage (CV) and constant current (CC) modes, and switches between the two depending on the set values and the applied load. if values are set which are outside the boundaries given by these circumstances, the instrument will ignore them (as far as I understand the manual). 
Depending on the mode (CC or CV), the reading gives either the current/voltage limit, or the actual value achieved, which is not immediately clear without further communication with the instrument. 

For a "full" implementation of this device to pymeasure, I guess it would be good to correctly observe these modes and read values for limits and actual achieved values separately (if possible). The implementation I can provide will necessarily lack this ability, and will only include the most prominent features of (most any keithley) power supplies, where some device-specific preferences and the convenience control `applied` are added.

pymeasure/instruments/keithley/keithley2260B.py
bklebel: I am not quite sure of the behavior of the validator with the `strict_discrete_set`, the given values and the `map_values` option. 
With the code as it is now, it accepts both `True`/`False` and `1`/`0` - i.e. with the version of pymeasure this was based on. possibly 1 and 0 are directly being interpreted as True/False? or was this changed recently?

pymeasure/instruments/keithley/keithley2260B.py
bklebel: thanks for the suggestion!

pymeasure/instruments/keithley/keithley2260B.py
bklebel: Done

pymeasure/instruments/keithley/keithley2260B.py
bklebel: Done

pymeasure/instruments/keithley/keithley2260B.py
bilderbuchi: OK, I suspected that, but it was not fully clear from the manual, and I don't have the device around to check.

When I implement power supplies (that have the necessary commands), I tend to define a quartet of `control`s of `voltage_setpoint` (the one you set), `voltage` (the actual voltage the device achieves, which can deviate in CC mode), `current` (the actual current, less than the limit in CV mode) and `current_limit` (the limit you set).
I found this to be the clearest way to express what is going on (and enabling expressive interaction/logging), maybe you also find that useful.

pymeasure/instruments/keithley/keithley2260B.py
bilderbuchi: Did you test correct operation with your device?
Yeah, I also did not think of that until now. :grin: 
Good idea! Could you add a remark about this possibility in https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#writing-properties (after the Python string format is mentioned)?

pymeasure/instruments/keithley/keithley2260B.py
bilderbuchi: I thought that giving e.g. `1` would fail when it looks up the input in the defined values map. It's not a big deal.

pymeasure/instruments/keithley/keithley2260B.py
```diff
@@ -0,0 +1,74 @@
...
+log.addHandler(logging.NullHandler())
+
+
+class Keithley2260B(Instrument):
+    """ Represents the Keithley 2260B Power Supply (minimal implementation)
+    and provides a high-level interface for interacting with the instrument.
+
+    For a connection through tcpip, the device only accepts
+    connections at port 2268, which cannot be configured otherwise.
+        example connection string: 'TCPIP::xxx.xxx.xxx.xxx::2268::SOCKET'
```
bilderbuchi, line 16: Hm, there is a weird jump in indentation here?

bilderbuchi (changes requested):

pymeasure/instruments/keithley/keithley2260B.py
bklebel: Yes I did test it with this device, it works perfectly well. 

I have created #410 for the change in the documentation

bilderbuchi: Great! You can just add this change to the present PR, no need to open a separate one!

pymeasure/instruments/keithley/keithley2260B.py
bklebel: Thank you for your suggestions! I have made changes accordingly.

pymeasure/instruments/keithley/keithley2260B.py
bklebel: fixed

pymeasure/instruments/keithley/keithley2260B.py
bilderbuchi: Never mind, I fixed the other PR already.

bilderbuchi:
LGTM, thank you for your contribution!

Issue #84, opened 2017-06-09
Can't exit with Ctrl+C


feinsteinben:
Hi,

I'm trying to run a procedure in a safe manner, such that it shuts down all the instruments when closed.
It based on this example: https://github.com/ralph-group/pymeasure/blob/master/examples/Basic/script.py
However, when I try exiting the script with Ctrl+C, it doesn't seem to recognize it. Do you have any suggestions regarding this issue?

Thanks,
Ben.

cjermain:
When the code in `script.py` gets to `worker.start()`, any call to Keyboard Interrupt should give you something like the printout below. (I'm using Ctrl+C on Linux.)

```
^C05:16:06 PM: User stopped Worker join prematurely (pymeasure.experiment.workers, WARNING)
05:16:06 PM: Worker has joined (root, INFO)
05:16:06 PM: Stopping the logging (root, INFO)
05:16:06 PM: User stopped Worker execution prematurely (root, WARNING)
05:16:06 PM: Emitting message: b'status' 1 (root, DEBUG)
05:16:06 PM: Finished (root, INFO)
05:16:06 PM: Recorder caught stop command (pymeasure.experiment.listeners, INFO)
```

That first line of the printout is where the interrupt is being caught.

If you insert a Keyboard Interrupt before `worker.start()`, the code is not designed to handle it. You can improve this by simply adding a try-catch statement that catches the interrupt. Since this section is just setting up the experiment, there won't be any instruments in use yet.

@feinsteinben, what OS are you using? The Keyboard Interrupt may be defined differently.

feinsteinben:
@cjermain Thanks for your response.
I tried to Ctrl+C after the worker had started.
I'm running that on windows 7.

jrr1984:
have you fixed this bug? It seems not to be working the keyboard interrup feature.

cjermain:
@jrr1984, can you give more specifics on your observations? What OS are you using? What versions of PyMeasure?

jrr1984:
Windows 10 64 bits, using pycharm IDE (i realised this IDE has a bug itself, maybe thats the problem!). The pymeasure version is the latests one from github, tomorrow i can give you more details. I know this is no the place to ask this but do you have any example using the results object when we emit a list as a result like the trace of an spectrometer or an oscilloscope? I would like the Plotter and then the GUI to plot the intensity vs wavelength traces from the spectrometer im using. Thank you very much in advance.

bilderbuchi:
[...]

PR #396, opened 2021-01-30
Fix for keyboard interrupt

CasperSchippers:
Propose a fix for the KeyboardInterrupt (i.e. ctrl + C) that is not working on some systems when running a measurement script (issue #84).

It seems to be related to a python bug that causes the `wait` function of an event to be uninterruptable (https://bugs.python.org/issue35935).

I proposed a fix (/workaround) by introducing an `InterruptableEvent` where the `wait` function is divided in many shorter waits.
Note that this fix only works when running a measurement script (e.g. the script.py in examples); it is not possible to interrupt the GUI with ctrl+C (though I also think that that would not be required, as there is an abort button in the software).
However, since closing the GUI using the close-button can still leave the measurement running when it is not finished yet, I here also propose to issue an abort when the GUI window is closed.

### DIFF
--- a/pymeasure/display/windows.py
+++ b/pymeasure/display/windows.py
@@ -301,6 +301,9 @@ class ManagedWindow(QtGui.QMainWindow):
         self.resize(1000, 800)
 
     def quit(self, evt=None):
+        if self.manager.is_running():
+            self.abort()
+
         self.close()
 
     def browser_item_changed(self, item, column):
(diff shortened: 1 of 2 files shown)


bilderbuchi:
Sounds good to me. One thing: Is there a noticeable CPU useage impact of the wait for 10 ms? I feel like, if we are waiting for a keyboard interrupt, a latency of 100 ms until it registers is still acceptable, and would result in 10x fewer wait cycles.

CasperSchippers:
Yeah, that might be a good idea. I'll adjust it and merge it.

cjermain:
Thanks @CasperSchippers!

PR #408, opened 2021-02-21
Add write binary values serial adapter

msmttchr:
This PR add the method `write_binary_values `in `SerialAdapter `and `PrologixAdapter` with same behavior of the version provided in `VISAAdapter`

### DIFF
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -104,7 +104,7 @@ class PrologixAdapter(SerialAdapter):
         self.connection.write(command.encode())
 
     def _format_binary_values(self, values, datatype='f', is_big_endian=False, header_fmt = "ieee"):
-        """Format values in binary format in order to be used with instrument commands.
+        """Format values in binary format, used internally in :meth:`.write_binary_values`.
 
         :param values: data to be writen to the device.
         :param datatype: the format string for a single element. See struct module.
@@ -138,7 +138,7 @@ class PrologixAdapter(SerialAdapter):
 
         :param command: SCPI command to be sent to the instrument
         :param values: iterable representing the binary values
-        :param kwargs: Key-word arguments to pass onto :meth:`~._format_binary_values`
+        :param kwargs: Key-word arguments to pass onto :meth:`._format_binary_values`
         :returns: number of bytes written
         """
         if self.address is not None:
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -103,6 +103,44 @@ class PrologixAdapter(SerialAdapter):
         command += "\n"
         self.connection.write(command.encode())
 
+    def format_binary_values(self, values, datatype='f', is_big_endian=False):
+        """Format values in binary format in order to be used with instrument commands.
+
+        :param values: data to be writen to the device.
+        :param datatype: the format string for a single element. See struct module.
+        :param is_big_endian: boolean indicating endianess.
+        :return: binary string.
+        :rtype: bytes
+        """
+
+        block = super().format_binary_values(values, datatype, is_big_endian)
+        # Prologix needs certian characters to be escaped.
+        # Special care must be taken when sending binary data to instruments. If any of the
+        # following characters occur in the binary data -- CR (ASCII 13), LF (ASCII 10), ESC
+        # (ASCII 27), '+' (ASCII 43) - they must be escaped by preceding them with an ESC
+        # character.
+        special_chars = b'\x0d\x0a\x1b\x2b'
+        new_block = b''
+        for b in block:
+            escape = b''
+            if b in special_chars:
+                escape = b'\x1b'
+            new_block += (escape + bytes((b,)))
+
+        return new_block
+
+    def write_binary_values(self, command, values, **kwargs):
+        """ Writes the command to the GPIB address stored in the
+        :attr:`.address`
+
+        :param command: SCPI command string to be sent to the instrument
+        """
+        if self.address is not None:
+            address_command = "++addr %d\n" % self.address
+            self.connection.write(address_command.encode())
+        super().write_binary_values(command, values, **kwargs)
+        self.connection.write('\n'.encode())
+
     def read(self):
         """ Reads the response of the instrument until timeout
 
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -103,7 +103,7 @@ class PrologixAdapter(SerialAdapter):
         command += "\n"
         self.connection.write(command.encode())
 
-    def format_binary_values(self, values, datatype='f', is_big_endian=False):
+    def _format_binary_values(self, values, datatype='f', is_big_endian=False):
         """Format values in binary format in order to be used with instrument commands.
 
         :param values: data to be writen to the device.
@@ -113,7 +113,7 @@ class PrologixAdapter(SerialAdapter):
         :rtype: bytes
         """
 
-        block = super().format_binary_values(values, datatype, is_big_endian)
+        block = super()._format_binary_values(values, datatype, is_big_endian)
         # Prologix needs certian characters to be escaped.
         # Special care must be taken when sending binary data to instruments. If any of the
         # following characters occur in the binary data -- CR (ASCII 13), LF (ASCII 10), ESC
@@ -130,10 +130,15 @@ class PrologixAdapter(SerialAdapter):
         return new_block
 
     def write_binary_values(self, command, values, **kwargs):
-        """ Writes the command to the GPIB address stored in the
-        :attr:`.address`
+        """ Write binary data to the instrument, e.g. waveform for signal generators.
 
-        :param command: SCPI command string to be sent to the instrument
+        values are encoded in a binary format according to
+        IEEE 488.2 Definite Length Arbitrary Block Response Data block.
+
+        :param command: SCPI command to be sent to the instrument
+        :param values: iterable representing the binary values
+        :param kwargs: Key-word arguments to pass onto `_format_binary_values`
+        :returns: number of bytes written
         """
         if self.address is not None:
             address_command = "++addr %d\n" % self.address
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -103,17 +103,18 @@ class PrologixAdapter(SerialAdapter):
         command += "\n"
         self.connection.write(command.encode())
 
-    def _format_binary_values(self, values, datatype='f', is_big_endian=False):
+    def _format_binary_values(self, values, datatype='f', is_big_endian=False, header_fmt = "ieee"):
         """Format values in binary format in order to be used with instrument commands.
 
         :param values: data to be writen to the device.
         :param datatype: the format string for a single element. See struct module.
         :param is_big_endian: boolean indicating endianess.
+        :param header_fmt: Format of the header prefixing the data ("ieee", "hp", "empty").
         :return: binary string.
         :rtype: bytes
         """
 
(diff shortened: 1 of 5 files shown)


bilderbuchi (changes requested):
Nice addition, I welcome the drive towards a more homogeneous API of our adapters.
Could you maybe add some tests so that we can confirm that this works as expected? You could probably use pyserials [loopback handler](https://pythonhosted.org/pyserial/url_handlers.html#loop), write some stuff and read back what arrived to see if it's as expected.

pymeasure/adapters/serial.py
```diff
@@ -78,5 +78,31 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
...
+        block = to_ieee_block(values, datatype, is_big_endian)
+
+        return block
+
+    def write_binary_values(self, command, values, **kwargs):
+        """ Write binary data to the instrument, e.g. waveform for signal generators
+
+        :param command: SCPI command to be sent to the instrument
+        :param values: iterable representing the binary values
+        :param kwargs: Key-word arguments to pass onto `write_binary_values`
```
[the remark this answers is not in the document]
msmttchr: Absolutely, I will fix it

pymeasure/adapters/serial.py
```diff
@@ -78,5 +78,31 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
         header, data = binary[:header_bytes], binary[header_bytes:]
         return np.fromstring(data, dtype=dtype)
 
+    def format_binary_values(self, values, datatype='f', is_big_endian=False):
```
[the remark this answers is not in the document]
msmttchr: Yes, I think your suggestion is right. The format_binary_values has been introduced to "mimic" behaviour of `write_binary_values `provided by `pyvisa `(see https://github.com/pyvisa/pyvisa/blob/main/pyvisa/resources/messagebased.py). This method is not intended to be called directly by users and its implementation depends on some internal API of `pyvisa`.
The motivation to have it are:
-  If `pyvisa`, in the future, will change this internal API, we could simply rewrite/adjust the `format_binary_values `  
- In the `PrologixAdapter`, we need to make some further adjustment to the binary data and this is achieved by specializing  `format_binary_values ` avoiding code duplication.

pymeasure/adapters/serial.py
```diff
@@ -78,5 +78,31 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
...
+    def write_binary_values(self, command, values, **kwargs):
+        """ Write binary data to the instrument, e.g. waveform for signal generators
+
+        :param command: SCPI command to be sent to the instrument
+        :param values: iterable representing the binary values
+        :param kwargs: Key-word arguments to pass onto `write_binary_values`
+        :returns: number of bytes written
+        """
+
+        block = self.format_binary_values(values, **kwargs)
```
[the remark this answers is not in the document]
msmttchr: Initially, I did not have the `format_binary_values`, but for the motivations described in https://github.com/pymeasure/pymeasure/pull/408#discussion_r584095442, I think having this internal API (with leading `_`) is probably better. In addition to this, the SCPI instruments understand ieee format (e.g. prefix by #, etc. see http://ena.support.keysight.com/e5061b/manuals/webhelp/eng/programming/remote_control/reading-writing_measurement_data/data_transfer_format.htm). Indeed, the implementation provided does not support HP format (`to_hp_block`) and pure binary format (`to_binary_block`) as in the original pyvisa write_binary_values API. This could be added, but I wanted to limit the dependency on function from `pyvisa`. What do you think ?

msmttchr:
> Nice addition, I welcome the drive towards a more homogeneous API of our adapters.
> Could you maybe add some tests so that we can confirm that this works as expected? You could probably use pyserials [loopback handler](https://pythonhosted.org/pyserial/url_handlers.html#loop), write some stuff and read back what arrived to see if it's as expected.

@bilderbuchi, thanks for the suggestions I have implemented a couple of tests using the loopback.

msmttchr:
Closed by mistake

pymeasure/adapters/serial.py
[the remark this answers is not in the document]
bilderbuchi: OK, I see your point about specializing PrologixAdapter.

pymeasure/adapters/serial.py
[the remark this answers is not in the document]
bilderbuchi: I think in case users have instruments expecting a different format, it would be valuable (and not too complicated) to have this user-configurable (and with equivalent API) as in [pyvisa write_binary_values](https://github.com/pyvisa/pyvisa/blob/9cab3576fb6b7c4aa18538d4337c67f8d971fcc7/pyvisa/resources/messagebased.py#L307-L314), with an iee default.
I think if we pull in one pyvisa util function, we might also pull in the other equivalent ones.

pymeasure/adapters/serial.py
```diff
@@ -45,7 +45,7 @@ class SerialAdapter(Adapter):
 
     def __init__(self, port, preprocess_reply=None, **kwargs):
         super().__init__(preprocess_reply=preprocess_reply)
-        if isinstance(port, serial.Serial):
+        if isinstance(port, serial.SerialBase):
```
bilderbuchi, line 48: I'm curious why this change was needed -- afaict, `serial_for_url` [returns a `Serial` instance](https://github.com/pyserial/pyserial/blob/0e7634747568547b8a7f9fd0c48ed74f16af4b23/serial/__init__.py#L61), not `SerialBase`.

tests/adapters/test_prologix.py
```diff
@@ -0,0 +1,49 @@
...
+from pymeasure.adapters import PrologixAdapter
+
+prefix="\n".join(["++auto 0", "++eoi 1", "++eos 2"])+"\n"
+
+def make_adapter(**kwargs):
+    return PrologixAdapter(serial.serial_for_url("loop://", **kwargs))
+
+
+@pytest.mark.parametrize("test_input,expected", [("OUTP", prefix + "OUTP\n"),
+                                                 ("POWER 22 dBm", prefix+ "POWER 22 dBm\n")])
```
bilderbuchi, line 37: missing space in `prefix+`.

tests/adapters/test_prologix.py
```diff
@@ -0,0 +1,49 @@
...
+def test_adapter_write(test_input, expected):
+    adapter = make_adapter(timeout=0.2)
+    adapter.write(test_input)
```
bilderbuchi, line 40: As long as you are testing non-binary write, wouldn' it be better to use `adapter.ask()` (and rename the test), as we would then also get to test that often-used part of the adapter API (and it's basically what you do, anyway)?
```
returned = adapter.ask(test_input)
assert returned == expected
```
Same for the serial test below.

tests/adapters/test_prologix.py
```diff
@@ -0,0 +1,49 @@
...
+    assert(adapter.connection.read(len(expected)+10) == expected.encode())
+
+@pytest.mark.parametrize("test_input,expected", [([1,2,3], prefix.encode() + b'OUTP#13\x01\x02\x03' + "\n".encode()),
+                                                 ([43, 27, 10,13, 97, 98, 99], prefix.encode() + b'OUTP#17\x1b\x2b\x1b\x1b\x1b\x0a\x1b\x0dabc'+ "\n".encode())])
```
bilderbuchi, line 44: Parameterized tests are great! :-)

bilderbuchi (changes requested):

pymeasure/adapters/prologix.py
```diff
@@ -103,6 +103,49 @@ def write(self, command):
...
+
+    def write_binary_values(self, command, values, **kwargs):
+        """ Write binary data to the instrument, e.g. waveform for signal generators.
+
+        values are encoded in a binary format according to
+        IEEE 488.2 Definite Length Arbitrary Block Response Data block.
+
+        :param command: SCPI command to be sent to the instrument
+        :param values: iterable representing the binary values
+        :param kwargs: Key-word arguments to pass onto `_format_binary_values`
```
bilderbuchi, line 140: If you change this to 
```
:meth:`~._format_binary_values`
```
the rendered documentation should get a link to _format_binary_values so that users can find out what kwargs are actually allowed.
I don't know if a private method gets documented by default -- maybe you can check that if you are set up for this? The documentation can be generated with sphinx by running `make html` in the docs/ directory, you need the sphinx and sphinx_rtd_theme packages installed.

pymeasure/adapters/serial.py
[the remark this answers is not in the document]
msmttchr: Ok

pymeasure/adapters/serial.py
msmttchr: @bilderbuchi , here what I found in my ubuntu machine:
```
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import serial
>>> serial.__version__
'3.4'
>>> s=serial.serial_for_url("loop://")
>>> s
Serial<id=0x7fe03456bcd0, open=True>(port='loop://', baudrate=9600, bytesize=8, parity='N', stopbits=1, timeout=None, xonxoff=False, rtscts=False, dsrdtr=False)
>>> s.__class__
<class 'serial.urlhandler.protocol_loop.Serial'>
>>> s.__class__.__base__
<class 'serial.serialutil.SerialBase'>
>>> s.__class__.__mro__
(<class 'serial.urlhandler.protocol_loop.Serial'>, <class 'serial.serialutil.SerialBase'>, <class 'io.RawIOBase'>, <class '_io._RawIOBase'>, <class 'io.IOBase'>, <class '_io._IOBase'>, <class 'object'>)
>>> isinstance(s, serial.Serial)
False
```
As you can see, loopback class is not an instance of `serial.Serial`, at least in the version I have.

tests/adapters/test_prologix.py
msmttchr: Ok

pymeasure/adapters/prologix.py
msmttchr: I have the documentation setup, but apparently internal API does not get documented in my test.

tests/adapters/test_prologix.py
msmttchr: Here, my intent is not to check that an hypothetical command is working by performing an read+write operation.
Indeed my aim is to check that when I send a write, the data I am writing to the device is what I have in the test vector.
As such, the loopback is very nice since it reports me exactly what has been written.
Using the `ask` or a `write`+`read`, would change the bytes read, because, for example, a `PrologixAdapter`, prefix all the read with a write of "`++read eoi`" string. I think writing a fully comprehensive test suite, it is beyond the scope of this PR.

bilderbuchi: I see, agreed. Thanks for the clarification, and for your patience with my many remarks! :-)

pymeasure/adapters/serial.py
bilderbuchi: Thank you for the analysis, I can confirm; this is fine I think. Pyserial chose an interesting design with several *different* classes called `Serial` :grin: .

pymeasure/adapters/prologix.py
bilderbuchi: I have pushed a doc fix to your PR branch. :-)

bilderbuchi:
LGTM, thank you for your great contribution (again), Michele!

A minor thing for the future: I think it's not necessary to prefix commits with "Change after @bilderbuchi feedback." or similar preambles; just write an expressive commit message describing the change.

msmttchr:
@bilderbuchi, thank you again for your valuable feedback

PR #410, opened 2021-03-02
Documentation for passing tuples to control property

bklebel:
This adds a comment in the documentation, which outlines how a control property can be used with multiple values at once, if the command of the instrument so allows it. 

Refer to the discussion in PR #395.


bilderbuchi:
I fixed the doctests that were failing the build. 
One curious thing is the number of commits here - are you branching your feature branches off of the current pymeasure master, or some weird/old version? It doesn't matter here because I'll just squash them, but maybe check this out on your end for future PRs.

bklebel:
Yes, you are correct, I have had some issues with organizing my fork properly - I hope it will be better for future PRs :)

Issue #296, opened 2020-06-30
Question regarding prefered implementation of non-SCPI devices


Applepi:
What is the preferred approach to add support for instruments that do not use SCPI to interface with the adapters. My thinking would be to write an interpreter to convert the non-ascii commands to SCPI commands, the reason for this being that some devices such as thorlabs rotation stages and such require to use either DLL or to send direct serial commands using APT and they possible bytes for the commands do not fall within the readable ASCII table. 

I want to add support for GRBL serial motion controllers as well as thorlabs APT devices, and I just wanted to touch base and see what the preferred approach for these would be. 

I think for the GRBL side of things it should be easy as all the commands are ascii sent over serial, but the main issue would be for me to write some sort of Throlabs APT to SPCI interpreter.

bilderbuchi:
Some discussion around frame-based communication can be found in the comments of #239. We think it implementing a separate Adapter class for frame-based communication would be most useful, maybe based on pyVISA RegisterBasedResource, but I have not yet investigated this in detail due to lack of time.

moritzj29:
since we discussed in the PR mentioned above: there appears to be an implementation of Thorlabs APT devices in Python, based on the underlying serial protocol (instead of using the DLL):

https://gitlab.com/yaq/thorlabs-apt-protocol

maybe this is useful for you.

ksunden:
Hi, author of that `thorlabs-apt-protocol` package here. 

If you have any questions, feel free to ask. It is mostly just wrapping the messages in python functions to be readable/handle the struct packing/unpacking. It itself does _precisely zero_ state maintenance or even direct communication (though you can hand it a pyserial or other file like serial pipe and it will allow you to iterate over responses).

If you want to see an example of it being used, I have an implementation of a yaq (https://yaq.fyi) daemon for thorlabs motors [here](https://gitlab.com/yaq/yaqd-thorlabs/-/blob/master/yaqd_thorlabs/_thorlabs_apt_motor.py).

bilderbuchi:
Thanks, appreciate the offer and the link to your implementation, that is quite useful. Oh, the APT implementation is quite large! :-o

In any case, I think we need to keep in mind that we have some base layer that can be used by ~arbitrary protocols, not only APT - it's big but not the only thing out there.

bilderbuchi:
#621 add an instrument with a command_word-register-length kind of syntax, so this fits into this frame-based communication scope.

PR #239, opened 2019-12-16
regatron power supply with scpi interface

tinix84:
driver scpi for regatron power supply adapted and tested in the lab

PR #417, opened 2021-03-22
Thorlabs DLL instruments: Generic Powermeter and Chrolis LED source

moritzj29:
This adds support for Thorlabs instruments based on a DLL driver:

1. `ThorlabsDLLAdapter` provides common methods of the two DLLs from Throlabs I had at hand. Hopefully, also other DLLs from Thorlabs match the pattern.
2. Two instruments are added, the Chrolis 6-wavelength LED source and a general Thorlabs Powermeter Instrument.

Pymeasure already supports the PM100USB powermeter via VISA communication, but only a few of the functionalities are implemented. Furthermore, Thorlabs switched from NI-VISA based drivers to libusb based drivers, which are used here via a DLL. Using the DLL for instrument communication it is possible to also use the most up-to-date Thorlabs software (Optical Power Monitor) without the need to switch drivers.

The DLL for the Chrolis device is not yet officially released, but was provided by Thorlabs customer support. Therefore, I marked the instrument as experimental in the documentation. But I think anyone interested may contact Thorlabs to get it. It works as good as the propietary software.

I'm happy for comments on improving my code!


bilderbuchi:
Thanks for this! 

I agree it would be great to be able to support DLL based drivers. It is a departure from the current scope of pymeasure, though, which has centered around scpi devices so far, so I think we need to be careful how this all fits together with the rest of the codebase. 

Some rough off-the cuff remarks/comments I had from skimming this: 
* I think we should take care to keep communication details abstracted away from instrument drivers and be confined to adapter class(es). 
* For example, `ctypes` stuff should not turn up in instrument implementations. Instruments use their `read`/`write`/`ask` methods to communicate with the device. Probably the usual SCPI get_command/set_command will have other near-equivalents for dll communication, e.g. command names (like `dll_indexed` uses) - maybe coupled with expected data type/size?.
* We should probably have a generic `DLLAdapter` with an API/method set compatible with `Adapter` (if possible/sensible). A derived `ThorlabsAdapter` could then implement thorlabs-specific stuff. 
* We should remain compatible with all OSes, not only Windows.
* This stuff is using VISA error codes (and other things?), but we cannot use pyvisa for communicating with it? Is there another python library available that is abstracting dll communication for us?
* Other relevant/related issues: #239 #296


For more nuanced feedback I need to find the time, which is difficult nowadays -- sorry!

moritzj29:
Thanks for the remarks!

## DLL instruments in general
I agree that pymeasure relies on SCPI commands so far and a lot of the features, like abstracting adapters, are only possible with "universal" commands.
The DLL commuincation is different in a sense that you need to tell e.g. the `ask` method what to expect as an answer. This can be several different objects of different type. For SCPI you just read and return the complete stream. (and parse it yourself). I think the whole adapter concept relies on the return data be some kind of datastream. With vendor specific libraries you don't have access to this datastream, since all the interpretation is handled in the library.

So I think it won't be possible to get as easy as:
`ret = self.ask('*IDN?')`
at least you need something like:
`ret = self.ask('IDcommand', 'stringbuffer', 256)`

I don't see much of a difference compared to: (except from the need to import the ctypes here instead of in the adapter file)
```
result = c_uint16(0)
message = create_string_buffer(256)
self.dll('selfTest', byref(result), byref(message))
```
One option would be to provide the parameters as a class variable instead of within the dll call:
```
param = {'selfTest': ('stringbuffer', 256), ...}
# or if ctypes is allowed within the instrument file
param = {'selfTest': (create_string_buffer, 256), ...}
ret = self.ask('selfTest')
```
`self.ask` would then look up the needed parameter buffers in `param` and return e.g. a tuple of the parameters. This would make the call more similar to the current style. On the other hand, the readability of the code suffers, since it is not in one place.

I think `ctypes` is the library to go for, since it is a Python standard library and also available for all platforms. If you get the instrument libraries for other platforms is a different topic... (at least the ones in this PR are only available for Windows). I have no idea if it's possible to port such stuff yourself and its definitely out of scope of this project.

So I guess, the first question is whether pymeasure wants to open it's scope towards more specific instruments which do not offer cross-protocol, cross-plattform usage. Don't get me wrong, either way would be fine. Development resources are limited...

## Thorlabs DLL instruments
The powermeter DLL originates from a VISA based driver, I think. Thorlabs switched the default communication protocol from VISA to libusb. This is probably the reason for the VISA error codes. In principle it would be possible to communicate via SCPI commands over the DLL, but I found this to be less realiable and much more time consuming to implement. (especially, since the existing PM100USB instrument has only a few commands implemented) I think Thorlabs provides this option to make it more easy for existing users to switch to the new driver.

The Chrolis device is much more recent and does not offer SCPI communication at all.

So I guess there will be some more discussion 😃

untzag:
For what it's worth, I'm pretty sure you can force the new drivers to support serial. In Windows there is a feature in the device manager called Virtual COM Port. If you check the checkbox you should be able to send serial commands directly over USB to that device. I don't know anything about the Chrolis device, but I suspect that the power meter is just talking SCPI under the hood anyway. Forcing serial where possible makes things much more portable on the hardware abstraction side.

Copying the following from https://gitlab.com/yaq/thorlabs-apt-protocol

On Windows, you must toggle a driver setting to make the COM port appear:

Within Device Manager, right click on the APT device (under USB devices), and go to Properties. On the Advanced tab, check the box that says Load VCP (VCP stands for Virtual COM Port). Unplug and replug the USB cable to make it load the COM Port. If the Advanced tab does not appear, I was able to use FTDI's FT_PROG. With all other Thorlabs programs not running, click Devices > Scan for devices. Review over the provided ID information to ensure that you are editing the correct device, FTDI provides serial communication for many different devices from several manufacturers, it is very possible you have multiple devices which show up. Under the Hardware Specific, uncheck the Load D2XX box. This box is a bit of a misnomer, as it really means "Load only D2XX", as when unchecked the D2XX driver (which the Thorlabs provided programs use) is still loaded, but the VCP driver is also loaded. Click Devices > Program.

Upon reconnecting (unplug and replug USB) the Advanced tab should appear as above, but the Load VCP driver option may not be checked yet.

moritzj29:
interesting point! Unfortunately, I was not able to activate a VCP for these devices. The Advanced Tab is not shown and also FT_PROG does not show anything... Probably Thorlabs uses different hardware then for the APT devices?

Checking the USB traffic with wireshark, the Thorlabs powermeter communication seems to be somehow based on SCPI commands, indeed. For the Chrolis I don't see any human readable commands.

Basing everything on serial is of course the most universal approach. And would probably also provide cross-platform use even though not initially supported by the manufacturer? On the contrary, it would be necessary to reverse engineer the protocol and reprogram the devices EEPROM to support the VCP.

untzag:
@moritzj29 that's disappointing. I want to believe it's possible but I have no direct experience with those drivers so I don't know for sure. Glad you gave it a shot :smile:

paulgoulain:
Also note that for the PM16 family of powermeter, it is possible to communicate with the instrument using the libusb driver with the python usbtmc library. It should also be possible with the PM100USB.

moritzj29:
Hello @paulgoulain, thanks for your comment! When I started working on this, I tried getting along with usbtmc (on windows) and was not very successful... I was in touch with Thorlabs support and they recommended using either the old VISA based driver or the new TLPM driver with the methods `TLPM_readRaw` / `TLPM_writeRaw` to issue SCPI based commands. Since I want to use the new driver to provide compatibility with the Thorlabs software, I went with accessing the methods through the driver directly instead of passing everything as SCPI commands.

Since cross-platform appears to be more desirable as I thought, looking into SCPI based communication may be worth another try... USBTMC seems to be the way to go for Linux based systems: https://github.com/WasatchPhotonics/FastPM100/blob/7fcab5575d51fc3ce460bc78d9c3cf08ba9f9c40/fastpm100/devices.py#L27

So maybe an abstract instrument driver with SCPI commands. And an adapter via USBTMC on Linux and based on the DLL on windows?

or just use an external library and leave it out of pymeasure completely...
https://pythonhosted.org/ThorlabsPM100/

paulgoulain:
Well if you want to try usbtmc on windows, here is the method I used :

- Install [pyusb](https://pypi.org/project/pyusb/)
- Download the lastest windows binary from [libusb](https://github.com/libusb/libusb/releases/download/v1.0.24/libusb-1.0.24.7z), extract it and copy the DLL (libusb-1.0.dll) corresponding to the compiler used for your python distrib (for mine, 3.8.x it is VS2019) to your system32 folder. Pyusb will use this DLL as a backend
- Install [python-usbtmc](pip install python-usbtmc)

Now you can try to connect to your device. Below is an example with the [PM16-401](https://www.thorlabs.com/thorproduct.cfm?partnumber=PM16-401) :

```Python
>>> import usbtmc
>>> usbtmc.list_devices()
[<DEVICE ID 1313:807c on Bus 002 Address 003>]
>>> instr = usbtmc.Instrument(0x1313, 0x807c)
>>> instr.open()
>>> print(instr.ask('*IDN?'))
Thorlabs,PM16-401,1902164,1.5.0
>>> print(instr.ask('READ?'))
2.5762554E-04
```

So I think it is possible to have a cross platform driver that uses the same code for Linux and Windows. But I think if one want to have a sturdy and full feature driver, it is best to use the Thorlabs DLL. And if it doesn't fit Pymeasure, outside Pymeasure is fine ;)

bilderbuchi:
Have you tried to talk to these usb devices via PyVISA? I see at least that USB [resource string examples are available](https://pyvisa.readthedocs.io/en/latest/introduction/names.html#visa-resource-syntax-and-examples), and folks in the issue tracker seem to talk to USB devices, probably automagically when pyusb is installed.
If this works, great, then we can probably work through our VISAAdapter. :+1: 
Considering our maintenance availability, I indeed don't think we can reasonably support the dll-based approach suggested here, I suspect this will lead to many unknowns, edge-cases, OS-problems etc. :-(

moritzj29:
Thanks everyone for the comments and suggestions, I really appreciated the discussion! I will close this PR due to the various reasons mentioned (and since I have no time diving deeper into the instrument communication on protocol level to eventually skip the DLL and make it cross-platform). Additionally there have been some improvements on the existing PM100 instrument in #504 in the meantime.

Maybe someone else will need the functionality and is willing to invest some more time on this. On Windows, the drivers in this PR are working fine now for >12 months in our lab. So I will keep my branch for anyone interested/in need.

Issue #180, opened 2019-02-14
Faulty function get_buffer() in dsp7265


badarirao:
I am trying to make the get_buffer()  function work, but unable to do so.
There are many issues in it like:
- self.length is not defined.
- the line "while int(self.values("M")) != 0:" should have self.values("M")[0] ?
- how is the variable count used? I don't see it changing anywhere, so how will it become greater than maxCount?
- also the command "self.values("DC. %d" % self.curve_bits[quantity])" is giving timeout error. 

Can you please check out these problems?

Thanking you,
Sincerely,
Badari

cjermain:
@badarirao, thanks for raising this issue! The `get_buffer` method indeed has some errors as implemented. The `self.length` is indeed not defined, but appears as if it should be a property that returns the length of the buffer. I would recommend that you get the DSP7265 manual and try to manually run the commands. Once we have the working set, we can revise this method.

badarirao:
@cjermain ,Thanks for looking into this bug.
Could you give me the algorithm on how the function is supposed to work? I am not exactly able to understand the given code.
Also, the signal recovery manual says something about implementation of serial poll in order to extract the data. I am not exactly sure on how to do this. The exact note in the manual is as follows:
" Note that when using this command with the GPIB interface the serial poll must be
used. After sending the DC command, perform repeated serial polls until bit 7 is set,
indicating that the instrument has an output waiting to be read. Then perform repeated
reads in a loop, waiting each time until bit 7 is set indicating that a new value is
available. The loop should continue until bit 1 is set, indicating that the transfer is
completed. "

Can you point me in the right direction on how to implement this?

Thanks,
Badari

cjermain:
Unfortunately, I did not write the original function, nor do I have access to that instrument.

I would suggest determining which command provides the status, and then do a bit comparison to find the value the manual suggests. You will need to use the `DC` command to pull back the values. All of this can be done by directly calling read/write/ask on the instrument, and manually stepping through the process.

bilderbuchi:
@badarirao did you every manage to fix the issues you report here?

badarirao:
@bilderbuchi  No I could not fix this issue.

bilderbuchi:
@CasperSchippers do I remember correctly that you are using such a device? If so, could you maybe take care of fixing this?

CasperSchippers:
I have used it in the past, and I think I can get access to this system; I'll have a look

CasperSchippers:
[...]

PR #418, opened 2021-03-22
Dsp7265

CasperSchippers:
PR to solve the issues in issue #180.

### DIFF
--- a/examples/Basic/LI_test.py
+++ b/examples/Basic/LI_test.py
@@ -1,0 +1,5 @@
+from pymeasure.instruments.signalrecovery import DSP7265
+
+if __name__ == '__main__':
+    li = DSP7265("visa://phys8173.campus.tue.nl/GPIB0::11::INSTR")
+    print(li.curve_buffer_length)
--- a/pymeasure/instruments/signalrecovery/dsp7265.py
+++ b/pymeasure/instruments/signalrecovery/dsp7265.py
@@ -169,9 +169,10 @@ class DSP7265(Instrument):
     )
 
     def __init__(self, resourceName, **kwargs):
-        super(DSP7265, self).__init__(
+        super().__init__(
             resourceName,
             "Signal Recovery DSP 7265",
+            includeSCPI=False,
             **kwargs
         )
         self.curve_bits = {
@@ -183,7 +184,6 @@ class DSP7265(Instrument):
             'ADC2': 64,
             'ADC3': 128
         }
-
 
     def values(self, command):
         """ Rewrite the method because of extra character in return string."""
--- a/pymeasure/instruments/signalrecovery/dsp7265.py
+++ b/pymeasure/instruments/signalrecovery/dsp7265.py
@@ -274,19 +274,55 @@ class DSP7265(Instrument):
         validator=truncated_range,
     )
 
+    curve_buffer_length = Instrument.control(
+        "LEN", "LEN %d",
+        """ An integer property that controls the length of the curve buffer.
+        Valid values are values between 1 and 32,768, but the actual maximum
+        amount of points is determined by the amount of curves that are stored,
+        as set via the curve_buffer_bits property (32,768 / n) """,
+        values=[1, 32768],
+        validator=truncated_range,
+    )
+
+    curve_buffer_interval = Instrument.control(
+        "LEN", "LEN %d",
+        """  An integer property that controls Sets the time interval between
+        successive points being acquired in the curve buffer. The time interval
+        is specified in ms with a resolution of 5 ms; input values are rounded
+        up to a multiple of 5. Valid values are values between 0 and 1,000,000,000
+        (corresponding to 12 days).
+        The interval may be set to 0, which sets the rate of data storage to the
+        curve buffer to 1.25 ms/point (800 Hz). However this only allows storage
+        of the X and Y channel outputs. There is no need to issue a CBD 3 command
+        to set this up since it happens automatically when acquisition starts.
+        """,
+        values=[1, 1000000000],
+        validator=truncated_range,
+    )
+
+    def init_curve_buffer(self):
+        """Initializes the curve storage memory and status variables. All record
+        of previously taken curves is removed.
+        """
+        self.write("NC")
+
     def set_buffer(self, points, quantities=['x'], interval=10.0e-3):
         num = 0
         for q in quantities:
             num += 2**self.CURVE_BITS[q]
-        self.points = points
-        self.write("CBD %d" % int(num))
-        self.write("LEN %d" % int(points))
+
+        self.curve_buffer_bits = num
+        self.curve_buffer_length = points
+
         # interval in increments of 5ms
-        interval = int(float(interval)/5.0e-3)
-        self.write("STR %d" % interval)
-        self.write("NC")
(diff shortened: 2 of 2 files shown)


examples/Basic/LI_test.py
```diff
@@ -0,0 +1,5 @@
+from pymeasure.instruments.signalrecovery import DSP7265
+
+if __name__ == '__main__':
+    li = DSP7265("visa://phys8173.campus.tue.nl/GPIB0::11::INSTR")
```
bilderbuchi, line 4: I think maybe don't include that test file specific to your setup?

CasperSchippers: Yes. That was a mistake indeed.

pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -166,13 +158,41 @@ class DSP7265(Instrument):
...
         """ A floating point property that controls the sensitivity
-        range in Volts, which can take discrete values from 2 nV to
-        1 V. This property can be set. """,
-        validator=truncated_discrete_set,
-        values=SENSITIVITIES,
+        range in Volts (for voltage mode) or Amps (for current modes).
+        When in Volts it takes discrete values from 2 nV to 1 V. When in
+        Amps it takes discrete values from 2 fA to 1 µA (for normal current
+        mode) or up to 10 nA (for low noise current mode).
+        This property can be set; setting requires a call to `imode`. """
```
bilderbuchi, line 169: "setting requires a call to `imode`." maybe delete that, since it is not something relevant for the user, and could cause confusion?

pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -199,19 +219,12 @@ def __init__(self, resourceName, **kwargs):
             resourceName,
```
bilderbuchi, line 219: would you mind moving the __init__ to the top of the class definition? I always find it frustrating to hunt for it.

pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -199,19 +219,12 @@ def __init__(self, resourceName, **kwargs):
...
             "Signal Recovery DSP 7265",
             includeSCPI=False,
+            # Remove extra unicode character
+            # TODO: (find a way to) test if the line underneath indeed correctly replaces the reimplemented values method
+            preprocess_reply=lambda r: r.replace('\x00', ''),
```
bilderbuchi, line 224: funny, I had another instrument model where some of the devices would return spurious \x00 bytes, do you know what causes this in your case?

bilderbuchi: In principle this functionality is already being tested [here](https://github.com/pymeasure/pymeasure/blob/0670a8ab4174fdeaf498c92f1d81042b7f705de1/tests/instruments/test_instrument.py#L175-L192). I guess you can test with your physical device to confirm the bytes get swallowed?

pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -341,29 +388,212 @@ def start_buffer(self):
...
+        returned data or an exception will be raised, depending on the value of
+        raise_error.
+
+        The requirements for converting the data are as follows:
+        - Converting X, Y, magnitude and noise requires sensitivity data, which
+            can either be part of the provided data or can be provided via the
+            sensitivity argument
+        - The same holds for X2, Y2, magnitude2 and noise2 with sensitivity2.
+        - Converting the frequency requires both 'frequency part 1' and
+            'frequency part 2'.
```
bilderbuchi, line 496:
```suggestion
        The requirements for converting the data are as follows:
        
        - Converting X, Y, magnitude and noise requires sensitivity data, which
          can either be part of the provided data or can be provided via the
          sensitivity argument
        - The same holds for X2, Y2, magnitude2 and noise2 with sensitivity2.
        - Converting the frequency requires both 'frequency part 1' and
          'frequency part 2'.
```
I think this should fix the documentation linter messages reported at the start of the file.

pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -341,29 +388,212 @@ def start_buffer(self):
...
+            if sensitivity is not None:
+                pass
+            elif "sensitivity" in data:
+                sensitivity = data["sensitivity"]
+            elif raise_error:
+                raise ValueError("X, Y, magnitude and noise cannot be converted as no "
+                                 "sensitivity is provided, neither as argument or as "
+                                 "part of the buffer_data. ")
+
+            if sensitivity is not None:
```
bilderbuchi, line 543: I think you can skip this check? Either sensitivity is already something (first if clause above), or it gets assigned from data, or we raise so never reach this code line.

pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -341,29 +388,212 @@ def start_buffer(self):
...
+            if sensitivity2 is not None:
+                pass
+            elif "sensitivity2" in data:
+                sensitivity2 = data["sensitivity2"]
+            elif raise_error:
+                raise ValueError("X2, Y2, magnitude2 and noise2 cannot be converted as no "
+                                 "sensitivity2 is provided, neither as argument or as "
+                                 "part of the buffer_data. ")
+
+            if sensitivity2 is not None:
```
bilderbuchi, line 591: see comment above.

bilderbuchi (changes requested):
I could not review the new additions in detail, it's a lot of code. I left a couple of remarks, if those are addressed this LGTM.

pymeasure/instruments/signalrecovery/dsp7265.py
CasperSchippers: Yes, I'll remove it.

pymeasure/instruments/signalrecovery/dsp7265.py
CasperSchippers: Sure

pymeasure/instruments/signalrecovery/dsp7265.py
CasperSchippers: Ahh thanks, I was looking for this but didn't yet completely understand that this would be the reason for the message

pymeasure/instruments/signalrecovery/dsp7265.py
CasperSchippers: No not really, in case there is no sensitivity provided but `raise_error=False` is also passed, sensitivity will be None in the second if clause; you'd want to use some sort of if-else-finally logic, but that doesn't exist in python.

pymeasure/instruments/signalrecovery/dsp7265.py
CasperSchippers: Well, my device (the one I used for testing this PR) doesn't do this. I left it there because originally, in this instrument the `values` method was reimplemented just for this purpose; maybe it also depends on the way you communicate with the system (GPIB vs. serial e.g.).

pymeasure/instruments/signalrecovery/dsp7265.py
bilderbuchi: Ah, sure, d'oh!

PR #421, opened 2021-03-28
Add HP6060B Electronic Load

bsdphk:
As it says on the tin... :-)

### DIFF
--- a/pymeasure/instruments/hp/hp6060b.py
+++ b/pymeasure/instruments/hp/hp6060b.py
@@ -1,0 +1,89 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2021 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Instrument, validators
+
+class HP6060B(Instrument):
+    """ Represents the HP HP6060B instrument.
+    """
+
+    voltage_measured = Instrument.measurement(
+        ":MEAS:VOLT?",
+        "Measured DC voltage, in Volts"
+    )
+
+    current_measured = Instrument.measurement(
+        ":MEAS:CURR?",
+        "Measured DC current, in Amps"
+    )
+
+    power_measured = Instrument.measurement(
+        ":MEAS:POW?",
+        "Measured DC power, in Watts"
+    )
+
+    mode = Instrument.control(
+        ":MODE?", ":MODE:%s",
+        """ Regulation mode: VOLT[AGE], CURR[ENT] or RES[ISTANCE]. """,
+        validator=validators.strict_discrete_set,
+        values=["VOLT", "VOLTAGE", "CURR", "CURRENT", "RES", "RESISTANCE"],
+    )
+
+    input = Instrument.control(
+        ":INP?", ":INP:STAT %s",
+        """ Input state: ON or OFF. """,
+        validator=validators.strict_discrete_set,
+        values=["ON", "OFF"],
+    )
+
+    current = Instrument.control(
+        ":CURR:LEVEL?", ":CURR:LEVEL:IMM %g",
+        """ Current mode regulation target, in Amps. """,
+        validator=validators.strict_range,
+        values=[0, 60],
+        check_set_errors=True,
+    )
+
+    voltage = Instrument.control(
+        ":VOLT:LEVEL?", ":VOLT:LEVEL:IMM %g",
+        """ Voltage mode regulation target, in Volts. """,
+        validator=validators.strict_range,
+        values=[0, 60],
+        check_set_errors=True,
+    )
+
+    resistance = Instrument.control(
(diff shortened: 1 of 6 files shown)


pymeasure/instruments/hp/hp6060b.py
```diff
@@ -0,0 +1,89 @@
...
+        ":VOLT:LEVEL?", ":VOLT:LEVEL:IMM %g",
+        """ Voltage mode regulation target, in Volts. """,
+        validator=validators.strict_range,
+        values=[0, 60],
+        check_set_errors=True,
+    )
+
+    resistance = Instrument.control(
+        ":RES:LEVEL?", ":RES:LEVEL:IMM %g",
+        """ Resistance mode regulation target, in Ohms. """,
```
CasperSchippers, line 78: It would be nice if the docstring included the range of valid values.

pymeasure/instruments/hp/hp6060b.py
```diff
@@ -0,0 +1,89 @@
...
+        ":CURR:LEVEL?", ":CURR:LEVEL:IMM %g",
+        """ Current mode regulation target, in Amps. """,
+        validator=validators.strict_range,
+        values=[0, 60],
+        check_set_errors=True,
+    )
+
+    voltage = Instrument.control(
+        ":VOLT:LEVEL?", ":VOLT:LEVEL:IMM %g",
+        """ Voltage mode regulation target, in Volts. """,
```
CasperSchippers, line 70: It would be nice if the docstring included the range of valid values.

pymeasure/instruments/hp/hp6060b.py
```diff
@@ -0,0 +1,89 @@
...
+        """ Input state: ON or OFF. """,
+        validator=validators.strict_discrete_set,
+        values=["ON", "OFF"],
+    )
+
+    current = Instrument.control(
+        ":CURR:LEVEL?", ":CURR:LEVEL:IMM %g",
+        """ Current mode regulation target, in Amps. """,
+        validator=validators.strict_range,
+        values=[0, 60],
```
CasperSchippers, line 64: It would be nice if the docstring included the range of valid values.

CasperSchippers (changes requested):
Thanks for the PR.
I've had a look and don't have much to remark,
Apart from the small remarks, could you add yourself to `AUTHORS.txt` and `docs/about/authors.rst`.

pymeasure/instruments/hp/hp6060b.py
bsdphk: done

pymeasure/instruments/hp/hp6060b.py
bsdphk: done

pymeasure/instruments/hp/hp6060b.py
bsdphk: done

pymeasure/instruments/hp/hp6060b.py
```diff
@@ -0,0 +1,89 @@
...
+        ":VOLT:LEVEL?", ":VOLT:LEVEL:IMM %g",
+        """ Voltage mode regulation target, [0...60] Volts. """,
+        validator=validators.strict_range,
+        values=[0, 60],
+        check_set_errors=True,
+    )
+
+    resistance = Instrument.control(
+        ":RES:LEVEL?", ":RES:LEVEL:IMM %g",
+        """ Resistance mode regulation target, [.033 ... 10000] Ohms. """,
```
CasperSchippers, line 78: The range you mention in the docstring is not exactly the same as the range you pass to `values`. Is that a mistake or is there a reason for this?

pymeasure/instruments/hp/hp6060b.py
```diff
@@ -0,0 +1,89 @@
...
+    input = Instrument.control(
+        ":INP?", ":INP:STAT %s",
+        """ Input state: ON or OFF. """,
+        validator=validators.strict_discrete_set,
+        values=["ON", "OFF"],
+    )
+
+    current = Instrument.control(
+        ":CURR:LEVEL?", ":CURR:LEVEL:IMM %g",
+        """ Current mode regulation target, [0...6] Amps. """,
```
CasperSchippers, line 62: The range you mention in the docstring is not exactly the same as the range you pass to values. Is that a mistake or is there a reason for this?

bsdphk:
I have implemented two range functions which I had hoped to be able to ignore, because I'm confused about how error handling is supposed to work in pymeasure.

What precisely should the `.check_errors()` method do ?  Return a list or raise an exeception ?   (I see both behaviours in the existing instruments)

If it returns a list, that is just thrown away if one sets `check_set_errors=True` in `control` ?

Also: is there a reason for not having a default SCPI `.check_error` function which Does The Right Thing™ ?

Back to the HP6060:  Should the `.resistance_range` update the range of the `.resistance` range-check so that out-of-range can be detected at the python level without bus-traffic ?

pymeasure/instruments/hp/hp6060b.py
bsdphk: I have added range-selection functions, see other comment.

pymeasure/instruments/hp/hp6060b.py
bsdphk: I have added range-selection functions, see other comment.

CasperSchippers (changes requested):
I agree that the `check_error` method is not as streamlined as could be. I think the best approach is to get all errors, log each of them using `log.error(errmsg)` (this is the way it is implemented in most systems), but maybe also return them as a list. This will indeed be discarded if used in the case of `check_set_errors=True` or `check_get_errors=True` in the `control`, `setting` and `measurement`, but will allow users to also manually check and get the errors and handle them if needed.
As to the default implementation for SCPI-enabled devices, I would be inclined to think it would be a good idea, but maybe @bilderbuchi could comment if there is a specific reason this method is not implemented by default.

I'm not sure if you need the range functions you mention to check for errors, but if so, do not forget to add the `check_set_errors=True` option.

Regarding updating the range of the `resistance` range-check: there is currently work being done on making this possible (#250, #371, #376).

pymeasure/instruments/hp/hp6060b.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+    mode = Instrument.control(
+        ":MODE?", ":MODE:%s",
+        """ Regulation mode: VOLT[AGE], CURR[ENT] or RES[ISTANCE]. """,
+        validator=validators.strict_discrete_set,
+        values=["VOLT", "VOLTAGE", "CURR", "CURRENT", "RES", "RESISTANCE"],
+    )
+
+    input = Instrument.control(
```
bilderbuchi, line 67: Maybe a little clearer/idiomatic could be to call this e.g. `input_enabled` and use a boolean instead of strings to interact with the control. See e.g. [here](https://github.com/pymeasure/pymeasure/blob/c07ac91be2212c976641f2b44b38b8c401ca15b8/pymeasure/instruments/keithley/keithley2400.py#L75-L83) how you can achieve that.

bilderbuchi (commented):
Looks good to me from a quick review, I left one remark.

bilderbuchi:
As for the set of default SCPI methods: this was before my time, could be nobody needed that before. But I agree, if it's a standard SCPI command it should probably be implemented. However, we have to think how we fail if an instrument (i.e. the hardware) does not implement a standard scpi command, or if that would even be a problem.
Fwiw, there is the slight complication of how to nicely enable/disable the standard scpi command set, see #170.

bmoneke:
Hey @bsdphk would you like to address the last few issues in order to make this PR merge?

bmoneke:

PR #423, opened 2021-03-29
Add filename input and placeholder logic

paulgoulain:
Add a filename inpute similar to #322. Also implement placeholder support using a PlaceholderReplacer class that take a string and replace defined substring with parameter value or function return value.

Currently, substring must be inside a <> pattern (example : \<time\> will be replaced with 17h02m40s). Two built in placeholders are defined : time and date. 
Other are defined when parameter's placeholder argument is provided.

Here is a sample script that generate a ManagedWindow with a filename input. When the Queue button is pressed, it will return some information about found placeholders in the filename string and the string with the placeholders replaced with the expected value.

[test_placeholder.py](https://gist.github.com/paulgoulain/76244ba4ed7dec5284152ccb6f355adf)

I created this draft pull request to discuss the implementation. Many things can be changed in order to make this new feature easier to fit in Pymeasure and its philosophy.


pymeasure/display/placeholders.py
```diff
@@ -0,0 +1,63 @@
...
+class PlaceholderReplacer():
+    """ Allows to replace sequence of characters in a string with information
+    or parameter values.
+
+    :param inputs: Reference to the input widget
+    """
+
+    def __init__(self, inputs):
+        
+        self.pattern = r'\<.*?\>'
```
CasperSchippers, line 19: Would it be an idea to use the built-in `format` method of python strings. Might give some additional flexibility.

CasperSchippers (commented):
Thanks for this idea/contribution.
I had a look at the PR (more into the idea and structure than into the details at this moment) and have some suggestions, especially for the placeholder part:
- If you use the procedure to get the values for the placeholder, rather than the inputs, you can also use this tool together with the sequencer-widget.
- I think it could be a good idea to use the built-in `str.format` function of python to do the actual replacement, giving some additional control/flexibility in the formatting of the values.
- Would it not be a good idea to use the `name` of the `Parameters`, rather than adding a `placeholder` argument? I think this keeps the actual usage more simple, as the `name` is also what's showed on the input-widget.
With these suggestions in mind I toyed around with an implementation, for which I opened a PR (hope you don't mind): #439. I invite you to have a look and would like to hear your thoughts about it/the suggestions.

CasperSchippers:
I think that when this is merged, it might also be nice to have a default implementation for the `queue` method using the directory input and filename input.

CasperSchippers:
As #439 has been merged, I suggest to remove the placeholder-logic part from the PR and base the filename input around that.

bmoneke:
@paulgoulain do you want to pick up your work or may I close this PR?

bmoneke:

Issue #160, opened 2018-08-30
Add read_termination and write_termination to SerialAdapter


robmarkcole:
As identified in https://github.com/ralph-group/pymeasure/issues/158 my Newport power meters require that every command be terminated with `\r`. I create this issue to discuss the solution. 
One approach is to create an `adapter` sub-class where for every command I `command += '\r` but this is very hacky. Alternatives?

bilderbuchi:
I also remember running into this previously (don't remember which instrument, though). I circumevented by appending the respective characters manually to each command, which is quite inelegant. I also did the custom Adapter with overridden `write` function, but that's just not nice.

Maybe a good approach would be to expose `write_termination` and `read_termination` parameters across adapters? Ideally both of them, because some instruments can be quite exotic concerning line endings. I seem to remember one that expected strings with `\r\n`, but returns terminated with `\n` (can't remember which one, though.

The canonical way with pyserial [seems to be](https://stackoverflow.com/a/16961872/599884) with io.TextWrapper (but use BufferedRandom instead of BufferedRWPair!). InstrumentKit has the capability to change the terminator (see https://github.com/Galvant/InstrumentKit/pull/130).

robmarkcole:
@bilderbuchi thanks for your advice, indeed the danfysik instrument implements a [custom adapter](https://github.com/ralph-group/pymeasure/blob/master/pymeasure/instruments/danfysik/adapters.py)  which I can fork as a quick solution. Longer term I like the proposal of implementing `terminator` across all adapters. If we adopt pymeasure in our lab then I would be willing to do this.

Re InstrumentKit, whats the difference to pymeasure?

bilderbuchi:
I think you can leave this issue open for now, as the "right" way to do it seems to be clear but not yet implemented. I'll let @cjermain decide how he wants to handle this.

Re IK, check out #53 (warning: long read)

cjermain:
@robmarkcole, this is an interesting problem. The [Danfysik adapter](https://github.com/ralph-group/pymeasure/blob/master/pymeasure/instruments/danfysik/adapters.py#L30) is certainly a good example, which required some more sophisticated logic on the read side to handle error reporting.  If you need that level of sophistication, then I think a dedicated class makes sense. 

For simple line termination, I like @bilderbuchi's suggestion for adding read and write termination strings. We could add them to the Adapter base class, and make sure they are implemented in child classes.

bilderbuchi:
[...]

PR #430, opened 2021-04-20
Add instrument: Lakeshore 421

CasperSchippers:
Found a partial implementation of the LakeShore 421 which I'll complete.

I also propose to add a (simple) standard Implementation of the `__enter__` and `__exit__` methods, so the instruments can easily be used in a `with` statement. As discussed in https://github.com/pymeasure/pymeasure/pull/368#discussion_r615414529, the usage may be limited but I think it doesn't hurt to add the methods nevertheless: they are only used by the `with` statement and will not generate any issues.


CasperSchippers:
I found that for this instrument a delay of 50 ms is needed between commands; hence I proposed to add an optional `command_delay` argument to `Instrument`.

bilderbuchi:
> I found that for this instrument a delay of 50 ms is needed between commands; hence I proposed to add an optional command_delay argument to Instrument, which might also be useful for other systems.

Can you say more about this? Is this between consecutive writes to the device? reads from the device? between a write and subsequent read? 
If the latter, that's what [PyVISAs query_delay](https://pyvisa.readthedocs.io/en/latest/api/resources.html?highlight=delay#pyvisa.resources.MessageBasedResource.query_delay) is for.

Also, is this delay not dependent on the used adapter/comm (e.g. serial vs. tcp/ip), or is this imposed by the device itself?

pymeasure/instruments/instrument.py
```diff
@@ -74,6 +77,12 @@ class Object(object):
         self.isShutdown = False
         log.info("Initializing %s." % self.name)
 
+    def __enter__(self):
```
bilderbuchi, line 80: OK, let's put this in. Thank you for including a test! 
Please also add some explanation and short example to the documentation, so that this is actually discoverable. :-)

pymeasure/instruments/instrument.py
```diff
@@ -46,7 +47,7 @@ class Instrument(object):
     """
 
     # noinspection PyPep8Naming
-    def __init__(self, adapter, name, includeSCPI=True, **kwargs):
+    def __init__(self, adapter, name, includeSCPI=True, command_delay=None, **kwargs):
```
bilderbuchi, line 50: The new parameter will need an entry in the class docstring. Please ensure that you explain which commands will be delayed.

pymeasure/instruments/lakeshore/lakeshore421.py
```diff
@@ -0,0 +1,395 @@
...
+        super(LakeShore421, self).__init__(
+            resource_name,
+            "Lake Shore 421 Gaussmeter",
+            read_termination='\r',
+            write_termination='\n',
+            command_delay=0.05,
+            **kwargs
+        )
+
+        self.adapter.connection.set_visa_attribute(
```
bilderbuchi, line 63: Using this assumes that a VISAAdapter is used, right? What happens if folks pass a SerialAdapter instance as `resource_name` (as our API enables).

pymeasure/instruments/lakeshore/lakeshore421.py
```diff
@@ -0,0 +1,395 @@
...
+            visa_const.VI_ATTR_ASRL_BAUD, 9600)
+        self.adapter.connection.set_visa_attribute(
```
bilderbuchi, line 65: I wonder: why are you using the `set_visa_attribute` method, instead of the [simpler](https://pyvisa.readthedocs.io/en/latest/api/resources.html?highlight=delay#pyvisa.resources.SerialInstrument) `connection.baud_rate`, `connection.data_bits` etc. attributes? There'd also be no need to drag in visa_const, then.

pymeasure/instruments/lakeshore/lakeshore421.py
```diff
@@ -0,0 +1,395 @@
...
+        values=['G', 'T', 'Oe', 'A/m'],
+    )
+
+    field_range = Instrument.control(
+        "RANGE?", "RANGE %d",
+        """ A floating point property that controls the field range of the
+        meter. Valid values are 0.003, 0.03, 0.3, and 3 (in Tesla). """,
+        validator=truncated_discrete_set,
+        values=RANGES,
+        map_values=True
```
bilderbuchi, line 120: Generally, it's advisable to include a trailing comma in the last argument (or a dict entry). That way, when you add another argument, you don't have to change this line, too.

bilderbuchi (commented):
Some remarks from a quick read of the PR.

pymeasure/instruments/instrument.py
CasperSchippers: Yes, will do!

pymeasure/instruments/lakeshore/lakeshore421.py
CasperSchippers: Good one, I will check this and make sure it works for all cases

pymeasure/instruments/lakeshore/lakeshore421.py
CasperSchippers: Yes, you're right, I'd normally do this, but now I seemed to have forgotten.

pymeasure/instruments/instrument.py
CasperSchippers: Yes, will do!

pymeasure/instruments/lakeshore/lakeshore421.py
CasperSchippers: This is a remainder from way back when this instrument was last used by me (I then didn't finish the implementation); this was needed for some reason then and I didn't think too much about that part when finishing it; I'll change this into the simpler attributes.

CasperSchippers:
> Can you say more about this? Is this between consecutive writes to the device? reads from the device? between a write and subsequent read?
> If the latter, that's what [PyVISAs query_delay](https://pyvisa.readthedocs.io/en/latest/api/resources.html?highlight=delay#pyvisa.resources.MessageBasedResource.query_delay) is for. That is the only situation I have encountered.

It's a delay between consecutive writes to the device; I tried if the query_delay would suffice, but it didn't unfortunately.
 
> Also, is this delay not dependent on the used adapter/comm (e.g. serial vs. tcp/ip), or is this imposed by the device itself? So, I'm not sure if this should not happen at the adapter level. What do you think?

I think it is more of an instrument property, but for this instrument itself it is a bit ambiguous because it only has a serial port for communication. However, the newer edition (the Lakeshore 425) imposes a similar limitation (of 30 ms) according to it's documentation with a RS-232 over USB connection. I think also maybe other instruments can benefit from it (e.g. the Temptronics that is being implemented in PR #368; https://github.com/pymeasure/pymeasure/pull/368#r598807397_), hence I would place it in the instrument implementation to make it adapter-independent

pymeasure/instruments/lakeshore/lakeshore421.py
bilderbuchi: In the end, when we resolve the safe_keywords issue, these can be passed as kwargs to the adapter constructor.

bilderbuchi:
OK, that's weird, I see. I am a bit concerned about messiness in the API -- if we start placing sleeps here and there, it will be hard to keep the overview over what's going on and where speed bottlenecks are. Especially, as we now have `query_delay` and `command_delay`. But maybe this is the best way we can do it. Especially, as every adapter overrides `write` and does not call the base method, it will be hard to DRY if we put it in adapter. Instrumentkit has a nicer structure of the adapter abstraction (which also enables common comms logging defined only once).

If it is only before _writes_ to the device, maybe `write_delay` would be a more fitting name, btw? (also `last_write_time`)

pymeasure/instruments/lakeshore/lakeshore421.py
```diff
@@ -54,20 +54,29 @@ def __init__(self, resource_name, **kwargs):
...
+        elif isinstance(self.adapter, SerialAdapter):
+            self.adapter.connection.baudrate = 9600
+            self.adapter.connection.bytesize = 7
+            self.adapter.connection.stopbits = 1
+            self.adapter.connection.parity = "O"
+            self.adapter.connection.timeout = 1
+
+            # Ensure a \n is written after every command
+            write = self.adapter.write
+            self.adapter.write = lambda command: write(command + '\n')
```
bilderbuchi, line 79: Huh, I was not aware that the format for the arguments (parity, stop_bits) differs between pyvisa and serial!
We should think if we should reduce priority/support for SerialAdapter and long-term replace it with VISAAdapter. 

I think in this case, we can only hit the elif clause if people passed a finished SerialAdapter (instead of naming the port with a string). In this special case, I think it's safe to assume that people can configure/have configured the adapter correctly themselves (otherwise, why pass an adapter instance?), so I would drop the elif clause.
It's probably harder to argue in case of the patched read_termination, but placing this here is imo really ugly, we should rather fix #160 instead.

bilderbuchi: Argh this is so messy sometimes. :-/

CasperSchippers: Yes, I agree it is indeed a bit ugly.
You're argument about people passing a SerialAdapter is sound I think, so I can drop the else clause.

CasperSchippers:
For the read_termination (and also for the write_delay as you mentioned before and the uniformity of badparity/stopbits/baud/etc.) it would indeed be good to add/improve the abstraction of the Adapter class and its subclasses; I'll hapily start on this if you think/agree we should go ahead with that. (We'll only risk that part is going to look a bit like the way InstrumentKit has implemented it, though I don't know if that would be a problem).

bilderbuchi:
> I'll hapily start on this (in a new PR to keep things clear) if you think/agree we should go ahead with that. (We'll only risk that part is going to look a bit like the way InstrumentKit has implemented it, though I don't know if that would be a problem).

Maybe let's discuss/flesh the design out in an issue first? Or at least use that to keep track of the discussion, there are several other issues surrounding that topic, maybe we should do some work piece-by-piece first?
Also, to be honest, I'd rather bring the number of open PRs/bugs down first before attacking redesigns (if that is not necessary for said bugs). That way I think it's easier to keep an overview, and not too many parts are moving at the same time :-/

CasperSchippers:
Hmmm, yeah fair enough, then let's put this off for now.

bilderbuchi:
I think it's better to move this to the specific instrument. If we end up rolling this out generally, it's easy to migrate; the other way -- not so much.
I create #512 for the Adapter restructuring discussion.

I think we should discuss if there is an expectation that Instrument.shutdown does more. Right now it's ~ a no-op - should it at least close the connection to the hardware? I'll make an issue for this.

bilderbuchi:
By the way, is the reason behind the delay a problem with the comms hardware, or instrument saturation? If the latter, is it capable of using SCPI wait, and does that help (could be appended to all affected commands, separated with semicolon)
> *WAI - Wait
Prohibits the instrument from executing any new commands until all pending overlapped commands have been completed. See [Understanding Command Synchronization](https://na.support.keysight.com/pxi/help/latest/Programming/Learning_about_GPIB/Understanding_Command_Synchronization.htm)

CasperSchippers:
> By the way, is the reason behind the delay a problem with the comms hardware, or instrument saturation? If the latter, is it capable of using SCPI wait, and does that help (could be appended to all affected commands, separated with semicolon)

I'm not entirely sure what is the root of this delay problem; the instrument just seems to be slow to react.
I tried it, but it doesn't respond to the SCPI wait command.

CasperSchippers:
I moved the write-delay to the LakeShore 421 instrument class itself.

pymeasure/instruments/lakeshore/lakeshore421.py
```diff
@@ -0,0 +1,448 @@
...
+            asrl=dict(
+                baud_rate=9600,
+                data_bits=7,
+                stop_bits=10,
+                parity=1,
+                read_termination='\r',
+                write_termination='\n',
+            ),
+            **kwargs
+        )
```
msmttchr, line 72: According to updated documentation, I understand this way of setting defaults is suggested for instruments with multiple interfaces. For single interface, you could simply set the default with initialized keyword parameters.

CasperSchippers: Ahh, yes, I've changed it

bilderbuchi (approved):
LGTM!

Issue #428, opened 2021-04-18
tests fail because of QT


LongnoseRob:
Hello,
I was working on implementing a class for the HP3478A non-scpi multimeter.
After I had my implementation working, I tried the unit tests as recommended in the contribution section of the documentation.

with my system running on Archlinux I get the following errors, which seem to be related to some recent changes in the QT support modules:

`============================================================================================= short test summary info =============================================================================================
ERROR tests/display/test_inputs.py - AttributeError: module 'PySide2.QtCore' has no attribute 'pyqtSignal'
ERROR tests/display/test_plotter.py - AttributeError: module 'PySide2.QtCore' has no attribute 'pyqtSignal'
ERROR tests/display/test_windows.py - AttributeError: module 'PySide2.QtCore' has no attribute 'pyqtSignal'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 3 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
========================================================================================== 2 skipped, 3 errors in 31.27s ==========================================================================================
`

full console log is attached

version infomation on my used python etc:
`platform linux -- Python 3.9.3, pytest-6.2.3, py-1.10.0, pluggy-0.13.1
PySide2 5.15.2 -- Qt runtime 5.15.2 -- Qt compiled 5.15.2
rootdir: /home/robby/builds/pymeasure
plugins: qt-3.3.0
`
[20210418_test_console_log.txt](https://github.com/pymeasure/pymeasure/files/6330342/20210418_test_console_log.txt)

msmttchr:
This is probably due to some conflict between PyQt5 and PySide2.
I could suggest to have a look at this link https://pyqtgraph.readthedocs.io/en/latest/how_to_use.html#pyqt-and-pyside.

msmttchr:
Checked on my PC and confirmed your problem:
```
Python 3.9.2 (tags/v3.9.2:1a79785, Feb 19 2021, 13:44:55) [MSC v.1928 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import PySide2
[...]

PR #431, opened 2021-04-20
Fix pyside2

msmttchr:
Added support for PySide 2 while keeping PyQt5 functional. See also issue #428

### DIFF
--- a/pymeasure/display/Qt.py
+++ b/pymeasure/display/Qt.py
@@ -29,7 +29,7 @@ from pyqtgraph.Qt import QtGui, QtCore, loadUiType
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
-QtCore.QSignal = QtCore.pyqtSignal
+QtCore.QSignal = QtCore.Signal
 
 
 def fromUi(*args, **kwargs):
@@ -52,9 +52,11 @@ def fromUi(*args, **kwargs):
 def qt_min_version(major, minor=0):
     """
     Check for a minimum Qt version. For example, to check for version 4.11
-    or later, call ``check_qt_version(4, 11)``.
+    or later, call ``qt_min_version(4, 11)``.
 
-    :return bool: True if PyQt version >= min_version
+    :return bool: True if Python Qt bindings library version >= min_version
     """
-    return (QtCore.QT_VERSION >= ((major << 16) + (minor << 8)))
-
+    version = QtCore.qVersion().split(".")
+    version_major = int(version[0])
+    version_minor = int(version[1])
+    return (version_major > major) or ((version_major == major) and  (version_minor >= minor))
--- a/pymeasure/display/Qt.py
+++ b/pymeasure/display/Qt.py
@@ -48,15 +48,3 @@ def fromUi(*args, **kwargs):
         if isinstance(element, QtGui.QWidget):
             setattr(widget, name, element)
     return widget
-
-def qt_min_version(major, minor=0):
-    """
-    Check for a minimum Qt version. For example, to check for version 4.11
-    or later, call ``qt_min_version(4, 11)``.
-
-    :return bool: True if Python Qt bindings library version >= min_version
-    """
-    version = QtCore.qVersion().split(".")
-    version_major = int(version[0])
-    version_minor = int(version[1])
-    return (version_major > major) or ((version_major == major) and (version_minor >= minor))
--- a/pymeasure/display/Qt.py
+++ b/pymeasure/display/Qt.py
@@ -59,4 +59,4 @@ def qt_min_version(major, minor=0):
     version = QtCore.qVersion().split(".")
     version_major = int(version[0])
     version_minor = int(version[1])
-    return (version_major > major) or ((version_major == major) and  (version_minor >= minor))
+    return (version_major > major) or ((version_major == major) and (version_minor >= minor))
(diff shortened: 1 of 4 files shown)


pymeasure/display/log.py
```diff
@@ -32,12 +32,16 @@
 log.addHandler(logging.NullHandler())
 
 
-class LogHandler(QtCore.QObject, Handler):
-    record = QtCore.QSignal(object)
+class LogHandler(Handler):
+    class Emitter(QtCore.QObject):
```
bilderbuchi, line 36: Why the nested class? As you only end up using `record`, would it not be simpler to do `self.record = QtCore.QSignal(object)` in `LogHandler`s init, and use that?

pymeasure/display/inputs.py
```diff
@@ -80,18 +80,14 @@ def parameter(self):
         return self._parameter
 
 
-class StringInput(QtGui.QLineEdit, Input):
+class StringInput(Input, QtGui.QLineEdit):
```
bilderbuchi, line 83: If you have the time to comment, I'd be curious why the reordered superclasses support multiple inheritance in pyside2 better?

bilderbuchi (commented):
Looks good to me, I only have two small questions.
@CasperSchippers as this is GUI code, could you please review/merge this?

pymeasure/display/log.py
msmttchr: Qt signal needs to be class attribute, so this is why I was forced to create a dedicated class.
I preferred to put inside the LogHandler class to limit its visibility, since I do not see a general use for it.

pymeasure/display/inputs.py
msmttchr: You can have a look at https://github.com/pymeasure/pymeasure/pull/383#discussion_r614554704

Essentially using `super ` and multiple inheritance requires that the classes are designed to work cooperatively.
This boils down to the fact that `__init__ `method should call `super`. In fact the class `Input `does this, but the widget classes from `Pyside2 `apparently not.

pymeasure/display/log.py
bilderbuchi: OK, I see. `record` already was a class attribute inside the `LogHandler` class, though?

pymeasure/display/inputs.py
bilderbuchi: Ah, that makes sense, some shenanigans with `super` were the primary culprit I suspected.

pymeasure/display/log.py
msmttchr: Indeed the original code was good for `Qt `signal management and for `logging` `emit `method.
The changes are required to avoid problems linked to conflict with `emit `method in `PySide2` signal management.
See also https://stackoverflow.com/questions/66664542/conflicting-names-between-logging-emit-function-and-qt-emit-signal

bilderbuchi: I get the part about clobbering `emit`, and indeed that is resolved nicely. I just don't see how the `LogHandler` class attribute `record` is a part of that conflict, and we need to hide in in an `Emitter` class (and associated instance field `emitter`) that to me only seems to introduce additional API/interface surface for no reason I can discern. That's why I'm asking. :-)

bilderbuchi: In other words, `self.emitter` does not do much and looks to me similar to a pass-through method.

msmttchr: It's a bit weird, but an example can explain.
In `PySide2`, if an host class which contains a `Signal` class attribute and a method called `emit`, whenever you call the `emit` method of the `Signal` attribute, the host class method `emit` get called.
As a result, you can have either mismatch in the `emit` parameters or infinite recursion.
Have a look at the attached script, in particular, the class method `emit` method has a second parameter to avoid the signature mismatch.
The script goes into an infinite loop.
If the questions is why `PySide2` behaves like that then I do not know, but this is the best compromise I found to keep things running.

[test_pyside2.zip](https://github.com/pymeasure/pymeasure/files/6367586/test_pyside2.zip)

bilderbuchi: Ah, that was the context I was missing, thanks for the explanation! Can you maybe add that as a commit to the `Emitter` class so that our future selves can get their bearings quickly? (see [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence))

msmttchr: Added a comment to clarify purpose of class Emitter.

pymeasure/display/Qt.py
```diff
@@ -52,9 +52,11 @@ def fromUi(*args, **kwargs):
 def qt_min_version(major, minor=0):
```
CasperSchippers, line 52: It seems to me that this function is not used anymore now you've removed the calls in `inputs.py`. Is there a reason to keep the function around? Maybe for backwards compatilibity, though I do not see the point in this case, since it is not a real feature of pymeasure; what do you think @bilderbuchi?

CasperSchippers (approved):
Thanks for this update and the explanations you've provided. 
I don't have any remarks about the code, just a single question about `qt_min_version`.

pymeasure/display/Qt.py
bilderbuchi: Yeah I think it's fine to remove it if we don't use it in the code base.

msmttchr: OK, I will do this then.

PR #434, opened 2021-04-23
Keysight Technologies 34980A Switch Matrix library

Neel-Sancheti:
Keysight Technologies 34980A Multifunction Switch/Measure Unit with support for Keysight technologies 34923A Multifunction Switch/Measure Unit.

### DIFF
--- a/pymeasure/instruments/keysight/keysight34980A.py
+++ b/pymeasure/instruments/keysight/keysight34980A.py
@@ -28,7 +28,7 @@ from pyvisa import VisaIOError
 
 class Keysight34980A(Instrument):
     """ Represents the Keysight34980A Multifunction Switch/Measure Unit
-    and provides a high-level interface for taking scans with Keysight 34923A 34980A 
+    and provides a high-level interface for taking scans with Keysight 34923A 
     Multifunction Switch/Measure Unit.
     """
 
--- a/pymeasure/instruments/keysight/keysight34980A.py
+++ b/pymeasure/instruments/keysight/keysight34980A.py
@@ -1,0 +1,520 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2021 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Instrument, discreteTruncate, RangeException 
+import pyvisa as visa
+from pyvisa import VisaIOError
+
+class Keysight34980A(Instrument):
+    """ Represents the Keysight34980A Multifunction Switch/Measure Unit
+    and provides a high-level interface for taking scans with Keysight 34923A 34980A 
+    Multifunction Switch/Measure Unit.
+    """
+
+    def __init__(self,device_address):
+        print("Initializing Switch Matrix\n")
+        global rm 
+        rm = visa.ResourceManager()
+        self.switch_data = rm.open_resource(device_address)
+
+    #This command reads the instrument's (mainframe) identification string which contains 
+    #four comma-separated fields. The first field is the manufacturer's name, the second 
+    #field is the instrument model number, the third field is the serial number, and the 
+    #fourth field is a firmware revision code which contains four codes separated by dashes. 
+    def is_identification(self):
+        return(str(self.switch_data.query('*IDN?')))
+
+    #This command opens all channel relays and Analog Bus relays on the specified multiplexer 
+    #or switch module. Alternately by specifying the "ALL" parameter, this command opens 
+    #all channels on all installed modules. 
+    def set_channel_open_all(self, module = 1):    
+        self.switch_data.write(":ROUTe:OPEN:ALL " + str(module))
+
+    #This command opens the specified channel on a multiplexer or switch module.         
+    def set_channel_open(self, chan = ""):    
+        self.switch_data.write(":ROUTe:OPEN (@" + str(chan) + ")")
+        while(self.is_channel_open(chan) == 0):
+            self.switch_data.write(":ROUTe:OPEN (@" + str(chan) + ")")
+
+    #This command checks if the specified channel on a multiplexer or switch module is open. 
+    def is_channel_open(self, chan = ""):    
+        temp_values = self.switch_data.query_ascii_values(":ROUTe:OPEN? (@" + str(chan) + ")")
+        return(str(temp_values[0]))
+
+    #This command closes the connections to four Analog Buses on all installed modules. 
+    def set_abus_close(self, abus = 1):
+        chan = str("192") + str(abus)
+        self.set_channel_close(chan)
+
+    #This command opens the connections to four Analog Buses on installed modules. 
+    def set_abus_open(self, abus = 1):
+        self.switch_data.write(":ROUTe:OPEN:ABUS " + str(abus))
+        while(self.is_abus_open(abus) == 1):
+            self.switch_data.write(":ROUTe:OPEN:ABUS " + str(abus))
+
+    #This command opens all connections to all four Analog Buses on all installed modules.
+    def set_abus_open_all(self):
+        self.switch_data.write(":ROUTe:OPEN:ABUS ALL")
+        while((self.is_abus_open(1) == 1) and (self.is_abus_open(2) == 1) and (self.is_abus_open(3) == 1) and (self.is_abus_open(4) == 1)):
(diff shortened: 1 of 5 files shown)


Neel-Sancheti:
As per Christoph Buchner's comment. I have added a commit and am waiting for the approval. I am new to git, all the guidance is appreciated.

Thankyou!

pymeasure/instruments/keysight/keysight34980A.py
```diff
@@ -0,0 +1,520 @@
...
+    #field is the instrument model number, the third field is the serial number, and the 
+    #fourth field is a firmware revision code which contains four codes separated by dashes. 
+    def is_identification(self):
+        return(str(self.switch_data.query('*IDN?')))
+
+    #This command opens all channel relays and Analog Bus relays on the specified multiplexer 
+    #or switch module. Alternately by specifying the "ALL" parameter, this command opens 
+    #all channels on all installed modules. 
+    def set_channel_open_all(self, module = 1):    
+        self.switch_data.write(":ROUTe:OPEN:ALL " + str(module))
```
CasperSchippers, line 52: See if you can convert the methods into `control`, `measurement`, or `setting` properties.

pymeasure/instruments/keysight/keysight34980A.py
```diff
@@ -0,0 +1,520 @@
...
+
+    #This command opens the specified channel on a multiplexer or switch module.
```
CasperSchippers, line 54: This documentation string can better go into a docstring of one of the properties, or if a method cannot be converted into a property into the docstring format of that method, e.g.:
```
def some_method(self):
    """ Put the docstring over here
    Can also be multiline
    """

    # body of the method
```

This way, the documentation system that generates all the documentation knows where to find these docstrings and can use it for the doc-page of the instrument.

CasperSchippers (changes requested):
Thanks for the contribution.
The PyMeasure instruments section provides a number of features that make programming and maintaining an instrument easier. I've remarked upon a few in the file, but do also reccomend you to take a look at this part of the documentation: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html.

A lot of properties can converted into a `control`, `measurement`, or `setting` property. For those that cannot be converted into this, I would recommend using `self.write` in stead of `self.switch_data.write` and `self.values` (or in some cases `self.ask`) in stead of `self.switch_data.query` (though, if at all possible, it is always recomended to use the aforementioned properties).

Apart from that, you could also have a look at how the Keithley 2700 or 2750 is implemented, these are also matrix multiplexers and might be helpful to look at.

Finally, don't forget to add the instrument to __init__.py file in the Keysight folder, add the required documentation files in the docs/api/instruments/keysight folder, and add yourself to the author lists (one at AUTHORS.txt at the root of the repo and one in docs/about/authors.rst).

If you have any questions or anything is not (completely) clear, feel free to let me know!

Neel-Sancheti:
Hey Casper, I have updated the code according to your comments. Thank you for your help!

CasperSchippers:
> Hey Casper, I have updated the code according to your comments. Thank you for your help!

That's very good; you might want to push the changes, than I can have another look.

Neel-Sancheti:
> > Hey Casper, I have updated the code according to your comments. Thank you for your help!
> 
> That's very good; you might want to push the changes, than I can have another look.

I have updated them. Can you see the changes?

CasperSchippers:
> I have updated them. Can you see the changes?

Yes, I'll have a look asap

pymeasure/instruments/keysight/keysight34980A.py
```diff
@@ -0,0 +1,628 @@
...
+            self.write(":FORMat:READing:TIME ON")        
+
+    def format_reading_time_disable(self):
+        """
+        This command disables the inclusion of time stamp with data retrieved 
+        by the READ? command, the FETCh? command, or other queries of scan results.
+        """
+        self.write(":FORMat:READing:TIME OFF")
+        while((self.values(":FORMat:READing:TIME?")) == 1):
+            self.write(":FORMat:READing:TIME OFF")
```
CasperSchippers, line 463: I think these (or at least most of these) can be nicely converted into a properties (also holds for a number of other methods, but the ones over here seem like a good starting point): for example `Instrument.measurement` for the measure_xx methods and `Instrument.control` for set_temperature_unit and is_temperature_unit.

pymeasure/instruments/keysight/keysight34980A.py
```diff
@@ -0,0 +1,628 @@
...
+        temp = str(temp_values[0])    
+        return(temp)
+
+    def internal_dmm_enable(self):
+        """
+        This command enables the internal DMM. 
+        """
+        self.write(":INSTrument:DMM:STATe ON")
+        while(self.internal_dmm_state() == 0):
+            self.write(":INSTrument:DMM:STATe ON")
```
CasperSchippers, line 562: I see you often have such a construction; could you explain why this is needed? Is setting a value once not sufficient?

pymeasure/instruments/keysight/keysight34980A.py
```diff
@@ -0,0 +1,628 @@
...
+
+    def measure_frequency(self, chan, resolution = "DEFault", range_param = "DEFault"):    
+        """
+        The following command configures the instrument for frequency measurements, 
+        triggers the internal DMM to scan the specified channel.
+        """
+        temp_values = self.values(":MEASure:SCALar:FREQuency?")
+        return(str(temp_values[0]))
+
+    def measure_period(self, chan, resolution = "DEFault", range_param = "DEFault"):
```
CasperSchippers, line 358: The arguments (chan, resolution and range_param) can be removed, right? they're not used anyway

pymeasure/instruments/keysight/keysight34980A.py
```diff
@@ -0,0 +1,628 @@
...
+        temp_values = self.values(":MEASure:SCALar:TEMPerature?")
+        return(str(temp_values[0]))
+
+    def set_temperature_unit(self, chan = "", unit = 'K'):    
+        """
+        This command selects the temperature units (°C, °F, or Kelvins) on the specified channels. 
+        If you omit the optional <ch_list> parameter, this command applies to the internal DMM, 
+        independent of any channels or a scan list.
+        """
+        if (chan == ""):
```
CasperSchippers, line 387: I would suggest setting the default value to `None` rather than `""` and change this line into `chan is None`; I think it's a bit neater

pymeasure/instruments/keysight/keysight34980A.py
```diff
@@ -0,0 +1,628 @@
...
+        """
+        The following command configures the instrument for temperature measurements on the specified channel.
+        """
+        temp_values = self.values(":MEASure:SCALar:TEMPerature?")
+        return(str(temp_values[0]))
+
+    def set_temperature_unit(self, chan = "", unit = 'K'):    
+        """
+        This command selects the temperature units (°C, °F, or Kelvins) on the specified channels. 
+        If you omit the optional <ch_list> parameter, this command applies to the internal DMM,
```
CasperSchippers, line 384: for methods with arguments it is always nice to have at the end of the docstring a list of the parameters in the following format:
```
:param chan: A optional string parameter that defines the channel for which the temperature unit is set.
:param unit: An optional string parameter that defines which unit to set.
```

pymeasure/instruments/keysight/keysight34980A.py
```diff
@@ -0,0 +1,628 @@
...
+    def is_temperature_unit(self, chan = ""):    
+        """
+        The command checks the temperature units (°C, °F, or Kelvins) on the specified channels. 
+        If you omit the optional <ch_list> parameter, this command applies to the internal DMM, 
+        independent of any channels or a scan list. 
+        """
+        if (chan == ""):
+            self.write(":UNIT:TEMPerature?") 
+            temp_values = self.read()
+            return(str(temp_values))
```
CasperSchippers, line 405: Structures like this (i.e. writing a certain query, reading the response and casting it into a certain type) can be best performed using a `self.ask` (in case you expect a single string to be returned) or `self.values` (in case you expect one or more numeric values to be returned).

pymeasure/instruments/keysight/keysight34980A.py
```diff
@@ -0,0 +1,628 @@
...
+        else:
+            self.write(":UNIT:TEMPerature? (@" + str(chan) + ")") 
+            temp_values = self.read()
+            return(str(temp_values))
```
CasperSchippers, line 409: The fact that the instrument has an internal DMM or a channel for a lot of commands complicates the implementation quite a bit; I'm wondering if there would be a neater way to implement this.

You could consider defining an `Instrument.control` property for the temperature unit of the DMM and have a sub-class for the different channels such that you can call Keysight34980A34980A.ch1.temperature_unit (see for example the KeysightDSOX1102G class where there are 2 channels).

If there are really many channels, you could also consider another construction where you have a sort of channelmanager that manages the channels such that in the end you can call it like Keysight34980A34980A.ch(1).temperature_unit:
```
class Keysight34980A34980A:
    def __init__(self):
        self.ch = ChannelManager()


class ChannelManager:
    def __init__(self):
        self.channel = 0

    def __call__(self, channel):
        self.channel = channel
        return self

    def do_something(self):
        print(f"doing something to channel {self.channel}")


a = Keysight34980A34980A()
a.ch(1).do_something()
a.ch(2).do_something()

# Calling without the number will use the last entered channel number
a.ch.do_something()

# You can't just store a channel, because you constantly refer to the same object
ch1 = a.ch(1)
ch2 = a.ch(2)
ch1.do_something()
# but you can using copy if you'd want
import copy
ch1 = copy.copy(a.ch(1))
ch2 = copy.copy(a.ch(2))
ch1.do_something()
ch2.do_something()
```

pymeasure/instruments/keysight/keysight34980A.py
Neel-Sancheti: It is just a verification to check that the value is set correctly.

CasperSchippers: Do you know if it happens often/at all that the value isn't correctly set? I'd think that it would not be required in general and that it just clutters the communication. Moreover, if there is another reason a certain value cannot be set, it will try again indefinitely.

AidenDawn:
From what I know of this instrument, this pull request probably needs quite a lot of restructuring before it could be merged. The mainframe contains a DMM (or not, I think thats an option) and the control hardware for lots of different submodule types which are mostly relay matrices but also include some digital IO and on the more recent hardware revisions can contain digitizers too. Happy to have a look into this a bit more at some point but I believe storing the channel/module functions inside the main mainframe class would be oversimplifying the complexity of this instrument

bmoneke:
Hey @Neel-Sancheti do you intend to finish this contribution (thanks for opening it)?
If not, we can close it. The code will stay in case you want to continue the work later on.

Neel-Sancheti:
Hello,

Sorry for the delayed response. For now we can close it. I will come back
to it later. Or someone else can continue the work.

Thankyou!

bmoneke:
Thanks for your response.

Closed until someone wants to pick the work up again.

PR #437, opened 2021-04-24
New instrument HP3478A

LongnoseRob:
this adds the HP3478A 5.5 digit multimeter.
As this is a non-SCPI instrument, implementation may look/feel odd.
Also included is an example file using the plotter example

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -32,3 +32,4 @@ Mathieu Plante
 Michele Sardo
 Steven Siegl
 Benjamin Klebel-Knobloch
+Robert Eckelmann
--- a/pymeasure/instruments/hp/hp3478A.py
+++ b/pymeasure/instruments/hp/hp3478A.py
@@ -42,22 +42,42 @@ class HP3478A(Instrument):
 
     As this unit predates SCPI some tricks are required to get this working
     """
-    DIGITS=['n/a','5 digits','4 digits','3 digits']
 
-    MODES=['n/a','DC Volts','AC Volts','Ohms (2-wire)','Ohms (4-wire)',
-           'DC current','AC current','Extended Ohms']
+    DIGITS={3: "N3",
+            4: "N4",
+            5: "N5"
+            }
 
-    RANGES=[['n/a'],
-                ['n/a','30 mV DC','300 mV DC','3 V DC','30 V DC','300 V DC'],
-                ['n/a','300 mV AC','3 V AC','30 V AC','300 V AC'],
-                ['n/a', '30 Ohm', '300 Ohm', '3 kOhm', '30 kOhm', '300 kOhm',
-                 '3 MOhm','30 MOhm'],
-                ['n/a', '30 Ohm', '300 Ohm', '3 kOhm', '30 kOhm', '300 kOhm',
-                 '3 MOhm','30 MOhm'],
-                ['n/a','300 mA DC','3 A DC'],
-                ['n/a','300 mA AC','3 A AC'],
-                ['n/a','extended Ohms']]
+    AZ={"off": "Z0",
+        "on": "Z1"
+        }
 
+    MODES={"DCV": "F1",
+           "ACV": "F2",
+           "R2W": "F3",
+           "R4W": "F4",
+           "DCI": "F5",
+           "ACI": "F6",
+           "Rext": "F7"
+           }
+
+    RANGES={"DCV": {3E-2: "R-2",3E-1: "R-1",3: "R0",30: "R1",300:"R2","auto": "RA"},
+            "ACV": {3E-1: "R-1",3: "R0",30: "R1" ,300: "R2","auto": "RA"},
+            "R2W": {30: "R1",300: "R2", 3E3: "R3",3E4: "R4",3E5: "R5", 3E6: "R6", 3E7: "R7","auto": "RA"},
+            "R4W": {30: "R1",300: "R2", 3E3: "R3",3E4: "R4",3E5: "R5", 3E6: "R6", 3E7: "R7","auto": "RA"},
+            "DCI": {3E-1: "R-1",3: "R0","auto": "RA"},
+            "ACI": {3E-1: "R-1",3: "R0","auto": "RA"},
+            "Rext": {3E6: "R7","auto": "RA"}
+            }
+
+    TRIGGERS={
+        "auto": "T1",
+        "internal": "T1",
+        "external": "T2",
+        "single": "T3",
+        "hold": "T4",
+        "fast": "T5"
+        }
 
     def __init__(self, resourceName, **kwargs):
         super(HP3478A, self).__init__(
@@ -69,7 +89,7 @@ class HP3478A(Instrument):
             **kwargs
         )
 
-    @property
+
     def error_status(self):
         """
         get the Error-status register value
@@ -83,7 +103,6 @@ class HP3478A(Instrument):
         estatus=self.ask('E')
         return estatus
 
-    @property
     def status(self):
         """
         Reads the HP3478A status register (5-bytes)
@@ -94,35 +113,34 @@ class HP3478A(Instrument):
             DESCRIPTION.
 
         """
-        status_read=self.values('B')
+        status_read=self.adapter.connection.query_binary_values("B","b",data_points=5)
         status=[]
         for i in range(0,len(status_read)):
             # print(bin(status_read[i]))
             status.append(status_read[i])
         return status
 
-    @property
-    def status_readable(self):
-        """
-        Delivers a human-readble status based on the status() command
+    # def status_readable(self):
+    #     """
+    #     Delivers a human-readble status based on the status() command
 
-        Returns
-        -------
-        None.
+    #     Returns
+    #     -------
+    #     None.
 
-        """
-        status_read=self.status()
-        d_select=status_read[0]&3
-        r_select=(status_read[0]>>2)&7
-        m_select=(status_read[0]>>5)&7
-        print(RANGES[m_select][r_select])
-        print(DIGITS[d_select])
-        print(MODES[m_select])
-        #TODO: add decodung on status byte, e.g. AZ on/off, Front/rear selection
-        print('General status: ',status_read[1])
-        print('srq mask: ',status_read[2])
-        print('Error status: ',status_read[3])
-        print('DAC value: ',status_read[4])
+    #     """
+    #     status_read=self.status()
+    #     d_select=status_read[0]&3
+    #     r_select=(status_read[0]>>2)&7
+    #     m_select=(status_read[0]>>5)&7
+    #     print(RANGES[m_select][r_select])
+    #     print(DIGITS[d_select])
+    #     print(MODES[m_select])
+    #     #TODO: add decodung on status byte, e.g. AZ on/off, Front/rear selection
+    #     print('General status: ',status_read[1])
+    #     print('srq mask: ',status_read[2])
+    #     print('Error status: ',status_read[3])
+    #     print('DAC value: ',status_read[4])
 
     @property
     def auto_zero(self):
@@ -163,7 +181,6 @@ class HP3478A(Instrument):
             if on_off==1:
                 self.write('Z1')
 
-    @property
     def active_connectors(self):
         """
         checks the status of the front/back switch, selecting either the front
@@ -181,8 +198,8 @@ class HP3478A(Instrument):
         front_back=status_read[1]>>4&1
         return front_back
 
-    
-    def measure(self,mode='DC Volts',measurement_range='auto',auto_zero=1,digits=5, trigger='auto'):
+
+    def measure(self,mode='DCV',measurement_range='auto',auto_zero="on",digits=5, trigger='auto'):
         """
         returens a measurement result, depeding on the parameters
         see also the other measure_ functions for the HP3478A.
@@ -208,38 +225,32 @@ class HP3478A(Instrument):
             DESCRIPTION.
 
         """
-        measurement_string='F'
+        measurement_string=''
         if mode in self.MODES:
-            measurement_string=measurement_string+str(self.MODES.index(mode))
+            measurement_string=self.MODES[mode]
         else:
             raise Exception('Mode string not supported')
-        
-        if measurement_range=='auto':
-            measurement_string=measurement_string+('RA')
-            
-        if measurement_range in self.RANGES[self.MODES.index(mode)]:
-            measurement_string=measurement_string+'R'
-            if self.MODES.index(mode)==1:
-                measurement_string=measurement_string+str(self.RANGES[self.MODES.index(mode)].index(measurement_range)-3)
-            if self.MODES.index(mode)==2:
-                measurement_string=measurement_string+str(self.RANGES[self.MODES.index(mode)].index(measurement_range)-2)
-            if self.MODES.index(mode)==3:
-                measurement_string=measurement_string+str(self.RANGES[self.MODES.index(mode)].index(measurement_range)-1)
-            if self.MODES.index(mode)==4:
-                measurement_string=measurement_string+str(self.RANGES[self.MODES.index(mode)].index(measurement_range)-1)
-            if self.MODES.index(mode)==5:
-                measurement_string=measurement_string+str(self.RANGES[self.MODES.index(mode)].index(measurement_range)-2)
-            if self.MODES.index(mode)==6:
-                measurement_string=measurement_string+str(self.RANGES[self.MODES.index(mode)].index(measurement_range)-2)                
-            if self.MODES.index(mode)==7:
-                raise Exception("nbot yet implemented")
-                
-        # measurement_string='F1RAN5T1Z1'
-        measured_value=self.ask(measurement_string)
+
+        if measurement_range in self.RANGES[mode]:
+            measurement_string=measurement_string+self.RANGES[mode][measurement_range]
+
+        if auto_zero in self.AZ:
+            measurement_string=measurement_string+self.AZ[auto_zero]
+
+        if digits in self.DIGITS:
+            measurement_string=measurement_string+self.DIGITS[digits]
+
+        if trigger in self.TRIGGERS:
+            measurement_string=measurement_string+self.TRIGGERS[trigger]
+
+        # print(measurement_string)
+        measured_value=float(self.ask(measurement_string))
         return measured_value
-    
-    
+
+
     def display_reset(self):
+
+
         """
         Resets the display of the HP3478A to the normal measurement dusplay.
         Especially helpful after using D2 or D3 command
@@ -250,5 +261,14 @@ class HP3478A(Instrument):
 
         """
         self.write('D1')
-        
-        
+
+    def reset(self):
+        self.adapter.connection.clear()
+
+    def close(self):
+        self.adapter.connection.close()
+
+    def shutdown(self):
+        self.adapter.connection.clear()
+        self.adapter.connection.close()
+
--- a/pymeasure/instruments/hp/hp3478A.py
+++ b/pymeasure/instruments/hp/hp3478A.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2021 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+
+import logging
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set
+
+
+class HP3478A(Instrument):
+    """ Represents the Hewlett Packard 3748A 5 1/2 digit multimeter
+    and provides a high-level interface for interacting
+    with the instrument.
+    
+    As this unit predates SCIP some tricks are required to get this working
+    """
+    DIGITS=['n/a','5 digits','4 digits','3 digits']
+    
+    MODES=['n/a','DC Volts','AC volts','Ohms (2-wire)','Ohms (4-wire)',
+           'DC current','AC current','Extended Ohms']
+    
+    RANGES=[['n/a'],
+                ['n/a','30 mV DC','300 mV DC','3 V DC','30 V DC','300 V DC'],
+                ['n/a','300 mV AC','3 V AC','30 V AC','300 V AC'],
+                ['n/a', '30 Ohm', '300 Ohm', '3 kOhm', '30 kOhm', '300 kOhm', 
+                 '3 MOhm','30 MOhm'],
+                ['n/a', '30 Ohm', '300 Ohm', '3 kOhm', '30 kOhm', '300 kOhm', 
+                 '3 MOhm','30 MOhm'],
+                ['n/a','300 mA DC','3 A DC'],
+                ['n/a','300 mA AC','3 A AC'],
+                ['n/a','extended Ohms']]
+    
+    def __init__(self, resourceName, **kwargs):
+        super(HP3478A, self).__init__(
+            resourceName,
+            "Hewlett-Packard HP3478A",
+            includeSCPI=False,
+            send_end=True,
+            read_termination="\r\n",
+            **kwargs
+        )
--- a/pymeasure/instruments/hp/hp3478A.py
+++ b/pymeasure/instruments/hp/hp3478A.py
@@ -113,34 +113,37 @@ class HP3478A(Instrument):
             DESCRIPTION.
 
         """
-        status_read=self.adapter.connection.query_binary_values("B","b",data_points=5)
+        # status_read=self.adapter.connection.query_binary_values("B","b",data_points=5)
+        self.write("B")
+        status_read=self.adapter.connection.read_bytes(5)
         status=[]
         for i in range(0,len(status_read)):
             # print(bin(status_read[i]))
             status.append(status_read[i])
         return status
 
-    # def status_readable(self):
-    #     """
-    #     Delivers a human-readble status based on the status() command
+    def status_readable(self):
+        """
+        Delivers a human-readble status based on the status() command
 
-    #     Returns
-    #     -------
-    #     None.
+        Returns
+        -------
+        None.
 
-    #     """
-    #     status_read=self.status()
-    #     d_select=status_read[0]&3
-    #     r_select=(status_read[0]>>2)&7
-    #     m_select=(status_read[0]>>5)&7
-    #     print(RANGES[m_select][r_select])
-    #     print(DIGITS[d_select])
-    #     print(MODES[m_select])
-    #     #TODO: add decodung on status byte, e.g. AZ on/off, Front/rear selection
-    #     print('General status: ',status_read[1])
-    #     print('srq mask: ',status_read[2])
-    #     print('Error status: ',status_read[3])
-    #     print('DAC value: ',status_read[4])
+        """
+        status_read=self.status()
+        d_act=6-(status_read[0]&3)
+        
+        m_act=list(self.MODES.keys())[((status_read[0]>>5)&7)-1]
+        r_act=list(self.RANGES[m_act].keys())[(((status_read[0]>>2)&7)-1)]
+        #TODO improve the status display 
+        print('The actual range is '+str(r_act)+' '+m_act+' with '+str(d_act)+' 1/2 digits')
+            
+        #TODO: add decodung on status byte, e.g. AZ on/off, Front/rear selection
+        print('General status: ',status_read[1])
+        print('srq mask: ',status_read[2])
+        print('Error status: ',status_read[3])
+        print('DAC value: ',status_read[4])
 
     @property
     def auto_zero(self):
--- a/pymeasure/instruments/hp/hp3478A.py
+++ b/pymeasure/instruments/hp/hp3478A.py
@@ -63,8 +63,10 @@ class HP3478A(Instrument):
 
     RANGES={"DCV": {3E-2: "R-2",3E-1: "R-1",3: "R0",30: "R1",300:"R2","auto": "RA"},
             "ACV": {3E-1: "R-1",3: "R0",30: "R1" ,300: "R2","auto": "RA"},
-            "R2W": {30: "R1",300: "R2", 3E3: "R3",3E4: "R4",3E5: "R5", 3E6: "R6", 3E7: "R7","auto": "RA"},
-            "R4W": {30: "R1",300: "R2", 3E3: "R3",3E4: "R4",3E5: "R5", 3E6: "R6", 3E7: "R7","auto": "RA"},
+            "R2W": {30: "R1",300: "R2", 3E3: "R3",3E4: "R4",3E5: "R5", 3E6: "R6",
+                    3E7: "R7","auto": "RA"},
+            "R4W": {30: "R1",300: "R2", 3E3: "R3",3E4: "R4",3E5: "R5", 3E6: "R6",
+                    3E7: "R7","auto": "RA"},
             "DCI": {3E-1: "R-1",3: "R0","auto": "RA"},
             "ACI": {3E-1: "R-1",3: "R0","auto": "RA"},
             "Rext": {3E6: "R7","auto": "RA"}
@@ -117,7 +119,7 @@ class HP3478A(Instrument):
         self.write("B")
         status_read=self.adapter.connection.read_bytes(5)
         status=[]
-        for i in range(0,len(status_read)):
+        for i in range(0,4):
             # print(bin(status_read[i]))
             status.append(status_read[i])
         return status
@@ -133,15 +135,41 @@ class HP3478A(Instrument):
         """
         status_read=self.status()
         d_act=6-(status_read[0]&3)
-        
         m_act=list(self.MODES.keys())[((status_read[0]>>5)&7)-1]
+        if m_act=='DCV':
+            m_str='V'
+        if m_act=='ACV':
+            m_str='V'
+        if m_act=='DCI':
+            m_str='A'
+        if m_act=='ACI':
+            m_str='A'
+        if m_act=='R2W':
+            m_str='Ohm'
+        if m_act=='R4W':
+            m_str='Ohm'
+        if m_act=='Rext':
+            m_str='Ohm'
+
         r_act=list(self.RANGES[m_act].keys())[(((status_read[0]>>2)&7)-1)]
-        #TODO improve the status display 
-        print('The actual range is '+str(r_act)+' '+m_act+' with '+str(d_act)+' 1/2 digits')
-            
-        #TODO: add decodung on status byte, e.g. AZ on/off, Front/rear selection
-        print('General status: ',status_read[1])
-        print('srq mask: ',status_read[2])
+        if r_act <= 1:
+            r_str=str(r_act*1000)+' m'
+        if r_act >=1E6:
+            r_str=str(r_act/1E6)+' M'
+        if r_act >=1000:
+            r_str=str(r_act/1000)+' k'
+        print('----- HP3478 ----- Status -----')
+        print('The current range is '+r_str+m_str+' with '+str(d_act)+' 1/2 digits')
+
+        print('General status: ')
+        print('     Internal trigger: '+str(status_read[1]&1))
+        print('     Autorange: '+str(status_read[1]>>1&1))
+        print('     Auto Zero: '+str(status_read[1]>>2&1))
+        print('     50 Hz opereration: '+str(status_read[1]>>3&1))
+        print('     Front/Rear: '+str(status_read[1]>>4&1))
+        print('     CalibrationRAM en: '+str(status_read[1]>>6&1))
+        print('     External trigger: '+str(status_read[1]>>7&1))
+        print('Srq mask: ',status_read[2])
         print('Error status: ',status_read[3])
         print('DAC value: ',status_read[4])
 
@@ -200,12 +228,33 @@ class HP3478A(Instrument):
         status_read=self.status()
         front_back=status_read[1]>>4&1
         return front_back
+    
+    @property
+    def trigger(self):
+        status_read=self.status()
+        int_trig=status_read[1]&1
+        ext_trig=status_read[1]>>7&1
+        if int_trig==1 and ext_trig==0:
+            trigger_stat='internal'
+        if int_trig==0 and ext_trig==1:
+            trigger_stat='external'
+        if int_trig==0 and ext_trig==0:
+            trigger_stat='hold'
+        return trigger_stat
+    
+    @trigger.setter
+    def trigger(self, trigger_mode):
+        if trigger_mode in self.TRIGGERS.keys():
+            self.write(self.TRIGGERS[trigger_mode])
+        else:
+            print("fix me")
+      
 
 
     def measure(self,mode='DCV',measurement_range='auto',auto_zero="on",digits=5, trigger='auto'):
         """
         returens a measurement result, depeding on the parameters
-        see also the other measure_ functions for the HP3478A.
+        s
 
 
 
@@ -250,7 +299,6 @@ class HP3478A(Instrument):
         measured_value=float(self.ask(measurement_string))
         return measured_value
 
-
     def display_reset(self):
 
 
@@ -274,4 +322,3 @@ class HP3478A(Instrument):
     def shutdown(self):
         self.adapter.connection.clear()
         self.adapter.connection.close()
-
--- a/pymeasure/instruments/hp/hp3478A.py
+++ b/pymeasure/instruments/hp/hp3478A.py
@@ -27,7 +27,7 @@
 import logging
 
 from pymeasure.instruments import Instrument
-from pymeasure.instruments.validators import strict_discrete_set
+
 
 
 log = logging.getLogger(__name__)
@@ -228,9 +228,18 @@ class HP3478A(Instrument):
         status_read=self.status()
         front_back=status_read[1]>>4&1
         return front_back
-    
+
     @property
     def trigger(self):
+        """
+        Checks the triger settings on the HP3478
+
+        Returns
+        -------
+        trigger_stat : str
+            A string value describing the current status of the trigger
+
+        """
         status_read=self.status()
         int_trig=status_read[1]&1
         ext_trig=status_read[1]>>7&1
@@ -241,14 +250,28 @@ class HP3478A(Instrument):
         if int_trig==0 and ext_trig==0:
             trigger_stat='hold'
         return trigger_stat
-    
+
     @trigger.setter
     def trigger(self, trigger_mode):
+        """
+        Offers the possibility to set the trigger mode
+
+        Parameters
+        ----------
+        trigger_mode : str
+            allowed values are:
+                 auto, internal, external, single, hold or fast
+
+        Returns
+        -------
+        None.
+
+        """
         if trigger_mode in self.TRIGGERS.keys():
             self.write(self.TRIGGERS[trigger_mode])
         else:
             print("fix me")
-      
+
 
 
     def measure(self,mode='DCV',measurement_range='auto',auto_zero="on",digits=5, trigger='auto'):
@@ -314,11 +337,35 @@ class HP3478A(Instrument):
         self.write('D1')
 
     def reset(self):
+        """
+        Initatiates a reset of the HP3478A
+
+        Returns
+        -------
+        None.
+
+        """
         self.adapter.connection.clear()
 
     def close(self):
+        """
+        close the current connection to the HP3478A
+
+        Returns
+        -------
+        None.
+
+        """
         self.adapter.connection.close()
 
     def shutdown(self):
+        """
+        provides a way to gravcefully close the conention to the HP3478A
+
+        Returns
+        -------
+        None.
+
+        """
         self.adapter.connection.clear()
         self.adapter.connection.close()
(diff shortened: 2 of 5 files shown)


msmttchr (commented):
I think documentation should follow Sphinx docstring format https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html
I see a lot of template documentation not filled, eg:
```
Returns
        -------
        estatus : TYPE
            DESCRIPTION.
```
Some commented out code still present, e.g:
`#status_read=self.adapter.connection.query_binary_values("B","b",data_points=5)`

I am not sure whether having `print` statement in an instrument driver is appropriate. Possibly you can have an API that return a printable string and leave user application decide what to do with the string.

LongnoseRob:
@msmttchr thank you for the feedback and comments.
I must have overlooked to edit these docstrings before issuing the PR.

regarding the print statements, yes you are right, I'll rework the function and will make it return a string.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,389 @@
...
+        -------
+        None.
+
+        """
+        if on_off==0:
+            self.write('Z0')
+        else:
+            if on_off==1:
+                self.write('Z1')
+
```
msmttchr, line 216: What about self.write('Z{}'.format(on_off)) ?

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,389 @@
...
+        Returns
+        -------
+        None.
+
+        """
+        if trigger_mode in self.TRIGGERS.keys():
+            self.write(self.TRIGGERS[trigger_mode])
+        else:
+            print("fix me")
+
```
msmttchr, line 276: Is it a debug `print` left by mistake? Is it better to raise an exception ? or simply use the code:
`self.write(self.TRIGGERS[trigger_mode])`

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,389 @@
...
+        trigger : str, optional
+            Trigger mode selection, choices are: auto=internal, external, single, hold, fast (see manual for this)
+            
+            Default is 'auto'.
+
+        Returns
+        -------
+        measured_value : float
+            the value mesured based on the mode and other settings.
+
```
msmttchr, line 319: I am not familiar with this style of comments, I have seen that in the resto of the code the style used is as follow:
"""[Summary]

:param [ParamName]: [ParamDescription], defaults to [DefaultParamVal]
:type [ParamName]: [ParamType](, optional)
...
:raises [ErrorType]: [ErrorDescription]
...
:return: [ReturnDescription]
:rtype: [ReturnType]
"""
E.g.:
```
        """Formats a record as csv.
        :param record: record to format.
        :type record: dict
        :return: a string
        """
```

Are you sure the style you are using renders well with Sphinx ?

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,389 @@
...
+        Returns
+        -------
+        estatus : int
+            error status register value (confirm with manual for detail)
+
+        """
+        estatus=self.ask('E')
+        return estatus
+
```
msmttchr, line 107: Is it possible to implement the commands using the `Instrument.control/setting/measurement` class methods ?
For example, does the following works ?
```
    error_status = Instrument.measurement(
        "E",
        """ This property returns the Error-status register value""",
    )

```

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,389 @@
...
+        """
+        status_read=self.status()
+        front_back=status_read[1]>>4&1
+        return front_back
+
+    @property
+    def trigger(self):
+        """
+        Checks the triger settings on the HP3478
+
```
msmttchr, line 238: trigger is misspelled

LongnoseRob: Thank you, will be changed with the next commit

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Thank you very much, i have changed this to an exception

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: That was a good hint!
Thank you very much!

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: You were absolutely right, the way you descibe is the right way to do it.
I was just using the docstring template from spyder, it acutally compiled with sphinx, but there were some warnings.
I have changed the parmeters(&other descriptions).

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: This might work, need to dig into this a bit more

LongnoseRob: found that this works
'    error_status = Instrument.measurement(
        "E",
        """Checks the error status register""",
        get_process=(lambda x: int(x)),
        )
'

actually I am thinking to change the name to check_errors

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,316 @@
...
+        """Checks the error status register""",
+        get_process=(lambda x: int(x)),
+        )
+
+    status = Instrument.measurement(
+        "B",
+        """Checks the status registers""",
+        get_process=(lambda x: bytes(x.encode(encoding="ASCII"))),
+        )
+
```
msmttchr, line 96: `encode` returns `bytes` , why `bytes` is needed ?

In a wider review context, I would like to propose to access the status bits returned by command "B" in a more structured way. In this mode, your code is probably slower, but much more readable, in my opinion.
My proposal is to define a static method which decodes the bytes returned by "B" command through the use of ctypes and bitfield (see attached examples [test.zip](https://github.com/pymeasure/pymeasure/files/6411976/test.zip))

Let's call the static method `get_status(status, field=None)` with the following behaviour:
- `get_status(status, None)` return all the fields
- `get_status(status, "auto_zero")` return "auto_zero" field

Advantages:
- code is more readable
- `__str__` call can replace `status_readable` utility
- `status get_process` is defined as follow:` get_process=(lambda x: self.get_status(x.encode(encoding="ASCII")))`
- `auto_zero` can be rewritten as follow:
```
auto_zero = Instrument.control(
        "B",
        "Z%d",
        """ Return autozero settings on the HP3478, this property can be set """,
        get_process=(lambda x: self.get_status(x.encode(encoding="ASCII"), "auto_zero")),
        values=(0,1),
        )
```
- similar approach also for other property and and methods in the class, when possible
- Using `map_value` and dictionaries allow to use symbolic names as well (see `https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html`), so probably your dictionaries `MODES`, `RANGES`, etc. can be reused.
- Align code to class `Instrument` style and make it probably more homogenous to the rest of instruments.

Sorry, probably a lot of changes, but if you think are useful, it could be worth to make them.

LongnoseRob: @msmttchr  Thank you again for the code review, I really appreciate your support in this matter.

> encode returns bytes , why bytes is needed ?


The 'B' command returns 5 bytes which are not ment to be read as an ASCII encoding.
The Intstrument.control()/messurement()/setting() functions however do not offer a passing of binary values like Instrument.binary_read() does, this is why I used the encode call.
For reference I attach a exerpt of the manual 
3478A_manual_pg_3-14

> In a wider review context, I would like to propose to access the status bits returned by command "B" in a more structured way. In this mode, your code is probably slower, but much more readable, in my opinion.
My proposal is to define a static method which decodes the bytes returned by "B" command through the use of ctypes and bitfield (see attached examples test.zip)

I like your idea of a static method to decode the bits from the status bytes.
Speed is not an issue here, as the status of the instrument is normally not queried very quickly.
And btw this instrument is anyway not the fastest (being from the early '80ties)
Where speed was concerned was the measure routine, which now uses caching of the command to be sent.

Looking at your proposal with the ctypes, I have been looking at enum over the last days and am working on that route at the moment, I have attached an example 
[enum_test.zip](https://github.com/pymeasure/pymeasure/files/6412715/enum_test.zip)
(The function and range encoding still gives me some headache, as the bit-assignment is not unique)


**General question:** is it save to assume that a python version >3.6 will be used? Or is there is a feature set we should respect.
For my private use I am at Pyhton 3.9.4 (ArchLinux/ArchLinuxArm) and for my workstuff typically also on >3.8(WInpython), but I dont know if this can be expected for all pymeasure users.

msmttchr: @LongnoseRob, possibly I was too crytpic in my comment, here I meant that the code below is equivalent:
- `get_process=(lambda x: bytes(x.encode(encoding="ASCII"))),`
- `get_process=(lambda x: x.encode(encoding="ASCII")),`

In other word, I am not sure if the `bytes` instantiation is redundant since after call to method `encode` you have already `bytes` class data.

Regarding whether it is safe to assume on python version >3.6, from what I see in the CI tests that are run after a commit, the python version used are: 3.6, 3.7, 3.8, 3.9. In addition to this, file [setup.py](https://github.com/pymeasure/pymeasure/blob/master/setup.py) mention the same python versions.

I would assume then that python version >= 3.6.

AUTHORS.txt
```diff
@@ -32,3 +32,4 @@ Mathieu Plante
 Michele Sardo
 Steven Siegl
 Benjamin Klebel-Knobloch
+Robert Eckelmann
```
CasperSchippers, line 35: Don't forget to also add yourself to the second authors in the docs (docs/about/authors.rst)

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,428 @@
+# -*- coding: utf-8 -*-
```
CasperSchippers, line 1: Maybe remove this line; it doesn't really hurt, but it is not common to have this line in any of the other pymeasure files

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,428 @@
...
+
+    """
+    ret_val = Status(Status_bytes(*status_bytes))
+    if field is None:
+        return ret_val.b
+    else:
+        if field == "SRQ":
+            return getattr(ret_val.B, "byte3")
+        else:
+            return getattr(ret_val.b, field)
```
CasperSchippers, line 180: Can you not equally well have a single if - elif - else statement rather than a nested if-else statement?

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Will remove this line,

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Thank you, looking at the code again after a while, this make totally sense and looks better.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+
+    #TODO: fix validators
+    range = Instrument.control(
+        "B",
+        "%s",
+        """Returns the current measurment range, this property can be set""",
+        get_process = (lambda x: get_range(x.encode(encoding="ASCII"))),
+        set_process = (lambda x,y: RANGES[x][y]),
+        # validator = joined_validators(strict_discrete_set,strict_discrete_set),
+        # values = RANGES,
```
CasperSchippers, line 368: When using the `joined_validators`, the the argument for `values` is also expected to be a list that is equally long as the list of validators you pass to `joined_validators` (i.e. 2 long in this case)

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+        "B",
+        "%s",
+        """Return current selected trigger mode, this property can be set""",
+        get_process = (lambda x: get_trigger(x.encode(encoding="ASCII"))),
+        set_process = (lambda x: TRIGGERS[x]),
+        validator = strict_discrete_set,
+        values = TRIGGERS,
+        )
+
+    #Functions using low-level access via instrument.adapter.conneciton methods
```
CasperSchippers, line 405: typo in `connection`

CasperSchippers (commented):
Thanks for the contribution; I've had a look and left some remarks

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Thank you, will be fixed with the next submit

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+        set_process = (lambda x: str.upper(x[0:12])),
+        )
+
+    disp_text_no_symbol = Instrument.setting(
+        "D3%s",
+        """Displays up to 12 upper-case ASCII characters on the display and disables all symbols on the display.""",
+        set_process = (lambda x: str.upper(x[0:12])),
+        )
+
+    id = Exception(ValueError('Not implemented in this pre-SCPI instrument'))
```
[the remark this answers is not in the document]
LongnoseRob: WIll remove it

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+def get_trigger(status_bytes):
+    """Method to decode trigger mode
+    :param status_bytes:   list of bytes to be decoded
+    :return trigger_mode: string with the current trigger mode
+    :rtype trigger_mode: str
+
+    """
+    cur_stat= Status(Status_bytes(*status_bytes))
+    i_trig = cur_stat.b.int_trig
+    e_trig = cur_stat.b.int_trig
```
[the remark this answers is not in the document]
LongnoseRob: Thanks, of course that was a copy-paste-noedit mistake..

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+            includeSCPI = False,
+            send_end = True,
+            read_termination = "\r\n",
+            **kwargs
+        )
+
+
+    active_connectors = Instrument.measurement(
+        "B",
+        """Return selected connectors (1->front/0->back)), based on front-panel selector switch""",
```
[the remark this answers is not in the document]
LongnoseRob: The instrument only returns 0/1, so I will add the mapping to this and similiar commands

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+        values = (0,1),
+        )
+
+    cal_enable = Instrument.measurement(
+        "B",
+        """Return calibration enable switch setting (1->CAL / 0->disbaled), based on front-panel selector switch""",
+        get_process = (lambda x: get_status(x.encode(encoding="ASCII"), "cal_enable")),
+        )
+
+    check_errors = Instrument.measurement(
```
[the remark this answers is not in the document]
LongnoseRob: No problem, check_errors will be made a method and return error_status

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+        "Z%d",
+        """ Return autozero settings on the HP3478, this property can be set (0: disabled, 1: enabled) """,
+        get_process = (lambda x: get_status(x.encode(encoding="ASCII"), "auto_zero")),
+        validator = strict_discrete_set,
+        values = (0,1),
+        )
+
+    cal_enable = Instrument.measurement(
+        "B",
+        """Return calibration enable switch setting (1->CAL / 0->disbaled), based on front-panel selector switch""",
```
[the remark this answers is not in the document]
LongnoseRob: Mapping implemented

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+    auto_range = Instrument.measurement(
+        "B",
+        """ Return auto-ranging status """,
+        get_process = (lambda x: get_status(x.encode(encoding="ASCII"), "auto_range")),
+        )
+
+    auto_zero = Instrument.control(
+        "B",
+        "Z%d",
+        """ Return autozero settings on the HP3478, this property can be set (0: disabled, 1: enabled) """,
```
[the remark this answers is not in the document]
LongnoseRob: as above, mapping will be used from now on
Thanks for the recommendation

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+    measure_Rext = Instrument.measurement(
+        MODES["Rext"],
+        """Return the measured value for extended resistance mode (>30M, 2-wire) resistance""",
+        get_process = (lambda x: float(x)),
+        )
+
+    mode = Instrument.control(
+        "B",
+        "%s",
+        """Return current selected measurment mode, this propery can be set""",
```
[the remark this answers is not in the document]
LongnoseRob: Will have a close look at the doc strigs and add the allowed values

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,427 @@
...
+        }
+
+TRIGGERS={
+        "auto": "T1",
+        "internal": "T1",
+        "external": "T2",
+        "single": "T3",
+        "hold": "T4",
+        "fast": "T5",
+        }
```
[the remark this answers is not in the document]
LongnoseRob: Ill move them into the actual class.

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: For the time being I have removed the validators complete from the range property.
My problem with this property is that it is depending on the measurement mode, I have added a set-function as an intermediate, but I am not happy with that yet.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -361,17 +394,20 @@ def __init__(self, resourceName, **kwargs):
     range = Instrument.control(
```
CasperSchippers, line 394: You could also consider using a slightly different approach: you can also write a property where you query the present mode to the instrument and use that to translate the provided range (can be a float in that case as the user doesn't need to pass the mode to the property) to the correct command; see for example the sensitivity property in the DPS7265: https://github.com/pymeasure/pymeasure/blob/78c5ed3bed5ed6ffcbabadeeb13bbc6feb6f5ee7/pymeasure/instruments/signalrecovery/dsp7265.py#L171-L197

Maybe you can also achieve this in a get and set method rather than by writing a complete own property

CasperSchippers (changes requested):

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -247,17 +234,51 @@ def __init__(self, resourceName, **kwargs):
...
+    RANGES={"DCV": {3E-2: "R-2", 3E-1: "R-1",3: "R0",30: "R1",300:"R2",
+                    "auto": "RA"},
+            "ACV": {3E-1: "R-1", 3: "R0",30: "R1" ,300: "R2","auto": "RA"},
+            "R2W": {30: "R1", 300: "R2", 3E3: "R3",3E4: "R4",3E5: "R5",
+                    3E6: "R6", 3E7: "R7","auto": "RA"},
+            "R4W": {30: "R1",300: "R2", 3E3: "R3",3E4: "R4",3E5: "R5",
+                    3E6: "R6", 3E7: "R7","auto": "RA"},
+            "DCI": {3E-1: "R-1",3: "R0","auto": "RA"},
+            "ACI": {3E-1: "R-1",3: "R0","auto": "RA"},
+            "Rext": {3E6: "R7","auto": "RA"},
```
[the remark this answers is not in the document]
LongnoseRob: Thank you for pointing out my mistake,
Actually for the extended Ohm range it should be according to the manual: 
`"Rext": {3E7: "R7","auto": "RA"},`

Which make absolutly sense, as the extended Ohm function is used with an external parallel resistor (10M) to measure values >30M.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -247,17 +234,51 @@ def __init__(self, resourceName, **kwargs):
...
+            "internal": "T1",
+            "external": "T2",
+            "single": "T3",
+            "hold": "T4",
+            "fast": "T5",
+            }
 
     active_connectors = Instrument.measurement(
         "B",
         """Return selected connectors (1->front/0->back)), based on front-panel selector switch""",
```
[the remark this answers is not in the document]
LongnoseRob: Yes of course

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -266,16 +287,28 @@ def __init__(self, resourceName, **kwargs):
...
 
-    check_errors = Instrument.measurement(
+    def check_errors(self):
+        """
+        Method to read the error status register
+
+        :return error_status: one byte with the error status register content
+        :rtype error_status: int
+        """
+        return self.error_status
```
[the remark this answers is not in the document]
LongnoseRob: Good point,
Of the 8bit in the error byte only 6 are actually used, so this was quick to integrate.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -227,14 +196,32 @@ def get_trigger(status_bytes):
...
+    :return ranges_str: string indicating the range to the instrument
+    :rtype ranges_str: str
+    
+    """
+    mo = r_string.split(sep=",")[0]
+    ra = str.strip(r_string.split(sep=",")[1])
+    if ra == "auto":
+        rr = ra
+    else:
+        rr = int(ra)
```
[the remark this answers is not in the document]
LongnoseRob: Based on your comment further down, I will very probably go this way.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -247,17 +234,51 @@ def __init__(self, resourceName, **kwargs):
...
         """Return selected connectors (1->front/0->back)), based on front-panel selector switch""",
         get_process = (lambda x: get_status(x.encode(encoding="ASCII"), "front_rear")),
+        values={"back":0, "front":1},
+        map_values = True,
         )
 
     auto_range = Instrument.measurement(
         "B",
-        """ Return auto-ranging status """,
+        """ Return auto-ranging status, range can be set the range property""",
```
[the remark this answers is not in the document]
LongnoseRob: done

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: I changed the range property to use instrument.measurment and added a setter-function.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,491 @@
...
+        )
+
+    disp_text_no_symbol = Instrument.setting(
+        "D3%s",
+        """Displays up to 12 upper-case ASCII characters on the display and disables all symbols on the display.
+        """,
+        set_process = (lambda x: str.upper(x[0:12])),
+        )
+
+    #TODO: find a good way to cache the latest measurment and use it to incease speed (if actually needed)
```
CasperSchippers, line 341: I'm inclined to think that caching the measurement result is not something that should be done within the instrument implementation, because in the end that is up to the user to decide whether and when a cached result should be used, or do you disagree?

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,491 @@
...
+    disp_reset = Instrument.setting(
+        "D%d",
+        """Resets the display of the instrument if 1 is written to it.
+        """,
+        set_process = (lambda x: int(x)),
+        validator = strict_discrete_set,
+        values = [1],
+        )
+
+    #TODO: find a proper validator for the next two methods (without using regex, if possible)
```
CasperSchippers, line 326: What kind of validator do you want to implement (in a broad sense, so what is it that you want to check)

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,491 @@
...
+            )
+
+class Status(ctypes.Union):
+    """Union type element for the decoding of the status bit-fields"""
+    _fields_ = [
+        ("B", Status_bytes),
+        ("b", Status_bits)
+    ]
+
+# @staticmethod does not work ?!
```
CasperSchippers, line 133: That's a bit strange, I would expect this would work just fine as a staticmethod; can you tell me what happens?

CasperSchippers (commented):
I've had another look and left some comments. Thanks for the continued improvements.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -306,17 +302,22 @@ def check_errors(self):
         :return error_status: one byte with the error status register content
         :rtype error_status: int
         """
+
+        if self.error_status != 0:
+            log.critical("HP3478A error detected: $s", self.ERRORS[self.error_status] )
```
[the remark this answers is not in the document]
LongnoseRob: Changed to log.error

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -306,17 +302,22 @@ def check_errors(self):
...
         )
 
     disp_reset = Instrument.setting(
         "D%d",
-        """Resets the display of the instrument if 1 is written to it.""",
+        """Resets the display of the instrument if 1 is written to it.
+        """,
         set_process = (lambda x: int(x)),
         validator = strict_discrete_set,
         values = [1],
```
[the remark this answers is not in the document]
LongnoseRob: Yes makes more sense, will change it

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -390,23 +400,28 @@ def check_errors(self):
...
         get_process = (lambda x: get_range(x.encode(encoding="ASCII"))),
-        set_process = (lambda x: set_range(x)),
         )
 
+    @range.setter
+    def range(self,value):
+        cur_mode = self.mode
+        set_range = strict_discrete_set(value, self.RANGES[cur_mode])
+        set_range = self.RANGES[cur_mode][value]
+        print(set_range)
```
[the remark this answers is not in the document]
LongnoseRob: yes ,that was a left over from debugging, removed

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -176,7 +176,6 @@ def get_range(status_bytes):
     cur_range = list(HP3478A.RANGES[cur_mode].keys())[rnge-1]
     return cur_range
 
-
 def get_trigger(status_bytes):
     """Method to decode trigger mode
     :param status_bytes:   list of bytes to be decoded
```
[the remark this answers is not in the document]
LongnoseRob: done

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,491 @@
...
+    active_connectors = Instrument.measurement(
+        "B",
+        """Return selected connectors ("front"/"back"), based on front-panel selector switch
+        """,
+        get_process = (lambda x: get_status(x.encode(encoding="ASCII"), "front_rear")),
+        values={"back":0, "front":1},
+        map_values = True,
+        )
+
+    auto_range = Instrument.measurement(
```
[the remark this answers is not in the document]
LongnoseRob: yes, 
changed

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,491 @@
...
+        """ Returns autozero settings on the HP3478, this property can be set 
+        (False: disabled, True: enabled) 
+        """,
+        get_process = (lambda x: get_status(x.encode(encoding="ASCII"), "auto_zero")),
+        validator = strict_discrete_set,
+        values={False:0, True:1},
+        map_values = True,
+        )
+
+    cal_enable = Instrument.measurement(
```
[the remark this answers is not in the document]
LongnoseRob: I have changed the nameing to longer wording.

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Yes, 
I remove this TODO

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: The insturment allows for text to be shown on the display only tall ASCII-letter, numbers, colon and dot.
My Idea was to have some kind of validator to check for this.
The set process with  str.upper() is already taking care of this, so I willr emove this comment for the time being.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,491 @@
...
+        "B",
+        """ Return auto-ranging status, returns False if manual range and True if auto-range active.
+        For manual range control the range property can be set
+        """,
+        get_process = (lambda x: get_status(x.encode(encoding="ASCII"), "auto_range")),
+        values={False:0, True:1},
+        map_values = True,
+        )
+
+    auto_zero = Instrument.control(
```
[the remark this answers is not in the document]
LongnoseRob: naming changed

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -325,56 +326,65 @@ def check_errors(self):
...
         set_process = (lambda x: str.upper(x[0:12])),
         )
 
     #TODO: find a good way to cache the latest measurment and use it to incease speed (if actually needed)
     measure_ACI = Instrument.measurement(
         MODES["ACI"],
-        """Return the measured value for AC current""",
+        """Return the measured value for AC current
+        """,
         get_process = (lambda x: float(x)),
```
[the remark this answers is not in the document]
LongnoseRob: yes, a quick check without the get_process also worked. --> removed the get process on all measure elements

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,491 @@
...
+    def GPIB_trigger(self):
+        """
+        Initate trigger via low-level GPIB-command (aka GET- group execute trigger)
+
+        """
+        self.adapter.connection.assert_trigger()
+
+    def reset(self):
+        """
+        Initatiates a reset of the HP3478A
```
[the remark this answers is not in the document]
LongnoseRob: it is similiar to a power-on reset, changed the wording of the doc-string

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: with 

```
@staticmethod

def get_status(status_bytes, field=None):
    ...

```
I get the following error:

```

  File "/home/robby/RF_stuff/rf-detectors/soft/xxx/xxx/test.py", line 37, in <module>
    print(str(meter.status))

  File "/home/robby/builds/pymeasure/pymeasure/instruments/instrument.py", line 220, in fget
    value = get_process(vals[0])

  File "/home/robby/builds/pymeasure/pymeasure/instruments/hp/hp3478A.py", line 426, in <lambda>
    get_process = (lambda x: get_status(x.encode(encoding="ASCII"))),

TypeError: 'staticmethod' object is not callable

```
Could this come from the fact that get_status() is defined outside the HP3478A class?

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -465,7 +484,7 @@ def reset(self):
 
     def shutdown(self):
```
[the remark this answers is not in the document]
LongnoseRob: Done, 
```
In [6]: meter.shutdown()

In [7]: meter.isShutdown
Out[7]: True
```

pymeasure/instruments/hp/hp3478A.py
msmttchr: A staticmethod is defined within a class, so I am not sure what you mean with "outside the HP3478A class" ? Where get_status is defined?

Normally you should call the staticmethod as an attribute of the class where it's defined.

Have a look at https://pythonbasics.org/static-method/

pymeasure/instruments/hp/hp3478A.py
CasperSchippers: Check, perfect

pymeasure/instruments/hp/hp3478A.py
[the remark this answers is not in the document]
CasperSchippers: Perfect!!

pymeasure/instruments/hp/hp3478A.py
CasperSchippers: Ahh yes, that would indeed be the problem. I think, for the sake of code clarity, it'd be nice to place these 'get' functions as staticmethods in the class; that way they are nicely confined within the class that uses them and get nicely documented.

In some of the functions you actually use the class (e.g. `HP3478A.MODES`), so these you can best add as `@classmethod` rather than as `@staticmethod` such that you can access the `MODES`:
```
@classmethod
def get_mode(cls, status_bytes):
    ....
    ....
    inv_modes = dict(zip(cls.MODES.values(),cls.MODES.keys()))
    ....
    ....
```

I'd also suggest, to prevent confusion about what these methods do, to change the `get` into something like `decode` (as you use in the docstring) (i.e. `decode_range`).

(btw: I think for inverting a dictionary, a nicer/maybe clearer way is to use `inv_modes = {v: k for k, v in MODES.items()}`

LongnoseRob: Thanks @msmttchr & @CasperSchippers again for your input and sticking with me on my first bigger python project.

@msmttchr 
the function definition for get_status and the other get_functions is in the same file, but before the definition of the class HP3478A:
20210717_spyder

But now I understand that the staticmethod or classmethod only work within the class


I will do the folloing:
1. change the naming of the get_status etc functions to decode_status
2. move these functions into the class hp3478a
3. improve the inverting of the dictionary

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,489 @@
...
+            16: "AD converter error",
+            32: "AD link error",
+           }
+    
+
+
+    #decoder functions
+    @classmethod
+    def decode_status(self,status_bytes, field=None):
+        """Method to handle the decoding of the status bytes into something meaningfull
```
msmttchr, line 200: If you replace `@classmethod` with `@staticmethod` and remove the `self` parameter, you should be good to go with static method.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,489 @@
...
+            return ret_val.b
+        elif field == "SRQ":
+            return getattr(ret_val.B, "byte3")
+        else:
+            return getattr(ret_val.b, field)
+
+    @classmethod
+    def decode_mode(cls,status_bytes):
+        """Method to decode current mode
+
```
msmttchr, line 217: If you replace `@classmethod` with `@staticmethod` and remove the `cls` parameter, you should be good to go with static method.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,489 @@
...
+        """
+        cur_stat = Status(Status_bytes(*status_bytes))
+        function = cur_stat.b.function
+        inv_modes = {v: k for k, v in HP3478A.MODES.items()}
+        cur_mode = inv_modes["F"+str(function)]
+        return cur_mode
+
+    @classmethod
+    def decode_range(cls,status_bytes):
+        """Method to decode current range
```
msmttchr, line 231: If you replace `@classmethod` with `@staticmethod` and remove the `cls` parameter, you should be good to go with static method.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,489 @@
...
+        function = cur_stat.b.function
+        inv_modes = {v: k for k, v in HP3478A.MODES.items()}
+        cur_mode = inv_modes["F"+str(function)]
+        rnge = cur_stat.b.range
+        cur_range = list(HP3478A.RANGES[cur_mode].keys())[rnge-1]
+        return cur_range
+
+    @classmethod
+    def decode_trigger(cls,status_bytes):
+        """Method to decode trigger mode
```
msmttchr, line 248: If you replace `@classmethod` with `@staticmethod` and remove the `cls` parameter, you should be good to go with static method.

pymeasure/instruments/hp/hp3478A.py
CasperSchippers: Hmm, I think I'd actually suggest leaving this (and `decode_range`) a classmethod and changing `HP3478A.MODES` to `cls.MODES` (and also `HP3478A.RANGES` to `cls.RANGES` in `decode_range`); or is there a reason not to do that?

(`decode_status` and `decode_trigger` can indeed be converted to a staticmethod anyway)

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,489 @@
...
+        else:
+            trigger_mode = "internal"
+        return trigger_mode
+
+    #commands/properties for instrument control
+    active_connectors = Instrument.measurement(
+        "B",
+        """Return selected connectors ("front"/"back"), based on front-panel selector switch
+        """,
+        get_process = (lambda x: HP3478A.decode_status(x.encode(encoding="ASCII"), "front_rear")),
```
CasperSchippers, line 272: As you always use `x.encode(....)`, I think it would be nicest if you put the `encode` into the `decode_..` methods as well

pymeasure/instruments/hp/hp3478A.py
msmttchr: @CasperSchippers I think you are right, I didn't notice this. Sorry for the confusion

LongnoseRob: Thanks for the contious support.
I will chage in decode_mode & decode_range the HP3478A. references to cls.
Also I will change decode_status and decode_trigger to static methods.

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Will do.
Imrproves the readabibility of the code

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,490 @@
...
+        :rtype cur_range: float
+
+        """
+        cur_stat = Status(Status_bytes(*status_bytes.encode(encoding="ASCII")))
+        function = cur_stat.b.function
+        inv_modes = {v: k for k, v in cls.MODES.items()}
+        cur_mode = inv_modes["F"+str(function)]
+        rnge = cur_stat.b.range
+        cur_range = list(cls.RANGES[cur_mode].keys())[rnge-1]
+        return cur_range
```
msmttchr, line 245: I think the statement `list(cls.RANGES[cur_mode].keys())[rnge-1]` rely on the fact that dictionary keys are ordered which is not the case. You could try to use `OrderedDict `or some other method, I believe.

LongnoseRob: @msmttchr Thank for the reminder about the fact that the order of elements in dicts can not always be guranteed.
Probably because I am programming here on my ArchLinux machine with Python 3.9.x, I never encountered a problem with the dicts.

Over the next days Ill look at this range read out method again, I think ther emus be a better method to decuct the current range form the status bits, without the reversed dict.

CasperSchippers: I'm not sure what the best approach is. However, we've had this discussion recently for another PR as well; there in the end it was decided it would be safe to rely on the order of a dict. See https://github.com/pymeasure/pymeasure/pull/276#issuecomment-742036301_

bilderbuchi: indeed, assuming order should be safe enough until 3.6 drops off the list of supported versions at EOL.

bilderbuchi:
Wow, this is an intensive reviewing session, thank you all involved!
Please excuse me if I state something that was already discussed -- the only way to check seems to manually uncollapse 99 resolved conversations :-/
Seeing the implementations of `StatusBits`, `StatusBytest` and `__str__`, I'm wondering if using a `struct` plus a NamedTuple would not yield a more elegant and easily graspable interaction, see e.g. the [doc examples](https://docs.python.org/3/library/struct.html#examples) demonstrating how to unpack a struct record into a NamedTuple (which then automaticall yields nice string representation and member access.

msmttchr:
> Wow, this is an intensive reviewing session, thank you all involved!
> Please excuse me if I state something that was already discussed -- the only way to check seems to manually uncollapse 99 resolved conversations :-/
> Seeing the implementations of `StatusBits`, `StatusBytest` and `__str__`, I'm wondering if using a `struct` plus a NamedTuple would not yield a more elegant and easily graspable interaction, see e.g. the [doc examples](https://docs.python.org/3/library/struct.html#examples) demonstrating how to unpack a struct record into a NamedTuple (which then automaticall yields nice string representation and member access.
> Or maybe it comes out to the same thing, in the end.

I think `unpack` does not support bitfields.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,490 @@
...
+    def check_errors(self):
+        """
+        Method to read the error status register
+
+        :return error_status: one byte with the error status register content
+        :rtype error_status: int
+        """
+
+        if self.error_status != 0:
+            log.error("HP3478A error detected: $s", self.ERRORS[self.error_status])
```
bilderbuchi, line 319: If you turned self.ERRORS into an [`Enum.IntFlag`](https://docs.python.org/3/library/enum.html?highlight=intflag#intflag), you'd get the nice property of readable string output and that if multiple errors are present (e.g. if `self.error_status` is 6), this works transparently and gives readable output strings. The present implementation would just fail the dict lookup. 
I've used this in the past to great effect to readably decode a lot status flags out of a status integer.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,490 @@
...
+        return self.error_status
+
+    error_status = Instrument.measurement(
```
bilderbuchi, line 322: Just FYI: This would normally be an Instrument method `check_errors()` that could also be used by other properties to check errors after operation.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,490 @@
...
+    measure_Rext = Instrument.measurement(
+        MODES["Rext"],
+        """Return the measured value for extended resistance mode (>30M, 2-wire) resistance
+        """,
+        )
+
+    mode = Instrument.control(
+        "B",
+        "%s",
+        """Return current selected measurment mode, this propery can be set.
```
bilderbuchi, line 393: measur**e**ment (also below)

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,490 @@
...
+        "B",
+        """Returns the current measurment range, this property can be set.
+        Valid settings are 3*powers of ten (e.g 0.3,3,30)"
+        for all valid ranges look at HP3478A.RANGES structure
+        """,
+        get_process = lambda x: HP3478A.decode_range(x),
+        )
+
+    @range.setter
+    def range(self,value):
```
bilderbuchi, line 412: For reasons of symmetry, while the current arrangement of a measurement + separate setter works, it looks a bit weird - a classic property instead of the measurement would feel more natural to me.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,490 @@
...
+        cur_mode = self.mode
+        set_range = strict_discrete_set(value, self.RANGES[cur_mode])
+        set_range = self.RANGES[cur_mode][value]
```
bilderbuchi, line 415: Are you not overwriting `set_range` in the second line?

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -0,0 +1,490 @@
...
+        """,
+        get_process = (lambda x: HP3478A.decode_status(x)),
+        )
+
+    SRQ_mask = Instrument.control(
+        "B",
+        "M%o",
+        """Return current SRQ mask, this property can be set,
+
+        bit assigment for SQR:
```
bilderbuchi, line 441: "SRQ", I think.

bilderbuchi: And decoding that for the user could also be dealt with with an IntFlag enum.

bilderbuchi:
It's nice to see an implementation of an instrument that uses (at least partially) a frame-based approach for interaction! At one point I'd like to allow for this "properly" (maybe with a custom adapter subclass?), cf. #239 #296. I caution though that I see DLL-based interaction as out of scope for us right now.

bilderbuchi:
> I think unpack does not support bitfields.
Ah you're probably right. I'd probably pack those into IntFlag enums, but of course, then it's not only one struct to unpack anymore, so much of the benefit is gone...

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Thanks for the comments reading the order of the dict elements.
While testing the range method in more detail I found a nasty issue:
If the instrument is in DC or AC current measruement mode the first byte of the status bytes is >128, and therefore I get an error from the instrument.measurement method:
```
meter.measure_ACI
Out[2]: 0.00122

meter.range

  File "<ipython-input-3-09f873ffe539>", line 1, in <module>
    meter.range

  File "/home/robby/builds/pymeasure/pymeasure/instruments/instrument.py", line 235, in fget
    vals = self.values(command_process(get_command), **kwargs)

  File "/home/robby/builds/pymeasure/pymeasure/instruments/instrument.py", line 130, in values
    return self.adapter.values(command, **kwargs)

  File "/home/robby/builds/pymeasure/pymeasure/adapters/adapter.py", line 87, in values
    results = str(self.ask(command)).strip()

  File "/home/robby/builds/pymeasure/pymeasure/adapters/visa.py", line 112, in ask
    return self.connection.query(command)

  File "/home/robby/.local/lib/python3.9/site-packages/pyvisa/resources/messagebased.py", line 644, in query
    return self.read()

  File "/home/robby/.local/lib/python3.9/site-packages/pyvisa/resources/messagebased.py", line 486, in read
    message = self._read_raw().decode(enco)

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc6 in position 0: ordinal not in range(128)
```
I still need to do a more deep  investigation, but I might rework the range method to use a more low-level read method (e.g. pyvisas read_raw).

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Thanks, fixed

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: It looks like this,
Actually we are not using the return valuem so I will change it to keep the call to the validator function.

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Thanks, 
spelling fixed
and IntFlag enum added for SRQ mask decoding

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Thanks for the recommendation,
Changed ERRORS to a IntFlag Enum

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: I think I found the reason for this:
If I switch the instrument to AC-Current measurment (with 4 digit resolution) the first status byte has the value  0xc6(=198).
(In DC-Current measurment the value is 0xa6.)
As this value is 198 in decimal, it is bigger then the 128 allowed by the adatper.read() called adapter.ask().

for now I am working on a staticmethod get_status which communciates to the instrument and reads the 5 bytes of status bytes.

bilderbuchi: Adapters have a binary_values method that might be useful, although it returns a numpy array and not the raw bytes. maybe we should think about exposing a generic read_binary_values in correspondence to the write_binary_values that we already have.

LongnoseRob: @bilderbuchi I like your idea,
I have now reworked my code with a get_status function using the read_bytes() method.

CasperSchippers (approved):
Looks good to me; thanks for the work

LongnoseRob:
Thank you

PR #439, opened 2021-04-29
Placeholder logic

CasperSchippers:
An alternative implementation for replacing placeholders to #423. Credits to @paulgoulain for the idea.

This replacer uses `str.format` to replace placeholders with the name (as given via the `name` argument) of any `Parameter` of the `Procedure`. It also adds to other placeholders: date and time (both with a keyword-argument to choose the format).

An example of usage would be `filename = replace_placeholders("Measurement_{Parameter Name:.2f}", procedure)`.

It can also be called via `unique_filename` by passing the `procedure` keyword-argument.

### DIFF
--- a/pymeasure/experiment/results.py
+++ b/pymeasure/experiment/results.py
@@ -45,6 +45,9 @@ def replace_placeholders(string, procedure):
     parameters = procedure.parameter_objects()
     placeholders = {param.name: param.value for param in parameters.values()}
 
+    placeholders["date"] = now.strftime(date_format)
+    placeholders["time"] = now.strftime(time_format)
+
     # Check keys against available parameters
     invalid_keys = [i[1] for i in Formatter().parse(string)
                     if i[1] is not None and i[1] not in placeholders]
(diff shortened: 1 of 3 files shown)


msmttchr:
Very useful feature :) Thanks.

msmttchr (approved):

pymeasure/experiment/results.py
```diff
@@ -40,12 +41,62 @@
 log.addHandler(logging.NullHandler())
 
 
+def replace_placeholders(string, procedure, date_format="%Y-%m-%d", time_format="%H:%M:%S"):
+    """ Replaces the placeholders in the provided strings with the values of the
+    associated parameters, as provided by the procedure. This uses the standard
```
bilderbuchi, line 46:
```suggestion
    """Replace placeholders in string with values from procedure parameters.
    
    Replaces the placeholders in the provided string with the values of the
    associated parameters, as provided by the procedure. This uses the standard
```
For docstrings it's customary to have the first line as a "short summary" in imperative tone.

bilderbuchi:
Looks solid to me, except for a minor docstring remark.

pymeasure/experiment/results.py
CasperSchippers: Thanks for this remark, I wasn't aware of this.

PR #441, opened 2021-05-07
actually call should_stop in wait for trigger

neal-kepler:
`should_stop` is passed as a function and should actually be called in the loop breaking conditional of `wait_for_trigger`


CasperSchippers (approved):
Thanks for this improvement, I'll merge it right away

Issue #440, opened 2021-05-05
Keithley2000 instrument constructor error


martinjankoehler:
Hello, 

trying to construct `pymeasure.instruments.keithley.Keithley2000` results in an error:
```python
[MacPro]martin:~$ python3
Python 3.8.2 (default, Dec 21 2020, 15:06:04)
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pymeasure
>>> pymeasure.__version__
'0.9.0'
>>> from pymeasure.instruments.keithley import Keithley2000
>>> m = Keithley2000("ASRL4::INSTR")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/martin/Library/Python/3.8/lib/python/site-packages/pymeasure/instruments/keithley/keithley2000.py", line 439, in __init__
    self.adapter.config(
AttributeError: 'VISAAdapter' object has no attribute 'config'
```

LongnoseRob:
Looks like you have an VISA issue with your connection to the instrument.
Which connection are you using?
Which VISA backend are you using?

bklebel:
I don't think there is any problem with the visa backend. I think the function call for `config` here is problematic - I had the same problem once, and just commented the lines away, since I did not make the effort to see whether there exists an instrument which actually uses this, or somehow uses this. 
From a quick glance, I think this function call should be removed, and the functionality which it tries to achieve implemented correctly in the current version of pymeasure, however that might be.

bilderbuchi:
Closed by #443

PR #443, opened 2021-05-08
keithley2000 remove deprecated call to visa.config

bklebel:
Removing the deprecated function call to `self.adapter.config()` in the `keithley2000` multimeter. 
Refer to #440 and [commit where this function was removed from pymeasure](https://github.com/pymeasure/pymeasure/commit/21e659988fcb3a960bbce6143bb5dc806bf53a1b)

I am not 100% sure this would break something in the instrument functions, but it does not seem to, and the functionality of preprocessing output from the instrument is well integrated in pymeasure

### DIFF
--- a/pymeasure/instruments/keithley/keithley2000.py
+++ b/pymeasure/instruments/keithley/keithley2000.py
@@ -430,11 +430,6 @@ class Keithley2000(Instrument, KeithleyBuffer):
         values=[0, 999999.999]
     )
 
-    def __init__(self, adapter, **kwargs):
-        super(Keithley2000, self).__init__(
-            adapter, "Keithley 2000 Multimeter", **kwargs
-        )
-
     # TODO: Clean up error checking
     def check_errors(self):
         """ Read all errors from the instrument."""
--- a/pymeasure/instruments/keithley/keithley2000.py
+++ b/pymeasure/instruments/keithley/keithley2000.py
@@ -434,14 +434,6 @@ class Keithley2000(Instrument, KeithleyBuffer):
         super(Keithley2000, self).__init__(
             adapter, "Keithley 2000 Multimeter", **kwargs
         )
-        # Set up data transfer format
-        if isinstance(self.adapter, VISAAdapter):
-            self.adapter.config(
-                is_binary=False,
-                datatype='float32',
-                converter='f',
-                separator=','
-            )
 
     # TODO: Clean up error checking
     def check_errors(self):
--- a/pymeasure/instruments/keithley/keithley2000.py
+++ b/pymeasure/instruments/keithley/keithley2000.py
@@ -430,6 +430,11 @@ class Keithley2000(Instrument, KeithleyBuffer):
         values=[0, 999999.999]
     )
 
+    def __init__(self, adapter, **kwargs):
+        super(Keithley2000, self).__init__(
+            adapter, "Keithley 2000 Multimeter", **kwargs
+        )
+
     # TODO: Clean up error checking
     def check_errors(self):
         """ Read all errors from the instrument."""


pymeasure/instruments/keithley/keithley2000.py
```diff
@@ -430,11 +430,6 @@ class Keithley2000(Instrument, KeithleyBuffer):
         values=[0, 999999.999]
     )
 
-    def __init__(self, adapter, **kwargs):
-        super(Keithley2000, self).__init__(
-            adapter, "Keithley 2000 Multimeter", **kwargs
```
martinjankoehler, line 435: Isn't the init still necessary to set the instrument name?

bklebel: Thank you for pointing this out, I was a bit too quick there

bilderbuchi:
This looks fine I think, but I don't have a device to test. 
@martinjankoehler can you test as many features of the instrument as possible, to make sure this does not have a negative impact?

martinjankoehler:
```python
from pymeasure.instruments.keithley import Keithley2000
sourcemeter = Keithley2000("ASRL7::INSTR")
print(sourcemeter.id)
```
now delivers successfully
```
KEITHLEY INSTRUMENTS INC.,MODEL 2015,1197542,B17  /A02
```
Have yet to do more thorough tests.

caygen:
Hello,
I am trying to use Keithley 2000 over GPIB, so I need the VISA stuff and commenting out those lines didn't fix my problem. Is there anything else I can try?

caygen:
Hello I think I followed the fix instructions here correctly, yet I still cannot get it to work.
(I am using GPIB communication)

```
        super(Keithley2000, self).__init__(
            adapter, "Keithley 2000 Multimeter", **kwargs
        )
        # # Set up data transfer format
        # if isinstance(self.adapter, VISAAdapter):
        #     self.adapter.config(
        #         is_binary=False,
        #         datatype='float32',
        #         converter='f',
        #         separator=','
        #     )
```
and in the command line I have this to test it
```
from pymeasure.instruments.keithley import Keithley2000
K3 = Keithley2000("GPIB::3")
```

which results in this:
```
File "path of the file \keithley2000.py", line 434, in __init__
    
super(Keithley2000, self).__init__(
TypeError: super(type, obj): obj must be an instance or subtype of type
```

CasperSchippers:
> Hello I think I followed the fix instructions here correctly, yet I still cannot get it to work.

That is rather weird as what you used should indeed work; you could try changing `super(Keithhley2000, self)` into `super()` or using this PR rather than editing it yourself, though I have to admit that if the lines you mentioned are the only one that are changed it should just work.


> Removing the deprecated function call to self.adapter.config() in the keithley2000 multimeter.

Does this work for both GPIB and RS232 communication, or do we expect to create issues by not specifying the now-removed settings for RS232 communication?

bklebel:
I do expect this to be independent from the difference between GPIB and RS232. Indeed, from my experience most often instruments that have only RS232 as connection option do not follow the SCPI protocol, and the configuration for this instrument here looks like an early version of "I get an SCPI response, please reformat it into a list of floats", we do not specify the baud rate, number of data/stop bits, or the parity which would be issues with RS232. 
However, even it that were the case, I like the approach taken in general in pymeasure, to separate the communication layer from the device logic layer. For instruments which can only communicate over a certain interface, which additionally need special communication specifications being set, I would still advocate to only include this information in the respective documentation (clearly stated) - I don't think pymeasure should tell anyone how to connect to a certain instrument. 

The only issues I might expect are coming from functions which might directly rely on the output generated by this removed configuration. In this case, the only function which does something like this is `check_errors` in line 447, however this already uses the adapter function `self.values` in the default configuration, which separates with a `,` similar to other error checking functionsin so many other instruments defined in pymeasure. 
It should still be tested, I think - I sadly do not have such a device at hand.

bilderbuchi:
I agree that the removed config call should not have an impact on the serial communication settings.

I don't see how the error in the super call can occur, very weird (especially since it works for others) - does this also occur if you connect via serial instead of gpib? (not that I see how that could make a difference)

msmttchr (commented):
@bklebel could you please remove method `check_errors` since, for SCPI instruments, it is provided by `Instrument` class ?
I would replace also `super(Keithley2000, self)` with `super()`
For the rest, it looks good to me.

bilderbuchi:
So, I have no idea how @caygen triggered this error, since the pattern of the super call is the same as for all the other Keithley instruments using the KeithleyBuffer. As we have not heard back from them, I suggest we move on. Also, I think it's fine to keep the super call as it is as again, it follows the same pattern as its sibling instruments.

bklebel:
I am sorry it took me a while to get back to it - thanks for merging!

Issue #334, opened 2020-10-16
Allow passing serial parameters to PyVISA through PyVisaAdapter?


alterscape:
I'm trying to control an instrument through PyMeasure which uses a rather unorthodox serial setup: 57600 baud, 7 data bits, 1 stop bit, odd parity. All these settings are kwargs for PyVisa's `get_instrument` / `open_resource` but are not permitted by safeKeywords in visa.py.  I ended up hacking my local copy of the library to validate that adding them to safeKeywords works.

Is there any compelling reason not to allow these keyword args, since they make it possible to configure more instruments?

bilderbuchi:
Hm, you raise an interesting point. I'd have to dig for why the safe keywords are around. Right now I can't see why we should not extend the list, but [Chesterton's fence](https://en.wikipedia.org/wiki/Wikipedia%3AChesterton%27s_fence) and all that...

In the meantime, to define fixed serial params for your instrument without modifying pymeasure, the following pattern can be useful in your specific Instrument's init-- you create a `SerialAdapter` and pass _that_ to the `Instrument` init:
```
    def __init__(self, adapter, name='My unorthodox instrument',
                 baudrate=57600, bytesize=7, parity='O', stopbits=1, **kwargs):
        """Docstring
        """
        adapter = SerialAdapter(adapter, **kwargs)
        super().__init__(adapter, name, **kwargs)
```

dkriegner:
I am preparing a [pull request](https://github.com/dkriegner/pymeasure/tree/ibeam) (which waits for #327 to be finished) which is also suffering from this issue. 
My feeling was that the safe keywords are indeed only keywords which **every** pyvisa resource supports. this of course comes with some limitations.

My workaround was to set the communication parameters after opening the resource. see [here](https://github.com/dkriegner/pymeasure/blob/13bd9b2cf217294ea36c8d287c2ee083b8e626b7/pymeasure/instruments/toptica/ibeamsmart.py#L123)

bilderbuchi:
Also, [`baud_rate`](https://github.com/ralph-group/pymeasure/pull/352/files#diff-fc67f8adf8c97a2f75456448361816bffbb5c18d82cf31f709c6620638426ccbR47) is not getting through.

bilderbuchi:
> My feeling was that the safe keywords are indeed only keywords which every pyvisa resource supports

[...]

PR #444, opened 2021-05-10
Added Fluke 7341 temperature bath instrument

msmttchr:
Simple non SCPI serial interface

### DIFF
--- a/pymeasure/instruments/fluke/fluke7341.py
+++ b/pymeasure/instruments/fluke/fluke7341.py
@@ -53,7 +53,8 @@ class Fluke7341(Instrument):
 
     id = Instrument.measurement("*ver",
                                 """ Read the instrument model """,
-                                preprocess_reply=lambda x: "Fluke,{},NA,{}".format(x[0][4:], x[1])
+                                preprocess_reply=lambda x: x,
+                                get_process=lambda x: "Fluke,{},NA,{}".format(x[0][4:], x[1])
                                 )
 
     def __init__(self, resource_name, **kwargs):
--- a/pymeasure/instruments/fluke/fluke7341.py
+++ b/pymeasure/instruments/fluke/fluke7341.py
@@ -1,0 +1,68 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2021 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+from pymeasure.adapters import SerialAdapter
+
+class FlukeSerialAdapter(SerialAdapter):
+    def write(self, command):
+        super().write(command + "\r\n")
+
+class Fluke7341(Instrument):
+    """ Represents the compact constant temperature bath from Fluke
+    """
+
+    set_point = Instrument.control("s", "s=%g",
+                                   """ Set the bath temperature. This property can be read
+                                   """,
+                                   validator=strict_range,
+                                   values = (-40, 150),
+                                   get_process=lambda x: float(x.split()[1])
+                                   )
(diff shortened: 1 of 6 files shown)


pymeasure/instruments/fluke/fluke7341.py
```diff
@@ -0,0 +1,68 @@
...
+class Fluke7341(Instrument):
+    """ Represents the compact constant temperature bath from Fluke
+    """
+
+    set_point = Instrument.control("s", "s=%g",
+                                   """ Set the bath temperature. This property can be read
+                                   """,
+                                   validator=strict_range,
+                                   values = (-40, 150),
+                                   get_process=lambda x: float(x.split()[1])
```
CasperSchippers, line 42: As almost all of the properties have the same `get_process`, maybe it would be nice (in the DRY-mindset) to define a (static)method that performs the split; if you pass it as the `preprocess_reply` rather than the `get_process` keyword argument, you can even leave out the casting in the method, as the casting to float is attempted by default after the `preprocess_reply`

pymeasure/instruments/fluke/fluke7341.py
```diff
@@ -0,0 +1,68 @@
...
+class FlukeSerialAdapter(SerialAdapter):
+    def write(self, command):
+        super().write(command + "\r\n")
+
+class Fluke7341(Instrument):
+    """ Represents the compact constant temperature bath from Fluke
+    """
+
+    set_point = Instrument.control("s", "s=%g",
+                                   """ Set the bath temperature. This property can be read
```
CasperSchippers, line 38: I'd suggest to mention the valid values?

Is this value in the unit provided by the `unit` property?

pymeasure/instruments/fluke/fluke7341.py
```diff
@@ -0,0 +1,68 @@
...
+    unit = Instrument.control("u", "u=%s",
+                              """ Set the temperature unit. This property can be read
+                              """,
+                              validator=strict_discrete_set,
+                              values = ('c','f'),
+                              get_process=lambda x: x.split()[1]
+                              )
+    
+    temperature = Instrument.measurement("t",
+                                         """ Read the current bath temperature """,
```
CasperSchippers, line 53: I would mention that the unit is the same as the one provided by `unit` to prevent confusion

pymeasure/instruments/fluke/fluke7341.py
```diff
@@ -0,0 +1,68 @@
...
+
+    set_point = Instrument.control("s", "s=%g",
+                                   """ Set the bath temperature. This property can be read
+                                   """,
+                                   validator=strict_range,
+                                   values = (-40, 150),
+                                   get_process=lambda x: float(x.split()[1])
+                                   )
+    unit = Instrument.control("u", "u=%s",
+                              """ Set the temperature unit. This property can be read
```
CasperSchippers, line 45: Maybe mention the valid values

CasperSchippers (changes requested):
Thanks for the contribution; I've had a look and added some comments to the code

pymeasure/instruments/fluke/fluke7341.py
msmttchr: You mean using: `preprocess_reply=lambda x: x.split()[1]`. I wouldn't go for a static method since the `lambda` is almost as compact a method name. What do you think ?

pymeasure/instruments/fluke/fluke7341.py
msmttchr: OK, I will update documentation

pymeasure/instruments/fluke/fluke7341.py
msmttchr: OK, I will update documentation

pymeasure/instruments/fluke/fluke7341.py
msmttchr: OK, I will update documentation

pymeasure/instruments/fluke/fluke7341.py
```diff
@@ -0,0 +1,68 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+from pymeasure.adapters import SerialAdapter
+
+class FlukeSerialAdapter(SerialAdapter):
```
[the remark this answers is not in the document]
msmttchr: I chose `SerialAdapter` since it is much more simple than `VisaAdapter` and I felt could be more appropriate for that simple non SCPI instrument. I am not sure this is acceptable. What is your opinion ? On the other hand, I agree with your comment on removing the class `FlukeSerialAdapter` and I replaced it with a re-implementation of `write`.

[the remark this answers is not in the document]
CasperSchippers: I understand your argument, however, the advantage of the visa adaptor is that the communication is handled by visa; one of the advantages of this is that the having the instrument open in one script doesn't block opening another connection via visa.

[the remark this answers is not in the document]
msmttchr: I have tried to implement it using `VISAAdapter `as follow:
```
    def __init__(self, port):
        adapter = VISAAdapter(port, write_termination='\r\n', timeout=2000, preprocess_reply=lambda x: x.split()[1])
        adapter.connection.baudrate = 2400
        super(Fluke7341, self).__init__(
            adapter,
            "Fluke 7341",
            includeSCPI=False
        )
```
Is this OK, in your opinion ?
Few notes:
- I need to set the `baudrate` explicitly, since `safeKeywords` "feature" (see #334)
- The port parameter is restricted to serial port, but I am not sure how to implement this restriction. Is this OK ?

Thanks again for your time.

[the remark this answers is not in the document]
msmttchr: Ok. Done. Thanks for the suggestion

pymeasure/instruments/fluke/fluke7341.py
CasperSchippers: Yeah like this is also fine with me indeed

CasperSchippers (approved):
Thanks for the adjustments; looks good to me.

bilderbuchi:
Thank you for the contribution, looks good to me too! (Please excuse the delays, I'm super busy nowadays)

Issue #387, opened 2021-01-04
How to store commanded state when using Instrument.setting()?


pyMatJ:
Hi,
This is a general question regarding the best approach to create a driver for an instrument (here a current amplifier SR570 from Stanford Research Systems) that has no reading capabilities for its own state (AFAIK it does not even return a statement to acknowledge successfully setting a parameter). 
As stated in the manual, "the RS232 interface is configured as listen only", and all the functions only allow to set a given property. In my current implementation, I only use Instrument.setting() methods.  
This creates two problems: the user cannot check the state of a given property (in the eventuality that it was modified from the front interface), and there is no way to keep track of the current state of the instrument.

The first problem cannot be solved, obviously, unless the user manually checks and reset the properties through python if needed. I see however multiple ways to handle the second one, amongst which:
- creating instruments attributes in the code, and setting them accordingly in parallel to the Instrument.setting() 
- creating a dictionary of the instrument's properties
 
What would be the preferred way of dealing with this ? 
Trying to keep track of the changes with minimal intervention from the user implies maybe that the value should be stored during the Intrument.setting() call, maybe using set_processed ? Or using decorators (I'm new to these) to avoid the getter/setter scheme ?

I'm looking forward to your input !

bilderbuchi:
Hm, are not your two problems 2 sides of the same coin, really? What are you trying to accomplish with tracking the "state" of the instrument, when that seems impossible as the instrument cannot be queried, and you cannot confirm if a value was changed from the front panel? Are you purely trying to track the "state of remote user commands" so to say? Is that useful to you in operating the device?

I know that Lantz implements the concept of [caching](https://lantz.readthedocs.io/en/0.3/tutorial/using.html#the-cache), which could be similar to what you try to accomplish. If that sounds useful to you, depending on the complexity of the command set, one approach could be implement a custom [adapter](https://pymeasure.readthedocs.io/en/latest/api/adapters.html) based on the VISAAdapter, which records all the commands it receives and only passes on those whose values don't agree with the latest value. Pay attention that flip-flopping values get processed correctly (e.g. `GAIN 1`, `GAIN 5`, `GAIN1`). This does not afford you user-visible checks of set values, though, purely saves communication overhead on repeated commands.

Using `set_process` will be tricky because you will need a stateful check (keeping the last set value), there is currently no nice/elegant way to do this.

pyMatJ:
[...]

PR #445, opened 2021-05-11
Adding SR570 current amplifier driver

pyMatJ:
This adds the driver for the SR570 current amplifier (from Stanford Research Systems).
The SR570 communicates through a RS232 listen-only so it is up to the end user to keep track of the values of the set parameters, no function in the driver stores or returns a value (see https://github.com/pymeasure/pymeasure/issues/387 moved to https://github.com/pymeasure/pymeasure/discussions/406)

### DIFF
--- a/docs/api/instruments/srs/index.rst
+++ b/docs/api/instruments/srs/index.rst
@@ -9,5 +9,6 @@ This section contains specific documentation on the Stanford Research Systems (S
 .. toctree::
    :maxdepth: 2
 
+   sr570
    sr830
    sr860
--- a/pymeasure/instruments/srs/sr570.py
+++ b/pymeasure/instruments/srs/sr570.py
@@ -206,9 +206,9 @@ class SR570(Instrument):
     
     def blank_front(self):
         """"Blanks the frontend output of the device"""
-        self.write("BLNK 1")
+        self.front_blanked = True
         
     def unblank_front(self):
         """Un-blanks the frontend output of the device"""
-        self.write("BLNK 0")
+        self.front_blanked = False
         
--- a/pymeasure/instruments/srs/sr570.py
+++ b/pymeasure/instruments/srs/sr570.py
@@ -29,6 +29,14 @@ from pymeasure.instruments.validators import strict_discrete_set, \
 
 class SR570(Instrument):
     
+    def __init__(self, resourceName, **kwargs):
+        super(SR570, self).__init__(
+            resourceName,
+            "Stanford Research Systems SR570 Lock-in amplifier",
+            **kwargs
+        )
+    
+    
     SENSITIVITIES = [
         1e-12, 2e-12, 5e-12, 10e-12, 20e-12, 50e-12, 100e-12, 200e-12, 500e-12,
         1e-9, 2e-9, 5e-9, 10e-9, 20e-9, 50e-9, 100e-9, 200e-9, 500e-9,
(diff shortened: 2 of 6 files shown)


CasperSchippers (changes requested):
Thanks for the addition.

Could you also create the appropriate doc-files (docs/api/instruments/srs/sr570.rst, you can take an example from the other instruments) and add it to the index files (docs/api/instruments/srs/index.rst). Finally, don't forget to add yourself to the two authors lists (AUTHORS.txt and docs/about/authors.txt)

pyMatJ:
I tried to make all these modifications. 
- the file terminates on an empty line
- the init is at the beginning of the class 
- I added the allowed values in the docstring
- I added setting properties in addition to wrapper functions and used mapping values when recommended (it is indeed much more easy to read like this)

Thanks for your inputs !

pymeasure/instruments/srs/sr570.py
```diff
@@ -0,0 +1,214 @@
...
+    # Methods        #
+    ####################
+    
+    def enable_bias(self):
+        """Turns the bias voltage on"""
+        self.write("BSON 1")
+    
+    def disable_bias(self):
+        """Turns the bias voltage off"""
+        self.write("BSON 0")
```
CasperSchippers, line 189: As you have implemented the `bias_enabled` property, maybe you could change `self.write("BSON 1")` to `self.bias_enabled = True` (same holds for the other wrappers below)

CasperSchippers (changes requested):

docs/api/instruments/srs/index.rst
```diff
@@ -9,5 +9,6 @@ This section contains specific documentation on the Stanford Research Systems (S
 .. toctree::
    :maxdepth: 2
 
+   sr570
```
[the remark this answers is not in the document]
pyMatJ: ah sorry, it got lost in a commit -a that does not add files not previously indexed... I'm a bit new to this, my bad.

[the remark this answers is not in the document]
CasperSchippers: Ah, yes; no worries

pymeasure/instruments/srs/sr570.py
```diff
@@ -0,0 +1,214 @@
...
+    
+    def clear_overload(self):
+        """"Reset the filter capacitors to clear an overload condition"""
+        self.write("ROLD")
+    
+    def reset_instrument(self):
+        """Resets the amplifier to default settings"""
+        self.write('*RST')
+    
+    def blank_front(self):
```
CasperSchippers, line 207: Could you do the same for this as for the other wrappers with the property?

pymeasure/instruments/srs/sr570.py
```diff
@@ -0,0 +1,214 @@
...
+        """"Disables the offset current """
+        self.offset_current_enabled = False
+    
+    def clear_overload(self):
+        """"Reset the filter capacitors to clear an overload condition"""
+        self.write("ROLD")
+    
+    def reset_instrument(self):
+        """Resets the amplifier to default settings"""
+        self.write('*RST')
```
CasperSchippers, line 205: A method for sending `*RST` is standard implemented (named `reset`), so you can remove this to prevent code duplication.

CasperSchippers (commented):
Sorry, i missed this one thing last time i checked; after this it can be merged I think.

pymeasure/instruments/srs/sr570.py
pyMatJ: Ah yes I see, I did not check this. This is fixed now, thanks for your thorough reviewing work :)

PR #446, opened 2021-05-17
Added Superconducting Magnet Controller from ScientificMagnetics / Twickenham

DF9AS:
I've added the Superconducting Magnet Controller from Twickenham Scientific (which was also distributed by ScientificMagnetics) to the Instruments-Library. Therefore, I had to slightly change the adapter-module to accept Dict-Values, as the SMU is a non-SCPI Device and, in my opinion, the communication protocol is kind of a "Bastard Protocol From Hell"; which can be checked here: https://www.twicksci.co.uk/manuals/pdf/smc552+.pdf

I hope that this work will be useful for other scientists - but be  careful: mixing up the solenoid can lead to a quench and thus, depending on  your setup, to serious damage or injury! I've only tested the setter- and getter functions for the magnetic field and also the shutdown routine as I don't need other functions by the moment.

### DIFF
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -89,18 +89,21 @@ class Adapter(object):
             results = preprocess_reply(results)
         elif callable(self.preprocess_reply):
             results = self.preprocess_reply(results)
-        results = results.split(separator)
-        for i, result in enumerate(results):
-            try:
-                if cast == bool:
-                    # Need to cast to float first since results are usually
-                    # strings and bool of a non-empty string is always True
-                    results[i] = bool(float(result))
-                else:
-                    results[i] = cast(result)
-            except Exception:
-                pass  # Keep as string
-        return results
+        if isinstance(results, dict):
+            return results
+        else:
+            results = results.split(separator)
+            for i, result in enumerate(results):
+                try:
+                    if cast == bool:
+                        # Need to cast to float first since results are usually
+                        # strings and bool of a non-empty string is always True
+                        results[i] = bool(float(result))
+                    else:
+                        results[i] = cast(result)
+                except Exception:
+                    pass  # Keep as string
+            return results
 
     def binary_values(self, command, header_bytes=0, dtype=np.float32):
         """ Returns a numpy array from a query for binary data 
--- a/pymeasure/instruments/scientificmagnetics/scimag.py
+++ b/pymeasure/instruments/scientificmagnetics/scimag.py
@@ -31,15 +31,16 @@ log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 # Define RegExpressions to match the responses of the instrument
-_reG = re.compile(r'^I([\s+-]\d{3}.\d{3})V([\s+-]\d{2}.\d)R(\d[AV])\r\n$')
-_reJF = re.compile(r'^F([\s+-]\d{2}.\d{4})H(\d)\r\n$')
-_reJI = re.compile(r'^I([\s+-]\d{3}.\d{3})H(\d)\r\n$')
-_reKF = re.compile(r'^R(\d)M(\d)P(\d)X(\d)H(\d)Z0.00E(\d{2})Q([\s+-]\d{2}.\d{4})\r\n$')
-_reKI = re.compile(r'^R(\d)M(\d)P(\d)X(\d)H(\d)Z0.00E(\d{2})Q([\s+-]\d{3}.\d{3})\r\n$')
-_reN = re.compile(r'^F([\s+-]\d{2}.\d{4})V([\s+-]\d{2}.\d)R(\d[AV])\r\n$')
-_reO = re.compile(r'^A(\d{2}.\d{5})D(\d)T(\d)B(\d)W(\d{3}.)C(0.\d{6})\r\n$')
-_reSF = re.compile(r'^T(\d)U(\d{2}.\d{4})L(\d{2}.\d{4})Y(\d{2}.\d)\r\n$')
-_reSI = re.compile(r'^T(\d)U(\d{3}.\d{3})L(\d{3}.\d{3})Y(\d{2}.\d)\r\n$')
+_reG = re.compile(r'^I([\s+-]\d{3}.\d{3})V([\s+-]\d{2}.\d)R(\d[AV])$')
+_reJF = re.compile(r'^F([\s+-]\d{2}.\d{4})H(\d)$')
+_reJI = re.compile(r'^I([\s+-]\d{3}.\d{3})H(\d)$')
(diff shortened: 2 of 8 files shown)


pymeasure/instruments/scientificmagnetics/scimag.py
```diff
@@ -0,0 +1,448 @@
...
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+import logging
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+# Define RegExpressions to match the responses of the instrument
+_reG = re.compile(r'^I([\s+-]\d{3}.\d{3})V([\s+-]\d{2}.\d)R(\d[AV])$')
+_reJF = re.compile(r'^F([\s+-]\d{2}.\d{4})H(\d)$')
+_reJI = re.compile(r'^I([\s+-]\d{3}.\d{3})H(\d)$')
```
CasperSchippers, line 36: As I understand, you have a number of time a `F` and `I` version of the same Regular Expression where just the numbers differ (presumable, but maybe I misunderstand, because the full length of the double is constant but the position of the point may be different). Wouldn't it possible to combine them into a single expression?
I think that using `([\s+-]\d{2,3}.\d{3,4})` would work for example; another option is `([\s+-][\d.]{7})` to make it fully independent of the position of the point.

pymeasure/instruments/scientificmagnetics/scimag.py
```diff
@@ -0,0 +1,448 @@
...
+import time
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+import logging
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+# Define RegExpressions to match the responses of the instrument
+_reG = re.compile(r'^I([\s+-]\d{3}.\d{3})V([\s+-]\d{2}.\d)R(\d[AV])$')
```
CasperSchippers, line 34: To keep understanding the code in the future, it might be good to have a bit more descriptive variablenames for these regular expressions

pymeasure/instruments/scientificmagnetics/scimag.py
```diff
@@ -0,0 +1,448 @@
...
+        Unit can be "T" or "A" and must be set before.
+        """,
+        validator=strict_range,
+        values=[0, 200]
+    )
+
+    output_parameters_amps = Instrument.measurement(
+        "G",
+        """Returns dict of output parameters.""",
+        preprocess_reply=extract_value_output_parameters_amps
```
CasperSchippers, line 349: See my remark in `instruments.py`; I think you can better pass it as a `get_process`.

pymeasure/instruments/scientificmagnetics/scimag.py
```diff
@@ -0,0 +1,448 @@
...
+        WARNING: Unconditional on will turn on the heater without any interlocks! Use at own risk.
+        Standard mode to switch heater on is Conditional.
+        Simplified Commands are "True" (Conditional on) and "False" (Off).
+        """,
+        validator=strict_discrete_set,
+        values={"O": 0, "C": 1, "U": 2, "R": 9, False: 0, True: 1},
+        map_values=True
+    )
+
+    ramp_rate = Instrument.setting(
```
CasperSchippers, line 286: Can you mention the valid values or ranges in the docstrings?

pymeasure/instruments/scientificmagnetics/scimag.py
```diff
@@ -0,0 +1,448 @@
...
+        """Sets the External trip:
+         False = Off, True = Simple trip On, "Auto" = Auto-rampdown On.""",
+        validator=strict_discrete_set,
+        values={False: 0, True: 1, "Auto": 4},
+        map_values=True
+    )
+
+    terminal_voltage_limit = Instrument.setting(
+        "Y%04.1f",
+        """Sets the terminal voltage limit to nn.n V.""",
```
CasperSchippers, line 304: what do you mean by `nn.n`? If you mean that you can set it with 0.1V accuracy, it might be clearer to directly mention that.

pymeasure/instruments/scientificmagnetics/scimag.py
```diff
@@ -0,0 +1,448 @@
...
+                 **kwargs):
+        super().__init__(
+            resourceName,
+            "Scientific Magnetics SMC",
+            includeSCPI=False,
+            **kwargs
+        )
+        self._RAMP_RATE_MAX = ramp_rate_max  # A/s
+        self._VOLTAGE_LIMIT_MAX = voltage_limit_max  # V
+        self._T_MAX = field_max  # T
```
CasperSchippers, line 395: Please not that updating these variables doesn't have any impact on the validators where you use these variables (see  #250, and pull requestes #371 & #376)

pymeasure/instruments/scientificmagnetics/scimag.py
```diff
@@ -0,0 +1,448 @@
...
+
+    @property
+    def tesla(self) -> float:
+        """
+        High-Level property for setting and getting the field in the solenoid.
+
+        If installed, the setter-funktion toggles the
```
CasperSchippers, line 402: typo in 'funktion'

pymeasure/instruments/scientificmagnetics/scimag.py
```diff
@@ -0,0 +1,448 @@
...
+            self.ramp_target = "U"
+        else:
+            log.warning("Intended field was: %08.4f T.\nAllowed maximum field is: %08.4f T" % (value, self._T_MAX))
+            raise UserWarning("Intended field value exceeded range of the solenoid. Value was NOT set.")
+
+    def shutdown(self):
+        """Brings the instrument to a safe and stable state.
+        """
+        self.isShutdown = True
+        log.info("Shutting down %s" % self.name)
```
CasperSchippers, line 445: I think it is better to use `super().shutdown()` rather than copying the lines of code.

CasperSchippers (changes requested):
Thanks for the contribution and sorry for the late response. I've had a look at the code and left some comments.

pymeasure/adapters/adapter.py
```diff
@@ -89,18 +89,21 @@ def values(self, command, separator=',', cast=float, preprocess_reply=None):
...
+                try:
+                    if cast == bool:
+                        # Need to cast to float first since results are usually
+                        # strings and bool of a non-empty string is always True
+                        results[i] = bool(float(result))
+                    else:
+                        results[i] = cast(result)
+                except Exception:
+                    pass  # Keep as string
+            return results
```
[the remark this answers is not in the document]
bilderbuchi: I agree, preprocess_reply should be a string-to-string operation, maybe it makes sense to clarify that in the docs?

bmoneke:
@DF9AS do you want to pick up this PR or may I close it? It can be reopened whenever you want to continue.

bmoneke:

PR #449, opened 2021-05-21
Implement visibility groups in `InputsWidgets`

CasperSchippers:
Often in my measurement software I have a bunch of inputs which are only relevant if a certain checkbox is ticked; e.g. a boolean to control whether the temperature should be regulated or not, and if only if this is checked inputs like the temperature set-point, ramp-rate and hold-time are relevant.
To keep the interface clean, I here propose to "group" a couple of inputs under a single `BooleanInput`: if the checkbox is checked the inputs will be visible, if it is unchecked the 'grouped' inputs will be invisible. An additional key-word argument is added to the `Parameters` (`grouped_by`) that can refer to the name of the `BooleanParameter` that will control the visibility of the `Parameter`.


msmttchr:
This is a very useful feature. Some observations from my side:
- In some GUI design, inputs are greyed out rather than hidden. Is it possible to allow both modes under user selection ?
- I met the same issue also with other parameter type, e.g. list, floating number, etc. You can imagine to have a group of parameters that make sense only if a list parameter or a float parameter has a certain value. For example, you could have a list parameter with two choices 'voltage driven' or 'temperature driven' and according to user selection an appropriate list of relevant parameters is shown to user. Is possible to use any type of parameter and not just Boolean to control groups ?

CasperSchippers:
Thanks for the suggestions. I think both are good additions:
- I went for the invisible option as then you save space when a parameter is not valid, but graying out is definitely also possible. I could for example introduce a key-word argument to the `ManagedWindow` which controls this, defaulting to invisibility.
- Grouping for other parameters should also be possible; I so far did not do this as it would me a bit more complex than with a Boolean, and it would require more keyword-arguments in the parameters. I'll have a look at how to do this. One question though: for the `ListParameter` I understand it's usefulness, but could you give an example for a numeric parameter, e.g. the `FloatParameter`?

pymeasure/display/widgets.py
```diff
@@ -490,16 +492,72 @@ def _layout(self):
...
+            elif isinstance(group_el, StringInput):
+                group_el.textChanged.connect(toggle)
+                toggle(group_el.text())
+            elif isinstance(group_el, IntegerInput):
+                group_el.valueChanged.connect(toggle)
+                toggle(group_el.value())
+            elif isinstance(group_el, ScientificInput):
+                group_el.valueChanged.connect(toggle)
+                toggle(group_el.value())
+            elif isinstance(group_el, ListInput):
```
msmttchr, line 557: I noticed that the code for IntegerInput and ScientificInput cases is the same, what about using:
`elif isinstance(group_el, (IntegerInput, ScientificInput)):`
to avoid duplication

pymeasure/display/widgets.py
```diff
@@ -490,16 +492,72 @@ def _layout(self):
...
+                group_el.currentTextChanged.connect(toggle)
+                toggle(group_el.currentText())
+
```
msmttchr, line 560: There is no `else` condition on the `if`/`elif` sequence. Is this intentional ? I woud assume that if a new `Parameter` subclass is defined, this code fragment needs to be updated. What is the preference in such cases  ? Ignore the unknown type or to raise an exception ?
No strong opinion from my side, I just raised this as potential point of discussion.

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,38 @@ This adds the input line above the Queue and Abort buttons.
...
+This is done with the :code:`group_by` key-word argument.
+
+.. code-block:: python
+
+    toggle = BooleanParameter("toggle", default=True)
+    param = FloatParameter('some parameter', group_by='toggle')
+
+When both the :code:`toggle` and :code:`param` are visible in the :code:`InputsWidget` (via :code:`inputs=['iterations', 'delay', 'seed']` as demonstrated above) one can control whether the input-field of :code:`param` is visible by checking and unchecking the checkbox of :code:`toggle`.
+By default, the group will be visible if the value of the :code:`group_by` :code:`Parameter` is :code:`True` (which is only relevant for a :code:`BooleanParameter`, but is is possible to specify other value as conditions using the :code:`group_condition` keyword argument.
+
```
msmttchr, line 355: Typo "is is possible"

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,38 @@ This adds the input line above the Queue and Abort buttons.
...
+Here the input of :code:`param` is only visible if :code:`iterations` has a value of 99.
+This works with any type of :code:`Parameter` as :code:`group_by` parameter.
+
+To allow for even more flexibility, it is also possible to pass a (lambda)function as a condition:
+.. code-block:: python
+
+    iterations = IntegerParameter('Loop Iterations', default=100)
+    param = FloatParameter('some parameter', group_by='iterations', group_condition=lambda v: 50 > v > 99)
+
```
msmttchr, line 369: Should be `50 < v < 99` instead of `50 > v > 99`. Right ?

pymeasure/display/widgets.py
```diff
@@ -490,16 +492,72 @@ def _layout(self):
...
+            condition = parameter.group_condition
+            if group is None or group not in self._inputs or group == name:
+                continue
+
+            if isinstance(getattr(self, group), BooleanInput):
+                if condition:
+                    condition = 2
+                elif not condition:
+                    condition = 0
+
```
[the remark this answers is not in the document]
CasperSchippers: yeah fair enough

pymeasure/display/widgets.py
```diff
@@ -490,16 +492,72 @@ def _layout(self):
...
+                groups[group].append((self.labels[name], condition))
+
+        for name, group in groups.items():
+            def toggle(state):
+                for (el, con) in group:
+                    if callable(con):
+                        visible = con(state)
+                    else:
+                        visible = (state == con)
+
```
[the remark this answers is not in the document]
CasperSchippers: Compact, yes; clear, I don't know. I think I'd prefer the way it is

pymeasure/display/widgets.py
CasperSchippers: Yes, good one

pymeasure/display/widgets.py
CasperSchippers: Good point; I'd go for raising an exception: If the new type of parameter is not used, it will not brake the code, but if it is used the user is made aware of the limitation.

pymeasure/display/widgets.py
```diff
@@ -490,16 +492,69 @@ def _layout(self):
...
+        parameters = self._procedure.parameter_objects()
+        for name in self._inputs:
+            parameter = parameters[name]
+            group = parameter.group_by
+            condition = parameter.group_condition
+            if group is None or group not in self._inputs or group == name:
+                continue
+
+            if isinstance(getattr(self, group), BooleanInput):
+                coondition = 2 if condition else 0
```
msmttchr, line 519: Typo in `condition` spelling. Sorry it was in my original comment.

pymeasure/display/widgets.py
```diff
@@ -490,16 +492,69 @@ def _layout(self):
...
+            elif isinstance(group_el, (IntegerInput, ScientificInput)):
+                group_el.valueChanged.connect(toggle)
+                toggle(group_el.value())
+            elif isinstance(group_el, ListInput):
+                group_el.currentTextChanged.connect(toggle)
+                toggle(group_el.currentText())
+            else:
+                raise NotImplementedError(
+                    "Grouping based on % (%) is not implemented." % (name, group_el))
+
```
msmttchr, line 557: I don't know the format character `%` standalone. Did you mean `%s` ?

CasperSchippers: Yes, you're right; I could've been a bit more careful with these last few commits

docs/tutorial/graphical.rst
```diff
@@ -338,5 +338,38 @@ This adds the input line above the Queue and Abort buttons.
...
+.. code-block:: python
+
+    iterations = IntegerParameter('Loop Iterations', default=100)
+    param = FloatParameter('some parameter', group_by='iterations', group_condition=99)
+
+Here the input of :code:`param` is only visible if :code:`iterations` has a value of 99.
+This works with any type of :code:`Parameter` as :code:`group_by` parameter.
+
+To allow for even more flexibility, it is also possible to pass a (lambda)function as a condition:
+.. code-block:: python
```
msmttchr, line 365: Minor point: blank line before line 365 to make Sphinx happy

CasperSchippers: ahh, yes; fixed

CasperSchippers:
I also added the posibillity to provide multiple parameters as `group_by` arguments, for when the visibility of an input is to depend on multiple other inputs

docs/tutorial/graphical.rst
```diff
@@ -397,5 +397,51 @@ The estimates are automatically updated every 2 seconds.
...
+Finally, it is also possible to provide multiple parameters to the :code:`group_by` argument, in which case the input will only be visible if all of the conditions are true.
+Multiple parameters for grouping can either be passed as a dict of string: condition pairs, or as a list of strings, in which case the `group_conditions` can be either a single condition or a list of conditions:
+
+.. code-block:: python
+
+    iterations = IntegerParameter('Loop Iterations', default=100)
+    toggle = BooleanParameter('A checkbox')
+    param_A = FloatParameter('some parameter', group_by=['iterations', 'toggle'], group_condition=[lambda v: 50 < v < 99, True])
+    param_B = FloatParameter('some parameter', group_by={'iterations': lambda v: 50 < v < 99, 'toggle':, True]})
+
```
msmttchr, line 443: Some minor errors in `param_B` example.
The part:
 - `group_by={'iterations': lambda v: 50 < v < 99, 'toggle':, True]}`
should be replaced by:
 - `group_by={'iterations': lambda v: 50 < v < 99, 'toggle': True}`

docs/tutorial/graphical.rst
```diff
@@ -397,5 +397,51 @@ The estimates are automatically updated every 2 seconds.
...
+To allow for even more flexibility, it is also possible to pass a (lambda)function as a condition:
+
+.. code-block:: python
+
+    iterations = IntegerParameter('Loop Iterations', default=100)
+    param = FloatParameter('some parameter', group_by='iterations', group_condition=lambda v: 50 < v < 99)
+
+Now the input of :code:`param` is only shown if the value of :code:`iterations` is between 50 and 99.
+
```
msmttchr, line 431: Sorry a bit pedantic observation, actually should be between 51 and 98. Same also below.

pymeasure/experiment/parameters.py
```diff
@@ -32,14 +32,40 @@ class Parameter(object):
...
+            self.group_by = {group_by: group_condition}
+        elif isinstance(group_by, (list, tuple)) and all(isinstance(e, str) for e in group_by):
+            if isinstance(group_condition, (list, tuple)):
+                self.group_by = {g: c for g, c in zip(group_by, group_condition)}
+            else:
+                self.group_by = {g: group_condition for g in group_by}
+        elif group_by is not None:
+            raise TypeError("The provided group_by argument () of is not valid, should be either"
+                            "a string, a list of strings, or a dict of string: condition pairs.")
+
```
msmttchr, line 66: Minor point, I could not fully understand the part "() of" when an exception is raised.
Also  `either"` should have a trailing space `either "`

msmttchr (commented):
Great job!!!

docs/tutorial/graphical.rst
CasperSchippers: Ahh yes indeed, thanks

docs/tutorial/graphical.rst
CasperSchippers: Ahh yes indeed, thanks

pymeasure/experiment/parameters.py
CasperSchippers: Good catch, I have no clue where the "() of" came from; it's removed now.

msmttchr (approved):
Looks good to me

Issue #416, opened 2021-03-20
Instrument's classes


msmttchr:
It is a reality, for various reasons, that similar instruments from different vendors exist in the market. So when we do a lab activity, we often use generic name, eg: oscilloscope and not Tektronix TDS2000. In general, unless there's a specific need, we consider instruments belonging to same class interchangeable and so, if one is broken or in use, we can use a different one (model or vendor or both) to do the job.
Obviously, human beings have normally the ability to adapt to new situation and they would use the new instrument by reading manual or guessing based on experience.
One may wonder why this is not accomplished in an instrument automation framework.
Indeed, SCPI instrument defines a set of generic commands valid for every instrument and a set of commands valid for some identified instrument  classes (see https://en.m.wikipedia.org/wiki/Standard_Commands_for_Programmable_Instruments and http://www.ivifoundation.org/docs/scpi-99.pdf).
I have also looked at some specific instrument implementation which is often basic and/or partial. I think there's nothing wrong with that, we are often very busy and we tend to implement what's strictly needed, an object oriented language like python allow to manage these situations without any problem.
My proposal is to identify some very well known instrument classes and starts implementing them as a generic one, and for each applicable vendor subclass the generic one.
This would have the advantage of:
- Reducing code duplication
- Making testing simple
- Allowing to grow the list of instruments quickly
- Allowing to use an instrument of a certain class even though its implementation is available
- Allowing to implement only vendor specific enhancements in the specific subclass.
- Define common API in one place and allow instrument reuse/replacement.

If you have time, you could have a look at the example in development for spectrum analyser and RF signal generators (https://github.com/wnluke/pymeasure/blob/d7deaf6c8517788a200762e2c1adb9de8bb564fa/pymeasure/instruments/rf_signal_generator.py or https://github.com/wnluke/pymeasure/blob/6e4d78d71c4d8ad35ae949bd7c7599c83a5d36da/pymeasure/instruments/spectrum_analyzer.py)
Here, even though the instruments are extremely complex and no SCPI class (for spectrum analyser) is defined, it is possible to define a common API.

My proposal on a step by step fashion is:
  1 identify instrument classes
  2 identify difference between SCPI instruments and non SCPI instruments
  3 Implement base classes
  4 Refactor existing specific instrument implementation based on base classes

Obviously hierarchy can be deeper than just one level, for example some time communalities between same vendor instruments exists.

[...]

PR #453, opened 2021-06-06
Corrected property implementation

msmttchr:
Purpose of this pull request is double:
- Fix implementation of complete and status properties ( class attributes and not instance attributes)
- Add options property

Note that in the change, the TODO 'Determine case basis for the addition of these methods' has been removed


pymeasure/instruments/instrument.py
```diff
@@ -63,17 +63,36 @@ class Object(object):
...
         log.info("Initializing %s." % self.name)
 
+    @property
+    def complete(self):
+        """ This property allows synchronization between a controller and a device. The Operation Complete 
+        query places an ASCII character 1 into the device's Output Queue when all pending
+        selected device operations have been finished.
+        """
+        if self.SCPI:
+            return self.adapter.ask("*OPC?").strip()
```
CasperSchippers, line 76: It might be nicer to use `self.ask` (which in turn calls `self.adapter.ask`)

CasperSchippers (commented):
Nice contribution. Something for consideration: would it be good to also make the `clear` and `reset` methods into similar case-based methods? And to make a standard (for SCPI at least) implementation of the `check_errors` methods?

pymeasure/instruments/instrument.py
msmttchr: Ok. I will extend the change also to `id`

msmttchr:
> Nice contribution. Something for consideration: would it be good to also make the `clear` and `reset` methods into similar case-based methods? And to make a standard (for SCPI at least) implementation of the `check_errors` methods?

You mean adding the same check on `self.SCPI` for `clear` and `reset` methods ?
I think it is also a very useful idea to add a default implementation for `check_errors` method.

I would like to have your opinion on some aspects:

1. When the check on `self.SCPI` fails, is it better to raise an exception ?
2. For `check_errors`, the description is not very detailed, but following some implementations found in the tree, I would do something like:
```
    def check_errors(self):
        """ Read all errors from the instrument.
        
        :return: list of error entries
        """
        if self.SCPI:
            errors = []
            while True:
                err = self.values("SYST:ERR?")
                if int(err[0]) != 0:
                    errors.append(errmsg)
                else:
                    break

            return errors
        else:


```
do you agree ?

CasperSchippers:
> You mean adding the same check on self.SCPI for clear and reset methods ?

That's indeed what I meant (these are also standard SCPI commands as far as I understood, or are they used more common?).


> When the check on self.SCPI fails, is it better to raise an exception ?

If mean that `SYST:ERR?` returns an error (was that what you meant by failing?) I think an exception should not be raised, as the error-checking is also used within the `control`, `settings`, and `measurement` properties; I think the best handling is logging the error and returning them, but I'd also like to pull in @bilderbuchi for his opinion on this matter.

> For check_errors, the description is not very detailed, but following some implementations found in the tree, I would do something like:

That's indeed the implementation I would use; I would like to suggest to add logging of the error (as in with `log.error`) as `check_error` is also used in the `control`, `settings`, and `measurement`, where the returned value is not used. (Also I think `errors.append(errmsg)` should read `errors.append(err)`, as  `errmsg` is not defined.)

msmttchr:
> > You mean adding the same check on self.SCPI for clear and reset methods ?
> 
> That's indeed what I meant (these are also standard SCPI commands as far as I understood, or are they used more common?).
> 
> > When the check on self.SCPI fails, is it better to raise an exception ?
> 
> If mean that `SYST:ERR?` returns an error (was that what you meant by failing?) I think an exception should not be raised, as the error-checking is also used within the `control`, `settings`, and `measurement` properties; I think the best handling is logging the error and returning them, but I'd also like to pull in @bilderbuchi for his opinion on this matter.
> 

I am sorry, my question was not really clear, what I meant was if `self.SCPI` is `False`, at the moment a warning message is returned, but, in my opinion, it could be better to raise an exception similarly to what I proposed in the `check_errors` method. Do you agree ?

> > For check_errors, the description is not very detailed, but following some implementations found in the tree, I would do something like:
> 
> That's indeed the implementation I would use; I would like to suggest to add logging of the error (as in with `log.error`) as `check_error` is also used in the `control`, `settings`, and `measurement`, where the returned value is not used. (Also I think `errors.append(errmsg)` should read `errors.append(err)`, as `errmsg` is not defined.)

Thanks for spotting this typo. I will add it and also the logging.

pymeasure/instruments/instrument.py
```diff
@@ -63,22 +63,41 @@ class Object(object):
...
+            raise NotImplementedError("Non SCPI instruments require implementation in subclasses")
+
     @property
     def id(self):
         """ Requests and returns the identification of the instrument. """
         if self.SCPI:
-            return self.adapter.ask("*IDN?").strip()
+            return self.ask("*IDN?").strip()
         else:
             return "Warning: Property not implemented."
```
CasperSchippers, line 102: I think you missed a error message -> NotImplementedError

CasperSchippers:
> I am sorry, my question was not really clear, what I meant was if self.SCPI is False, at the moment a warning message is returned, but, in my opinion, it could be better to raise an exception similarly to what I proposed in the check_errors method. Do you agree ?

Ahh, is see; yes I definitely agree that raising a `NotImplementedError` is a lot better.

pymeasure/instruments/instrument.py
msmttchr: Fixed. Sorry about this.

pymeasure/instruments/instrument.py
```diff
@@ -293,23 +312,42 @@ def fset(self, value):
     def clear(self):
         """ Clears the instrument status byte
         """
-        self.write("*CLS")
+        if self.SCPI:
+            self.write("*CLS")
+        else:
+            raise NotImplementedError("Non SCPI instruments require implementation in subclasses")
 
     # TODO: Determine case basis for the addition of this method
```
bilderbuchi, line 320: that TODO can also be removed now :-)

pymeasure/instruments/instrument.py
```diff
@@ -293,23 +312,42 @@ def fset(self, value):
...
+            self.write("*RST")
+        else:
+            raise NotImplementedError("Non SCPI instruments require implementation in subclasses")
 
     def shutdown(self):
         """Brings the instrument to a safe and stable state"""
         self.isShutdown = True
         log.info("Shutting down %s" % self.name)
 
     def check_errors(self):
```
bilderbuchi, line 333: Great, thank you for adding this generic implementation!
Could you please also remove all the instrument-specific `check_errors` that are now being made redundant by this change?

bilderbuchi (changes requested):
This looks great! It's great that the behaviour is consistent in these properties. I only have two remarks/requests, when those are resolved this looks good to merge!

Raising NotImplementedErrors on non-scpi instruments is a great workaround around the class-based nature of properties. I think when this is merged we can finally close #170 (and also #313 whose changes are contained here).

bilderbuchi:
Thanks for pinging me! I reviewed the PR. Thank you for the consistent quality of your contributions, Michele!

CasperSchippers (approved):
Looks good to me when the aforementioned remarks are resolved.

pymeasure/instruments/instrument.py
msmttchr: Thanks for the review, I had a second thought: I really don't like all that `if self.SCPI` that are against proper use of classes. What about having having `Instrument` class as a base class for SCPI and non-SCPI instruments and then having a proper subclass `SCPIInstrument` which implements all the mandatory SCPI commands ?
That is essentially moving all the following properties to subclass `SCPIInstrument`:

- `complete`
- `status`
- `options`
- `id`

and changing methods: `clear`, `reset`, `shutdown`, `check_errors` with 'Subclass should implement this' exception and make the SCPI implementation in `SCPIInstrument`. Also parameter `includeSCPI` should be removed from `__init__`.
After this, all the implemented instruments that are SCPI compliant should be changed to inherits from `SCPIInstrument` and remove the possible parameter `includeSCPI` and `check_errors` implementation. All the other should be left untouched.

bilderbuchi: Fair point. I'm a bit hesitant to make the inheritance structure more complicated, but in this case it's probably straightforward enough - `SCPIInstrument` isa `Instrument` is a pretty clear-cut relationship. Also, the documentation needs to be updated.

Normally, I'd try nowadays to add this additional behaviour as a Mixin (to favour composition over inheritance), e.g. like [here](https://stackoverflow.com/a/39471532), but it makes the correct inheritance formulation more fragile for our instrument contributors (correct order, correct use of `super()`), and I currently don't expect that we need more classes in the inheritance tree.

However, *if* we go towards these "reuseable, common-functionality" instrument classes (SignalGenerator and such) like you proposed in #416, this will get more complicated to deal with cleanly. Suddenly there is another axis to consider - to we make `SignalGenerator` *and* `SCPISignalGenerator` (to me this feels quite unnecessarily complicated)? 
As the scpi command set can be considered a feature/trait, maybe it would be better to use an `SCPIMixin`, after all, and any concrete implementation would then be `class KeithleyXYZ(SCPIMixin, SignalGenerator)` or `class RSXYZ(SignalGenerator)`? We then have to educate contributors on the importance of `super()`, order of base classes/MRO, and `*args, **kwargs` pass-through in initialisers. Probably this can best be sorted out with good Instrument templates/documentation.

Think about it!

CasperSchippers: It sounds like a very straightforward implementation, and I indeed can see how it is a bit neater than the using the `includeSCPI` option. For the sake of the contributors I'm not sure if I'm a big fan of the Mixin approach (I understand how it would be nicer, but I also think that in order to keep contributing instruments as accessible as possible it would be nice if a specific instrument only has to inherit a single superclass).

> to we make SignalGenerator and SCPISignalGenerator (to me this feels quite unnecessarily complicated)

I'm not so sure if that would actually be required: are there actually any shared commands between non-SCPI signal generators, as I'd think that all/most shared commands have the stem from the SCPI standard (but maybe I'm wrong though).

bilderbuchi: If I understand correctly what you are saying, you argue that, should we implement #416, we base all the new instrument "categories" on `SCPIInstrument` (naming either `SCPISignalGenerator` or just the simpler `SignalGenerator` might be preferable), and any non-scpi devices (either just not complying to one or more standard commands, or e.g. thorlabs devices) are based solely on `Instrument`. Correct?
This means we lose the opportunity to classify those non-scpi devices, as they are `Instruments`, but that might be an acceptable drawback.

msmttchr: @bilderbuchi, thanks for highlighting these very sharp points which I did not see before your comment. I understand and I share your concerns regarding having multiple dimension and multiple inheritance with #416, my gut feelings is that we could implement the `SCPIInstrument` class and then adding instrument categories either as a Mixin or directly inheriting from `SCPIInstrument`. I see some tendency with T&M manufacturers to make more compact multi-function devices and in such case, I think, a Mixin approach would be good but with more complicated inheritance structure which, at the end of the day, probably reflect the fact that instrument is more complex.

I think some kind of decision needs to be made: shall I try to implement `SCPIInstrument` as described in https://github.com/pymeasure/pymeasure/pull/453#discussion_r651306658 ? Or just implement the changes requested in this PR and leave the `SCPIInstrument` changes for another PR ?
@CasperSchippers , @bilderbuchi any preference/suggestion from your side ?

bilderbuchi: Another point to consider: 
One thing we (I think) want from the instrument categories (#416) is instrument API consistency. 
With the possible exception of the standard SCPI commands, this should be independent from SCPI-conformance or not. So, we will need some kind of mechanism that prescribes the available commands. If that should happen via Mixin (SCPI or not), or via inheritance and subsequent overriding of properties in the subclass, is not clear at this point I would say. This needs more discussion/thinking IMO.

I think it's safer to 
* wait with the `SCPIInstrument` implementation (the current solution is already in place and not so ugly)
* implement the requested changes to get this PR merged
* In #416, link to this discussion/the most valuable points so that they are not lost
* and add the `SCPIInstrument` feature/improvement (mixin or subclass or ...) to the scope of #416.

Let's see if @CasperSchippers agrees :-)

CasperSchippers: I think I agree with that, @bilderbuchi. For now indeed I'd finish this and merge the PR and add the SCPIInstrument implementation to the scope of #416.

> If I understand correctly what you are saying..... 

Yes that was indeed what I meant (I was mainly thinking from the point of view of shared commands), though I agree with your point regarding API consistency (although I'm not yet sure how to best implement it at this moment).

bilderbuchi:
Looks good now, thank you @CasperSchippers @msmttchr for the discussion!

PR #455, opened 2021-06-10
init, start stop function

hududed:
added a simple start stop function for edwards NXDS 10i pump

PR #456, opened 2021-06-10
Lighthouse photonics

hududed:
added functions to set operation mode and power on the Lighthouse photonics sprout G-12W laser systems


msmttchr:
Hi, I noticed that this PR include also PR #455, is this intentional ? If yes, I think it would be better to separate to have a single PR for each instrument.

hududed:
this was an accident, not sure how to solve this - I reverted the commit and pushed to origin. Do I wait until #455 is pushed and then recommit PR #456?

msmttchr:
I am not sure what is the best approach, but if the instruments are
independent, I would do a branch for each instrument from my fork's master.
Then I will create a pull request from each independent branch. In your
case, I guess, you created the lighthouse branch from edwards branch and so
you included the changes from it.
I would close #456 and make a new lighthouse branch from your fork master
branch and create a new pull request from the new branch.

Il Ven 25 Giu 2021, 20:48 hududed ***@***.***> ha scritto:

PR #459, opened 2021-06-29
new instrument: Oxford IPS120-10

CasperSchippers:
Adds the oxford IPS120-10 power supply for superconducting magnets


pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+            raise Exception("IPS not at 0T, not disabling the supply")
+
+        self.switch_heater_status = 0
+        self.activity = "clamp"
+
+    def enable_persistent_mode(self):
+        """ Methods that enables the persistent magnetic field mode. """
+        # Check if system idle
+        if not self.sweep_status == "at rest":
+            raise Exception("Magnet not at rest")
```
bilderbuchi, line 235: Might make sense to create custom exceptions in this module (`MagnetError`, and `SwitchStatusError`, deriving from `ValueError`?) for easier handling. Otherwise, whoever handles these exceptions would need to check for the exception message to tell this apart from, e.g. a `TimeoutError`.

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+        """ A integer property that returns the contactor status of
+        the IPS in Tesla. """,
+        get_process=lambda v: int(v[14]),
+    )
+
+    switch_heater_status = Instrument.control(
+        "X", "$H%d",
+        """ A integer property that returns the switch heater status of
+        the IPS in Tesla. """,
+        get_process=lambda v: int(v[8]),
```
bilderbuchi, line 146: do these ints also map to human-understandable strings, like for sweep_status?

pymeasure/instruments/oxfordinstruments/ips120_10.py
bilderbuchi: similar for the next method

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+            try:
+                status = self.sweep_status
+            except Exception as e:
+                log.error("IPS: Issue with getting status (#%d): %s" % (error_ct, e))
+                error_ct = 0
+
+            if status == "at rest":
+                break
+
+            if error_ct > max_errors:
```
bilderbuchi, line 290: you seem to never increment `error_ct`, so can that ever become true (except for negative `max_errors`?) :-)

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+            self.wait_for_idle()
+            self.activity = "hold"
+            self.switch_heater_status = 1
+            log.info("IPS: Wait for 20s for switch heater")
+            sleep(self._SWITCH_HEATER_DELAY)
+        else:
+            raise Exception("Switch status returned %d" % switch_status)
+
+    def wait_for_idle(self, max_errors=10, delay=1):
+        error_ct = 0
```
bilderbuchi, line 276: I am assuming that this is supposed to wait for the status to become "at rest" after some operation. In that case, maybe a better interface would be for the user to define `max_wait_time` they are prepared to accept, and you'll then divide that into a reasonable number (>=10?) of sleep intervals between retries?

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+        status = None
+        while True:
+            sleep(delay)
+
+            try:
+                status = self.sweep_status
+            except Exception as e:
```
bilderbuchi, line 283: What errors would occur here? timeouts? is this a "flaky" instrument?

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+                error_ct = 0
+
+            if status == "at rest":
+                break
+
+            if error_ct > max_errors:
+                raise Exception("Too many exceptions occurred during getting IPS status.")
+
+    def set_field(self, field, sweep_rate=None, persistent_mode_control=True):
+        # Check if field needs changing
```
bilderbuchi, line 294: I feel like this method with involved logic would profit from a detailed docstring.

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+
+        switch_status = self.switch_heater_status
+        if switch_status == 1:
+            pass  # Magnet in demand mode
+        elif switch_status in [0, 2]:
+            # Magnet in persistent mode
+            if persistent_mode_control:
+                self.disable_persistent_mode()
+            else:
+                raise Exception("IPS in persistent mode and control of persistent mode not allowed")
```
bilderbuchi, line 306: So, if **not** `persistent_mode_control` the error message is "IPS in persistent mode"? I admit I am confused :smile:

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def set_field(self, field, sweep_rate=None, persistent_mode_control=True):
+        # Check if field needs changing
+        if self.field == field:
+            return
+
+        switch_status = self.switch_heater_status
+        if switch_status == 1:
+            pass  # Magnet in demand mode
+        elif switch_status in [0, 2]:
```
bilderbuchi, line 301: From a UX perspective I find it confusing that the switch heater status actually describes the magnet status/mode, but probably it's because I have no idea about the device :D

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+            self.field_setpoint = field
+
+        self.wait_for_idle()
+        sleep(10)
+
+        if persistent_mode_control and field != 0:
+            self.enable_persistent_mode()
+
+    def train_magnet(self, training_scheme=None):
+        """ Method that trains the magnet after cooling down """
```
bilderbuchi, line 324: Maybe define what `training_scheme` should look like in the docstring.

bilderbuchi (approved):
Good work as usual! I have some cosmetic comments and a couple of suggestions stemming from me being confused by the device intricacies. Feel free to merge after you think it's finished!

pymeasure/instruments/oxfordinstruments/ips120_10.py
bklebel: the switch heater does control whether the magnet is in the persistent mode or not (physically). if it is on, field can be changed in the magnet by the instrument, once you turn it off, the superconducting loop that is the magnet is being closed, and "control" is gone, but coolant consumption is reduced. turning the heater on breaks the loop, allowing the current to flow through the power supply, and thus to be controlled. Therefore, before turning the heater on, the current in the power supply must be the same as in the magnet. 
I was thinking about implementing this very same device here, @CasperSchippers beat me to it, and quite nicely so I would say! I like the implementation as far as I looked into it, but have not yet tested it.

pymeasure/instruments/oxfordinstruments/ips120_10.py
bklebel: I would prefer that too, over a pre-defined training scheme. The way it is now, the default scheme is good for only an (uncommonly) strong magnet, and I would much prefer it if the optional argument was given at initialisation, without an otherwise defined scheme. After all, this scheme is not a characteristic property of the device itself, but of the magnet it controls.

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+        (14.9, 0.2),
+        (16.0, 0.1),
+    ]
+
+    def __init__(self, resourceName, clear_buffer=True, switch_heater_delay=None, **kwargs):
+        super().__init__(
+            resourceName,
+            "Oxford IPS",
+            includeSCPI=False,
+            send_end=True,
```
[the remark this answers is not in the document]
bklebel: for comparison, we control our ips120 via rs232, and use the same termination for reading and sending (although what we have right now is rudimentary, but communication works with this termination)

docs/api/instruments/oxfordinstruments/IPS120_10.rst
```diff
@@ -0,0 +1,7 @@
+#############################################################################
+Oxford Instrument Intelligent Power Supply 120-10 for superconducting magnets
```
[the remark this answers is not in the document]
CasperSchippers: Thanks, fixed

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+        # values=_T_RANGE
+    )
+
+    sweep_rate = Instrument.control(
+        "R9", "$T%f",
+        """ A floating point property that controls the sweep-rate of
+        the IPS in Tesla/minute. """,
+        get_process=lambda v: float(v[1:]),
+        # validator=truncated_range,
+        # values=_T_RANGE
```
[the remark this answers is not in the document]
CasperSchippers: Indeed, I have removed this

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+        values=[0, _MAX_CURRENT]
+    )
+
+    field_setpoint = Instrument.control(
+        "R8", "$J%f",
+        """ A floating point property that controls the magnetic field set-point of
+        the IPS in Tesla. """,
+        get_process=lambda v: float(v[1:]),
+        # validator=truncated_range,
+        # values=_T_RANGE
```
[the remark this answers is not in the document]
CasperSchippers: sort of, yes; I added a `_FIELD_RANGE` attribute (field-range can be passed at initialization) and added it for validation here

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,330 @@
...
+        """ Method that enables active control of the IPS.
+        Sets control to remote and turns off the clamp. """
+        self.control_mode = "RU"
+
+        # Turn off clamping if still clamping
+        if self.activity == "clamp":
+            self.activity = "hold"
+
+        # Turn on switch-heater if field at zero
+        if self.field == 0:
```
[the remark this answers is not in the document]
CasperSchippers: That's a fair point, however, in this case I think the instrument does exactly return 0.0; if it doesn't I wouldn't turn on the switch-heater via the software (or at least not automatically as is done over here.

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: Makes perfect sense indeed; when I implemented it I simply did not have time for a nice implementation (just needed it to work properly since we had only access to the magnet for about a week).

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: Ah nice catch, thanks

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: It only gave me ValueError (from not being able to convert a string into a float because sometime it returned not completely the right format); I changed the `Exception` here to a `ValueError`

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: Yes, i completely agree, I've added a docstring to this and the other methods.

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: That's a good point; however, as I also like the possibility to have wait without maximum I've kept the delay and the max_wait_time as separate numbers. I've also added a `should_stop` argument.

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: You are right indeed; I've added to the docstring how the training_scheme should look like (and in the general-docstring of the class an example of how to use the instrument, including the train_magnet method. I've decided to remove the training scheme as an attribute from the class as I think you're right in stating that it is not an property of this device, and it is probably even not useful to define it for most instruments upon initialization, but only whenever used.

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: I've change the error to: "magnet is in persistent mode but cannot turn of persistent mode because persistent_mode_control == False. ", hopefully this is a bit more clear.

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: Yes they do; unfortunately, they're not as symmetric though: the value you set to the system do however not mean the same as the value you get from the system. I will try to make a property manually to get it a bit neater

pymeasure/instruments/oxfordinstruments/ips120_10.py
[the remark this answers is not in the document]
CasperSchippers: It indeed is a GPIB argument, though admittedly not very often used I think.

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: I've added an additional boolean property (`switch_heater_enabled`) that wraps around the `switch_heater_status` to take care of this.

bklebel:
yes, I think I will be able to test the implementation. Currently I have no access to the lab, but I will be back in the coming week(s), so I will go for it - in the end, I want to use our magnet eventually, and your implementation will help a lot in this direction!

bklebel:
Hi @CasperSchippers, I have started to look into this. 
Unfortunately I have run into a problem: The instruments from oxfordinstruments which we have in the lab are quite old, and sometimes rather misbehaving. it happens on a quite frequent but irregular basis, that the return value for a query is **empty**, or did not understand the command (i.e. sends back a question mark preceding a copy of the command which was not understood). 

to be sure, I have set additional serial connection specifics (I am not sure what the default values for these are): 
```
from pymeasure.instruments.oxfordinstruments import ips120_10
ips = ips120_10.IPS120_10("ASRL4::INSTR", query_delay=5e-1)  # clear_buffer=True)

from visa import constants as vconst
ips.adapter.connection.data_bits = 8
ips.adapter.connection.baud_rate = 9600
ips.adapter.connection.parity = vconst.Parity.none
ips.adapter.connection.stop_bits = vconst.StopBits.two


print(ips.version)
print(ips.version)
for _ in range(3):
    print(ips.ask("X"))

ips.control_mode = "RU"
print(ips.control_mode)

print("swtich heater")
ips.switch_heater_enabled = False
print(ips.switch_heater_enabled)

print("field_setpoint")
ips.field_setpoint = 0
print(ips.field_setpoint)

print("activity:")
print(ips.activity)

print("enable:")
ips.enable_control()
print(ips.activity)
print(ips.switch_heater_enabled)

ips.disable_control()
print(ips.activity)
print(ips.switch_heater_enabled)

```
- [x] I have configured all parameters for the serial connection for the adapter to be sure. 
- [x] I have introduced a query_delay to give it enough time after queries, does not help, although I have not gone beyond 2 seconds, but this starts to become unacceptable 

In the above example, the problem occurs at latest at `ips.enable_control()`, inside of which first the `control_mode` is set and then the `activity` is checked - I cannot go past this: `self.activity == "clamp"` . 
However, it happens from time to time that there is a problem earlier too. in order to get some feeling for it (using many more iterations than the 3 now depicted), I put the loop which requests the status of the device all the time, and sometimes I get an empty answer, sometimes a timeout error, and most of the time the answer is sane. 

I had the same problem with an oxford ITC 503 and ILM 211, which was one of the reasons why I did not take pymeasure for those in the first place: I don't know of a way in pymeasure to do a sanity check of an answer to a query, before this is then fed into the `get_process` lambda function. Currently, with an empty return message, I get an `IndexError`  when this is executed. 

To give some perspective, this is what I do (rather manually) to request a measured value (specified by `variable`) in our implementations for itc 503 and ilm 211: 
```
    def getValue(self, variable=0)
        value = self.query("R{}".format(variable))

        try:
            if value[0] != "R":
                try:
                    self.read()
                except VisaIOError as e_visa:
                    if (
                        isinstance(e_visa, type(self.timeouterror))
                        and e_visa.args == self.timeouterror.args
                    ):
                        pass
                return self.getValue(variable)
        except TypeError:
            try:
                self.read()
            except VisaIOError as e_visa:
                if (
                    isinstance(e_visa, type(self.timeouterror))
                    and e_visa.args == self.timeouterror.args
                ):
                    pass
            return self.getValue(variable)
        return float(value.strip("R+"))
```
Can we do such a kind of recursive checking? Or do you have some advice how to do this more elegantly? 
Or maybe there is some other issue with my connection you might have an idea about?

bilderbuchi:
Not getting a response is tough. I had some instruments that prepended nonsensical \0 or somesuch to replies sometimes - that I stripped out with `get_process`, but this here is not that easy.

One thing you could try is to subclass `VISAAdapter` and implement the error handling there (e.g. a `RetryingVISAAdapter`). Then you could use that adapter with a pymeasure instrument as usual.

By the way, in the above code, are those if statements doing anything? isn't ` if bla: pass, else pass` effectively a no-op?

bklebel:
Thanks, I think the suggestion with subclassing the `VisaAdapter` is a good approach! I will think about it, and make a separate PR. I guess I will put thoughts about "how to decide what a `sane` reply is" there too - after all, if I am going to make a general `RetryingVisaAdapter`, this should not be just only custom-tailored to our instruments. Although, for a start it might be enough to work there. 

Yes and no, this part of the code is one the one hand a reminder that I expect to see a `TimeoutError` potentially, which I want to ignore, however it should have raised the `e_visa` Error again in case it was something else, which I apparently forgot to put. Thanks for spotting it!

bilderbuchi:
I'm not sure if this is suitable as a general feature for pymeasure, I would suggest to first keep it local -- after all it only starts to be necessary because of some weirdness of your specific instruments.
Maybe later we can think how this can be made generic and flexible enough so that it's useful for a larger audience. 
So far, I fear that if we introduce helpers/tweaks/hacks based on many conceivable ways to go wrong, pymeasure scope (and maintenance burden) will explode, but we don't have enough tests to lend confidence that everything is correct.

bklebel:
Okay, sure, I understand - in the end, the prospect of thinking something up that works for everyone (i.e. flexible, maintainable, ...) was not all too pleasing to me. 

In any case, I will have to get this to work first, and will then come back to testing the magnet, I will let you know about my progress!

bklebel (commented):
I have managed to get the adapter going and test the device in a "dry-run" fashion, without the magnet actually being connected (using a short between the leads). All the simple properties seem to work fine, `set_field` fails almost silently right now. I get the same two notices from my custom adapter that some answer from the device was corrupt, but I do not really know what happens here, and will need to investigate more - I will introduce a bunch of debug logging messages, possibly fix whatever is wrong, and make a PR to your branch, @CasperSchippers. 

In the meantime, I have added just a few comments which came to my mind while testing. 
I hope I will be able to test the magnet "live" in at some point after mid-October.

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,457 @@
...
+        else:
+            if heater_on:
+                field = self.demand_field
+            else:
+                field = self.persistent_field
+
+        return field
+
+    def enable_control(self):
+        """ Method that enables active control of the IPS.
```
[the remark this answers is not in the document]
bilderbuchi: > "Do x by using y", in an active voice.

Indeed, that is best practice. However, we are glad if we get useful docstrings in the first place, and have not formulated a policy about docstring format -- at some point we can add `pydocstyle` to our CI and then invest the editing effort to get the docstrings cleaned up, but there's many more important things to do, first. :-)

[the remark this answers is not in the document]
bklebel: okay, sure, no big deal in the end :)

pymeasure/instruments/oxfordinstruments/ips120_10.py
bklebel: I guess this is also the problem which I faced. 
I have created a gist for the adapter I am using, possibly we can get rid of the error handling eventually:
https://gist.github.com/bklebel/1ccdc6ceae13fda3db78e75dee629988

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,457 @@
...
+    :param field_range: A numeric value or a tuple of two values to indicate the
+        lowest and highest allowed magnetic fields. If a numeric value is provided
+        the range is expected to be from :code:`-field_range` to :code:`+field_range`.
+
+    """
+
+    _MAX_CURRENT = 120  # Ampere
+    _MAX_VOLTAGE = 10  # Volts
+    _SWITCH_HEATER_DELAY = 20  # Seconds
+    _FIELD_RANGE = [-16, 16]  # Tesla
```
bklebel, line 95: I think it might be beneficial to reduce the default field range to a field range which is most common in magnets, and not necessarily one which is usually found in more special facilities. 
Therefore, I would recommend to change the range to [-7, 7]  # Tesla, since I don't know any magnets one would want to control with such a device which cannot achieve this.

bklebel:
just for reference, I have made a gist with my custom Adapter, it can be found here:
https://gist.github.com/bklebel/1ccdc6ceae13fda3db78e75dee629988
Possibly, using this might eliminate the need to do error handling within the function `wait_for_idle`.

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,457 @@
...
+        in the correct mode.
+    :param field_range: A numeric value or a tuple of two values to indicate the
+        lowest and highest allowed magnetic fields. If a numeric value is provided
+        the range is expected to be from :code:`-field_range` to :code:`+field_range`.
+
+    """
+
+    _MAX_CURRENT = 120  # Ampere
+    _MAX_VOLTAGE = 10  # Volts
+    _SWITCH_HEATER_DELAY = 20  # Seconds
```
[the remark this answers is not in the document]
CasperSchippers: Good suggestion, I'll make them separate.

pymeasure/instruments/oxfordinstruments/ips120_10.py
CasperSchippers: Fair enough, +-7 Tesla sounds like a decent range for most magnets.

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,457 @@
...
+    def set_field(self, field, sweep_rate=None, persistent_mode_control=True):
+        """ Method that changes the applied magnetic field.
+        If allowed (via `persistent_mode_control`) the persistent mode will be turned off
+        if needed and turned on when the magnetic field is reached.
+        When the new field set-point is 0, the set-point of the instrument will not be changed
+        but rather the `to zero` functionality will be used. Also, the persistent mode will not
+        turned on upon reaching the 0T field in this case.
+
+        :param field: The new set-point for the magnetic field.
+        :param sweep_rate: A numeric value that controls the rate with which to change
```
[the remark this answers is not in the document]
CasperSchippers: Indeed, fixed this, and added units to the docstrings in some other places as well.

pymeasure/instruments/oxfordinstruments/ips120_10.py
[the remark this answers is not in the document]
CasperSchippers: I've changed the docstrings of het methods to a more active form; for now, I left the properties as descriptive in line with all the other properties.

pymeasure/instruments/oxfordinstruments/ips120_10.py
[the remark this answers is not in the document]
bilderbuchi: for the record, when #539 lands, you can use this:
```
    def __init__(self, resourceName,
                 clear_buffer=True,
                 switch_heater_heating_delay=None,
                 switch_heater_cooling_delay=None,
                 field_range=None,
                 **kwargs):
        kwargs.setdefault('read_termination', '\r')
        super().__init__(
            resourceName,
            "Oxford IPS",
            includeSCPI=False,
            gpib={'send_end': True},
            **kwargs
        )
```
and this should work as expected for either connection.

bklebel:
Thanks to @CasperSchippers, it turns out that some of the problems I had with this PR were related to actually working on a slightly older model (PS instead of IPS), so that we agreed that it would be best to get both instruments implemented separately. 
Since the changes are minor, and mostly relate to the `get_process` and `set_process` parameters (floats vs ints), we agreed that it would be most elegant to use the features from #376 as soon as they are available, to base one instrument on the other. However, in the meantime, I hope it will be okay if we use a copy-paste approach, with two separate instruments (a small chunk of code duplication), and change it to cleaner code once #376 is through. 
@bilderbuchi and whoever else is responsible for sth like this is that okay with you?

bilderbuchi:
#376 is like one step from the finish line -- at this point we only need to add the documentation for the new feature, and modify/update any instruments we already have that could make use of the feature. 
I would wait for that to merge and then do it "properly". 
AFAICT, there is nothing left to do in this PR, or do you want to add some tooling based on #376 already to prepare for the other instrument? 
ping @CasperSchippers

pymeasure/instruments/oxfordinstruments/ips120_10.py
```diff
@@ -0,0 +1,470 @@
...
+            read_termination="\r",
+            **kwargs
+        )
+
+        if isinstance(self.adapter, VISAAdapter):
+            self.adapter.connection.baud_rate = 9600
+            self.adapter.connection.data_bits = 8
+            self.adapter.connection.parity = 0
+            self.adapter.connection.stop_bits = 20
+
```
bilderbuchi, line 134: ah, this should be updated accounting for the merged #539:
```suggestion
        kwargs.setdefault('read_termination', '\r')
        super().__init__(
            resourceName,
            "Oxford IPS",
            includeSCPI=False,
            gpib={'send_end': True},
            asrl={'baud_rate': 9600,
                  'data_bits': 8,
                  'parity': 0,
                  'stop_bits': 20,
                  },
            **kwargs
        )

```

bilderbuchi (changes requested):
Found an adaption to the now-removed safedefaults thing.

bklebel:
I have made a PR to CasperSchippers/pymeasure#5 for some additional updates (what I though was debugging, while we were working on similar but different devices), since I cannot change anything here afaik. Please see our small discussion there. Of course it would be possible to merge this now as is, with my updates being introduced in a later PR, after #376 has been merged.

bilderbuchi:
> I have made a PR to CasperSchippers#5 for some additional updates, since I cannot change anything here afaik.

Yes, this is the correct workflow. #376 has landed now.

CasperSchippers:
I reworked the adapter a bit (I hope you don't mind, @bklebel); main things are:
- Fixed some DRY
- Removed the "$"s that prefixed all the commands, which suppresses a response if it is not required; now always a response is expected, which is handled by the Oxford instruments adapter (this way you can check if a command is correctly set and it prevents you from waiting for a time-out every time you write something)

I tested it using the ITC503, which works perfectly so far. (I will add the adapter to the ITC in a separate PR, as I also have some other update to the ITC)

I think it is finished now.

bilderbuchi:
Looking good! I'll squash-merge this and compile a meaningful commit message. The large number of tweaking commits are useful for development, but in the pymeasure history it's just noise.

bklebel:
Well, I absolutely do not mind having the adapter reworked, and I am quite happy this is now merged, looking forward to using it in our system!

Issue #422, opened 2021-03-29
System Crash on Big Sur with Plotter


dph3dg:
I'm trying the tutorials and the moment I call plotter.start() in the example I get a segmentation fault.
The following is the errors thrown before the segmentation fault:
WARNING: QApplication was not created in the main() thread.
2021-03-29 14:21:59.858 python[23493:824564] WARNING: NSWindow drag regions should only be invalidated on the Main Thread! This will throw an exception in the future. Called from (
	0   AppKit                              0x00007fff22c8947f -[NSWindow(NSWindow_Theme) _postWindowNeedsToResetDragMarginsUnlessPostingDisabled] + 352
	1   AppKit                              0x00007fff22c74121 -[NSWindow _initContent:styleMask:backing:defer:contentView:] + 1375
	2   AppKit                              0x00007fff22c73bbb -[NSWindow initWithContentRect:styleMask:backing:defer:] + 42
	3   AppKit                              0x00007fff22f7e1f4 -[NSWindow initWithContentRect:styleMask:backing:defer:screen:] + 52
	4   libqcocoa.dylib                     0x000000011db44df0 qt_plugin_instance + 20859
)
2021-03-29 14:21:59.881 python[23493:824564] WARNING: nextEventMatchingMask should only be called from the Main Thread! This will throw an exception in the future.

CasperSchippers:
Hi. Thanks for raising this issue. What I understand is that the latest macOS don't like it when you try to render something (in this case the plotter-window) from anything else than the Mean Thread. While I'm trying to find a solution, I want to note that it is rather difficult for me as I don't have access to macOS.

Could you tell us if this problem only arises in this plotter example, or also with other examples, such as the ManagedWindow?

dph3dg:
I used the plotter example as it was the 1st one, so I assumed it was the easiest. The managed window opens as the example and I get a moving plot al'a Labview. The progress bar doesn't seem to update, just shows if it's finished or not.

CasperSchippers:
Yes the plotter is supposed to be the easiest (or at least most stripped down) version, but apparently the way it is called gives some problems on the newest macOS.

That the progress bar does not show the progress in this example is correct. It seems the example omits the line responsible for updating the progress: `self.emit('progress', 100*i/iterations)`; in the examples folder on GitHub this line is included in the GUI example.

msmttchr:
[...]

PR #460, opened 2021-07-05
Documentation bugfix

CasperSchippers:
Fixes a bug which causes the documentation-pages of all Qt-derived classes to not show in the documentation. Is fixed by updating the sphinx version to a version >3.4.2. Thanks to @msmttchr for reporting and finding the cause of the issue.

Also adds a warning to the `Using the Plotter` tutorial for macOS users (see issue #422).


msmttchr (approved):
Looks good to me.

bilderbuchi:
👍

CasperSchippers:
Yeah It might indeed be wise to generalize the warning a bit.

Issue #393, opened 2021-01-26
Information on plots


msmttchr:
I would like to plot two traces at the same time during an experiment, for example:
- plot a curve based on measurement at same time with a predicted one to appreciate the differences between model and measurement
- Plot a curve with experimental data and, when finished, make plot with a calculated interpolation curve 

Is this possible with pymeasure ?

CasperSchippers:
This can be quite challenging at the moment I think. As far as I know plotting two curves is not really possible.

It would of course be possible to add an additional data column to the experiment with a simulated curve, but then you'd still have to switch between the two columns (i.e. between the simulated and the measured curve). Another option would be to have first an "experiment" where you simulate the curve and then an experiment where you measure. But both are not ideal of course.

For the interpolated curve at the end, one possibility would be to add a column which will be NAN during the experiment, and then basically append the interpolation to the data (while then leaving the other real-data columns NaN). But this would of course pollute your data file with a lot of rows at the bottom just for the interpolated data (probably more rows than the data itself).

So for both cases there is i think not a really good solution at the moment.

msmttchr:
@CasperSchippers thanks for the feedback. I hope that #383 can progress in some way, so that people could redefine plot view.

PR #461, opened 2021-07-08
added a way to initialize multiple plot windows for 1 procedure

Casvharen:
Allows plotwidget to be initialized with multiple plot screens


pymeasure/display/manager.py
```diff
@@ -139,7 +139,9 @@ def __init__(self, plot, browser, port=5888, log_level=logging.INFO, parent=None
         self._monitor = None
         self.log_level = log_level
 
-        self.plot = plot
+        if not type(plots) == list:
```
bilderbuchi, line 142: I think it's more common to check this with `isinstance`

Casvharen: I will change that then

msmttchr:
> Allows plotwidget to be initialized with multiple plot screens

I think thi

> Allows plotwidget to be initialized with multiple plot screens

Interesting and useful extension (see also #393 )
I wonder if the modifications can be extended to:

- Have multiple curves on the same plot
- Have an option to plot on secondary y axis (similarly to Excel)

mcdo0486:
Hi,

I also was working on something similar https://github.com/mcdo0486/pymeasure/tree/multi_plot 

I went with subclassing ManagedWindow and Experiment and such for multi variants. 

But I'll checking out this implementation

mcdo0486:
@Casvharen I made a pull request for your repo to bring it up to date. There were some core changes in commit 3a97463822616d7414e25d59e6662c09bea00a0e that modified display/windows.py display/widgets.py display/manager.py

This pull request merges your multiple plot implementation with current master I believe.

I haven't updated my implementation yet. The one significant difference between our implementations is that I allow different X variable selection for the other graphs. Well, that and subclassing to make new files multiwindow and such. I wonder if I should make a new pull request for that setup or if we figure stuff out here? I don't know

Casvharen:
Thanks @mcdo0486 the updates look good. This is my first time contributing to pymeasure so I don't know which way would be preferred, since both our solutions seem to work. Maybe we could find the best solution if we discuss it in a discord call?

mcdo0486:
@Casvharen Same for me, just starting to contribute. 

I was thinking yesterday though that changes from 3a97463822616d7414e25d59e6662c09bea00a0e that were merged into core help organize the logic around widgets, experiments, and curves. I think we both found out how the logic of plot updates used to be spread out across windows and manager. Now with the TabWidget(object) metaclass helps manage that.

IMHO I think your implementation should be added so the default PlotWidget supports multiple plots w/ a shared X axis.

I'll put in the multi x axis implementation as a subclass of PlotWidget in another commit I think. 

Also IMHO, since I imagine there will be a lot more widgets incoming, it may make more sense to convert widgets.py to a folder and have individual widgets as files in that location. The current widgets.py is getting pretty long.

mcdo0486:
Also @Casvharen it's more important what the maintainers think, not so much myself, as they're the ones keeping everything together.

msmttchr:
I think a viable option could be to define a more powerful version of `PlotWidget` with an agreement on the features it should have. Then decide whether replace original `PlotWidget` or offer new one as a more powerful alternative.

See also whether https://github.com/pymeasure/pymeasure/pull/461#issuecomment-876649759 can be taken into account.

In general, I think this is great and very useful enhancement

Casvharen:
> I think a viable option could be to define a more powerful version of `PlotWidget` with an agreement on the features it should have. Then decide whether replace original `PlotWidget` or offer new one as a more powerful alternative.
> 
> See also whether [#461 (comment)](https://github.com/pymeasure/pymeasure/pull/461#issuecomment-876649759) can be taken into account.
> 
> In general, I think this is great and very useful enhancement

I don't really see the purpose in adding another plotwidget if we can make the changes such that with the default initialization, the basic functionality is kept. The extra additions you suggested do indeed seem nice, but with my understanding of pyqtgraph I have no idea yet how to that, so feel free to suggest something

msmttchr:
> > I think a viable option could be to define a more powerful version of `PlotWidget` with an agreement on the features it should have. Then decide whether replace original `PlotWidget` or offer new one as a more powerful alternative.
> > See also whether [#461 (comment)](https://github.com/pymeasure/pymeasure/pull/461#issuecomment-876649759) can be taken into account.
> > In general, I think this is great and very useful enhancement
> 
> I don't really see the purpose in adding another plotwidget if we can make the changes such that with the default initialization, the basic functionality is kept. The extra additions you suggested do indeed seem nice, but with my understanding of pyqtgraph I have no idea yet how to that, so feel free to suggest something

I am not an expert of pyqtgraph either :(, but I believe example https://github.com/pyqtgraph/pyqtgraph/blob/master/examples/MultiPlotWidget.py provide a reference for plot on secondary axis.
Hope this helps

mcdo0486:
For pyqtgraph examples, I recommend running the [ExampleApp.py ](https://github.com/pyqtgraph/pyqtgraph/blob/master/examples/ExampleApp.py) which loads all the examples and you can mess with them in a small IDE. Very cool.

For MultiPlotWidget, from what I can tell, that class isn't used much at in the other examples. They typically use win.addPlot() like what pymeasure does currently. My hunch is that MultiPlotWidget isn't as useful as just adding additional plots. 

As @Casvharen mentioned, this is about what should the defaults be for PlotWidget and when new classes should be created to extend functionality. I don't have an answer for defaults, if it should be multiplots or multicurves in a plot. It depends on your opinion, I guess, of what users would want for defaults. Many will probably just use PlotWidget. Maybe a discussion topic.

msmttchr:
Apparently a feature similar to this is being introduced in pyqtgraph: https://github.com/pyqtgraph/pyqtgraph/issues/1359. I just pointing out this, since it could make the implementation of this feature in pymeasure much simpler.

Interesting also https://github.com/pyqtgraph/pyqtgraph/issues/2010

bmoneke:
@Casvharen do you want to pick up this PR or may I close it? It can be reopened whenever you want to continue.

bmoneke:

PR #468, opened 2021-08-12
Fix bug in the Keithley 2700

CasperSchippers:
Fixes a bug where the `channels_from_rows_columns` method returns an `np.int32` rather than an `np.array` when both the `rows` and `columns` arguments are ints. This subsequently causes trouble with the validator when the output of this method is used for opening or closing channels.


msmttchr:
Sorry I closed it by mistake

PR #469, opened 2021-08-12
Add output low grounded property to Keithley 6221

CasperSchippers:
Adds a property to the Keithley 6221 that allows the low output of the triax connector to be connected to ground or float.


msmttchr (approved):
Looks good to me.

PR #474, opened 2021-09-02
Dev/tox tests

PeterPirog:
Tox framework has been added. It gives possibility to test code for multiple OS and multiple python versions.([Tox](https://tox.readthedocs.io/en/latest/index.html))

run: pip install tox
command to run tests:  tox -v

Specified python environments must be installed in your OS before runing tox.

```
[tox]
envlist={py36,py37,py38,py39}-{mylinux,mymacos,mywindows}

[testenv]

deps= pytest
platform = mylinux: linux
           mymacos: darwin
           mywindows: win32
commands_pre=python -m pip install --upgrade pip
commands=python setup.py test -v
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,6 @@
 # THE SOFTWARE.
 #
 
-
 from setuptools import setup, find_packages
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,8 @@ setup(
         "pandas >= 0.14",
         "pyvisa >= 1.8",
         "pyserial >= 2.7",
-        "pyqtgraph >= 0.9.10"
+        "pyqtgraph >= 0.9.10",
+        "PyQt5"
     ],
     extras_require={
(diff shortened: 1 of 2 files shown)


bilderbuchi:
Sorry, I don't see why we need this? We already have CI set up via Github actions and test in all OSes and Python versions, using Anaconda. This has been working for a long time! 
For what it's worth, I don't think we have discovered a python-version-specific issue in a looong time. I think it's similar but not that long for OS-specific issues,  so the benefit of using tox is limited.
 
Increasing test coverage would be something that would be welcomed, though! Admittedly, at least on the hardware-focused instrument side it's hard to write tests, but there is also the GUI half of pymeasure 😀

bilderbuchi:
Thank you for your interest! 
At the moment we are satisfied with our testing/CI setup, so I'd rather not be pulling tox into the project without a strong reason. As I said, testing extension would be much more urgently needed on the instrument/hardware side than on the multi-OS angle.

PR #481, opened 2021-09-18
Added LIA and ERR status byte read properties to the SRS Sr830 driver

samcondon4:
Small PR. Just adding two additional properties to read status bytes from the sr830. The queries for these status bytes don't follow SCPI so new properties need to be created.

### DIFF
--- a/pymeasure/instruments/srs/sr830.py
+++ b/pymeasure/instruments/srs/sr830.py
@@ -22,13 +22,40 @@
 # THE SOFTWARE.
 #
 
+import re
+import time
+import numpy as np
+from enum import IntFlag
 from pymeasure.instruments import Instrument, discreteTruncate
 from pymeasure.instruments.validators import strict_discrete_set, \
     truncated_discrete_set, truncated_range
 
-import numpy as np
-import time
-import re
+
+
+class LIAStatus(IntFlag):
+    """ IntFlag type that is returned by the lia_status property.
+    """
+    NO_ERROR = 0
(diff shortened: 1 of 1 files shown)


pymeasure/instruments/srs/sr830.py
```diff
@@ -89,6 +89,36 @@ class SR830(Instrument):
...
+           bit 2: RAM error
+           bit 3: unused
+           bit 4: ROM error
+           bit 5: GPIB error
+           bit 6: DSP error
+           bit 7: Math error
+           """,
+        get_process=lambda s: bin(int(s))[2:],
+    )
+
```
msmttchr, line 121: In both cases you return a bit mask which identifies the error conditions.
When users see the mask, they need to go back in the manual to decode the mask.
Now, python has some more descriptive data type for such cases.
For example, you could try to use `IntFlag` or `Flag` as suggested by @bilderbuchi
sometime ago in a similar situation.
In such a way users can directly see the meaning of the flags.

msmttchr (changes requested):
Thanks for your PR. I suggested some changes to ease readability of results for users.

pymeasure/instruments/srs/sr830.py
samcondon4: Good point. I had not used the IntFlag type before this. Just implemented this change.

msmttchr (approved):
Thanks for your work, looks good to me.

PR #484, opened 2021-10-02
Fix incorrect columns on showing data

CasperSchippers:
I stumbled on a bug where a dataset shows the wrong columns when that data is un-hidden.
It can be reproduced by unchecking a dataset (i.e. hiding that dataset), changing the shown column of the plotter (e.g. changing the x-column), and checking (i.e. showing) the dataset again.

Here I propose to fix this issue by re-setting the x and y columns (or z for an image) upon adding a curve to the plotframe.


msmttchr (approved):
Looks good to me

PR #485, opened 2021-10-04
Dev/instrument quick properties

samcondon4:
Added two properties to the instrument base class that allows multiple specific instrument subclass properties to be set with a single line as discussed in discussion #477. This implementation is exactly as described in my final comment of this discussion. I have found this feature to be quite useful for populating GUI elements as well as controlling instruments with a GUI without having to hard code sets and queries of every property that should be displayed/controlled.


pymeasure/instruments/instrument.py
```diff
@@ -101,6 +102,35 @@ def id(self):
...
+        return self.quick_props_list
+
+    @quick_properties_list.setter
+    def quick_properties_list(self, prop_list):
+        if type(prop_list) is str:
+            self.quick_props_list = [prop_list]
+        elif type(prop_list) is list:
+            self.quick_props_list = prop_list
+        else:
+            raise TypeError("Expected input of type str or list")
```
bilderbuchi, line 116: AFAICT, the type-checking is the only thing that this property does beyond a simple attribute access. However, you're already nicely signaling that you expect a _list_, so I think we can expect users to assign a list (single-element if need be).
I would drop the property and use a plain attribute `quick_properties_list` initialised to `[]` as you already do.

pymeasure/instruments/instrument.py
```diff
@@ -101,6 +102,35 @@ def id(self):
...
+
+    @property
+    def quick_properties(self):
+        prop_dict = {}
+        for prop_name in self.quick_props_list:
+            prop_dict[prop_name] = getattr(self, prop_name)
```
bilderbuchi, line 122: I don't know if you're a fan of comprehensions, but 
```
prop_dict = {prop_name: getattr(self, prop_name) for prop_name in self.quick_props_list}
```
should be a more concise way of generating this.

pymeasure/instruments/instrument.py
```diff
@@ -101,6 +102,35 @@ def id(self):
...
+        if type(prop_list) is str:
+            self.quick_props_list = [prop_list]
+        elif type(prop_list) is list:
+            self.quick_props_list = prop_list
+        else:
+            raise TypeError("Expected input of type str or list")
+
+    @property
+    def quick_properties(self):
+        prop_dict = {}
```
bilderbuchi, line 120: You can document the behaviour of the new property by adding a relevant docstring to this getter. You can also explain the relationship to `self.quick_properties_list`.

pymeasure/instruments/instrument.py
```diff
@@ -101,6 +102,35 @@ def id(self):
...
+
+        return prop_dict
+
+    @quick_properties.setter
+    def quick_properties(self, setter_dict):
+        for key in setter_dict:
+            if key in self.quick_props_list:
+                setattr(self, key, setter_dict[key])
+            else:
+                raise RuntimeError("Property {} has not been placed in the quick properties list!".format(key))
```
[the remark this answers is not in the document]
bilderbuchi: I agree with Casper here, on both counts. 
Another point to caution against a `quick_properties` property is that now we're "nesting" properties, we have properties acting on other properties, which introduces some mental complexity - up until now we could be ~sure that properties are instrument-related, which could be useful e.g. when enumerating an instrument's capabilities - with that suddenly there's this "meta"property which is not "like the others" (not sure if I convey that efficiently).

What do you think about this, instead:
`self.get_attributes(names=None)` queries one or more properties or attributes (getattr doesn't care). If `names` is not used, it falls back to a list `self.quick_attribute_list` as defined by the user.
`self.set_attributes(values)` sets the class attributes/properties as defined by the key-value pairs in the `values`.
Docstrings on both methods should explain the context and usage.

I think this achieves what you want, while at the same time being more generally useful, which is important because `Instrument` is the root of our driver hierarchy.

bilderbuchi (changes requested):
As expressed in the discussion, I'm not fully sold on the need for this, but if it helps and is useful for many people - fair enough! I have a couple of suggestions how to improve your proposed addition. See especially the last comment, which makes some of the previous remarks redundant.

BenediktBurger:
@samcondon4  do you want to continue this PR or may I close it?

samcondon4:
@BenediktBurger you can close this. Thanks!

Issue #486, opened 2021-10-05
Can not perform basic math operations on Parameters


bicarlsen:
I am getting strange errors when trying to perform math on Parameters.
+ When trying to do basic mathematical operations with a Parameter and a built in Python data type I get a `TypeError: unsupported operand type(s)`.
+ When trying to use `parameter.value` in the math I get the error `AttributeError: 'float' object has no attribute 'value'`

```python
class My_Procedure( Procedure ):
    f_param = FloatParameter( 'Float Parameter',   default = 0 )
    g_param = FloatParameter( 'Another Float Parameter',   default = 0 )

    def math( self ):
        total = self.f_param + self.g_param  # no error

        res_1 = 1* self.f_param  # TypeError: unsupported operand type(s) for *: 'int' and 'FloatParameter'
        
        f_val = self.f_param.value  # no error
        res_2 = 1* self.f_param.value  # AttributeError: 'float' object has no attribute 'value'
```

msmttchr:
Strange problem, I would expect the following:
- line `sum = f_param + g_param` should actually give an error of variable not defined
- line `res_1 = 1* self.f_param` should be ok.
- line `f_val = self.f_param.value` and `res_2 = 1* self.f_param.value` both give errors

I have checked also my assumption using the `gui.py` basic example, adding your code and a call to `math` method.
Something weird going on, how do you trigger the problem ?

bicarlsen:
I think I've narrowed down the problem. It seems to be occurring because I am doing doing the arithmetic in an `@property` method. i.e.
```python

class My_Procedure( Procedure ):
    f_param = FloatParameter( 'Float Parameter',   default = 0 )
    g_param = FloatParameter( 'Another Float Parameter',   default = 0 )

     @property
     def my_property( self ):
        total = self.f_param + self.g_param  # no error

        res_1 = 1* self.f_param  # TypeError: unsupported operand type(s) for *: 'int' and 'FloatParameter'
        
        f_val = self.f_param.value  # no error
        res_2 = 1* self.f_param.value  # AttributeError: 'float' object has no attribute 'value'

    def my_method( self ):
        # everything works fine
        total = self.f_param + self.g_param 
        res_1 = 1* self.f_param  
        f_val = self.f_param.value 
        res_2 = 1* self.f_param.value
```

I forgot to include that detail in my original post. Apologies about that.

To answer your question `print (self.voltage_step, type(self.voltage_step))` outputs `0.01 <class 'float'>`.

[...]

PR #489, opened 2021-10-07
Fix procedure property issue

msmttchr:
Fix Issue #486


msmttchr:
Sorry a bit of mess with GIT, but now PR should be OK for review.

PR #494, opened 2021-10-14
Change of pen width for pyqtgraph

maederan201:
Following this discussion: https://github.com/pymeasure/pymeasure/discussions/405#discussioncomment-1367841 I implemented the proposed change to a width of 1 for all pyqtgraph pens. I included a comment stating the previous value and a pointer to the related pyqtgraph issue (https://github.com/pyqtgraph/pyqtgraph/issues/533). This increases the performance of the PlotterWidget for Plots with many (>1k) datapoints.

This PR is only intended as a temporary fix and should be reverted once the issue mentioned above is fixed.

### DIFF
--- a/docs/about/authors.rst
+++ b/docs/about/authors.rst
@@ -47,3 +47,4 @@ The following developers have contributed to the PyMeasure package:
 | Nicola Corna
 | Robert Eckelmann
 | Sam Condon
+| Andreas Maeder


bilderbuchi:
Would you be so kind as to open an issue for undoing this workaround,  linking to the pyqtgraph issue, so that we don't lose track of this and can check from time to time? Thanks!

Issue #248, opened 2020-02-25
Unit support


bilderbuchi:
I think it could be quite useful to add unit support, using pint, similar to how [InstrumentKit](https://instrumentkit.readthedocs.io/en/latest/devguide/util_fns.html#unitful-property) does it (currently transitioning from quantities to pint): Accept and return unitful values (pint `Qantity`), but also deal with plain floats optionally (then assuming a unit).

I think it should be possible to make this optional, so not break existing instruments, but give the opportunity to add new ones/upgrade existing ones with units.

Any opinions on this?

maederan201:
#499 (Keysight N7776C) would be an instrument which would benefit from this implementation and could be used for testing as well.

bmoneke:
I'm working on an _adapter_ to make instrumentKit instruments appear as if they were from pymeasure.
The largest difficulty, I see, is, that the instrumentKit instruments always return a pint `Quantity` (number and unit) and not numbers alone. Right now, the pymeasure GUI wouldn't know how to handle them.

What's about making pymeasure able to handle pint `Quantity`?
I would add it as an option. That way instruments may use pint or not.

I did not use pymeasure's GUI and cannot estimate the amount of work.

I think it is good to be unit aware, but would not enforce it (but recommend it) for new instruments.

bilderbuchi:
Great to hear! I'm assuming you mean "adapter" in the general sense, not in `Adapter` subclass? (I couldn't make sense of the latter interpretation)

> always return a pint Quantity (number and unit) and not numbers alone. Right now, the pymeasure GUI wouldn't know how to handle them.

Two points: 
* Pint has [wrappers](https://pint.readthedocs.io/en/stable/wrapping.html) to bridge between unit-aware and unitless code sections.
* Pint Quantities have [`m_as()`](https://pint.readthedocs.io/en/stable/developers_reference.html?highlight=m_as#pint.Quantity.m_as) to convert a unitful quantity to a plain float.

Maybe you can use either of those to adapt between pymeasure and IK?

> What's about making pymeasure able to handle pint Quantity?
I would add it as an option. That way instruments may use pint or not.

Indeed, that's what this issue is about! We should first find a good, working design before implementing it. 
Some thoughts about potential approaches that I had in the back of my head were to
* Add a `unit` attribute to the property factories to indicate which unit the property is using. Now that we have dynamic properties, we can even change that at runtime! Very useful e.g. when you can change the unit in the instrument.
[...]

PR #499, opened 2021-10-20
New Instrument: Keysight N7776C Swept Laser Source

maederan201:
Basic instrument driver for Keysight N7776C swept laser source

### DIFF
--- a/pymeasure/instruments/keysight/keysightN7776C.py
+++ b/pymeasure/instruments/keysight/keysightN7776C.py
@@ -35,7 +35,7 @@ WL_RANGE = [1480,1620]
 
 class N7776C(Instrument):
     """
-    This represents the Keysight N7776C Tunable Laser Source interface
+    This represents the Keysight N7776C Tunable Laser Source interface.
     .. code-block:: python
         laser = N7776C( address )
         laser.sweep_wl_start = 1550
@@ -125,6 +125,13 @@ class N7776C(Instrument):
                                     """ Sweep mode of the swept laser source """,
                                     validator=strict_discrete_set,
                                     values=['STEP','MAN','CONT'])
+    sweep_twoway = Instrument.control('sour0:wav:swe:rep?','sour0:wav:swe:rep %s',
+                                    """Sets the repeat mode. Applies in stepped,continuous and manual sweep mode.""",
+                                    map_values=True,
+                                    validator=strict_discrete_set,
+                                    values=[True,False],
+                                    set_process=lambda v: ['ONEW','TWOW'][int(v)],
+                                    get_process=lambda v: bool(['ONEW','TWOW'].index(v)))
     _sweep_check = Instrument.measurement('sour0:wav:swe:chec?',
                                     """Returns whether the currently set sweep parameters (sweep mode, sweep start, stop, width, etc.) are consistent. If there is a
                                     sweep configuration problem, the laser source is not able to pass a wavelength sweep.""")
--- a/pymeasure/instruments/keysight/keysightN7776C.py
+++ b/pymeasure/instruments/keysight/keysightN7776C.py
@@ -56,30 +56,30 @@ class KeysightN7776C(Instrument):
             address, "N7776C Tunable Laser Source", **kwargs)
 
     locked = Instrument.control(':LOCK?', ':LOCK %g,'+str(LOCK_PW),
-                                """ Boolean property controlling the lock state (True/False) of \
+                                """ Boolean property controlling the lock state (True/False) of
                                 the laser source""",
                                 validator=strict_discrete_set,
                                 map_values=True,
                                 values={True: 1, False: 0})
 
     output_enabled = Instrument.control('SOUR0:POW:STAT?', 'SOUR0:POW:STAT %g',
-                                        """ Boolean Property that controls the state (on/off) of \
+                                        """ Boolean Property that controls the state (on/off) of
                                         the laser source """,
                                         validator=strict_discrete_set,
                                         map_values=True,
                                         values={True: 1, False: 0})
 
     _output_power_mW = Instrument.control('SOUR0:POW?', 'SOUR0:POW %f mW',
-                                          """ Floating point value indicating the laser output power \
+                                          """ Floating point value indicating the laser output power
                                           in mW.""",
                                           get_process=lambda v: v*1e3)
 
     _output_power_dBm = Instrument.control('SOUR0:POW?', 'SOUR0:POW %f dBm',
-                                           """ Floating point value indicating the laser output power \n
+                                           """ Floating point value indicating the laser output power
                                            in dBm.""")
 
     _output_power_unit = Instrument.control('SOUR0:POW:UNIT?', 'SOUR0:POW:UNIT %g',
-                                            """ String parameter controlling the power unit used internally \n
+                                            """ String parameter controlling the power unit used internally
                                             by the laser.""",
                                             map_values=True,
                                             values={'dBm': 0, 'mW': 1})
@@ -103,7 +103,7 @@ class KeysightN7776C(Instrument):
         self._output_power_dBm = new_power
 
     trigger_out = Instrument.control('TRIG0:OUTP?', 'TRIG0:OUTP %s',
-                                     """ Specifies if and at which point in a sweep cycle an output trigger \n
+                                     """ Specifies if and at which point in a sweep cycle an output trigger
                                      is generated and arms the module. """,
                                      validator=strict_discrete_set,
                                      values=['DIS', 'STF', 'SWF', 'SWST'])
@@ -145,41 +145,42 @@ class KeysightN7776C(Instrument):
                                     validator=strict_discrete_set,
                                     values=['STEP', 'MAN', 'CONT'])
     sweep_twoway = Instrument.control('sour0:wav:swe:rep?', 'sour0:wav:swe:rep %s',
-                                      """Sets the repeat mode. Applies in stepped,continuous and \
+                                      """Sets the repeat mode. Applies in stepped,continuous and
                                       manual sweep mode.""",
                                       validator=strict_discrete_set,
                                       map_values=True,
                                       values={False: 'ONEW', True: 'TWOW'})
-    _sweep_check = Instrument.measurement('sour0:wav:swe:chec?',
-                                          """Returns whether the currently set sweep parameters (sweep mode, sweep start, \
-                                          stop, width, etc.) are consistent. If there is a \
-                                          sweep configuration problem, the laser source is not \
-                                          able to pass a wavelength sweep.""")
+    _sweep_params_consistent = Instrument.measurement(
+        'sour0:wav:swe:chec?',
+        """Returns whether the currently set sweep parameters (sweep mode, sweep start,
+        stop, width, etc.) are consistent. If there is a
+        sweep configuration problem, the laser source is not
+        able to pass a wavelength sweep.""")
 
     sweep_points = Instrument.measurement('sour0:read:points? llog',
-                                          """Returns the number of datapoints that the :READout:DATA? \
+                                          """Returns the number of datapoints that the :READout:DATA?
                                           command will return.""")
 
     sweep_state = Instrument.control('sour0:wav:swe?', 'sour0:wav:swe %g',
-                                     """ State of the wavelength sweep. Stops, starts, pauses \
-                                     or continues a wavelength sweep. Possible state values are \
+                                     """ State of the wavelength sweep. Stops, starts, pauses
+                                     or continues a wavelength sweep. Possible state values are
                                      0 (not running),
                                      1 (running) and
                                      2 (paused).
-                                     Refer to the N7776C user manual for exact usage of the \
+                                     Refer to the N7776C user manual for exact usage of the
                                      paused option. """,
                                      validator=strict_discrete_set,
                                      values=[0, 1, 2])
 
     wl_logging = Instrument.control('SOUR0:WAV:SWE:LLOG?', 'SOUR0:WAV:SWE:LLOG %g',
-                                    """ State (on/off) of the lambda logging feature of the \
+                                    """ State (on/off) of the lambda logging feature of the
                                     laser source.""",
                                     validator=strict_discrete_set,
                                     map_values=True,
                                     values={True: 1, False: 0})
 
     def check_sweep_params(self):
-        response = self._sweep_check
+        response = self._sweep_params_consistent
         if response[0] == 0.0:
             return True
         elif response[0] == 368:
--- a/pymeasure/instruments/keysight/keysightN7776C.py
+++ b/pymeasure/instruments/keysight/keysightN7776C.py
@@ -57,48 +57,48 @@ class KeysightN7776C(Instrument):
         super(KeysightN7776C, self).__init__(
             address, "N7776C Tunable Laser Source",**kwargs)
 
-    def set_power(self,value,unit='mW'):
-        """
-        Function wrapper to set the power in an arbitrary unit
-        """
-        if not unit in ['dBm','mW']:
-            raise ValueError('Unknown Power Unit.')
+    # def set_power(self,value,unit='mW'):
+    #     """
+    #     Function wrapper to set the power in an arbitrary unit
+    #     """
+    #     if not unit in ['dBm','mW']:
+    #         raise ValueError('Unknown Power Unit.')
 
-        self._output_power_unit = unit
-        if unit == 'mW':
-            self._output_power = value*1e-3 #Conversion to mW
-        else:
-            self._output_power = value
+    #     self._output_power_unit = unit
+    #     if unit == 'mW':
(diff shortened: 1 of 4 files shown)


pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+    def set_power(self,value,unit='mW'):
+        """
+        Function wrapper to set the power in an arbitrary unit
+        """
+        if not unit in ['dBm','mW']:
+            raise ValueError('Unknown Power Unit.')
+
+        self._output_power_unit = unit
+        if unit == 'mW':
+            self._output_power = value*1e-3 #Conversion to mW
```
bilderbuchi, line 71: so the device has a power setting "mW", but then wants the value to be sent in "W"? Alternatively, for a conversion _to mW_ (as your comment indicates) with *1e-3, `value` has to be in µW - I have to admit I'm confused by this.

I would expect with such a signature, that an `instr.set_power(35, 'mW')` sets a power of 0.035 W (i.e. 35 mW) - in that context I don't know why you divide by 1000 again, or what unit the device expects the power in (if not "mW" as its `_output_power_unit` is being set).

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+        if power_unit == 'mW':
+            if unit_output:
+                return (power_reading*1e3,power_unit)
+            else:
+                return power_reading*1e3
+        else:
+            if unit_output:
+                return (power_reading,power_unit)
+            else:
+                return power_reading
```
bilderbuchi, line 90: I'm not sure, this unit handling feels a bit clunky, do you really want to change the return value type depending on the unit_output? I'd prefer to wait for #248 and handle this properly, then.

bilderbuchi: Also, same conversion confusion as above. Maybe I haven't had enough coffee?

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+        
+    def locked(self):
+        return self._locked
+
+    _locked = Instrument.control(':LOCK?',':LOCK %g,'+str(LOCK_PW),
+                                   """ Boolean property controlling the lock state (True/False) of the laser source""",
+                                   validator=strict_discrete_set,
+                                   values=[True,False],
+                                   set_process=lambda v: int(v),
+                                   get_process=lambda v: bool(v))
```
bilderbuchi, line 105: A more pymeasure-idiomatic way for these 4 methods would be to just have a `locked = Instrument.control...` as you already have; instead of `instr.lock()` you use `instr.locked = True`.

Also, a smoother way to implement the booleanness is
```
        map_values=True,
        values={True: 1, False: 0},
```
(I have not commented again if that appears elsewhere)

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+    
+    output = Instrument.control('SOUR0:POW:STAT?','SOUR0:POW:STAT %g',
```
bilderbuchi, line 107: A clearer name could be `output_enabled`.

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+                                    """ Boolean Property that controls the state (on/off) of the laser source """,
+                                    validator=strict_discrete_set,
+                                    values=[True,False],
+                                    set_process=lambda v: int(v),
+                                    get_process=lambda v: bool(v))
+
+    _output_power = Instrument.control('SOUR0:POW?','SOUR0:POW %f',
```
bilderbuchi, line 114: Why indicate this as a private property (leading underscore)?

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+                                    """ Floating point value indicating the laser output power in the unit set by the user (see _output_power_unit).""")
+    
+    _output_power_unit = Instrument.control('SOUR0:POW:UNIT?','SOUR0:POW:UNIT %g',
```
bilderbuchi, line 117: Why indicate this as a private property (leading underscore)?

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+                                    values=WL_RANGE,
+                                    get_process=lambda v: v*1e9)
+    sweep_wl_stop = Instrument.control('sour0:wav:swe:stop?','sour0:wav:swe:stop %fnm',
+                                    """ End Wavelength (in nanometers) for a sweep.""",
+                                    validator=strict_range,
+                                    values=WL_RANGE,
+                                    get_process=lambda v: v*1e9)
+
+    sweep_step = Instrument.control('sour0:wav:swe:step?','sour0:wav:swe:step %fnm',
+                                    """ Step width of th[e sweep (in nanometers).""",
```
bilderbuchi, line 149:
```suggestion
                                    """ Step width of the sweep (in nanometers).""",
```

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+    sweep_mode = Instrument.control('sour0:wav:swe:mode?','sour0:wav:swe:mode %s',
+                                    """ Sweep mode of the swept laser source """,
+                                    validator=strict_discrete_set,
+                                    values=['STEP','MAN','CONT'])
+    sweep_twoway = Instrument.control('sour0:wav:swe:rep?','sour0:wav:swe:rep %s',
+                                    """Sets the repeat mode. Applies in stepped,continuous and manual sweep mode.""",
+                                    validator=strict_discrete_set,
+                                    values=[True,False],
+                                    set_process=lambda v: ['ONEW','TWOW'][int(v)],
+                                    get_process=lambda v: bool(['ONEW','TWOW'].index(v)))
```
bilderbuchi, line 166: Similar to above, I think
```
        map_values=True,
        values={True: 'ONEW', False: 'TWOW'},
```
should do the trick...

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+    _sweep_check = Instrument.measurement('sour0:wav:swe:chec?',
```
bilderbuchi, line 167: More expressive name `sweep_params_status`? (also why private, this seems useful for a user?)

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+                                    """Returns whether the currently set sweep parameters (sweep mode, sweep start, stop, width, etc.) are consistent. If there is a
+                                    sweep configuration problem, the laser source is not able to pass a wavelength sweep.""")
+
+    sweep_points = Instrument.measurement('sour0:read:points? llog',
+                                    """Returns the number of datapoints that the :READout:DATA? command will return.""")
+    
+    sweep = Instrument.control('sour0:wav:swe?','sour0:wav:swe %g',
```
bilderbuchi, line 174: Clearer: `sweep_state`? `sweep` could as well be a verb/command.

bilderbuchi: Also, how is this used?

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+                                    validator=strict_discrete_set,
+                                    values=[True,False],
+                                    set_process=lambda v: int(v),
+                                    get_process=lambda v: bool(v))
+
+    def check_sweep_params(self):
+        response = self._sweep_check
+        if response[0] == 0.0:
+            return True
+        elif response[0] == 368:
```
bilderbuchi, line 189: Maybe?
```suggestion
    def check_sweep_params_ok(self):
        response = self._sweep_check[0]
        if response == 0.0:
            return True
        elif response == 368:
        etc. etc.
```
However, does `self._sweep_check` return a list, not an int?

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+            log.log(30,'Warning: End Wavelength <= Start Wavelength.')
```
bilderbuchi, line 190: please instead
```suggestion
            log.warning('End Wavelength <= Start Wavelength.')
```
etc.

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+            return False
+        else:
+            log.log(30,'Warning: Unknown Error!')
+            return False
+            
+
+
+    def next_step(self):
+        """
+        Performs the next sweep step in stepped sweep if its paused or in manual mode.
```
bilderbuchi, line 236:
```suggestion
        Performs the next sweep step in stepped sweep if it is paused or in manual mode.
```

bilderbuchi (changes requested):
Thank you for your contribution! I tried to give feedback how to better align this with usage of other instruments, and some questions/change requests.

pymeasure/instruments/keysight/keysightN7776C.py
maederan201: Changed to sweep_state.

It is basically a state indicator giving either 0 (not running), 1 (running) or 2 (paused). Mostly it will be used for checking if a continuous sweep is finished, but other use cases are possible

maederan201 (commented):
Thanks for your comments!

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+import numpy as np
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+
+WL_RANGE = [1480,1620]
+LOCK_PW = 1234
+
+class KeysightN7776C(Instrument):
+    """
+    This represents the Keysight N7776C Tunable Laser Source interface.
```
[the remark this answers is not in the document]
bilderbuchi: hm, no, still occurs. I'll see if I can fix it.

pymeasure/instruments/keysight/keysightN7776C.py
bilderbuchi: As you have noticed, passing multiple objects to/from properties starts to feel strange quickly. 
Maybe while you wait for #248, for now instead of the current get_power/set_power/_output_power define 2 properties `output_power_dBm` and `output_power_mW`, and have the user use floats when getting/setting with those -- the property name makes the unit clear. 
You can look at other instruments to check how to manually implement properties with validators etc. Our property factory will not be suitable as you have the additional complication of the unit setting/getting. If you don't get/set power at a high rate, or your instrument is fast, you can get/set the unit every time. If not, you might cache it in an attribute (e.g. `self._current_unit` to avoid at least some redundant communication.

When #248 is available (could take time, though), you can add a generic unitful `output_power`. Btw please make a short note in that issue that this part of your instrument would be a candidate for test-implementing a unitful property.

pymeasure/instruments/keysight/keysightN7776C.py
bilderbuchi: `valid_sweep_params()` sounds great!

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,256 @@
...
+    .. code-block:: python
+        laser = N7776C( address )
+        laser.sweep_wl_start = 1550
+        laser.sweep_wl_stop = 1560
+        laser.sweep_speed = 1
+        laser.sweep_mode = 'CONT'
+        laser.output = 1
+        while laser.sweep == 1:
+            continue
```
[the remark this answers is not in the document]
bilderbuchi: I see. Well, there is the SCPI-compatible `Instrument.complete()`, but I don't know if your instrument implements that.

pymeasure/instruments/keysight/keysightN7776C.py
bilderbuchi: Thanks, best document those values in the docstring, and use a `strict_discrete_set` validator to make sure that only valid values get through. Trinary logic is tricky (can't easily be converted to true/false) - I'd suggest either staying with the integers, or map them to strings "stopped"/"running"/"paused". Consider how a user determines which is which.

pymeasure/instruments/keysight/keysightN7776C.py
maederan201: So I double checked this, it is indeed giving back a list, e.g.
´´´[368.0, 'LambdaStop<=LambdaStart']´´´
So this should be resolved with the newest version

pymeasure/instruments/keysight/keysightN7776C.py
maederan201: I finally got around to testing this, generally I like they way of having two properties ´´´output_power_dBm´´´ and ´´´output_power_mW´´´. However there was a flaw in the logic: For setting all is fine, but getting is a problem as the value given back is always in the unit of ´´´_output_power_unit´´´. 

Specifically the following script does pose problems:
´´´
l._output_power_unit #Response: 'mW'
l.output_power_dBm = 7.2
l.output_power_dBm #Resonse: 0.00524807442
l._output_power_unit = 'dBm'
l.output_power_dBm #Resonse: 7.2
´´´

I tried to fix it by writing a function which comapres the desired unit to what the current unit is and converts if neccessary (which I tried to hide) and then passing this for the get_process, but apparently I have no access to the base class in the get_process function. I need to look deeper into this, but if you have any suggestions of how to make this work, let me know. 

For now, I commited this version with two options under 87a8437. If you have a suggestion or a hint if this could be a way, let me know. I will think about it in more detail too

maederan201: Found out how to fix it with decorators!

maederan201:
All requested changes should be implemented. PR should be ready for merging.

pymeasure/instruments/keysight/keysightN7776C.py
bilderbuchi: Resolved? The code has not changed afaict, or am I missing something?

Also it might make sense to convert `response[0]` to int, as it's not guaranteed that 234234.0 == 234234

pymeasure/instruments/keysight/keysightN7776C.py
bilderbuchi: Maybe rename to `_sweep_params_consistent`?

pymeasure/instruments/keysight/keysightN7776C.py
```diff
@@ -0,0 +1,239 @@
...
+            log.info('Sweep in progress.')
+        laser.output_enabled = 0
+
+    """
+    def __init__(self, address, **kwargs):
+        super(KeysightN7776C, self).__init__(
+            address, "N7776C Tunable Laser Source", **kwargs)
+
+    locked = Instrument.control(':LOCK?', ':LOCK %g,'+str(LOCK_PW),
+                                """ Boolean property controlling the lock state (True/False) of \
```
bilderbuchi, line 59: What's up with the line continuation `\` here? These are multiline strings, I'd suggest to leave the natural linebreaks in place. Also below. Also, you should not need to add a `\n` at the end of lines, that feels weird. 

Just write the docstrings with natural linebreaks, and if you absolutely want it to fit on one line without hitting the line length limit, you can use an indentation pattern of
```
    output_enabled = Instrument.control(
        'SOUR0:POW:STAT?', 
        'SOUR0:POW:STAT %g',
        """ Boolean Property that controls the state (on/off) of the laser source """,
        map_values=True,
        values={True: 1, False: 0},
    )
```

pymeasure/instruments/keysight/keysightN7776C.py
maederan201: Renamed as suggested, I chose to make it private because I'd want a user to use `check_sweep_params` function as this handles warnings and errors

pymeasure/instruments/keysight/keysightN7776C.py
maederan201: You are right, now it is really renamed. I also included the int conversion and got rid of the repeated `response[0]` and redefined the response to be the first element of the list

pymeasure/instruments/keysight/keysightN7776C.py
maederan201: I changed the docstrings, at one point a `\` seems to be unavoidable, or I'm to inexperienced with the linting and linebreaks. Feel free to change it if you know a better way.

```
        elif response == 374:
            log.warning('The only allowed modulation source with lambda logging \
                        function is coherence control.')
```

bilderbuchi:
LGTM, thank you for your contribution!

Issue #478, opened 2021-09-12
Run experiment in console/batch mode


msmttchr:
I think it can be extremely useful to be able to write script that can run both in GUI and console batch mode.
The example provided in pymeasure shows that the same Procedure class can be run both in GUI mode and in console mode, but in console mode all the parameters have to be provided inside the script.

My idea would be to create a sort of `ConsoleScript` class which is the console version of `ManagedWindow`. `ConsoleScript` will take care of creating an `argparse` initialization to parse the users's input parameters via command line.

Does anyone thought about this ? Anyone who started such pull request ? Any other comment ?

msmttchr:
Indeed, my idea was to make the parameters available as command line options to be able to configure them. Obviously the graphic part will not be available, but log messages will be available in the console and the same results files (.csv) will be generated as the regular experiment in GUI mode.

CasperSchippers:
Sounds like it could be useful indeed. One thing to keep in mind is to ensure that interrupting the measurement (e.g Keyboard Interrupt) works as an abort, and hence also runs the shutdown method of the procedure.

LongnoseRob:
I like the idea.:+1:
These are two points which I would like o add to the discussion:
1. instead/in parallel to offering the parameters as command line options also offer the possibility to use a "config"-file[^1]
2. use something like [progressbar](https://progressbar-2.readthedocs.io/en/latest/) to visualize the progress (if neccessary/possible/wanted)


[^1]: some easy flavor

msmttchr:
@LongnoseRob, if you have time, could You please give a try to PR #500 ? It is still draft, but pretty much usable.
I would be interested in your preliminary feedback

CasperSchippers:
It looks very promising indeed; I tried some things (not yet all) and most works quite well and looks promising

> Have you tried some of the other command line options (e.g. to set parameter from log file)?

There is a small problem when using the --use-log-file option, see my remark in the PR.

I will continue testing and try to review asap

msmttchr:
Completed as PR #500 is merged. Thanks to all contributors and reviewers, especially @CasperSchippers, @LongnoseRob, @mcdo0486

PR #500, opened 2021-10-20
Add console mode

msmttchr:
Address Issue #478


pymeasure/display/console.py
```diff
@@ -0,0 +1,268 @@
...
+            raise NotImplementedError("Sequencer not yet implemented")
+
+        bar_enabled = not args['no_progressbar']
+
+        # Set procedure parameters
+        parameter_values = {}
+
+        if args['use_log_file'] != None:
+            # Special case set parameters from log file
+            results = Results(args['use_log_file'])
```
CasperSchippers, line 232: At this moment, this raises an error; it works when you change this line to `results = Results.load(args['use_log_file'])`

msmttchr: Thanks well spotted, this problem was introduced during a code clean up. Possibly, I need to extend the test suite to cover also this.

bilderbuchi:
Github limits the number of annotations in the diff view (to 10?), so if you have a lot of them you'll need a number of commits until they are all gone -- better to run `flake8 .` locally.

msmttchr:
> Github limits the number of annotations in the diff view (to 10?), so if you have a lot of them you'll need a number of commits until they are all gone -- better to run `flake8 .` locally.

Good tip! Thanks 👍

examples/Basic/console.py
```diff
@@ -0,0 +1,98 @@
...
+        experiment = self.new_experiment(results)
+
+        self.manager.queue(experiment)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) > 1:
+        app = Console(sys.argv)
+    else:
+        app = QtGui.QApplication(sys.argv)
```
CasperSchippers, line 94: `QtGui.QApplication` is being deprecated, I think it should be replaced by:
```suggestion
        app = QtWidgets.QApplication(sys.argv)
```

examples/Basic/console.py
```diff
@@ -0,0 +1,98 @@
...
+        window = MainWindow()
+        window.show()
```
CasperSchippers, line 96: I like this construction!!! One thing I started wondering (but maybe that'd make it more/too complex) is if this could not be integrated into the MainWindow by default.

pymeasure/display/console.py
```diff
@@ -0,0 +1,276 @@
...
+    def abort(self):
+        """ Aborts the currently running Experiment, but raises an exception if
+        there is no running experiment
+        """
+        self._worker.update_status(Procedure.ABORTED)
+        self._worker.stop()
+
+    def exec(self):
+        # Parse command line arguments
+        args = vars(self.parser.parse_args(self.args[1:]))
```
CasperSchippers, line 216: I think the `self.args[1:]` is not required; it should normally automatically handle the sys.argv correctly

pymeasure/display/console.py
```diff
@@ -0,0 +1,276 @@
...
+            for name in results.parameters:
+                parameter_values[name] = results.parameters[name].value
+        else:
+            for name in args:
+                opt_name = name.replace("_", "-")
+                if not (opt_name in self.parser.special_options):
+                    parameter_values[name] = args[name]
+
+        procedure.set_parameters(parameter_values)
+        if (bar_enabled):
```
CasperSchippers, line 250: since you use `bar_enabled` only here, there seems no need to define the variable at all and just directly access `not args['no_progressbar']` here

CasperSchippers (commented):
Turns out I'm a bit of a moron: I thought I submitted some review already a while ago and didn't come back and check better: turns out you do really have to push the 'submit' button.

Anyways, went over it again and here are some comments.

examples/Basic/console.py
msmttchr: Here I followed the style used in the basic GUI demo. Are you hinting to change this demo structrure ?
I think this can be done, but probably in a dedicated PR ? Do you agree ?

examples/Basic/console.py
```diff
@@ -0,0 +1,98 @@
...
+
+    sys.exit(app.exec_())
```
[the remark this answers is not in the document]
msmttchr: Indeed. Done.

pymeasure/display/console.py
```diff
@@ -0,0 +1,276 @@
...
+        special_options = copy.deepcopy(self.special_options)
+        for option, kwargs in special_options.items():
+            help_fields = [('units are', 'units')] + kwargs['help_fields']
+            desc = kwargs['desc']
+            kwargs['help'] = self._cli_help_fields(desc, kwargs, help_fields)
+            del kwargs['help_fields']
+            del kwargs['desc']
+            self.add_argument("--" + option, **kwargs)
+
+    def _cli_help_fields(self, name, inst, help_fields):
```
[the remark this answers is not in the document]
msmttchr: Thanks. This is absolutely correct.

requirements/pymeasure.yml
```diff
@@ -23,3 +23,5 @@ dependencies:
   - pip  # don't pin, to gain newest conda compatibility fixes
   - pip:
     - pyvisa-sim==0.4.0
+    - progressbar2==3.53.3
```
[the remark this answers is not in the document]
msmttchr: Really no reason, except I do not know conda very well.
Since I made the module optional, I will remove it.

pymeasure/display/console.py
```diff
@@ -0,0 +1,276 @@
...
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class ConsoleArgumentParser(argparse.ArgumentParser):
+    special_options = {
+        "no-progressbar":  {"default": False,
+                            "desc": "Disable progressbar",
+                            "help_fields": ["default"],
+                            "action": 'store_true'},
```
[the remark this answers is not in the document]
msmttchr: @CasperSchippers, I think the progressbar could make sometime the logging messages not easily readable, so users may want to disable it in these cases.
I used a boolean command line option, but this I think allow you only to set to True or False and it does not accept parameters, that is why I used the no-progressbar, that is set to False by default, but set to True if user pass the command line option.

Another option would be to use something similar to this:
`parser.add_argument('--progress-bar', action='store', default=1, type=int)
`
In this case, you have progressbar enabled by default and using `--progressbar=0` should disable it.
I would think that this mode has the advantage to be more direct. but it is a little more difficult to use.

What is your opinion about this solution ? or did you have something different in mind ?

pymeasure/display/console.py
```diff
@@ -0,0 +1,276 @@
...
+            kwargs = {}
+            parameter = parameter_objects[name]
+            default, help_fields, _type = parameter.cli_args
+            kwargs['help'] = self._cli_help_fields(parameter.name, parameter, help_fields)
+            kwargs['default'] = default
+            if _type is not None:
+                kwargs['type'] = _type
+            self.add_argument("--"+name, **kwargs)
+
+        special_options = copy.deepcopy(self.special_options)
```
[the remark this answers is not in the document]
msmttchr: Done, thanks for the suggestion.

pymeasure/display/console.py
```diff
@@ -0,0 +1,267 @@
...
+    :param inputs: list of :class:`~pymeasure.experiment.parameters.Parameter` instance variable names, which the display will generate graphical fields for
+    :param log_channel: :code:`logging.Logger` instance to use for logging output
+    :param log_level: logging level
+    :param sequence_file: simple text file to quickly load a pre-defined sequence with the :code:`Load sequence` button
+    :param directory_input: specify, if present, where the experiment's result will be saved.
+    """
+    def __init__(self,
+                 args,
+                 procedure_class,
+                 inputs=(),
```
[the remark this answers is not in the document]
msmttchr: @CasperSchippers, I originally assumed that `input` parameter was to select parameters that user can change, but now I realize you may be right. This may be used to avoid to limit space in the UI, so I would vote to remove the parameter.

pymeasure/display/console.py
```diff
@@ -0,0 +1,276 @@
...
+        self.directory_input = directory_input
+        self.log = logging.getLogger(log_channel)
+        self.log_level = log_level
+        log.setLevel(log_level)
+        self.log.setLevel(log_level)
+
+        # Check if the get_estimates function is reimplemented
+        self.use_estimator = not self.procedure_class.get_estimates == Procedure.get_estimates
+        self.parser = ConsoleArgumentParser(procedure_class)
+        self.parser.setup_parser(self.inputs)
```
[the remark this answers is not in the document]
msmttchr: Done. Thanks.

pymeasure/display/console.py
```diff
@@ -0,0 +1,276 @@
...
+                            "help_fields": ["default"]},
+        "log-directory":   {"default": ".",
+                            "desc": "Log directory",
+                            "help_fields": ["default"]},
+        "log-file":        {"default": None,
+                            "desc": "Log filename (string or callable which return a string)",
+                            "help_fields": ["default"]},
+        "use-log-file":    {"default": None,
+                            "desc": "File to retrieve params from",
+                            "help_fields": ["default"]},
```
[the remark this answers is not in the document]
msmttchr: Indeed you are right, rather confusing. I will change it.

pymeasure/display/console.py
msmttchr: Ok. I will make the change

pymeasure/display/console.py
```diff
@@ -0,0 +1,276 @@
...
+        procedure = self.procedure_class()
+
+        self.directory = args['log_directory']
+        self.filename = args['log_file']
+        try:
+            log_level = int(args['log_level'])
+        except ValueError:
+            # Ignore and assume it is a valid level string
+            log_level = args['log_level']
+        self.log_level = log_level
```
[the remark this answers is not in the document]
msmttchr: Ok. Done. Thanks

pymeasure/display/console.py
msmttchr: OK, done.

msmttchr:
> Turns out I'm a bit of a moron: I thought I submitted some review already a while ago and didn't look better: turns out you do really have to push the 'submit' button.
> 
> Anyways, went over it again and here are some comments.
> 
> My sincere apologies for the (enormously) late review.

No worries at all. We are all very busy and for sure I am not lighting fast 😁.

pymeasure/display/console.py
```diff
@@ -0,0 +1,281 @@
...
+# THE SOFTWARE.
+#
+
+import logging
+
+import os
+import copy
+import argparse
+try:
+    import progressbar
```
bilderbuchi, line 31: This snagged my attention in passing: I haven't checked what that package does specifically, but in the last years, whenever the matter was including "progress bars" in Python, `tqdm` came up first and foremost (in articles, blogs, etc): https://tqdm.github.io/ What was the reason you chose `progressbar`?
(To be fully clear I have used neither up to now)

pymeasure/display/console.py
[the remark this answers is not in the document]
CasperSchippers: Ah no that is a good point; I didn't realize that for an "enable-progressbar" option, you'd have to add an argument to the option, which is not as nice as just being able to add the option to disable it. Let's just leave it as it is.

pymeasure/display/console.py
msmttchr: Done

pymeasure/display/console.py
```diff
@@ -0,0 +1,276 @@
...
+        # Handle Ctrl+C nicely
+        signal.signal(signal.SIGINT, lambda sig, _: self.abort())
+
+    def get_filename(self, directory):
+        """ Return filename for logging.
+
+        User can override this method to define their own filename
+        """
+        if self.filename is not None:
+            return os.path.join(directory, self.filename)
```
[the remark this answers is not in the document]
CasperSchippers: Hmmm, that also makes sense; on the other hand, I could also expect someone to run it with passing just a base-filename and expecting it to append a number (e.g. for repeating measurements you can then just in the console hit the up button and enter to start a new measurement)

pymeasure/display/console.py
msmttchr: > This snagged my attention in passing: I haven't checked what that package does specifically, but in the last years, whenever the matter was including "progress bars" in Python, `tqdm` came up first and foremost (in articles, blogs, etc): https://tqdm.github.io/ What was the reason you chose `progressbar`? (To be fully clear I have used neither up to now)

@bilderbuchi, thanks for having a look. Indeed, I did some google search and using progressbar2 I had good visual results. If there is no strong reason, I would leave this optional package instead of moving to tqdm.

bilderbuchi: Fair enough, ok with me.

pymeasure/display/console.py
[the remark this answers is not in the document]
msmttchr: I understand your concern, I think I have fixed it with 6e8e9ae384b73917c7ff51e453b6e07af76e88b7

CasperSchippers (commented):
Here's some remarks from me; I think it's almost finished, good work!

Maybe it is also good to add a remark about this feature and reference to the example in the documentation.

pymeasure/display/browser.py
```diff
@@ -34,7 +34,24 @@
 log.addHandler(logging.NullHandler())
 
 
-class BrowserItem(QtWidgets.QTreeWidgetItem):
+class BaseBrowserItem(object):
```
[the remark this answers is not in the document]
msmttchr: Indeed

pymeasure/experiment/parameters.py
```diff
@@ -264,10 +289,13 @@ class VectorParameter(Parameter):
     :param ui_class: A Qt class to use for the UI of this parameter
     """
 
+    _type = eval
```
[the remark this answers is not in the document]
msmttchr: I think you are totally right in raising this point on eval. Actually, I believe that this further inspired some changes to rationalize the code, could you please have a look at 34ac7ca8948c40cea30be20a0e6044c6b95f90cd at your convenience?

pymeasure/experiment/parameters.py
```diff
@@ -46,9 +46,12 @@ class Parameter:
 
     def __init__(self, name, default=None, ui_class=None, group_by=None, group_condition=True):
         self.name = name
-        self._value = default
+        self._value = None
+        if default is not None:
+            self._value = self.convert(default)
```
CasperSchippers, line 51: You can consider using the setter here; that way you ensure setting the default value gets the same treatment as does normal value setting (for now not so much difference, but you never know):
```suggestion
            self.value = default
```

pymeasure/experiment/parameters.py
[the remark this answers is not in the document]
CasperSchippers: That looks like a good improvement, I like the centralization (inheritance) of the value property

pymeasure/display/console.py
```diff
@@ -0,0 +1,272 @@
...
+            results = Results.load(args['use_result_file'])
+            for name in results.parameters:
+                self.parameter_values[name] = results.parameters[name].value
+        else:
+            for name in args:
+                opt_name = name.replace("_", "-")
+                if not (opt_name in parser.special_options):
+                    self.parameter_values[name] = args[name]
+
+        if (progressbar and not args['no_progressbar']):
```
CasperSchippers, line 217: Came accross some reduntant parentheses

examples/Basic/console.py
CasperSchippers: Well, I was even thinking of adding it into the class itself or something; but you are definately right, if at all, this would be something for another PR;

CasperSchippers:
Seems almost finished, doesn't it?

pymeasure/display/console.py
msmttchr: Ok. Fixed

msmttchr:
> Seems almost finished, doesn't it?

Indeed! Only documentation needs some update.
I hope to find some time soon to complete it.

msmttchr:
> > Seems almost finished, doesn't it?
> 
> Indeed! Only documentation needs some update. I hope to find some time soon to complete it.

@CasperSchippers, I finally added some section in the tutorial. Could you please have a final look and then I believe we can merge?

pymeasure/display/console.py
```diff
@@ -0,0 +1,272 @@
...
+                             "action": 'store_true'},
+        "log-level":        {"default": 'INFO',
+                             "choices": list(logging._nameToLevel.keys()),
+                             "desc": "Set log level (logging module values)",
+                             "help_fields": ["default"]},
+        "sequence-file":    {"default": None,
+                             "desc": "Sequencer file",
+                             "help_fields": ["default"]},
+        "result-directory": {"default": ".",
+                             "desc": "directory where experiment's result are saved",
```
CasperSchippers, line 77: Something I just noted when looking at the help

``` suggestion
                             "desc": "Directory where experiment's result are saved",
```

pymeasure/display/console.py
```diff
@@ -0,0 +1,272 @@
...
+        "no-progressbar":   {"default": False,
+                             "desc": "Disable progressbar",
+                             "help_fields": ["default"],
+                             "action": 'store_true'},
+        "log-level":        {"default": 'INFO',
+                             "choices": list(logging._nameToLevel.keys()),
+                             "desc": "Set log level (logging module values)",
+                             "help_fields": ["default"]},
+        "sequence-file":    {"default": None,
+                             "desc": "Sequencer file",
```
CasperSchippers, line 74: As you're referring to the help string of the console mode (which is nice, since it helps keeping the tutorial concise), I'd suggest slightly expanding this string.

```suggestion
                             "desc": "Sequencer file as used/defined by the sequencer widget to "
                                     "execute a sequence of measurements",
```

pymeasure/display/console.py
```diff
@@ -0,0 +1,272 @@
...
+                             "help_fields": ["default"]},
+        "result-directory": {"default": ".",
+                             "desc": "directory where experiment's result are saved",
+                             "help_fields": ["default"]},
+        "result-file":      {"default": None,
+                             "desc": "File name where results are stored",
```
CasperSchippers, line 80: As you're referring to the help string of the console mode (which is nice, since it helps keeping the tutorial concise), I'd suggest slightly expanding this string.

```suggestion
                             "desc": "File name where results are stored; this string is handled "
                                     "by the `unique_filename` function and hence allows for "
                                     "filling in parameter values and is suffixed by the date "
                                     "(`YYYY-MM-DD`) and an index number",
```

CasperSchippers (approved):
Nice job!!!

I just have a few suggestions for the help strings, but that is just some really minor nitpicking.

I think it is finished; feel free to merge!!

PR #503, opened 2021-10-28
Store conditions in file

CasperSchippers:
PR to resolve #501 

Adds a Parameter-like object (Metadata). After the `startup` method of the Procedure has ran, the Metadata objects are evaluated, thereby retrieving and fixing their value. Subsequently they are stored stored to an optional section of the datafile.

The conditions can be filled in various manners:
- By defining an fget method that will be evaluated after the startup phase
- The fget method can also be defined as a string; in that case it will assume the string is the name of a method of the Procedure that is to be called
- By manually setting the value in the startup phase; just like a Parameter, Metadata objects are replaced by their value in the init of the Procedure; this value can be set (just like an ordinary attribute) and after the startup the values of these ordinary attributes are collected and will replace (if set) the value of the Metadata
- If none of the above, and if a default value is set, it will return to the default value


bilderbuchi:
Judging from
> allow storing information (e.g. the actual starting time, instrument parameters) about the measurement

I think "Metadata" -- data _about_ something -- would be a good fit.

bleykauf:
I think Metadata captures both user controlled things (instrument settings) and environmental conditions (temperature, measurement time).

CasperSchippers:
Thanks @bilderbuchi for the suggestion; indeed 'metadata' to me sounds like the best suggestion.

docs/tutorial/procedure.rst
```diff
@@ -249,6 +249,52 @@ Let's extend our SimpleProcedure with logging. ::
...
+                sleep(0.01)
+                if self.should_stop():
+                    break
+
+
+As with a Parameter, PyMeasure swaps out the Metadata with their values behind the scene, which makes accessing the values of Metadata very convenient.
+
+The value of a Metadata can be set either using an :python:`fget` method or manually in the startup method.
+The :python:`fget` method, if provided, is ran after startup method.
+It can also be provided as a string; in that case it is assumed that the string contains the name of a method of the Procedure class which returns the value that is to be stored.
```
bilderbuchi, line 294: Why this duplicated way of defining the callback? Can't you just say `Metadata('bla', fget=self.mymethod)` instead of `Metadata('bla', fget='mymethod')`? 
The only limitation, I think, is that the Metadata class attribute has to be defined _after_ that method, but that kinda makes sense anyway. 
Additionally, why have the Metadata as class-level attributes? Would it not be simpler to just define instance attributes in the __init__? That would also avoid the above limitation.

CasperSchippers: The reason is exactly as you mention, the fact that you can only define it after defining the the method. Why I chose for this (and also to make it a class attribute) is to make it similar to the Parameter and Measurable objects, such that you can define this at the beginning of the Parameters class: everything that is stored in the file can be nicely defined at a single place, making it (in my opinion) neater/clearer.

msmttchr:
> Thanks @bilderbuchi for the suggestion; indeed 'metadata' to me sounds like the best suggestion.
> 
> As a matter of fact, I also still want to add one additional way of getting the metadata, by grabbing it from an instrument property.

Nice and useful addition. Thanks.
I have one question: if I would like to store the instrument model in the metadata, how would you do this ?

CasperSchippers:
> if I would like to store the instrument model in the metadata, how would you do this ?

At this moment, you would define a Metadata object without an f-get method (maybe a default value if you want) and, in the startup method, get the instrument model and assign it   to that Metadata object.
I am thinking of making an additional option to, when defining the Metadata object, pass an argument stating which property from which instrument to retrieve and store. Not yet sure though what the best way to do this is.

bilderbuchi:
You could use the fget arg for that -- if it's a method object, call it. If it's a string, try to call the respective method, in an except try to `getattr` (captures both properties and attributes), if that fails raise an exception?

CasperSchippers:
That's a good suggestion, thanks

pymeasure/experiment/procedure.py
```diff
@@ -180,6 +182,38 @@ def set_parameters(self, parameters, except_missing=True):
                     raise NameError("Parameter '{}' does not belong to '{}'".format(
                         name, repr(self)))
 
+    def _update_metadata(self):
+        """ Collects all the Metadata objects for the procedure and stores
+        them in a meta dictionary so that the actual values can be set and used
+        in their stead
+        """
+        if not self._metadata:
+            self._metadata = {}
```
msmttchr, line 191: I am not sure why this check is needed.

pymeasure/experiment/parameters.py
```diff
@@ -508,3 +508,83 @@ def value(self):
...
+    If no fget function is specified, the value property will return the
+    latest set value of the parameter (or default if never set).
+
+    :var value: The value of the parameter. This returns (if a value is set)
+        the value obtained from the `fget` (after evaluation) or a manually
+        set value. Returns `None` if no value has been set
+
+    :param name: The parameter name
+    :param fget: The parameter fget function; can be provided as a callable,
+        or as be a string, in which case it is assumed to be the name of a
```
msmttchr, line 525:
```suggestion
        or as a string, in which case it is assumed to be the name of a
```

pymeasure/experiment/procedure.py
```diff
@@ -180,6 +182,38 @@ def set_parameters(self, parameters, except_missing=True):
...
+
```
msmttchr, line 192: I am not sure why this check is needed. Is it to copy class attribute `_metadata` at instance level ?
I think internal method `_update_metadata` is called only once at `__init__`, so why not to replace the code with simply: `self._metadata = {}` avoiding the `if` and removing line 72.

msmttchr (commented):
@CasperSchippers looks good and very useful to me.
Few comments/clarifications and then it is OK to approve for me.

pymeasure/experiment/procedure.py
CasperSchippers: I think that this if statement is indeed not necessary and can be replaced by `self._metadata = {}`. I don't completely understand what you mean by removing line 72 though??

msmttchr: Sorry I meant line 67 where dictionary `_metadata` is declared as class attribute.

CasperSchippers: Hmm ahh yeah that makes sense. Thank!

msmttchr (approved):
Ok for me. Thanks for this useful feature

PR #504, opened 2021-10-28
Fix setting of sensor flags for Thorlabs PM100D

### LINKED ISSUE #498: Thorlabs PM100 doesn't work with sensor head S130C
When using the Thorlabs PM100 with the measurement head S121C everything works as expected.

However, if a measurement head S130C is used instead, I get the following error message upon inialization of the ThorlabsPM100USB class:

```
ValueError                                Traceback (most recent call last)
<ipython-input-15-b33dbcfde9d7> in <module>
----> 1 pm = ThorlabsPM100USB('USB0::0x1313::0x8078::P0032734::INSTR')

~\Miniconda3\envs\lab\lib\site-packages\pymeasure\instruments\thorlabs\thorlabspm100usb.py in __init__(self, adapter, **kwargs)
     49             adapter, "ThorlabsPM100USB powermeter", **kwargs)
     50         self.timout = 3000
---> 51         self.sensor()
     52
     53     def measure_power(self, wavelength):

~\Miniconda3\envs\lab\lib\site-packages\pymeasure\instruments\thorlabs\thorlabspm100usb.py in sensor(self)
     82         # setting the flags; _dn are empty
     83         self.is_power, self.is_energy, _d4, _d8, \
---> 84         self.resp_settable, self.wavelength_settable, self.tau_settable, _d128, self.temperature_sens = self._flags
     85
     86     @property

ValueError: not enough values to unpack (expected 9, got 6)
```

So the problem is in this line: https://github.com/pymeasure/pymeasure/blob/658d8fb9a02bdb62f64cc3838875c0de12f49ca1/pymeasure/instruments/thorlabs/thorlabspm100usb.py#L84

I have to admit that I don't really understand what exactly is happing in this specific code block. The two different measurement heads identify as follows  ("SYST:SENSOR:IDN?"):

```
'S130C,14102240,22-Oct-2014,1,3,33\n'
```

and 

```
'S121C,18030824,08-Mrz-2018,1,18,289\n'
```

respectively.

bleykauf:
Note: Has not yet been tested with an actual instrument, will report after I had the chance and remove the draft status if succesful.

Fixes the setting of the sensor head flags for older(?) sensors where flags were not 9 bit, yet.

Also, fixes a bug where the order of the flags was reversed. Also cleaned up the code in general.

Closes #498.


pymeasure/instruments/thorlabs/thorlabspm100usb.py
```diff
@@ -23,74 +23,97 @@
...
+        """Energy in J."""
+        if self.is_energy:
+            return self.values("MEAS:ENER?")[0]
+        else:
+            raise AttributeError(f"{self.sensor_name} is not an energy sensor.")
+
+    def _set_flags(self):
+        """Get sensor info and write flags."""
+        response = self.ask("SYST:SENSOR:IDN?").split(",")
         self.sensor_name = response[0]
```
msmttchr, line 88: What about using `self.values` ?

pymeasure/instruments/thorlabs/thorlabspm100usb.py
```diff
@@ -23,74 +23,97 @@
...
-        self._flags_str = response[-1][:-1]
-        # interpretation of the flags
-        # rough trick using bin repr, maybe something more elegant exixts
-        # (bitshift, bitarray?)
-        self._flags = tuple(
-            map(lambda x: x == '1', bin(int(self._flags_str))[2:]))
-        # setting the flags; _dn are empty
-        self.is_power, self.is_energy, _d4, _d8, \
-        self.resp_settable, self.wavelength_settable, self.tau_settable, _d128, self.temperature_sens = self._flags
+        _flags_str = response[5].rstrip("\n")
```
msmttchr, line 94: I think the `rstrip` is not needed if you use `self.values` above or you need to preserve trailing spaces?

msmttchr (changes requested):
Few comments from my side.
I noticed also there is a conflict with `authors.txt`. Please remember to update also `authors.rst`

pymeasure/instruments/thorlabs/thorlabspm100usb.py
```diff
@@ -23,74 +23,97 @@
...
+    def wavelength(self):
+        """Wavelength in nm."""
+        value = self.values("SENSE:CORR:WAV?")[0]
+        return value
+
+    @wavelength.setter
+    def wavelength(self, value):
+        """Wavelength in nm."""
+        if self.wavelength_settable:
+            value = truncated_range(value, [self.wavelength_min, self.wavelength_max])
```
bilderbuchi, line 62: watch out here, this means every time you set the wavelength, you have 3 communications with the instrument, as wavelength_min and wavelength_max will be queried anew. If those are constant properties (e.g. instrument or sensor head parameters), it might be better to store the values in, e.g. `self._wavelength_min`, `self._wavelength_max` and fill those values once in the initializer of the class.

pymeasure/instruments/thorlabs/thorlabspm100usb.py
```diff
@@ -23,74 +23,97 @@
...
+
+        # Convert to boolean.
+        self.flags = [x == "1" for x in _flags_str]
 
-    @energy.setter
-    def energy(self, val):
-        raise Exception("Energy not settable!")
+        # setting the flags; _dn are unused; decimal values as comments
+        (
+            self.is_power,  # 1
```
bilderbuchi, line 110: while we're improving the code, maybe `self.measures_power` and `self.measures_energy` are better names?

bilderbuchi (changes requested):
Looks good, thanks for the PR. I have a couple of suggestions.

pymeasure/instruments/thorlabs/thorlabspm100usb.py
bleykauf: I renamed all flags the names from the manual: c2e6e1f

pymeasure/instruments/thorlabs/thorlabspm100usb.py
bleykauf: I added a check to only request the minimum and maximum wavelength once: a76423a

I put it in the wavelength setter and not in the class init, since wavelength might not be settable by some instruments. This way, `AttributeError` is raised in this case without the need for an additional check in `__init__`.

bleykauf:
So apparently the conflict in the `AUTHORS.txt` did not get resolved...

pymeasure/instruments/thorlabs/thorlabspm100usb.py
bleykauf: Agree, I changed it in 27bb492.

Will test soon with an actual instrument to make sure it works as expected.

pymeasure/instruments/thorlabs/thorlabspm100usb.py
bleykauf: Changed it in 27bb492.

pymeasure/instruments/thorlabs/thorlabspm100usb.py
bilderbuchi: Nicely spotted, good point!

bilderbuchi:
Don't worry, I fixed the conflict.

bilderbuchi:
LGTM to me now, thanks for your contribution! After you've tested with a real instrument as you indicated, to confirm that everything works as intended, ping us and we can merge.

msmttchr (approved):
Thanks for your contribution

PR #508, opened 2021-11-03
Add pendulum CNT-91 frequency counter.

bleykauf:
I added some functionality for a new device: the pendulum CNT-91 frequency counter.

There is only a small fraction of commands implemented so far. It has a convenience function for taking time series of frequency data (that's what I use the device for). Validators and value mapping are implemented for the commands.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -41,3 +41,4 @@ Nicola Corna
 Robert Eckelmann
 Sam Condon
 Andreas Maeder
+Bastian Leykauf
--- a/docs/api/instruments/pendulum/index.rst
+++ b/docs/api/instruments/pendulum/index.rst
@@ -1,0 +1,12 @@
+.. module:: pymeasure.instruments.rohdeschwarz
+
+########
+Pendulum
+########
+
+This section contains specific documentation on the Pendulum hwarz instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments>`.
+
+.. toctree::
+   :maxdepth: 2
+
+   cnt91
--- a/pymeasure/instruments/pendulum/cnt91.py
+++ b/pymeasure/instruments/pendulum/cnt91.py
@@ -1,0 +1,191 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2021 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import logging
+from time import sleep
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import (
+    strict_discrete_set,
+    strict_range,
+    truncated_range,
+)
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class CNT91(Instrument):
+    """Represents a Pendulum CNT-91 frequency counter."""
+
+    CHANNELS = {"A": 1, "B": 2, "C": 3, "E": 4, "INTREF": 6}
+    MAX_BUFFER_SIZE = 32000  # User Manual 8-38
+
+    def __init__(self, resourceName, **kwargs):
+        super().__init__(resourceName, "Pendulum CNT-91", **kwargs)
+        self.adapter.connection.timeout = 120000
+
+    def ask(self, command):
+        """
+        Writes the command to the instrument through the adapter
+        and returns the read response.
+
+        :param command: command string to be sent to the instrument
+        """
+        # Infinite loop occures if not redefined.
+        return super().ask(command).rstrip("\n")
+
+    @property
+    def operation_complete(self):
+        """Is True if operation is complete."""
+        return self.ask("*OPC?") == "1"
+
+    @property
+    def batch_size(self):
+        """Maximum number of buffer entries that can be transmitted at once."""
+        if not hasattr(self, "_batch_size"):
+            self._batch_size = int(self.ask("FORM:SMAX?"))
+        return self._batch_size
+
+    def read_buffer(self):
+        """
+        Read out the entire buffer.
+
+        :return : Frequency values from the buffer.
+        """
+        while not self.operation_complete:
+            # Wait until the buffer is filled.
+            sleep(0.01)
+        data = []
+        # Loop until the buffer is completely read out.
+        while True:
+            # Get maximum number of buffer values.
+            new = self.values(":FETC:ARR? MAX")
+            data += new
+            # Last values have been read from buffer.
+            if len(new) < self.batch_size:
+                break
+        return data
+
+    external_start_arming_source = Instrument.control(
+        "ARM:SOUR?",
+        "ARM:SOUR %s",
+        (
+            "Select arming input or switch off the start arming function."
+            "Options are 'A', 'B' and 'E' (rear). 'IMM' turns trigger off."
+        ),
+        validator=strict_discrete_set,
+        values={"A": "EXT1", "B": "EXT2", "E": "EXT4", "IMM": "IMM"},
+        map_values=True,
+    )
+
+    external_arming_start_slope = Instrument.control(
+        "ARM:SLOP?",
+        "ARM:SLOP %s",
+        "Set slope for the start arming condition.",
+        validator=strict_discrete_set,
+        values=["POS", "NEG"],
+    )
+
+    continuous = Instrument.control(
+        "INIT:CONT?",
+        "INIT:CONT %s",
+        "Turn continious measurement 'ON' or 'OFF'",
+        strict_discrete_set,
+        values=["ON", "OFF"],
+    )
+
+    measurement_time = Instrument.control(
+        ":ACQ:APER?",
+        ":ACQ:APER %f",
+        "Gate time for one measurement.",
+        validator=strict_range,
+        values=[2e-9, 1000],  # Programmer's guide 8-92
+    )
+
+    format = Instrument.control(
+        "FORM?",
+        "FORM %s",
+        "Reponse format (ASCII or REAL).",
+        validator=strict_discrete_set,
+        values=["ASCII", "REAL"],
+    )
+
+    interpolator_calibration = Instrument.control(
+        ":CAL:INT:AUTO?",
+        "CAL:INT:AUTO %s",
+        "Controls if interpolators should be calibrated automatically.",
+        strict_discrete_set,
+        values=["ON", "OFF"],
+        get_process=lambda v: "ON" if v else "OFF",
+    )
+
+    def configure_frequency_array_measurement(self, n_samples, channel):
+        """
+        Configure the counter for an array of measurements.
+
+        :param n_samples : The number of samples
+        :param channel : Measurment channel (A, B, C, E, INTREF)
+        """
+        n_samples = truncated_range(n_samples, [1, self.MAX_BUFFER_SIZE])
+        channel = strict_discrete_set(channel, self.CHANNELS)
+        channel = self.CHANNELS[channel]
+        self.write(f":CONF:ARR:FREQ {n_samples},(@{channel})")
+
+    def buffer_frequency_time_series(
+        self, channel, n_samples, sample_rate, trigger_source=None
+    ):
+        """
+        Record a time series to the buffer and read it out after completion.
+
+        :param channel: Channel that should be used.
+        :param n_samples : The number of samples
+        :param sample_rate : Sample rate in Hz.
+        :param trigger_source: Optionally specify a trigger source to start the
+                               measurement.
+        """
+        if self.interpolator_calibration:
+            max_sample_rate = 125e3
+        else:
+            max_sample_rate = 250e3
+        log.info(
+            "Calibration of interpolation is {}".format(
+                self.interpolator_calibration
+            )
+        )
+        sample_rate = strict_range(sample_rate, [0, max_sample_rate])
+        log.info("sample rate is {}".format(sample_rate))
+        measurement_time = 1 / sample_rate
+
+        self.clear()
+        self.format = "ASCII"
+        self.configure_frequency_array_measurement(n_samples, channel)
+        self.continuous = "OFF"
+        self.measurement_time = measurement_time
+
+        if trigger_source:
+            self.external_start_arming_source = trigger_source
+
+        # start the measurement (or wait for trigger)
+        self.write(":INIT")
--- a/docs/api/instruments/pendulum/cnt91.rst
+++ b/docs/api/instruments/pendulum/cnt91.rst
@@ -1,0 +1,7 @@
+################################
+Pendulum CNT91 frequency counter
+################################
+
+.. autoclass:: pymeasure.instruments.pendulum.cnt91.CNT91
+    :members:
+    :show-inheritance:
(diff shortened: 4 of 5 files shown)


pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+    def ask(self, command):
+        """
+        Writes the command to the instrument through the adapter
+        and returns the read response.
+
+        :param command: command string to be sent to the instrument
+        """
+        # Infinite loop occures if not redefined.
+        return super().ask(command).rstrip("\n")
+
```
msmttchr, line 58: Why infinite loop occurs?

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+    @property
+    def operation_complete(self):
+        """Is True if operation is complete."""
+        return self.ask("*OPC?") == "1"
+
```
msmttchr, line 63: SCPI instruments can use the `complete` property. I think there's no need to define this method.

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+        return self._batch_size
+
+    def read_buffer(self):
+        """
+        Read out the entire buffer.
+
+        :return : Frequency values from the buffer.
+        """
+        while not self.operation_complete:
+            # Wait until the buffer is filled.
```
msmttchr, line 78: `self.complete` should do the job

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+    external_start_arming_source = Instrument.control(
+        "ARM:SOUR?",
+        "ARM:SOUR %s",
+        (
+            "Select arming input or switch off the start arming function."
+            "Options are 'A', 'B' and 'E' (rear). 'IMM' turns trigger off."
+        ),
+        validator=strict_discrete_set,
+        values={"A": "EXT1", "B": "EXT2", "E": "EXT4", "IMM": "IMM"},
+        map_values=True,
```
msmttchr, line 100: Just a curiosity, why mapping names to A, B and E ? Do you need to match instrument labelling?

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+        "Turn continious measurement 'ON' or 'OFF'",
+        strict_discrete_set,
+        values=["ON", "OFF"],
+    )
+
+    measurement_time = Instrument.control(
+        ":ACQ:APER?",
+        ":ACQ:APER %f",
+        "Gate time for one measurement.",
+        validator=strict_range,
```
msmttchr, line 123: I think you could specify the units.

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+    )
+
+    interpolator_calibration = Instrument.control(
+        ":CAL:INT:AUTO?",
+        "CAL:INT:AUTO %s",
+        "Controls if interpolators should be calibrated automatically.",
+        strict_discrete_set,
+        values=["ON", "OFF"],
+        get_process=lambda v: "ON" if v else "OFF",
+    )
```
msmttchr, line 142: What about mapping ON and OFF to 1 and 0?

msmttchr (changes requested):
@bleykauf thanks for your contribution.
I made few comments and suggestions.
I think some further review will be needed.

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+        ),
+        validator=strict_discrete_set,
+        values={"A": "EXT1", "B": "EXT2", "E": "EXT4", "IMM": "IMM"},
+        map_values=True,
+    )
+
+    external_arming_start_slope = Instrument.control(
+        "ARM:SLOP?",
+        "ARM:SLOP %s",
+        "Set slope for the start arming condition.",
```
LongnoseRob, line 106: This should be a docstring if I am not mistaken

bilderbuchi: Do you mean a triple-quote delimited string `"""bla bla"""`? If so, that is only Python's way of defining an optionally multiline string. A normal string is fine here.

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+        validator=strict_discrete_set,
+        values=["POS", "NEG"],
+    )
+
+    continuous = Instrument.control(
+        "INIT:CONT?",
+        "INIT:CONT %s",
+        "Turn continious measurement 'ON' or 'OFF'",
```
bilderbuchi, line 114: *continuous

pymeasure/instruments/pendulum/cnt91.py
bilderbuchi: or calling it `interpolator_autocalibrated` and accepting boolean True/False.

AUTHORS.txt
```diff
@@ -41,3 +41,4 @@ Nicola Corna
 Robert Eckelmann
 Sam Condon
 Andreas Maeder
+Bastian Leykauf
```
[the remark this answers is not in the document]
bleykauf: Yes, I have forked and branched before I was added to the author file. I still don't really get why there is a conflict, as the files look the same. Anyway, this can be resolved upon merge.

docs/api/instruments/pendulum/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.rohdeschwarz
+
```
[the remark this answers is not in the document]
bleykauf: Oops... fixed in 4e477c9 .

docs/api/instruments/pendulum/index.rst
```diff
@@ -0,0 +1,12 @@
...
+########
+Pendulum
+########
+
+This section contains specific documentation on the Pendulum hwarz instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments>`.
+
```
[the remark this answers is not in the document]
bleykauf: Another oops. Fixed in d491777.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: Infinite loop is the wrong term. When I ran a `Procedure` without modifying `ask` it got stuck.

But I will have a closer look at this, i.e. check what is actually returned by the device and why it is not properly taken care of by the the `ask` method of  `Instrument`.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: I did have some trouble with using `complete`. At least for the CNT-91 it did not return "1" but just a property object.

Will check more thoroughly and report back.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: Yes, A, B, E are the labels on the instrument.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: I have not yet lost hope completely that at some point in the future I will be able to write this word correctly... Fixed in b158c7c.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: True, fixed in e4d4d0c.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: If using boolean for this kind of on/off switches I can change that. 

With this instrument you can set this switch with either "ON", "OFF" or 1 and 0 using SCPI. I am not sure what would be the best way here: should it be possible to also pass 1 and 0 in the python implementation? What about True/False for setting?

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+        values=["ON", "OFF"],
+        get_process=lambda v: "ON" if v else "OFF",
+    )
+
+    def configure_frequency_array_measurement(self, n_samples, channel):
+        """
+        Configure the counter for an array of measurements.
+
+        :param n_samples : The number of samples
+        :param channel : Measurment channel (A, B, C, E, INTREF)
```
LongnoseRob, line 149: trailing  space in  `:param n_samples :` breaks the formatting.

pymeasure/instruments/pendulum/cnt91.py
LongnoseRob: Yes exactly.
I had some issues with the display of short docstrings in the documentation in the past and have defaulted to use triple-qouted docstrings with a new line at the end.

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def ask(self, command):
+        """
+        Writes the command to the instrument through the adapter
+        and returns the read response.
+
+        :param command: command string to be sent to the instrument
+        """
+        # Infinite loop occures if not redefined.
+        return super().ask(command).rstrip("\n")
```
LongnoseRob, line 57: instead of using the `.rstrip()` here, it might be better to set `read_termination = "\r\n",` as part of the `super().__init()` call.
 more details on the terminations [here](https://pyvisa.readthedocs.io/en/latest/introduction/communication.html#getting-the-instrument-configuration-right)

bleykauf: Thanks! Missed that part of the documentation. This is indeed much cleaner. Fixed in 5e2e222.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: I don't have a strong opinion either way. Changed to triple quotes since in 1704e41 since it seems a little bit more of the standard formatting and number of lines is the same.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: Fixed in 57ef117 also in some other instances.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: So it's a simple line ending thing. I thought that this was automatically stripped by `Instrument.ask`. 

@LongnoseRob clarified this in https://github.com/pymeasure/pymeasure/pull/508#discussion_r742808399.

Should be clearer now.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: Could not reproduce so the problem I had earlier must have been something else.

Fixed in 6804299.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: Fixed in 6804299.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: I think in the end accepting and returning booleans is the cleanest and most pythonic method. I changed this also for `continuous` in 0a0415b.

bleykauf:
I noticed that sometimes there is some garbage data at the end of the buffer. I added an aptional argument to check for the length of the read out buffer and trucate it if necessary, see adf50d4. Basically the entire code was written bei github Copilot (including docstring and comments) with the only input from my side was the addition of the parameter and the `if` statement at the beginning of the check...

I also removed two of the log messages that are not really useful,  653155c.

Thanks everyone for the feedback!

I think the only remaining issue is with the `AUTHORS.txt` that can be resolved upon merge.

AUTHORS.txt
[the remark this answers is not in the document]
msmttchr: I could suggest to visit you branch dev/pendulum-cnt91 with web browser and click on fetch upstream to realign to master changes.
Screenshot_20211104-200306_Chrome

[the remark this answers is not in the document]
bleykauf: Thanks. I actually did that to my forked master branch. I thought I then integrated these changes into `bleykauf:dev/pendulum-cnt91` (where I also squashed some changes from an intermediate branch to not have too many commits with unverified code (did a lot from home without the actual device present).

I already have a new device in the pipeline, so I will try to get it right this time.

Thank you for all the feedback, it's really great contributing to this project :)

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+class CNT91(Instrument):
+    """Represents a Pendulum CNT-91 frequency counter."""
+
+    CHANNELS = {"A": 1, "B": 2, "C": 3, "E": 4, "INTREF": 6}
+    MAX_BUFFER_SIZE = 32000  # User Manual 8-38
+
+    def __init__(self, resourceName, **kwargs):
+        super().__init__(resourceName, "Pendulum CNT-91", **kwargs)
```
msmttchr, line 46:
```suggestion
        super().__init__(resourceName, "Pendulum CNT-91", timeout = 120000, read_termination = '\n', **kwargs)
```
I think you can apply change above and remove following two lines.

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,182 @@
...
+    continuous = Instrument.control(
+        "INIT:CONT?",
+        "INIT:CONT %s",
+        "Controls whether to perform continuous measurements.",
+        strict_discrete_set,
+        values=[True, False],
+        set_process=lambda x: "ON" if x else "OFF",
+        get_process=lambda x: bool(x),
+    )
+
```
msmttchr, line 114: What about using
`values = {True: "1", False : "0"}` and `map_values = True` ?

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,182 @@
...
+    interpolator_autocalibrated = Instrument.control(
+        ":CAL:INT:AUTO?",
+        "CAL:INT:AUTO %s",
+        "Controls if interpolators should be calibrated automatically.",
+        strict_discrete_set,
+        values=[True, False],
+        set_process=lambda x: "ON" if x else "OFF",
+        get_process=lambda x: bool(x),
+    )
```
msmttchr, line 139: Similar comment to `continuos` property

pymeasure/instruments/pendulum/cnt91.py
bleykauf: Fixed in 3575d75.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: Yes, much cleaner. I thought for some reason that booleans are not allowed as dictionary keys. Also just realized that the mapping works in both directions.

Fixed in 9885993.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: Fixed in 9885993.

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,184 @@
...
+        if not hasattr(self, "_batch_size"):
+            self._batch_size = int(self.ask("FORM:SMAX?"))
+        return self._batch_size
+
+    def read_buffer(self, expected_length=0):
+        """
+        Read out the entire buffer.
+
+        :param expected_length: The expected length of the buffer. If more
+            data is read, values at the end are be removed. Defaults to 0,
```
bilderbuchi, line 66:
```suggestion
            data is read, values at the end are removed. Defaults to 0,
```

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -0,0 +1,184 @@
...
+        :param n_samples: The number of samples
+        :param sample_rate: Sample rate in Hz
+        :param trigger_source: Optionally specify a trigger source to start the
+            measurement
+        """
+        if self.interpolator_autocalibrated:
+            max_sample_rate = 125e3
+        else:
+            max_sample_rate = 250e3
+        sample_rate = strict_range(sample_rate, [0, max_sample_rate])
```
bilderbuchi, line 171: 0 as a sample rate is really admissible? I think the lower limit is formed by the upper of `measurement_time`, i.e. 1/1000

bilderbuchi (changes requested):
LGTM, except for some small remarks.

pymeasure/instruments/pendulum/cnt91.py
bleykauf: I agree, added a `MAX_MEASUREMENT_TIME` in 7ec480e to avoid having two dependend magic numbers in the code. Note that `Instrument.control` needs access to this value but it doesn't have access to `self`. Since I didn't find a clean build-in solution to define class attributes, I put this constant in the main body of the module.

Open for ideas of how to improve this.

docs/api/instruments/pendulum/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.pendulum
+
```
bilderbuchi, line 2: whoops, the above warning ^ means that you need to add a line to pymeasure/docs/api/instruments/index.rst as this is a new manufacturer [docs](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#adding-documentation)

bilderbuchi (approved):
When this doc warning is fixed it's good to go from my side, thank you! :-)

docs/api/instruments/pendulum/index.rst
bleykauf: Fixed in 6915d29.

bilderbuchi:
Thank you for bringing this to merge readiness! :-)

PR #511, opened 2021-11-05
Add Rohde&Schwarz FSL spectrum analyzer.

bleykauf:
I added the Rohde&Schwarz FSL spectrum analyzer, including the most common commands and a dedicated `Marker` class. 

The FSL handles both numerical value as well as physical quantities like "1.5 GHz". The latter feature is implemented by the device itself. It also automatically chooses the closest available value for some settings or ignores them if they are out of range (like setting the span wider than the maximum value). I thus do not perform validation in the interface itself in these cases to not change the behaviour compared to pure SCPI commands.

I also added documenation including an example above the API documentation.

I tested the commands with an FSL-18.

### DIFF
--- a/docs/api/instruments/rohdeschwarz/fsl.rst
+++ b/docs/api/instruments/rohdeschwarz/fsl.rst
@@ -1,0 +1,138 @@
+######################################
+R&S FSL spectrum analyzer
+######################################
+
+Connecting to the instrument via network
+----------------------------------------
+Once connected to the network, the instrument's IP address can be found by
+clicking the "Setup" button and navigating to "General Settings" -> "Network
+Address".  
+
+It can then be connected like this:
+
+.. code:: python
+    from pymeasure.instruments.rohdeschwarz import FSL
+    fsl = FSL("TCPIP::192.168.1.123::INSTR")
+
+Getting and setting parameters
+------------------------------
+
+Most parameters are implemented as properties, which means they can be read and
+written (getting and setting) in a consistent and dimple way. If numerical
+values are provided, base units are used (seconds, hertz, decibel, ...). 
+Alternatively, the values can also be provided with a unit, e.g. :code:"1.5 GHz" or
+:code:"1.5GHz". Return values are always numerical.
+
+.. code:: python
+
+    # Getting the current center frequency
+    fsl.freq_center
+
+    9000000000.0
+
+.. code:: python
+
+
+    # Changing it to 10 MHz by providing the numerical value 
+    fsl.freq_center = 10e6
+
+.. code:: python
+
+    # Verifying:
+    fsl.freq_center
+
+    10000000.0
+
+.. code:: python
+
+    # Changing it to 9 GHz by providing a string and verifying the result
+    fsl.freq_center = '9GHz'
+    fsl.freq_center
+
+    9000000000.0
+
+.. code:: python
+
+    # Setting the span to maximum
+    fsl.freq_span = '7 GHz'
+
+Reading a trace
+---------------
+
+We will read the current trace
+
+.. code:: python
+
+    x, y = fsl.read_trace()
+
+Markers
+-------
+
+Markers are implemented as their own class. You can create them like
+this:
+
+.. code:: python
+
+    m1 = fsl.create_marker()
+
+Set peak exursion:
+
+.. code:: python
+
+    m1.peak_excursion = 3
+
+Set marker to a specific position:
+
+.. code:: python
+
+    m1.x = 10e9
+
+Find the next peak to the left and get the level:
+
+.. code:: python
+
+    m1.to_next_peak('left')
+    m1.y
+
+    -34.9349060059
+
+Delta markers
+~~~~~~~~~~~~~
+
+Delta markers can be created by setting the appropriate keyword.
+
+.. code:: python
+
+    d2 = fsl.create_marker(is_delta_marker=True)
+    d2.name
+
+    'DELT2'
+
+Example program
+---------------
+
+Here is an example of a simple script for recording the peak of a signal.
+
+.. code:: python
+
+    m1 = fsl.create_marker() # create marker 1
+
+    # Set standard settings, set to full span
+    fsl.continuous_sweep = False
+    fsl.freq_span = '18 GHz'
+    fsl.rbw = "AUTO"
+    fsl.vbw = "AUTO"
+    fsl.sweep_time = "AUTO"
+
+    # Perform a sweep on full span, set the marker to the peak and some to that marker
+    fsl.single_sweep()
+    m1.to_peak()
+    m1.zoom('20 MHz')
+
+    # take data from the zoomed-in region
+    fsl.single_sweep()
+    x, y = fsl.read_trace()
+
+.. autoclass:: pymeasure.instruments.rohdeschwarz.fsl.FSL
+    :members:
+    :show-inheritance:
--- a/docs/api/instruments/rohdeschwarz/fsl.rst
+++ b/docs/api/instruments/rohdeschwarz/fsl.rst
@@ -19,7 +19,7 @@ Getting and setting parameters
 ------------------------------
 
 Most parameters are implemented as properties, which means they can be read and
-written (getting and setting) in a consistent and dimple way. If numerical
+written (getting and setting) in a consistent and simple way. If numerical
 values are provided, base units are used (s, Hz, dB, ...). 
 Alternatively, the values can also be provided with a unit, e.g. ``"1.5 GHz"``
 or ``"1.5GHz"``. Return values are always numerical.
@@ -32,8 +32,6 @@ or ``"1.5GHz"``. Return values are always numerical.
     9000000000.0
 
 .. code:: python
-
-
     # Changing it to 10 MHz by providing the numerical value 
     fsl.freq_center = 10e6
 
--- a/docs/api/instruments/rohdeschwarz/fsl.rst
+++ b/docs/api/instruments/rohdeschwarz/fsl.rst
@@ -11,6 +11,7 @@ Address".
 It can then be connected like this:
 
 .. code:: python
+
     from pymeasure.instruments.rohdeschwarz import FSL
     fsl = FSL("TCPIP::192.168.1.123::INSTR")
 
@@ -19,9 +20,9 @@ Getting and setting parameters
 
 Most parameters are implemented as properties, which means they can be read and
 written (getting and setting) in a consistent and dimple way. If numerical
-values are provided, base units are used (seconds, hertz, decibel, ...). 
-Alternatively, the values can also be provided with a unit, e.g. :code:"1.5 GHz" or
-:code:"1.5GHz". Return values are always numerical.
+values are provided, base units are used (s, Hz, dB, ...). 
+Alternatively, the values can also be provided with a unit, e.g. ``"1.5 GHz"``
+or ``"1.5GHz"``. Return values are always numerical.
 
 .. code:: python
 
--- a/docs/api/instruments/rohdeschwarz/fsl.rst
+++ b/docs/api/instruments/rohdeschwarz/fsl.rst
@@ -119,8 +119,8 @@ Here is an example of a simple script for recording the peak of a signal.
     # Set standard settings, set to full span
     fsl.continuous_sweep = False
     fsl.freq_span = '18 GHz'
-    fsl.rbw = "AUTO"
-    fsl.vbw = "AUTO"
+    fsl.res_bandwidth = "AUTO"
+    fsl.video_bandwidth = "AUTO"
     fsl.sweep_time = "AUTO"
 
     # Perform a sweep on full span, set the marker to the peak and some to that marker
--- a/docs/api/instruments/rohdeschwarz/fsl.rst
+++ b/docs/api/instruments/rohdeschwarz/fsl.rst
@@ -32,6 +32,7 @@ or ``"1.5GHz"``. Return values are always numerical.
     9000000000.0
 
 .. code:: python
+
     # Changing it to 10 MHz by providing the numerical value 
     fsl.freq_center = 10e6
 
--- a/pymeasure/instruments/rohdeschwarz/fsl.py
+++ b/pymeasure/instruments/rohdeschwarz/fsl.py
@@ -78,29 +78,28 @@ class FSL(Instrument):
         "Attenuation in dB.",
     )
 
-    @property
-    def res_bandwidth(self):
-        """Resolution bandwidth in Hz. Can be set to 'AUTO'."""
-        return float(self.ask("BAND:RES?"))
-
-    @res_bandwidth.setter
-    def res_bandwidth(self, value):
-        if type(value) is str and value.upper() == "AUTO":
-            self.write("BAND:RES:AUTO ON")
+    @staticmethod
+    def _number_or_auto(value):
+        # helper for the bandwidth setting
+        if isinstance(value, str) and value.upper() == "AUTO":
+            return ":AUTO ON"
         else:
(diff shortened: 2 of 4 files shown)


LongnoseRob (commented):
Thank you for adding this new instrument, good work!
I like the enhanced documentation with the examples

Left some comments, mostly on the documentation side.

docs/api/instruments/rohdeschwarz/fsl.rst
```diff
@@ -0,0 +1,138 @@
...
+Connecting to the instrument via network
+----------------------------------------
+Once connected to the network, the instrument's IP address can be found by
+clicking the "Setup" button and navigating to "General Settings" -> "Network
+Address".  
+
+It can then be connected like this:
+
+.. code:: python
+    from pymeasure.instruments.rohdeschwarz import FSL
```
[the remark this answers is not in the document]
bleykauf: Fixed in b560615.

docs/api/instruments/rohdeschwarz/fsl.rst
```diff
@@ -0,0 +1,138 @@
...
+    fsl = FSL("TCPIP::192.168.1.123::INSTR")
+
+Getting and setting parameters
+------------------------------
+
+Most parameters are implemented as properties, which means they can be read and
+written (getting and setting) in a consistent and dimple way. If numerical
+values are provided, base units are used (seconds, hertz, decibel, ...).
```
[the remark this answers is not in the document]
bleykauf: Agreed. Fixed in b560615.

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,252 @@
...
+    )
+
+    # Markers --------------------------------------------------------------------
+
+    def create_marker(self, num=1, is_delta_marker=False):
+        """
+        Create a marker.
+        :param num: The marker number (1-4)
+        :param is_delta_marker: True if the marker is a delta marker, default
+                            is False.
```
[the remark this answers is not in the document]
bleykauf: This clashes with PEP8's 79 character line limit. I instead reduced the indention level to 1 throughout the whole module and added blank lines before the parameter section. It now renders properly.

docs/api/instruments/rohdeschwarz/fsl.rst
```diff
@@ -0,0 +1,138 @@
...
+    fsl = FSL("TCPIP::192.168.1.123::INSTR")
+
+Getting and setting parameters
+------------------------------
+
+Most parameters are implemented as properties, which means they can be read and
+written (getting and setting) in a consistent and dimple way. If numerical
+values are provided, base units are used (seconds, hertz, decibel, ...). 
+Alternatively, the values can also be provided with a unit, e.g. :code:"1.5 GHz" or
+:code:"1.5GHz". Return values are always numerical.
```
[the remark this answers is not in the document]
bleykauf: Fixed in b560615.

docs/api/instruments/rohdeschwarz/fsl.rst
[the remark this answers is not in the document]
bilderbuchi: this is a general thing about rst "[directives](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html)" -- first line the directive name, next any options/arguments, then a blank line, content (indented more), another blank line.

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+            self.write(f"SWE:TIME {value}")
+
+    continuous_sweep = Instrument.control(
+        "INIT:CONT?",
+        "INIT:CONT %s",
+        "Continuous (True) or single sweep (False)",
+        validator=strict_discrete_set,
+        values=[True, False],
+        set_process=lambda x: "ON" if x else "OFF",
+        get_process=lambda x: bool(x),
```
bilderbuchi, line 126: not sure what you're aiming for here and what the device returns on command_get, but maybe
```
        strict_discrete_set,
        values={True: "ON", False: "OFF"},
        map_values=True,
```

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+    def read_trace(self, n_trace=1):
+        """
+        Read trace data.
+
+        :param n_trace: The trace number (1-6). Default is 1.
+        :return: Nnumpy arrays of the trace data (frequency and amplitude).
+        """
+        y = np.array(self.values("TRAC{n_trace}? TRACE{n_trace}"))
+        x = np.linspace(self.freq_start, self.freq_stop, len(y))
+        return x, y
```
bilderbuchi, line 148: if you're already going to the trouble to return numpy objects, why not assemble into a 2D (len(y),2) array and return that?

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+
+    trace_mode = Instrument.control(
+        "DISP:TRAC:MODE?",
+        "DISP:TRAC:MODE %s",
+        "Trace mode ('WRIT', 'MAXH', 'MINH', 'AVER' or 'VIEW'",
```
bilderbuchi, line 153:
```suggestion
        "Trace mode ('WRIT', 'MAXH', 'MINH', 'AVER' or 'VIEW')",
```

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+        :param is_delta_marker: True if the marker is a delta marker, default
+            is False.
+        :return: The marker object.
+        """
+        return self.Marker(self, num, is_delta_marker)
+
+    class Marker:
+        def __init__(self, instrument, num, is_delta_marker):
+            """
+            Marker and Delte Marker class.
```
bilderbuchi, line 174:
```suggestion
            Marker and Delta Marker class.
```

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+        "DISP:TRAC:MODE?",
+        "DISP:TRAC:MODE %s",
+        "Trace mode ('WRIT', 'MAXH', 'MINH', 'AVER' or 'VIEW'",
+        validator=strict_discrete_set,
+        values=["WRIT", "MAXH", "MINH", "AVER", "VIEW"],
+    )
+
+    # Markers -----------------------------------------------------------------
+
+    def create_marker(self, num=1, is_delta_marker=False):
```
bilderbuchi, line 160: What happens if you create the same marker twice?

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+                self.name = "MARK"
+                if num > 1:
+                    # Marker 1 doesn't get a number.
+                    self.name = self.name + str(num)
+
+            self.activate()
+
+        def activate(self):
+            """Activate a marker."""
+            self.instr.write(f"CALC:{self.name}:STAT ON")
```
bilderbuchi, line 197: The way to avoid having to manually prefix this channel ID everywhere, and get a nicely composable channel/Marker with properties that you can create with the control/setting/measurement factories is to [implement read/write/ask methods that delegate to the instrument](https://github.com/pymeasure/pymeasure/blob/7580edd2cc5dfaafb9c0d6462d6f95cd87a3d468/pymeasure/instruments/tektronix/afg3152c.py#L127-L144). (although you're free to use modern f-strings instead of the % syntax, of course)

bilderbuchi (changes requested):
Thank you for the contribution, this looks good, I found a couple opportunities to improve the implementation.

pymeasure/instruments/rohdeschwarz/fsl.py
bleykauf: My main reason behind using numpy here is actually that there is no build-in `linspace`...

Not sure what is better here, importing and using numpy or creating teh linspace manually in this instance?

pymeasure/instruments/rohdeschwarz/fsl.py
bleykauf: Nothing special. You get back another Marker instance. Both do work. Do you see an issue with that?

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+
+    attenuation = Instrument.control(
+        "INP:ATT?",
+        "INP:ATT %s",
+        "Attenuation in dB.",
+    )
+
+    @property
+    def rbw(self):
```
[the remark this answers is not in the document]
bleykauf: Fixed in f0c402d.

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+
+    @rbw.setter
+    def rbw(self, value):
+        if type(value) is str and value.upper() == "AUTO":
+            self.write("BAND:RES:AUTO ON")
+        else:
+            self.write(f"BAND:RES {value}")
+
+    @property
+    def vbw(self):
```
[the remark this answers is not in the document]
bleykauf: Fixed in f0c402d.

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+    attenuation = Instrument.control(
+        "INP:ATT?",
+        "INP:ATT %s",
+        "Attenuation in dB.",
+    )
+
+    @property
+    def rbw(self):
+        """Resolution bandwidth in Hz. Can be set to 'AUTO'."""
+        return self.values("BAND:RES?")[0]
```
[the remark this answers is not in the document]
LongnoseRob: (The Manual)[https://cdn.rohde-schwarz.com/pws/dl_downloads/dl_common_library/dl_manuals/gb_1/f/sfl_1/FSL_OperatingManual_en_12.pdf] page 904 does unfortunately not give a query example.

pymeasure/instruments/rohdeschwarz/fsl.py
bleykauf: Cool, I like this. Changed in 6ea22bb .

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,256 @@
...
+                self.name = "DELT" + str(max(2, num))
+            else:
+                self.name = "MARK"
+                if num > 1:
+                    # Marker 1 doesn't get a number.
+                    self.name = self.name + str(num)
+
+            self.activate()
+
+        def write(self, command):
```
bilderbuchi, line 195: you have overridden `write`, but not the `read` method. If that is intentional (e.g. because the instrument does not include this CALC... prefix when replying), just leave a comment here so that future readers are not confused. :-)

bilderbuchi (changes requested):

pymeasure/instruments/rohdeschwarz/fsl.py
bleykauf: In that case the marker switches between being a delta marker and a regular marker every time one of the two `Marker` objects is used (indicated by the label M3 and D3 on the display of the instrument).

pymeasure/instruments/rohdeschwarz/fsl.py
bleykauf: True. Also, calling `x, y = fsl.read_trace` still works so I changed it in 1746fb1.

pymeasure/instruments/rohdeschwarz/fsl.py
bleykauf: In 86744f8 I added `read` and `ask` and also fixed the `value` method (I didn't actually delegate to `instrument` causing an infinite loop. Now you can actually read out marker positions etc.

One thing that I noticed is that the [link you provided earlier](https://github.com/pymeasure/pymeasure/blob/7580edd2cc5dfaafb9c0d6462d6f95cd87a3d468/pymeasure/instruments/tektronix/afg3152c.py#L127-L144) doesn't return anything for the `read` and `ask` methods. Maybe this didn't get caught because `values` was always used by the other methods?

bilderbuchi: Wow, nice catch about those methods not returning! That looks like an error - can you also include a fix while you're at it?

pymeasure/instruments/rohdeschwarz/fsl.py
bilderbuchi: thanks for clarifying!

pymeasure/instruments/rohdeschwarz/fsl.py
bleykauf: I opened [another pull request](https://github.com/pymeasure/pymeasure/pull/519) since I also took the opportunity to fix some formatting issues.

pymeasure/instruments/rohdeschwarz/fsl.py
```diff
@@ -0,0 +1,259 @@
...
+    def sweep_time(self):
+        """Sweep time in s. Can be set to 'AUTO'."""
+        return self.values("SWE:TIME?")[0]
+
+    @sweep_time.setter
+    def sweep_time(self, value):
+        if type(value) is str and value.upper() == "AUTO":
+            self.write("SWE:TIME:AUTO ON")
+        else:
+            self.write(f"SWE:TIME {value}")
```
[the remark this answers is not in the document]
bilderbuchi: @bleykauf I don't know how it is for you, but Github "helpfully" hid this and the following second unresolved conversations from view, so I only realised now that this is still open.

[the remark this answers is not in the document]
bleykauf: Yeah, I just got the email, otherwise I would never have seen this...

I just made a change, will check tomorrow if it works correctly.

pymeasure/instruments/rohdeschwarz/fsl.py
bilderbuchi: @bleykauf same here - ping so you see this.

pymeasure/instruments/rohdeschwarz/fsl.py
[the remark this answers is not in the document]
bleykauf: Didn't work, staticmethods apparently are not callable in the way I was hoping for. `_number_or_auto` is now a module level function, 18efbc3. This works now.

See also https://stackoverflow.com/questions/41921255/staticmethod-object-is-not-callable.

pymeasure/instruments/rohdeschwarz/fsl.py
bleykauf: Implemented in 97198bc.

bleykauf:
All requested changes have now been implemented.

If there are now further issues I would suggest to merge this. I think it is a good idea to squash the commits before merging since this is a completely new device so I don't see a lot of value in documenting all the history of fixes of typos and mistakes.

bilderbuchi:
Thank you for the changes! The last (I promise :D) thing: Sorry for not being more explicit, but can't the new `_number_or_auto` be used to also simplify the `sweep_time` property definition? (I'm guessing you just missed that?)
Indeed, when that last thing is done I'll squash before merging, thanks for bringing that up. :-)

bleykauf:
@bilderbuchi: Yes, I just missed that. It's changed now.

bilderbuchi:
Thanks for iterating on the fixes/updates!

PR #515, opened 2021-11-06
Fix typo SCIP->SCPI

mfikes:


msmttchr (approved):
Thanks for spotting the typo

PR #516, opened 2021-11-07
Agilent 11713A Attenuator / Switch Driver

mfikes:


pymeasure/instruments/agilent/agilent11713A.py
```diff
@@ -0,0 +1,88 @@
...
+    .. code-block:: python
+
+        asd = Agilent11713A("GPIB::28")
+
+        asd.reset()                             # Reset the instrument
+
+        asd.attenuator_x = [0, 1, 0, 1]         # Sets the attenuator X values
+        asd.attenuator_y = [1, 1, 0, 1]         # Sets the attenuator Y values
+        asd.switches = [False, True]            # Sets the switch values
+
```
msmttchr, line 48: Some settings are 1 or 0 and others are True or False. Is this intentional ?

pymeasure/instruments/agilent/agilent11713A.py
```diff
@@ -0,0 +1,88 @@
...
+from pymeasure.instruments import Instrument
+
+def on_off_command(v):
+    if v:
+        return "A"
+    else:
+        return "B"
+
+def command_string(designators, v):
+    return ''.join(on_off_command(v[i]) + designators[i] for i in range(0, len(designators)))
```
msmttchr, line 34: What about adding a dictionary to map 1 to 'A' and 0 to 'B' instead of `on_off_command` ?

pymeasure/instruments/agilent/agilent11713A.py
```diff
@@ -0,0 +1,88 @@
...
+        """,
+        set_process=lambda v: command_string("5678", v)
+    )
+
+    switches = Instrument.setting(
+        "%s",
+        """ A property that sets the switch values via a two-element array.
+        """,
+        set_process=lambda v: command_string("90", v)
+    )
```
msmttchr, line 88: I think you are providing the basic support for controlling the switch.
I am not sure whether some convenience method can be added to ease users experience.
I was thinking something which receives the desired attenuation and translate it into relevant switch configuration.

msmttchr (commented):
Thanks for the contribution. I added few comments

pymeasure/instruments/agilent/agilent11713A.py
mfikes: The intent was to simply allow any truthy / falsy values. But perhaps simply allowing 0 / 1 for now would keep things simple.

pymeasure/instruments/agilent/agilent11713A.py
mfikes: That level of abstraction could indeed be useful to end users who simply want to use the attenuator / switch driver as it is intended to be used.

For example, table 6-3 in the manual shows that, in order to set 30 dB of attenuation on a 3-section attenuator attached to the X port, you would issue the GPIB command `A12B3` and the same level of attenuation for a 4-section attenuator would involve `A14B23`. (Not so user-friendly!)

You could envision an API that allows attenuation values to be set for attenuators connected to either the 4-bit X or Y port. It would seem to require arguments such as the port (X or Y), how many sections the attenuator has (3 or 4), the maximum attenuation provided by the attenuator (11 dB, 70 dB, or 110 dB), and the desired attenuation value.

```python
asd.set_attenuation("X", 3, 70, 30)
```

Alternatively, since the first 3 arguments define an attenuator connected to a port, maybe an API that allows you to partially apply or "capture" the first three arguments and create an "attenuator" with an attenuation property might be viewed more a more useful approach:

```python
attenuator = asd.create_attenuator("X", 3, 70)
attenuator.attenuation = 30
attenuator.attenuation = 10
```

For the switch buttons (S9 and S0), perhaps there might be a similar table-driven approach for common switches. (I don't see any in the manual). On the other hand, perhaps a simple API that exposes those two particular buttons as properties might be useful:

```python
asd.s9 = 1
asd.s0 = 0
```

msmttchr: @mfikes, thanks for the clear explanation. I thought initially that attenuators were connected to instrument. Nevertheless your proposal to have a create_attenuator can be interesting. On the other hand I am not sure about the parameters, I would supply the 'X' parameter as well as a list indicating the attenuation per segment and the list length would indicate the number of segments (I am not sure how to infer the attenuation per segment having the number of segments and  max attenuation value).

In alternative, you could add 3 optional parameter in the instrument constructor to indicate the X, Y and switch segments. Then a method or a property allows to set the attenuation performing the required math and finally issuing the relevant commands to the instrument. I am not sure to propose something for the switch though. In any case, consider this comment as a sort of brainstorming and not a strong indication to make changes.

mfikes: @msmttchr Yes, the attenuators are indeed separate from the instrument and connect to it via ports on the rear. There are two ports on the back where you can connect attenuators, port X and port Y, and on the front of the instrument there are 4 buttons for the X port and 4 buttons for the Y port.

For aid in brainstorming, here's the table from the manual for a 3-segment 70 dB attenuator, but represented as binary (note that the 4th button is not used for a 3-segment attenuator):

`000`: 0 dB
`100`: 10 dB
`010`: 20 dB   (note for this one, ignoring what appears to be typos in the manual)
`110`: 30 dB
`001`: 40 dB
`101`: 50 dB
`011`: 60 dB
`111`: 70 dB

You can see that the first segment has 10 dB, the second has 20 dB, and the third has 40 dB. In other words it is counting from 0 to 7 in binary. So, really, any higher level API would seem to be helping the user by letting them pass, a parameter in dB, and it determining the associated binary representation and setting it as appropriate.

It seems like the hard part here is determining a clean, useful API. (Implementing that API is probably the easy part, IMHO.)

I _do_ like the idea of having the implementation to the binary conversion for the user, who may be simply wanting to run an experiment involving setting various attenuation levels in terms of dB (rather than low-level binary).

msmttchr: In table 6-3, I see the following combinations:
4 segments:
1,2,4,4              max 11dB
10,20,40,40    max 110dB
10,20,20,20    max 70dB
3 segments:
10,20,40.          max 70dB

and the table quotes them as "various three and four-section attenuators", so I did conclude (I may be wrong on this) that there could be other attenuators with:
- different attenuation per segment, e.g 1,2,4,8
- different order of attenuation, e.g: 40,20,10
- different number of segments, e.g 10,20
So I was thinking that simply specifying the list of attenuators could be sufficient to capture both the number of segments and the attenuation per segment.

bilderbuchi:
Thank you for supplying yet another interestingly strange instrument :grin: -- it seems there is no danger at all of pymeasure just obtaining the N+1st driver for a power supply or oscilloscope.. :D

pymeasure/instruments/agilent/agilent11713A.py
bilderbuchi: > Alternatively, since the first 3 arguments define an attenuator connected to a port, maybe an API that allows you to partially apply or "capture" the first three arguments and create an "attenuator" with an attenuation property might be viewed more a more useful approach

I agree that this seems the most promising approach. We are currently in the process of merging another instrument (#511) that uses the instrument to [define](https://github.com/pymeasure/pymeasure/pull/511/files#diff-a8aafadb326eb6887d766db276b907fdf28095eb7b4a7211b83f27f7379ad6edR159-R168) a [Marker class](https://github.com/pymeasure/pymeasure/pull/511/files#diff-a8aafadb326eb6887d766db276b907fdf28095eb7b4a7211b83f27f7379ad6edR170) in a manner that seems useful here.

The proposed API looks good to me, maybe like this:
```
attenuator = asd.create_attenuator(port="X", sections= (10, 20, 40), model=None)
attenuator.attenuation = 30
attenuator.attenuation = 10
```
`create_attenuator` creates and returns a class instance that stores a reference to its instrument parent, e.g. in `self.instrument`.
The `sections` sequence defines the segment attenuations for the later math to find the correct switch config. 
Alternatively, `model` can give the model ID of a specific attenuator (strings like "HP8496"), which consults a lookup table to figure out the segments (which could be more convenient for the user).
Considering the complexity of figuring out admissible values and switch configurations, it's probably better to implement a plain property `Attenuator.attenuation` that 
* accepts a desired attenuation value
* checks if the present segments can fulfill that attenuation, and bail otherwise
* assembles the command string and uses its `instrument` attribute to delegate sending to `Agilent11713A.write`
There is no need for a getter, as the device apparently cannot send a data message.

Alternatively, and (considering that there will only ever be 2 attenuators) maybe even easier, would be to implement a method `Agilent11713A.configure_attenuator()` with the same signature, that writes to a segment list, and then have two properties `x_attentuation` and `y_attenuation` that internally dispatch to a private method to determine the data message and check validity of the input.

bilderbuchi:
Also, I love those old instruments that include servicing/calibration info, circuit diagrams and even PCB shots in the user manual :D ...reminds me of old slide projectors...

mfikes:
> Thank you for supplying yet another interestingly strange instrument 😁 -- it seems there is no danger at all of pymeasure just obtaining the N+1st driver for a power supply or oscilloscope.. :D

Hah, yeah. This one ostensibly looked like an easy one to start with, given it only has 10 toggle buttons on it. But, it turns out the real challenge is in creating the API for it. :) I've got some more mainstream instruments to add apart from this one. :)

bmoneke:
@mfikes do you want to pick up this PR or may I close it? It can be reopened whenever you want to continue.

mfikes:
It can be closed.

PR #517, opened 2021-11-07
Fix SCPI used for Keithley 2400 voltage NPLC

mfikes:


pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -191,7 +191,7 @@ class Keithley2400(Instrument, KeithleyBuffer):
         values=[-210, 210]
     )
     voltage_nplc = Instrument.control(
-        ":SENS:CURRVOLT:NPLC?", ":SENS:VOLT:NPLC %g",
+        ":SENS:VOLT:NPLC?", ":SENS:VOLT:NPLC %g",
         """ A floating point property that controls the number of power line cycles
```
msmttchr, line 195: Most likely a copy & paste error and probably command not much used.

msmttchr (approved):
Mike thanks for your fix.
In the future, if you have time, also some testing can be added to catch these problems.
If you are curious, have a look at the tests directory.

PR #519, opened 2021-11-11
Fix missing return statements in Tektronix AFG3152C.

bleykauf:
As noticed in https://github.com/pymeasure/pymeasure/pull/511#discussion_r746893753, the implementation of the instrument was missing two `return` statements. This has been fixed here.

In addition I also fixed the code formatting by applying black (with 79 character line length). Before, flake8 was complaining about missing blank lines. Also removed an unused import.


bilderbuchi:
Thanks, looking good, although hard to judge considering how Github chops up that diff view -- thanks for meaningful commit separation!

PR #520, opened 2021-11-11
AnritsuMS9740A

md12g12:
Hi, I'm very new to Git but hopefully I have done this correctly. 
The AnritsuMS9740A instrument I am adding is very similar to an existing instrument on pymeasure, the AnritsuMS9710C so I have mostly copied that whilst updating the commands that no longer work. Running my code as a script works well enough so hopefully this is useful to someone.

### DIFF
--- a/docs/api/instruments/anritsu/anritsuMS9740A.rst
+++ b/docs/api/instruments/anritsu/anritsuMS9740A.rst
@@ -1,0 +1,3 @@
+.. autoclass:: pymeasure.instruments.anritsu.AnritsuMS9740A
+    :members:
+    :show-inheritance:
--- a/docs/api/instruments/anritsu/index.rst
+++ b/docs/api/instruments/anritsu/index.rst
@@ -11,3 +11,4 @@ This section contains specific documentation on the Anritsu instruments that are
 
    anritsuMG3692C
    anritsuMS9710C
+   anritsuMS9740A 
--- a/pymeasure/instruments/anritsu/AnritsuMS9740A.py
+++ b/pymeasure/instruments/anritsu/AnritsuMS9740A.py
@@ -1,0 +1,351 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Wed Oct 13 15:38:48 2021
+
+@author: Matthew Delaney
+"""
+
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2021 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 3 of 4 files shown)


bilderbuchi:
Hi, and welcome! I think you're doing good on the Git front. :-)
Concerning creating this instrument from another very similar one: Copy-pasting is less than ideal -- it increases the maintenance effort because now any future fixes have to happen in 2 places (and we have to remember that!). Please derive yours from the existing one instead, add any functionality that is missing and override/replace any that is different. This way it's much clearer what the difference between the two models is, and we can reuse code between them.
If the other one is more featureful we can also go in the other direction, or find a common subset and push that into a base class. Go into your notes/recollection of what you had to change, and replay that onto a derived class, please, and we will assist with any problems. :-)

Something like this should get you started:
```
class AnritsuMS9740A(AnritsuMS9710C):
    new_control = Instrument.control(
        "bla?", "bla %s", 
        "docs",
        # etc
    )

md12g12:
Hi Christopher,

Thanks for the feedback - I guess its about time I learn about
inheritance (I think thats what you are asking for?). I will have a crack
at it then submit a new pull request with the updated version. In the
meantime has my previous pull request been deleted (I cant see it anymore
so I assume so?). And a final question - I cant find anything on the
contribution page about writing documentation outside of the inline
comments. Is the readthedocs.io page automatically generated or is this
manually created?

Many thanks,
Matt

bilderbuchi:
> Thanks for the feedback - I guess its about time I learn about
inheritance (I think thats what you are asking for?). 

Indeed, yes, this is inheritance. I find the realpython materials quite approachable, e.g. https://realpython.com/inheritance-composition-python/, but you can also go straight for the docs: https://docs.python.org/3/tutorial/classes.html

> In the meantime has my previous pull request been deleted (I cant see it anymore
so I assume so?).

Not by us, but I think it happened when you updated your `master` branch with our master (and both were at the same commit). A good way to handle pull requests is to have them live in a separate branch that you start from the master - then any subsequent updates to master are not interfering, and the PR branch can live as long as it needs to. 

> And a final question - I cant find anything on the
contribution page about writing documentation outside of the inline
comments. Is the readthedocs.io page automatically generated or is this
manually created?

There a little info [here](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#adding-documentation). 
Teh docs autogenerated by CI from our codebase when we push branches, see the "docs and linting" job under the "Checks" tab of a PR. You can also do this, after installing `sphinx` and `sphinx-rtd_theme` by running `make html` in the `docs` directory.

PR #521, opened 2021-11-12
2nd attempt at a first pull request - AnritsuMS9740A

md12g12:
I have updated the changed commands from the AnritsuMS9710, a very similar instrument. It is my first time using class inheritance so hopefully it looks right. I will want to add some new commands in the future but for now I want to get this live so I am familiar with the process and im not sure when I will be able to come back to this project.
Best,
Matt


pymeasure/instruments/anritsu/AnritsuMS9740A.py
```diff
@@ -0,0 +1,140 @@
...
+    ONOFF_MAPPING = {True: 'ON', False: 'OFF', 1: 'ON', 0: 'OFF'}
+
+    ###########
+    #  Modes  #
+    ###########
+
+    measure_mode = Instrument.measurement(
+        "MOD?", "Returns the current Measure Mode the OSA is in.",
+        values={None: 0, "SINGLE": 1.0, "REPEAT": 2.0, "POWER": 3.0},
+        map_values=True
```
bilderbuchi, line 85: It is good practice in Python to also add a trailing comma to the last element of a argument list/dict/collection. This way, should you add another argument later, the git diff only has that new line (because you don't have to add a comma to this line, then).

pymeasure/instruments/anritsu/AnritsuMS9740A.py
```diff
@@ -0,0 +1,140 @@
...
+
+    resolution_vbw = Instrument.control(
+        "VBW?", "VBW %s", "Video Bandwidth Resolution",
+        validator=strict_discrete_set,
+        values=["1MHz", "100kHz", "10kHz", "2kHz", "1kHz", "200Hz", "100Hz", "10Hz"]
+    )
+
+    average_sweep = Instrument.control(
+        "AVS?", "AVS %d",
+        "Number of averages to make on a sweep (1-1000)",
```
bilderbuchi, line 106: so this instrument does not take an "OFF" setting, correct?

pymeasure/instruments/anritsu/AnritsuMS9740A.py
```diff
@@ -0,0 +1,140 @@
...
+    ##########################
+
+    data_memory_select = Instrument.control(
+        "TTP?", "TTP %s",
+        "Memory Data Select.",
+        validator=strict_discrete_set,
+        values=["A", "B", "C", "D", "E", "F", "G", "H", "I", "J"]
+    )
+
+    def __init__(self, adapter, **kwargs):
```
bilderbuchi, line 128: Can you please place the __init__ at the beginning of the class? It's not really homogeneous over the codebase yet, but i try to do that for new instruments.

pymeasure/instruments/anritsu/AnritsuMS9740A.py
```diff
@@ -0,0 +1,140 @@
...
+        """Constructor."""
+        self.analysis_mode = None
+        super(AnritsuMS9740A, self).__init__(adapter, "Anritsu MS9740A Optical Spectrum Analyzer", **kwargs)
+
+    def repeat_sweep(self, **kwargs):
```
bilderbuchi, line 133: Is that a thing that the `AnritsuMS9710C` can do, too (and it was just not exposed, yet)? If so, it makes sense to place it in that class, so both can use it.

Also, as it's only 2 arguments, it's probably clearer to just repeat the args (and default arguments) that `wait_for_sweep` expects. I.e. `repeat_sweep(self, n=20, delay=0.5)`, and below `self.wait_for_sweep(n=n, delay=delay)`

bilderbuchi (changes requested):
Great work, this is looking good! Do you see how now the instrument definition is much shorter?
It could be even less because most changes are only tweaking some allowable ranges, but there are some limitations of Python properties at play that make that a bit difficult, so don't worry about that.

Have you already added yourself to the Authors.txt and docs/about/authors.rst files?

pymeasure/instruments/anritsu/AnritsuMS9740A.py
```diff
@@ -0,0 +1,140 @@
...
+    l, p = vals
+    res = [l]
+    m = r_value_units.match(p)
+    if m is not None:
+        data = list(m.groups())
+        data[0] = float(data[0])
+        res.extend(data)
+    else:
+        res.append(float(p))
+    return res
```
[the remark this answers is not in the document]
md12g12: Thanks, I didnt realize this but it makes sense - and they are not actually used so I can remove them completely

pymeasure/instruments/anritsu/AnritsuMS9740A.py
md12g12: Yes, its an annoying change but it now uses 1 in the same way that "off" used to work on the older models.

pymeasure/instruments/anritsu/AnritsuMS9740A.py
md12g12: I have moved it, however I am actually having isses with the code. I have only just got access to the instrument and when i try to call my new class I get an error in the new __init__ statement which I was hoping you could help with? When running my class locally as a stand alone (ie no parents) the __init__works fine, however when adding (AnritsuMS9710C) in the Class line, I get a "TypeError: __init_ takes 2 positional arguments but 3 were given". 
As far as I can tell it was receiving 3 arguments (namely adapter, a string and **kwargs) in the parent class and that works so im unsure what the issue is. I guess its something to do with adapter (which I have no idea what it is or where it is defined/imported)

pymeasure/instruments/anritsu/AnritsuMS9740A.py
md12g12: I'm not sure - I dont have access to the MS9710C so cant easily test. I guess to be on the safe side I would only add it here to ensure that the previous Class is not "broken"

md12g12:
Hi Chris,

I tried to reply to this on Git but my pull request has failed: "PR run
failed: PyMeasure CI, Attempt #2". Im still learning what it all means but
in the meantime can I ask for your help getting the module running as a
child?

When I place the new instrument controls in a class that has a dependency
on the AnritsuMS9710C I get an error when initiating my new class. Im sure
there is a sensible way to share code on git for you to review but im not
sure which wat is best yet. In the meantime here is my code and error if
you are able to take a look as my class works in isolation so im hoping its
an easy fix.

class AnritsuMS9740A(AnritsuMS9710C):
        self.analysis_mode = None
        super(AnritsuMS9740A, self).__init__(adapter, "Anritsu MS9740A
Optical Spectrum Analyzer", **kwargs)

The error I get is TypeError: __init__() takes 2 positional arguments but 3
were given. This error is for the line with the super().__init__
I dont know what the point of this __init__ statement is, and dont know
what the adapter / string / **kwargs really mean in this. I guessed it is
something that pymeasure defines? I will keep learning but in the meantime
if this is an easy fix please let me know.

Once this is fixed the code should be ready. Do I open a new pull request
with the complete set of changes, or can I edit the failed one and try that
again?

Many thanks,
Matt

p.s I have access to the instrument all of this week so should be able to
reply far quicker than I have in the past

md12g12:
Never mind, of course straight after sending that email I have the genius
idea to remove the string from the __init__ which I thought was weird in
the first place. It all works now so I should have a complete module to
upload. If I dont hear from you I will open a new request with a working
copy, otherwise please let me know if there is a better way.

Thanks for all your help,
Matt

bilderbuchi:
> If I dont hear from you I will open a new request with a working
copy, otherwise please let me know if there is a better way.

Real quick before I find the time to dig deeper: Just add commits to your already existing branch (`master`) and then push, and the PR will automatically be updated! [Github docs for further reading](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests)

bilderbuchi:
Also watch out, Python is case-sensitive, and your file is called `AnritsuMS9740A.py`, but you try to import it with `from .anritsuMS9740A import AnritsuMS9740A` (the first is the module/file, the second the class within, here). Please lowercase the filename.

bilderbuchi:
re: the arg problem: this is a problem in the parent class, the approach you suggest is a good workaround for the meantime, we'll get the issue straightened out properly later!

docs/api/instruments/anritsu/anritsuMS9740A.rst
```diff
@@ -0,0 +1,3 @@
+.. autoclass:: pymeasure.instruments.anritsu.AnritsuMS9740A
```
bilderbuchi, line 1:
```suggestion
#########################################
Anritsu MS9740A Optical Spectrum Analyzer
#########################################

.. autoclass:: pymeasure.instruments.anritsu.AnritsuMS9740A
```

you'll need this, too to make the warning below go away. 
I'll try to use these suggestions as much as possible, you basically only have to press an accept button somewhere here to use what I suggest.

pymeasure/instruments/anritsu/AnritsuMS9740A.py
bilderbuchi: Changing the definition of the parent class in anritsuMS9710C.py to read
```
    def __init__(self, adapter, name="Anritsu MS9710C Optical Spectrum Analyzer", **kwargs):
        """Constructor."""
        self.analysis_mode = None
        super(AnritsuMS9710C, self).__init__(adapter, name=name, **kwargs)
```

and the super call here in your daughter class init to
```
super(AnritsuMS9740A, self).__init__(adapter, name="Anritsu MS9740A Optical Spectrum Analyzer", **kwargs)
```
should do the trick, and when you construct an instance, it should have the correct ` myinstr.name` attribute.

Previously, the parent class did not expect to have its name overridden with a constructor argument, so that caused the problems you were seeing.

pymeasure/instruments/anritsu/AnritsuMS9740A.py
bilderbuchi: fair enough. still, the argument tip is unrelated to that.

pymeasure/instruments/anritsu/AnritsuMS9740A.py
bilderbuchi: maybe makes sense to document that in the docstring, for the next person scratching their head :D

bilderbuchi (changes requested):

md12g12:
@bilderbuchi I have made all the changes you suggested, not sure how to push this but I might be waiting for approval as a first time contributor?
Thanks for all your help!

bilderbuchi:
Yes the approval thing is something annoying that Github had to do because of crypto miners abusing the free resources.

One thing I don't see yet is this change in the definition of the parent class in anritsuMS9710C.py to read
```
    def __init__(self, adapter, name="Anritsu MS9710C Optical Spectrum Analyzer", **kwargs):
        """Constructor."""
        self.analysis_mode = None
        super(AnritsuMS9710C, self).__init__(adapter, name=name, **kwargs)
```
then we're done :-)

md12g12:
Ah sorry, I thought when I accepted changes that it had done it already. That should be good now. Cheers!

bilderbuchi:

PR #524, opened 2021-11-13
Keithley 2400: Add line frequency controls

mfikes:


pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -117,6 +117,23 @@ class Keithley2400(Instrument, KeithleyBuffer):
         map_values=True,
     )
 
+    line_frequency = Instrument.control(
+        ":SYST:LFR?", ":SYST:LFR %d",
+        """ An integer property that controls the line frequncy in Hertz. 
+        Valid values are 50 and 60. """,
+        validator=strict_discrete_set,
+        values={50: 50, 60: 60},
+        map_values=True,
```
mfikes, line 126: Note that a dictionary and `map_values=True` is used here instead of `values=[50, 60]` because the latter results in the property returning floating point values instead of integers.

CasperSchippers: For this you could also consider using `cast=int`

bilderbuchi:
Looks good, I'll merge.

PR #525, opened 2021-11-14
BK Precision 8500 Programmable DC Electronic Load

mfikes:
This particular instrument doesn't support SCPI, but instead accepts and returns 26-byte packets of information via a TTL DB9. The need to work in terms of raw binary makes for a bit of a challenge for this one (perhaps there are other instruments in the pipeline that encounter similar characteristics, needing to talk directly to the adapter, override methods in instrument, etc.).

### DIFF
--- a/pymeasure/instruments/bkprecision/bkprecision8500.py
+++ b/pymeasure/instruments/bkprecision/bkprecision8500.py
@@ -1,0 +1,283 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2021 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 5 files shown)


bilderbuchi:
Ah, great! In fact there were some initiatives to get frame-based communication going. See #296, #239, #437, especially [this conversation](https://github.com/pymeasure/pymeasure/pull/239#discussion_r386124016).
I want to get that in in a way that fits well into the rest of the package, but I haven't had time to think and iterate on that. :-(

I had a really quick look at the PR, and I think it's a great (re)start for the above. Great that you include tests, very appreciated!
If it's OK with you I'd like to take some time doing it right (putting us in a place to resolve #296, #239, too), which means it will probably be a while until this PR gets merged. 

Do you maybe want to take a stab at this? Considering this device and the Regatron thing from #239, maybe there is already a clear path to extract common behaviour/API that makes sense for "all"/many frame-based instruments. 
If yes, some remarks off the top of my head. I'd like to see this kind of communication method extracted to a specific `Adapter` subclass -- it should not remain with specific instruments, and it should fit into the `Adapter` API that `Instrument` uses. 
Maybe pass standardised "special" command strings or -tuples to the property factories, and the Adapter de-/encodes those and assembles frames to send out.

mfikes:
@bilderbuchi Yeah, I don' mind taking a stab at this. With #516 I encountered a non-SCPI instrument, but that one didn't present so much difficulty since it is ASCII based only sends commands. With this one, in addition to being non-SCPI, it is binary. I think I have another instrument (HP 8657A) that is non-SCPI so I might be in a good position to test such changes.

bilderbuchi:
Great, thanks!
Yeah, that's actually what I meant (sorry for being imprecise) - not SCPI-compliant (e.g. not implementing the standard commands) is fine; I consider instruments exchanging "commands" (either ascii/plaintext or binary frames/registers/arrays that one can assemble) as in scope for us. 
Where we draw the line is DLL-based communication, that is out of scope for pymeasure, it's just too different and not compatible with our maintenance resources.

bilderbuchi:
Also check [this out](https://pyvisa.readthedocs.io/en/latest/api/resources.html#pyvisa.resources.RegisterBasedResource) - might be useful. It would be fine for me if we restrict the frame-based Adapter to using pyvisa/inherits from the VISAAdapter if that makes our work easier.

pymeasure/instruments/bkprecision/bkprecision8500.py
```diff
@@ -0,0 +1,283 @@
...
+class BKPrecision8500(Instrument):
+    """ Represents the BK Precision 8500 Programmable DC Electronic Load and
+    provides a high-level interface for interacting with the instrument.
+    """
+
+    def __init__(self, resourceName, baud_rate=4800, **kwargs):
+        super().__init__(
+            resourceName,
+            "BK Precision 8500",
+            **kwargs
```
msmttchr, line 135: Being a non SCPI instrument, you need to add `includeSCPI=False`, in parameters.
I would also try to add `write_termination=""` and `read_termination=""`. This could be sufficient to handle binary values. Also you may want to check `binary_values` and `write_binary_values` in the `VISAAdapter` and `SerialAdapter` classes.

bmoneke:
@mfikes do you want to pick up this PR or may I close it? It can be reopened whenever you want to continue.

mfikes:
It can be closed.

PR #528, opened 2021-11-19
Keithley2600 | Fixed error when retrieving error code

bicarlsen:
In Keithley2600.error the error code is sometimes printed in scientific notation which raises an error when converted to an integer. This commit allows scientific notation to be used.


bilderbuchi:
Looking good, thanks!

PR #529, opened 2021-11-21
Keithley 2306 Dual Channel Battery/Charger Simulator

mfikes:


pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -0,0 +1,608 @@
...
+        values={'limit': 'LIM', 'trip': 'TRIP'},
+        map_values=True,
+    )
+
+    source_current_limit_enabled = Instrument.measurement(
+        ":SOUR<ch>:CURR:STAT?",
+        """A boolean property that returns the source current limit state. """
+    )
+
+    read = Instrument.measurement(
```
mfikes, line 339: Perhaps a better name for this property would be `reading`. There is another SCPI command that returns an array, and maybe that could be `readings`.

bilderbuchi: also `Instrument.read()` is already an pretty central method, and `read` sounds more like a verb to me, and properties are often/typically nouns/adjectives (in the spirit of command-query-separation).
(drive by comment to the comment, it will take time until I get around for a proper review.)

mfikes: I took a deeper look at the SCPI API and there are roughly 3 modes: 

1. Trigger and return a measurement based on the previously set sense mode (voltage, current, etc.), 
2. Trigger and return a measurement while also specifying the sense mode (which sets the sense mode as a side-effect)
3. Just fetch existing measurements (without triggering, or specifying sense mode)

Additionally, you can request either a single value or an array (where the number of values returned is tied to the number of values used for averaging).

So, given all of the above, I added a whole slew of properties allowing the user to obtain data in a way that fairly closely adheres to the way the SCPI API works.

lgarbarini:
@mfikes Is there anything blocking this merge? I was about to submit a pull request to upstream our internal keithley2306 instrument, but this one looks more feature complete and has tests! 

Happy to run some tests if that's what's blocking.

mfikes:
@lgarbarini I don't believe there are any outstanding issues or improvements identified by the PyMeasure maintainers. IMHO, the current PR is good for review and potential acceptance.

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -0,0 +1,692 @@
...
+        elif relay_number == 2:
+            return self.relay2
+        elif relay_number == 3:
+            return self.relay3
+        elif relay_number == 4:
+            return self.relay4
+        else:
+            raise ValueError("Invalid relay number. Must be 1, 2, 3, or 4")
+
+    def check_errors(self):
```
msmttchr, line 678: `check_errors` should be available from parent class `Instrument`, there should be no need to reimplement it here, unless you need some customization

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -0,0 +1,692 @@
...
+        """A floating point property that controls the long integration time
+        in seconds, which takes on values in the range of 0.850 for 60 Hz and
+         0.840 for 50 Hz up to 60. """,
+        validator=truncated_range,
+        values=[0.840, 60],
+    )
+
+    def long_integration_time_auto(self):
+        """Arranges for the instrument to control integration times. """
+        self.write(":SENS<ch>:LINT:TIME:AUTO")
```
msmttchr, line 242: I think you could avoid this method by using the `joined_validator`  in property `long_integration_time`

msmttchr (changes requested):
@mfikes huge and high quality contribution.
Thanks a lot.
I put two minor remarks, could you please have a look.

pymeasure/instruments/keithley/keithley2306.py
mfikes: I've removed the custom `check_errors` implementation.

I may have encountered a situation where the instrument returned an empty space as in the following, and tweaked the overriding `check_errors` to handle it.

```
:SYST:ERR?
,"Foo"
```

But perhaps I had driven the instrument itself into a bad state and a custom `check_errors` is not warranted. The unit tests don't repro this situation.

pymeasure/instruments/keithley/keithley2306.py
mfikes: If I understand the suggestion, this would allow clients write either

```python
instr.ch(1).long_integration_time = 30
```
or
```python
instr.ch(1).long_integration_time = 'AUTO'
```

where the former would result in this SCPI being issued

```
:SENS1:LINT:TIME 30
```

and the latter involving this SCPI

```
:SENS1:LINT:TIME:AUTO
```

An aspect to consider, though, is that there is a similar `pulse_current_time_auto` for automatically setting pulse current measurement parameters. Where long integration involves a single property, pulse current mode involves three distinct properties (`pulse_current_time_high`, `pulse_current_time_low`, and `pulse_current_time_average`).

So a question would be whether allowing a user to set any one of the three pulse current properties to `'AUTO'`, detecting this case, and  make it issue `:SENS<ch>:PCUR:TIME:AUTO` (thus effectively having it automatically set all three), or alternatively leaving things as-is, with methods for each "automatic" setting, adhering a bit more closely to the underlying SCPI exposed by the instrument.

msmttchr: Regarding your first comment on `long_integration_time`, yes, indeed this what I meant, but looking at your explanation, I am no longer sure that my suggestion would be applicable, since:

1. The type of the two kind of parameters is different (`str `and `float`), so I am not sure which specifier would go in `":SENS<ch>:LINT:TIME %g"`
2. The syntax is a little bit different, if the parameter is `float`, this is indeed a parameter for SCPI command `:SENS1:LINT:TIME` while in the second case is a different command `:SENS1:LINT:TIME:AUTO`

So, given the above and based on my knowledge on the code and in python, I would leave thing as they are.

Regarding the second question (`pulse_current_time_auto`, sorry I did not spot this), I definitively agree to leave the things as they are.

mfikes:
Cool, at this point I have nothing else I think can be done for this PR and it is ready for review and potential acceptance.

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -0,0 +1,676 @@
...
+        """A floating point property that controls the source voltage protection 
+        offset in volts, which takes values between 0 and 8. """,
+        validator=truncated_range,
+        values=[0, 8],
+    )
+
+    source_voltage_protection_enabled = Instrument.measurement(
+        ":SOUR<ch>:VOLT:PROT:STAT?",
+        """A boolean property that returns the source voltage protection state. """
+    )
```
msmttchr, line 306: Is this returning a boolean ? I think you could pass the parameter `cast=bool` to make a casting.
How do you enable/disable the voltage protection ?
Similar comment for `source_current_limit_enabled`.

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -0,0 +1,676 @@
...
+    source_current_limit_enabled = Instrument.measurement(
+        ":SOUR<ch>:CURR:STAT?",
+        """A boolean property that returns the source current limit state. """
+    )
+
+    last_reading = Instrument.measurement(
+        ":FETCH<ch>?",
+        """A floating point property that returns the last reading. """,
+        get_process=lambda v: float(v)
+    )
```
msmttchr, line 343: In this situation, you could use the keyword parameter `cast=float` which is anyway the default, so the `get_process` parameter can be removed altogether. I must admit though that is very difficult to infer this and I know since I have seen it used in other instruments definition. Have a look at https://github.com/pymeasure/pymeasure/blob/029f8e3a6cba18730692094e64c7cbc455456b36/pymeasure/adapters/adapter.py#L75

Same comment for other similar  casting needs.

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -0,0 +1,676 @@
...
+
+    last_readings = Instrument.measurement(
+        ":FETCH<ch>:ARR?",
+        """A floating point array property that returns the last readings. """,
+        get_process=lambda v: v if type(v) == list else [v]
+    )
```
msmttchr, line 349: I think in python `isinstance` is preferred over `type` due to the fact that can understand also inheritance.
Same comment for other similar comments below.

msmttchr (commented):
@mfikes, I had some more time to have a second look at your code and I have few other comments/clarifications. Please note there are also some lint warning and errors, could You please check and fix them ?

pymeasure/instruments/keithley/keithley2306.py
mfikes: Good catch... it is not:

```python
INSTR.ch(1).source_voltage_protection_enabled
0.0
```

Will add a test and fix.

pymeasure/instruments/keithley/keithley2306.py
mfikes: Removed unnecessary `get_process` lambdas that do nothing more than cast to `float`.

pymeasure/instruments/keithley/keithley2306.py
mfikes: Changed `type` usages to `isinstance`.

bilderbuchi:
Sorry, I'm totally blocked ATM, will take some time to be able to review this. :-/

pymeasure/instruments/keithley/keithley2306.py
msmttchr: @mfikes looking a the instrument manual, I understand this property is probably set when voltage protection has triggered. I infer this from user manual notes on the other similar property `source_current_limit_enabled  ` which allow to understand if the current limit has been reached.
Initially I thought, looking at doc string, these properties were indicating whether the user has enabled or not the specific protection.
Am I correct in this ? Do you have the same understanding ?

mfikes: @msmttchr Indeed, this is essentially a property that evaluates to `True` when voltage protection has triggered. (And in that case the voltage should have been shut off by the instrument.) And indeed, this—and the analogous current limit property—are not settable by the user. 

Unfortunately, for `source_current_limit_enabled`, the explanation is a little more complex because when it evaluates to `True` this could mean that the source is actually being current-limited or the source has been disabled (the current protection has _tripped_, in Keithley's terminology). Given this, the name of the property feels less than ideal. In the SCPI portion of the manual, the full command is `SOURCE:CURRENT:LIMIT:STATE?` where `1` means that the output is in limit mode or has tripped, depending on what mode the user put the device in.

So, I agree, these properties names and/or doc strings could be improved. I looked around for other PSU instruments in PyMeasure and didn't find anything like this being exposed, so perhaps this is a chance to set a good precedent.

Does the use of `_enabled` in a property name imply something that can be set (with some other naming being used for properties that are read-only)? (I had only chosen `_enabled` to go along with the fact that the property is Boolean-valued, no more rationale than that.)

In any case, it seems easy to expand upon the doc string to further clarify the meanings of each of these properties, perhaps something along these lines: 

Voltage:

"A boolean property that returns the source voltage protection state. If this property is True, the source has been shut off in accordance with the source voltage protection settings. If this property is False, the source has not been shut off due to voltage protection."

Current:

"A boolean property that returns the source current limit state. If this property is True, the source is in either in current limit mode, or has tripped (shut off), based on the source_current_limit_type setting. If this property is False, the source is not being limited and has not been tripped."

msmttchr: @mfikes I am not aware of any convention for _enabled suffix, so I would not suggest to make any name change.
Regarding the clarification on the doc string, I agree with your proposal.

mfikes: Doc strings updated.

bilderbuchi: > Does the use of _enabled in a property name imply something that can be set (with some other naming being used for properties that are read-only)? (I had only chosen _enabled to go along with the fact that the property is Boolean-valued, no more rationale than that.)

There is no rationale for naming indicating settability -- your chosen rationale was correct. In the end, we're happy if people choose clear names, and adjective-like names for boolean properties typically fit well and are clear (if you imagine them being a question "is power enabled?" or command "make power enabled")

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -0,0 +1,683 @@
...
+        is enabled and the current range is set to 5 A. Takes values 'HIGH' or
+        'LOW'. If the output is disabled or the current range is set to 5 mA the
+        bandwidth is 'LOW'. """,
+        validator=strict_discrete_set,
+        values={'low': 'LOW', 'high': 'HIGH'},
+        map_values=True,
+    )
+
+    sense_mode = Instrument.control(
+        ":SENS<ch>:FUNC?", ":SENS<ch>:FUNC \"%s\"",
```
bilderbuchi, line 57: Just a Python detail you are maybe already aware of anyway, you can mix and match quotes in strings, so the second string here does not need escapes, `':SENS<ch>:FUNC "%s"'` should work, too.

bilderbuchi (approved):
Thank you for the high-quality contribution, another complex Keithley instrument with a large API surface. :yum: 
I love that you went the extra mile to implement a test suite!
Looking very tidy and good to me.

bilderbuchi:
Thank you @msmttchr for the previous review loop! We can merge this now.

PR #534, opened 2021-11-30
Heidenhain ND287 Position Display Unit Driver

samcondon4:
This PR implements a basic driver for the Heidenhain ND287 Position Display Unit. This driver is pretty basic, though I would like to hear feedback on the `id` and `error` properties. Both of the queries that these properties trigger return strings with several lines of characters. That is, the instrument returns several carriage returns and line feeds in the return string. By just creating a normal property via `Instrument.measurement`, the read operation is terminated on the first carriage return, so the whole string is not retrieved. To account for this I used the low-level `adapter.connection.write()` and `adapter.connection.read_bytes()` methods, though I am not sure that this is the cleanest implementation. Do any of the reviewers have suggestions of better ways to implement these properties? Perhaps a few additions to the `Instrument` base class are in order to accomplish this kind of query in a cleaner way?

Thanks in advance!


docs/api/instruments/heidenhain/index.rst
```diff
@@ -0,0 +1,14 @@
+.. module:: pymeasure.instruments.heidenhain
+
```
bilderbuchi, line 2: this warning above means that you need to add a reference in  docs/api/instruments/index.rst

pymeasure/instruments/heidenhain/nd287.py
```diff
@@ -0,0 +1,118 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import logging
+from pyvisa.errors import VisaIOError
+from pymeasure.instruments import Instrument
+
+
+class Nd287(Instrument):
```
bilderbuchi, line 30: You can call it ND287

pymeasure/instruments/heidenhain/nd287.py
```diff
@@ -0,0 +1,118 @@
...
+        """ Apply the appropriate scaling factor to the position read-out from the encoder based on the value of the
+            units property.
+        """
+        if cls._units == "mm":
+            pos *= 1e-4
+        elif cls._units == "inch":
+            pos *= 1e-5
+        return pos
+    
+    def __init__(self, resourceName, **kwargs):
```
bilderbuchi, line 59: please move the init to the top of the class

pymeasure/instruments/heidenhain/nd287.py
```diff
@@ -0,0 +1,118 @@
...
+    # additional configuration properties that cannot be controlled through software, but should be configurable
+    # to match the values on the local device. #
+    _units = "mm"
+    
+    @classmethod
+    def position_proc(cls, pos):
+        """ Apply the appropriate scaling factor to the position read-out from the encoder based on the value of the
+            units property.
+        """
+        if cls._units == "mm":
```
bilderbuchi, line 53: Have you tested that this works correctly, i.e. you can switch units after creating an instance by using self.units?

Also, be aware of the caveat that, if you have more than one ND287 at the same time, I think they then can't have different settings. A proper fix of this will be (WIP, but slowly), see #250.

pymeasure/instruments/heidenhain/nd287.py
```diff
@@ -0,0 +1,118 @@
...
+    @property
+    def id(self):
+        """ String identification property for the device.
+        """
+        self.adapter.connection.write("\x1BA0000")
+        id_str = self.adapter.connection.read_bytes(36).decode("utf-8")
+        return id_str
+
+    @property
+    def error(self):
```
bilderbuchi, line 80: do you need this property? Is it not better to pull this into the `check_errors` method?

pymeasure/instruments/heidenhain/nd287.py
```diff
@@ -0,0 +1,118 @@
...
+    def check_errors(self):
+        """ Method to read an error status message and log when an error is detected.
+
+        :return: String with the error message as its contents.
+        """
+        err_msg = self.error
+        if err_msg is not None:
+            logging.error("Heidenhain ND287 error message received: %s" % err_msg)
+        return err_msg
+
```
bilderbuchi, line 118: v there is an empty file  pymeasure/instruments/nd287.py shown below - please delete.

pymeasure/instruments/heidenhain/nd287.py
```diff
@@ -0,0 +1,118 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import logging
+from pyvisa.errors import VisaIOError
+from pymeasure.instruments import Instrument
+
```
bilderbuchi, line 28:
```suggestion
import logging
from pyvisa.errors import VisaIOError
from pymeasure.instruments import Instrument

log = logging.getLogger(__name__)
log.addHandler(logging.NullHandler())
```
then use `log` instead of `logging` below.

bilderbuchi (changes requested):
Looking good, some small things, really.

docs/api/instruments/heidenhain/index.rst
samcondon4: Oops. Fixed that.

pymeasure/instruments/heidenhain/nd287.py
samcondon4: Got it.

pymeasure/instruments/heidenhain/nd287.py
samcondon4: I did test this and it worked (at least for one device). I believe that you are correct though that multiple ND287 devices could not have different settings. To address this, I've implemented the `position` property manually so that it can query the instance `_units` attribute to apply the appropriate scale factor. I will keep an eye on #250 though and perhaps this property can be updated when #250 is accepted.

pymeasure/instruments/heidenhain/nd287.py
samcondon4: Nope, I think it is better to pull it into `check_errors`. The latest commit addresses that.

pymeasure/instruments/heidenhain/nd287.py
samcondon4: Not sure how I did that. It has been removed.

pymeasure/instruments/heidenhain/nd287.py
samcondon4: Got it.

pymeasure/instruments/heidenhain/nd287.py
bilderbuchi: A solution for #250 is nearing completion now, so I'd wait until #376 is merged, and then use the new functionality here.

bilderbuchi: In the meantime, we have activated a linter, so there are a couple of linting reports in the "Files changed" tab, which you could fix when you have an opportunity.

bilderbuchi: Great to hear! A couple of linter issues have reappeared -- you can run flake8 locally (`flake8 .`) to check for issues (if your editor does not do that automatically)

samcondon4: Do you still see the linter issues on your end? I believe my latest commit has addressed them. I ran flake8 locally and don't see any issues in the remote or local checks on my end anymore.

bilderbuchi: Hm, I guess some UI state in the browser must have been stale? Anyway, nothing to see anymore, thanks!

PR #535, opened 2021-12-02
new Instruments: Andeen Hagerling capacitance bridges

dkriegner:
This PR implements two precision capacitance bridges used in our lab. Devices are typically connected via GPIB or serial connection.

The devices are rather trivial and provide only two important readout values (the capacity and the connected loss).

Note that we tested only the GPIB communication, but except of the obvious serial communication settings (which a user would need to provide) the serial version should work equally with this code.


pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,76 @@
...
+#
+
+from pymeasure.instruments import Instrument
+from .util import parse_reply
+
+
+class AH2500A(Instrument):
+    """ Andeen Hagerling 2500A Precision Capacitance Bridge implementation
+    """
+    id = "Andeen Hagerling AH2500A"  # device does not support '*IDN?'
```
LongnoseRob, line 32: if this instrument does not offer SCPI capabilities, it is recommended to add `includeSCPI=False,` to the `super(..).__init__` parameters

pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,76 @@
...
+        get_process=parse_reply
+    )
+
+    def __init__(self, adapter, **kwargs):
+        if "timeout" not in kwargs:
+            kwargs["timeout"] = 3000
+        if "write_termination" not in kwargs:
+            kwargs["write_termination"] = "\n"
+        if "read_termination" not in kwargs:
+            kwargs["read_termination"] = "\n"
```
LongnoseRob, line 52: I like the idea of checking if these parameters are set.
However if these parameters are always required (e.g. for RS232 and GPIB communication), int hat case I would add them to the parameter list ofthe `super(...).__init__()`

LongnoseRob (commented):
Hello and gthank you for adding  these interesting instruments.
I never had head of this company or these instruments before.

I have left a few comments which could be the start of a discussion for further improvement of the implementations.

Do these istruments store the last operation conditions like measurement-frequency and mode?
I did not see any methods (yet) to set e.g. the measurement frequency.

pymeasure/instruments/andeenhagerling/util.py
```diff
@@ -0,0 +1,43 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import math
+import re
+
+mcl = re.compile("[FHZ0-9.=\s]*C=\s*(-?[0-9.]+)\s*PF L=\s*(-?[0-9.]+)\s*NS")
```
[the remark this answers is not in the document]
dkriegner: likely I should use a raw r"" string here...

pymeasure/instruments/andeenhagerling/ah2700a.py
```diff
@@ -0,0 +1,52 @@
...
+            kwargs["timeout"] = 5000
+        super(AH2700A, self).__init__(adapter, **kwargs)
+
+    def trigger(self):
+        """
+        Triggers a new measurement without blocking and waiting for the return
+        value.
+        """
+        self._triggered = True
+        self.write("*TRG")
```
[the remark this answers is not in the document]
dkriegner: The trigger command is different for the two. 
One uses "TRG" (AH2500A), The other uses "*TRG" (AH2700A).

dkriegner:
Thanks for the comments. I'll look into them and will make according changes.

pymeasure/instruments/andeenhagerling/util.py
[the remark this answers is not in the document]
bilderbuchi: yes, using a raw string for regexes typically makes it more readable.

pymeasure/instruments/andeenhagerling/ah2700a.py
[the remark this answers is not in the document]
LongnoseRob: Oh my bad, did not see the small difference.

pymeasure/instruments/andeenhagerling/ah2500a.py
bilderbuchi: you can also use the builtin `dict.setdefault()` method.

pymeasure/instruments/andeenhagerling/util.py
```diff
@@ -0,0 +1,43 @@
...
+
+def parse_reply(string):
```
bilderbuchi, line 30: why not have these as `_parse_reply` methods with the respective classes, then you could use class-specific regexes? Or is this not a big bother?

pymeasure/instruments/andeenhagerling/ah2700a.py
bilderbuchi, line 52: might be worth it to reverse the order of these two - if something goes wrong during the write, you otherwise end up with a _triggered==True state that is incorrect. (same above)

pymeasure/instruments/andeenhagerling/ah2700a.py
```diff
@@ -0,0 +1,52 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Instrument
+from .ah2500a import AH2500A
+
+
+class AH2700A(AH2500A):
```
bilderbuchi, line 29: Nice that you used inheritance to eliminate duplication!

pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,76 @@
...
+    def triggered_caploss(self):
+        """
+        reads the measurement value after the device was triggered by the
+        trigger function.
+        """
+        if self._triggered:
+            self._triggered = False
+            return parse_reply(self.read())
+        else:
+            raise Exception("Device was not triggered previously!")
```
bilderbuchi, line 76: Is that a harsh error condition (worth raising an exception) to not have triggered before? Otherwise, you could do the more forgiving
```
        if not self._triggered:
            self.trigger()  # maybe log a complaint :D
        return parse_reply(self.read())
```

bilderbuchi (changes requested):
some comments from my side.

pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, adapter, **kwargs):
+        if "timeout" not in kwargs:
+            kwargs["timeout"] = 3000
+        if "write_termination" not in kwargs:
+            kwargs["write_termination"] = "\n"
+        if "read_termination" not in kwargs:
+            kwargs["read_termination"] = "\n"
+        super(AH2500A, self).__init__(
+            adapter, "Andeen Hagerling 2x00A Precision Capacitance Bridge",
```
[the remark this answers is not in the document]
dkriegner: I think your suggestion won't work as suggested because `name` is not a keyword argument of `Instrument`. But I adopted an equal idea from `FakeInstrument`.

pymeasure/instruments/andeenhagerling/util.py
dkriegner: I agree. the util.py file should be removed and replaced by a classmethod

docs/api/instruments/andeenhagerling/ah2700a.rst
```diff
@@ -0,0 +1,7 @@
+###########################################
+Andeen Hagerling AH2700A capacitance bridge
+###########################################
+
+.. autoclass:: pymeasure.instruments.andeenhagerling.AH2700A
+    :members:
+    :show-inheritance:
```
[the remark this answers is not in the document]
dkriegner: This should be fixed in c1cf9e0 by adding `":inherited-members: Instrument"` which adds members of `AH2500A` but not the ones of Instrument. From `Instrument` it might be interesting to add only some selected ones like `id`. I see that in some devices you show all inherited members and then exclude a list of those which one does not want. not sure how you propose to handle this here?

pymeasure/instruments/andeenhagerling/util.py
```diff
@@ -0,0 +1,43 @@
...
+    :param string: reply string from the instrument. This commonly could be
+      2500A: "C=123.123456 PF L=0.12345 NS"
+      2700A: "F= 1000.00 HZ C= 4.20188     PF L=-0.0260      NS V= 15.0     V"
+    """
+    m = mcl.match(string)
+    if m is not None:
+        cap, loss = map(float, m.groups())
+        return cap, loss
+    # if an invalid string is returned ('EXCESS NOISE')
+    return math.nan, math.nan
```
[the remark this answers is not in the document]
dkriegner: I implemented the logging, but would prefer not to abort the measurement right away by an exception.

pymeasure/instruments/andeenhagerling/ah2500a.py
dkriegner: I decided to go for the `dict.setdefault` options. I think its most readable and I also added the properties to the `__init__` call expression so that the options are visible in the documentation.

pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,76 @@
...
+    config = Instrument.measurement(
+        "SHOW",
+        """ Read out configuration """,
+    )
+
+    caploss = Instrument.measurement(
+        "Q",
+        """ Perform a single capacitance, loss measurement and return the
+        values in units of pF and nS.""",
+        get_process=parse_reply
```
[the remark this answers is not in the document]
dkriegner: done

pymeasure/instruments/andeenhagerling/ah2700a.py
dkriegner: true this was a logical error. its now reversed

pymeasure/instruments/andeenhagerling/ah2500a.py
dkriegner: I implemented the forgiving version in c1cf9e0

dkriegner:
I still need to look into the other scpi properties issue and implement some of the other properties. 
I'll re-request a code review afterwards

docs/api/instruments/andeenhagerling/ah2700a.rst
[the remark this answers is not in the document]
LongnoseRob: I think I would do it depening on which number is smaller: the of the elements to exclude or the of the elements to include.

pymeasure/instruments/andeenhagerling/ah2500a.py
dkriegner: I decided to include `includeSCPI=False` for both instruments since even the newer one only supports a small subset of the SCPI commands implemented in `Instrument`. I therefore reimplemented the `id` property and a `reset` function. Let me know if this is ok for you.

pymeasure/instruments/andeenhagerling/util.py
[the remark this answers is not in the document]
dkriegner: I now explicitly test the return string for the `EXCESS NOISE` message and only produce the warning in this case. If the string can't be parsed for other reasons an Exception is raised. see 0b0dfb3

dkriegner:
This is ready for another review round. I hope I addressed all raised issues. Not sure if I always found the best solution, but you will let me know! :)

I left those comments which I think could remain controversial as unresolved.

docs/api/instruments/andeenhagerling/ah2700a.rst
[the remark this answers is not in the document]
dkriegner: with the latest changes (of "includeSCPI=False") this seems to be solved without further changes. things from AH2500A are shown and this should cover all necessary things.

pymeasure/instruments/andeenhagerling/ah2500a.py
bilderbuchi: Sounds good.

pymeasure/instruments/andeenhagerling/ah2500a.py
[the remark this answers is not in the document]
bilderbuchi: > I think your suggestion won't work as suggested because name is not a keyword argument of Instrument

d'oh, indeed!

bilderbuchi:
> Could you specify what is the preferred order of 
* class variables
* `__init__`
* properties (`Instrument.control/measurement/setting`)
* other instance/class/static methods

-> that is a good order!

pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,121 @@
...
+    def _parse_reply(cls, string):
+        """
+        parse reply string from Andeen Hagerling capacitance bridges.
+
+        :param string: reply string from the instrument. This commonly is:
+          2500A:
+            "C= 1.234567    PF L= 0.000014    NS V= 0.750   V"
+          2700A:
+            "F= 1000.00 HZ C= 4.20188     PF L=-0.0260      NS V= 15.0     V"
+        """
```
bilderbuchi, line 52:
```suggestion
            "F= 1000.00 HZ C= 4.20188     PF L=-0.0260      NS V= 15.0     V"
        :returns: tuple with C, L, V values
        """
```
?

pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,121 @@
...
+        m = cls._reclv.match(string)
+        if m is not None:
+            values = tuple(map(float, m.groups()))
+            return values
+        # if an invalid string is returned ('EXCESS NOISE')
+        if string.strip() == "EXCESS NOISE":
+            log.warning("Excess noise, check your experiment setup")
+            return (math.nan, math.nan, math.nan)
+        else:  # some unknown return string (e.g. misconfigured units)
+            raise Exception("Measurement value not understood")
```
bilderbuchi, line 62: Does 
```suggestion
            raise Exception(f'Returned string "{string}" could not be parsed')
```
work better?

pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,121 @@
...
+        "SHOW",
+        """ Read out configuration """,
+    )
+
+    caplossvolt = Instrument.measurement(
+        "Q",
+        """ Perform a single capacitance, loss measurement and return the
+        values in units of pF and nS. The used measurement voltage is returned
+        as third value.""",
+        get_process=lambda v: AH2500A._parse_reply(v),
```
bilderbuchi, line 74:
```suggestion
        get_process=AH2500A._parse_reply,
```
is a bit simpler. This is already a callable, no need to wrap in a lambda

pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, adapter, name=None, timeout=3000,
+                 write_termination="\n", read_termination="\n",
+                 **kwargs):
+        kwargs.setdefault("write_termination", write_termination)
+        kwargs.setdefault("read_termination", read_termination)
+        kwargs.setdefault("timeout", timeout)
+        kwargs.setdefault("includeSCPI", False)
+        super(AH2500A, self).__init__(
+            adapter,
+            name or "Andeen Hagerling 2500A Precision Capacitance Bridge",
```
bilderbuchi, line 98: :+1:

pymeasure/instruments/andeenhagerling/ah2500a.py
```diff
@@ -0,0 +1,121 @@
...
+        # "VOLTAGE HIGHEST 1.00    V"
+        get_process=lambda v: float(AH2500A._renumeric.search(v).group(0)),
+    )
+
+    def __init__(self, adapter, name=None, timeout=3000,
+                 write_termination="\n", read_termination="\n",
+                 **kwargs):
+        kwargs.setdefault("write_termination", write_termination)
+        kwargs.setdefault("read_termination", read_termination)
+        kwargs.setdefault("timeout", timeout)
```
bilderbuchi, line 94: Actually, you don't need to do this if the arguments are present in the signature already. Just do below
```
        super(AH2500A, self).__init__(
            adapter,
            name or "Andeen Hagerling 2500A Precision Capacitance Bridge",
            write_termination=write_termination,
            read_termination=read_termination,
            timeout=timeout,
            **kwargs,
```
not a big deal, though, either way works.

pymeasure/instruments/andeenhagerling/ah2700a.py
```diff
@@ -0,0 +1,71 @@
...
+        get_process=lambda v: float(AH2500A._renumeric.search(v).group(0)),
+    )
+
+    def __init__(self, adapter, timeout=5000, **kwargs):
+        super(AH2700A, self).__init__(
+            adapter,
+            name="Andeen Hagerling 2700A Precision Capacitance Bridge",
+            timeout=timeout,
+            **kwargs
+        )
```
bilderbuchi, line 59: Actually, _here_ it's a good idea to use kwargs.setdefault() or give `name` as an explicit __init__ argument with default. The idea is that users can then give custom names to their instruments, which can be useful if e.g. you have 2 identical devices in a lab setup and want to differentiate between them.
```suggestion
    def __init__(self, adapter, name="Andeen Hagerling 2700A Precision Capacitance Bridge", timeout=5000, **kwargs):
        super(AH2700A, self).__init__(
            adapter,
            name=name,
            timeout=timeout,
            **kwargs
        )
```

bilderbuchi (changes requested):
Final round I think, a couple of small remarks.

pymeasure/instruments/andeenhagerling/ah2500a.py
dkriegner: as you see in the CI runs below your suggested code fails. I think I also tested this before and it does not work because you can't use a class method here before the class definition is finished... Maybe there is a different way this can be written? I will revert back tot he lambda function in the meantime

pymeasure/instruments/andeenhagerling/ah2500a.py
dkriegner: i agree, i went with your suggestion

dkriegner:
I hope I have addressed the final remarks and this is ready for merging. If you see another option to avoid this "ugly"-lambda function I would be certainly interested.

pymeasure/instruments/andeenhagerling/ah2500a.py
bilderbuchi: Yeah, sorry for not catching this -- sometimes coding "blind" and just making suggestions without trying them does not work. :-P
It's fine as it is. There are a couple of possibilities to do this differently, but they are not more elegant.

bilderbuchi:
LGTM, thanks!

Issue #531, opened 2021-11-24
Python 3.10: ManagedWindow TypeError


Mommes:
Hi all,
I am trying to use pymeasure for the first time. But when running the example "ManagedWindow" I get the following error message as soon as I press the "queue" button in the GUI and nothing will be plotted. 

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
C:\Program Files\Python 3.10\lib\site-packages\pymeasure\display\manager.py in _update_progress(self, progress)
    158     def _update_progress(self, progress):
    159         if self.is_running():
--> 160             self._running_experiment.browser_item.setProgress(progress)
    161 
    162     def _update_status(self, status):

C:\Program Files\Python 3.10\lib\site-packages\pymeasure\display\browser.py in setProgress(self, progress)
     73 
     74     def setProgress(self, progress):
---> 75         self.progressbar.setValue(progress)
     76 
     77 class Browser(QtGui.QTreeWidget):

TypeError: setValue(self, int): argument 1 has unexpected type 'float'---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
C:\Program Files\Python 3.10\lib\site-packages\pymeasure\display\manager.py in _finish(self)
    253         experiment = self._running_experiment
    254         self._clean_up()
--> 255         experiment.browser_item.setProgress(100.)
    256         experiment.curve.update()
    257         self.finished.emit(experiment)

C:\Program Files\Python 3.10\lib\site-packages\pymeasure\display\browser.py in setProgress(self, progress)
     73 
     74     def setProgress(self, progress):
---> 75         self.progressbar.setValue(progress)
     76 
     77 class Browser(QtGui.QTreeWidget):

TypeError: setValue(self, int): argument 1 has unexpected type 'float'An exception has occurred, use %tb to see the full traceback.
```

I'm using Windows and Python 3.10.

Do you have an idea what is wrong? Probably something with the data type of progress. But what should I change? 
Thanks for your help.

msmttchr:
Are you using PyQt or PySide2 as Qt bindings ? Which version ?

Mommes:
Thank your for help.
I'm using PyQT5.

msmttchr:
It seems that the `setValue` method of `QProgressBar()` refuse to accept `float` value. This is indeed very strange since, it should be automatically converted to `int`. I could suggest to have a look at link:
[...]

PR #538, opened 2021-12-08
Bump pytqtgraph minimum version, avoid implicit conversion warnings.

bilderbuchi:
This avoids a deprecation warning "DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python. c.setHsv(h, sat, v)"


bilderbuchi:
Note: This will not solve the problem reported in #531

bilderbuchi:
This requires Python>3.6, I'll do this in a separate PR with a larger CI update.

PR #540, opened 2021-12-08
Update CI & annotations, remove Python 3.6

### LINKED ISSUE #346: Start using linting
With the new CI setup, we should start using linting to keep a clean code base.
I propose using flake8 and pydocstyle for a reasonable set of checks. 

Default configuration shall be set via config file (probably setup.cfg is the most universal), so that users can already make use of that locally during coding.

We need to define a reasonable set of defaults so that linting is not too noisy. I will analyse the code base to identify good settings and then clean up the codebase in one commit.

One possible improvement that i can still do is to use [Problem matchers](https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md) to tease out warnings from sphinx and pytest to make them available in the [annotations view](https://github.com/ralph-group/pymeasure/actions/runs/322901274), so that people can easily discover them without going through logs. 

Note to self: There is an action for flake8 checking available, should we want that in the future. (we need to define and clean up linting for the repo first): Problem match for flake8 [here](https://github.com/home-assistant/core/blob/dev/.github/workflows/matchers/flake8.json).
Also, this: https://github.com/marketplace/actions/flake8-with-annotations or https://github.com/marketplace/actions/flake8-annotator

Note: the high number of warnings is due to some bugs in the setup-miniconda action integration that should mostly disappear with the next release.

bilderbuchi:
There's some maintenance needed on the CI configuration.

Don't fetch this branch, it will be rebased.

* [x] fix remaining 270 linter reports in instruments/
* [x] write docs about how to use flake8 locally, and how to deal with CI reports in PRs
* [x] manually cross-check pytest logs for warnings

Closes #346


bilderbuchi:
@CasperSchippers @msmttchr the linter finds the `Experiment.pcolor()` and `Experiment.update_pcolor()` methods (see Changed Files view). These are obviously broken, so cannot have been used in the last years?! Is it OK to just remove them (and associated code in `Experiment.update_plot()`?

CasperSchippers:
I'm inclined to say these can be safely removed indeed; I must say I find it a bit strange however, that it seems that from the moment these methods were committed, they were already broken in the same way as they are now.

bilderbuchi:
Haha, I also fixed some invalid escapes that seem to have tripped up the linter, now the dam is broken and I got 2k lines of warnings from flake8 :rofl: 
I'll try to come up with a sensible flake8 config and then fix a bunch.

bilderbuchi:
I'm already anticipating that with these linter changes a lot of PRs will get merge conflicts, but I think we should bite the bullet because I don't see a situation happening where we don't have a lot of PRs open. 
I'll just do my best fixing as many conflicted PRs as possible in/over the Christmas break.

bilderbuchi:
@CasperSchippers in d76de7b I refactored `DSP7265.buffer_to_float()` to bring its complexity down, can you take a look if that solution looks alright to you?

pymeasure/instruments/signalrecovery/dsp7265.py
```diff
@@ -556,41 +558,29 @@ def buffer_to_float(self, buffer_data, sensitivity=None, sensitivity2=None, rais
...
         # conversion for, adc1, adc2, dac1, dac2, ratio, and log ratio
-        for key in ["adc1", "adc2", "dac1", "dac2", "ratio", "log ratio"]:
-            if key in buffer_data:
-                data[key] = buffer_data[key] / 1000
+        convert_if_present(["adc1", "adc2", "dac1", "dac2", "ratio", "log ratio"], 1 / 1000)
 
         # adc3 (integrating converter); requires a call to adc3_time
-        if "adc3" in buffer_data:
-            data["adc3"] = buffer_data["adc3"] / (50000 * self.adc3_time)
+        convert_if_present(["adc3"], 1 / (50000 * self.adc3_time))
```
CasperSchippers, line 569: Downside of this approach is that a call to adc3_time (and hence communication with the instrument) is now always executed, even if adc3 not present.

CasperSchippers (approved):
> I refactored `DSP7265.buffer_to_float()` to bring its complexity down, can you take a look if that solution looks alright to you?

It looks quite ok to me (as far as I can tell from here, don't have a DSP7265 available at this moment), I only have a minor point with converting adc3.

PS What a tremendous amount of work; incredible job, @bilderbuchi!!

pymeasure/instruments/signalrecovery/dsp7265.py
bilderbuchi: Good point, I undid that.

bilderbuchi:
@msmttchr @CasperSchippers this is now finished despite the scope explosion of correcting all linter reports :D 
I decided to put python 3.10 support into another PR.

Do you want to review this? It's a huge amount of changes, but largely mechanistic or automated (with autopep8). Otherwise I'll just merge it as-is. It probably makes most sense to look at the commit history and the diff view [with whitespace changes disabled](https://github.com/pymeasure/pymeasure/pull/540/files?diff=unified&w=1).

pymeasure/instruments/signalrecovery/dsp7265.py
CasperSchippers: Yes, I will!

CasperSchippers (approved):
I've had a quick look, looks very good to me.

One small thing: there're two annotations in the tests/test_log.py file.

pymeasure/display/widgets.py
```diff
@@ -484,7 +490,7 @@ def _setup_visibility_groups(self):
 
                 if isinstance(getattr(self, group_name), BooleanInput):
                     # Adjust the boolean condition to a condition suitable for a checkbox
-                    condition = QtCore.Qt.CheckState.Checked if condition else QtCore.Qt.CheckState.Unchecked
+                    condition = QtCore.Qt.CheckState.Checked if condition else QtCore.Qt.CheckState.Unchecked  # noqa
```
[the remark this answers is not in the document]
bilderbuchi: Good catch! Indeed it does, I'll make this more specific.

bilderbuchi:
> two annotations in the tests/test_log.py file.

no idea how I missed that :grimacing:

bilderbuchi:
OK, I'll merge this, then, and fix the fallout. :grin:

msmttchr:
> @msmttchr @CasperSchippers this is now finished despite the scope explosion of correcting all linter reports :D I decided to put python 3.10 support into another PR.
> 
> Do you want to review this? Otherwise I'll just merge it as-is. It's a huge amount of changes, but largely mechanistic or automated (with autopep8). It probably makes most sense to look at the commit history and spot-check, and look at the diff view [with whitespace changes disabled](https://github.com/pymeasure/pymeasure/pull/540/files?diff=unified&w=1).

@bilderbuchi, very huge amount of work, unfortunately I cannot give any valuable feedback since I am not familiar at all with CI in GitHub. Nevertheless congratulations for the big effort.
I think it is also right to have the support for 3.10 in a separate PR, since I remember there are other issues with pyqtgraph and PySide2.

pymeasure/instruments/signalrecovery/dsp7265.py
CasperSchippers: > Good point, I undid that. Maybe when you are with the instrument again you can do a quick check?

I just did a test with the instrument and as far as I can tell right now it works perfectly fine.

PR #541, opened 2021-12-10
fix Attocube ANC300 with new SCPI Instrument properties

dkriegner:
in commit bbc0354 the not implemented SCPI properties were changed to emit an error. The properties of an instrument can therefore not be browsed anymore as before which made a code change in the Attocube code necessary.

Likely this iterating over all attributes (`for attr in dir(self)`) was never good as it was implemented before.


bilderbuchi:
Looks straightforward enough, thank you!

PR #545, opened 2021-12-14
Add procedure flag for start and end of sequence

jmittelstaedt:
Hi,

This is a minor change to the process of queuing a sequence which adds a flag to the created `procedure` which tells the procedure whether it is the first or last procedure in a sequence. 

I had independently created the same functionality (although in a much cruder way) that was added with the sequencers, and found that often there is either special setup needed for the instruments at the beginning of a sequence or some shutdown steps which only need to be done at the end of a sequence. As I don't believe there is a way to tell from within a `procedure` whether there are more queued, adding this flag is an easy workaround.

I suppose there is the risk of a name conflict with a user-defined `Parameter`, but the names `first_in_sequence` and `last_in_sequence` seem like unlikely `Parameter` names.

Please let me know if there is an existing way to do this or if there is a possibly better way.

### DIFF
--- a/pymeasure/display/widgets.py
+++ b/pymeasure/display/widgets.py
@@ -899,14 +899,12 @@ class SequencerWidget(QtGui.QWidget):
 
                 procedure = self._parent.make_procedure()
                 procedure.set_parameters(parameters)
-                
                 procedure.first_in_sequence = False
                 procedure.last_in_sequence = False
                 if i == 0:
-                    procedure.last_in_sequence = True
+                    procedure.first_in_sequence = True
                 if i == len(sequence)-1:
                     procedure.last_in_sequence = True
-                
                 self._parent.queue(procedure=procedure)
 
         finally:
--- a/pymeasure/display/widgets.py
+++ b/pymeasure/display/widgets.py
@@ -893,12 +893,20 @@ class SequencerWidget(QtGui.QWidget):
                 "Queuing %d measurements based on the entered sequences." % len(sequence)
             )
 
-            for entry in sequence:
+            for i, entry in enumerate(sequence):
                 QtGui.QApplication.processEvents()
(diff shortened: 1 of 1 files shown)


msmttchr:
In general, I believe that a sequence as defined today is nothing more than a convenient way to queue a set of experiments (with certain parameters sweep) without the burden of doing it by hand one by one 
As such I would find peculiar the fact that behaviour maybe different whether a sequence is queued "by hand" or by the sequencer mechanism.

Having said that, a possibility could be to add some sequence property to the procedure, eg. something like:
- `sequence_info` which return whether a procedure is part of a sequence, ordinal in the sequence and length of the sequence
- `set_sequence_info` which set the information when the procedure is queued.

Please consider this just an idea and not a recommendation to change your PR as other people may have some further suggestion.

I think also documentation and tests need to be updated.

LongnoseRob:
> As such I would find peculiar the fact that behaviour maybe different whether a sequence is queued "by hand" or by the sequencer mechanism.

This is a good point for the discussion, as there might be use-cases where procedures are added while the sequence is already running.
Like it can be done with the current example of `examples/Basic/gui_sequencer.py`
With that we also should think about how to refresh the flag(s) on the last procedure if we add to the queue while the sequence is working on the already exisiting entries to the queue.

Maybe adding a "start-up" and "shutdown" element/feature to the queue could be another way to implement the idea of this PR.
("start-up" would be the 0th procedure to be called and "shutdown" would be the N+1st procedure)

Havbing such "shutdown" element could also be called in case there is a failure condition during the procedure or the user aborts the sequence. (e.g. turn off all power supply outputs after the user aborted by Ctrl-C).

BenediktBurger:
Closed as stale. It can be reopened, whenever you want to continue your work.

PR #547, opened 2021-12-17
Switch on Python 3.10 testing

bilderbuchi:


bilderbuchi:
As more issues start being reported that this fixes, I'll merge it right away.

PR #549, opened 2021-12-19
Adding HP3437A system-voltmeter & dynamic bitfield creation

LongnoseRob:
Adds support for the HP3437A system-voltmeter, another vintage instrument from the time before SCPI.


msmttchr:
@LongnoseRob, I didn't have time yet to make s full review, but based on the instrument's name I would expect it's similar to hp3478a. Is this the case? If yes can't you use some form of inheritance between instrument classes to minimize code duplication?

LongnoseRob:
@msmttchr thank you for the comment.
Honestly, I had similiar hopes, but as the 3437A was released in 1977 and the 3478A in 1983, the technology inside is different and with that the command structure (HP NanoProcessor vs. IIRC 6809).

The 3437A is a system volt-meter which can only masure voltages with 3/12 digits, but with high speed (the manual mentions up to 5800sps)
On the otherhand the 3478A is 5 1/2 digit DCV-ACV-DCI-ACI-OHMS DMM.
I reused much of the bitfields decoding part from the 3478A, but had add e.g. BCD decoding.

I will have a tought about if I could create a `HP-pre-SCPI` helper class/module which could be used in both instrument-classes.

LongnoseRob:
moved the bitfield/bytefield stuff from both meters to a new file.
Had a look at class_factories with ctype, but could not figure out a nice way until know how to have  a more compact and better maintainable way for these bitfields..

LongnoseRob:
This PR now includes a HPsupport class which can create the bit-&bytefields in a more dynamic way,
adding further instruments with such bitfields should be easier now.

Basic tests with my instruments passed, but will test the implmentatation further.

pymeasure/instruments/hp/hphelper.py
```diff
@@ -0,0 +1,260 @@
+#
```
msmttchr, line 1: My proposal is to add in this common file, the support for managing bitfield and byte and remove instrument specific information.
In order word, you could create a class (let's call for the sake of discussion `HPLegacyInstrument`) inheriting from `Instrument` with static methods implementing the factories.
Moving the specific instrument's definition and initialization to specific instrument file.
Specific instrument classes are now inheriting from `HPLegacyInstrument`.
I noticed also that `__float__` method seems specific to HP3437A. Is this the case ?
I have to say that common part is not so big, so, people may argue that at the end it may not be worth having a separate file. Also some people don't like long chain of inheritance

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,412 @@
...
+        return np.array(read_data[:-2].decode("ASCII").split(","), dtype=float)
+
+    # commands/properties for instrument control
+    def check_errors(self):
+        """
+        As this instrument does not have a error indication bit,
+        this function alwyas returns 0.
+
+        """
+        return 0
```
msmttchr, line 259: This method normally return a list, so would it be more appropriate to return `[]` ?

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,412 @@
...
+        """
+        return bool(self._decode_status(self, self._fetch_status(), "Format"))
+
+    @talk_ascii.setter
+    def talk_ascii(self, value):
+        if value is True:
+            self.write("F1")
+        else:
+            self.write("F2")
+
```
msmttchr, line 282:
```
talk_ascii = Instrument.control("B", "%s",
                                   """ Return True if the instrument is set to ASCII communication,
        this property can be set.
                                   This property can be read
                                   """,
                                   values={True: "F1", False: "F2"},
                                   map_values = True,
                                   get_process=<process output of B and return "F1" or "F2">
                                   )
```
I was thinking about the above sketch implementation, but I believe there are reasons why this cannot be done.
Could you please comment on this point ?
Similar consideration are valid also for other properties.

pymeasure/instruments/hp/hp3437A.py
LongnoseRob: > This method normally return a list, so would it be more appropriate to return `[]` ?
Yes good catch, will change this.

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,412 @@
...
+        :param data: list of bytes to be decoded
+        :return ret_data: float value
+
+        """
+        ret_data = self.packed_data(self.packed_bytes(*data))
+        return float(ret_data.b)
+
+    # commands overwriting the base implementaiton
+    def read(self):
+        """
```
msmttchr, line 218: I think, this method is used to read measurements from instrument, so I would think it is a bit misleading to call it `read` as the method with same name in the parent class is meant to read generic response from instrument.
My proposal would be either to rename the method to something like `read_measurement` (or similar) or have an `Instrument.measurement` property that would simply read from instrument (I am not sure the latter proposal is technically feasible, though)

pymeasure/instruments/hp/hp3437A.py
LongnoseRob: The problem on this device is that the status information queried with the `B` command is encoded in a raw byte string of 7 bytes. and I think from earlier tests instrument.control is not 8bit transparent as a `read_raw()` is.

I could try switch the `@xxx.setter` methods  to `Instrument.setting,` this would clean up the code a bit (?).

I will try a bit more maybe I come up with some good compromise solution.

pymeasure/instruments/hp/hp3437A.py
LongnoseRob: My Intention of overwriting the `read` method was to catch two problems:

1. the problem of data decoding if the packed data format is selected
2. find a way to handle the GPIB-timeout related problems if a high number of readings and a long delay is set. Which I still need to debug further


I could of course rename the function to `read_measurement`, but then we should aslso add the warning that the `read` method may not correctly deliver data

pymeasure/instruments/hp/hphelper.py
LongnoseRob: Thank you for the further explanation/proposal, that makes much more sense, then the way I tried with adding elements from an external class.

The `__float__`method is specific for the HP3437A at the mooment, as the HP3478A does not use fractional numbers encoded in BCD.

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,412 @@
...
+
+    def __init__(self, resourceName, **kwargs):
+        super(HP3437A, self).__init__(
+            resourceName,
+            "Hewlett-Packard HP3437A",
+            includeSCPI=False,
+            send_end=True,
+            read_termination="\r\n",
+            write_termination="\r\n",
+            **kwargs,
```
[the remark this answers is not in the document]
LongnoseRob: > I think, `read_termination` is not used anywhere, since when you are reading from instrument, you use `read_raw`. Correct ? From what I understood from your code, all message to instrument requires a termination '\r\n' except binary program ('B'). Is this correct ? If this the case, it would be possible override `write` and `values` methods as follow:

You are right. `read_termination` is not needed in this case.
However when reading data in ASCII format it is used, I think. 
Will give it a closer look.

> ```
> def write(self, command):
>     """ This is needed only if 'B' cannot be followed by a termination character (to be checked) """
>     if command == "B":
>         self.adapter.connection.write("B", termination="")
>     else:
>         super().write(command)
> ```
This could be probably be omittied, as I do not plan at the moment to implement the write part of the `B`-command.


>```
> def values(self, command, **kwargs):
>     if command == "B":
>         self.write(command)
>         return self.adapter.connection.read_raw(**kwargs)
>     else:
>         return super().values(command, **kwargs)
> ```
Yeah that would be one idea, will givit ia try.

> Then a each field of the status property could be defined as follow:
> 
> ```
> <field name> = Instrument.measurement("B",
>                                 get_process = <function to decode the specific field in the status>,
>                         )
> ```
> Again, some ideas as a starting point for discussions and not a definitive request to make changes.
> 
Your feedback is **much appreciated** and I will definitely give your proposal a try after I got the basic things working with the `HPLegacyInstrument `class.

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,412 @@
...
+        """
+        self.adapter.connection.assert_trigger()
+
+    def reset(self):
+        """
+        Initatiates a reset (like a power-on reset) of the HP3437A
+
+        """
+        self.adapter.connection.clear()
+
```
msmttchr, line 404: Does make sense to send the command "DCL" to reset the instrument ?

LongnoseRob: `self.adapter.connection.clear()` sends `IFC`.
which works in the same way as a DCL,
sending DCL shoudl work with pyvisa and the commercial visa-backends, but pyvisa-py still has some open issues on that.

msmttchr: Ok, thanks for the clarification

pymeasure/instruments/hp/hp3437A.py
msmttchr: What about redefining `read `as a wrapper to `read_raw`, something like:
```
def read(self):
     return self.adapter.connection.read_raw()
```
In this mode, you could have ` get_process` parameter that can make the necessary post processing to extract information from binary value.
I think you need also to redefine the `values` though.

LongnoseRob: @msmttchr that it is a good idea, I might try that later this week,

LongnoseRob:
@bilderbuchi my mistake, I messed up with git.
I wanted to rebase to the latest master and something got messed up.

At this stage, I will close this PR, check my stuff locally and prepare a new PR

bilderbuchi:
Sure! If you need help sorting git stuff out, then ping me.

PR #551, opened 2021-12-19
Fix browser progress bar fails to process float() signals

### LINKED ISSUE #550: Progress bar fails to update if signals are emitted as float()
`setValue` call in `BrowserItem` must use an `int` to update the UI https://github.com/pymeasure/pymeasure/blob/32423d20ac13ee1ae889764ff46ad83271444b93/pymeasure/display/browser.py#L77

However, `Monitor` emits progress signals as `float` https://github.com/pymeasure/pymeasure/blob/32423d20ac13ee1ae889764ff46ad83271444b93/pymeasure/display/listeners.py#L95

Causing cast errors like:

```
  File "/home/user/miniconda3/envs/eos2/lib/python3.10/site-packages/pymeasure/display/manager.py", line 258, in _finish
    experiment.browser_item.setProgress(100.)
  File "/home/user/miniconda3/envs/eos2/lib/python3.10/site-packages/pymeasure/display/browser.py", line 77, in setProgress
    self.progressbar.setValue(progress)
TypeError: setValue(self, int): argument 1 has unexpected type 'float'
```

I did not experience this error before. I could not pinpoint why is now being triggered (python 3.10? A recent commit?), but in any case forcing a cast to `int` doesn't hurt, so I will make a PR.

carez:
Fixes #550

### DIFF
--- a/pymeasure/display/browser.py
+++ b/pymeasure/display/browser.py
@@ -74,7 +74,7 @@ class BrowserItem(QtGui.QTreeWidgetItem):
             """)
 
     def setProgress(self, progress):
-        self.progressbar.setValue(progress)
+        self.progressbar.setValue(int(progress))
 
 
 class Browser(QtGui.QTreeWidget):


msmttchr:
This is being addressed in #547, so I think this PR could be closed

PR #552, opened 2021-12-21
Added image data generation to Mock Instrument class

samcondon4:
Added a few properties to the Mock instrument class to generate numpy arrays with random data that can be used as image data in subsequent testing. I plan to do some work on the ManagedImageWindow class in the near future and will plan to use the mock instrument class with these simple new additions.


pymeasure/instruments/mock.py
```diff
@@ -100,3 +106,52 @@ def set_output_voltage(self, value):
         self._output_voltage = value
 
     output_voltage = property(fget=get_output_voltage, fset=set_output_voltage)
+
+    def get_frame_width(self):
```
bilderbuchi, line 110: Is there a specific reason you're not using the `@property` decorator that [comes with Python](https://docs.python.org/3.9/library/functions.html?highlight=@property#property)? That would make for a more concise definition of these methods, and would not add e.g. the `set_frame_width` to the class namespace.

pymeasure/instruments/mock.py
```diff
@@ -100,3 +106,52 @@ def set_output_voltage(self, value):
...
+    def get_frame_format(self):
+        """ Format for image data returned from the get_frame() method. Allowed values are:
+                mono_8: single channel 8-bit image.
+                mono_16: single channel 16-bit image.
+        """
+        time.sleep(self._wait) 
+        return self._frame_format 
+
+    def set_frame_format(self, form):
+        if not form in ["mono_8", "mono_16"]:
```
bilderbuchi, line 141: you could use our `strict_discrete_set` [validator](https://pymeasure.readthedocs.io/en/latest/api/instruments/validators.html) for this.

pymeasure/instruments/mock.py
```diff
@@ -22,15 +22,17 @@
...
 
-import numpy
 import time
+import numpy as np
 
 from pymeasure.adapters import FakeAdapter
 from pymeasure.instruments import Instrument
 
 
 class Mock(Instrument):
```
bilderbuchi, line 32: Can you rename this to `FakeInstrument` while you're in there? I think it reduces potential confusion with ` mock.Mock`, and is more consistent with `FakeAdapter`.

bilderbuchi (changes requested):
Sure, why not. This class is not really used atm, so it's good that it gets some love :-)
I have a couple of requests, and there are a number of linter reports (now that we have set this up).

pymeasure/instruments/mock.py
samcondon4: I was just following the format that properties were already written in the class before. I agree that using the `@property` decorator is cleaner though. I've moved all of the properties to use this.

pymeasure/instruments/mock.py
samcondon4: Yep. Using this now

pymeasure/instruments/mock.py
samcondon4: Would you like the `Mock` class renamed to `FakeInstrument` or would you like `Mock` to inherit from the currently existing `FakeInstrument` class? I have assumed the latter for the latest commit.

bilderbuchi: Oh, yeah, of course, that would collide otherwise, sorry about the confusion. 
I suggest the following:
* Inherit from `FakeInstrument` as you chose already.
* Give `Mock` a more expressive name. It's a bit hard at this point as this mixes I/O features between different instrument types, so it does not cleanly matcha single one, say, an oscilloscope. The best I could come up with is `MultiFake` or `SwissFake` (like a multitool or Swiss knife). I'm open to better suggestions, otherwise `SwissFake` is probably most appropriate, as its ridiculousness also stands out a little, which is good for a dummy instrument. :D 
* Rename mock.py to fakes.py and move the `FakeInstrument` implementation there. This improves the structure a bit by keeping dummy/testing classes a bit away from `Instrument`. AFAICT, this only necessitates a changed import in `test_instruments.py` (I know `FakeAdapter` still resides with `Adapter`, but there are no other dummy classes around, so we IMO haven't reached the point where that should be extracted.)

pymeasure/instruments/__init__.py
```diff
@@ -24,9 +24,9 @@
 
 from ..errors import RangeError, RangeException
 from .instrument import Instrument
-from .mock import Mock
 from .resources import list_resources
 from .validators import discreteTruncate
+from .fakes import SwissArmyFake, FakeInstrument
```
bilderbuchi, line 29: I would actually not even include them in the `instruments` namespace, as they are used for testing only.

tests/instruments/test_instrument.py
```diff
@@ -23,12 +23,13 @@
...
 import pytest
-from pymeasure.instruments.instrument import Instrument, FakeInstrument
+from pymeasure.instruments.fakes import SwissArmyFake
+from pymeasure.instruments.instrument import Instrument
 from pymeasure.instruments.validators import strict_discrete_set, strict_range
 
 
 def test_fake_instrument():
-    fake = FakeInstrument()
+    fake = SwissArmyFake()
```
bilderbuchi, line 32: hm, I actually would prefer to keep using the simpler `FakeInstrument` for the general tests (reduces the number of "moving parts"), and use the `SwissArmyFake` only where needed.

bilderbuchi (changes requested):
I found two small things. I'll try and fix the linter things directly.

pymeasure/instruments/fakes.py
```diff
@@ -0,0 +1,200 @@
...
+        im_format_type_dict = {"8": np.uint8, "16": np.uint16}
+        bit_depth = self.frame_format.split("_")[1]
+        time.sleep(self._wait)
+        return np.array(
+            im_format_maxval_dict[bit_depth] * np.random.rand(self.frame_height, self.frame_width),
+            dtype=im_format_type_dict[bit_depth]
+        )
+
+    def __getitem__(self, keys):
+        return keys
```
bilderbuchi, line 200: Btw, I was wondering - what is this method for?

pymeasure/instruments/__init__.py
samcondon4: Good point. Changing this now.

pymeasure/instruments/fakes.py
samcondon4: I am not sure actually. It looks like that was implemented before I made any changes to the class. Shall we remove it?

tests/instruments/test_instrument.py
samcondon4: Ok, yeah good point. Changing it back.

samcondon4:
@bilderbuchi Thanks for taking care of the Linter report issues. Sorry for my delay. Things have been real busy in the lab the past few weeks. The latest commit reflects your recent comments. I think the last thing that remains is what to do with the `__getitems__` method. This was implemented before I modified the class at all so I am hesitant to remove it, though it is not clear to me what its purpose is.

pymeasure/instruments/fakes.py
bilderbuchi: Should have checked myself instead of asking you, sorry. I'll remove that, I don't think it serves any purpose anymore.

bilderbuchi:
No worries! Let's see if this builds correctly, then I'll squash-merge it.

PR #553, opened 2021-12-22
Added management of progress bar

msmttchr:
In the examples, add also management of progress bar to address misunderstanding similar to those reported in https://github.com/pymeasure/pymeasure/issues/531#issuecomment-979969959
Implement solution proposed by @CasperSchippers


CasperSchippers (approved):
LGTM, thanks for adding this.

PR #556, opened 2021-12-23
Creating independant InstrumentControlWindow

maederan201:
I have to preface this PR by saying that it is largely based on #336 by @CasperSchippers, which uses an addition to the ManagedWindow. As I'm not that familiar with Qt I chose perhaps a bit a simpler route. I create a new PR not to compromise the old one, if at some point this will be followed up with again.

This PR mainly adds `InstrumentControlWidget` and `InstrumentControlWindow` to be used for controlling single instruments. This can be useful for instruments not providing a GUI interface themself or for interacting with an instrument while or in between procedures. Also, some provided GUIs are bulky and frequently used options can be hidden. This InstrumentControlWidget could possibly implement only a few settings/options/measurements and simplify operations this way.

Similar to #336 I differentiated between five types of elements:

- Measurements: Read-only instrument properties (basically `Instrument.Measurement`s)
- Controls: Read- and Write instrument properties (basically `Instrument.Control`s)
- Settings: Write-only instrument properties (basicall `Instrument.Setting`s)
- Options: BooleanParameters or Instrument properties which are accept boolean values
- Functions: Either Instrument specific functions or external functions

There is a new basic example `gui_instrument.py` based on the `MockInstrument` class with some additions to play around with and explore.

Globally, there is the choice between automatically reading and writing values or doing so upon pressing the corresponding button. In terms of the layout, I put settings and options in one dock under the header _Settings_. The reason I need the option-types separately is to have the option of putting checkboxes instead of e.g. a float 0/1 for booleans. There could be a way of figuring this out without needing an extra type of element, hints are welcome.

Now, I marked this still as a draft as I would appreciate some inputs from different users on what would be choices in the design before I go on and fix all the details (handling of edge cases, layout finalization, documentation, etc.). In my opinion this should be something which is largely general in particular in terms of the layout. But one could inherit from this class and overwrite e.g. the layout function to create his/her own instrument specific layout. How would you differentiate between the different types of instrument properties? How do you see this being used in the lab?

Any other comments, hints and suggestions are welcome of course.


pymeasure/instruments/mock.py
```diff
@@ -49,6 +49,13 @@ def __init__(self, wait=.1, **kwargs):
                        'output_voltage': 'V',
                        'time': 's',
                        'wave': 'a.u.'}
+        self.running = False
```
samcondon4, line 52: Are lines 52-58 here needed for all instruments with your InstrumentControlWidget? If so, a better place for this would probably be the Instrument base class.

pymeasure/display/windows.py
```diff
@@ -681,10 +713,54 @@ class ManagedImageWindow(ManagedWindow):
...
 
         if "widget_list" not in kwargs:
             kwargs["widget_list"] = ()
-        kwargs["widget_list"] = kwargs["widget_list"] + (self.image_widget, )
+        kwargs["widget_list"] = kwargs["widget_list"] + (self.image_widget,)
 
         super().__init__(procedure_class, x_axis=x_axis, y_axis=y_axis, **kwargs)
+
+
+class InstrumentControlWindow(QtGui.QMainWindow):
```
samcondon4, line 726: I'm sure you plan to do this in a subsequent commit, but adding docstrings to this class in a manner similar to the ManagedWindow classes should definitely be added.

samcondon4:
My feeling is that this PR would be an excellent new feature. Perhaps you could include a screenshot of the gui that your example produces in a comment? I think that would be helpful for the reviewers. I ran your example on my computer and it looks pretty good, though I will say that the layout that @CasperSchippers has implemented in PR #336 in his `InstrumentWidget` feels a bit neater, more compact, and nicer to plug into the more elaborate Qt classes already implemented (all of the ManagedWindow classes).

That being said, something that I have not seen utilized in pymeasure.display widgets at all yet is `pyqtgraph's Parameter Trees`: https://pyqtgraph.readthedocs.io/en/latest/parametertree/index.html

I have used these classes just a bit in the past to generate user interfaces for interactive data visualization applications and they may be worth looking into for this as they make it quite convenient to create nice, compact guis for interacting with several parameters at once.

pymeasure/instruments/mock.py
maederan201: No, I just added this to have an example for functions implemented in an Instrument class and to see that the function actually works.

pymeasure/display/windows.py
maederan201: Of course! Did not do it yet because the structure might still change

maederan201:
Here is a screenshot of what it looks like right now

Screenshot_20220105_135230
(Running on Linux, that's why it looks like this.

There are improvements needed for the layout for sure. I thought a bit and it might make more sense to put everything in one column. One might have an easier time adding it to e.g. a ManagedWindow, similar to what @CasperSchippers has done in #336. However I still think it should also be possible to have it as a standalone window as well.

With regards to the `ParameterTrees`: I'm not familiar at all, so I would need some time to get familiar, or someone to help getting started. However I can see it a bit over-cluttering for simple examples, no? I agree that it could be nice for instruments with many variables you'd want to control. The use case I am aiming at is more tailored to a relatively small number of properties (let's say 10 to 20 max.). However I can see one nice benefit: You should be able to automatically and easily create a control widget with _all_ implemented members of the InstrumentClass. 

I can see use in both, its also a design choice that needs to be made I guess. And I am not the person to make it. However I will continue tinkering around with the layout and some other improvements

maederan201:
I like it better like this, recent pushes reflect this change of layout:

Screenshot_20220105_145525

maederan201:
I tried to generalize things a bit but ran into some issues. I would want a control window to automatically detect what is the best option to display a certain variable type. At least three types come to mind

1. Floats, for all kinds of settings, controls, measurements, the most obvious ones
2. Strings/Text, e.g. one might want to have the state (READY, RUNNING,...) of an instrument updating live
3. Boolean Values which should be Checkboxes to check/uncheck certain things

Would you expect more possibilities? I think the existing parameter classes should be used for that as far as possible to build on existing code (e.g. as done with the `input_from_parameter` function). What would you expect as a user to input? I think passing strings with the name of instrument class variables is straight forward and should be the default way to do it. For other things one could either allow any parameter or limit it. One option could be that the user needs to pass a Parameter if he wants something else than the default FloatParameters.

What would be a smart way to achieve this while keeping it looking nice? Maybe one has to work with subwidgets for each property, so one would have a ListSubwidget, FloatSubwidget, BooleanSubwidget which could then implement the specific SpinBox, CheckBox, Label, etc.

Let me know what seems useful and what you would expect from an InstrumentControlWindow

PR #559, opened 2022-01-01
Driver for the HP 8116A 50 MHz Pulse/Function Generator

CodingMarco:
This PR adds a driver for the HP 8116A. I have tested it as good as possible with two units, one with and one without option 001 (sweep/burst capability).

### DIFF
--- a/pymeasure/instruments/hp/hp8116a.py
+++ b/pymeasure/instruments/hp/hp8116a.py
@@ -25,11 +25,30 @@
 import logging
 import time
 import numpy as np
+from enum import IntFlag
 from pymeasure.instruments import Instrument
-from pymeasure.instruments.validators import strict_discrete_set, strict_range, truncated_discrete_set
+from pymeasure.instruments.validators import (
+    strict_discrete_set, strict_range, truncated_discrete_set
+)
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
+
+
+class Status(IntFlag):
+    """ IntFlag type for the GPIB status byte which is returned by the :py:attr:`status` property.
+    When the timing_error or programming_error flag is set, a more detailed error description
+    can be obtained by calling :py:method:`check_errors()`.
+    """
+    timing_error = 1 << 0
+    programming_error = 1 << 1
+    syntax_error = 1 << 2
+    system_failure = 1 << 3
+    autovernier_in_progress = 1 << 4
+    sweep_in_progress = 1 << 5
+    service_request = 1 << 6
+    buffer_not_empty = 1 << 7
+
 
 class HP8116A(Instrument):
     """ Represents the Hewlett-Packard 8116A 50 MHz Pulse/Function Generator
@@ -47,13 +66,13 @@ class HP8116A(Instrument):
             includeSCPI=False,
             **kwargs
         )
-    
+
     OPERATING_MODES = {
         'normal': 'M1',
         'triggered': 'M2',
         'gate': 'M3',
         'external_width': 'M4',
-        
+
         # Option 001 only
         'internal_sweep': 'M5',
         'external_sweep': 'M6',
@@ -118,41 +137,30 @@ class HP8116A(Instrument):
         'kilo': 1e3,
         'mega': 1e6,
     }
-    
-    @property
-    def haversine(self):
-        """ With triggered, gate or E.BUR operating mode selected and shape
-        set to sine or triangle, this setting shifts the start phase of the
-        waveform is -90°. As a result, haversine and havertriangle signals
-        can be generated.
-        """
-        raise NotImplementedError
-    
-    @haversine.setter
-    def haversine(self, haversine):
-        haversine_set = int(haversine)
-        haversine_cmd = "Z" + str(int(strict_discrete_set(haversine_set, [0, 1])))
-        self.write(haversine_cmd)
 
-    ## Instrument communication ##
+    @property
+    def status(self):
+        """ Returns the status byte of the 8116A as an IntFlag-type enum. """
+        return Status(self.adapter.connection.read_stb())
+
+    # Instrument communication #
 
     def write(self, command):
-        """ Write a command to the instrument.
-        If the command is a query (interrogate or CST for instrument state),
-        wait for the 8116A to respond before returning.
-        """
+        """ Write a command to the instrument and wait until the 8116A has interpreted it. """
         self.adapter.write(command)
 
-        if command.strip().lower()[0] in ('i', 'c'):
+        # We need to read the status byte and wait until this bit is cleared
+        # because some older units lock up if we don't.
+        while self.status & Status.buffer_not_empty:
             time.sleep(0.001)
-    
+
     def read(self):
         """ Some units of the 8116A don't use the EOI line (see service note 8116A-07A).
         Therefore reads with automatic end-of-transmission detection will timeout.
         Instead, :code:`adapter.read_bytes()` has to be used.
         """
         raise NotImplementedError('Not supported, use adapter.read_bytes() instead')
-    
+
     def ask(self, command, num_bytes):
         """ Write a command to the instrument, read the response, and return the response.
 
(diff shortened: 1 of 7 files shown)


pymeasure/instruments/hp/hp8116a.py
```diff
@@ -0,0 +1,176 @@
...
+        self.write(freq_cmd)
+
+    @property
+    def amplitude(self):
+        """ A floating point property that controls the amplitude of the
+        output in V. The allowed amplitude range is -10 V to 10 V.
+        """
+        raise NotImplementedError
+
+
```
msmttchr, line 176: @CodingMarco, thank for your draft contribution.
I understand you still need to complete it, but as an early comment, I was curious whether there is a specific reason to avoid usage of property factories like `Instrument.control` etc.

CodingMarco: The problem is that this is another instrument from the pre-SCPI era. Not only that, things are even more difficult since some units of this instrument have the GPIB "EOI" line cut which results in a timeout on all reads with pyvisa (see service note 8116A-07A "Solving EOI Prablems" in the 8116A service manual). Because of that, only `adapter.read_bytes(nr_of_bytes)` can be used to read the instrument state or query specific parameters (nr_of_bytes is different for each command).
Also it is important to have a short delay between writing a command and reading the response, otherwise the 8116A answert "NO MESSAGE".

Would this be possible with `Instrument.control`?

msmttchr: I think it's possible, all the problems you mentioned are related to the physical access to the instrument while `Instrument.control` is in charge of formatting the messages to instrument and decoding the  messages from instrument in a flexible and standardized way.

Regarding your specific points:
- non SCPI: This possible, have a look at the work don't by @LongnoseRob in pymeasure/instruments/hp/hp3478A.py
- For the EOI and delay problem, it should be possible to customize `write`, `read` and `ask` methods of `Instrument` class to serve your needs

Hope this helps

LongnoseRob (commented):
@CodingMarco that looks very promising!

One thing I did not see is the `check_errors` function, to see if the instrument reoirts any errors (if such a status or command is available on the instrument)

Another item, that you could still add is the hp8116A.rst in docs/api/instruments/hp/
For that you could copy the hp3478A.rst and change it to match for the 8116, also than please add the hp8116A.rst to the index.rst

pymeasure/instruments/hp/hp8116a.py
```diff
@@ -0,0 +1,506 @@
...
+            **kwargs
+        )
+    
+    OPERATING_MODES = {
+        'normal': 'M1',
+        'triggered': 'M2',
+        'gate': 'M3',
+        'external_width': 'M4',
+        
+        # Option 001 only
```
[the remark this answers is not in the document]
CodingMarco: No, unfortunately there is no direct way like a command. I think I'd have to send a command that's only valid with Option 001 and see if there's an error.
But now that you mentioned it, I definetely have to know wether the connected unit has Option 001, since the status responses have different lengths. However I would say it would be better to have it as a constructor argument, what do you think?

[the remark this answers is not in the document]
CodingMarco: Yes, I still have to implement `check_errors`. There is an "IERR" query and also a status byte that contains error information. Also, there are still some other small features that I have to implement like autovernier mode.
Thanks for pointing out the documentation, I'm going to do that when the implementation is ready.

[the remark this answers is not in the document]
LongnoseRob: @CodingMarco Wanted to propose some enum action for the errors but had a look at the manual of the 8116A, and the `IERR` comand is actually quite comfortable with the clear ASCII replies (e.g. `LIMIT ERROR`, tables 3-6 and 3-7 of the manual).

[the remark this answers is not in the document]
LongnoseRob: > No, unfortunately there is no direct way like a command. I think I'd have to send a command that's only valid with Option 001 and see if there's an error. But now that you mentioned it, I definetely have to know wether the connected unit has Option 001, since the status responses have different lengths. However I would say it would be better to have it as a constructor argument, what do you think?

Actually there is are two ways to see if option 001 is installed:

- like you mentioned before: Set an operation mode (e.g. M5-M8 which are Option 001 only) and if the option is not installed, we should get an errorstatus.

- length of the CST (current status, Paragraph 3-68 of the manual) return data, without option 001 the length should be 89 bytes, with option 001 its length is 161 bytes. so using this command and parsing it during the initialization of the instrument instrance could give us this info and also fill the instrument properties with the current status of the instrument.

pymeasure/instruments/hp/hp8116a.py
CodingMarco: Thanks for the suggestion, as you can see I have made extensive use of the predefined properties which were very helpful in the end. However, I had to essentially copy `values()` from `adapter` to my class because the method in adapter calls `self.ask()` which I have to override. Maybe this is a general design issue that can be addressed in the future.

pymeasure/instruments/hp/hp8116a.py
```diff
@@ -0,0 +1,541 @@
...
+            elif command[0] == 'I':
+                num_bytes = 14
+
+        # The first character is always a space or a leftover character from the previous command,
+        # when the number of bytes read was too large or too small.
+        bytes = self.adapter.read_bytes(num_bytes)[1:]
+        return bytes.decode('ascii').strip(' ,\r\n')
+
+    def values(self, command, separator=',', cast=float, preprocess_reply=None, **kwargs):
+        # I had to copy the values method from the adapter class since we need to call
```
msmttchr, line 212: I don't think you need to reimplement `values`, since you already changed the `ask` method. Or I am missing something?

CodingMarco: See https://github.com/pymeasure/pymeasure/pull/559#discussion_r780658782

msmttchr: I am sorry @CodingMarco, you are right and I missed your comment.
I will possibly file an issue such that this can be discussed/fixed in future.
One last point `ask` method is doing the strip so I am not sure it is also needed in `values`

CodingMarco: Yes you're right, I'll remove it in `values`.

pymeasure/instruments/hp/hp8116a.py
```diff
@@ -0,0 +1,608 @@
...
+        'sine': 'W1',
+        'triangle': 'W2',
+        'square': 'W3',
+        'pulse': 'W4',
+    }
+
+    SHAPES_INV = {v: k for k, v in SHAPES.items()}
+
+    _units_freqency = {
+        'milli': 'MZ',
```
msmttchr, line 130: Minor spelling error _units_freqency

pymeasure/instruments/hp/hp8116a.py
```diff
@@ -0,0 +1,608 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2021 PyMeasure Developers
+#
```
msmttchr, line 5: Update copyright to 2022?

tests/instruments/hp/test_hp8116a.py
```diff
@@ -0,0 +1,266 @@
...
+    def test_given_invalid_duty_cycle_when_check_errors_then_error(self, make_resetted_instr):
+        instr = make_resetted_instr
+        instr.frequency = 50e6
+        instr.duty_cycle = 90
+        errors = instr.check_errors()
+        assert len(errors) == 1
+        assert errors[0] == 'DUTY C. ERROR'
+
+    def test_given_two_error_conditions_when_check_errors_then_two_errors(self,
+                                                                          make_resetted_instr):
```
msmttchr, line 229: Self descriptive function name :)

msmttchr (changes requested):
Wonderful work. Congratulations and thanks for contributing to pymeasure. Some *really* minor remarks

tests/instruments/hp/test_hp8116a.py
CodingMarco: Yeah "we" as computer scientists / software engineers (not electrical engineers just doing some programming) sometimes do that :joy: (Not meant to be offensive)

pymeasure/instruments/hp/hp8116a.py
CodingMarco: Thanks for catching that.

msmttchr (approved):
Very good contribution. Thanks 👍

PR #569, opened 2022-01-13
Remove errorbars

### LINKED ISSUE #507: Wrong right errorbar
More by accident I stumbled upon a bug/typo/copy-paste-error in ```update_data``` of the ```ResultsCurve``` class when looking for something unrelated. The following code-block sets errorbar options if enabled:

```
if hasattr(self, '_errorBars'):
            self._errorBars.setOpts(
                x=data[self.x],
                y=data[self.y],
                top=data[self.yerr],
                bottom=data[self.yerr],
                left=data[self.xerr],
                right=data[self.yerr],
                beam=max(data[self.xerr], data[self.yerr])
            )
```
I think the line ```right=data[self.yerr]``` should be changed to ```right=data[self.xerr]```. Can someone confirm my suspicion? I have never worked with errorbars and I don't have an example to test it ready.

CasperSchippers:
PR to remove the unused and not-working errorbar feature.
Resolves #507

### DIFF
--- a/pymeasure/display/curves.py
+++ b/pymeasure/display/curves.py
@@ -44,22 +44,17 @@ def _greyscale_colormap(x):
 
 
 class ResultsCurve(pg.PlotDataItem):
-    """ Creates a curve loaded dynamically from a file through the Results
-    object and supports error bars. The data can be forced to fully reload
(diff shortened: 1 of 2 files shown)


msmttchr (approved):
Looks good to me

PR #577, opened 2022-01-23
Modernise the code base to use Python 3.7 features

bilderbuchi:
This mainly starts using f-strings and .format instead of the % notation, and removes superfluous super() arguments.

Used the pyupgrade tool to do this: `find . -name '*.py' -exec pyupgrade --py37-plus {} +`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 #
 # PyMeasure documentation build configuration file, created by
 # sphinx-quickstart on Mon Apr  6 13:06:00 2015.
@@ -55,8 +54,8 @@ source_suffix = '.rst'
 master_doc = 'index'
 
 # General information about the project.
-project = u'PyMeasure'
-copyright = u'2013-2022, PyMeasure Developers'
+project = 'PyMeasure'
+copyright = '2013-2022, PyMeasure Developers'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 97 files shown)


msmttchr (commented):
I think, pyugrade is a very nice tool to bring syntax up to speed with latest python versions.
From skim reading, looks good to me, but it is also true that  these changes are very large and they can introduce subtle bugs. At the end of the day everything boils down to the degree of reliability of pyupgrade.
Are you confident on it ? Do you have already some experience ? (I am asking also because I plan to use it on some personal projects :) )

bilderbuchi:
* I have no previous experience
* The creator, Anthony Sottile, is a known Python dev
* The package has existed for a couple of years
* The ratio of open to closed issues and pull requests is extremely low
* The changes seem mostly mechanical (maybe generated from inspecting the AST? I don't know), so this should be inherently safe
* This is a one-shot change, not an automatic one based on CI
* The creator even recommends the tool for [pre-commit](https://pre-commit.com/) usage.
* I inspected all the changes manually

So overall, I think the potential risk should be quite low. :-)
I'm thinking if we should even include it in our linting stage (to generate remark bubbles in the Changes view like flake8) to maintain a modern code style. In that case we should research first if there is a Github Action plugin for the decorations, or if the tool can even report back specific changes instead of changing in-place (the default behaviour).

msmttchr (approved):
Thanks, nice idea, in my opinion.

bilderbuchi:
Thanks!

CasperSchippers:
Ah, nice work; thanks!

Issue #102, opened 2017-07-20
Improve documentation


bilderbuchi:
Migrated from TODO

cjermain:
PRs welcome!

bilderbuchi:
Doctests in both pymeasure code and runnable examples in the documentation would be useful, see [here](https://github.com/ralph-group/pymeasure/issues/98#issuecomment-317273099).

bilderbuchi:
I found this a good, short, read about what constitutes good documentation: https://www.divio.com/en/blog/documentation/

bilderbuchi:
Some doctests executing code in the documentation has been added in #111, so this can be used as a guide for adding documentation doctests elsewhere. 

I have also looked a bit into adding doctests to the pymeasure code docstrings, but since most things are very interconnected and often depend on GUI features, I did not find feasible/sensible opportunities.

bilderbuchi:
document that
* [ ] `__init__` should be at the top of the class
* [ ] common pattern: bool properties with map_values
* [ ] common pattern: Channels for instruments
* [ ] common pattern: inheritance between instrument models -> nice example AnritsuMS9740A!
* [ ] PR single topic (1 instrument), small size makes reviews much easier

bilderbuchi:
* [ ] Maybe we should put some recommendation how to deal with atypical communication needs of some instruments (and pointer to examples) in the docs?  One [c|sh]ould encapsulate the device-specific comms requirements in a custom Adapter derived e.g. from VISAAdapter, instead of overriding the methods of `Instrument`. See e.g. #352/ [TopticaAdapter](https://github.com/dkriegner/pymeasure/blob/3d5f57b4d8209dfa2ae32f8794eb94e1f36fa794/pymeasure/instruments/toptica/adapters.py#L31), there are also other cases of this.  Cf .#357

LongnoseRob:
> * [ ]  Maybe we should put some recommendation how to deal with atypical communication needs of some instruments (and pointer to examples) in the docs?  One [c|sh]ould encapsulate the device-specific comms requirements in a custom Adapter derived e.g. from VISAAdapter, instead of overriding the methods of `Instrument`. See...

Very good idea, maybe we could extend this into a kind of best-practice-section

bilderbuchi:
Yeah, if we re-organize the docs according to the scheme linked above ( tutorials, how-to guides, explanation, reference), this should naturally fall out of that.

bilderbuchi:
also see #364.

[...]

PR #578, opened 2022-01-24
add docs for naming derived instruments

bklebel:
This naming scheme was suggested by @bilderbuchi in CasperSchippers/pymeasure#5 for derived classes. I think it would be good to add this info to the docs, since it has now become easier (in general) to derive instruments from each other. 

Possibly I could also put this heading one level higher, introducing a new heading for "Deriving new instruments from existing ones" and then rearrange the docs about how to use the dynamic properties from above to be included in the section about derived instruments. 
Apart from that, it might be good to change the minimal implementation shown in the docs, including the `name=Extreme 5000` as default in the `__init__`, to facilitate deriving from those classes in the future. 

Also, since I am not very familiar with Sphinx and the automatic generation of the docs, I wonder whether in the docs for derived instruments we will see all the properties and functions, and not just those which are implemented in the derived class?

### DIFF
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -698,7 +698,9 @@ In the above example, :code:`MultimeterA` and :code:`MultimeterB` use a differen
 Naming derived instruments
 **************************
 
-When deriving an instrument from an existing implementation, you might want to give separate names to any derived instrument. Two possible schemes for this are the following: 
+When deriving an instrument from an existing implementation, the derived instrument typically has a different name (e.g. another model number).
+In addition, a user might want to give instantiated instruments setup-specific names (e.g. `Helmholtz_coil_driver` instead of `GenericPSU123`), especially if there is more than one device of the same make in a setup.
+To accomodate these needs, there are two similar ways: 
 
 .. code-block:: python
 
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -698,7 +698,7 @@ In the above example, :code:`MultimeterA` and :code:`MultimeterB` use a differen
 Naming derived instruments
 **************************
 
-When deriving an instrument from an existing implementation, you might want to give separate names to any derived instrument. The suggested scheme for this is the following: 
+When deriving an instrument from an existing implementation, you might want to give separate names to any derived instrument. Two possible schemes for this are the following: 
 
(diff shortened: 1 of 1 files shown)


bklebel:
I just now found #102 which is all about documentation. I guess I will change this PR here to fully address @bilderbuchi's point about "common pattern: inheritance between instrument models", make this its own heading and so on

docs/dev/adding_instruments.rst
```diff
@@ -693,3 +693,29 @@ Another use case involves maintaining compatibility between instruments with com
...
 :code:`MultimeterB` can be defined subclassing :code:`MultimeterA` and just implementing the difference.
+
+
+Naming derived instruments
+**************************
+
+When deriving an instrument from an existing implementation, you might want to give separate names to any derived instrument. Two possible schemes for this are the following: 
+
+.. code-block:: python
+    class BaseClass:
```
bilderbuchi, line 704:
```suggestion
.. code-block:: python

    class BaseClass:
```
This needs at least this to work -- in RST there are blank lines between directives/their options and the content.

docs/dev/adding_instruments.rst
```diff
@@ -693,3 +693,29 @@ Another use case involves maintaining compatibility between instruments with com
 
 In the above example, :code:`MultimeterA` and :code:`MultimeterB` use a different command to read the voltage, but the rest of the behaviour is identical.
 :code:`MultimeterB` can be defined subclassing :code:`MultimeterA` and just implementing the difference.
+
+
+Naming derived instruments
+**************************
+
+When deriving an instrument from an existing implementation, you might want to give separate names to any derived instrument. Two possible schemes for this are the following:
```
bilderbuchi, line 701:
```suggestion
When deriving an instrument from an existing implementation, the derived instrument typically has a different name (e.g. another model number).
In addition, a user might want to give instantiated instruments setup-specific names (e.g. `Helmholtz_coil_driver` instead of `GenericPSU123`), especially if there is more than one device of the same make in a setup.
To accomodate these needs, there are two similar ways: 
```

bilderbuchi:
waiting for resolution of #659 before adapting/including this.

bilderbuchi:
I'll close this in favour of #659, as we're not sure if that will land in this form, anyway, and I want to clean up PRs a bit. I copied the text you made over there.

PR #579, opened 2022-01-27
Modified navigation of ManagedWindow directory widget

jarvas24:
This is a very small PR with two changes to the `ManagedWindow`'s directory widget:
1. The QFileDialog now opens at the currently entered directory (if there is one)
2. Added the kwarg `default_directory_path` to `ManagedWindowBase`. If provided, the directory widget will be populated with this path when the GUI loads

Please let me know if you'd like me to make any changes.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -44,3 +44,4 @@ Andreas Maeder
 Bastian Leykauf
 Matthew Delaney
 Marco von Rosenberg
+Jack Van Sambeek
--- a/docs/about/authors.rst
+++ b/docs/about/authors.rst
@@ -51,3 +51,4 @@ The following developers have contributed to the PyMeasure package:
 | Bastian Leykauf
 | Matthew Delaney
 | Marco von Rosenberg
+| Jack Van Sambeek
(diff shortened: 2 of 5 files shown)


msmttchr:
@jarvas24, thanks for your proposal. Looking at the code, I was curious if you had a look at https://pymeasure.readthedocs.io/en/latest/tutorial/graphical.html#using-the-directory-input
which can possibly be the default directory for the `QFileDialog`. This could avoid to add a new parameter to the `ManagedWindowBase`.

jarvas24:
Hi @msmttchr, that's an excellent point! I removed the parameter, and instead added a setter method for `ManagedWindowBase.directory`.  The default directory can now be set like this:

```python3

        def __init__(self):
            super().__init__(...)
            self.directory = r'C:/Path/to/default/directory'
```

I've also updated the docs to reflect this change

msmttchr (commented):
Looks good to me, but I feel more comfortable if also @CasperSchippers could have a look at it.
Also, don't forget to add yourself in the author's files (`AUTHORS.txt` and `authors.rs`t) before this PR can be merged.

CasperSchippers (approved):
Hi @jarvas24, Thanks for this PR; it's a very good idea and nice implementation.
As I have no further remarks, I will merge it right away!

PR #580, opened 2022-01-28
Base spectrum analyzer implementation

msmttchr:
A generic spectrum analyzer implementation, intentionally simple and basic.

The key features of the generic instrument implementation are as follow:
- interface is intentionally simple to be easily adapted to wide range of spectrum analyzers.
- all the instrument functions that could efficiently implemented in python are not implemented (e.g. peak search, mathematical operations on the traces, unit translation, etc.), instead they are provided as python method and this can be referred as pure software implementation of part of the instrument.
- all the features which are relevant for interactive use of the instrument are not implemented (e.g. markers, display control, multiple traces, etc.)
- a default implementation, using commands based on SCPI specification 1999, is provided, but all properties are declared dynamic so subclasses can easily modify them.

A possible approach to address #416

### DIFF
--- a/pymeasure/instruments/spectrum_analyzer.py
+++ b/pymeasure/instruments/spectrum_analyzer.py
@@ -1,0 +1,278 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2022 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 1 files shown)


bilderbuchi:
Thanks for that! At a quick glance this looks useful, if only to drive the discussion and help us shape an approach to resolve #416!

I'm not too fond of dragging in scipy as a pymeasure dependency (it's huge!), so if that can be avoided, or the peak finder an optional feature only available if scipy is already installed, that would be good.

msmttchr:
> Thanks for that! At a quick glance this looks useful, if only to drive the discussion and help us shape an approach to resolve #416!
> 
> I'm not too fond of dragging in scipy as a pymeasure dependency (it's huge!), so if that can be avoided, or the peak finder an optional feature only available if scipy is already installed, that would be good.

Ok I made `scipy` optional, even though I would love to see the instrument software part growing in the future, regardless of the use of `scipy`

bmoneke:
Thanks for that example. It helps discussing (in new issues) the traits / default instruments. 

I prefer the approach you took according to @bilderbuchi s suggestion, to make scipy optional.

msmttchr:
I think it is better to wait completion of #905, before resume work on this PR.

PR #581, opened 2022-01-28
New Instrument Driver: Newport Conex-CC Motion Controller

maederan201:
This PR adds a new Instrument Driver: Newport's Conex-CC Motion Controller for Linear Stages

Only the basic functionalities are implemented: Getting/Setting velocity and acceleration, absolute and relative moves (`move_to` and `move_by`) and all functionalities for closed loop operation.

Right now, I'm only using 1 driver but the device offers the possibility to connect multiple drivers and multiple axis (similar to the EPS300 which is already implemented), that's why I want to allow this with the `index` attribute. I was contemplating an implementation of axes similar to the EPS300 but decided against it, as there, the number of axes is hardcoded, but can of course vary on the application. How is such a case generally handled within the pymeasure frame?

### DIFF
--- a/pymeasure/instruments/newport/conexcc.py
+++ b/pymeasure/instruments/newport/conexcc.py
@@ -24,7 +24,7 @@
 
 import logging
 from pymeasure.instruments import Instrument
-from pymeasure.instruments.validators import strict_range
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


pymeasure/instruments/newport/conexcc.py
```diff
@@ -0,0 +1,147 @@
...
+        map_values=True, values={True: 1, False: 0},
+        validator=strict_discrete_set
+    )
+
+    def move_to(self, pos):
+        """ Move to absolute position
+        :param pos Absolute position to move to.
+        """
+        cmd = 'PA%f' % (pos)
+        self.write(cmd)
```
bilderbuchi, line 112: `self.write(f'{cmd:f}') should do the same thing and is a more modern string formatting approach. Similar for other occurences.

bilderbuchi (changes requested):
Thank you, this is looking good. I have some small change requests.
Are you already in the authors files for credit?

> How is such a case generally handled within the pymeasure frame?

We don't have a common API definition for instrument classes yet (see #416). Until that time it's fine to do as you did, your reasoning sounds fine.

BenediktBurger:
Hey @maederan201 , would you like to finish your almost finished work?

maederan201:
Thanks @BenediktBurger for reminding me. I will try and tackle this in the upcoming month. Thanks for the patience

BenediktBurger:
> Hey @maederan201 , would you like to finish your almost finished work?

BenediktBurger:
Closed due to lack of response. It might be reopened whenever there is interest.

PR #582, opened 2022-01-28
Add Rohde&Schwarz HMP4040 power supply.

bleykauf:
I added most commands for the Rohde&Schwarz HMP4040 with the exception of the fuse commands (since I could not test those commands). 

For some commands I implemented both `Instrument.control` and defined a callable member function since it seems a little odd to for example select a channel via `hmp.selected_channel = 1`; I think it is more natural to do this via `hmp.select_channel(1)`. These commands are annotated in the source code. Let me know what you think :)

Manual for the device for reference: https://scdn.rohde-schwarz.com/ur/pws/dl_downloads/dl_common_library/dl_manuals/gb_1/h/hmp_serie/HMP_GettingStarted_en_02.pdf

### DIFF
--- a/pymeasure/instruments/rohdeschwarz/hmp.py
+++ b/pymeasure/instruments/rohdeschwarz/hmp.py
@@ -190,24 +190,17 @@ class HMP4040(Instrument):
         map_values=True,
     )
 
-    # The following commands are for making it easier to change selected the
-    # selected channel and activate/deactivate them.
+    # The following commands are for making it easier to change the selected
+    # channels and activate/deactivate them.
 
-    def activate_channel(self, channel):
-        """Activate a channel."""
+    def set_channel_state(self, channel: int, state: bool):
+        """
+        Set the state of the channel to active (True) or inactive (False).
+        """
         # Save current selected channel before switching.
         selected_channel = self.selected_channel
         self.selected_channel = channel
-        self.selected_channel_active = True
-        # Restore previously selected channel.
-        self.selected_channel = selected_channel
-
-    def deactivate_channel(self, channel):
-        """Deactivate a channel."""
-        # Save current selected channel before switching.
-        selected_channel = self.selected_channel
-        self.selected_channel = channel
-        self.selected_channel_active = False
+        self.selected_channel_active = state
         # Restore previously selected channel.
         self.selected_channel = selected_channel
 
--- a/pymeasure/instruments/rohdeschwarz/hmp.py
+++ b/pymeasure/instruments/rohdeschwarz/hmp.py
@@ -61,7 +61,7 @@ class HMP4040(Instrument):
     # System Setting Commands -------------------------------------------------
 
     def beep(self):
-        """Return a single beep immediately."""
+        """Emit a single beep immediately."""
         self.write("SYST:BEEP")
 
     control = Instrument.setting(
@@ -125,11 +125,11 @@ class HMP4040(Instrument):
         values=[0, 32.050],
     )
 
-    def voltage_up(self):
+    def step_voltage_up(self):
         """Increase voltage by one step."""
         self.write("VOLT UP")
 
-    def voltage_down(self):
+    def step_voltage_down(self):
         """Decrease voltage by one step."""
         self.write("VOLT DOWN")
 
@@ -161,11 +161,11 @@ class HMP4040(Instrument):
         "CURR:STEP?", "CURR:STEP %s", "Current step in A."
     )
 
-    def current_up(self):
+    def step_current_up(self):
         """Increase current by one step."""
         self.write("CURR UP")
 
-    def current_down(self):
+    def step_current_down(self):
         """Decreases current by one step."""
         self.write("CURR DOWN")
 
@@ -187,7 +187,7 @@ class HMP4040(Instrument):
         map_values=True,
     )
 
-    output_on = Instrument.control(
+    output_enabled = Instrument.control(
         "OUTP:GEN?",
         "OUTP:GEN %s",
         "Set the output on or off or check the output status.",
@@ -219,11 +219,11 @@ class HMP4040(Instrument):
     # Convenience functions to make turning on/off the output more natural.
     def turn_output_on(self):
         """Turn output on."""
-        self.output_on = True
+        self.output_enabled = True
 
     def turn_output_off(self):
         """Turn output off."""
-        self.output_on = False
+        self.output_enabled = False
 
     # Measurement Commands ----------------------------------------------------
 
@@ -253,7 +253,7 @@ class HMP4040(Instrument):
         "ARB:REP?",
         "ARB:REP %s",
         "Number of repetitions (0...255). If 0 is entered, the sequence is"
-        "repeated infinitely.",
+        "repeated indefinitely.",
         validator=strict_discrete_set,
         values=range(256),
         get_process=lambda x: int(x),
--- a/pymeasure/instruments/rohdeschwarz/hmp.py
+++ b/pymeasure/instruments/rohdeschwarz/hmp.py
@@ -87,7 +87,7 @@ class HMP4040(Instrument):
         "Selected channel.",
         validator=strict_discrete_set,
         values=[1, 2, 3, 4],
-        get_process=lambda x: int(x),
+        cast=int
     )
 
     # Voltage Settings --------------------------------------------------------
@@ -242,7 +242,7 @@ class HMP4040(Instrument):
         "repeated indefinitely.",
         validator=strict_discrete_set,
         values=range(256),
-        get_process=lambda x: int(x),
+        cast=int,
     )
 
     def load_sequence(self, slot):
--- a/pymeasure/instruments/rohdeschwarz/hmp.py
+++ b/pymeasure/instruments/rohdeschwarz/hmp.py
@@ -251,7 +251,10 @@ class HMP4040(Instrument):
         self.write(f"ARB:REST {slot}")
 
     def save_sequence(self, slot):
-        """Save the sequence to internal memory (slot 1, 2 or 3)."""
+        """
+        Save the sequence defined in the sequence property to internal memory
+        (slot 1, 2 or 3).
+        """
         slot = strict_discrete_set(slot, [1, 2, 3])
         self.write(f"ARB:SAVE {slot}")
 
--- a/pymeasure/instruments/rohdeschwarz/hmp.py
+++ b/pymeasure/instruments/rohdeschwarz/hmp.py
@@ -257,11 +257,15 @@ class HMP4040(Instrument):
         self.write(f"ARB:START {channel}")
 
     def stop_sequence(self, channel):
-        """Stop the sequence of the selected channel."""
+        """Stop the sequence defined in the sequence property of the selected
+        channel."""
         channel = strict_discrete_set(channel, [1, 2, 3, 4])
         self.write(f"ARB:STOP {channel}")
 
     def transfer_sequence(self, channel):
-        """Transfer the defined sequence to the selected channel."""
+        """
+        Transfer the sequence defined in the sequence property to the selected
+        channel.
+        """
         channel = strict_discrete_set(channel, [1, 2, 3, 4])
         self.write(f"ARB:TRAN {channel}")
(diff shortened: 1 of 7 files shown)


b-leconte:
Hello,
 
Just tested your patch.
It is a good thing to have this instrument available in pymeasure.

I had small issue while testing the instrument:
using the function `def activate_channel(self, channel):` is causing an exception.
But if i comment the lines in this function:
`"selected_channel = self.selected_channel" `
and
`"self.selected_channel = selected_channel"` 
Then there is no issue anymore (except that you loose the backup of previously selected channel).

Regards,

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+
+    voltage_step = Instrument.control(
+        "VOLT:STEP?",
+        "VOLT:STEP %s",
+        "Voltage step in V. Default 1 V.",
+        validator=truncated_range,
+        values=[0, 32.050],
+    )
+
+    def voltage_up(self):
```
bilderbuchi, line 131: better `step_voltage_up`?

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+        """Increase voltage by one step."""
+        self.write("VOLT UP")
+
+    def voltage_down(self):
```
bilderbuchi, line 135: better `step_voltage_down`?

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def current_to_max(self):
+        """Set current of the selected channel to its maximum value."""
+        self.write("CURR MAX")
+
+    current_step = Instrument.control(
+        "CURR:STEP?", "CURR:STEP %s", "Current step in A."
+    )
+
+    def current_up(self):
```
bilderbuchi, line 167: better `step_current_up`?

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+        """Increase current by one step."""
+        self.write("CURR UP")
+
+    def current_down(self):
```
bilderbuchi, line 171: better `step_current_down`?

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+
+    selected_channel_active = Instrument.control(
+        "OUTP:SEL?",
+        "OUTPUT:SEL %s",
+        "Set the selected channel to active or inactive or check its status.",
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    output_on = Instrument.control(
```
bilderbuchi, line 193: `output_enabled` would echo some other instruments that use that name already (there is none that has `output_on`)

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+        self.selected_channel = selected_channel
+
+    # Convenience functions to make turning on/off the output more natural.
+    def turn_output_on(self):
+        """Turn output on."""
+        self.output_on = True
+
+    def turn_output_off(self):
+        """Turn output off."""
+        self.output_on = False
```
bilderbuchi, line 229: I have to disagree that these are "convenience functions". Also, in a pymeasure context, using properties is actually more "natural".
While we have methods in our instruments to do things that are difficult/complex with properties, we do not just mirror functionality that already exists in properties, as this is just confusing for the users. I'd rather you remove those.

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+
+    measured_current = Instrument.measurement(
+        "MEAS:CURR?", "Measured current in A."
+    )
+
+    # Arbitrary Sequence Commands ---------------------------------------------
+
+    def clear_sequence(self, channel):
+        """Clear the sequence of the selected channel."""
+        channel = strict_discrete_set(channel, [1, 2, 3, 4])
```
bilderbuchi, line 245: good way of using our validators!

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+        "Define sequence of triplets of voltage (V), current (A) and dwell "
+        "time (s).",
+        set_process=process_sequence,
+    )
+
+    repetitions = Instrument.control(
+        "ARB:REP?",
+        "ARB:REP %s",
+        "Number of repetitions (0...255). If 0 is entered, the sequence is"
+        "repeated infinitely.",
```
bilderbuchi, line 259: *indefinitely

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+        validator=strict_discrete_set,
+        values=range(256),
+        get_process=lambda x: int(x),
```
bilderbuchi, line 262: this can be replaced by `cast=int`.

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+    )
+
+    def load_sequence(self, slot):
+        """Load a saved waveform from internal memory (slot 1, 2 or 3)."""
+        slot = strict_discrete_set(slot, [1, 2, 3])
+        self.write(f"ARB:REST {slot}")
+
+    def save_sequence(self, slot):
+        """Save the sequence to internal memory (slot 1, 2 or 3)."""
```
bilderbuchi, line 271: I guess this means the "current"/active sequence (that has been defined with `sequence`)? Should we point this out in the docstring?

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+        "OUTP:GEN %s",
+        "Set the output on or off or check the output status.",
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    # The following commands are for making it easier to change selected the
+    # selected channel and activate/deactivate them.
+
+    def activate_channel(self, channel):
```
bilderbuchi, line 204: See how this and the following method duplicate all code except one Boolean? This could be united into one method, e.g. `def set_channel_active(self, channel, state):`

TBH, I am a bit unclear about the meanings of the different terms in the manual - "select", "activate" and "turn on output" seem to be three different things. It's clear that "select" designates a channel for some subsequent operations, but the difference between "activating" a (selected) channel and turning its output on is not clear to me (manual p 69)

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+    selected_channel = Instrument.control(
+        "INST:NSEL?",
+        "INST:NSEL %s",
+        "Selected channel.",
+        validator=strict_discrete_set,
+        values=[1, 2, 3, 4],
+        get_process=lambda x: int(x),
+    )
+
+    # Implemented as a more natural way of selecting a channel.
```
bilderbuchi, line 96: Actually, in our usage a method is not "more natural" than a property. In any case, having both purely for duplicating functionality is confusing, so I think the method should be dropped.

bilderbuchi (changes requested):
Thank you for your effort! I reviewed the PR and left suggestions/change requests.

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,288 @@
...
+    currents in A, and times in s. Dwell times are between 0.06 and 10 s.
+    """
+    if not len(sequence) % 3 == 0:
+        raise ValueError("Sequence must contain multiple of 3 values.")
+    if any(t > 10 or t < 0.06 for t in sequence[2:3:-1]):
+        raise ValueError("Dwell times must be between 0.06 and 10 s.")
+    # turn sequence data into a string
+    sequence = str(sequence).strip('[](){}')
+    sequence.replace(" ", "")
+    print(sequence)
```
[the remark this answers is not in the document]
bleykauf: Simply missed to remove this, no need for debug log I think.

pymeasure/instruments/rohdeschwarz/hmp.py
bleykauf: Removed the mirrored functionality.

pymeasure/instruments/rohdeschwarz/hmp.py
bleykauf: Removed duplicated functionality.

bleykauf:
I also added some instrument tests, i.e. tests where an instrument is necessary.

For this, I modified `conftest.py` and added a new marker `needs_instrument` and the possibility to specify the resource address via the command line with an command line option `--resource-name`. There is also a command line flag `--instrument-present` to indicate that an instrument is present (this might actually be a little redundant and I am not 100% happy with the way it is right now, will have another look soon).

I think in general pymeasure could benefit from such a functionality since it would allow to clearly separate tests that could be run during CI and tests that need an instrument with the possibility so specify the resource name from the command line so everybody with a specific device can run the tests. The existing tests could be adjusted accordingly.

What do you think?

bleykauf:
Opened an issue about the instrument specific tests, #632.

I would also be happy to work on this in a separate PR since it is out of scope for this PR (that just wants to add an instrument).

bilderbuchi:
> I would also be happy to work on this in a separate PR since it is out of scope for this PR (that just wants to add an instrument).

Agreed, please do split this out. In general PRs should have as small a scope as useful, as the reviewing effort goes up enormously with diff size.

bleykauf:
Simplified the tests and moved them to a separate file that is skipped by default. Ready for re-review :)

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,271 @@
...
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import (strict_discrete_set,
+                                              truncated_range)
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+def process_sequence(sequence: List[float]) -> str:
```
bilderbuchi, line 36: We [prefer](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#python-style-guides) (probably mainly for historical reasons) to put type hints into the (sphinx-style, like e.g. [here](https://github.com/pymeasure/pymeasure/blob/8086835d8a211268a26451c28a054770b3907d40/pymeasure/instruments/keithley/keithley2400.py#L387)) docstrings, not the signature/code.
Also in `set_channel_state`

bilderbuchi (changes requested):
One minor thing, then we're done here :-)

pymeasure/instruments/rohdeschwarz/hmp.py
```diff
@@ -0,0 +1,271 @@
...
+    sequence = ",".join(str(s) for s in sequence)
+    return sequence
+
+
+class HMP4040(Instrument):
+    """Represents a Rohde&Schwarz HMP4040 power supply."""
+
+    def __init__(self, resourceName, name="Rohde&Schwarz HMP4040", **kwargs):
+        super().__init__(
+            resourceName, name, includeSCPI=True, **kwargs
```
bilderbuchi, line 58:
```suggestion
    def __init__(self, adapter, **kwargs):
        kwargs.setdefault("name", "Rohde&Schwarz HMP4040")
        super().__init__(
            adapter, name, includeSCPI=True, **kwargs
```
anticipating some changes already in the pipeline :-)

pymeasure/instruments/rohdeschwarz/hmp.py
bleykauf: Moved the type hints to the docstring.

I think, however, that using the built-in type hints should be allowed. in the future. pymeasure is now Python 3.7+, they are less verbose and easier to read in plain-text format than sphinx-style docstrings and can be checked by your IDE. There is also a plugin for sphinx to make use of type hints: https://pypi.org/project/sphinx-autodoc-typehints/

bilderbuchi: Good to know, thanks!

Issue #583, opened 2022-01-29
Warning in documentation build: SwissArmyFake not found


msmttchr:
When building hml documentation an error is reported as follow:
```
Traceback (most recent call last):
  File "/home/micheles/.local/lib/python3.8/site-packages/sphinx/util/inspect.py", line 412, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: module 'pymeasure.instruments' has no attribute 'SwissArmyFake'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/micheles/.local/lib/python3.8/site-packages/sphinx/ext/autodoc/importer.py", line 110, in import_object
    obj = attrgetter(obj, mangled_name)
  File "/home/micheles/.local/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 318, in get_attr
    return autodoc_attrgetter(self.env.app, obj, name, *defargs)
  File "/home/micheles/.local/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 2606, in autodoc_attrgetter
    return safe_getattr(obj, name, *defargs)
  File "/home/micheles/.local/lib/python3.8/site-packages/sphinx/util/inspect.py", line 428, in safe_getattr
    raise AttributeError(name) from exc
AttributeError: SwissArmyFake

```
I believe this has been inadvertently introduced by PR #552, but I am not sure why it was not detected by automatic checks.
@bilderbuchi, any idea ?

bilderbuchi:
Agree on both points. Don't know how this was missed, maybe the squash merge was not built?

The reason is mist probably that we removed the fakes from the instruments namespace late in the PR, although I see a green checkmark after that.

bilderbuchi:
In any case I agree, we need to find out why this was not discovered automatically,  and fix that.

msmttchr:
Actually, looking at the build logs, the warning is there, but possibly is not considered a failure condition.

bilderbuchi:
Yeah, sphinx considers [this a warning](https://github.com/pymeasure/pymeasure/runs/4942688875?check_suite_focus=true#step:10:154). The thing is, sphinx warnings should not stop the build, I'll see if I can tune to make exceptions an error, instead.

bilderbuchi:
So, it's possible with a -W flag to have sphinx treat warnings as errors (then we could fail the build to detect this) with `make html SPHINXOPTS="-W --keep-going"`. 
[...]

PR #584, opened 2022-01-29
Add autodoc warnings to the problem matcher.

bilderbuchi:
This should enable us to at least flag autodoc
import errors in the diff view.
Catching error otherwise is not feasible, see #583.

### DIFF
--- a/.github/sphinx.json
+++ b/.github/sphinx.json
@@ -21,6 +21,18 @@
                     "message": 3
                 }
             ]
+        },
+        {
+            "owner": "sphinx-extension",
+            "pattern": [
+                {
+                    "regexp": "^(?<severity>WARNING|ERROR): (?<code>\\S+): (?<message>.+)(?<dummy_file_entry>)$",
+                    "severity": 1,
+                    "code": 2,
+                    "message": 3,
+                    "file": 4
+                }
+            ]
         }
     ]
 }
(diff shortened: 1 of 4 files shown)


bilderbuchi:
Great, I can match the warning, but it does not generate an annotation, but is only visible in the "checks" screen, where nobody notices it:
image
Maybe because it has no file location attached (because there is none) -- that cannot be hardcoded (e.g. to docs/conf.py) afaict

bilderbuchi:
OK, I don't know how to make those annotations appear in the PR diff view, or otherwise notify the user about problems.

@msmttchr  @CasperSchippers should we instead fail the build if we get any warnings in the docs step? That I could accomplish.
So far I did not want to stop the whole CI build just because of linting problems, because we could show those via annotations and not delaying pytest runs accelerates the feedback loop. If this causes us to miss warnings of incomplete docs that is a bit more severe, though.

msmttchr:
I would vote for failing the build

bilderbuchi:
OK, I'm failing the docs job now if we get any warnings, the rest of the workflow continues to run.

CasperSchippers:
I indeed agree on this; if there is no other way to ensure we don't miss issues with the docs, failing the build is the best way to go. Good work.

samcondon4:
Looks like mine was the problem PR. Yeah, I only very recently learned about how the linter reports were working so I had no idea that my commit caused any issues at the time. Good call to fail the build on this so that this doesn't happen again in the future.

PR #585, opened 2022-01-30
Fix issues #583

msmttchr:
Another option is to remove SwissArmyFake from documentation, but I am not sure whether it is better or nor.

### DIFF
--- a/docs/api/instruments/instruments.rst
+++ b/docs/api/instruments/instruments.rst
@@ -5,6 +5,6 @@ Instrument classes
 .. autoclass:: pymeasure.instruments.Instrument
     :members:
 
-.. autoclass:: pymeasure.instruments.SwissArmyFake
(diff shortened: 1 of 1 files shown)


msmttchr:
@samcondon4, could You please have a look and comment on the change ?

bilderbuchi:
Erm, I'm currently in the process of fixing this in #584? (actually have, now)

bilderbuchi:
> Another option is to remove SwissArmyFake from documentation, but I am not sure whether it is better or nor.

This fixes the current error, yes, but similarly malformed autodoc statements (for stuff that should be documented) would still slip through.

msmttchr:
> Erm, I'm currently in the process of fixing this in #584? (actually have, now)

Closing it then, easy  :)

msmttchr:
Addressed by #584

Issue #567, opened 2022-01-08
Adapter and Instrument methods


msmttchr:
As highlighted by in https://github.com/pymeasure/pymeasure/pull/559#discussion_r780658782 by @CodingMarco, there could be some design issue in the way `Instrument` and `Adapter` class implement some methods.
In particular, `values` and `ask` methods implemented in `Adapter` class are simply calling other methods or making text processing which has nothing to do with low level instrument access.
This requires, in certain circumstances, to reimplement them at `Instrument` class level even though no changes are needed.
A possible solution is to move the default implementation of these methods from `Adapter` to `Instrument`

msmttchr:
Possibly related to #512

bilderbuchi:
I see you point I think, although I'm wondering if the [problem in #559](https://github.com/pymeasure/pymeasure/pull/559#discussion_r780658782) (i.e., the need to copy the `values()` implementation) could have been avoided by encapsulating the device-specific comms requirements in a custom Adapter derived e.g. from VISAAdapter, instead of overriding the methods of `Instrument`. See also #352/ [TopticaAdapter](https://github.com/dkriegner/pymeasure/blob/3d5f57b4d8209dfa2ae32f8794eb94e1f36fa794/pymeasure/instruments/toptica/adapters.py#L31).

msmttchr:
@bilderbuchi I agree some guidelines could be extremely useful, prior to this I would make some effort to clarify the software architecture and what we target with `Instrument` and what with `Adapter` classes or may be is the same task.
Also, I have the feeling that there are too many `Adapter` subclasses and we are using mainly `VISAAdapter`.

msmttchr:
In my understanding, looking at the code, I see the following:
- Adapter class/subclasses: they allow to abstract different physical instrument interfaces, for example: VISA, Serial, Telnet, Prologix. Probably Serial and Telnet only exist to support the two variant of Prologix gpib interfaces.
- Instrument class encapsulate behaviour and control interface.

If an adapter is not available, user can develop its own and pass to `Instrument` class creator.

According to this interpretation, putting instrument behaviour/control inside an adapter could be inappropriate.

Finally, `Instrument` class normally provide a set of properties and methods to drive it and `read` and `write` methods as a way to have a lower level interface.

Some open points:
- What is the recommended paradigm to implement instrument using a binary interface (as opposed to text based interface) ?
- Are users allowed to reimplement `read` and `write` method when subclassing `Instrument`? In which cases ?
[...]

PR #586, opened 2022-02-02
Update to ITC503

CasperSchippers:
Obviously, I needed properties of the ITC503 that were not yet implemented because I thought. So now I implemented all (user-intended) commands.

This PR also includes a couple of other changes (in the group of Oxford instruments); an overview of the changes:
- Implemented additional features 
- Updated docs 
- Moved to using the OxfordInstrumentsAdapter
- Implemented preprocess_reply as suggested per PR #345 
- Implemented dynamic properties in the ITC503
- Adapted the wait_for_temperature (of the ITC503) and the wait_for_idle (of the IPS120-10) to the OxfordInstrumentsAdapter
- Fix a few issues in the OxfordInstrumentsAdapter I came across when implementing it in the ITC503 (related to querying the status and the version)

One thing I want to note: when using a dynamic property (in this case the temperature_setpoint property, of which I wanted to change the values parameter), it is not possible to change just one of the values in case values is a mutable. e.g. in this case I wanted to change either the maximum or minimum value you can set to the temperature_setpoint, but changing one of the two values (using `self.temperature_setpoint_values[1] = new_maximum`) requires getting the value of this attribute, which is forbidden for this kind of special attributes. This can be kind of annoying, is this something we want to keep like this?


msmttchr:
> One thing I want to note: when using a dynamic property (in this case the temperature_setpoint property, of which I wanted to change the values parameter), it is not possible to change just one of the values in case values is a mutable. e.g. in this case I wanted to change either the maximum or minimum value you can set to the temperature_setpoint, but changing one of the two values (using `self.temperature_setpoint_values[1] = new_maximum`) requires getting the value of this attribute, which is forbidden for this kind of special attributes. This can be kind of annoying, is this something we want to keep like this?

@CasperSchippers, this restriction was introduced to limit risk of people using some of the reserved names as attribute name and face surprising side effects. I think the probability is low, but impact is a bit annoying.
A possible recommendation, we could draft in these cases, is to get the entire `values` parameter, modify it and reassign to the `values` parameter, but I wonder if there is a proper way to get the limits defined by a property (regardless whether it is dyanmic or not).

pymeasure/instruments/oxfordinstruments/itc503.py
```diff
@@ -26,17 +26,38 @@
...
+    :param values: A range of values (passed to strict_range)
+    :raises: TypeError if the value is not a tuple or a list
+    :raises: IndexError if the value is not of length 2
+    """
+
+    if not isinstance(value, (list, tuple)):
+        raise TypeError('{:g} is not a list or tuple'.format(value))
+    if not len(value) == 2:
+        raise IndexError('{:g} is not of length 2'.format(value))
+    return tuple(strict_range(v, values) for v in value)
```
bilderbuchi, line 58: It could be argued that `strict_range` itself should work like this -- I was a bit confused when I realised that it takes the min and max of the collection, not just first and second value. ^.^

bilderbuchi: The question is, though, if we should change it now, it's surely been in use for a long time.

pymeasure/instruments/oxfordinstruments/itc503.py
```diff
@@ -54,220 +75,417 @@ class ITC503(Instrument):
...
+        if clear_buffer:
+            self.adapter.connection.clear()
+
+        self.temperature_setpoint_values = [min_temperature, max_temperature]
+
+    class FLOW_CONTROL_STATUS(IntFlag):
+        """ IntFlag class for decoding the flow control status. Contains the following
+        flags:
+
+        === ======================  ==============================================
```
bilderbuchi, line 118: nice table formatting!

bilderbuchi (approved):
LGTM, I left a couple of comments.

bilderbuchi:
> I wonder if there is a proper way to get the limits defined by a property (regardless whether it is dyanmic or not).

Not afaik for the non-dynamic case, as the property effectively is a bag of defined functions.

The question is, should we account for the use case of modifying only _one_ element of `values` (is that a frequent desire), or should we leave the read-restriction in place for now and wait what kind of user feedback we receive? I lean toward the latter, but not firmly.

pymeasure/instruments/oxfordinstruments/itc503.py
```diff
@@ -54,220 +75,417 @@ class ITC503(Instrument):
...
+                asrl={
+                    'baud_rate': 9600,
+                    'data_bits': 8,
+                    'parity': 0,
+                    'stop_bits': 20,
+                },
+                preprocess_reply=lambda v: v[1:],
+                **kwargs,
+            )
+
```
msmttchr, line 101: In my opinion, this approach will prevent users with a Prologix GPIB adapter to control the instrument, since part of the instrument behaviour is embedded in the `OxfordInstrumentsAdapter` which requires `VISAAdapter`.
Perhaps this is intended, but I just wanted to double check.
I think it is important to clarify the role of `Adapter` and `Instrument` classes (see also #567).

bilderbuchi: I think we should at least try and see if we can get Prologix into VISA, see https://github.com/pymeasure/pymeasure/issues/567#issuecomment-1024996200, that would resolve that issue most neatly.

msmttchr: If we succeed in having Prologix inside VISA (which is good), the entire `Adapter` classes are no longer needed, since all interfaces are handled by PyVISA. But, if we want to keep the `Adapter` classes, we need to define the scope and purpose of it. Of course, clarifying the role of `Adapter` is important anyway, regardless that Prologix is integrated or not in PyVISA.

bklebel: It was certainly not intended to restrict controlling this instrument to the `VISAAdapter` - I have to admit, I just started with it since I use pyvisa, and did not see this problem. I included my thoughts on this in [#567 (comment)](https://github.com/pymeasure/pymeasure/issues/567#issuecomment-1038857075).

CasperSchippers: That's actually quite a good point. I'm not sure what is the best approach: should we try to fix this now, or hold this off until #567 is sorted out (in which case, should we wait with merging until it is resolved or not)?

bilderbuchi: > I'd be inclined to leave it as it is now, merge it and modifying whatever's needed when https://github.com/pymeasure/pymeasure/issues/567 is sorted out.

Agreed. Other instruments also have the problem of not being able to use PrologixAdapter, which we will resolve with #567. Having this merged has the added benefit that a PR solving #567 has this instrument as a ready example in the codebase, to guide the implementation (as you immediately see how it works/"feels" in practice)

bklebel: I guess we could merge this now, since it conforms with the current way of doing things (which is sub-optimal as discussed in #567), and I think moving the custom adapter to a mixin class either on the `Adapter` or `Instrument` level will not be a lot of work, so we can then transition this one as an example, and add a prominent documentation of how to implement quirks of instruments/manufacturers.

msmttchr: I agree with the proposal to merge, a bit of pragmatism over high level design discussions sometime is needed.
If the outcome of #567 will requires some further adaption, it could be done at that time, but for the time being people could enjoy the enhancements.

bilderbuchi: Let's do it, then!

CasperSchippers: Perfect, thanks!

PR #587, opened 2022-02-06
Customizable error query as instrument constructor argument

CodingMarco:
This PR adds the option to specify a custom command for querying errors.
For example, the HP 54540A needs the command `SYST:ERR? STR` instead of just `SYST:ERR?` to return the error number and error description.
Since the `check_errors()` method contains quite a bit of logic, I would argue that it is cleaner to specify just the command instead of copying the whole method to a subclass with just the command changed.

### DIFF
--- a/pymeasure/instruments/instrument.py
+++ b/pymeasure/instruments/instrument.py
@@ -119,6 +119,7 @@ class Instrument:
     :param adapter: A string, integer, or :py:class:`~pymeasure.adapters.Adapter` subclass object
     :param string name: The name of the instrument. Often the model designation by default.
     :param includeSCPI: A boolean, which toggles the inclusion of standard SCPI commands
+    :param error_query: The string to be sent to the instrument to query for errors.
     :param \\**kwargs: In case ``adapter`` is a string or integer, additional arguments passed on
         to :py:class:`~pymeasure.adapters.VISAAdapter` (check there for details).
         Discarded otherwise.
@@ -145,7 +146,7 @@ class Instrument:
     __reserved_prefix = "___"
 
     # noinspection PyPep8Naming
-    def __init__(self, adapter, name, includeSCPI=True, **kwargs):
+    def __init__(self, adapter, name, includeSCPI=True, error_query="SYST:ERR?", **kwargs):
         try:
             if isinstance(adapter, (int, str)):
                 adapter = VISAAdapter(adapter, **kwargs)
@@ -156,6 +157,7 @@ class Instrument:
         self.name = name
         self.SCPI = includeSCPI
         self.adapter = adapter
+        self.error_query = error_query
 
         self.isShutdown = False
         self._special_names = self._setup_special_names()
@@ -519,7 +521,7 @@ class Instrument:
         if self.SCPI:
             errors = []
             while True:
-                err = self.values("SYST:ERR?")
+                err = self.values(self.error_query)
                 if int(err[0]) != 0:
                     log.error(f"{self.name}: {err[0]}, {err[1]}")
                     errors.append(err)
--- a/pymeasure/instruments/instrument.py
+++ b/pymeasure/instruments/instrument.py
@@ -144,9 +144,6 @@ class Instrument:
     # Prefix used to store reserved variables
     __reserved_prefix = "___"
 
-    # The command to be sent to the instrument to query for errors
-    _error_query = "SYST:ERR?"
-
     # noinspection PyPep8Naming
     def __init__(self, adapter, name, includeSCPI=True, **kwargs):
         try:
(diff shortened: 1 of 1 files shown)


msmttchr:
I think this PR could address several cases similar to this. I recently had same problem with Anritsu MS2668C. I also think that same problem could happen for other commands (e.g. reset, etc), so I would evaluate to structure the parameter in a way similar to the one proposed by @bilderbuchi in https://github.com/wnluke/pymeasure/blob/f03315b3ae9149370ac4f6b00652e08ce8656add/pymeasure/adapters/visa.py#L53

bilderbuchi:
Huh, I was under the impression that the command format is prescribed/defined in the SCPI standard. Anyway, good idea to expose just the error string, that indeed makes implementing deviating instruments cleaner. 👍 

However, this error string is always the same for a given instrument model. Therefore, I argue that is should not be in the constructor signature - after all, when would a user instantiating an instrument ever pass that argument?

Instead, what do you think about a private attribute `Instrument._error_query`, that is set to the default value in `Instrument` as a class attribute. (private because it will no be useful to a user of the class)
Then, any derived instrument needing a different value just needs to set the desired value (as a class attribute or in the ctor, does not matter).

CodingMarco:
Thank you both for your suggestions. I have now implemented it as a class attribute as suggested by @bilderbuchi.
@msmttchr did you have a problem with the error query or with another IEEE 488 root command? As of now, I haven't made the other root commands customizable because
1. the methods don't contain any logic and can be implemented in a subclass without much code duplication
2. I think it's very unlikely that these root commands are different for an instrument supporting SCPI

msmttchr:
> * the methods don't contain any logic and can be implemented in a subclass without much code duplication

I think this make sense and also the comment of @bilderbuchi about standard compliance is very relevant. so I think this instrument is not following the rules.
I still see the change specific to `check_errors()` method still very specific, but I understand the logic.
Another option could be to implement an `error` method command (in a similar fashion of `reset` or `clear`, etc.), have the `check_errors` using `error` method. In such way, it is much more obvious to users how to customize the `error` command by reimplementing the `error` method.

This could have also the advantage that, if we decide to go down the road with an `SCPIInstrument `class using dynamic properties, changes will be easier.

msmttchr (approved):
Looks good to me

pymeasure/instruments/instrument.py
```diff
@@ -528,3 +528,13 @@ def check_errors(self):
             return errors
         else:
             raise NotImplementedError("Non SCPI instruments require implementation in subclasses")
+
+    def next_error(self):
+        """ Read the next error from the instrument.
+
+        :return: error entry
+        """
```
bilderbuchi, line 536: I suggest making this a private method, as users should always use  `check_errors()` for consistency.

```suggestion
    def _next_error(self):
        """ Read the next error from the instrument.
        
        Internal method -- use :meth:`check_errors` to query for error status.
        Reimplement this method in instruments using a different command.

        :return: error entry
        """
```

bilderbuchi (changes requested):
I have one request, otherwise LGTM. Can you also check if there are instruments already in the codebase that could use your improved method of customization (to make sure our codebase reflects current best practice).

pymeasure/instruments/instrument.py
msmttchr: In my opinion, `next_error` is consistent with all the methods already present for SCPI commands. On the other hand `check_errors` is a utility function that extract all the errors and log them.
There could be several use cases:
- SCPI instruments: retrieve all errors using check_errors.
- SCPI instruments: you just need next error, use next_error
- Non SCPI instruments: customize next_error or check_errors or both as needed.

bilderbuchi: To provide context: I want to make sure we don't add public API to _all_ our instruments without a clear use case for it. I have seen in other code bases how messy an unchecked proliferation of public API in an inheritance tree can become, and I _very much_ want to avoid that here.

msmttchr:
As a use case, I can imagine a situation where the instrument return an
error after a specific command and may be you need to implement some
recovery action.
So in this case, you don't want to log this error condition, but just check
if a specific error condition happen, so assuming your error queue is
empty, you just want to read next error.

Really my point is from a consistency point of view, SCPI defines this
logic based on error queue, so next_error offer access to the related SCPI
command, and this would be consistent to other defined SCPI commands.

On top of this basic command, check_errors is a utility which allow to read
and log instrument errors presumably, but not necessarily, at the end of an
experiment and return them as a list.

Il Ven 11 Feb 2022, 14:34 Christoph Buchner ***@***.***> ha
scritto:

msmttchr:
This, in my opinion a strong and valid argument, which I share.

In the specific case, the next error is a standard command, so I would not
consider it an exhotic addition, but just a way to keep consistency with
what it is already implemented.

Il Ven 11 Feb 2022, 14:39 Christoph Buchner ***@***.***> ha
scritto:

bilderbuchi:
OK, then, you're more familiar with SCPI standard-related things, so I'll defer to your judgement -- i ou still think it's a good idea we can go ahead with `next_error()`. 
However, the point about converting currently customized methods to the "new way" still stands before we merge this.

BenediktBurger:
Implemented in #1024 . Thanks for the discussion.

PR #588, opened 2022-02-11
Phase control of Tektronix AFG3152C

FlyingCurryMonster:
I was using the AFG3152C and noticed I couldn't control the phase of the signal with pymeasure, so I added in Instrument.control method for that.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -82,3 +82,4 @@ Till Zürner
 J. A. Wilcox
 Nick James Kirkby
 Konrad Gralher
+David Ziliak
--- a/docs/api/instruments/hp/hp8753E.rst
+++ b/docs/api/instruments/hp/hp8753E.rst
@@ -1,0 +1,8 @@
+################################
+HP 8753E Vector Network Analyzer
+################################
+
+
+.. autoclass:: pymeasure.instruments.hp.HP8753E
+    :members:
+    :show-inheritance:
--- a/docs/api/instruments/hp/index.rst
+++ b/docs/api/instruments/hp/index.rst
@@ -18,6 +18,7 @@ If the instrument you are looking for is not here, also check :doc:`Agilent<../a
    hp8116A
    hp856xx
    hp8657B
+   hp8753E
    hp11713A
    hp437B
    hplegacyinstrument
--- a/pymeasure/instruments/hp/__init__.py
+++ b/pymeasure/instruments/hp/__init__.py
@@ -30,6 +30,7 @@ from .hp8116a import HP8116A
 from .hp8657b import HP8657B
 from .hp856Xx import HP8560A
 from .hp856Xx import HP8561B
+from .hp8753e import HP8753E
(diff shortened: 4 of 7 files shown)


pymeasure/instruments/tektronix/afg3152c.py
```diff
@@ -114,6 +120,22 @@ class Channel:
...
+        validator = strict_range,
+        values = PHASE_LIMIT['RAD']
+    )
+    
+    phase_deg = Instrument.control(
+        "phase:adjust?", "phase:adjust %e DEG",
+        """ A floating point property that controls the phase in radian units.
+        This property can be set.""",
+        validator=strict_range,
+        values = PHASE_LIMIT['DEG']
```
LongnoseRob, line 137: Probably a simple c&p mistake, for this property the text should be changed to `"...  in degrees."`

LongnoseRob (commented):
Thank you for the addition of the phase-control properties!

I found a small mistake in the doc-string for the `phase_deg` property.

pymeasure/instruments/tektronix/afg3152c.py
FlyingCurryMonster: Oops yea it was a copy and paste mistake.  And the string is actually different for phase_rad and phase_deg.  phase_deg ends with %e DEG, while phase_rad ends with %e RAD.

bilderbuchi: no, the _query_ string, meaning the `get_command` argument is the same -- what's up with that? what happens if you do 
```
print(instr.phase_deg)
print(instr.phase_rad)
```
?

bilderbuchi:
The linter found some formatting issues (check in the changed files view, or run flake8 locally)

pymeasure/instruments/tektronix/afg3152c.py
FlyingCurryMonster: It looks like phase_deg won't return the phase in degrees unless it's been set previously to degrees.  This also seems to be an issue with the get commands of amp_vpp and amp_vrms (haven't tried amp_dbm).  My AFG is being used in an ongoing measurement right now, but I wonder if just added the units RAD/DEG like so:

'phase:adjust RAD?'
'phase:adjust DEG?' 

would give the query back in the desired units.

FlyingCurryMonster: also apologies for the late replies, I've been working on my final required graduate course for my degree program.

bilderbuchi: Good to have this confirmed. Such statefulness in instruments/command sets is always annoying.
To answer your question, I guess you need to study the manual and/or try it out when the instrument is free again.

BenediktBurger:
Hey @FlyingCurryMonster , do you want to pick this PR up again or not?

FlyingCurryMonster:
For sure, apologies I've been busy with apparatus development.

PR #594, opened 2022-02-17
Breakout widgets into separate modules.

mcdo0486:
Breakout widgets into separate modules. Auto-import current modules at pymeasure/display/widgets/__init__.py for backwards compatibility


msmttchr:
> Breakout widgets into separate modules. Auto-import current modules at pymeasure/display/widgets/**init**.py for backwards compatibility

Thanks for your proposal. I like the idea. I would suggest to run `flake8` in your local PC before pushing the changes, otherwise you could have a lot of iterations before converging (this happened to me as well).

mcdo0486:
> > Breakout widgets into separate modules. Auto-import current modules at pymeasure/display/widgets/**init**.py for backwards compatibility
> 
> Thanks for your proposal. I like the idea. I would suggest to run `flake8` in your local PC before pushing the changes, otherwise you could have a lot of iterations before converging (this happened to me as well).

Thanks for that tip. I think all linting issues are caught now, I'll definitely use flake8 next time.

msmttchr (commented):
@mcdo0486, this looks good to me, even though I am not entirely sure I can understand all the ramifications of the changes.
I wonder, if there could be any issue of backward compatibility. Do you have any comment on this ?
Before merging, I would prefer to have a second opinion from other maintainers.

msmttchr:
@CasperSchippers and @bilderbuchi, at your convenience, could you please have a look at this PR?

bilderbuchi:
Yes I will. It's the "convenience" that I'd hard to come by. 😝

CasperSchippers:
I like the idea of splitting the file into separate modules.

I'll try having a look tomorrow evening (sorry for the wait, as mentioned already, time is hard to come by).

One concern I can already raise from a git point of view: splitting a file causes us to lose the history/blame of the code, right? I think there are ways to ensure we keep the history, though from top of my head I don't know how easy/difficult that is and if it's worth the effort (also a question to @bilderbuchi and @msmttchr); I myself do sometimes use the history, but not very often.

bilderbuchi:
Git blame tracks moved/renamed files. There is a [flag](https://stackoverflow.com/a/10664943/599884) to track moved code portions across files, as well:  [`-C`](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt--Cltnumgt).
In any case, if I had to choose, I'd rather argue for better-organised code even at the cost of some history features.

pymeasure/display/widgets/__init__.py
```diff
@@ -0,0 +1,12 @@
+from .browser_widget import BrowserWidget
```
CasperSchippers, line 1: Could you also add the copyright text over here

CasperSchippers (approved):
I've had a thorough look and looks good to me; no issues that I could find. Only question is to add the copyright text to the init file

msmttchr (approved):

pymeasure/display/widgets/__init__.py
mcdo0486: Yup pushed that up

PR #595, opened 2022-02-22
DPSeriesMotorController bug fix

samcondon4:
In the overriden `ask()`, `write()`, and `values()` methods I was using `self.address` to prepend command strings with the address of the motor controller. This causes collisions when multiple controllers are connected to the same rs485 line. Thus I have changed this to `self._address`.


msmttchr (approved):
Ok for me. Thanks for your contribution

PR #596, opened 2022-02-22
Fixed broken links in documentation.

samcondon4:
Ran a sphinx `make linkcheck` and fixed all of the reported broken links in the documentation.


msmttchr (approved):
Very useful fixes for the documentation.
I wonder whether this can be applied to our CI checks. @bilderbuchi any opinion about this?

bilderbuchi:
In principle yes, but it then is dependent on website availability, so I expect the tests to be quite flaky. Also, I don't know how well the CI machines can call out into the internet. 
Overall, I think it should be enough to run this manually from time to time and fix some things that broke.

Thanks for the fixes!

PR #598, opened 2022-03-02
Updated widget documentation

mcdo0486:
This adds:
1. Widget modules to `docs/api/display/widgets.rst` documentation file.
1. Missing docstrings to widget classes
1. Spelling and linking fixes to existing widget class docstrings


mcdo0486:
The github actions seems to have bugged out as it errored on "Python 3.7, windows-latest" `tests\experiment\test_workers.py:144: AssertionError`

bilderbuchi:
Let's rerun and check if it is a flaky test, that test would be a prime candidate. Would be interesting why it starts to fail now, though, as it has been working flawlessly for years (?) now.

msmttchr (approved):
Very useful/handy updates. LGTM.

pymeasure/display/widgets/results_dialog.py
```diff
@@ -37,6 +37,14 @@
 
 
 class ResultsDialog(QtGui.QFileDialog):
+    """
+    Widget that displays a dialog box for loading a past experiment run.
+    It shows a preview of curves from the results file when selected in the dialog box.
+
+    This widget is implemented in the `open_experiment` method in
```
CasperSchippers, line 44: Maybe rather "This widget is **used by** the `open_experiment` method in"

CasperSchippers (approved):
Thanks for this improvement; generally LGTM (added just a single linguistic remark).

What about this test that failed suddenly, is that something we need to do something with or is it good as it is as it runs fine now?

pymeasure/display/widgets/results_dialog.py
mcdo0486: I pushed up that change.

And to the CI test, it just failed in the same pytest for workers, at a different line:
`tests/experiment/test_workers.py:111: AssertionError`

bilderbuchi:
I suspect that something about this PR sometimes exposes a race condition or somesuch in our ZMQ handling or threading setup. We have to investigate and fix that before merging.

bilderbuchi:
@msmttchr before merging let's re-run the whole suite, not just the failing case, to be on the safe side.

bilderbuchi:
argh, a manually triggered re-run still [uses](https://github.com/pymeasure/pymeasure/pull/598/checks#step:3:473) the old master HEAD, so the fix from #601 is not in. 
@mcdo0486 can you merge the current master branch into the PR branch and push?

msmttchr:
Thanks for the contrubution

msmttchr:
> @msmttchr before merging let's re-run the whole suite, not just the failing case, to be on the safe side. (note: I just did)

Sorry, I missed your message. In any case, the merge should be ok.

PR #600, opened 2022-03-14
New usbtmc adapter for communication with usb instruments based on th…

pierricmora:
New usbtmc adapter for communication with usb instruments based on the python-usbtmc library. The motivation is to provide an alternative option when pyvisa-py fails to communicate with an usb instrument.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -45,3 +45,4 @@ Bastian Leykauf
 Matthew Delaney
 Marco von Rosenberg
 Jack Van Sambeek
+Pierric Mora
--- a/docs/api/adapters.rst
+++ b/docs/api/adapters.rst
@@ -67,6 +67,16 @@ Telnet adapter
     :show-inheritance:
 
 ============
+USBTMC adapter
+============
+
+.. autoclass:: pymeasure.adapters.USBTMCAdapter
+    :members:
+    :undoc-members:
+    :inherited-members:
+    :show-inheritance:
+
+============
 Fake adapter
 ============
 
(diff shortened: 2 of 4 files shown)


bilderbuchi:
Hi there! I'd be curious to know about the problems you encountered with pyvisa-py. It would be great if they could be avoided in another way (e.g. by using pyvisa itself), as we have been drifting towards _fewer_ Adapters, not more, and also because usbtmc is essentially unmaintained now (last commit 2017).
(Ideally, it's better to discuss such things in an issue first to avoid unnecessary work creating a PR that won't get merged in the end).

pierricmora:
Hi!
Ok I surely understand your point, sorry for not discussing it first. My problem is similar with that one being discussed here:
[https://github.com/pyvisa/pyvisa-py/issues/256](https://github.com/pyvisa/pyvisa-py/issues/256)
My instrument is one of the Keysight 33500B arb generator family. Apparently other Keysight / Agilent / Tektronix instruments give similar problems to pyvisa-py. The issue in pyvisa-py was opened 6 years ago. It is not solved yet (although seems to be active again now). Installing other visa backends can be real pain in linux, I finally got ni-visa compiled but it did not solve my issue. Usbtmc makes the job. I perfectly agree that it would be much better to have a pyvisa-py fix. So, should this PR be canceled and should I keep my lazy solution for myself, or should it be made a temporary alternative? Which one is less worse?

bilderbuchi:
I would prefer not to pull this in, "temporary" things quickly become non-temporary/relied upon, so I'd prefer if you kept it for yourself for the time being.
Actually, seeing as you have a problematic device at hand, and the skills to write PRs, maybe you are in a good position to help out over at that pyvisa-py issue? From skimming the discussion, it seems what Matthieu is lacking is 
* a (wireshark?) communication log for a device that does not work with pyvisa-py, but works with usbtmc.
* someone to submit a PR for review when the problem has been pinpointed.

bilderbuchi:
No need to be sorry! Thank you for the interest, and good luck!

PR #601, opened 2022-03-15
Test workers

mcdo0486:
A recent github action ran into an error when running pytest for the test function `test_zmq_does_not_crash_worker`  in [`tests/experiment/test_workers.py`](https://github.com/pymeasure/pymeasure/blob/135711821d3a4c5f8de709161411303f301c4a16/tests/experiment/test_workers.py#L96). The `RandomProcedure` was given 4.0 seconds to iterate over 100 randomly generated numbers with a delay of 0.001 seconds.

For the [Python 3.7, macos-latest](https://github.com/pymeasure/pymeasure/runs/5485426432?check_suite_focus=true#logs) action runner, it was only able to reach 65 out of 100 numbers

This pull request Increases the timeout from 4 or 5 seconds to 20 seconds `worker.join(timeout=20.0)`.
This will allow slow github action runner containers to complete the tests.

I replicated the timeout issue in this github repo:
https://github.com/mcdo0486/action_test/runs/5545060857?check_suite_focus=true#step:8:466

When I increased the number of seconds from 4 or 5 seconds to 20 seconds, the issue didn't resurface across 18 action runs.

There is a different issue where `test_zmq_topic_filtering_works` in [`tests/experiment/test_workers.py`](https://github.com/pymeasure/pymeasure/blob/135711821d3a4c5f8de709161411303f301c4a16/tests/experiment/test_workers.py#L118) times out while waiting for data to be received. This pull doesn't fix that issue.


msmttchr:
@mcdo0486, I think this fix is very important, without it in place we cannot push forward any PR since the CI will fail.
Do you have any idea why this is happening now ? I saw in the past that the timeout has been increased from 1 second to 4 second and now a further increase is needed. I think this test with [Python 3.7, macos-latest](https://github.com/pymeasure/pymeasure/runs/5485426432?check_suite_focus=true#logs) should be in place since long time. Possibly macos version has changed ?
Is it better to increase the timeout to 180 seconds, just to be sure we do not meet this problem again, in this case successful run will not be delayed, but failing run will need to wat for 2 minutes before showing the problem ? Or is there any possibility to change the test structure ?

msmttchr:
> 
> There is a different issue where `test_zmq_topic_filtering_works` in [`tests/experiment/test_workers.py`](https://github.com/pymeasure/pymeasure/blob/135711821d3a4c5f8de709161411303f301c4a16/tests/experiment/test_workers.py#L118) times out while waiting for data to be received. This pull doesn't fix that issue.

Is it possible to log it in the project issues ? Or is it already present ?

mcdo0486:
Why is this happening now? I'm not sure, maybe it is a change with Github action runner infrastructure, like a change to the scheduler or the demand for macos-latest has grown. In my testing, I found that my commits on Saturday morning didn't fail as often as commits during middle of a workday. So maybe it is a load issue.

Maybe running a non-blocking threaded test will always be problematic on github action runner, where the time it takes for work to be done on a spawned thread is highly variable because of process scheduling.

However, after increasing the sleep and timeout, there hasn't been a failure* in the past 60 tests I've run here: https://github.com/mcdo0486/action_test/actions

*there are two failures at a different step, with github rate limiting when install pymeasure is installed. But that isn't related.

msmttchr:
> Why is this happening now? I'm not sure, maybe it is a change with Github action runner infrastructure, like a change to the scheduler or the demand for macos-latest has grown. In my testing, I found that my commits on Saturday morning didn't fail as often as commits during middle of a workday. So maybe it is a load issue.
> 
> Maybe running a non-blocking threaded test will always be problematic on github action runner, where the time it takes for work to be done on a spawned thread is highly variable because of process scheduling.
> 
> However, after increasing the sleep and timeout, there hasn't been a failure* in the past 60 tests I've run here: https://github.com/mcdo0486/action_test/actions
> 
> *there are two failures at a different step, with github rate limiting when install pymeasure is installed. But that isn't related.

Ok, thanks for the clarification. I guess your changes are on to me, but possibly @bilderbuchi who is managing the CI can give a definitive green light.

msmttchr (commented):
LGTM, but if this is the only option to make the test passing, I would evaluate to further increase the timeout.

mcdo0486:
> LGTM, but if this is the only option to make the test passing, I would evaluate to further increase the timeout.

That sounds good to me, should I push up a change with increased values? Let me know what values, there are three functions where I changed values. Generally I multiplied previous value by four:

`worker.join()`
  - timeout from 4 or 5 to 20
  - This is max time for a worker thread to run before joining it

`listener.message_waiting()`
  - timeout from .1 to 4
  - This is max time for zmq.Poller() to poll for events before returning

`sleep()`
  - sleep from .5 to 4
  - Leave time for subscriber and publisher to establish a connection

msmttchr:
> > LGTM, but if this is the only option to make the test passing, I would evaluate to further increase the timeout.
> 
> That sounds good to me, should I push up a change with increased values? Let me know what values, there are three functions where I changed values:
> 
> `worker.join()`
> 
> * timeout from 4 or 5 to 20
> * This is max time for a worker thread to run before joining it
> 
> `listener.message_waiting()`
> 
> * timeout from .1 to 4
> * This is max time for zmq.Poller() to poll for events before returning
> 
> `sleep()`
> 
> * sleep from .5 to 4
> * Leave time for subscriber and publisher to establish a connection

I have checked the `RandomProcedure` and apparently runs 100 loops with 1 ms delay, this should results in a nominal time of 0.1 seconds (not including the random number generation), so I guess the original value (4 sec) was assumed to be safe and moving to 20 should handle all the reasonable delays, I think. For the other delays, I am not sure. Let's wait a little bit mode for @bilderbuchi to comment further or otherwise we can proceed and merge the changes, since they can only improve the situation.

bilderbuchi:
I'll review this, hopefully on the weekend.

bilderbuchi:
Thank you for the detailed investigation, also with your separate CI runs, this is quite thorough! It's always hard to debug/diagnose these threading problems. 

Yes, I had originally increased the timeout already, did not think it would not be enough :D (as you can see from the comment).

The changes are straightforward, restricted to testing code, and it seems like this does not mask an underlying issue (fingers crossed!), just accounts for slowness on the CI hosts, so let's merge this.

PR #607, opened 2022-04-05
Add numpy's bool_ to the sequencer

neal-kepler:
This addresses the edge case where you might want to do an array of booleans in the sequencer. As far as I tried
``[True, True, False]`` does not play nicely with the output type of the BooleanInput (a numpy style bool). A hack is to do something like ``[bool(1),bool(1),bool(0)]`` by allowing bool as a safe funtion.


neal-kepler:
I just realized that I may have overlooked a better way to do this

Issue #512, opened 2021-11-05
Restructure Adapter communication methods


bilderbuchi:
If we restructured `Adapter` and subclasses so that we can put common functionality into `Adapter`, that delegates to subclasses' specific implementations, we could gain several useful features without duplicating across all adapters:
* Logging of commands going/coming from adapters. (#219)
* Thread safety measures (#506)
* Delays between subsequent writes or reads (#430)
* Maybe unify read/write termination (https://github.com/pymeasure/pymeasure/pull/430#issuecomment-830623044)

InstrumentKit uses an approach where in the base class there are [concrete methods](https://github.com/Galvant/InstrumentKit/blob/6d216bd7f8e9ec7918762fe5fb7a306d5bd0eb1f/instruments/abstract_instruments/comm/abstract_comm.py#L168-L236) that delegate to [abstract methods](https://github.com/Galvant/InstrumentKit/blob/6d216bd7f8e9ec7918762fe5fb7a306d5bd0eb1f/instruments/abstract_instruments/comm/abstract_comm.py#L111-L158) that are overridden by subclasses.
The concrete methods can have common functionality (like logging) in one place, so they can observe DRY.

Issue #506, opened 2021-11-02
Thread safe communication


svihra:
Dear all,

I was trying to create a control tool on top of the Keithley2450 which would continuously take data while the user/code could apply changes. I created an update method which is providing the [readback/measured/status] values in one go. While testing, I ran into an issue where the internal methods are not written thread-safe, breaking the run. Debugging further (at least in case of this particular device) I was able to locally implement thread locks in the underlying Instrument class that provides the communication.

That said, I think the best place would be to modify the VISAAdapter class to ensure the thread-safe behaviour (which I think I should be able to do myself and create a merge request). Is there something that I miss and could cause major issues elsewhere?

bilderbuchi:
Note that pyvisa is [thread-safe](https://pyvisa.readthedocs.io/en/latest/faq/faq.html?highlight=thread#is-pyvisa-thread-safe), so the communication to the instrument should already be unproblematic. Where are you stumbling? I'd be interested in a repro procedure (ideally without needing a Keithley2450 :D).

svihra:
First of all, these are the library versions:
```
Python 3.7.11
pymeasure                 0.9              py37hf985489_2    conda-forge
pyvisa                    1.11.3           py37hf985489_1    conda-forge
pyvisa-py                 0.5.2              pyh44b312d_0    conda-forge
```


I have not tested with a dummy device but I think the issue is caused by the real lag in the communication.
The dummy code would be:

```
import threading, asyncio, time
from pymeasure.instruments.keithley import Keithley2450

class k2450(Keithley2450):
    def __init__(IP):
        super().__init__(f"TCPIP0::{IP}::inst0::INSTR")
        print(f"-> Connected {self.id}")
        self.loop = asyncio.get_event_loop()
        self._thread = threading.Thread(target=self.loop.run_forever).start()

    async def update(self):
        while True:
            self.ask('READ? "defbuffer1", SOUR, READ, SOURSTAT') # defbuffer1 is the default k2450 buffer name, this is the only way I found how to get back the source readback, measurement and source status values in one go
            await asyncio.sleep(0.25)

dev = k2450("1.1.1.1")
asyncio.run_coroutine_threadsafe(dev.update(), dev.loop) # placed in a separate loop to provide user access to the rest of the code

for _ in range(10):
    print(dev.source_voltage) # this should do an independent reading from the one in update, mimicking user access
    time.sleep(1)
```

[...]

PR #608, opened 2022-04-07
Adapter thread safety

samcondon4:
This PR addresses the thread safety measures discussed in #512 and #506. This is not a full restructure of the `Adapter` class, but rather I've simply added an attribute, `connection_lock` of type `threading.Lock`. I've also overridden the `__getattribute__` and `__setattr__` methods so that the lock is held for the duration of attribute access. In this way, multiple threads can read/write from the same instrument as all `Adapter` attributes are protected by this lock.

### DIFF
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -122,7 +122,8 @@ class Adapter:
     def __getattribute__(self, name):
         """ Override to acquire the 'connection_lock' lock through the duration of attribute access.
         """
-        if name not in ("connection_lock", "connection_lock_initialized") and self.connection_lock_initialized:
+        if name not in ("connection_lock", "connection_lock_initialized") \
+                and self.connection_lock_initialized:
             with self.connection_lock:
                 ret = super().__getattribute__(name)
         else:
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -22,6 +22,7 @@
 # THE SOFTWARE.
 #
 
+import threading
 import numpy as np
 from copy import copy
 
@@ -38,9 +39,13 @@ class Adapter:
     :param kwargs: all other keyword arguments are ignored.
     """
 
+    connection_lock_initialized = False
+
     def __init__(self, preprocess_reply=None, **kwargs):
         self.preprocess_reply = preprocess_reply
         self.connection = None
+        self.connection_lock = threading.Lock()
+        self.connection_lock_initialized = True
 
     def __del__(self):
         """Close connection upon garbage collection of the device"""
@@ -113,6 +118,26 @@ class Adapter:
         """
         raise NameError("Adapter (sub)class has not implemented the "
                         "binary_values method")
+
+    def __getattribute__(self, name):
+        if not name == "connection_lock" and object.__getattribute__(self, "connection_lock_initialized"):
+            with object.__getattribute__(self, "connection_lock"):
+                ret = object.__getattribute__(self, name)
+        else:
+            ret = object.__getattribute__(self, name)
+
+        return ret
+
+    def __setattr__(self, name, value):
+        if not name == "connection_lock" and object.__getattribute__(self, "connection_lock_initialized"):
+            with object.__getattribute__(self, "connection_lock"):
+                object.__setattr__(self, name, value)
+
+        elif name == "connection_lock" and object.__getattribute__(self, "connection_lock_initialized"):
+            raise ValueError("Adapter connection_lock cannot be modified!")
+
+        else:
+            object.__setattr__(self, name, value)
 
 
 class FakeAdapter(Adapter):
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -120,6 +120,8 @@ class Adapter:
                         "binary_values method")
 
     def __getattribute__(self, name):
+        """ Override to acquire the 'connection_lock' lock through the duration of attribute access.
+        """
         if not name == "connection_lock" and object.__getattribute__(self, "connection_lock_initialized"):
             with object.__getattribute__(self, "connection_lock"):
                 ret = object.__getattribute__(self, name)
@@ -129,6 +131,8 @@ class Adapter:
         return ret
 
     def __setattr__(self, name, value):
+        """ Override to acquire the 'connection_lock' lock through the duration of attribute setting.
+        """
         if not name == "connection_lock" and object.__getattribute__(self, "connection_lock_initialized"):
             with object.__getattribute__(self, "connection_lock"):
                 object.__setattr__(self, name, value)
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -122,7 +122,8 @@ class Adapter:
     def __getattribute__(self, name):
         """ Override to acquire the 'connection_lock' lock through the duration of attribute access.
         """
-        if not name == "connection_lock" and object.__getattribute__(self, "connection_lock_initialized"):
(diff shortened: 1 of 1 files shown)


pymeasure/adapters/adapter.py
```diff
@@ -114,6 +119,33 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
...
+                object.__getattribute__(self, "connection_lock_initialized"):
+            with object.__getattribute__(self, "connection_lock"):
+                object.__setattr__(self, name, value)
+
+        elif name == "connection_lock" and \
+                object.__getattribute__(self, "connection_lock_initialized"):
+            raise ValueError("Adapter connection_lock cannot be modified!")
+
+        else:
+            object.__setattr__(self, name, value)
```
msmttchr, line 147:
```suggestion
            super().__setattr__(name, value)
```
Why not using `super` ?

pymeasure/adapters/adapter.py
```diff
@@ -114,6 +119,33 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
...
+        else:
+            ret = object.__getattribute__(self, name)
+
+        return ret
+
+    def __setattr__(self, name, value):
+        """ Override to acquire the 'connection_lock' lock through the duration of attribute setting.
+        """
+        if not name == "connection_lock" and \
+                object.__getattribute__(self, "connection_lock_initialized"):
```
msmttchr, line 138:
```suggestion
                self.connection_lock_initialized:
```

pymeasure/adapters/adapter.py
```diff
@@ -114,6 +119,33 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
...
+            with object.__getattribute__(self, "connection_lock"):
```
msmttchr, line 139:
```suggestion
            with self.connection_lock:
```

pymeasure/adapters/adapter.py
```diff
@@ -114,6 +119,33 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
...
+                object.__setattr__(self, name, value)
```
msmttchr, line 140:
```suggestion
                setattr(self, name, value)
```

pymeasure/adapters/adapter.py
```diff
@@ -114,6 +119,33 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
...
+
+        elif name == "connection_lock" and \
+                object.__getattribute__(self, "connection_lock_initialized"):
```
msmttchr, line 143:
```suggestion
                self.connection_lock_initialized:
```

pymeasure/adapters/adapter.py
```diff
@@ -114,6 +119,33 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
         raise NameError("Adapter (sub)class has not implemented the "
                         "binary_values method")
 
+    def __getattribute__(self, name):
```
msmttchr, line 122:
```
def __getattribute__(self, name):
        """ Override to acquire the 'connection_lock' lock through the duration of attribute access.
        """
        if not name in ("connection_lock, "connection_lock_initialized") and self.connection_lock_initialized:
            with self.connection_lock:
                ret = super().__getattribute__(name)
        else:
            ret = super().__getattribute__( name)
```

msmttchr (changes requested):
Hi, I have put some suggested changes to simplify code, could you please check them?
I think it would needed also to add some test to show that everything is working as expected.

pymeasure/adapters/adapter.py
samcondon4: I like this version better as well. I've changed it to this.

pymeasure/adapters/adapter.py
samcondon4: In this case, using `setattr` leads to `RecursionError: maximum recursion depth exceeded` since everytime `setattr` is called, the overridden version of `__setattr__` is called. Similar to in the new version of `__getattribute__` I believe we should use `super().__setattr__()`.

pymeasure/adapters/adapter.py
samcondon4: Yep, this is definitely better. I've changed it to use `super()`

pymeasure/adapters/adapter.py
samcondon4: I was worried about a `RecursionError` with this at first, but with the `if name not in ("connection_lock", "connection_lock_initialized")` logic, this works just fine :)

pymeasure/adapters/adapter.py
samcondon4: This works as well.

pymeasure/adapters/adapter.py
msmttchr: Indeed. Sorry I overlooked this.

msmttchr (commented):
@samcondon4, I have reviewed changes you made and they are ok to me. On the other hand I tried to figure out what use cases are addressed by this PR. Could you please try to list them and add associated tests ? I was unable to find use cases on my side.

pymeasure/adapters/adapter.py
```diff
@@ -38,9 +39,13 @@ class Adapter:
     :param kwargs: all other keyword arguments are ignored.
     """
 
+    connection_lock_initialized = False
```
bilderbuchi, line 42: I'm wondering: why do you need this separate sentinel attribute that is effectively always true? 
You are creating the lock when initialising the `Adapter` - is it not enough to do that as the first action in `__init__`, and then just use the now existing lock? We would only need to make sure that all subclasses call the superclass init first thing.

To be a bit more defensive, you could set `self.connection_lock = None` here at class level, and check with `is None` (or just `and self.connection_lock`) in the attribute methods, or does that not work?

pymeasure/adapters/adapter.py
```diff
@@ -114,6 +119,29 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
         raise NameError("Adapter (sub)class has not implemented the "
                         "binary_values method")
 
+    def __getattribute__(self, name):
+        """ Override to acquire the 'connection_lock' lock through the duration of attribute access.
+        """
+        if name not in ("connection_lock", "connection_lock_initialized") \
```
bilderbuchi, line 125: I'm curious about the choice to intercept _all attributes_, when the stated intention (and lock name) indicate that you only want to lock on _connection_ access. I'd be more comfortable with a tighter-scoped version where you only put a lock around `self.connection`, which is what we care about. 
Otherwise, you are also locking e.g. `PrologixAdapter.address`, which does not make sense.

pymeasure/adapters/adapter.py
```diff
@@ -114,6 +119,29 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
...
+                and self.connection_lock_initialized:
+            with self.connection_lock:
```
bilderbuchi, line 127: Good to use the with-statement variant here!

bilderbuchi (changes requested):
Thank you for the PR! I left two comments.
In addition to that, I agree that we absolutely need some tests to confirm that this works as expected, and does not have unintended side effects. You can e.g. subclass a `ConnectionTestAdapter` with some minimal implementation to write some unit tests.

Also, I'm wondering if this actually does what you intend to do. Specifically, we are only locking _attribute access_, but all the communication works happen in _calls_ of attribute methods. 
I don't know off the top of my hat, but I suspect that, when you call `self.connection.write()`, Python _getting_ `self.connection` is protected, but _not_ the subsequent _method call_. Some tests should confirm that.

PR #613, opened 2022-04-10
PEP517 compatibility & dynamically obtaining a version number

bilderbuchi:
There were requests to get version numbers that include git information (e.g. for development builds). Instead of vendoring versioneer (see #574), I found out that setuptools_scm can do this, too. At the same time, finally moving to pyproject.toml-style modern packaging will simplify things, so that will be the first part of this PR.

Don't pull this yet, I'm probably going to rebase to keep history clean.


bilderbuchi:
Huh, I'm actually not sure why we're shipping that many files in the source distribution (examples, docs, etc), but I've kept that state for now. Maybe it's useful to ship more files than strictly necessary for other consumers of the tarball, like Linux distributions (not that anyone is packaging pymeasure for those)?

bilderbuchi:
Well that was a "nice" excursion into the depths of Python packaging. :D

It's finished and quite flexible now, can deal with pymeasure being not installed at all, editably-installed, or conventionally installed. Basically, if `setuptools_scm` is installed, you get a current version string computed at runtime, otherwise it falls back to metadata from installation.

@bleykauf I'd be obliged if you could test if this fits the use case you described in https://github.com/pymeasure/pymeasure/discussions/566

setup.cfg
```diff
@@ -1,5 +1,49 @@
...
+    numpy >= 1.6.1
+    pandas >= 0.14
+    pyvisa >= 1.8
+    pyserial >= 2.7
+    pyqtgraph >= 0.9.10
+    importlib-metadata; python_version<"3.8"
+
+[options.extras_require]
+matplotlib = matplotlib>=2.0.2
+tcp =
```
msmttchr, line 39: Is matplotlib needed by pymeasure ?

msmttchr (approved):
@bilderbuchi, wonderful feature (long awaited). From what I was able to test, it works nicely. Thank You. I left just a minor question on matplotlib.

CasperSchippers (approved):
Looks good to me! Thanks for this feature.

setup.cfg
bilderbuchi: Hm, I copied/moved over what was already there. That's maybe a leftover from earlier. Nowadays we require pandas, and probably assume matplotlib will be there (but it's not required by pandas).
Question to @CasperSchippers : is matplotlib mandatory? Does all our (experiment/GUI code) code work (meaning: could also fall back to something else if mpl is not found)? 
If optional, but needed to get plotting capability: should we rename that extra require target/feature selector `plotting` to be clearer?

bleykauf:
> @bleykauf I'd be obliged if you could test if this fits the use case you described in #566 as soon as I fix the last build errors and CI is green.

Sorry for the late reply: I just tested and it does indeed work as expected. Actually, I like the produced version strings better, since they follow PEP440 more closely than versioneer that I suggested in #566. Thank you!

I had no issues installing with `pip install -e .` . I guess the reason for still having the `setup.cfg` around is that support for defining everything in the `pyproject.toml` is compatability with older versions of pip and setuptools? Note that https://setuptools.pypa.io/en/stable/userguide/pyproject_config.html states that a `setup.py` might still be necessary for editable installs (it is not at least for the latest verson of pip).

bilderbuchi:
Thanks for the review!
> I guess the reason for still having the setup.cfg around is that support for defining everything in the pyproject.toml is compatability with older versions of pip and setuptools?

Partly that, and the warning on the linked page: 
> Support for declaring [project metadata](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/) or configuring setuptools via pyproject.toml files is still experimental and might change (or be removed) in future releases.

I read somewhere in my investigation that the part about setup.py still being necessary is not true any more since a recent pip version (I think [21.3](https://pip.pypa.io/en/latest/news/#v21-3)). 
We are pulling latest pip in the CI, I'm not sure if/how we should prescribe a minimum pip version for the editable install feature only, or should just wait if there are bug reports, first.

setup.cfg
msmttchr: I think the only place where `matplotlib` is used is here https://github.com/pymeasure/pymeasure/blob/0400c37c9fa474afe8fbddb041de6f5bb708a259/pymeasure/display/curves.py#L36

This is optional and used to import the colormap

bilderbuchi: ...is used _directly_. What I did not have the time to check, yet, is what happens if we want to use pymeasure GUI plotting in an environment that does not have matplotlib installed.

msmttchr: The complete code is as follow;
```
try:
    from matplotlib.cm import viridis
except ImportError:
    log.warning("Matplotlib not found. Images will be greyscale")
```
So when `matplotlib` is not available, GUI will still work and the `ResultsImage` will be grey scale.

In any case, I have submitted a PR #622 to clean all this and use colormap from `pyqtgraph`

bilderbuchi: I just merged #622 and will remove the matplotlib extras_require entry.

bilderbuchi: Huh, I double-checked, and there is [another use](https://github.com/pymeasure/pymeasure/blob/7369ff12cd969b744035839e443ee8894e1d3b56/pymeasure/experiment/config.py#L48) of matplotlib. 
However, this only gets hit if a user has defined an experiment config entry "matplotlib.rcParams", so presumably then they'll have matplotlib installed anyway. 

I think considering this, and the fact that the minimum version we require is very old by now, I think it is still safe to drop the explicit target.

bilderbuchi:
:shipit:

PR #621, opened 2022-05-07
New instruments: HC Photonics crystal ovens TC038, TC038D

bmoneke:
I added two crystal ovens from HC Photonics (HCP).

I tried to follow the "Adding instruments" guideline and looked at different present devices.

Both files have been tested.

A question of curiosity: Is this logging good or relevant? I found it in another file.

### DIFF
--- a/pymeasure/instruments/hcp/tc038.py
+++ b/pymeasure/instruments/hcp/tc038.py
@@ -1,0 +1,139 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2022 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import logging
+
+from pymeasure.instruments import Instrument
+import pyvisa
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class TC038(Instrument):
+    """
+    Communication with the HCP TC038 oven.
+
+    This is the older version with an AC power supply and AC heater.
+
+    It has parity or framing errors from time to time. Handle them in your
+    application.
+    """
+
+    def __init__(self, resourceName, address=1, timeout=1000):
+        """
+        Initialize the communication.
+
+        Parameters
+        ----------
+        resourceName : str
+            name COM-Port.
+        address : int
+            address of the device. Should be between 1 and 99.
+        timeout : int
+            Timeout in ms.
+        """
+        super().__init__(resourceName, "TC038", timeout=timeout,
+                         write_termination="\r", read_termination="\r",
+                         parity=pyvisa.constants.Parity.even,)
+        self.address = address
+
+        self.monitorTemperature()  # start to monitor the temperature
+
+    def write(self, command):
+        """
+        Send "command" to the oven with "commandData".
+
+        Parameters
+        ----------
+        command : string, optional
+            Command to be sent, three chars. The default is "WRM".
+        commandData : string, optional
+            Additional data for the command. The default is "".
+        """
+        super().write(chr(2) + f"{self.address:02}" + "010"
+                      + command + chr(3))
+        # 010 is CPU (01) and time to wait (0), which are fix
+
+    def ask(self, command):
+        """
+        Send a command to the oven and read its response.
+
+        Parameters
+        ----------
+        command : string, optional
+            Command to be sent, three chars. The default is "WRM".
+        commandData : string, optional
+            Additional data for the command. The default is "".
+
+        Returns
+        -------
+        string
+            response of the system.
+
+        """
+        return super().ask(chr(2) + f"{self.address:02}" + "010"
+                           + command + chr(3))
+
+    def monitorTemperature(self):
+        """Configure the oven to monitor the current temperature."""
+        self.ask(command="WRS" + "01" + "D0002")
+        # WRS in order to setup to monitor a word
+        # monitor 1 word
+        # monitor the word in register D0002
+
+    @property
+    def setpoint(self):
+        """Read and return the current setpoint in °C."""
+        got = self.ask(command="WRD" + "D0120" + ",01")
+        # WRD: read words
+        # start with register D0003
+        # read a single word, separated by space or comma
+        return self.dataToTemp(got)
+
+    @property
+    def temperature(self):
+        """Read and return the current temperature in °C."""
+        got = self.ask(command="WRD" + "D0002" + ",01")
+        return self.dataToTemp(got)
+
+    def getMonitored(self):
+        """Read and return the monitored value.
+        Normally it's the current temperature in °C."""
+        # returns the monitored words
+        got = self.ask(command="WRM")
+        return self.dataToTemp(got)
+
+    @setpoint.setter
+    def setpoint(self, setpoint=27):
+        """Set the setpoint to a temperature in °C."""
+        commandData = f"D0120,01,{int(round(setpoint*10)):04X}"
+        # Temperature without decimal sign in hex representation
+        got = self.ask(command="WWR" + commandData)
+        return got[5:7]
+
+    def dataToTemp(self, data):
+        """Convert the returned hex value "data" to a temperature in °C."""
+        return int(data[7:11], 16) / 10.
+        # get the hex number, convert to int and shift the decimal sign
--- a/pymeasure/instruments/hcp/tc038.py
+++ b/pymeasure/instruments/hcp/tc038.py
@@ -22,13 +22,8 @@
 # THE SOFTWARE.
 #
 
-import logging
-
 from pymeasure.instruments import Instrument
-import pyvisa
-
-log = logging.getLogger(__name__)
-log.addHandler(logging.NullHandler())
+from pyvisa.constants import Parity
 
 
 class TC038(Instrument):
@@ -41,7 +36,8 @@ class TC038(Instrument):
     application.
     """
 
-    def __init__(self, resourceName, address=1, timeout=1000):
+    def __init__(self, resourceName, address=1, timeout=1000,
+                 includeSCPI=False):
         """
         Initialize the communication.
 
@@ -56,7 +52,7 @@ class TC038(Instrument):
         """
         super().__init__(resourceName, "TC038", timeout=timeout,
                          write_termination="\r", read_termination="\r",
-                         parity=pyvisa.constants.Parity.even,)
+                         parity=Parity.even,)
         self.address = address
 
         self.monitorTemperature()  # start to monitor the temperature
@@ -68,9 +64,8 @@ class TC038(Instrument):
         Parameters
         ----------
         command : string, optional
-            Command to be sent, three chars. The default is "WRM".
-        commandData : string, optional
-            Additional data for the command. The default is "".
+            Command to be sent. Three chars indicating the type, and data for
+            the command, if necessary.
         """
         super().write(chr(2) + f"{self.address:02}" + "010"
                       + command + chr(3))
@@ -83,9 +78,8 @@ class TC038(Instrument):
         Parameters
         ----------
         command : string, optional
-            Command to be sent, three chars. The default is "WRM".
-        commandData : string, optional
-            Additional data for the command. The default is "".
+            Command to be sent. Three chars indicating the type, and data for
+            the command, if necessary.
 
         Returns
         -------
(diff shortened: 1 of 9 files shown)


bilderbuchi:
> A question of curiosity: Is this logging good or relevant? I found it in another file.

If you are using logging in the application, and want your instrument to communicate something not by return value or exception, it can be useful, and definitely preferable over print statements.
You are not using it in the implementation, though, so there's no need to keep it around.

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -0,0 +1,136 @@
...
+
+    functions = {'read': 0x03, 'writeMultiple': 0x10,
+                 'writeSingle': 0x06, 'echo': 0x08}
+
+    def __init__(self, resourceName, address=1, timeout=1000):
+        """Initialize the device."""
+        super().__init__(resourceName, "TC038D", timeout=timeout)
+        self.address = address
+
+    def CRC16(self, data):
```
bilderbuchi, line 53: This could be generally useful; it would be good to have some tests that document expected input/output of this function. Doesn't CRC have different "variants" that could slightly change outcomes? If so, which one is that?
btw, do you know about Python's `binascii.crc_hqx()`?

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -0,0 +1,136 @@
...
+            for j in range(8):
+                lsb = CRC & 0x1  # least significant bit
+                CRC = CRC >> 1
+                if lsb:
+                    CRC ^= 0xA001
+        return [CRC & 0xFF, CRC >> 8]
+
+    def readRegister(self, address, count=2):
+        """Read count variables from start address on."""
+        # Count has to be double the number of elements in 4-byte-mode.
```
bilderbuchi, line 67: Is there a way to make this mode configurable (e.g. with a predefined class-level attribute like `byteMode` below), so that manual details like this do not have to be "remembered", and you could use natural "counts"?

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -0,0 +1,136 @@
...
+        data = [self.address]
+        data.append(0x03)  # function code
```
bilderbuchi, line 69: `self.functions[read]` would be clearer and not duplicate that binary code, no?
Same below.

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -0,0 +1,136 @@
...
+        data += [address >> 8, address & 0xFF]  # 2B address
+        data += [count >> 8, count & 0xFF]  # 2B number of elements
+        data += self.CRC16(data)
+        self.adapter.connection.write_raw(bytes(data))
+        got = self.adapter.connection.read_bytes(3)
+        # Slave address, function, length
+        if got[1] == 0x03:
+            length = got[2]
+            read = self.adapter.connection.read_bytes(length + 2)
+            # data, CRC
```
bilderbuchi, line 79: Should this check CRC consistency before returning only the data?

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -0,0 +1,136 @@
...
+        """Write multiple variables."""
+        data = [self.address]
+        data.append(0x10)  # function code
+        data += [address >> 8, address & 0xFF]  # 2B address
+        if isinstance(values, int):
+            data += [0x0, byteMode // 2]  # 2B number of elements
+            data.append(byteMode)  # 1B number of write data
+            for i in range(byteMode - 1, -1, -1):
+                data.append(values >> i * 8 & 0xFF)
+        elif hasattr(values, "__iter__"):
```
bilderbuchi, line 99: `isinstance(values, collections.abc.Iterable)`?

bilderbuchi (changes requested):
A first pass with some suggestions. 
Great that you implemented this frame-based command scheme! 
We intended for quite some time to make this more generic, usable for more instruments, see #296, #239. For that at least it's useful as a baseline (so I'm not saying I expect this to be done!).

pymeasure/instruments/hcp/tc038.py
```diff
@@ -0,0 +1,139 @@
...
+        # monitor the word in register D0002
+
+    @property
+    def setpoint(self):
+        """Read and return the current setpoint in °C."""
+        got = self.ask(command="WRD" + "D0120" + ",01")
+        # WRD: read words
+        # start with register D0003
+        # read a single word, separated by space or comma
+        return self.dataToTemp(got)
```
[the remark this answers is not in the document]
bmoneke: I did choose to implement my own ask / read methods, which contain some more command forming (adding signs in front and back of the command). To ensure, that these new ask /read methods are called, I made my own properties.

I guess I could make some get_process method which adds these extra code. On the other hand, I need some flexibility, as these extra bytes take the device address:
```
def write(self, command):
    super().write(chr(2) + f"{self.address:02}" + "010"
                         + command + chr(3))```

File: pymeasure/instruments/hcp/tc038d.py (hunk 1 above)
Comment (bmoneke) [reply pymeasure/instruments/hcp/tc038d.py:53]: Yeah, finding the right variant is not easy. Therefore I had chosen to implement a method according to the manuals instruction (which describes a step by step procedure), that way I know, that it will work.

**Comment by bmoneke** [comment]
@bilderbuchi  thanks for your helpful comments.
I'll try to apply them to the code.

File: pymeasure/instruments/hcp/tc038.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (bilderbuchi) [reply pymeasure/instruments/hcp/tc038.py:113]: The pymeasure properties use read/write/ask in the background, so this is already accounted for. :-)

File: pymeasure/instruments/hcp/tc038d.py (hunk 1 above)
Comment (bilderbuchi) [reply pymeasure/instruments/hcp/tc038d.py:53]: It would be good to compare the output of your function with `binascii.crc_hqx()`. For that (and generally), it would be useful if your `CRC16` were a stand-alone function, it does not use the class' state (i.e. `self`), anyway.
Either extract it to method level (preferred), or add a ` @staticmethod` decorator so you can run it without creating a `TC038` instance first. 

If behaviour is identical you can use that one from the Python standard library, instead.

File: pymeasure/instruments/hcp/tc038.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (bmoneke) [reply pymeasure/instruments/hcp/tc038.py:113]: Implementing the change, Spyder marks "self" in the get_process as unknown. Now I remember, that this was the reason I opted for python properties.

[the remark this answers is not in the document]
Comment (bilderbuchi) [reply pymeasure/instruments/hcp/tc038.py]: > Spyder marks "self" in the get_process as unknown

Hah, of course! That's what I get for copy-pasting. :D
` _data_to_temp` and `_temp_to_data` should be standalone functions (you see they are not using their `self` argument). Best pull them out to module level (before the instrument definition), and drop the `self.` from the invocation.

Conceptually, they should be `@staticmethod`s (and accessed with `TC038._data_to_temp`), but the problem with that is that when we are invoking `Instrument.control`, the class definition for `TC038` is not yet complete, so you can't access those functions this way. A module-level function is the next best thing. If you only have one use, and the function is simple, a lambda is preferrable as you end up with less stuff in the module namespace.

File: pymeasure/instruments/hcp/tc038.py [hunk 7]
```diff
@@ -0,0 +1,139 @@
...
+
+    @property
+    def temperature(self):
+        """Read and return the current temperature in °C."""
+        got = self.ask(command="WRD" + "D0002" + ",01")
+        return self.dataToTemp(got)
+
+    def getMonitored(self):
+        """Read and return the monitored value.
+        Normally it's the current temperature in °C."""
```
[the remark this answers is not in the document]
bilderbuchi: What I meant is: Under which circumstances is it _not_ the current temperature? Document how to achieve that.

pymeasure/instruments/hcp/tc038.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        return super().ask(chr(2) + f"{self.address:02}" + "010"
+                           + command + chr(3))
+
+    def monitor_temperature(self):
+        """Configure the oven to monitor the current temperature."""
+        self.ask(command="WRS" + "01" + "D0002")
+        # WRS in order to setup to monitor a word
+        # monitor 1 word
+        # monitor the word in register D0002
```
bilderbuchi, line 98: A suggestion how this could be more generic (if in fact you can monitor other things than temperatures). The name reflects the association to `monitored_value`.

```suggestion
    def set_monitored_quantity(self, quantity='temperature'):
        """Configure the oven to monitor a certain register, by default the current temperature.        
        
        Valid entries are "temperature",...
        """
        monitored_registers: {'temperature': 'D002', 'something_else': 'D123',...}
        self.ask(command="WRS" + "01" + monitored_registers[quantity])
        # WRS in order to setup to monitor a word
        # monitor 1 word
        # monitor the word in the desired register
```

bmoneke: You could monitor any value of the temperature controller, but I did not find anything useful to monitor besides the temperature.
Accessing the monitored value is a little faster and easier than accessing the values directly (you just request the monitored value, which is a single command, instead of requesting a specific value which requires a longer message). I do not see the benefit of reading the setpoint, or other internal configuration, with a fast readout.

Nevertheless, I edited it according to your suggestions, with a dictionary which could be expanded containing more addresses.

pymeasure/instruments/hcp/tc038d.py
bmoneke: Just a comment, why length + 2

pymeasure/instruments/hcp/tc038d.py
bmoneke: That was the way I found to look for an iterable. Is it preferred to use the collections?

pymeasure/instruments/hcp/tc038d.py
bilderbuchi: OK, don't worry about it then.

pymeasure/instruments/hcp/tc038d.py
bilderbuchi: Either way is fine (although there are slight differences we probably won't hit) -- I guess maybe I'm personally more inclined towards isinstance.

In any case, though, are you really looking for an iterable, here? You are using `len()` on `values`, but not every iterator has a length (e.g. `count()`). Are you not looking for a *sequence*, instead?

pymeasure/instruments/hcp/tc038d.py
bilderbuchi: yeah, but you are getting a CRC sum back in the message, should you not check if the payload came back intact/fits the CRC?

bmoneke:
I just tried using the pymeasure properties for tc38, but they do not invoke my ask method, therefore they skip the additional chars and communication fails.
Maybe I did miss something.

bmoneke:
@bilderbuchi thank you very much for all your comments on my code. It's a great way to learn how to write better code.

bilderbuchi:
> I just tried using the pymeasure properties for tc38, but they do not invoke my ask method, therefore they skip the additional chars and communication fails.
Maybe I did miss something.

That is unexpected. I'll see if I can find out what is going wrong ~tonight. AFAICT the code looks correct, your overridden `write`and `ask`should be used. Did you try to debug/step through the execution?

bmoneke:
> That is unexpected. I'll see if I can find out what is going wrong ~tonight. AFAICT the code looks correct, your overridden `write`and `ask`should be used. Did you try to debug/step through the execution when you use a property?

The error message showed, that the super ask method was called, but not my own (verified with a print command in my own, which was not executed). Maybe the properties are generated before the overriden methods are created as part of the instance?

bilderbuchi:
I found it! You _also_ have to override `Instrument.values` as this is what the properties' fget/fset use internally.
This should do the trick:
```
        """ Reads a set of values from the oven with an adapted command."""
        return super().values(
            chr(2) + f"{self.address:02}" + "010" + command + chr(3),
            **kwargs
        )
```
Sorry I forgot about this :-/ 

I see you're familiar with force-pushing/revising commit sets - if you're successful maybe you can clean up the commit history to remove the now (hopefully) unnecessary revert.

pymeasure/instruments/hcp/tc038.py
```diff
@@ -0,0 +1,135 @@
...
+            the command, if necessary.
+        """
+        # 010 is CPU (01) and time to wait (0), which are fix
+        return chr(2) + f"{self.address:02}" + "010" + command + chr(3)
+        # Response is chr(2) + address:02 + "01" + response + chr(3)
+
+    def write(self, command):
+        """Send a "command" in its own protocol and clear the buffer."""
+        # The oven responds, use therefore ask to clear the buffer.
+        self.ask(command)
```
bilderbuchi, line 85: watch out, an  `ask` is (effectively) a write+read; this should be afaict `super().write(self._adjust_command(command)))`

pymeasure/instruments/hcp/tc038.py
```diff
@@ -0,0 +1,135 @@
...
+            Timeout in ms.
+        """
+        super().__init__(resourceName, "TC038", timeout=timeout,
+                         write_termination="\r", read_termination="\r",
+                         parity=Parity.even,)
+        self.address = address
+
+        self.set_monitored_quantity()  # start to monitor the temperature
+
+    def _adjust_command(self, command):
```
bilderbuchi, line 71: Great that you extract a method to avoid repetition!
This does not use `self`, so it should be better drop that and decorate it with `@staticmethod`.
```suggestion
    @staticmethod
    def _adjust_command(command):
```

pymeasure/instruments/hcp/tc038.py
```diff
@@ -0,0 +1,135 @@
...
+        self.ask(command)
+
+    def ask(self, command):
+        """Send a command to the oven and read its string response."""
+        got = super().ask(self._adjust_command(command))
+        if got[5:7] != "OK":
+            raise ConnectionError(f"Ask failed with message {got}")
+        return got
+
+    def values(self, command):
```
bilderbuchi, line 94: Watch out, with this override you changed the function signature/behaviour compared to what I suggested - you should forward any kwargs, too!

bilderbuchi (changes requested):

pymeasure/instruments/hcp/tc038.py
bmoneke: I know. The problem is, that the device responds to a change. Therefore, if I set a variable, an acknowledgment will sit in the buffer. For the next query, this acknowledgement is returned instead of the desired response.

The setters invoke write, and therefore the buffer has to be cleared afterwards. Maybe I could write "clear the read buffer" and hint to another (write_only or so) method, which does not read.

pymeasure/instruments/hcp/tc038.py
bmoneke: It does use `self` in the format string. That is exactly the reason, why I had to do this extra work.

bilderbuchi: Ha, shows I haven't had enough coffee, yet. :-/

pymeasure/instruments/hcp/tc038.py
bilderbuchi: If I don't misunderstand, the `check_set_errors`/`check_get_errors` toggles are there for exactly that behaviour -- you implement an overridden [`check_errors`](https://pymeasure.readthedocs.io/en/latest/api/instruments/instruments.html#pymeasure.instruments.Instrument.check_errors) method (base one [here](https://github.com/pymeasure/pymeasure/blob/master/pymeasure/instruments/instrument.py#L514)) that checks such an acknowledgement (and also responds when it is not "OK").

We should probably extend [the docs](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#writing-properties) to explain more about the various options?

bmoneke: Oh, I did not think about using error checking as a read method. Documentation is geared towards SCPI and I did not want to "request" the errors. Nevertheless, the documentation ist good enough, that I directly understood how to make it work. It's just a task to get into such a framework and all of its possibilities.

It is not optimal, because the same check_errors can not be applied to reading, but nevertheless, it works.
Now a meaningful error message is given for the ConnectionError.

bilderbuchi:
Now the only thing remaining is to fix the linter errors in tc038.rst and tc038d.rst -- there is an overlay in the files view telling you what's wrong :-)

bmoneke:
That's done. Thank you very much for your thorough review.

bilderbuchi:
Thank you for your patience, and the instrument implementation! 🚀

PR #628, opened 2022-05-23
Fleshed out version of ProtocolAdapter.

bmoneke:
I implemented a version of the ProtocolAdapter, which can read and write according to the message_pairs.
This implementation maintains the message pairs throughout, and only the read/write is done in bytes.

Tests are added for the normal read/write.

Frame-based communication needs new methods (I used instr.adapter.connection.read_bytes in the tc038d) in the adapters themselves in order to write and read bytes. Therefore we have to decide on the adapter implementation first, before doing to much work regarding the ProtocolAdapter Version. For that reason, I did not add tests for those methods.

In order to use some adapter configuration (here `preprocess_reply`, as I learnt writing tests for another driver), I had to let `Instrument` open the ProtocolAdapter (with the adaptername "test", but the modality can be changed).

### DIFF
--- a/pymeasure/adapters/protocol.py
+++ b/pymeasure/adapters/protocol.py
@@ -39,10 +39,7 @@ def to_bytes(command):
     elif isinstance(command, str):
         return command.encode("utf-8")
     elif isinstance(command, (list, tuple)):
-        try:
-            return bytes(command)
-        except TypeError:
-            raise
+        return bytes(command)
     elif isinstance(command, (int, float)):
         return str(command).encode("utf-8")
     raise TypeError(f"Invalid input of type {type(command).__name__}.")
@@ -52,8 +49,8 @@ class ProtocolAdapter(Adapter):
     """ Adapter class testing command exchange without instrument hardware.
 
     :param list of lists comm_pairs: List of message pairs. First message is
-        the written one, the second one is the read message, it is optional.
-        'None' indicates that this message (write or read) does not exist.
+        the write one, the second one is the read message.
+        'None' indicates that a pair member (write or read) does not exist.
     :param kwargs: TBD key-word arguments
     """
 
@@ -66,46 +63,30 @@ class ProtocolAdapter(Adapter):
         self._write_buffer = b""
         self.comm_pairs = comm_pairs
         self._index = 0
-        # TODO: Make this skeleton implementation workable
 
     def write(self, command):
-        """Compare the command with the expected one and fill the buffer."""
-        pair = self.comm_pairs[self._index]
-        self._index += 1
-        assert to_bytes(pair[0]) == self._write_buffer + to_bytes(command), (
-                f"Command '{command}' written, but {pair[0]} expected.")
-        self._write_buffer = b""
-        try:
-            self._read_buffer = to_bytes(pair[1])
-        except IndexError:
-            # No response in the pair.
-            self._read_buffer = b""
+        """Compare the command with the expected one and fill the read."""
+        self.write_bytes(to_bytes(command))
+        assert self._write_buffer == b"", (
+            f"Written bytes '{self._write_buffer}' do not match expected "
+            f"'{self.comm_pairs[self._index][0]}'.")
 
     def write_bytes(self, content):
         """Write the bytes `content`. If a command is full, fill the read."""
         self._write_buffer += content
-        pair = self.comm_pairs[self._index]
-        if self._write_buffer == to_bytes(pair[0]):
-            assert self._read_buffer == b"", "Responses have not been read yet."
+        p_write, p_read = self.comm_pairs[self._index]
+        if self._write_buffer == to_bytes(p_write):
+            assert self._read_buffer == b"", (
+                f"Unread response '{self._read_buffer}' present when writing. "
+                "Read the response or enable the property's 'check_set_errors'")
             # Clear the write buffer
             self._write_buffer = b""
+            self._read_buffer = to_bytes(p_read)
             self._index += 1
-            try:
-                self._read_buffer = to_bytes(pair[1])
-            except IndexError:
-                self._read_buffer = b""
 
     def read(self):
-        """Read the prepared read bufferr and return it as a string."""
-        if self._read_buffer:
-            buffer = self._read_buffer.decode("utf-8")
-            self._read_buffer = b""
-            return buffer
-        else:
-            pair = self.comm_pairs[self._index]
-            assert pair[0] is None, "Unexpected read without prior write."
-            self._index += 1
-            return to_bytes(pair[1]).decode("utf-8")
+        """Return an already present or freshly fetched read buffer as a string."""
+        return (self.read_bytes(-1) + self.read_bytes(1)).decode("utf-8")
 
     def read_bytes(self, count):
         """Read `count` number of bytes."""
@@ -114,11 +95,11 @@ class ProtocolAdapter(Adapter):
             self._read_buffer = self._read_buffer[count:]
             return read
         else:
-            pair = self.comm_pairs[self._index]
-            assert pair[0] is None, "Unexpected read without prior write."
+            p_write, p_read = self.comm_pairs[self._index]
+            assert p_write is None, "Unexpected read without prior write."
+            self._read_buffer = to_bytes(p_read)[count:]
             self._index += 1
-            self._read_buffer = to_bytes(pair[1])[count:]
-            return to_bytes(pair[1])[:count]
+            return to_bytes(p_read)[:count]
 
     # TODO: Harmonise ask being write+read (i.e., remove from VISAAdapter),
     #   use protocol tests to confirm it works correctly
--- a/pymeasure/adapters/protocol.py
+++ b/pymeasure/adapters/protocol.py
@@ -30,15 +30,88 @@ log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 
+def to_bytes(command):
+    """Change `command` to a bytes object"""
+    if isinstance(command, (bytes, bytearray)):
+        return command
+    elif command is None:
+        return b""
+    elif isinstance(command, str):
+        return command.encode("utf-8")
+    elif isinstance(command, (list, tuple)):
+        try:
+            return bytes(command)
+        except TypeError:
+            raise
+    elif isinstance(command, int):
+        return bytes((command,))
+    raise TypeError("Invalid input")
+
+
 class ProtocolAdapter(Adapter):
     """ Adapter class testing command exchange without instrument hardware.
 
     :param kwargs: TBD key-word arguments
     """
 
-    def __init__(self, preprocess_reply=None, **kwargs):
+    def __init__(self, comm_pairs, preprocess_reply=None, **kwargs):
         super().__init__(preprocess_reply=preprocess_reply)
+        self.read_buffer = b""
+        self.write_buffer = b""
+        self.comm_pairs = comm_pairs
+        self.index = 0
         # TODO: Make this skeleton implementation workable
+
+    def write(self, command):
+        """Compare the command with the expected one and fill the buffer."""
+        pair = self.comm_pairs[self.index]
+        self.index += 1
+        assert to_bytes(pair[0]) == to_bytes(command), (
+                f"Command '{command}' written, but {pair[0]} expected.")
+        print(pair)
+        try:
+            self.read_buffer = to_bytes(pair[1])
+        except IndexError:
+            # No response in the pair.
+            self.read_buffer = b""
+
+    def write_bytes(self, content):
+        """Write the bytes `content`. If a command is full, fill the read."""
+        self.write_buffer += content
+        pair = self.comm_pairs[self.index]
+        if self.write_buffer == pair[0]:
+            assert self.read_buffer == b"", "Responses have not been read yet."
+            # Clear the write buffer
+            self.write_buffer = b""
+            self.index += 1
+            try:
+                self.read_buffer = pair[1]
+            except IndexError:
+                self.read_buffer = b""
+
+    def read(self):
+        """Read the prepared read bufferr and return it as a string."""
+        if self.read_buffer:
+            return self.read_buffer.decode("utf-8")
+        else:
+            pair = self.comm_pairs[self.index]
+            assert pair[0] is None, "Unexpected read without prior write."
+            self.index += 1
+            return to_bytes(pair[1]).decode("utf-8")
+
+    def read_bytes(self, count):
+        """Read `count` number of bytes."""
+        if self.read_buffer:
+            read = self.read_buffer[:count]
+            self.read_buffer = self.read_buffer[count:]
+            return read
+        else:
+            pair = self.comm_pairs[self.index]
+            assert pair[0] is None, "Unexpected read without prior write."
+            self.index += 1
+            read = pair[1][:count]
+            self.read_buffer = pair[1][count:]
+            return read
 
     # TODO: Harmonise ask being write+read (i.e., remove from VISAAdapter),
     #   use protocol tests to confirm it works correctly
--- a/pymeasure/adapters/protocol.py
+++ b/pymeasure/adapters/protocol.py
@@ -43,75 +43,80 @@ def to_bytes(command):
             return bytes(command)
         except TypeError:
             raise
-    elif isinstance(command, int):
-        return bytes((command,))
-    raise TypeError("Invalid input")
+    elif isinstance(command, (int, float)):
+        return str(command).encode("utf-8")
+    raise TypeError(f"Invalid input of type {type(command).__name__}.")
 
 
 class ProtocolAdapter(Adapter):
     """ Adapter class testing command exchange without instrument hardware.
 
+    :param list of lists comm_pairs: List of message pairs. First message is
+        the written one, the second one is the read message, it is optional.
+        'None' indicates that this message (write or read) does not exist.
     :param kwargs: TBD key-word arguments
     """
 
-    def __init__(self, comm_pairs, preprocess_reply=None, **kwargs):
+    def __init__(self, comm_pairs=[], preprocess_reply=None, **kwargs):
+        """Generate the adapter and initialize internal buffers."""
+        assert isinstance(comm_pairs, (list, tuple)), (
+            "Parameter comm_pairs has to be a list or tuple.")
         super().__init__(preprocess_reply=preprocess_reply)
-        self.read_buffer = b""
-        self.write_buffer = b""
+        self._read_buffer = b""
+        self._write_buffer = b""
         self.comm_pairs = comm_pairs
-        self.index = 0
+        self._index = 0
         # TODO: Make this skeleton implementation workable
 
     def write(self, command):
         """Compare the command with the expected one and fill the buffer."""
-        pair = self.comm_pairs[self.index]
-        self.index += 1
-        assert to_bytes(pair[0]) == to_bytes(command), (
+        pair = self.comm_pairs[self._index]
+        self._index += 1
+        assert to_bytes(pair[0]) == self._write_buffer + to_bytes(command), (
                 f"Command '{command}' written, but {pair[0]} expected.")
-        print(pair)
+        self._write_buffer = b""
         try:
-            self.read_buffer = to_bytes(pair[1])
+            self._read_buffer = to_bytes(pair[1])
         except IndexError:
             # No response in the pair.
-            self.read_buffer = b""
+            self._read_buffer = b""
 
     def write_bytes(self, content):
         """Write the bytes `content`. If a command is full, fill the read."""
-        self.write_buffer += content
-        pair = self.comm_pairs[self.index]
-        if self.write_buffer == pair[0]:
-            assert self.read_buffer == b"", "Responses have not been read yet."
+        self._write_buffer += content
+        pair = self.comm_pairs[self._index]
+        if self._write_buffer == to_bytes(pair[0]):
+            assert self._read_buffer == b"", "Responses have not been read yet."
             # Clear the write buffer
-            self.write_buffer = b""
-            self.index += 1
+            self._write_buffer = b""
+            self._index += 1
             try:
-                self.read_buffer = pair[1]
+                self._read_buffer = to_bytes(pair[1])
             except IndexError:
-                self.read_buffer = b""
+                self._read_buffer = b""
 
     def read(self):
         """Read the prepared read bufferr and return it as a string."""
-        if self.read_buffer:
-            return self.read_buffer.decode("utf-8")
+        if self._read_buffer:
+            return self._read_buffer.decode("utf-8")
         else:
-            pair = self.comm_pairs[self.index]
+            pair = self.comm_pairs[self._index]
             assert pair[0] is None, "Unexpected read without prior write."
-            self.index += 1
+            self._index += 1
             return to_bytes(pair[1]).decode("utf-8")
 
     def read_bytes(self, count):
         """Read `count` number of bytes."""
-        if self.read_buffer:
-            read = self.read_buffer[:count]
-            self.read_buffer = self.read_buffer[count:]
+        if self._read_buffer:
+            read = self._read_buffer[:count]
+            self._read_buffer = self._read_buffer[count:]
             return read
         else:
-            pair = self.comm_pairs[self.index]
+            pair = self.comm_pairs[self._index]
             assert pair[0] is None, "Unexpected read without prior write."
-            self.index += 1
-            read = pair[1][:count]
-            self.read_buffer = pair[1][count:]
-            return read
+            self._index += 1
+            self._read_buffer = to_bytes(pair[1])[count:]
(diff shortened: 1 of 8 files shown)


pymeasure/test.py
```diff
@@ -48,9 +48,13 @@ def expected_protocol(instrument_cls, comm_pairs):
...
     """
-    protocol = ProtocolAdapter(comm_pairs)
-    instr = instrument_cls(protocol, "Virtual instrument")
+    instr = instrument_cls("test", name="Virtual instrument",
+                           comm_pairs=comm_pairs)
     yield instr
+    protocol = instr.adapter
+    assert protocol._index == len(comm_pairs), "Not all messages exchanged."
+    assert protocol._write_buffer == b"", "Non empty write buffer."
+    assert protocol._read_buffer == b"", "Non empty read buffer."
```
bilderbuchi, line 57: Generally, in exception messages, be helpful and try to give the user information they will need to deal with the exception (often: the content of the variables triggering the exception). It can be quite annoying to get an exception that leaves you only with the knowledge that something weird is going on, and then you have to set a breakpoint there, re-do what triggered the problem, and _then_ inspect the frame contents to find out what's going on. Sometimes, getting the a couple of critical bits of information is enough to unstick you.

If I may suggest something here:
```suggestion
    assert protocol._index == len(comm_pairs), f"Unprocessed protocol definitions remain: {comm_pairs[protocol._index:]}."
    assert protocol._write_buffer == b"", f'Non-empty write buffer: "{protocol._write_buffer}".'
    assert protocol._read_buffer == b"", f'Non-empty read buffer: "{protocol._read_buffer}".'
```
Now the user will immediately see what has been left behind, and maybe already realizes their mistake.

pymeasure/adapters/protocol.py
```diff
@@ -30,16 +30,96 @@
...
+        self._write_buffer = b""
+        try:
+            self._read_buffer = to_bytes(pair[1])
+        except IndexError:
+            # No response in the pair.
+            self._read_buffer = b""
+
+    def write_bytes(self, content):
+        """Write the bytes `content`. If a command is full, fill the read."""
+        self._write_buffer += content
```
bilderbuchi, line 86: Is there a valid situation when we _append_ to the write buffer? Shouldn't whatever is in there already been consumed in a previous operation?

tests/adapters/test_protocol.py
```diff
@@ -22,10 +22,177 @@
...
+    a = ProtocolAdapter([("write", "read"), ("write_only",)])
+    assert a.comm_pairs == [("write", "read"), ("write_only",)]
     # TODO: Make this first test assert something useful
+
+
+class Test_write:
+    @mark.parametrize("command, written", (("hey Ho", b"hey Ho"),
+                                           (b"superX", "superX")
+                                           ))
+    def test_write(self, command, written):
```
bilderbuchi, line 54: `test_write_does_not_distinguish_bytestring_from_string` is maybe a better name?
Maybe a discussion point for later if we should actually allow that...

tests/adapters/test_protocol.py
```diff
@@ -22,10 +22,177 @@
...
+        a = ProtocolAdapter([(written, 5)])
+        a.write(command)
+        assert a._read_buffer == b"5"
+
+    def test_write_without_response(self):
+        a = ProtocolAdapter([("Hey ho",)])
```
bilderbuchi, line 60: As I said above, I think it will make everything more consistent if a one-tuple becomes disallowed, and we used `("Hey ho", None)` here.

tests/adapters/test_protocol.py
```diff
@@ -22,10 +22,177 @@
...
+        a = ProtocolAdapter([("written", 5)])
+        a.write_bytes(b"written")
+        return a
+
+    @fixture(scope="class")
+    def written2(self):
+        """Write in two turns."""
+        a = ProtocolAdapter([("written", 5)])
+        a.write_bytes(b"writ")
+        a.write_bytes(b"ten")
```
bilderbuchi, line 89: Is that a use case you see already in your instruments? Writing two bits of one "command" separately, that should actually be one? Would you not test the protocol for that with `[("writ", None), ("ten", 5)]`?

tests/adapters/test_protocol.py
```diff
@@ -22,10 +22,177 @@
...
+        return a
+
+    def test_write_write_buffer1(self, written1):
+        assert written1._write_buffer == b""
+
+    def test_write_write_buffer2(self, written2):
+        assert written2._write_buffer == b""
```
bilderbuchi, line 96: To avoid the duplicated tests, you could parametrize the fixture itself: https://docs.pytest.org/en/6.2.x/fixture.html#parametrizing-fixtures

tests/adapters/test_protocol.py
```diff
@@ -22,10 +22,177 @@
...
+class Test_read:
+    @mark.parametrize("buffer, returned", ((b"\x03\x65", "\x03e"),
+                                           (b"Bytes", "Bytes"),
+                                           ))
+    def test_works(self, buffer, returned):
+        a = ProtocolAdapter()
+        a._read_buffer = buffer
+        assert a.read() == returned
+
+    def test_read_buffer_emptied(self):
```
bilderbuchi, line 136:
```suggestion
    def test_read_empties_read_buffer(self):
```

tests/adapters/test_protocol.py
```diff
@@ -22,10 +22,177 @@
...
+
+
+def test_read_write_sequence():
+    """Test several consecutive writes and reads, including ask."""
+    a = ProtocolAdapter([("c1", "a1"), ("c2",), (None, "a3"), ("c4", "a4")])
+    a.write("c1")
+    assert a.read() == "a1"
+    a.write("c2")
+    assert a.read() == "a3"
+    assert a.ask("c4") == "a4"
```
bilderbuchi, line 189: Does that make sense to you (simulating additional unprompted messages after an ask())?
```suggestion
    a = ProtocolAdapter([("c1", "a1"), ("c2",), (None, "a3"), ("c4", "a4"), (None, "a5")])
    a.write("c1")
    assert a.read() == "a1"
    a.write("c2")
    assert a.read() == "a3"
    assert a.ask("c4") == "a4"
    assert a.read() == "a5"
```

tests/adapters/test_protocol.py
```diff
@@ -22,10 +22,177 @@
...
+
+
+def test_write_and_read_with_and_without_bytes():
+    """Test writing and reading, normal and bytes in combination."""
+    a = ProtocolAdapter([("c1", "a1"), ("c4", "a4")])
+    a.write_bytes(b"c")
+    a.write("1")
```
bilderbuchi, line 196: The second pair should be deleted, not part of this test.

I'm not sure if we should allow this write sequence. I think I can see a use case for piecewise binary writes (and probably reads). However, mixing binary and normal writes, and especially `write("1")` that _succeeds_ against an expected protocol of "c1" looks veryyy weird to me.

tests/instruments/hcp/test_tc038d.py
```diff
@@ -0,0 +1,122 @@
...
+    ) as inst:
+        with pytest.raises(ConnectionError):
+            inst.setpoint = 32.1
+
+
+def test_write_multiple_wrong_values():
+    with expected_protocol(
+        TC038D, []
+    ) as inst:
+        with pytest.raises(ValueError):
```
bilderbuchi, line 41: maybe a comment might be useful here what kind of ValueError is raised here? Granted, it's a binary protocol, but what exactly is wrong here?

tests/instruments/hcp/test_tc038d.py
```diff
@@ -0,0 +1,122 @@
...
+
+def test_write_multiple_Value_error():
+    with expected_protocol(
+        TC038D,
+        [(b"\x01\x10\x01\x06\x00\x02\x04\x00\x00\x01A\xbf\xb5",
+          b"\x01\x90\x02\x06\x00")],
+    ) as inst:
+        with pytest.raises(ValueError) as exc:
+            inst.setpoint = 32.1
+            assert str(exc) == "Wrong start address"
```
bilderbuchi, line 53: You can use the `match=` kwarg of `raises` to test for the correct message: https://docs.pytest.org/en/6.2.x/reference.html#pytest.raises 
I love pytest :-)

tests/instruments/hcp/test_tc038d.py
```diff
@@ -0,0 +1,122 @@
...
+
+def test_write_multiple_wrong_values():
+    with expected_protocol(
+        TC038D, []
+    ) as inst:
+        with pytest.raises(ValueError):
+            inst.writeMultiple(0x010A, 5.5)
+
+
+def test_write_multiple_Value_error():
```
bilderbuchi, line 45:
```suggestion
def test_write_multiple_handle_wrong_start_address():
```
?

bilderbuchi: Also, maybe those tests checking internal error responses (CRC, start address, etc below) can use a docstring explaining what exactly is being tested for, how and why?

tests/test_expected_protocol.py
```diff
@@ -43,6 +50,41 @@ def test_simple_protocol():
...
 
+def test_not_all_communication_used():
+    """Test whether unused communication raises an error."""
+    with raises(AssertionError) as exc:
+        with expected_protocol(BasicTestInstrument,
+                               [('VOLT?', 3.14),
+                                ('VOLT 4.5 V',),
+                                ]) as instr:
+            assert instr.simple == 3.14
+    assert str(exc.value) == "Not all messages exchanged."
```
bilderbuchi, line 61: as above, `with pytest.raises(AssertionError, match=...` should be able to do this in one go.

tests/test_expected_protocol.py
```diff
@@ -43,6 +50,41 @@ def test_simple_protocol():
...
+def test_non_empty_read_buffer():
+    with raises(AssertionError) as exc:
+        with expected_protocol(BasicTestInstrument,
+                               [('VOLT?', 3.14),
+                                ]) as instr:
+            instr.write("VOLT?")
+    assert str(exc.value) == "Non empty read buffer."
+
+
+def test_preprocess():
```
bilderbuchi, line 83: Good that we have that one, the external behaviour/usage should be like this, however we deal with this internally (if we find a better way)

bilderbuchi (changes requested):
Thank you for this, you obviously put a lot of thought into it already! :heart:
I got more remarks/questions than I thought, as the design was also not finished in my head and the review got me thinking.

Great tests that you included. Nonetheless I added some suggestions, as those will be trailblazing tests and I figure others will refer to what's already there as a template, so it pays to polish that!

I like how you put most of the protocol logic tests with the ProtocolAdapter, which keeps the number of tests for expected_protocol quite small!

pymeasure/adapters/protocol.py
```diff
@@ -30,16 +30,96 @@
...
+        return str(command).encode("utf-8")
+    raise TypeError(f"Invalid input of type {type(command).__name__}.")
+
+
 class ProtocolAdapter(Adapter):
     """ Adapter class testing command exchange without instrument hardware.
 
+    :param list of lists comm_pairs: List of message pairs. First message is
+        the written one, the second one is the read message, it is optional.
+        'None' indicates that this message (write or read) does not exist.
```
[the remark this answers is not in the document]
bmoneke: Yes, it will make it easier, especially if you forget the comma, such that you do not have a tuple despite the brackets.

pymeasure/adapters/protocol.py
```diff
@@ -30,16 +30,96 @@
...
+        super().__init__(preprocess_reply=preprocess_reply, **kwargs)
+        self._read_buffer = b""
+        self._write_buffer = b""
+        self.comm_pairs = comm_pairs
+        self._index = 0
         # TODO: Make this skeleton implementation workable
 
+    def write(self, command):
+        """Compare the command with the expected one and fill the buffer."""
+        pair = self.comm_pairs[self._index]
```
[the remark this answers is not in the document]
bmoneke: Yes, with compulsory tuples of two, it is possible, but in the former implementation of tuples with one or two members, it was not possible.

pymeasure/adapters/protocol.py
```diff
@@ -30,16 +30,96 @@
...
+        except IndexError:
+            # No response in the pair.
+            self._read_buffer = b""
+
+    def write_bytes(self, content):
+        """Write the bytes `content`. If a command is full, fill the read."""
+        self._write_buffer += content
+        pair = self.comm_pairs[self._index]
+        if self._write_buffer == to_bytes(pair[0]):
+            assert self._read_buffer == b"", "Responses have not been read yet."
```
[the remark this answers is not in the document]
bmoneke: If you write only (for example a property setter), but the device responds with some confirmation, the read_buffer remains filled with that confirmation. If you now try to write again (setting something or as part of `ask`), the exception will be raised, because the next reading will (with the real instrument) return that confirmation and not, what you expected.

In order to fix it, you have to read the buffer, either with "read" itself, or with the property's check_set_errors which does read.

pymeasure/adapters/protocol.py
```diff
@@ -30,16 +30,96 @@
...
+            self._read_buffer = to_bytes(pair[1])
+        except IndexError:
+            # No response in the pair.
+            self._read_buffer = b""
+
+    def write_bytes(self, content):
+        """Write the bytes `content`. If a command is full, fill the read."""
+        self._write_buffer += content
+        pair = self.comm_pairs[self._index]
+        if self._write_buffer == to_bytes(pair[0]):
```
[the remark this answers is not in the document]
bmoneke: No, because you can compose a message with several writes, see communication description below.

pymeasure/adapters/visa.py
```diff
@@ -123,6 +123,10 @@ def read(self):
         """
         return self.connection.read()
 
+    def write_bytes(self, data):
```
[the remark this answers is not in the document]
bmoneke: In fact, I already mentioned it in my last post in #567

pymeasure/instruments/instrument.py
```diff
@@ -146,12 +147,14 @@ class Instrument:
 
     # noinspection PyPep8Naming
     def __init__(self, adapter, name, includeSCPI=True, **kwargs):
-        try:
-            if isinstance(adapter, (int, str)):
+        if adapter == "test":
+            adapter = ProtocolAdapter(**kwargs)
```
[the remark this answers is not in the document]
bmoneke: Done. I proposed to move self.preprocess to the protocol (or instrument), or even move the whole values there.

pymeasure/test.py
bmoneke: Under normal circumstances, I agree. In this case, we're talking about a method used by pytest, which already shows the difference of the failing asserted comparison. For that reason I did not repeat it. On the other hand, it does not hurt.

tests/adapters/test_protocol.py
bmoneke: I see reading a message in different "batches", but never writing, you're right. Out of symmetry I made this option.
I cannot think of any scenario, where you cannot compose fully a message before sending it.

tests/adapters/test_protocol.py
bmoneke: I tried parametrizing the tests, which does not work with fixtures. Thanks for showing me the parametrization of fixtures.

tests/adapters/test_protocol.py
bmoneke: It might be, that a device sends a command by itself, therefore, I wanted to test, whether the ProtocolAdapter handles a combination of several message pairs, including that possibility.

tests/adapters/test_protocol.py
bmoneke: See below about my communication thoughts.

bmoneke:
Thank you for you comments.

Regarding my other driver (Ophir power meters, once its conversion is done, I'll make a PR), it works. Just the tests do not work with the preprocess given to the adapter. I just ran the tests with preprocess defined in my Instrument's values (according to my suggestion how to handle it in the future): everything works. So, don't worry.

Regarding the design of write and write_bytes:
As far as I know, received bytes are stored in a read buffer until a message is detected, which is then processed. A message is either complete when a termchar is sent (done via write), or after a certain amount of bytes in frame-based communication. For a frame-based communication, I do have to check, whether the accumulated sent bytes corresponds to the expected message, as the ProtocolAdapter does not know how many bytes to expect and according to which rules.

In order to simulate, that sending bytes individually, and maybe finishing it off with a normal write, may constitute a real message, I designed that write_buffer to hold all bytes not yet read by the device.

This leads to your unexpected results, that `write_bytes(b"c"); write("1")` succeeds in querying ("c1"), but it is right, as the previously sent b"c" has to be taken in account, as well. For example `write_bytes(b"x"); write("c1")` would cause an error if the device expects "c1" only and not "xc1".

I agree, that you always should be able to compose a message beforehand and write it in a single write/write_bytes. Nevertheless, you can send it with several write_bytes (for example the individual parts of frame-based communication). If it works with the real device, it should work with the ProtocolAdapter, shouldn't it? The expected message is the same, so if you switch to a single write, the test remains the same.

tests/instruments/hcp/test_tc038d.py
bmoneke: It should be a TypeError

tests/test_expected_protocol.py
bmoneke: Thanks for that hint.

tests/test_expected_protocol.py
bmoneke: Defining the preprocess on Instrument level, fixes all problems and is a tiny change: adding it to the Instruments values method instead of the Adapter instantiation.

Requiring to define preprocess in the Instrument for tests, does not break anything (there are no tests yet). And if you write tests, you can make that change in the process.

tests/adapters/test_protocol.py
bmoneke: I just wanted to have some different inputs. In the end, to_bytes is already tested, so we do not need to test different things here.

Basically, this discussion is about whether to_bytes is applied to the comm_pairs or not. In the latter case, the protocol has to be specified in bytes.

pymeasure/adapters/protocol.py
```diff
@@ -30,16 +30,96 @@
...
+        self.comm_pairs = comm_pairs
+        self._index = 0
         # TODO: Make this skeleton implementation workable
 
+    def write(self, command):
+        """Compare the command with the expected one and fill the buffer."""
+        pair = self.comm_pairs[self._index]
+        self._index += 1
+        assert to_bytes(pair[0]) == self._write_buffer + to_bytes(command), (
+                f"Command '{command}' written, but {pair[0]} expected.")
```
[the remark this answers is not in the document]
bmoneke: See my comment about Protocol thoughts in the conversation part.

pymeasure/test.py
bilderbuchi: Ah, of course, pytest takes care of that :sweat_smile:

tests/test_expected_protocol.py
```diff
@@ -33,16 +35,56 @@ class BasicTestInstrument(Instrument):
     )
 
 
+class InstrumentWithPreprocess(BasicTestInstrument):
+    def values(self, command, **kwargs):
+        return super().values(command, preprocess_reply=lambda v: v+"2345", **kwargs)
+
+    def __init__2(self, adapter, **kwargs):
```
bilderbuchi, line 42: Oops, that 2 should not be there I think.

pymeasure/adapters/protocol.py
```diff
@@ -30,15 +30,76 @@
...
+                f"Unread response '{self._read_buffer}' present when writing. "
+                "Read the response or enable the property's 'check_set_errors'")
+            # Clear the write buffer
+            self._write_buffer = b""
+            self._read_buffer = to_bytes(p_read)
+            self._index += 1
+
+    def read(self):
+        """Return an already present or freshly fetched read buffer as a string."""
+        return (self.read_bytes(-1) + self.read_bytes(1)).decode("utf-8")
```
bilderbuchi, line 89: I'll have to take some time to step through _why_ this works  as expected :D

pymeasure/adapters/protocol.py
bilderbuchi: Understood, thanks!

pymeasure/adapters/protocol.py
bmoneke: We're reading lists in this Adapter. - 1 returns all bytes but the last one, 1 returns the last byte.

A first implementation reading len(buffer) bytes failed, if the pair was (None, resp), because the length is zero in this case.

tests/test_expected_protocol.py
bmoneke: No, the 2 deactivates the init, depending how we further decide to impement stuff.
If we decide to not support (in tests) an adapter preprocessor, we can drop this init altogether.

bilderbuchi:
Guten Morgen!

> Regarding my other driver (Ophir power meters, once its conversion is done, I'll make a PR), it works. Just the tests did not work with the preprocess given to the adapter. I just ran the tests with preprocess defined in my Instrument's values (according to my suggestion how to handle it in the future): everything works. So, don't worry.

Great!

> Regarding the design of write and write_bytes:
As far as I know, received bytes are stored in a read buffer until a message is detected, which is then processed. A message is either complete when a termchar is sent (done via write), or after a certain amount of bytes in frame-based communication. For a frame-based communication, I do have to check, whether the accumulated sent bytes corresponds to the expected message, as the ProtocolAdapter does not know how many bytes to expect and according to which rules.

> In order to simulate, that sending bytes individually, and maybe finishing it off with a normal write, may constitute a real message, I designed that write_buffer to hold all bytes not yet read by the device.

> This leads to your unexpected results, that `write_bytes(b"c"); write("1")` succeeds in querying ("c1"), but it is right, as the previously sent b"c" has to be taken in account, as well. For example `write_bytes(b"x"); write("c1")` would cause an error if the device expects "c1" only and not "xc1".

OK that makes sense. I think/hope that mixing write and write_bytes will remain a synthetic test situation, if even that.

> I agree, that you always should be able to compose a message beforehand and write it in a single write/write_bytes. Nevertheless, you can send it with several write_bytes (for example the individual parts of frame-based communication). If it works with the real device, it should work with the ProtocolAdapter, shouldn't it? The expected message is the same, so if you switch to a single write, the test remains the same.

"If it works with the real device, it should work with the ProtocolAdapter, shouldn't it?" this is a good point, agreed!

> EDIT: Point of discussion: How have the comm_pairs to be defined? As bytes-object or also as strings, even as numbers? As an array of ints (basically bytes array)?

I would say: either, as appropriate. For frame-based devices, we need to be able to prescribe bytes, but for message-based/plaintext protocols bytes will be very inconvenient and we should use strings. 
Single numbers, I'm not sure and don't know the use case.
Int arrays, as you say that is basically byte arrays, that could be useful if one checks those byte-based data transfers (oscilloscope traces etc)

> I'd like to point out, how much I love tests: changing read/write to use the bytes version was super easy, as the tests confirmed equal performance with different code.

I know, right?! This has been on my wish list for ages, especially because 
* it's the "right thing" to do to have tests in a code base
* I have no hardware available at all anymore (since changing jobs) and don't use pymeasure myself anymore day-to-day. :-/
* Testing hardware is hard/complicated
I'm very happy that this is coming in now, we will be able to increase pymeasure's quality with this, and also will be able to do some refactoring tasks much more easily when we have a couple of strategic tests in place!

pymeasure/adapters/protocol.py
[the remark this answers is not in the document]
bilderbuchi: Thanks, I see.

pymeasure/adapters/protocol.py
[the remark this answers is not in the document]
bilderbuchi: OK, I'll add a comment about that.

tests/adapters/test_protocol.py
bilderbuchi: I think it's good to have the freedom to use either. I have not yet come up with a case where that's a bad idea.

tests/adapters/test_protocol.py
bilderbuchi: Ok, I'll add a remark about his.

tests/adapters/test_protocol.py
bilderbuchi: What I meant is that I don't think we have tested all combinations, here. I'll add a separate test for the check_set_error situation.

tests/test_expected_protocol.py
bilderbuchi: That's a good workaround until we resolve this properly.

tests/test_expected_protocol.py
bilderbuchi: I'll fix this.

bilderbuchi:
This is ready IMO, so I'll merge it. 
Then, I'll tweak a couple of things based on my notes from this PR, and push an update later.
After that, I'll check the TODOs and see what I can attack next.

PR #634, opened 2022-06-02
Add feature: Protocol tests without hardware present

bilderbuchi:
This will be useful to test pymeasure instrument implementations without having the hardware present (see #627).

@bmoneke I have now gone through my notes for now and opened this PR so we can iterate further and knock out the remaining `TODO` items together. If you're up to it we can to PRs against this branch again, I think this worked well for collaborative design/implementation.

* [x] ~~What I still have noted down to do is "Debug/investigate and comment on what ProtocolAdapter.read() is doing with those indices" because it was not clear to me at review time :sweat_smile:, but I didn't put that as a formal TODO.~~ let's skip this to not get bogged down even longer
* [x] Also, when the implementation is converged, we need to write useful user-facing documentation, with examples, that we can point people at when we ask them to write protocol tests. Adding those tests being a painless process, with little boilerplate/distractions, is I think important to get people to add instrument tests.
* [x] Add protocol tests for a (real) simple instrument
* [x] Add protocol tests for a (real) instrument with multiple channels

Closes #627 
Closes #249

PR #637, opened 2022-06-13
New Instrument: MKS937B vacuum gauge controller

dkriegner:
This implements a device of a new vendor: [MKS Instruments 937B vacuum gauge controller](https://www.mks.com/f/937b-combination-vacuum-gauge-controller)

Connection to the device is made through an RS232/RS485 serial connection. 
The communication protocol of this device is as follows:

Query: `@<aaa><Command>?;FF` with the response `@<aaa>ACK<Response>;FF`
Set command: `@<aaa><Command>!<parameter>;FF` with the response `@<aaa>ACK<Response>;FF`

`<aaa>` is an address from `001` to `254`

I did not see a way to implement this without overloading VISAAdapter since the terminator `';FF'` is invalid in pyvisa as `read_terminator` (`ValueError: ambiguous ending in termination characters`). But if you see a more elegant solution, please let me know.

Using the code in this pull request, the device can be initialized via:
```
import pymeasure.instruments
gauge = pymeasure.instruments.mksinst.MKS937B('ASRL1::INSTR')  # or any other serial device
```

Edit:
I should also point out that I propose some changes to `Instrument` to allow map values to be used also in case multiple values are returned by an instrument. I guess this needs particular attention during the code review. Maybe even some test case has to be added to test this functionality? (How would this be done?)


bmoneke:
We're implementing a testing suite, with which you can test the communication protocol of your device. PR #634 contains the current work in progress.
The tests do work, we're writing the documentation (first version is already done, so look at [adding_instruments](docs/dev/adding_instruments.rst) in the protocol-test branch.

Instead of writing your own adapter, you could add reading those two bytes to the instruments read/values method.
I think that would be more in line with the current plan to have all instrument related stuff in the instrument and not the adapter itself. I'm not a maintainer, however, so wait for some official answer.

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -0,0 +1,288 @@
...
+        for prop in ["serial", "pressure1", "pressure2", "pressure3",
+                     "pressure4", "pressure5", "pressure6", "all_pressures",
+                     "combined_pressure1", "combined_pressure2",
+                     "sensor_typeA", "sensor_typeB", "sensor_typeC",
+                     "ion_gauge_status1", "ion_gauge_status3",
+                     "ion_gauge_status5", "unit",
+                     "power_status1", "power_status2", "power_status3",
+                     "power_status4", "power_status5", "power_status6",
+                     ]:
+            setattr(self, f"{prop}_command_process", self.command_process)
```
bmoneke, line 90: You want to include the device address in the command sent to the device, don't you?
Alternatively, you can rewrite the instrument write/values, to do that for you. The values/read methods could check for those two additional "F" chars, as well.

That way you do not make all the properties dynamic.
```suggestion

    def values(self, command, **kwargs):
        ret = super().values(self.command_process(command), **kwargs)
        self._check_termination()
        return ret

    def _check_termination(self):
        """Check the read termination to correspond to the protocol."""
        t = self.adapter.read_bytes(2)  # read extra termination chars 'FF'
        if t != b'FF':
            raise ValueError(f"unexpected termination string received {t}")
```

dkriegner:
> We're implementing a testing suite, with which you can test the communication protocol of your device. PR #634 contains the current work in progress. The tests do work, we're writing the documentation (first version is already done, so look at [adding_instruments](docs/dev/adding_instruments.rst) in the protocol-test branch.

I have not yet had the time to look into that.
 
> Instead of writing your own adapter, you could add reading those two bytes to the instruments read/values method. I think that would be more in line with the current plan to have all instrument related stuff in the instrument and not the adapter itself. I'm not a maintainer, however, so wait for some official answer.

I followed your advise and revised the code to avoid the definition of an extra adapter. I am not sure if I should not also overload the Instrument.ask method. Its not needed for the present code, but for the user it could be best to provide a consistent interface where ask/write/values performs the adding of the address part in the command.

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -0,0 +1,288 @@
...
+        :returns: string with only the response, or the original string
+        """
+        rvalue = self._re_response_value.search(reply)
+        if rvalue:
+            return rvalue.groups()[0]
+        else:
+            full_reply = self._re_fullresponse.search(reply)
+            if full_reply:
+                return full_reply.groups()[0]
+        return reply
```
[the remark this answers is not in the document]
dkriegner: This method depends on self (see self._re_response_value, self.re_fullresponse) which need to be instance variables since they depend on the device address. They could be made generic regular expressions, but I find it best to check it with the actual device address and not just with any 3 digit number. If this device would be used on a real bus connection (RS485) it is likely best in this way.

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -0,0 +1,288 @@
...
+            if reply.groups()[0].startswith('ACK'):
+                return
+        # no valid acknowledgement message found
+        raise ValueError(f"invalid reply '{ret}' found in check_errors")
+
+    def command_process(self, cmd):
+        """
+        create command string by adding device address
+        """
+        return f"@{self.address:03d}{cmd}"
```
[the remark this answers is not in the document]
dkriegner: `check_errors` has to be exposed since this is how it used inside the `Instrument.control` properties code. The other one I mark private.

pymeasure/instruments/mksinst/mks937b.py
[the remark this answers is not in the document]
bmoneke: Sorry, I missed that self, so you're fine.

bmoneke:
We plan to make Instrument.ask call write+read of the instrument instead of calling Adapter.ask, because we want to have the whole "protocol logic" in these two methods: write and read.
If you're concerned somebody might use ask, implement it as self.write+self.read.

dkriegner:
> We plan to make Instrument.ask call write+read of the instrument instead of calling Adapter.ask, because we want to have the whole "protocol logic" in these two methods: write and read. If you're concerned somebody might use ask, implement it as self.write+self.read.

This sounds like a sensible change. I already had to generate an Adapter because the Instrument read/write are effectively bypassed. I think it would make sense to do this on the Instrument level and not in every device. I think this can wait for a central solution of this issue.

> 
> It would be great, if you added tests for the expanded Instrument.control to the corresponding [test file](https://github.com/pymeasure/pymeasure/blob/master/tests/instruments/test_instrument.py).

will do. please wait with the merge for my tests

dkriegner:
I have added a test to the mentioned file. 

Since the feature of mapping was only implemented in `Instrument.measurement` I limited the test to the a `measurement` property. In order to extend the mapping of multiple values also to the `setting` or `control` properties I think more extensive changes are needed which need to be discussed.

- validators would need to handle iterable objects
- either the default separator (which now seems only available in `Adapter.values`) would need to be accessible to `Instrument.control.fset` in order that the setting string can be generated. 
Or 
- the set string would need to have multiple variable placeholders (which for really long arrays would be likely impractical)

It seems to me this would be beyond this scope of this PR. From my point of view this Instrument implementation would be ready to be merged.

bilderbuchi:
Looks interesting, but I'll need more time until I get around for a review, summer time is crazy :D (also, I'd like to get #634 in, first)

bilderbuchi:
> Looks interesting, but I'll need more time until I get around for a review, summer time is crazy :D (also, I'd like to get #634 in, first)

just FYI, #634 has been merged now, so [protocol tests are possible now](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#protocol-tests).

bilderbuchi:
> I did not see a way to implement this without overloading VISAAdapter since the terminator `';FF'` is invalid in pyvisa as `read_terminator` (`ValueError: ambiguous ending in termination characters`). But if you see a more elegant solution, please let me know.

yeah, that one will be challenging to get around.

bilderbuchi (changes requested):
First round of review, without the changes to map multiple values -- I might have a better approach, but I will need some more thought.

pymeasure/instruments/instrument.py
```diff
@@ -377,7 +377,24 @@ def fget(self,
...
+                            if v == val:
+                                ret.append(k)
+                        if len(ret) != i + 1:
+                            raise KeyError(f"Value {val} not found in mapped values")
+                    return ret
+                else:
+                    raise ValueError(
+                        'Values of type `{}` are not allowed '
+                        'for Instrument.control'.format(type(values))
+                    )
```
bilderbuchi, line 397: What I don't like is that this duplicates a lot of code and logic. I have come up with the following better alternative, which passes your test but does not duplicate the code. 
```
        def fget(self,
                 get_command=get_command,
                 values=values,
                 map_values=map_values,
                 get_process=get_process,
                 command_process=command_process,
                 check_get_errors=check_get_errors,
                 ):
            if get_command is None:
                raise LookupError("Instrument property can not be read.")
            vals = self.values(command_process(get_command), **kwargs)
            if check_get_errors:
                self.check_errors()
            vals = [get_process(v) for v in vals]  # TODO: should we run get_process on all elements or the whole list?
            if map_values:
                ret = []
                for value in vals:
                    if isinstance(values, (list, tuple, range)):
                        ret.append(values[int(value)])
                    elif isinstance(values, dict):
                        for k, v in values.items():
                            if v == value:
                                ret.append(k)
                                break
                        else:  # for-loop ran completely
                            raise KeyError(f"Value {value} not found in mapped values")
                    else:
                        raise ValueError(
                            'Values of type `{}` are not allowed '
                            'for Instrument.control'.format(type(values))
                        )
            else:
                ret = vals
            if len(ret) == 1:
                return ret[0]
            else:
                return ret
```
(basically, I am delaying the distinction between 1 and multiple value elements to later and do mapping on all elements)

However, I am not yet 100% confident that there are not further implications that I did not consider. 
* Also, I have not yet thought through at which point get_process should apply. 
* I'm not sure about mapping now becoming asymmetric, as mapping lists of items now only applies on get. Does it even make sense on set?

It would be good to have the extended amount of tests from #665 (will merge soon) before merging this.

ping @bmoneke what do you think about the impact of this?

bilderbuchi (changes requested):
Some thoughts on the values mapping change.

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def _check_extra_termination(self):
+        """
+        Check the read termination to correspond to the protocol
+        """
+        t = self.adapter.read_bytes(2)  # read extra termination chars 'FF'
```
bilderbuchi, line 107: @bmoneke that approach might be more challenging to test with our current testing setup (either ProtocolAdapter or the Mock when checking all instruments). I guess we will see on the next CI run :D

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def _command_process(self, cmd):
+        """
+        create command string by adding device address
+        """
+        return f"@{self.address:03d}{cmd}"
+
+    def _check_extra_termination(self):
+        """
+        Check the read termination to correspond to the protocol
```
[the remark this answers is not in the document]
bilderbuchi: We could probably to an instrument-specific test that this all works correctly using pyvisa-sim like [here](https://github.com/pyvisa/pyvisa/issues/120#issuecomment-74980018), where we can create simple returns, and include all termination characters. not sure if the eom entry can accept `;FF`, though.

pymeasure/instruments/mksinst/mks937b.py
bmoneke: You have to make the communication pairs longer, but it should work.
E.g.
comm_pairs = [(b"writing something", b"the response"), (None, b"FF")]

Remember, that you must not use the termination_characters defined at adapter level (i.e. ";FF" for writing and ";" for reading).

pymeasure/instruments/instrument.py
bmoneke: I'd say make `get_process(vals)`, because it gives flexibility without loosing possibility:

- If you want to modify every entry of a list, get_process can take that into account. e.g. `lambda vals: [val**2 for val in vals]`.
- If you want to take the list as a whole, e.g. get the average or standard deviation or sum or min, max or a specific value (e.g. the first one) etc., you can do it as well.

bmoneke: I just noticed problem with my suggestion:
If you had `get_process = lambda v: 2*v` before, you need to work with the list now.

I do not know by heart how many instruments implement get_process.

On the other hand, applying the get process to every single value should not brake any present code.

bilderbuchi: Good point on the flexibility!

Backwards compatibility is a good point (get_process is used a lot)! 
Until now, get_process is acting on the string entry if vals was length 1, and on the list if length >1 (which may be less than ideal in itself). We could preserve that behaviour with some small code tweaks.

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+
+    def __init__(self, resource_name, address=253, **kwargs):
+        kwargs.setdefault("preprocess_reply", self._extract_reply)
+        kwargs.setdefault("write_termination", ";FF")
+        kwargs.setdefault("read_termination", ";")  # in reality its ";FF"
+        # which is, however, invalid for pyvisa. Therefore extra bytes have to
+        # be read in the read() method.
+        super().__init__(
+            resource_name,
```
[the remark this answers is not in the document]
dkriegner: used the suggested code (and also fixed the docstring) accordingly

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -0,0 +1,291 @@
...
+        map_values=True,
+        values={"Torr": "TORR",
+                "mBar": "mBAR",
+                "Pascal": "PASCAL",
+                "Micron": "MICRON",
+                },
+        check_set_errors=True,
+    )
+
+    power_status1 = Instrument.control(
```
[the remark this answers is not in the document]
dkriegner: thanks for the suggestion. I have improved the code accordingly

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -0,0 +1,291 @@
...
+        super().__init__(
+            resource_name,
+            "MKS 937B vacuum gauge controller",
+            includeSCPI=False,
+            **kwargs
+        )
+        self.address = address
+        # compiled regular expression for finding numerical values in reply strings
+        self._re_fullresponse = re.compile(fr"@{self.address:03d}(.*)")
+        self._re_response_value = re.compile(fr"@{self.address:03d}ACK(.*)")
```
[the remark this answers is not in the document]
dkriegner: I went with your suggestion and simplified the code

pymeasure/instruments/instrument.py
dkriegner: I used the code suggestion from above. Its true that is easier like you propose it. It seems to work for my device. 
I did not understand from your discussion if you still want to handle get_process differently?

pymeasure/instruments/mksinst/mks937b.py
dkriegner: I tried to understand how the protocol test could work for this device but did not succeed. From the [documentation](https://github.com/pymeasure/pymeasure/blob/master/docs/dev/adding_instruments.rst) I assumed I can write something like:

```
import pytest

from pymeasure.instruments.mksinst.mks937b import MKS937B

def test_pressure():
    """Verify the communication of the voltage getter."""
    with expected_protocol(
        MKS937B,
        [("@253PR1?", "@253ACK1.10e-9"),
         (None, b"FF")],
    ) as inst:
        assert inst.pressure1 == 1.1e-9
```

but this failed with:
` AssertionError: assert '@253ACK1.10e-9' == 1.1e-09`

Can you give me a hint on how this is done for this instrument? I guess it makes sense I write few basic tests for this to be included in this PR

dkriegner:
I can not see how the latest CI run fail is related to this pull request... If I should merge main (which might be needed anyway to have the latest protocol tests included) or perform any other changes please let me know

pymeasure/instruments/mksinst/mks937b.py
bmoneke: The test looks right and I guess it does its job, exposing coding errors 😀: `inst.pressure1` returns the string the device responded and probably you made somewhere in your code a mistake (or values does not use the `preprocess_reply`).
Maybe your `_extract_reply` does not work as expected. I'd write a test for that method as well.

dkriegner: Not sure i can agree here. testing the same with the hardware seems to work
```
>>> import pymeasure.instruments
>>> inst = pymeasure.instruments.mksinst.MKS937B("ASRL3::INSTR")
>>> inst.pressure1
1.6e-08
>>>
```
So with the hardware all functions seem to be called as I intend it. Something seems different in the test case?

dkriegner: As extra info here the raw hardware communication for the same command:
```
>>> import pyvisa
>>> rm = pyvisa.ResourceManager()
>>> inst = rm.open_resource("ASRL3::INSTR")
>>> inst.write("@253PR1?;FF")
13
>>> inst.read_bytes(18)
b'@253ACK1.50E-08;FF'
>>>
```

bmoneke: I found the problem: The ProtocolAdapter does not get the global preprocess_reply configuration, as only a newly created VISAAdapter gets that information. (I'm reworking the adapters, so that's going to change in the future).

As a workaround you can add `inst.adapter.preprocess_reply = inst._extract_reply` directly after the colon in the test. That way you add the preprocess to the adapter and the test succeeds.
In theory you can hand parameters to the ProtocolAdapter if you define them after the communication pairs. In you case, that does not work, as the preprocess_reply is a method of the instrument instance.

dkriegner: thanks with the workaround the tests succeed. I have added some to test the different types of properties of this new device.

dkriegner:
> I can not see how the latest CI run fail is related to this pull request... If I should merge main (which might be needed anyway to have the latest protocol tests included) or perform any other changes please let me know

after merging master and adding the protocol tests no more error appaers...

@bilderbuchi: The MKS937B code is ready for merging from my point of view. You certainly still want to check the changes in the Instrument. measurement getter method, but the tests present on the master branch seem to succeed

pymeasure/instruments/mksinst/mks937b.py
bmoneke: Thanks for the tests. It makes future maintenance easier. You have some style errors (one instead of two newlines etc.), see the file with the comments made by CI.

bilderbuchi: Great troubleshooting, folks! Thanks for supplying complete information; thanks for helping out getting a good test running! "This is the way!" :D

pymeasure/instruments/mksinst/mks937b.py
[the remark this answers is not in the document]
bilderbuchi: The added test takes care of this already.

pymeasure/instruments/instrument.py
bilderbuchi: > I did not understand from your discussion if you still want to handle get_process differently?

> Event though it does not offer that much flexibility (e.g. averaging), I think your implementation is good, due to backwards compatibility.

> On the other hand, applying the get process to every single value should not brake any present code.

I think it's the other way round, we have changed behaviour of `get_process` when multiple elements are present:

> Until now, get_process is acting on the string entry if vals was length 1, and on the list if length >1 (which may be less than ideal in itself). We could preserve that behaviour with some small code tweaks.

I'll think on this, come up with a test and will suggest a tweak of that part of the code.

bilderbuchi (changes requested):
We are nearly there, I think the main thing that's left is the `get_process` behaviour.

pymeasure/instruments/instrument.py
bilderbuchi: OTOH, the docstring of `get_process` is "A function that take a value and allows processing before value mapping, returning the processed value", suggesting it acts on single values. 

In addition, we have the `preprocess_reply` that acts on the whole message (before it runs through `values()`, though)

Also, having  `get_process` always act on single elements simplifies the situation when the returned list can have a variable number of elements (but I don't know how frequent that is).

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -0,0 +1,301 @@
...
+        is returned.
+        :param reply: reply string
+        :returns: string with only the response, or the original string
+        """
+        rvalue = self._re_response.search(reply)
+        if rvalue:
+            return rvalue.group('msg')
+        return reply
+
+    def _prepend_process(self, cmd):
```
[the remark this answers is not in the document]
dkriegner: my bad. fixed

dkriegner:
> We are nearly there, I think the main thing that's left is the `get_process` behaviour.

thanks for the review. I think I will leave the `get_process` for you since I do not have the bigger picture what is wanted for this particular function. Personally I think applying it to a single value makes more sense since as you wrote `preprocess_reply` covers the full message already.

If more changes are needed from me just let me know.

pymeasure/instruments/instrument.py
bilderbuchi: @msmttchr @bmoneke what do you think the behaviour of `get_process()` with multiple values should be? Act on the whole list (like now, would make code slightly more complex, but is more flexible), or individual elements (fits smoother into the control flow, more consistent with single-element list use).
I'm on the fence, and my dice are a bad basis for decisions here. :grin:

dkriegner:
something in the latest master branch breaks this code.

It can be fixed by this patch:
```
--- a/pymeasure/instruments/mksinst/mks937b.py
+++ b/pymeasure/instruments/mksinst/mks937b.py
@@ -71,7 +71,6 @@ class MKS937B(Instrument):
     """

     def __init__(self, adapter, address=253, **kwargs):
-        kwargs.setdefault("preprocess_reply", self._extract_reply)
         kwargs.setdefault("write_termination", ";FF")
         kwargs.setdefault("read_termination", ";")  # in reality its ";FF"
         # which is, however, invalid for pyvisa. Therefore extra bytes have to
@@ -108,7 +107,7 @@ class MKS937B(Instrument):
         """
         Check the read termination to correspond to the protocol
         """
-        t = self.adapter.read_bytes(2)  # read extra termination chars 'FF'
+        t = super().read_bytes(2)  # read extra termination chars 'FF'
         if t != b'FF':
             raise ValueError(f"unexpected termination string received {t}")

@@ -137,8 +136,8 @@ class MKS937B(Instrument):

         """
+        kwargs.setdefault("preprocess_reply", self._extract_reply)
         ret = super().values(self._prepend_address(command), **kwargs)
-        self._check_extra_termination()
         return ret

     def check_errors(self):
```

Is this expected to be needed due to recent changes? And is it the best way (should I push a commit with these changes and rebase on master)?

bmoneke:
In the recent change, we moved all those "special things" from a preprocess_reply into the read / write methods of the instrument itself. And `values` calls `write` and `read` (via ask), therefore you can (as you already did) remove the check from your `values` (last section of your changes).

If `_extract_reply` is useful for every read (I'd guess so), add it to `read` itself (and you can delete your `values` implementation), otherwise you can do it, as it is right now.

Another thing: You can include "_prepend_address" already in `write`, as every writing action (values, ask, or directly 'write') goes through this method.

I'd say, merging master is fine, and rebase is a better.

dkriegner:
I agree that these recent changes make a lot of sense. I now removed my overloading of `values` completely and simplified the rest as recommended. I would keep `_extract_reply` and `_prepend_address` in own methods which makes them better documented.

bilderbuchi:
I think we need some more thought on how we want `get_process` and mapping multiple values to behave (see the unresolved conversation). It's a pity that this drags out things, but changes to pymeasures core logic are always much more hairy than simply adding another instrument. I ask for your patience.

pymeasure/instruments/instrument.py
bmoneke: Sorry for not coming back. It was hard for me to decide, what to do. I definitively like the idea of doing some fancy method on the returned values list (e.g. sum and standard deviation...), but that would require to have a more complex `get_process = lambda v: 5 * v[0]` instead of just multiplying with some number.

Whenever you need complex mathematics on the return values (I'd say rarely), you can use a python property. In the many cases of a single return value, we should have a more simple `get_process`.

Therefore I propose that `get_process` acts on every single value.

bilderbuchi: Thank you! I opened #740 for nailing down the current behaviour, integrating the PR, and writing docs and changelog for this change. Let us continue the discussion there.

bilderbuchi:
Thank you, we consider your contribution complete. We will hash out the remaining details in #740, where I have also integrated your branch. This PR will close when we merge #740.

bilderbuchi:
Superseded by #772 (and later #740 for the sensor types properties)

PR #640, opened 2022-06-21
Add Digital IO Functionality to Keithley 2600 Series Driver

brandonaltieri1:
This PR adds the `Keithley2602B` driver, which inherits from `Keithley2600`. The new driver adds digital I/O (hardware triggering) functionality. This PR also adds some compatible common commands to the base `Keithley2600` class to overwrite the `Instrument` classes standard methods (which are based on SCPI commands, the commands for this driver are written in Lua). Note my pre-commit hooks did a lot of reformatting - I can undo these changes if necessary.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -49,3 +49,4 @@ Jack Van Sambeek
 JC Arbelbide
 Florian Jünger
 Benedikt Moneke
+Brandon Altieri
--- a/pymeasure/instruments/keithley/keithley2602B.py
+++ b/pymeasure/instruments/keithley/keithley2602B.py
@@ -39,10 +39,10 @@ class Keithley2602B(Keithley2600):
     """
 
     number_of_pins = 14
-    _event_descriptions = []
+    event_descriptions = []
 
     for channel in ["a", "b"]:
-        _event_descriptions += [
+        event_descriptions += [
             f"smu{channel}.trigger.SOURCE_COMPLETE_EVENT_ID",
             f"smu{channel}.trigger.MEASURE_COMPLETE_EVENT_ID",
             f"smu{channel}.trigger.PULSE_COMPLETE_EVENT_ID",
@@ -56,27 +56,6 @@ class Keithley2602B(Keithley2600):
         self.dio_pins = [
             Keithley2600DigitalIOPin(self, i + 1) for i in range(self.number_of_pins)
         ]
-
-    @staticmethod
-    def get_trigger_event_description_strings():
-        """Returns a list of the valid event IDs that can be used to select the event that
-        causes a trigger to be asserted on the digital output line. The list can be indexed
-        to set a digitial I/O line to assert a trigger given the described conditions. E.g.
-        to set digital line 4 to assert a trigger when the SMU completes a source
-        action on channel A,
-        use the following:
-
-        .. code-block:: python
-
-            #Assume a Keithley2602B object called "smu" has been successully instaniated
-            trigger_event_lists = smu.get_trigger_event_description_strings()
-            smu.dio_pins[3].stimulus = trigger_event_lists[0]
-
-        See page 9-61 of the Reference Manual for more details about the various event
-        descriptions.
-        """
-
-        return Keithley2602B._event_descriptions
 
 
 class Keithley2600DigitalIOPin:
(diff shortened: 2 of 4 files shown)


pymeasure/instruments/keithley/keithley2602B.py
```diff
@@ -0,0 +1,213 @@
...
+        detected.
+        """
+
+        log.info(
+            "Waiting for trigger for %f on pin number %d." % (timeout, self.pin_number)
+        )
+        self.write("wait(timeout)")
+        self.check_errors()
+
+    trigger_mode = Instrument.control(
```
brandonaltieri1, line 170: Is there a way to return this property as an integer? I can't cast the return value in my `ask` method as it might needs to return floats, integers, and strings, depending on the wrapper method used.

bmoneke: With "map_values" you will turn strings to integers on setting and integers to strings on getting. Right now, you cannot separate those two directions directly.

As a hack, you could have a validator, which looks up the string in the dictionary and returns the integer. Now you can set map_values to False (default) and you will get integers.

At least that is how I understand `Instrument.control`.

brandonaltieri1: I didn't realize mapping worked both with getting and setting - in this case, I want the user to get the string description (in this case, the dictionary keys) when getting this property, so I can leave it the way it is. Thank you @bmoneke !

pymeasure/instruments/keithley/keithley2602B.py
```diff
@@ -0,0 +1,212 @@
...
+
+            #Assume a Keithley2602B object called "smu" has been successully instaniated
+            trigger_event_lists = smu.get_trigger_event_description_strings()
+            smu.dio_pins[3].stimulus = trigger_event_lists[0]
+
+        See page 9-61 of the Reference Manual for more details about the various event
+        descriptions.
+        """
+
+        return Keithley2602B._event_descriptions
```
brandonaltieri1, line 79: Is there a better way to return to the user the list of acceptable event description strings?

pymeasure/instruments/keithley/keithley2602B.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def wait_for_trigger(self, timeout):
+        """This method waits for a trigger for up to a maximum of the timeout value (in seconds).
+        Returns True if a trigger was detected, false if the timout was reached and no trigger was
+        detected.
+        """
+
+        log.info(f"Waiting for trigger for {timeout} on pin number {self.pin_number}.")
+        self.write("wait(timeout)")
+        self.check_errors()
```
brandonaltieri1, line 167: Do these make sense as methods, or should they be `Instrument.control` properties?

bmoneke: I reckon properties to be appropriate getting or setting values, properties etc. So get_event_id and get_oberrun could be properties (measurement, I guess).

Waiting on a trigger or clearing etc are, however, methods which I execute on the device, therefore, it's appropriate to leave them methods. At least in my understanding.

I'd say as a rule of thumb: If you call it get_x or set_x, it should be a property. If your method neither takes a parameter nor returns something, it should be a method.

pymeasure/instruments/keithley/keithley2602B.py
bmoneke: Just make the list a public property (without the underscore), such that the user may access it via smu.event_descriptions (or with the class name in front of the dot).

I don't see the point in a static method returning an publicly accessible class variable.

AUTHORS.txt
```diff
@@ -49,4 +49,5 @@ Jack Van Sambeek
 JC Arbelbide
 Florian Jünger
 Benedikt Moneke
+Brandon Altieri
```
bilderbuchi, line 52: Until someone gets around and fixes this duplication, please also add your name to docs/about/authors.rst

bilderbuchi:
> Note my pre-commit hooks did a lot of reformatting - I can undo these changes if necessary.

Yes, please avoid this if you can ... it's blowing up the diff with unrelated changes, which makes it harder to review. 
I see the attraction of pre-commit hooks, but for now we are content with a flake8 config (maybe you can make your hooks use that one to check your edits?) that also runs in CI.

PR #642, opened 2022-06-28
Add storing unitful experiment results

bmoneke:
In #248 pint was proposed for unitful quantities.

In this pull request, I want to make the GUI elements understand pint quantities (number with units).


bmoneke:
According to the discussion in #248, I made the csv parser understand pint quantities.

pymeasure/experiment/results.py
```diff
@@ -33,11 +33,13 @@
 from string import Formatter
 
 import pandas as pd
+import pint
 
 from .procedure import Procedure, UnknownProcedure
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
+u = pint.get_application_registry()
```
bilderbuchi, line 42: I think the canonical way is to call this `ureg`.

pymeasure/experiment/results.py
```diff
@@ -129,16 +131,54 @@ def __init__(self, columns, delimiter=','):
...
 
+    @staticmethod
+    def _parse_columns(columns):
+        """Parse the columns to get units in parenthesis."""
+        units_pattern = r"\((?P<units>[\w/\(\)\*\t]+)\)"
+        units = {}
+        for column in columns:
+            try:
+                match = re.search(units_pattern, column)
+            except TypeError:
```
bilderbuchi, line 145: In what situation do you expect to get a TypeError?

pymeasure/experiment/results.py
```diff
@@ -129,16 +131,54 @@ def __init__(self, columns, delimiter=','):
...
+                match = None
+
+            if match:
+                m = match.groupdict()
+                if 'units' in m:
+                    units[column] = pint.Unit(m['units'])
```
bilderbuchi, line 151: Don't you want to duplicate the unit information in the column key? You end up with `{"accel_voltage (kV)": "kV"}`

bilderbuchi: also, for consisteny, don't you want to assign, e.g. `None` for the entries without a unit? (e.g. to keep the size/length of `CSVFormatter.columns` and `CSVFormatter.units` the same (or actually unify them in one data structure)

pymeasure/experiment/results.py
```diff
@@ -129,16 +131,54 @@ def __init__(self, columns, delimiter=','):
...
+        units = {}
+        for column in columns:
+            try:
+                match = re.search(units_pattern, column)
+            except TypeError:
+                match = None
+
+            if match:
+                m = match.groupdict()
+                if 'units' in m:
```
bilderbuchi, line 150: Can you fail that if check? Is there a way to have a not-None `match`, but then don't match the capture group?

pymeasure/experiment/results.py
```diff
@@ -129,16 +131,54 @@ def __init__(self, columns, delimiter=','):
...
 
         :param record: record to format.
         :type record: dict
         :return: a string
         """
-        return self.delimiter.join(f'{record[x]}' for x in self.columns)
+        line = []
+        for x in self.columns:
+            value = record.get(x, float("nan"))
+            units = self.units.get(x, u.dimensionless)
```
bilderbuchi, line 164: careful, dimensionless (column `Mach number (1)`) and no unit (`device_mode`) are not the same. The former will have an entry in units (I suspect `u.dimensionless`), but here you also assign it to the latter. Maybe `None` is a better default?

pymeasure/experiment/results.py
```diff
@@ -129,16 +131,54 @@ def __init__(self, columns, delimiter=','):
...
+                    value = pint.Quantity(value)
+                except pint.UndefinedUnitError:
+                    pass  # Keep it as a string
+            if isinstance(value, pint.Quantity):
+                try:
+                    line.append(f"{value.m_as(units)}")
+                except pint.DimensionalityError:
+                    line.append("nan")
+            elif units == u.dimensionless:
+                # Dimensionless quantities can be given as numbers.
```
bilderbuchi, line 176: Logic alright here? A dimensionless quantity would match in the previous clause (as it is a pint.Quantity) already. I suspect this clause can be dropped?

setup.cfg
```diff
@@ -29,6 +29,7 @@ packages = find:
 install_requires =
     numpy >= 1.6.1
     pandas >= 0.14
+    pint >= 0.18
```
bilderbuchi, line 32: We will also need an entry in requirements/pymeasure.yml

tests/experiment/test_results.py
```diff
@@ -24,10 +24,12 @@
 
 import os
 import pickle
+import pytest
```
bilderbuchi, line 27: pytest import should be with the non-stdlib group below (with pint)

tests/experiment/test_results.py
```diff
@@ -36,6 +38,9 @@
 from data.procedure_for_testing import RandomProcedure
 
 
+u = pint.get_application_registry()
```
bilderbuchi, line 41: I got this usage pattern in my bones to not duplicate uregs, so reuse the existing one from pymeasure.experiment.results instead. Is that not needed anymore?

pymeasure/experiment/results.py
bmoneke: I took the inspiration from instrumentkit (the only pint-using project I know). I guess the reasoning was, that u is far shorter and u.m as unit of meters is more concise than ureg.m.
I'm fine with `ureg` as well.

pymeasure/experiment/results.py
bmoneke: > Do you want to duplicate the unit information in the column key? You end up with `{"accel_voltage (kV)": "kV"}`

yes, I duplicate the information. Removing 'kV' silently from the column could create confusion addressing it with data. The dictionary with the "kV" entry is for faster lookup of the unit, such that I have to parse just once and not for every data point.



> also, for consisteny, don't you want to assign, e.g. `None` for the entries without a unit? (e.g. to keep the size/length of `CSVFormatter.columns` and `CSVFormatter.units` the same (or actually unify them in one data structure)

It's a dictionary. I know, that a not existing entry means, I did not specify anything. Yes, we could make it into a list of the same size with Nones, but I prefer a dictionary, which contains just the necessary information and not a row of None.

pymeasure/experiment/results.py
bmoneke: The regex part is taken from the parsing for the plotter (only with the regex pattern expanded).
I did not try to improve it.

pymeasure/experiment/results.py
bmoneke: At the point of writing that code, I had "dimensionless" as a synonym of "I don't care", such that Mach number and device_mode both fit.
Let's talk about the implementation again, after having decided on the usage.

tests/experiment/test_results.py
bmoneke: get_application_registry does exactly what you want: If the application has already a registry, it returns it, such that you use that already existing registry again. If no registry exists, it creates one.
With that method you do not have to remember, where you defined it in the first place.

setup.cfg
bmoneke: which version? I see discrepancies between setup.cfg and pymeasure.yml.

tests/experiment/test_results.py
bmoneke: I checked it, and get_application_registry works that way only after having called set_application_registry somewhere else.
So we have to declare it at some single point of entry (later on).

pymeasure/experiment/results.py
bilderbuchi: I did not say turn it into a list, a dict is the correct choice here. 
> I know, that a not existing entry means, I did not specify anything.

Yes, _you_ know. :-) Someone else might not and will wonder what the deal with the missing column entries in a dict column->unit might mean. "Explicit is better than implicit" and all that.

In the end it's probably a style/taste question. :-)

pymeasure/experiment/results.py
bilderbuchi: Yeah, but you _should_ care about dimensionless things -- in the end the way you can think of it is that it has a unit of `1`. It still has to fit unit compatibility in operations, something which device_mode does not, so there is a difference, too.
>Let's talk about the implementation again, after having decided on the usage.

:+1:

setup.cfg
bilderbuchi: pymeasure.yml gives _specific_ fixed versions that e.g. our CI is using.
setup.cfg gives _minimum_ versions (because you make use of a feature/rely on a fix that only appeared at some point. 
Speaking of which, is there a reason for your minimum of 0.18, or could it be just `pint`? If the latter, that is preferable as it aids package compatibility resolution when installing many into an environment.

bmoneke: I just used the latest version which is still compatible with python 3.7

pymeasure/experiment/results.py
bmoneke: I adjusted the implementation (and tests) according to the table in the discussion #248

pymeasure/experiment/results.py
bmoneke: I don't know, whether an explicit None makes it more clearer that you did not specify a unit to check. Basically dict.get() will return None, if something does not exist, such that it is not explicit anymore.
If we want to be explicit, I'd use '-' to indicate no unit.

tests/experiment/test_results.py
bilderbuchi: From this [documentation section](https://pint.readthedocs.io/en/stable/tutorial.html#using-pint-in-your-projects) and [this comment](https://github.com/hgrecco/pint/issues/1273#issuecomment-805369430) I take away that the best way is to put this into our codebase somewhere and use `ureg` it from there
```
import pint
ureg = pint.get_application_registry()
```
(we maybe need to put it into a separate moduel to avoid circular imports, but I haven't confirmed that)

cf also [this issue](https://github.com/hgrecco/pint/issues/1365).

bmoneke: pint 0.19 requires python>=3.8, but pymeasure still supports python3.7

pymeasure/experiment/results.py
bilderbuchi: OK. I'll leave the choice to you, whatever fits better into the workflow (also with pint parsing etc).

setup.cfg
bilderbuchi: > I just used the latest version which is still compatible with python 3.7, which is 0.18.

I see. In this case just use plain `pint`. 
As we still want to support python 3.7, we can't demand pint >=0.19 as proposed above, so use 0.18 in pymeasure.yml

>  What are the criteria to choose a version?
for the CI testing yml file, I simply use the current version at a time when I add a dependency. We can then update the versions manually from time, or when incompatibilities arise (e.g. because the numpy version becomes very old, or somesuch).

tests/experiment/test_results.py
bmoneke: > (we maybe need to put it into a separate moduel to avoid circular imports, but I haven't confirmed that)

If there is not an application_registry (via set_application_registry), it will create one. as we want to keep the registry a user thing. I propose to move (later on) `ureg = pint.get_application_registry()` in the `pymeasure.__init__.py`, such that all pymeasure modules use this registry, which in turn gets the application registry set by the user (if any).

bmoneke:
New logic (see #248 ) and requested changes are implemented.

bilderbuchi:
It would be great to have user-facing documentation about the new behaviour, especially around how to format headers, and the logic flow described in the table in #248.

bmoneke:
I put pint in a central place and added documentation for the GUI and for writing instruments.

docs/dev/adding_instruments.rst
```diff
@@ -510,6 +510,27 @@ Similar to `set_process` the :func:`Instrument.control <pymeasure.instruments.In
     >>> extreme.current
     3.1
 
+Another use-case of `set-process`, `get-process` is conversion from/to a :code:`pint.Quantity`. Modifying above example to set or return a quantity, we get:
```
bilderbuchi, line 513: Awesome! I'd like to have a big ` .. note::` here emphasizing that this is a workaround for now -- I want to add proper support for properties like outlined in https://github.com/pymeasure/pymeasure/issues/248#issuecomment-1160009874 (i.e., have a `unit` parameter for our property factories, and some TBD features to control behaviour so we can transition nicely). I think we can either open a new issue on unit support for properties themselves, and point to that, or point to #248 (the former preferred).

docs/tutorial/procedure.rst
```diff
@@ -333,4 +333,9 @@ Now that you have a background on how to use the different features of the Proce
         worker.join(timeout=3600) # wait at most 1 hr (3600 sec)
         log.info("Finished the measurement")
 
+The parentheses in the :code:`COLUMN` entries indicate the physical unit of the data in the corresponding column, e.g. :code:`'Voltage Std (V)'` indicates Volts. If you want to indicate a unitless value, e.g. Mach number, you can use `(1)` instead. Combined units like `(m/s)` or the long form `(meter/second)` are also possible. The class :class:`Results` ensures, that the data is stored in the correct unit, here Volts. For example a :python:`pint.Quantity` of 500 mV will be stored as 0.5 V. A string will be converted first to a `Quantity` and a mere number (e.g. float, int, ...) is assumed to be already in the right unit (e.g 5 will be stored as 5 V).
```
bilderbuchi, line 336: I can't see a "suggestions" button, so this way
> If you want to indicate a **unitless** value

please use "dimensionless" for such. (https://en.wikipedia.org/wiki/Dimensionless_quantity)

docs/tutorial/procedure.rst
```diff
@@ -333,4 +333,9 @@ Now that you have a background on how to use the different features of the Proce
...
+If the data entry is not compatible, either because it has the wrong unit, e.g. meters which is not a unit of voltage, or because it is no number at all, a warning is logged and `'nan'` will be stored in the file.
+If you do not specify a unit (i.e. no parentheses), no unit check is performed for this column, unless the data entry is a `Quantity` for that column. In this case, this column's unit is set to the base unit (e.g. meter if unit of the data entry is kilometers) of the data entry. From this point on, unit checks are enabled for this column. Use columns without unit checks (i.e. without parentheses) for strings or booleans.
```
bilderbuchi, line 338: *Also* use columns without unit checks...

docs/dev/adding_instruments.rst
```diff
@@ -510,6 +510,27 @@ Similar to `set_process` the :func:`Instrument.control <pymeasure.instruments.In
...
+
+.. testcode::
+
+	from pymeasure.units import ureg
```
bilderbuchi, line 517: there's a tab indent here.

docs/dev/adding_instruments.rst
bmoneke: #666

tests/instruments/test_instrument.py
```diff
@@ -213,6 +215,61 @@ def __init__(self):
...
+            super().__init__(adapter, "test", **kwargs)
+        x = Instrument.measurement(
+            "x", "doc", get_process=lambda v: ureg.Quantity(v, ureg.m))
+    with expected_protocol(Fake, [("x", "5.5")]) as instr:
+        y = instr.x
+        assert y.m_as(ureg.m) == 5.5
+
+
+@pytest.mark.parametrize("dynamic", [False, True])
+def test_measurement_dict_str_map(dynamic):
```
bilderbuchi, line 255: wait is this test just copied from below?

bmoneke: Funny, my local file does not have it twice...

bilderbuchi: Huh, that's weird!

docs/dev/adding_instruments.rst
```diff
@@ -510,6 +510,31 @@ Similar to `set_process` the :func:`Instrument.control <pymeasure.instruments.In
...
+    from pymeasure.units import ureg
+
+    Extreme5000.current = Instrument.control(
+        ":CURR?", ":CURR %g",
+        """ A floating point quantity property representing the measurement current 
+        """,
+        values=[0, 10],
+        set_process=lambda v: v.m_as(ureg.mA),  # send the value as mA to the device
+        get_process=lambda v: ureg.Quantity(v, ureg.mA),  # convert to quantity
+
```
bilderbuchi, line 526: Missing a closing parens. BTW, you can run "make doctest" in der docs directory locally, too.

bmoneke: Thanks

tests/instruments/test_instrument.py
bmoneke: Now it's here (and is going to be deleted)

docs/dev/adding_instruments.rst
bilderbuchi: ah yeah, sure.
I don't notice such anymore, my pymeasure environment has been accumulating packages for quite some time. ^^
Are you using conda or pip for environment managing?

bilderbuchi: ah yeah, sure.
I don't notice such anymore, my pymeasure environment has been accumulating packages for quite some time. ^^
Are you using conda or pip for environment managing?

bilderbuchi:
I think this is now complete. :-) Agreed?

bmoneke:
Agreed! :-)

docs/dev/adding_instruments.rst
bmoneke: pip. Anaconda made problems with packages etc, such that I changed to pip.

PR #644, opened 2022-06-30
Dev/emcenter

IanCharboneau:
Added basic support for controlling the positioner card of the ETS-lindgaren emcenter moudular rf platform


pymeasure/instruments/etslindgren/m7006_001.py
```diff
@@ -0,0 +1,174 @@
...
+    @property
+    def acceleration(self):
+        """ Returns the acceleration of the device, im seconds """
+        return self.adapter.ask(f'{self._slot}lf{self._device}ACC?')
+    
+    @acceleration.setter
+    def acceleration(self, value):
+        """ Sets the acceleration of the device, in seconds. vaule must be between 0.1 and 30.0 """
+        value = strict_range(value, 0.1, 30.0)
+        return self.adapter.ask(f'{self._slot}{self._device}ACC {value}')
```
bmoneke, line 73: As coding style, we try to use our own [style of properties](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#advanced-properties).

Besides, you do not have to rely on the adapter directly, but can use self.read/self.write/self.values. These methods can contain the string formatting as well.

Here is an example for the rewritten methods and how to write a property.

```suggestion
    def write(self, command):
        super().write(f"self._slot}{self._device}{command}")

    def values(self, command, **kwargs):
        return super().values(f"self._slot}{self._device}{command}", **kwargs)

    acceleration = Instrument.control(
        "ACC?", "ACC %g",
        """Acceleration of the device in seconds. Settable 0.1 and 30, and gettable.""",
        values=(0.1, 30),
        validator=strict_range,
        )

**Review by bmoneke — COMMENTED** [review]
Thanks for your suggestion.

Try to follow the hints of the Docs and Linting hinter.

We try to use Properties in many cases. For that reason we have specific methods, which generate our properties (Instrument.control, Instrument.measurement - read only, Instrument.setting - write only). Please use those, where applicable. I gave you an example down below, linking to the tutorial as well.

You can move all the "slot" formatting into a single write (or values for our properties) method.

File: pymeasure/instruments/etslindgren/emcenter.py [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+
+       
+
+    
+    """
+
+    local = Instrument.control(
+            'LOCAL',
+            """ Sets the EMCenter to local mode.""",
+    )
```
bmoneke, line 60: If you set something, but cannot read it, Instrument.setting is your method of choice.

pymeasure/instruments/etslindgren/emcenter.py
```diff
@@ -0,0 +1,83 @@
...
+    reboot = Instrument.control(
+            'REBOOT SYSTEM',
+            """ Reboots the EMCenter.""",
+    )
```
bmoneke, line 64: I'd keep reboot as a method, as you call it and don't want to set "reboot" to a value, nor want to read the value of "reboot".

pymeasure/instruments/etslindgren/m7006_001.py
bmoneke: You can enable "check_errors" for the getter and write a check_errors method (I have to look up the correct names tomorrow), which reads the response.

That is the preferred way dealing with setters which return an answer.

IanCharboneau: I ended up not needing errors.

pymeasure/instruments/etslindgren/m7006_001.py
```diff
@@ -47,127 +47,110 @@ def __init__(self, resource_name, slot=1, device="A", **kwargs):
...
+
+    down = Instrument.setting(
+        "DN",
+        """Move the mast down.""",
+    )
+    up = Instrument.setting(
+        "UP",
+        """Move the mast up.""",
+    )
+    stop = Instrument.setting("ST", """Stop the device.""")
```
bmoneke, line 109: A setting is called in the style `instr.cw = 5`. These settings, however, seem to set not a value, at least they do not have a parameter.
Would it be better to have a setting `instr.rotation = "cw"/"ccw"`? and `direction = up/down`? You can use `values={'down': "DN", 'up': "UP"}` and `map_values=True` in order to convert english words to the abbreviation.

Stop should be a method, I think, as you do not set anything or get anything.

pymeasure/instruments/etslindgren/emcenter.py
```diff
@@ -0,0 +1,53 @@
...
+import logging
+
+from pymeasure.instruments import Instrument
+
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class EMCenter(Instrument):
```
msmttchr, line 36: The purpose of this class is to implement the "LOCAL" command and `reboot` method only. Is this correct ?
Is there any link between main frame and sub-modules ?

For the order of elements in a class, I would suggest to use the one mentioned here:
https://github.com/pymeasure/pymeasure/pull/535#issuecomment-1009202829

pymeasure/instruments/etslindgren/emcenter.py
```diff
@@ -0,0 +1,53 @@
...
+    """Represents ETS-Lindgren EMCenter main frame.
+    used for direct access to the instrument, does not control sub-modules
+    """
+
+    local = Instrument.setting(
```
msmttchr, line 41: This property is a "setting" one, but it just ignores what is passed. Is this intentional ?
Normally you need to use this as follow:
`inst.local = some_value`
The command "LOCAL" will be issued regardless of the value of some_value.

pymeasure/instruments/etslindgren/m7006_001.py
```diff
@@ -0,0 +1,154 @@
...
+log.addHandler(logging.NullHandler())
+
+
+class M7006_001(Instrument):
+    """Position control card of EMCenter modular system.
+    Allows for controll and feedback of the positioner card.
+    Comunication happes transparently through the EMCenter mainframe.
+    devices share the resource name of the mainframe, and
+    are identified by the slot number and the device number
+    (if applicable to the modual).
```
msmttchr, line 40:
```suggestion
    """Position control card of EMCenter modular system.
    
    Allows for control and feedback of the positioner card.
    Communication happens transparently through the EMCenter mainframe.
    Devices share the resource name of the mainframe, and
    are identified by the slot number and the device number
    (if applicable to the module).
```

pymeasure/instruments/etslindgren/m7006_001.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+
+    def __init__(self, resource_name, slot=1, device="A", **kwargs):
+
+        kwargs.setdefault("write_termination", "\n")
+        super().__init__(resource_name, "test", **kwargs)
```
msmttchr, line 46: "test" probably needs to be updated with something more meaningful. Right?

pymeasure/instruments/etslindgren/m7006_001.py
```diff
@@ -0,0 +1,154 @@
...
+
+import logging
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class M7006_001(Instrument):
```
msmttchr, line 34: For the order of elements in a class, I would suggest to use the one mentioned here:
https://github.com/pymeasure/pymeasure/pull/535#issuecomment-1009202829

msmttchr (changes requested):
@IanCharboneau thank for your contribution. Please find attached first batch of comments.

docs/api/instruments/etslindgren/m7006_001.rst
```diff
@@ -0,0 +1,7 @@
+###################################################################
+Ets-lindgren emcenter modualr rf platorm , Postioner control modual
```
bilderbuchi, line 2:
```suggestion
Ets-lindgren emcenter modular RF platform , Positioner control module
```

pymeasure/instruments/etslindgren/emcenter.py
bilderbuchi: Agreed, assuming there is a "mode" other than "local", this should be a setting/control called "mode" (which then accepts values "local" or "something else". 
If there is only local mode, this should be a method (e.g. `set_local_mode(self)`)

pymeasure/instruments/etslindgren/emcenter.py
bilderbuchi: I'm guessing this is WIP and will still get the modular card class instances (like channels)?

pymeasure/instruments/etslindgren/m7006_001.py
```diff
@@ -0,0 +1,154 @@
...
+        values=(0.0, 100.0),
+        validator=strict_range,
+    )
+
+    aux1 = Instrument.control(
+        "AUX1?",
+        "AUX1 %g",
+        """Auxiliary input 1, Settable to ON or OFF, and gettable.""",
+        values=("ON", "OFF"),
+        validator=strict_discrete_set,
```
bilderbuchi, line 80: A better pattern is to call this property `aux1_enabled` and use `map_values=True` and `values={True: "ON", False: "OFF"}` -- then you can use Python booleans to more naturally interact with this.

Same below.

pymeasure/instruments/etslindgren/m7006_001.py
```diff
@@ -0,0 +1,154 @@
...
+        values=("CW", "CCW"),
+        validator=strict_discrete_set,
+    )
+
+    movement = Instrument.setting(
+        "%g",
+        """Move the tower up or down.""",
+        values={"up": "UP", "down": "DN"},
+        map_values=True,
+    )
```
bilderbuchi, line 103: Do these move the device by one increment/step or so? If yes, I think methods `up()/down()/cw()/ccw()` (or `move(dist=+1)/rotate(angle=1)` could be more appropriate -- with the added benefit that you could pass a step size/distance to those methods, too.

bilderbuchi (changes requested):
some remarks from my side.

pymeasure/instruments/etslindgren/emcenter.py
IanCharboneau: Thank you for the suggestions and comments. I've made most of the changes. 
Can you give me some more details about what you mean by modular card class instances? How this "instrument" works is that it is a linux box that has intefaces for control (LAN/USBSERIAL/GBIP<I think it might be able to control other intr though this too>). It also Has expantion slots on the back for the modular cards (controlled by <slot><sub-card device id><command>).
Right now I control it without using the mainframe itself and just connecting "directly" to the card.
I also only have a positional card onhand, and have implemented most of the basic features that I have equipment for. 
The ones I haven't implement are mainly used for inital setup and wont be used often in normal op, but I can implement them if desired
Thanks

BenediktBurger:
I close this stale pull request. It can be opened, whenever you wish to continue.

Issue #94, opened 2017-07-13
Introduce a compatibility layer to enable pymeasure to use InstrumentKit drivers


bilderbuchi:
A "compatibility layer" that would enable pymeasure to use existing [InstrumentKit](https://github.com/Galvant/InstrumentKit) drivers would be a great first step to move the pymeasure driver portion to InstrumentKit, and use that package's drivers, If necessary, incompatible information (e.g. units) could be discarded in a first iteration, to reduce the complexity of this endeavour.
cc @cjermain @scasagrande #53

cjermain:
@bilderbuchi, PR welcome. I plan to focus my time on better test coverage at this point, but would be happy to see this effort progress.

bilderbuchi:
Better test coverage is good! It will definitely take me quite some time to get a feel for both codebases enough so that I can try to contribute something.

bmoneke:
I started a first version in my branch [IKcompatibility](https://github.com/bmoneke/pymeasure/tree/IKcompatibility). It uses the testing environment, we are implementing right now (branch [pymeasure/protocol-tests](https://github.com/pymeasure/pymeasure/tree/protocol-tests), therefore I did not do a pull request.

At least using the ProtocolAdapter works as expected. I did not yet connect a real instrument.

bmoneke:
Looking more into the GUI of pymeasure, I did not find anything, which required a specific type of instrument.

Does an instrument require anything special, in order to be able to be used by pymeasure GUI? @CasperSchippers @msmttchr

If not, it could be advertised, that any python instrument may be used, increasing the coverage of pymeasure GUI.

msmttchr:
> Looking more into the GUI of pymeasure, I did not find anything, which required a specific type of instrument.
> 
> Does an instrument require anything special, in order to be able to be used by pymeasure GUI? @CasperSchippers @msmttchr
> 
> If not, it could be advertised, that any python instrument may be used, increasing the coverage of pymeasure GUI.

I think this is correct. The GUI provide a framework to execute experiments and data is usually retrieved from Pymeasure instrument, but this not strictly required.

neal-kepler:
> Looking more into the GUI of pymeasure, I did not find anything, which required a specific type of instrument.
> 
> Does an instrument require anything special, in order to be able to be used by pymeasure GUI? @CasperSchippers @msmttchr
> 
> If not, it could be advertised, that any python instrument may be used, increasing the coverage of pymeasure GUI.

[...]

PR #647, opened 2022-07-02
Add an example and documentation regarding using a foreign instrument

bmoneke:
I adapted a GUI example to using an instrument of another package (here InstrumentKit) in order to show the compatibility (see #94 ).
I also added a sentence in the docu in order to show, that the GUI is not restricted to the included instruments.

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -34,3 +34,5 @@ These features are explored in the :doc:`Using a graphical interface <tutorial/g
 
 .. image:: tutorial/pymeasure-managedwindow-running.png
     :alt: ManagedWindow Running Example
+
+The GUIs are not restricted to the instruments included in this package. Any python instrument may be used.
--- a/examples/Basic/gui_foreign_instrument.py
+++ b/examples/Basic/gui_foreign_instrument.py
@@ -1,0 +1,99 @@
+"""
+This example demonstrates how to use an instrument from another library, here
+InstrumentKit, in the pymeasure graphical interface.
+It is a modification of the `gui_custom_inputs.py`: The `MainWindow` is the
+same, the `TestProcedure` was adjusted to use an instrumentKit instrument
+instead of a random number generator.
+
+Run the program by changing to the directory containing this file and calling:
+
+python gui_foreign_instrument.py
+"""
+
+import sys
+import tempfile
+from time import sleep
+
+from pymeasure.experiment import Procedure, IntegerParameter, FloatParameter
+from pymeasure.experiment import Results
+from pymeasure.display.Qt import QtGui, fromUi
+from pymeasure.display.windows import ManagedWindow
+
+# Import the InstrumentKit
+from instruments.thorlabs.pm100usb import PM100USB
+import instruments.units as u
+
+
+import logging
+log = logging.getLogger('')
+log.addHandler(logging.NullHandler())
+
+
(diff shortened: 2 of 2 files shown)


examples/Basic/gui_foreign_instrument.py
```diff
@@ -0,0 +1,99 @@
...
+
+class TestProcedure(Procedure):
+
+    iterations = IntegerParameter('Loop Iterations', default=100)
+    delay = FloatParameter('Delay Time', units='s', default=0.2)
+
+    DATA_COLUMNS = ['Iteration', 'Power (W)']
+
+    def startup(self):
+        log.info("Setting up random number generator")
```
bilderbuchi, line 40:
```suggestion
        log.info("Setting up the power meter")
```

examples/Basic/gui_foreign_instrument.py
```diff
@@ -0,0 +1,99 @@
...
+import tempfile
+from time import sleep
+
+from pymeasure.experiment import Procedure, IntegerParameter, FloatParameter
+from pymeasure.experiment import Results
+from pymeasure.display.Qt import QtGui, fromUi
+from pymeasure.display.windows import ManagedWindow
+
+# Import the InstrumentKit
+from instruments.thorlabs.pm100usb import PM100USB
```
bilderbuchi, line 23: Does IK have a "fake instrument" feature one could use to test this even when not having the hardware around? I'm not familiar enough to know.

docs/introduction.rst
```diff
@@ -34,3 +34,5 @@ These features are explored in the :doc:`Using a graphical interface <tutorial/g
 
 .. image:: tutorial/pymeasure-managedwindow-running.png
     :alt: ManagedWindow Running Example
+
+The GUIs are not restricted to the instruments included in this package. Any python instrument may be used.
```
bilderbuchi, line 38:
```suggestion
The GUIs are not restricted to the instruments included in this package. Any python instrument may be used.
For example,  `this script <https://github.com/pymeasure/pymeasure/tree/master/examples/Basic/gui_foreign_instrument.py>`_ demonstrates how to use an `InstrumentKit <instrumentkit.readthedocs.io/>`_ instrument.
```
(I hope I didn't make any Sphinx mistakes blindly :D)

examples/Basic/gui_foreign_instrument.py
```diff
@@ -0,0 +1,99 @@
...
+import sys
+import tempfile
+from time import sleep
+
+from pymeasure.experiment import Procedure, IntegerParameter, FloatParameter
+from pymeasure.experiment import Results
+from pymeasure.display.Qt import QtGui, fromUi
+from pymeasure.display.windows import ManagedWindow
+
+# Import the InstrumentKit
```
bilderbuchi, line 22:
```suggestion
# Import the InstrumentKit package
```

bilderbuchi (changes requested):
Thanks for this! Looks good with some minor tweaks.

examples/Basic/gui_foreign_instrument.py
bmoneke: I changed it to use a "Test-Adapter", such that it works, even without an actual device connected.

Further I adjusted it to be based on the most basic gui example.

bilderbuchi:
Thanks!

PR #649, opened 2022-07-15
Add support for Active Technologies AWG-401x

garzetti:
Active Technologies AWG-401x is a dual mode Arbitrary Function
Generator (AFG) and Arbitrary Waveform Generator (AWG) 1.2GS/s
instrument. This instrument is available with two (AWG-4012), four
(AWG-4014), and eight (AWG-4018) channel configurations, all supported
by this module. Depending on the model, 8, 16, or 32 digital channels
are available as options.

This module implements the interface of both modes in two distinct
classes: AWG401x_AFG and AWG401x_AWG.

In AWG401x_AFG, most of the functionality of AFG mode has been
implemented: selection of waveshape, amplitude, frequency, offset,
phase, ... . All the parameters are validated at runtime based on the
validity ranges provided by the instrument, which may change based on
hardware configuration and license status.

In AWG401x_AWG, the basic functionality of AWG mode has been
implemented. This includes the selection of basic amplitude and time
parameters and the selection of the entries' waveforms. Additionally,
the possibility to upload custom waveforms has been included.

The module has been implemented using the official documentation
available at https://www.activetechnologies.it. Portions of the
documentation have been included in the methods' docstring after
written permission from Active Technologies.

Notable missing features (currently) are the control of digital
channels, the flow control of the entries in AWG mode, and the chaining
of multiple instruments.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -49,3 +49,4 @@ Jack Van Sambeek
 JC Arbelbide
 Florian Jünger
 Benedikt Moneke
+Fabio Garzetti
--- a/pymeasure/instruments/activetechnologies/AWG401x.py
+++ b/pymeasure/instruments/activetechnologies/AWG401x.py
@@ -40,7 +40,7 @@ from pymeasure.instruments.validators import strict_discrete_set,\
     strict_range, joined_validators
 
 
-class Channel_base(Instrument):
+class ChannelBase(Instrument):
     """Implementation of a base Active Technologies AWG-4000 channel."""
 
     enabled = Instrument.control(
@@ -104,7 +104,7 @@ class Channel_base(Instrument):
                                               str(self.channel_number)))
 
 
-class Channel_AFG(Channel_base):
+class ChannelAFG(ChannelBase):
     """Implementation of a Active Technologies AWG-4000 channel in AFG mode."""
 
     load_impedance = Instrument.control(
@@ -357,7 +357,7 @@ class Channel_AFG(Channel_base):
             [self.baseline_offset_min, self.baseline_offset_max]))
 
 
-class Channel_AWG(Channel_base):
+class ChannelAWG(ChannelBase):
     """Implementation of a Active Technologies AWG-4000 channel in AWG mode."""
 
     # Default delay override
@@ -450,7 +450,7 @@ class AWG401x_base(Instrument):
         else:
             raise ValueError("position value outside permitted range [0,4]")
 
-    def waitLast(self):
+    def wait_last(self):
         """Wait for last operation completition"""
 
         self.write("*WAI")
@@ -510,7 +510,7 @@ class AWG401x_AFG(AWG401x_base):
 
         self.ch = {}
         for i in range(1, num_ch+1):
-            self.ch[i] = Channel_AFG(self, i)
+            self.ch[i] = ChannelAFG(self, i)
 
 
 class AWG401x_AWG(AWG401x_base):
@@ -716,7 +716,7 @@ class AWG401x_AWG(AWG401x_base):
 
         self.setting_ch = {}
         for i in range(1, self.num_ch+1):
-            self.setting_ch[i] = Channel_AWG(self, i)
+            self.setting_ch[i] = ChannelAWG(self, i)
 
         self.entries = self.DummyEntriesElements(self, self.num_ch)
         # self.sequencer = Sequencer(self, self.num_ch)
@@ -734,45 +734,45 @@ class AWG401x_AWG(AWG401x_base):
         """Force a trigger event to occour."""
         self.write("TRIGger:SEQuence:IMMediate")
 
-    def saveFile(self,
-                 fileName: str,
-                 data,
-                 path: str = None,
-                 overrideExisting=False):
+    def save_file(self,
+                  file_name: str,
+                  data,
+                  path: str = None,
+                  override_existing=False):
         """Write a string in a file in the instrument"""
 
         if path is not None:
             raise NotImplementedError("Specify path is not implemented")
 
-        if fileName in [file.name
-                        for file in self.listFiles(path=path)
-                        if file.type == '']:
-            if overrideExisting:
-                self.removeFile(fileName, path=path)
+        if file_name in [file.name
+                         for file in self.list_files(path=path)
+                         if file.type == '']:
+            if override_existing:
+                self.remove_file(file_name, path=path)
             else:
                 raise ValueError("File already exist and override is disabled")
 
-        self.write('MMEM:DATA "' + fileName + '", 0, ' +
+        self.write('MMEM:DATA "' + file_name + '", 0, ' +
                    IEEE488DataBlock.encode(data))
 
         # HACK: Send an unuseful command to ensure the last command was
         # executed because if it is more than 1024 bytes it doesn't work
-        self.waitLast()
+        self.wait_last()
 
-    def removeFile(self, fileName: str, path: str = None):
+    def remove_file(self, file_name: str, path: str = None):
         """Remove a specified file"""
 
         if path is not None:
             raise NotImplementedError("Specify path is not implemented")
 
-        if fileName not in [file.name
-                            for file in self.listFiles(path=path)
-                            if file.type == '']:
+        if file_name not in [file.name
+                             for file in self.list_files(path=path)
+                             if file.type == '']:
             raise ValueError("File do not exist")
 
-        self.write('MMEMORY:DELETE "' + fileName + '"')
+        self.write('MMEMORY:DELETE "' + file_name + '"')
 
-    def listFiles(self, path: str = None):
+    def list_files(self, path: str = None):
         """Return a List of tuples with all file found in a directory. If the
         path is not specified the current directory will be used"""
 
@@ -803,7 +803,7 @@ class AWG401x_AWG(AWG401x_base):
         def __getitem__(self, key):
             """Load data from instrument if not present"""
             if self._data[key] is None:
-                self._data[key] = self._getWaveform(key)
+                self._data[key] = self._get_waveform(key)
             return self._data[key]
 
         def __setitem__(self, key, value):
@@ -829,9 +829,9 @@ class AWG401x_AWG(AWG401x_base):
                     f"which is "
                     f"{self.instrument.entries[1].ch[1].voltage_low_min}V")
 
-            self.instrument.saveFile(f"{key}.txt",
-                                     "\n".join(map(str, value)),
-                                     overrideExisting=True)
+            self.instrument.save_file(f"{key}.txt",
+                                      "\n".join(map(str, value)),
+                                      override_existing=True)
 
             try:
                 del self[key]
@@ -841,9 +841,9 @@ class AWG401x_AWG(AWG401x_base):
             self.instrument.write(f'WLISt:WAVeform:IMPort "{key}",'
                                   f'"{key}.txt",ANAlog')
 
-            self.instrument.waitLast()
+            self.instrument.wait_last()
 
-            self.instrument.removeFile(f"{key}.txt")
+            self.instrument.remove_file(f"{key}.txt")
 
             self._data[key] = None
             return
@@ -872,13 +872,13 @@ class AWG401x_AWG(AWG401x_base):
 
         def reset(self):
             """Reset the class reloading the waveforms from instrument"""
-            waveformsName = self.instrument.values("WLISt:LIST?")
-            self._data = {v: None for v in waveformsName}
+            waveforms_name = self.instrument.values("WLISt:LIST?")
+            self._data = {v: None for v in waveforms_name}
 
-        def _getWaveform(self, waveformName):
+        def _get_waveform(self, waveform_name):
             """Get the waveform point of a specified waveform"""
             bin_value = self.instrument.binary_values(
-                'WLISt:WAVeform:DATA? "' + waveformName + '"',
+                'WLISt:WAVeform:DATA? "' + waveform_name + '"',
                 dtype=np.byte)
 
             bin_wave = IEEE488DataBlock.decode(bin_value)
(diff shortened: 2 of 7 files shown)


AUTHORS.txt
```diff
@@ -49,3 +49,4 @@ Jack Van Sambeek
 JC Arbelbide
 Florian Jünger
 Benedikt Moneke
+Fabio Garzetti
```
msmttchr, line 52: Please update also docs/about/authors.rst.
Unfortunately we have this duplication.

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1145 @@
...
+        get_process=lambda v: AWG401x_AWG._run_state_dict[v]
+    )
+
+    enabled = Instrument.control(
+        "AWGControl:RSTATe?", "AWGControl:%s",
+        """A boolean property that enables the generation of signals.""",
+        validator=strict_discrete_set,
+        values={True: "RUN", False: "STOP"},
+        map_values=True,
+        get_process=lambda v: "STOP" if v == 0 else "RUN"
```
msmttchr, line 692: This property has different commands to set and get. Is this intentional?

msmttchr (changes requested):
@garzetti, first batch of comments on this very interesting PR. I like you tried to address problems like instrument channels with dynamic properties.

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1145 @@
...
+        ValueError
+            If position is outside permitted limit [0,4].
+        """
+
+        if position >= 0 or position <= 4:
+            self.write(f"*RCL {position}")
+        else:
+            raise ValueError("position value outside permitted range [0,4]")
+
+    def waitLast(self):
```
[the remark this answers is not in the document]
msmttchr: Similar comment for other method's name

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1145 @@
...
+        "AFGControl:STATus?", "AFGControl:%s",
+        """A boolean property that enables the generation of signals.""",
+        validator=strict_discrete_set,
+        values={True: "START", False: "STOP"},
+        map_values=True,
+        get_process=lambda v: "START" if v == 1 else ("STOP" if v == 0 else v)
+    )
+
+    def __init__(self, adapter, **kwargs):
+        super().__init__(adapter)
```
msmttchr, line 498: kwargs are silently discarded. Is this intentional?

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1145 @@
...
+            self.voltage_high_values = list(map(
+                float,
+                [self.voltage_high_max, self.voltage_high_min]))
+
+            self.voltage_low_values = list(map(
+                float,
+                [self.voltage_low_max, self.voltage_low_min]))
+
+
+class IEEE488DataBlock():
```
msmttchr, line 1112: I wonder if this class can be removed by using utilities provided by `pyvisa`.
See pymeasure/adapters/serial.py for some example

physicsninja: @msmttchr For some context, this instrument can also accept data in the form of a text string of numbers separated by newlines, not just arrays of numbers. When I wrote a much more hacky implementation of this instrument (under the US distributer name Berkeley Nucleonics 675)  [here](https://github.com/neal-kepler/pymeasure/blob/master/pymeasure/instruments/berkeleynucleonics/bn675.py), I made the same choice. In light of your comment I think [write_binary](https://pyvisa.readthedocs.io/en/latest/introduction/rvalues.html) with datatype='s' may actually work, I didn't realize at the time that you could pass raw strings. @garzetti can correct me if they have already tried.

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1145 @@
...
+    )
+
+    polarity = Instrument.control(
+        "OUTPut<ch>:POLarity?", "OUTPut<ch>:POLarity %s",
+        """This property inverts the output waveform relative to its average
+        value: (High Level – Low Level)/2. NORM for normal, INV for inverted
+        """,
+        validator=strict_discrete_set,
+        values=["NORMAL", "NORM", "INVERTED", "INV"],
+        get_process=lambda v: "NORM" if v == 0 else ("INV" if v == 1 else v)
```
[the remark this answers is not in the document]
garzetti: Because the get and set arguments are different. The set use `NORMAL` and `INVERTED` while get use `0` and `1`

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1145 @@
...
+    )
+
+    delay = Instrument.control(
+        None, None,
```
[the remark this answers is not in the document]
garzetti: Yes, since the command for the AFG and AWG channel's delay are different i will set in the dedicated class derived from `Channel_base`. e.g., https://github.com/pymeasure/pymeasure/blob/9e0d254463e05259eff6093f21b7c761fc5aba42/pymeasure/instruments/activetechnologies/AWG401x.py#L145-L149

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1145 @@
...
+        """This property queries the minimum delay that can be set to the
+        output waveform.""",
+        dynamic=True
+    )
+
+    def __init__(self, instrument, channel_number):
+        self.instrument = instrument
+        self.channel_number = channel_number
+
+        self._special_names = self._setup_special_names()
```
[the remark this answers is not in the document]
garzetti: Yes, right

pymeasure/instruments/activetechnologies/AWG401x.py
garzetti: Yes, it is.  Here there are tree different command for a single property:
 * AWGControl: RSTATe? (Query Only)
 * AWGControl:RUN (No Query Form)
 * AWGControl:STOP (No Query Form)

AUTHORS.txt
garzetti: Done

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1145 @@
...
+
+    def __init__(self, adapter, **kwargs):
+
+        try:
+            if isinstance(adapter, (int, str)):
+                adapter = VISAAdapter(adapter, **kwargs)
+                # Insert an higher timeout because often, when starting the
+                # instrument, can pass some time
+                adapter.connection.timeout = 7500
+        except ImportError:
```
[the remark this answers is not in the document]
garzetti: Done

pymeasure/instruments/activetechnologies/AWG401x.py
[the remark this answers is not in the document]
garzetti: Done

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1145 @@
...
+        """This property queries the minimum sample rate for the Sampling
+        Clock."""
+    )
+
+    _run_state_dict = {0: "STOPPED", 1: "WAITING_TRIGGER", 2: "RUNNING"}
+
+    run_status = Instrument.measurement(
+        "AWGControl:RSTATe?",
+        """This property returns the run state of the AWG. The possible values
+        are: STOPPED, WAITING_TRIGGER, RUNNING""",
```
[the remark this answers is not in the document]
garzetti: Done

pymeasure/instruments/activetechnologies/AWG401x.py
garzetti: Fixed

pymeasure/instruments/activetechnologies/AWG401x.py
garzetti: I have removed the `IEEE488DataBlock` class, as suggested by @msmttchr , and used `from_ieee_block` and `to_ieee_block` methods from pyvisa.util . 
As suggested by @physicsninja , probably using `query_binary_values` and `write_binary_values` from pyvisa will work, but in my opinion, it is better to modify the `pymeasure.adapter.visa` class to include these two methods instead of calling it directly from pyvisa.

garzetti:
Feel free to squash the commits at the end of PR review process

pymeasure/instruments/activetechnologies/AWG401x.py
msmttchr: What about passing the keyword parameter `datatype='s'` to `binary_values` and `write_binary_values` ? Does this the trick?

garzetti: The `write_binary_values` will work because it simply call the pyvisa `write_binary_values` method in which specifying `datatype='s'` works correctly. I can not use `binary_values` because it does not expose directly the pyvisa `query_binary_values` in which i must use `header_fmt='ieee'` and `datatype='d'` (yes, the data type are different read direction).
I can suggest two solution:
 * Use directly the `connection` object in visa adapter (i.e., `self.instrument.connection.query_binary_values(...)` )
 * Modify the visa adapter to expose directly a `query_binary_values` method like the `write_binary_values` one

msmttchr: @physicsninja and @garzetti thanks for your valuable feedback and clarification.
Let me share my opinion below.

> The `write_binary_values` will work because it simply call the pyvisa `write_binary_values` method in which specifying `datatype='s'` works correctly. I can not use `binary_values` because it does not expose directly the pyvisa `query_binary_values` in which i must use `header_fmt='ieee'` and `datatype='d'` (yes, the data type are different read direction). I can suggest two solution:
> 
>     * Use directly the `connection` object in visa adapter (i.e., `self.instrument.connection.query_binary_values(...)` )
> 
This has the disadvantage to restrict support (when using GPIB interface) to adapters supported by pyvisa (e.g. cheap Prologix is not supported), but the instrument comes by default with LAN interface, so it should be OK.

>     * Modify the visa adapter to expose directly a `query_binary_values` method like the `write_binary_values` one
The adapter are under restructuring, so, I would avoid to add more methods, especially because this needs to be kept consistent with other adapters. I would be in favor of doing this extension in a dedicated PR. If you have time have a look at the ongoing discussion in #617.

In summary, I would vote for the first option you suggested.

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1106 @@
...
+
+    def __init__(self, adapter, **kwargs):
+        super().__init__(adapter, **kwargs)
+
+        self.setting_ch = {}
+        for i in range(1, self.num_ch+1):
+            self.setting_ch[i] = ChannelAWG(self, i)
+
+        self.entries = self.DummyEntriesElements(self, self.num_ch)
+        # self.sequencer = Sequencer(self, self.num_ch)
```
msmttchr, line 715:
```suggestion
```

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -0,0 +1,1106 @@
...
+
+    def trigger(self):
+        """Force a trigger event to occour."""
+        self.write("TRIGger:SEQuence:IMMediate")
+
+    def save_file(self,
+                  file_name: str,
+                  data,
+                  path: str = None,
+                  override_existing=False):
```
msmttchr, line 734: type hinting is not encouraged in `PyMeasure`, at least from some discussion dated back in 2020 (see #259).

pymeasure/instruments/activetechnologies/AWG401x.py
garzetti: @msmttchr thanks for your suggestion. Modified to use the first solution.

garzetti:
In the commit 2c618132d86e11383795cff38130902e07f8dfb1 I have modified how the properties are converted in float. Now the string are converted directly via `get_process` lambda. @msmttchr let me know if you prefer as it was before

pymeasure/instruments/activetechnologies/AWG401x.py
garzetti: type hinting removed

msmttchr:
> In the commit [2c61813](https://github.com/pymeasure/pymeasure/commit/2c618132d86e11383795cff38130902e07f8dfb1) I have modified how the properties are converted in float. Now the string are converted directly via `get_process` lambda. @msmttchr let me know if you prefer as it was before

Sorry, I have missed this point during my original review. Actually, measurements properties cast everything to `float` by default and you can override this by passing a keyword parameter 'cast=<something>'. This is achieved because the property `fget` calls `values` method in the adapter and it passes `cast` parameter to it.
A bit weird for my taste.
Long story to say that there's no need to add extra code to cast to `float`.

garzetti:
> > In the commit [2c61813](https://github.com/pymeasure/pymeasure/commit/2c618132d86e11383795cff38130902e07f8dfb1) I have modified how the properties are converted in float. Now the string are converted directly via `get_process` lambda. @msmttchr let me know if you prefer as it was before
> 
> Sorry, I have missed this point during my original review. Actually, measurements properties cast everything to `float` by default and you can override this by passing a keyword parameter 'cast='. This is achieved because the property `fget` calls `values` method in the adapter and it passes `cast` parameter to it. A bit weird for my taste. Long story to say that there's no need to add extra code to cast to `float`.

Sorry, I have not seen this cast in `values` function. I have removed every float cast in properties

msmttchr (approved):
Looks good to me.
Many thanks for your high quality contribution

PR #651, opened 2022-07-18
HP legacy devices

LongnoseRob:
This PR is the contiuation of #549

- creation of a class `HPLegacyInstrument` based on `pymeasure.instrument`
   -  rebasing of multimeter **HP3478A** to `HPLegacyInstrument`
   -  addiing of system voltmeter **HP3437A** based on `HPLegacyInstrument`

- adding system power supplies **HP6632A/6633A/6634A** (single channel 100W class power supplies)

I have some verification code which I could transform into some examples, might add this to this PR later.


pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,218 @@
...
+        """Method to read the status bytes from the instrument
+        :return current_status: a byte array representing the instrument status
+        :rtype current_status: bytes
+        """
+        self.write("B")
+        current_status = self.adapter.read_bytes(self.status_bytes_count)
+        return current_status
+
+    @staticmethod
+    def decode_status(self, status_bytes, field=None):
```
msmttchr, line 93: This is a bit weird: a static method requiring the instance parameter.
I would say that you can change it in a regular method

msmttchr (changes requested):
@LongnoseRob, please find initial review of 1 file for your comments.
Thanks

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,218 @@
...
+        bcd = int.from_bytes(bcd, "big")
+        place, decimal = 1, 0
+        while bcd > 0:
+            nibble = bcd & 0xF
+            decimal += nibble * place
+            bcd >>= 4
+            place *= 10
+        return decimal
+
+    # facotry components
```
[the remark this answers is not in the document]
LongnoseRob: Thank you for pointing this out.

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,215 @@
...
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+c_uint8 = ctypes.c_uint8
+c_uint16 = ctypes.c_uint16
+c_uint32 = ctypes.c_uint32
+
+
+class HPLegacyInstrument(Instrument):
```
msmttchr, line 38: @LongnoseRob, I think your idea of putting the common part in this file is great. If you don't mind, let me share few idea to increase code reuse. I think that status register is in some way pivotal in this instruments, as such I would like to propose the following:
- Add a `StatusBitsBase` class that represent generic status register and associated methods
- `StatusBitsBase` includes a mechanism to decode and encode bitfields
- In the instrument file, create a proper `Status` class derived from `StatusBitsBase`, by adding all the relevant fields and the functions to encode/decode.
-  Status information can be accessed as property, e.g. `inst.status.Format` (`inst `is an instance of an instrument)

See below some sketch implementation.

```
class StatusBitsBase(ctypes.BigEndianStructure):
    """
    A bitfield structure containing the assignments for the status decoding
    """
    _pack_ = 1
    _get_process_ = {}
    _set_process_ = {}

    @staticmethod
    def _convert_from_bcd(bcd):
        """Converts a bcd value to a decimal value

        :param value: The value to unpack from bcd
        :returns: The number in decimal form
        """
        place, decimal = 1, 0
        while bcd > 0:
            nibble = bcd & 0xF
            decimal += nibble * place
            bcd >>= 4
            place *= 10
        return decimal

    def __str__(self):
        """
        Returns a pretty formatted string showing the status of the instrument

        """
        ret_str = ""
        for field in self._fields_:
            ret_str = ret_str + f"{field[0]}: {hex(getattr(self, field[0]))}\n"

        return ret_str

    def __getattribute__(self, name):
        val = super().__getattribute__(name)
        if name == "fields":
            return val
        if name in self.fields():
            process = super().__getattribute__('_get_process_')
            if name in process:
                val = process[name](val)

        return val

    def __setattr__(self, name, value):
        val = super().__getattribute__(name)
        if name in self.fields():
            process = super().__getattribute__('_set_process_')
            if name in process:
                value = process[name](value)

        super().__setattr__(name, value)

    def fields(self):
        return [desc[0] for desc in super().__getattribute__('_fields_')]

class HPLegacyInstrument(Instrument):
    """
    Class for legacy HP instruments from the era before SPCI, based on `pymeasure.Instrument`

    """

    status_desc = StatusBitsBase # To be overriden by subclasses

    def __init__(self, adapter, name, **kwargs):
        super().__init__(
            adapter, name,
            includeSCPI=False,
            **kwargs,
        )

        self.status_bytes_count = ctypes.sizeof(self.status_desc)
        log.info(f"Initializing {self.name}")

.
.
.
    @property
    def status(self):
        """
        Returns an object representing the current status of the unit.

        """
        
        self.write("B")
        reply = self.adapter.read_bytes(self.status_bytes_count)
        return self.status_desc.from_buffer(reply)

```

in the instrument file, you could do something like this:
```
class Status(StatusBitsBase):
    """
    A bitfield structure containing the assignments for the status decoding
    """
    _pack_ = 1
    _fields_ = [
        # Byte 0: Function, Range and Number of Digits
        ("Format", c_uint8, 1),  # Bit 7
        ("SRQ", c_uint8, 3),  # bit 4..6
        ("trigger", c_uint8, 2),  # bit 2..3
        ("range", c_uint8, 2),  # bit 0..1
        # Byte 1 & 2:
        ("Number", c_uint16, 16),
        # Byte 1:
        # ("NRDGS_MSD", c_uint8, 4),
        # ("NRDGS_2SD", c_uint8, 4),
        # Byte 2:
        # ("NRDGS_3SD", c_uint8, 4),
        # ("NRDGS_LSD", c_uint8, 4),
        ("not_used", c_uint8, 4),
        ("Delay", c_uint32, 28),
        # Byte 3:
        # ("Not_Used", c_uint8, 4),
        # ("Delay_MSD", c_uint8, 4),
        # Byte 4:
        # ("Delay_2SD", c_uint8, 4),
        # ("Delay_3SD", c_uint8, 4),
        # Byte 5:
        # ("Delay_4SD", c_uint8, 4),
        # ("Delay_5SD", c_uint8, 4),
        # Byte 6:
        # ("Delay_6SD", c_uint8, 4),
        # ("Delay_LSD", c_uint8, 4),
        ]

    _get_porocess_ = {
        "SRQ" : None, # TBD
        "Number" : StatusBitsBase._convert_from_bcd,
        "Delay" : StatusBitsBase._convert_from_bcd
        }

    _set_porocess_ = {
        # TBD
        }
.
.
.
class HP3437A(HPLegacyInstrument):
    """Represents the Hewlett Packard 3737A system voltmeter
    and provides a high-level interface for interacting
    with the instrument.
    """

    status_desc = Status

    def __init__(self, resourceName, **kwargs):
        super().__init__(
            resourceName,
            "Hewlett-Packard HP3437A",
            **kwargs,
        )
        self.packed_bits = Packed_bits # Is this needed ?
        self.packed_bytes = self.bytefield_factory(PACKED_BYTES) # Is this needed ?
        self.packed_data = self.union_factory(self.packed_bytes, self.packed_bits) # Is this needed ?
        log.info("Initialized HP3437A")


```

I think also the `packed_*` data can be avoided in some way, but I am not sure.
For fields like `SRQ`, take into account that bitfields can be nested.

Sorry for the long comment, please let me know your opinion.

LongnoseRob: @msmttchr Thank you very much for the constructive feedback.
I will try to implement the idea and hopefully we will see it working in the next days..

LongnoseRob: @msmttchr  have started to implement the changes as proposed:
_hplegacydevices.py_:
```class HPLegacyInstrument(Instrument):
    """
    Class for legacy HP instruments from the era before SPCI, based on `pymeasure.Instrument`

    """
    status_desc = StatusBitsBase  # To be overriden by subclasses

    def __init__(self, adapter, name, status_bytes, status_bitfield, **kwargs):
        super().__init__(
            adapter, name,
            includeSCPI=False,
            **kwargs,
        )

        self.status_bytes_count = ctypes.sizeof(self.status_desc)
        self.status_bits = self.status_desc
        self.status_bytes = self.bytefield_factory(status_bytes)
        self.status_union = self.union_factory(self.status_bytes, self.status_bits)

        log.info(f"Initializing {self.name}")

    def write(self, command):
        if command == "B":
            self.adapter.connection.write("B", termination="")
        else:
            # super().write(command)
            self.adapter.connection.write(command)

        if command == "B":
            self.write(command)
            return self.adapter.connection.read_raw(**kwargs)
        else:
            return super().values(command, **kwargs)

    @property
    def status(self):
        """
        Returns an object representing the current status of the unit.

        """
        # current_status = self.decode_status(self.fetch_status())
        # return current_status
        self.write("B")
        reply = self.adapter.read_bytes(self.status_bytes_count)
        return self.status_desc.from_buffer(reply).
.
.
.

```
for easier testing I have started working on the HP3478A (which is fan-less, so easier debugging in the living room)
_exerpt from HP3478a.py:_
```
STATUS_BYTES = 5


class Status(StatusBitsBase):
    """
    Support-Class with the bit assignments for the 5 status byte of the HP3478A
    """

    _fields_ = [
        # Byte 1: Function, Range and Number of Digits
        ("digits", c_uint8, 2),  # bit 0..1
        ("range", c_uint8, 3),  # bit 2..4
        ("function", c_uint8, 3),  # bit 5..7

        # Byte 2: Status Bits
        ("int_trig", c_uint8, 1),
        ("auto_range", c_uint8, 1),
        ("auto_zero", c_uint8, 1),
        ("fifty_hz", c_uint8, 1),
        ("front_rear", c_uint8, 1),
        ("cal_enable", c_uint8, 1),
        ("ext_trig", c_uint8, 1),
        ("res1", c_uint8, 1),

        # Byte 3: Serial Poll Mask (SRQ)
        ("SRQ_data_rdy", c_uint8, 1),
        ("res2", c_uint8, 1),
        ("SRQ_syntax_error", c_uint8, 1),
        ("SRQ_internal_error", c_uint8, 1),
        ("SRQ_front_panel", c_uint8, 1),
        ("SRQ_cal_error", c_uint8, 1),
        ("res3", c_uint8, 1),
        ("SRQ_PON", c_uint8, 1),

        # Byte 4: Error Information
        ("ERR_cal", c_uint8, 1),
        ("ERR_RAM", c_uint8, 1),
        ("ERR_ROM", c_uint8, 1),
        ("ERR_slope", c_uint8, 1),
        ("ERR_AD", c_uint8, 1),
        ("ERR_AD_Link", c_uint8, 1),
        ("res4", c_uint8, 1),
        ("res5", c_uint8, 1),

        # Byte 5: DAC Value
        ("DAC_value", c_uint8, 8),
    ]

    def _decode_trigger(self):
        i_trig = self.int_trig
        e_trig = self.ext_trig
        if i_trig == 0:
            if e_trig == 0:
                trigger_mode = "hold"
            else:
                trigger_mode = "external"
        else:
            trigger_mode = "internal"
        return trigger_mode

    _get_process_ = {
         "SRQ": None,  # TBD
         "Trigger": _decode_trigger,
         "Delay": StatusBitsBase._convert_from_bcd
         }

    # _set_process_ = {
    #     # TBD
    #     }


class HP3478A(HPLegacyInstrument):
    """ Represents the Hewlett Packard 3478A 5 1/2 digit multimeter
    and provides a high-level interface for interacting
    with the instrument.
    """
    status_desc = Status

    def __init__(self, resourceName, **kwargs):
        kwargs.setdefault('read_termination', '\r\n')
        kwargs.setdefault('send_end', True)
        super().__init__(
            resourceName,
            "Hewlett-Packard HP3478A",
            status_bytes=STATUS_BYTES,
            status_bitfield=self.status_desc,
            **kwargs,
        )
.
.
.

```

However when I now try to call `meter.status.Trigger` I get the following error:
```
In [5]: meter.status.Trigger

  Input In [5] in <cell line: 1>
    meter.status.Trigger

  File ~/builds/pymeasure/pymeasure/instruments/instrument.py:209 in __getattribute__
    return super().__getattribute__(name)

  File ~/builds/pymeasure/pymeasure/instruments/hp/hplegacyinstrument.py:159 in status
    return self.status_desc.from_buffer(reply)

TypeError: underlying buffer is not writable
```
Is this TypeError comming from the fact that I commented out the `_set_command`above?

msmttchr: No, I don't think so. I am sorry I was not able to test my sketch code, but looking at the documentation the error should be fixed by updating the `status` property as follow:
```
    @property
    def status(self):
        """
        Returns an object representing the current status of the unit.

        """
        # current_status = self.decode_status(self.fetch_status())
        # return current_status
        self.write("B")
        reply = bytearray(self.adapter.read_bytes(self.status_bytes_count))
        return self.status_desc.from_buffer(reply).

```
On the other hand, after this fix you should be able to access all the bitfields defined in `Status` class (with optional decoding function), but `Trigger` is not one of them.
The way you have defined the decode_trigger function leads me to think that it should be a property similar to this (I changed `Trigger` to lower case to respect code style conventions):
```
  @property
  def trigger(self):
        status = self.status
        i_trig = status.int_trig
        e_trig = status.ext_trig
        if i_trig == 0:
            if e_trig == 0:
                trigger_mode = "hold"
            else:
                trigger_mode = "external"
        else:
            trigger_mode = "internal"
        return trigger_mode
 
```

LongnoseRob: Hello,

The addition of the `bytearray(..)` helped, now the status field are accessible.

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,218 @@
...
+            return val
+        if name in self.fields():
+            process = super().__getattribute__('_get_process_')
+            if name in process:
+                val = process[name](val)
+
+        return val
+
+    def fields(self):
+        return [desc[0] for desc in super().__getattribute__('_fields_')]
```
msmttchr, line 151:
```suggestion
    def __getattribute__(self, name):
        val = super().__getattribute__(name)
        if name == "fields":
            return val
        if name in self.fields():
            process = super().__getattribute__('_get_process_')
            if name in process:
                val = process[name](val)

        return val

    def fields(self):
        return [desc[0] for desc in super().__getattribute__('_fields_')]
```
I think this code is not needed

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,218 @@
...
+        if self.range == 3:
+            cur_range = 1.0
+
+        signbit = 1
+        if self.sign_bit == 0:
+            signbit = -1
+
+        return (
+            cur_range * signbit * (
+                self.MSD + float(self.SSD) / 10 + float(self.TSD) / 100 + float(self.LSD) / 1000
```
msmttchr, line 135:
```suggestion
                self.MSD + self.SSD / 10 + self.TSD / 100 + self.LSD / 1000
```
I think the conversion to float is not needed. Right ?

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,218 @@
...
+        self.status_bits = self.status_desc
+
+        log.info(f"Initializing {self.name}")
+
+    def write(self, command):
+        if command == "B":
+            self.adapter.connection.write("B", termination="")
+        else:
+            # super().write(command)
+            self.adapter.connection.write(command)
```
msmttchr, line 178: I am not sure why `super` is commented out. To me using `super  `is a preferable alternative, but I am not sure if there is any issue here.

msmttchr (changes requested):
@LongnoseRob, please find my second round of comments.
Thanks for your patience.

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,358 @@
...
+            cur_range = math.nan
+        # 1 indicates 0.1V range
+        if r == 1:
+            cur_range = 0.1
+        # 2 indicates 10V range
+        if r == 2:
+            cur_range = 10.0
+        # 3 indicates 1V range
+        if r == 3:
+            cur_range = 1.0
```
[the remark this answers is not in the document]
LongnoseRob: very elegant solution, I like that proposal

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,358 @@
...
+    @staticmethod
+    def _decode_trigger(t):
+        """Method to decode trigger mode
+
+        :param status_bytes: list of bytes to be decoded
+        :return trigger_mode: string with the current trigger mode
+        :rtype trigger_mode: str
+
+        """
+        if t == 0:
```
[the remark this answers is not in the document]
LongnoseRob: yes, the proposed map dictionary is the way better solution

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,358 @@
...
+        """
+        Returns a pretty formatted string showing the status of the instrument
+
+        """
+        ret_str = ""
+        for field in self._fields_:
+            ret_str = ret_str + f"{field[0]}: {getattr(self, field[0])}\n"
+
+        return ret_str
+
```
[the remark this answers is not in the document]
LongnoseRob: yes that was an unneeded double entry

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,218 @@
...
+c_uint32 = ctypes.c_uint32
+
+
+class StatusBitsBase(ctypes.BigEndianStructure):
+    """
+    A bitfield structure containing the assignments for the status decoding
+    """
+    _pack_ = 1
+    _get_process_ = {}
+    _set_process_ = {}
```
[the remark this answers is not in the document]
LongnoseRob: yes, a set_process is currently not needed,
could still be added later if actually required

pymeasure/instruments/hp/hplegacyinstrument.py
LongnoseRob: yes that was not needed

pymeasure/instruments/hp/hplegacyinstrument.py
LongnoseRob: changed where possible to the super call.
was probably a left over from earlier

pymeasure/instruments/hp/hpsystempsu.py
```diff
@@ -0,0 +1,341 @@
...
+        CCCV_ZERO_HIGH = 18
+        CCCV_ZERO_LOW = 19
+        CV_REF_HIGH = 20
+        CV_REF_LOW = 21
+        CC_REF_HIGH = 22
+        CC_REF_LOW = 23
+        DAC_FAIL = 24
+        EEPROM_CHKSUM = 51
+
+    class SRQ (IntFlag):
```
msmttchr, line 145: I think this class is not used. Right?

pymeasure/instruments/hp/hp3437A.py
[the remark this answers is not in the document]
msmttchr: @LongnoseRob, indeed you are right.
I think that difference is too tiny that the two methods can be merged in the parent class.
Some proposals (alternatives):

1. Check that field value is an integer and in such case convert to hex
2. Or just use string representation of every field without converting to hex.

I believe in such way code will be more maintenable in the future.
Make sense for you?

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,198 @@
...
+    def fields(self):
+        return [desc[0] for desc in super().__getattribute__('_fields_')]
+
+
+class PackedBitsBase(ctypes.BigEndianStructure):
+    """
+    A bitfield structure containing the assignments for the status decoding
+    """
+    _pack_ = 1
+    _get_process_ = {}
```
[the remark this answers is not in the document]
LongnoseRob: agreed, not needed anymore

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,198 @@
...
+            if name in process:
+                val = process[name](val)
+
+        return val
+
+    def fields(self):
+        return [desc[0] for desc in super().__getattribute__('_fields_')]
+
+
+class PackedBitsBase(ctypes.BigEndianStructure):
```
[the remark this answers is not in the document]
LongnoseRob: Also agreed, I will put it in the file hp34737a as it is instrument specific (at the moment)

pymeasure/instruments/hp/hpsystempsu.py
LongnoseRob: correct, removed

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,166 @@
...
+        """Converts a decimal value to a bcd value
+
+        :param value: The decimal value to to pack into bcd
+        :returns: The number in bcd form
+        """
+        place, bcd = 0, 0
+        while decimal > 0:
+            nibble = decimal % 10
+            bcd += nibble << place
+            decimal /= 10
```
msmttchr, line 73:
```suggestion
            decimal //= 10
```
I think you need integer division here, right?

pymeasure/instruments/hp/hpsystempsu.py
```diff
@@ -0,0 +1,330 @@
...
+        return output_status
+
+    def reset_OVP_OCP(self):
+        """
+        Resets Overvoltage and Overcurrent protections
+
+        """
+        self.write("RST")
+
+    rom_Version = HPLegacyInstrument.measurement(
```
msmttchr, line 280: I am not a big fan (and expert) of code style, but sometime improves code readability.
`pymeasure` follows https://www.python.org/dev/peps/pep-0008/

See also https://peps.python.org/pep-0008/#function-and-variable-names

In this case, I think it would be better to use `rom_version`.

Could You please check also other names in this module (take into account that exceptions are allowed when necessary) ?

pymeasure/instruments/hp/hpsystempsu.py
```diff
@@ -0,0 +1,330 @@
...
+        output_status = bool(self.status.CV or self.status.CCpos or
+                             self.status.CCneg or self.status.Unregulated)
+        return output_status
+
+    def reset_OVP_OCP(self):
+        """
+        Resets Overvoltage and Overcurrent protections
+
+        """
+        self.write("RST")
```
msmttchr, line 278: This looks to me a general reset of the instrument. Is this correct ?

pymeasure/instruments/hp/hpsystempsu.py
```diff
@@ -0,0 +1,330 @@
...
+class HP6633A(HP6632A):
+    """ Represents the Hewlett Packard 6633A system power supply
+    and provides a high-level interface for interacting
+    with the instrument.
+    """
+
+    name = "Hewlett Packard HP6633A"
+    current_values = [0, limits["HP6633A"]["Cur_lim"]]
+    OVP_values = [0, limits["HP6633A"]["OVP_lim"]]
+    voltage_values = [0, limits["HP6633A"]["Volt_lim"]]
```
msmttchr, line 318: In order to have this working, you need to declare the corresponding property with parameter `dynamic=True`.
Same comment for class `HP6634A`

pymeasure/instruments/hp/hpsystempsu.py
```diff
@@ -0,0 +1,330 @@
...
+        MASK_PGM_ERR = 46
+        MULT_CSAVE_ERR = 50
+        EEPROM_CJLSUM_ERR = 51
+        CALMODE_DISABLED = 52
+        CAL_CHNL_ERR = 53
+        CAL_FS_ERR = 54
+        CAL_OFFSET = 55
+        CAL_JMP_ERR = 59
+
+    class ST_ERRORS(Enum):
```
msmttchr, line 119: This class is not used anywhere. From comment it seems that it should be used in method `check_selftest_errors`

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,166 @@
...
+
+        """
+        self.adapter.connection.assert_trigger()
+
+    def reset(self):
+        """
+        Initatiates a reset (like a power-on reset) of the HP3478A
+
+        """
+        self.adapter.connection.clear()
```
msmttchr, line 157: Is this enough to start a reset as described in the docstring ?

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    TRIGGERS = {
+        "internal": "T1",
+        "external": "T2",
+        "hold": "T3",
+        "manual": "T3",
+    }
+
+    class SRQ(IntFlag):
```
msmttchr, line 200: I think this class is not used anywhere. Right ?
Take into account that if you need to further split the field `SRQ`, you could use nested `ctypes` enums.

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,372 @@
...
+
+        (This function also takes care of unpacking the data if required)
+
+        :return data: np.array containing the data
+        """
+        # Adjusting the timeout to match the number of counts and the delay
+
+        current_timeout = self.adapter.connection.timeout
+        time_needed = self.number_readings * self.delay
+        new_timeout = time_needed * 3 * 1000  # safety factor 3
```
msmttchr, line 231:
```suggestion
        new_timeout = min(1e6, time_needed * 3 * 1000)  # safety factor 3
```

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,372 @@
...
+        if new_timeout > current_timeout:
+            if new_timeout >= 1e6:
+                # Disables timeout if measurement would take more then 1000 sec
+                new_timeout = 1000000
```
msmttchr, line 235:
```suggestion
```

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,372 @@
...
+            processed_data = []
+            read_data_length = int(len(read_data) / 2)
+            for i in range(0, read_data_length):
+                _from = i * 2
+                _to = _from + 2
+                processed_data.append(self._unpack_data(read_data[_from:_to]))
+            self.adapter.connection.timeout = current_timeout
+            return np.array(processed_data)
+        self.adapter.connection.timeout = current_timeout
+        return np.array(read_data[:-2].decode("ASCII").split(","), dtype=float)
```
msmttchr, line 251:
```suggestion
        if self.talk_ascii:
            return_value = np.array(read_data[:-2].decode("ASCII").split(","),
                                    dtype=float)
        else:
            processed_data = []
            for i in range(0, len(read_data), 2):
                processed_data.append(self._unpack_data(read_data[i:i+2]))
            return_value = np.array(processed_data)
        self.adapter.connection.timeout = current_timeout
        return return_value

```

pymeasure/instruments/hp/hp3437A.py
```diff
@@ -0,0 +1,372 @@
...
+    def talk_ascii(self):
+        """
+        A boolean property, True if the instrument is set to ASCII-based communication.
+        This property can be set.
+        """
+        return bool(self.status.Format)
+
+    @talk_ascii.setter
+    def talk_ascii(self, value):
+        if value is True:
```
msmttchr, line 272:
```suggestion
        if value:
```

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -37,126 +37,68 @@
...
-    Support-Class for the 5 status byte of the HP3478A
-    """
-    _fields_ = [
-        ("byte1", c_uint8),
-        ("byte2", c_uint8),
-        ("byte3", c_uint8),
-        ("byte4", c_uint8),
-        ("byte5", c_uint8)
-    ]
+# STATUS_BYTES = 5
```
msmttchr, line 40:
```suggestion
```

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -37,126 +37,68 @@
...
 
     def __init__(self, resourceName, **kwargs):
         kwargs.setdefault('read_termination', '\r\n')
         kwargs.setdefault('send_end', True)
         super().__init__(
             resourceName,
             "Hewlett-Packard HP3478A",
-            includeSCPI=False,
-            **kwargs
+            status_bitfield=self.status_desc,
```
msmttchr, line 100:
```suggestion
```
This line is not needed, since parameter `status_bitfield` has been replaced by class variable `status_desc`

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -552,13 +412,12 @@ def SRQ_mask(self):
         =========  ==========================
 
         """
-        mask = self.decode_status(self.get_status(), "SRQ")
-        return mask
+        mask = self.status.SRQ
+        return self.SRQ(mask)
```
msmttchr, line 416: I think, you can have the same behavior using a nested `ctypes` bitfield class, but I haven't tried it.

msmttchr (changes requested):
@LongnoseRob, last round of feedback (I hope :) ).
Please have a look also at the conversation https://github.com/pymeasure/pymeasure/pull/651#discussion_r932902746

pymeasure/instruments/hp/hpsystempsu.py
LongnoseRob: agreed, will change to `rom_version`

pymeasure/instruments/hp/hpsystempsu.py
LongnoseRob: Yes you are correct, that was a probably a copy& paste with some missing edit.
changed the call in `check_selftest_errors()` to use the `ST_ERRORS` class

pymeasure/instruments/hp/hpsystempsu.py
LongnoseRob: Thank you for the reminder, added a `dymaic=True,` to the tree properties

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: Sorry that was a lfetover, which I did not clean up properly

pymeasure/instruments/hp/hpsystempsu.py
LongnoseRob: It might look like this, 
but on this instrument it only resets the over-voltage-protection & over-current-protection.
Please see also paragraph 6-13, pdf page 29 of the [manual](https://userequip.com/files/specs/1253/HP%206632A,%2033A,%2034A%20Operating.pdf)

pymeasure/instruments/hp/hplegacyinstrument.py
LongnoseRob: Yes it is,

Please see  pdf page 43 of the [manual](https://xdevs.com/doc/HP_Agilent_Keysight/HP%203478A%20Operations.pdf).

The HP3437 & HP663xA react in the same way

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: > I think, you can have the same behavior using a nested ctypes bitfield class, but I haven't tried it.
I have not tried it either, and I was not able to find a good example of nested ctyoes bitfields (yet)

Let me see if I can find some better way to simplify this.

pymeasure/instruments/hp/hp3437A.py
LongnoseRob: Thanks for the reminder, i was returning only the numeric value of the SRQ mask, now changed to the enum return, as in HP3478A

pymeasure/instruments/hp/hp3478A.py
msmttchr:
```
import ctypes
c_uint8 = ctypes.c_uint8

class F(ctypes.BigEndianStructure):
	_fields_ = [
        ("SRQ1", c_uint8, 1),
        ("SRQ2", c_uint8, 1),
     ]
class C(ctypes.BigEndianStructure):
	_fields_ = [
        ("byte1", c_uint8, 8),
        ("SRQ", F),
     ]
a=C.from_buffer(bytearray([0xff, 0x40]))
print(a.byte1)
print(a.SRQ.SRQ2)
print(a.SRQ) # This is not accessed as number
```
I have made this example, please note that SRQ field cannot be accessed as byte (unless a union is used)

LongnoseRob: Thank you for the example, I'll have a look at it during the week and hopefully have an update torards the end of the week

pymeasure/instruments/hp/hpsystempsu.py
bilderbuchi: I'm very glad that you drove the dynamic property implementation, @msmttchr ! This example shows how this enables really concise model variants without code duplication! :-)

pymeasure/instruments/hp/hp3478A.py
LongnoseRob: I think I found a good compromise,

pymeasure/instruments/hp/hplegacyinstrument.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class HPLegacyInstrument(Instrument):
+    """
+    Class for legacy HP instruments from the era before SPCI, based on `pymeasure.Instrument`
+
+    """
+    status_desc = StatusBitsBase  # To be overriden by subclasses
+
+    def __init__(self, adapter, name, status_bitfield, **kwargs):
```
msmttchr, line 110:
```suggestion
    def __init__(self, adapter, name, **kwargs):
```

This parameter is not used (I believe left from previous implementation)

pymeasure/instruments/hp/hpsystempsu.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+    status_desc = Status
+
+    def __init__(self, resourceName, **kwargs):
+        kwargs.setdefault('read_termination', '\r\n')
+        kwargs.setdefault('send_end', True)
+        super().__init__(
+            resourceName,
+            "Hewlett-Packard HP6632A",
+            status_bitfield=self.status_desc,
```
msmttchr, line 83:
```suggestion
```
This is not needed, right?

msmttchr (changes requested):
A couple of comments from my side.
Thanks

pymeasure/instruments/hp/hplegacyinstrument.py
LongnoseRob: yes that was an left-over, cleaned uo now..

msmttchr (approved):

msmttchr (changes requested):
@LongnoseRob, changes are ok for me.
Thanks a lot for your contribution.
There's a small detail that popped up recently in another PR. Could you please rename `resourceName` to `adapter` in the code ?

msmttchr (approved):

bilderbuchi:
I was not very involved here, if it looks good to you, go ahead with the merge. :-)

bilderbuchi:
@msmttchr @LongnoseRob this apparently broke master CI (https://github.com/pymeasure/pymeasure/commit/9ebc63d19c7a20dac2987b2bbb59f6d813df62c8), can you take a look maybe?

msmttchr:
I have checked quickly and apparently the problem is in the line
`cls(adapter=MagicMock())`.
The class `HPLegacyInstrument` requires also `name` parameter and so the test fails.

I am not sure which is best approach to solve this. I have seen that the test skips some instrument that are problematic, but possibly @bmoneke has more clear idea.
It looks to me a test issue more than a driver code issue.

bilderbuchi:
Yeah, probably we had some bad timing w.r.t master updating after the last CI run for this PR happened, I think this is how this slipped through the cracks. 
However, what this test checks (maybe it could have a more expressive name) is homogeneity of API across our instrument base. Contrary to most (all?) other instruments, `HPLegacyInstrument` requires a mandatory positional argument `name` when being constructed. So, this should be fixed afaict, most easily by making `name` a kwarg with a default value.

The skipped tests are for instruments that were not possible to test or are non-compliant for other reasons (with the intention to winnow that list down over time).

bilderbuchi:
I PRd a quick fix in #684 and will merge it if CI passes.

LongnoseRob:
@bilderbuchi thank you for the quick fix #684 
I did a quick run with my verification script on the HP3478A and I can report the change did not have any impact on the operation of the instrument.

bilderbuchi:
Great, thanks!

PR #655, opened 2022-07-26
Documentation overhaul

bmoneke:
I improved the documentation wording and added a few lines to the tests with an actual instrument.

With that, I think we can merge protocol-tests into the master.

### DIFF
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -709,12 +709,13 @@ In the above example, :code:`MultimeterA` and :code:`MultimeterB` use a differen
 Writing tests
 =============
 
-Tests are an important method for writing good code. We distinguish two groups of tests for instruments: Tests which verify the working of the code against expectation (device manual) only, and tests with a device connected.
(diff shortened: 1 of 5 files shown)


bmoneke:
Small changes to the expected protocol

- no name, because several instruments cannot handle a name given.
- handing kwargs to the instrument in order to allow more parameters at instantiation. Some manuals give examples for different device addresses, which are included in the communication. If the instrument stores the address given at the instantiation, expected_protocol can test these protocol examples.

I also added the test for a different device.

bilderbuchi:
:heart: thank you, that is very useful!
I have a couple minor tweaks, but I'll merge as-is and do those in the target branch.
Good point regarding kwargs, quite useful!

> no name, because several instruments cannot handle a name given.

If you have time, could you file an issue/PR on the instruments you found, they should follow the best practice of allowing the user to override `name`. We should also check if the documentation reflects that.

bmoneke:
> If you have time, could you file an issue/PR on the instruments you found, they should follow the best practice of allowing the user to override `name`. We should also check if the documentation reflects that.

Most instruments do not have a name argument, but give the `super.__init__(adapter, "my funny name")` the name as a string like in this example.
I'll make the necessary changes and do a single PR for all of them together.

PR #656, opened 2022-07-26
GUI: Enable spinbox functionality for IntergerParameter and FloatParameter

jarvas24:
I like to use spin boxes for my `IntegerParameter` and `FloatParameters`, which I've previously been implementing with the `ui_class` kwarg. This seems like a feature that could have general appeal, so I made this PR with a potential implementation:

The `IntegerParameter` and `FloatParameter` classes now accept a `step` kwarg. If provided, the corresponding GUI input will be a spinbox with step size = `step`. The current behavior is unchanged if `step` is not provided.

I also removed the unused `FloatInput` class. It seems to have been replaced by the `ScientificInput` class.

Please let me know if you'd like me to make any changes or add an example of this to the docs.

### DIFF
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -109,13 +109,24 @@ class IntegerInput(Input, QtGui.QSpinBox):
 
     def __init__(self, parameter, parent=None, **kwargs):
         super().__init__(parameter=parameter, parent=parent, **kwargs)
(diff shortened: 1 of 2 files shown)


msmttchr (approved):
Looks good to me. Nice addition and also cleanup of unused stuff.
I wonder whether @CasperSchippers has some time to double check.

PR #61, opened 2017-03-03
Updating conda package for PyQt in Python 3.4

cjermain:
Working to solve issue #60 to resolve building in Appveyor for Python 3.4.

PR #659, opened 2022-08-02
Harmonise instrument name definition pattern, consistently name the instrument connection argument "adapter"

### LINKED ISSUE #658: Many instruments miss name argument
Many instruments miss a name argument and pass their name directly to the super constructor.

bmoneke:
Fixes #658 and adjustes the documentation accordingly.

- [x] Verify that the whole adding_instruments documentation is up to date.
- [ ] Create a follow up Issue for changing all instruments. (correct the super invocation in Keysight N7776, LakeShore421), maybe together with the change to a "SCPI" mixin.

### DIFF
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -104,10 +104,10 @@ This is a minimal instrument definition:
     class Extreme5000(Instrument):
         """Control the imaginary Extreme 5000 instrument."""
 
-        def __init__(self, adapter, **kwargs):
+        def __init__(self, adapter, name="Extreme 5000", **kwargs):
             super().__init__(
                 adapter,
-                "Extreme 5000",
+                name,
                 **kwargs
             )
 
@@ -236,10 +236,10 @@ The simplest version, suitable when the instrument connection needs default sett
 
 .. code-block:: python
 
-    def __init__(self, adapter, **kwargs):
+    def __init__(self, adapter, name="Extreme 5000", **kwargs):
         super().__init__(
             adapter,
-            "Extreme 5000",
+            name,
             **kwargs
         )
 
@@ -248,10 +248,10 @@ This is suitable when the instrument has one type of interface, or any defaults 
 
 .. code-block:: python
 
-    def __init__(self, adapter, baud_rate=2400, **kwargs):
+    def __init__(self, adapter, name="Extreme 5000", baud_rate=2400, **kwargs):
         super().__init__(
             adapter,
-            "Extreme 5000",
+            name,
             baud_rate=baud_rate,
             **kwargs
         )
@@ -260,11 +260,11 @@ If you want to set defaults, but they don't need to be prominently exposed for r
 
 .. code-block:: python
 
-    def __init__(self, adapter, **kwargs):
+    def __init__(self, adapter, name="Extreme 5000", **kwargs):
         kwargs.setdefault('timeout', 1500)
         super().__init__(
             adapter,
-            "Extreme 5000",
+            name,
             **kwargs
         )
 
@@ -279,11 +279,11 @@ These then contain a *dictionary* with the settings specific to the respective i
 
 .. code-block:: python
 
-    def __init__(self, adapter, baud_rate=2400, **kwargs):
+    def __init__(self, adapter, name="Extreme 5000", baud_rate=2400, **kwargs):
         kwargs.setdefault('timeout', 1500)
         super().__init__(
             adapter,
-            "Extreme 5000",
+            name,
             gpib=dict(enable_repeat_addressing=False,
                       read_termination='\r'),
             asrl={'baud_rate': baud_rate,
@@ -1003,8 +1003,8 @@ Additionally, the device needs some time after it received a command, before it 
         :param address: The device address for the communication.
         :param query_delay: Wait time after writing and before reading in seconds.
         """
-        def __init__(self, adapter, address=0, query_delay=0.1):
-            super().__init__(adapter, "ExtremeCommunication")
+        def __init__(self, adapter, name="ExtremeCommunication", address=0, query_delay=0.1):
+            super().__init__(adapter, name)
             self.address = f"{address:03}"
             self.query_delay = query_delay
     
@@ -1052,8 +1052,8 @@ Some devices do not expect ASCII strings but raw bytes. In those cases, you can 
 
     class ExtremeBytes(Instrument):
         """Control the ExtremeBytes instrument with byte-based communication."""
-        def __init__(self, adapter):
-            super().__init__(adapter, "ExtremeBytes")
+        def __init__(self, adapter, name="ExtremeBytes"):
+            super().__init__(adapter, name)
     
         def write(self, command):
             """Write to the device according to the comma separated command.
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -211,13 +211,14 @@ As our signal values are often integers, the most appropriate enum types are :co
 :code:`IntFlags` are used by many instruments for the purpose just demonstrated.
 
 The status property could look like this:
+
 .. testcode::
 
     status = Instrument.measurement(
         "STB?", 
         """Measure the status of the device as enum.""",
         get_process=lambda v: ErrorCode(v), 
-   )
+    )
 
 .. _default_connection_settings:
 
--- a/tests/instruments/test_all_instruments.py
+++ b/tests/instruments/test_all_instruments.py
@@ -54,14 +54,97 @@ proper_adapters = ["IBeamSmart", "ANC300Controller"]
 init = ["ThorlabsPM100USB", "Keithley2700", "TC038", "Agilent34450A",
         "AWG401x_AWG", "AWG401x_AFG", "VARX", "HP8116A"]
 # Instruments which require more input arguments
-init.extend(["Instrument", "ATSBase"])
+name_required = ["ATSBase"]
 
 
 @pytest.mark.parametrize("cls", devices)
-def test_args(cls):
+def test_adapter_arg(cls):
     "Test that every instrument has adapter as their input argument"
     if cls.__name__ in proper_adapters:
         pytest.skip(f"{cls.__name__} does not accept an Adapter instance.")
-    elif cls.__name__ in init:
+    elif cls.__name__ in (*init, *name_required, "Instrument"):
         pytest.skip(f"{cls.__name__} requires communication in init.")
     cls(adapter=MagicMock())
+
+
+# Instruments which do not yet accept "name" argument
+nameless_instruments = [
+    "AdvantestR3767CG",
+    "Agilent33220A",
+    "Agilent33500",
+    "Agilent33521A",
+    "Agilent34410A",
+    "Agilent4156",
+    "Agilent8257D",
+    "Agilent8722ES",
+    "AgilentB1500",
+    "AgilentE4408B",
+    "AgilentE4980",
+    "Ametek7270",
+    "AMI430",
+    "DPSeriesMotorController",
+    "APSIN12G",
+    "AnritsuMG3692C",
+    "AnritsuMS9740A",
+    "Danfysik8500",
+    "SM7045D",
+    "EurotestHPP120256",
+    "Fluke7341",
+    "FWBell5080",
+    "ND287",
+    "HP33120A",
+    "HP3437A",
+    "HP34401A",
+    "HP3478A",
+    "HP6632A",
+    "HP6633A",
+    "HP6634A",
+    "HP8657B",
+    "Keithley2000",
+    "Keithley2306",
+    "Keithley2400",
+    "Keithley2450",
+    "Keithley2600",
+    "Keithley2750",
+    "Keithley6221",
+    "Keithley6517B",
+    "KeysightDSOX1102G",
+    "KeysightN5767A",
+    "KeysightN7776C",
+    "LakeShore331",
+    "LakeShore421",
+    "LakeShore425",
+    "ESP300",
+    "ParkerGV6",
+    "razorbillRP100",
+    "FSL",
+    "SFM",
+    "SPD1168X",
+    "SPD1305X",
+    "DSP7265",
+    "SG380",
+    "SR510",
+    "SR570",
+    "SR830",
+    "SR860",
+    "AFG3152C",
+    "TDS2000",
+    "ATS525",
+    "ATS545",
+    "ECO560",
+    "Thermotron3800",
+    "ThorlabsPro8000",
+    "Yokogawa7651",
+    "YokogawaGS200",
+]
+
+
+@pytest.mark.parametrize("cls", devices)
+def test_name_argument(cls):
+    "Test that every instrument accepts a name argument"
+    if cls.__name__ in (*proper_adapters, *init):
+        pytest.skip(f"{cls.__name__} cannot be tested without communication.")
+    elif cls.__name__ in nameless_instruments:
+        pytest.skip(f"{cls.__name__} does not accept a name argument.")
+    inst = cls(adapter=MagicMock(), name="Name_Test")
+    assert inst.name == "Name_Test"
--- a/tests/instruments/test_all_instruments.py
+++ b/tests/instruments/test_all_instruments.py
@@ -50,11 +50,17 @@ for manufacturer in dir(instruments):
 
 # Instruments unable to accept an Adapter instance.
 proper_adapters = ["IBeamSmart", "ANC300Controller"]
-# Instruments with communication in their __init__ which consequently fails.
-init = ["ThorlabsPM100USB", "Keithley2700", "TC038", "Agilent34450A",
-        "AWG401x_AWG", "AWG401x_AFG", "VARX", "HP8116A"]
-# Instruments which require more input arguments
-name_required = ["ATSBase"]
+# Instruments with communication in their __init__, which consequently fails.
+need_init_communication = [
+    "ThorlabsPM100USB",
+    "Keithley2700",
+    "TC038",
+    "Agilent34450A",
+    "AWG401x_AWG",
+    "AWG401x_AFG",
+    "VARX",
+    "HP8116A",
+]
 
 
 @pytest.mark.parametrize("cls", devices)
@@ -62,8 +68,10 @@ def test_adapter_arg(cls):
     "Test that every instrument has adapter as their input argument"
     if cls.__name__ in proper_adapters:
         pytest.skip(f"{cls.__name__} does not accept an Adapter instance.")
-    elif cls.__name__ in (*init, *name_required, "Instrument"):
+    elif cls.__name__ in need_init_communication:
         pytest.skip(f"{cls.__name__} requires communication in init.")
+    elif cls.__name__ == "Instrument":
+        pytest.skip("`Instrument` requires a `name` parameter.")
     cls(adapter=MagicMock())
 
 
@@ -142,9 +150,9 @@ nameless_instruments = [
 @pytest.mark.parametrize("cls", devices)
 def test_name_argument(cls):
     "Test that every instrument accepts a name argument"
-    if cls.__name__ in (*proper_adapters, *init):
+    if cls.__name__ in (*proper_adapters, *need_init_communication):
         pytest.skip(f"{cls.__name__} cannot be tested without communication.")
     elif cls.__name__ in nameless_instruments:
-        pytest.skip(f"{cls.__name__} does not accept a name argument.")
+        pytest.skip(f"{cls.__name__} does not accept a name argument yet.")
     inst = cls(adapter=MagicMock(), name="Name_Test")
     assert inst.name == "Name_Test"
--- a/tests/instruments/test_all_instruments.py
+++ b/tests/instruments/test_all_instruments.py
@@ -94,8 +94,10 @@ nameless_instruments = [
     "APSIN12G",
     "AnritsuMG3692C",
     "AnritsuMS9740A",
+    "BKPrecision9130B",
     "Danfysik8500",
-    "SM7045D",
+    "SM7045D",  # deltaelektronika
+    "Nxds",  # edwards
     "EurotestHPP120256",
     "Fluke7341",
     "FWBell5080",
@@ -124,11 +126,12 @@ nameless_instruments = [
     "LakeShore425",
     "ESP300",
     "ParkerGV6",
+    "CNT91",  # pendulum
     "razorbillRP100",
     "FSL",
     "SFM",
-    "SPD1168X",
-    "SPD1305X",
+    "SPD1168X",  # siglenttechnologies
+    "SPD1305X",  # siglenttechnologies
     "DSP7265",
     "SG380",
     "SR510",
@@ -136,10 +139,11 @@ nameless_instruments = [
     "SR830",
     "SR860",
     "AFG3152C",
-    "TDS2000",
-    "ATS525",
-    "ATS545",
-    "ECO560",
+    "TDS2000",  # tectronix
+    "ATS525",  # temptronic
+    "ATS545",  # temptronic
+    "ECO560",  # temptronic
+    "TexioPSW360L30",
     "Thermotron3800",
     "ThorlabsPro8000",
     "Yokogawa7651",
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@ Upcoming version
 New adapter and instrument mechanics
 ------------------------------------
 - Channel class added. Instrument.channels and Instrument.ch_X (X is any channel name) are reserved for channel implementations.
+- All instruments are required to accept a :code:`name` argument.
 
 Version 0.11.0 (2022-11-19)
 ===========================
--- a/pymeasure/instruments/__init__.py
+++ b/pymeasure/instruments/__init__.py
@@ -38,8 +38,10 @@ from . import anapico
 from . import andeenhagerling
 from . import anritsu
 from . import attocube
+from . import bkprecision
 from . import danfysik
 from . import deltaelektronika
+from . import edwards
 from . import eurotest
 from . import fluke
 from . import fwbell
@@ -53,6 +55,7 @@ from . import newport
 from . import ni
 from . import oxfordinstruments
 from . import parker
+from . import pendulum
 from . import razorbill
 from . import rohdeschwarz
 from . import siglenttechnologies
@@ -60,6 +63,7 @@ from . import signalrecovery
 from . import srs
 from . import tektronix
 from . import temptronic
+from . import texio
 from . import thermotron
 from . import thorlabs
 from . import toptica
--- a/pymeasure/instruments/temptronic/temptronic_base.py
+++ b/pymeasure/instruments/temptronic/temptronic_base.py
@@ -556,8 +556,8 @@ class ATSBase(Instrument):
         dynamic=True
     )
 
-    def __init__(self, adapter, **kwargs):
-        super().__init__(adapter, query_delay=0.05, **kwargs)
+    def __init__(self, adapter, name="ATSBase", **kwargs):
+        super().__init__(adapter, name=name, query_delay=0.05, **kwargs)
 
     def reset(self):
         """Reset (force) the System to the Operator screen.


bilderbuchi:
I'm actually not sure if we should always put the default name into the constructor signature, as it pollutes that (and displayed docstrings, presumably) with a pretty long entry that is not that central.
I'd prefer we use the `setdefault` to set a reasonable default name (i.e. the model identifier) in the constructor, and allow the user to easily override it with a setup-specific name if desired:
```python
def test_name_setting():
    class NamedInstrument(Instrument):
            kwargs.setdefault('name', 'Default Instrument')
            super().__init__(adapter, **kwargs)

    adapter = FakeAdapter()
    instr = NamedInstrument(adapter)
    assert instr.name == 'Default Instrument'
    # Let's use a user-specific name (e.g. to disambiguate multiple identical devices)
    instr2 = NamedInstrument(adapter, name='Vacuum Frobnicator')
    assert instr2.name == 'Vacuum Frobnicator'
```

What do you think?

bmoneke:
Normally I prefer to know, what values I might change and what the default is. setdefault hides the possibility to rename the device.
As changing the name is rarely used (almost no instrument allowed for that), I'm fine with hiding the renaming option.

One caveat though: The documentation does not hand the kwargs to the instrument, but to the adapter in case of a custom adapter!
We should take care of that as well.
We might even remove that example, as custom adapters are rarely necessary and should not be used, instead read/write.

Once I'm remodeling all instruments. Should we name the adapter consistently adapter? In some instruments it is called "resourceName", which is taken from the documentation.

bilderbuchi:
> One caveat though: The documentation does not hand the kwargs to the instrument, but to the adapter in case of a custom adapter!

I'm not sure what you mean?

> Once I'm remodeling all instruments. Should we name the adapter consistently adapter? In some instruments it is called "resourceName", which is taken from the documentation.

I think this comes from pyvisa nomenclature, where it's "resource name". I'm not sure. I guess in 95% of cases people supply a resource name string, so that would be a more fitting name. However, if people pass an `Adapter` instance, `resourceName` is a weird name. :-P -- `adapter` is delightfully ambiguous :unamused:
@msmttchr what do you think?

`adapter` is also consistent with `Instrument`. Renaming the Instrument posarg will probably break a lot of code I suspect.

msmttchr:
> > Once I'm remodeling all instruments. Should we name the adapter consistently adapter? In some instruments it is called "resourceName", which is taken from the documentation.
> 
> I think this comes from pyvisa nomenclature, where it's "resource name". I'm not sure. I guess in 95% of cases people supply a resource name string, so that would be a more fitting name. However, if people pass an `Adapter` instance, `resourceName` is a weird name. :-P -- `adapter` is delightfully ambiguous 😒 @msmttchr what do you think?
> 
> `adapter` is also consistent with `Instrument`.

I agree, in my opinion `adapter` is the correct name.

bmoneke:
I made the talked about changes (to adapter and using kwargs).

There are some doctest errors in the documentation though.

docs/dev/adding_instruments.rst
```diff
@@ -138,10 +138,10 @@ This is suitable when the instrument has one type of interface, or any defaults
 
 .. code-block:: python
 
-    def __init__(self, resourceName, baud_rate=2400, **kwargs):
+    def __init__(self, adaoter, baud_rate=2400, **kwargs):
```
bilderbuchi, line 141:
```suggestion
    def __init__(self, adapter, baud_rate=2400, **kwargs):
```

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -367,15 +367,14 @@ class AWG401x_base(Instrument):
     """AWG-401x base class"""
 
     def __init__(self, adapter, **kwargs):
+        kwargs.setdefault('name', "Active Technologies AWG-4014 1.2GS/s Arbitrary Waveform Generator")
```
bilderbuchi, line 370:
```suggestion
        kwargs.setdefault('name', 
                          "Active Technologies AWG-4014 1.2GS/s Arbitrary Waveform Generator")
```

pymeasure/instruments/agilent/agilent33521A.py
```diff
@@ -40,6 +40,7 @@ class Agilent33521A(Agilent33500):
     """
 
     def __init__(self, adapter, **kwargs):
+        kwargs.setdefault('name', "Represents the Agilent 33521A Function/Arbitrary Waveform Generator")
```
bilderbuchi, line 43:
```suggestion
        kwargs.setdefault('name', 
                          "Represents the Agilent 33521A Function/Arbitrary Waveform Generator")
```

docs/dev/adding_instruments.rst
```diff
@@ -196,15 +196,16 @@ If, however, you are unable to use the :py:class:`~pymeasure.adapters.VISAAdapte
 
 .. code-block:: python
 
-    def __init__(self, resourceName, baud_rate=2400, **kwargs):
+    def __init__(self, adapter, baud_rate=2400, **kwargs):
         kwargs.setdefault('timeout', 0.5)
         kwargs.setdefault('xonxoff', True)
-        adapter = SerialAdapter(resourceName, 
+        kwargs.setdefault('name', "Extreme 5000")
+        adapter = SerialAdapter(adapter,
```
bilderbuchi, line 203: Here it's a bit more complicated, as `SerialAdapter` does not accept a `name` kwarg.
This should work correctly:
```suggestion
        name = kwargs.pop('name', "Extreme 5000")
        adapter = SerialAdapter(adapter, 
```
If the user has given a `name`, remove it from the dict and assign it to the variable. 
Below, use that `name` as the 2nd posarg instead of the string literal.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -221,9 +221,9 @@ class ANC300Controller(Instrument):
 
     def __init__(self, host, axisnames, passwd, query_delay=0.05, **kwargs):
         kwargs['query_delay'] = query_delay
+        kwargs.setdefault('name', "attocube ANC300 Piezo Controller")
```
bilderbuchi, line 224: Same as in the SerialAdapter example above, the `AttocubeConsoleAdapter` does not accept  a `name` kwarg afaik.

docs/dev/adding_instruments.rst
bilderbuchi: I just realise: We should add a test to confirm correct behaviour for this usage pattern when the user passes in an `Adapter` instance instead of a string! I'm mad that I didn't think of that before. :-/

bilderbuchi (changes requested):

bilderbuchi:
> I'm actually not sure if we should always put the default name into the constructor signature, as it pollutes that (and displayed docstrings, presumably) with a pretty long entry that is not that central. I'd prefer we use the `setdefault` to set a reasonable default name (i.e. the model identifier) in the constructor, and allow the user to easily override it with a setup-specific name if desired:
> 
> ```python
> def test_name_setting():
>     class NamedInstrument(Instrument):
>         def __init__(self, adapter, **kwargs):
>             kwargs.setdefault('name', 'Default Instrument')
>             super().__init__(adapter, **kwargs)
> 
>     adapter = FakeAdapter()
>     instr = NamedInstrument(adapter)
>     assert instr.name == 'Default Instrument'
>     # Let's use a user-specific name (e.g. to disambiguate multiple identical devices)
>     instr2 = NamedInstrument(adapter, name='Vacuum Frobnicator')
>     assert instr2.name == 'Vacuum Frobnicator'
> ```
> 
> This way the signature of `NamedInstrument` stays clean, but we remain as flexible as possible. What do you think?

Please also add that test to the test suite, it's good to have tests that define best practices, so they constantly get checked/confirmed.

docs/dev/adding_instruments.rst
bilderbuchi: > I just realise: We should add a test to confirm correct behaviour for this usage pattern when the user passes in an `Adapter` instance instead of a string! I'm mad that I didn't think of that before. :-/

I hope this does not invalidate this whole usage pattern :-(

bilderbuchi:
I don't know what's up with the doctest, first I thought that it's because we swap in FakeInstrument in adding_instruments.rst L89 (to enable doctesting), but I can't seem to reproduce this locally, and the error message does not make sense to me.

docs/dev/adding_instruments.rst
bilderbuchi: > > I just realise: We should add a test to confirm correct behaviour for this usage pattern when the user passes in an `Adapter` instance instead of a string! I'm mad that I didn't think of that before. :-/
> 
> I hope this does not invalidate this whole usage pattern :-(

Some quick tests indicate that we're fine for this :sweat_smile: -- Python uses the kwarg `name` element as the posarg called `name`, and in the Instrument initialiser it does not show up in that method's kwargs any more.
It's only problematic if an adapter is constructed inside the specific instrument's initialiser, as here and in one instrument below.

pymeasure/instruments/advantest/advantestR3767CG.py
```diff
@@ -67,10 +67,10 @@ class AdvantestR3767CG(Instrument):
         "TRAC:DATA? FDAT1", """ Reads the Data array from trace 1 after formatting """
     )
 
-    def __init__(self, resourceName, **kwargs):
+    def __init__(self, adapter, **kwargs):
```
bilderbuchi, line 70: Since we are renaming posargs in a large number of instrument implementations, we should add an entry to a fresh section (upcoming release) in the CHANGES.rst where we make note of this. 
I suspect we could break user code if it refers to the posarg by name, but the name changed? (I.e. `myinstr = AdvantestR3767CG(resourceName='COM2')`.)

docs/dev/adding_instruments.rst
bmoneke: > I just realise: We should add a test to confirm correct behaviour for this usage pattern when the user passes in an `Adapter` instance instead of a string! I'm mad that I didn't think of that before. :-/
> > 
> > I hope this does not invalidate this whole usage pattern :-(
> 
> Some quick tests indicate that we're fine for this 😅 -- Python uses the kwarg `name` element as the posarg called `name`, and in the Instrument initialiser it does not show up in that method's kwargs any more.

Yes, I also had tested, that a kwarg element supplants a posarg.

> It's only problematic if an adapter is constructed inside the specific instrument's initialiser, as here and in the ANC300 below.

What problem do you see? If the developer provides their own adapter, they have to either make the adapter take the `name` kwarg (just adding **kwargs is enough) or give a subset of the kwargs (without name) to their adapter.

> I just realise: We should add a test to confirm correct behaviour for this usage pattern when the user passes in an `Adapter` instance instead of a string! I'm mad that I didn't think of that before. :-/

Here again the developer should take care, if they modify the adapter behaviour.

pymeasure/instruments/attocube/anc300.py
bmoneke: Ok.

bmoneke:
Thanks for your reviews. I implemented the changes.

I also implemented a test whether a named argument removes the corresponding entry from kwargs.

docs/dev/adding_instruments.rst
bilderbuchi: > What problem do you see? If the developer provides their own adapter, they have to either make the adapter take the name kwarg (just adding **kwargs is enough) or give a subset of the kwargs (without name) to their adapter.

Yes, only problematic for instrument driver developers that create custom adapters inside their instrument's init. The docs specify the correct behaviour, so I think we're good now.

bilderbuchi:
I think we're good now! Only the doctests are still broken (and a small linting issue remains). :crying_cat_face: 

Afterwards, we should check/update all open PRs for the new usage (`adapter` and `name` handling). I don't know how to meaningfully automate those checks.

docs/dev/adding_instruments.rst
```diff
@@ -96,10 +96,10 @@ This is a minimal instrument definition:
         """ Represents the imaginary Extreme 5000 instrument.
         """
 
-        def __init__(self, resourceName, **kwargs):
+        def __init__(self, adapter, **kwargs):
+            kwargs.setdefault('name', "Extreme 5000")
             super().__init__(
-                resourceName,
-                "Extreme 5000",
+                adapter
```
bilderbuchi, line 102: Gotcha! :D
```suggestion
                adapter,
```

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -367,15 +367,15 @@ class AWG401x_base(Instrument):
     """AWG-401x base class"""
 
     def __init__(self, adapter, **kwargs):
+        kwargs.setdefault('name',
```
bilderbuchi, line 370:
```suggestion
        kwargs.setdefault('name',
```

pymeasure/instruments/agilent/agilent33521A.py
```diff
@@ -40,6 +40,8 @@ class Agilent33521A(Agilent33500):
     """
 
     def __init__(self, adapter, **kwargs):
+        kwargs.setdefault('name',
```
bilderbuchi, line 43:
```suggestion
        kwargs.setdefault('name',
```

bilderbuchi (changes requested):

bmoneke:
> Afterwards, we should check/update all open PRs for the new usage (`adapter` and `name` handling). I don't know how to meaningfully automate those checks.

Testing with something like
```suggestion
def test_args():
    adapter = FakeAdapter()
    instr = NewInstrument(adapter=adapter, **{'name': "test successful"})
    assert NewInstrument.name == "test successful"
```
This test verifies, whether the first argname is adapter, because if it is not adapter, it won't receive a value and will raise an error.
It also verifies, that the name argument is present somehow.

bilderbuchi:
> This test verifies, whether the first argname is adapter, because if it is not adapter, it won't receive a value and will raise an error.
> It also verifies, that the name argument is present somehow.

Sorry, I was unclear. Yeah such a test tests what we need, but it does not do that (dynamically) for any instruments changed in a PR. Adding this same test to all instruments individually seems wasteful. I think we could enumerate all instruments, and feed those to that test, though! (or one test for the adapter, one for the name) :-)

bmoneke:
> Sorry, I was unclear. Yeah such a test tests what we need, but it does not do that (dynamically) for any instruments changed in a PR. Adding this same test to all instruments individually seems wasteful. I think we could enumerate all instruments, and feed those to that test, though! (or one test for the adapter, one for the name) :-)

I wrote a test, which searches for all instruments in pymeasure and does this test.
The problem is, however, that the inits of some devices fail, because they need some initial communication (adapter.connection.clear or setting/getting something etc.) or because they require some obligatory arguments (port, baud_rate...), which we do not supply.
Right now we pass 58 and fail 26.

bilderbuchi:
I'd say push the test, if there are obvious fixes make them, if we need to add something to FakeAdapter (if only that some instrument has something to call) we can do that. Finally, assuming you store the found instruments in a list for parameterize, we can make a filter list where we can skip (`pytest.skipif` or mark xfail with strict mode) an instrument test if it's on the list. That way we have a filter list we can slowly reduce. What do you say?

bilderbuchi:
In general, that list of all instruments (module-level `ALL_INSTRUMENTS`?) could come in handy for further API tests like this. :-)

tests/instruments/test_all_instruments.py
```diff
@@ -0,0 +1,88 @@
...
+devices = []
+for manufacturer in dir(instruments):
+    if manufacturer.startswith("__"):
+        continue
+    manu = getattr(instruments, manufacturer)
+    for dev in dir(manu):
+        if dev.startswith("__") or dev[0].islower():
+            continue
+        d = getattr(manu, dev)
+        if isclass(d):
```
bilderbuchi, line 63: `isinstance(d, Instrument)`?

bilderbuchi, line 63: `issubclass(d, Instrument)`?

tests/instruments/test_all_instruments.py
```diff
@@ -0,0 +1,88 @@
...
+
+from inspect import isclass
+
+import pytest
+
+from pymeasure.adapters import FakeAdapter
+from pymeasure import instruments
+
+
+class Adapter(FakeAdapter):
```
bilderbuchi, line 34: Minor nitpick: please don't name this class `Adapter` as it's identical to our adapter base class. `TestAdapter`?
I also later realised, we don't care what the adapter does in this test, at all, I think, so we might as well use a `MagicMock` instead of defining our own skeleton class.

bilderbuchi:
Awesome addition with that test, we can work with that!

tests/instruments/test_all_instruments.py
bilderbuchi: this should also avoid the failing test, as `DAQMx` is not an Instrument subclass for some reason? I'll report that separately.

tests/instruments/test_all_instruments.py
bmoneke: A consequence of being tired and wanting to finish...

bilderbuchi: Hey, I understand you -- I still have 10 unread pymeasure notifications (all PR updates)... not today... :-)

tests/instruments/test_all_instruments.py
bmoneke: thanks for hinting at issubclass

docs/dev/adding_instruments.rst
```diff
@@ -196,15 +196,17 @@ If, however, you are unable to use the :py:class:`~pymeasure.adapters.VISAAdapte
 
 .. code-block:: python
 
-    def __init__(self, resourceName, baud_rate=2400, **kwargs):
+    def __init__(self, adapter, baud_rate=2400, **kwargs):
         kwargs.setdefault('timeout', 0.5)
         kwargs.setdefault('xonxoff', True)
-        adapter = SerialAdapter(resourceName, 
+        name = kwargs.pop('name', "Extreme 5000")
+        adapter = SerialAdapter(adapter,
```
msmttchr, line 203: @bmoneke, I am sorry for late comment, especially because it is not related to your changes.
I think this section is not so relevant and I have difficulty to find a matching use case.
In my understanding, the controller (PC) uses the `Adapter` to talk to instrument and there are two cases:
- User passes an address (string) and this is translated into a `VISAAdapter` instance 
- User pass it's own `Adapter` subclass instance that identifies the adapter being used.

An instrument driver developer should not make any assumption regarding the interface being used, since this is a user choice.

I would propose to remove this section if there is an agreement on this

bmoneke: Some instruments have their own adapter, which is specific to the Instrument (in most cases it should be possible to move these implementations to the instrument, that is what I try in #660 ).

Removing this section prevents people from writing their own adapters, which is good, in my opinion, so I'm with you removing this section.

msmttchr:
> 

I am not entirely clear what is the advantage, I think the name of the instrument is something fixed and defined by instrument driver writer. Why do we want to allow user to override it ?
If user wants to do this, he can still do something like `inst.name = "Something"`.
On the other hand, implementating this suggestion has triggered a lot of changes in all instruments making, in my opinion, the code more obscure

bmoneke:
> > 
> 
> I am not entirely clear what is the advantage, I think the name of the instrument is something fixed and defined by instrument driver writer. Why do we want to allow user to override it ? If user wants to do this, he can still do something like `inst.name = "Something"`. On the other hand, implementating this suggestion has triggered a lot of changes in all instruments making, in my opinion, the code more obscure

For a user yes, `instr.name = "xy"` works, but for subclassing, you can just hand the name to the parent class as if it were `Instrument` itself.

docs/dev/adding_instruments.rst
msmttchr: @bmoneke, I have to say that I am not against people using their own adapters. For example, if you have a serial instrument, you can use a `VISAAdapter` or `SerialAdapter`.
The first one is selected by user specifying a string address, but I can also do:
```
my_adapter = SerialAdapter (...)
inst = InstClass(adapter=my_adapter)
```
I just don't see the use case where the `Adapter` is imposed by the instrument driver writer.
I also agree 100% with your comment that instrument behaviour should not be encapsulated in a custom adapter

bmoneke: > Removing this section prevents people from writing their own adapters, which is good, in my opinion, so I'm with you removing this section.

I meant preventing the instrument authors to write specific adapters for their instruments and probably encapsulating these adapters in the instrument init. The reason, why I don't like this kind of adapter use (imposed by the instrument author), is exactly what you say, that you cannot use another adapter (e.g. our ProtocolTest) or Serial instead of pyvisa etc. in this case.

In #660 I try to remove all those special adapters enforced in `instrument.__init__`s in order to allow different (generally defined or personal) adapters to be used.

bilderbuchi: > > Removing this section prevents people from writing their own adapters, which is good, in my opinion, so I'm with you removing this section.
> 
> I meant preventing the instrument authors to write specific adapters for their instruments and probably encapsulating these adapters in the instrument init. The reason, why I don't like this kind of adapter use (imposed by the instrument author), is exactly what you say, that you cannot use another adapter (e.g. our ProtocolTest) or Serial instead of pyvisa etc. in this case.
> 
> In #660 I try to remove all those special adapters enforced in `instrument.__init__`s in order to allow different (generally defined or personal) adapters to be used.

:+1: on this. This is relevant when you need to force use of specific/special adapters in an instrument. Dealing with this (as Benedikt says, ideally removing this) should be dealt with in #660, where this is better in scope.

bilderbuchi:
> I am not entirely clear what is the advantage, I think the name of the instrument is something fixed and defined by instrument driver writer. Why do we want to allow user to override it ?

One use case for users giving custom names is when you have an experiment with several identical devices, e.g. several power supplies controlling coil current, bias voltage, excitation voltage (random example), then you would name the instances as such so that it's clearer/more convenient to see further down (e.g. in logging, GUI use, etc).
E.g. I have previously written code that uses the instrument name when naming a data file name or data column name.

The other use case is when subclassing, as already mentioned.

bmoneke:
With that, I think, we're good to go.

bilderbuchi:
> With that, I think, we're good to go.

Me too. 
@msmttchr, agree to merge this?

msmttchr:
> > With that, I think, we're good to go.
> 
> Me too. @msmttchr, agree to merge this?

Sorry a bit busy in this period, I understand the use cases, but I still feel the changes are too much for the benefit we have, essentially:
1) Using the `name` of the instrument to solve application problems, to me looks like a workaround, since instrument driver should be separated from application.
2) For inheritance, here I see only the problem when similar instruments are inherited from a common instrument and still you can use the approach of patching the name after calling `super().__init__(...)`

Probably, the way I understand the `name` attribute (the name identifiying the instrument) is that it should be more appropriately a class attribute and not an instance attribute (passed via `__init__` as it is today).
So, if we would go under this restructuring, I would probably suggest to change name into a class attribute.

Just sharing my opinion, but if you feel changes are still worth to do, please go ahead.

bilderbuchi:
Thank you for your input!

> since instrument driver should be separated from application.

Well, the driver is still separate from the application. The application uses the driver to create instrument instances (as you are aware), with specific properties like different adapter ports, but also different names for the instruments. We want to enable users to be expressive about their instruments.

Furthermore, if you collect all instruments of a setup in e.g. a list, you lose even the variable names to disambiguate between two different instruments of the same class, so then the `name` becomes even more important. Otherwise, you'd have to backwards-infer which instrument you're addressing from its `adapter` value.

> when similar instruments are inherited from a common instrument 

Indeed, and we want to encourage that approach to avoid code duplication. Patching the name afterwards feels "dirty" to me.

> the way I understand the name attribute (the name identifiying the instrument) is that it should be more appropriately a class attribute and not an instance attribute

Sorry, I have to disagree. If you want a class-attached means of identifying an instrument, you should use the class type imo. 
AFAIK (but have not investigated), in pymeasure the name was always attached to an instance.

Finally, the main reason why this PR triggers so many changes related to `name` is that its usage was inhomogenous before, also because we only recently gained the more flexible kwargs handling, and this PR makes usage homogeneous across pymeasure.

msmttchr:
> Indeed, and we want to encourage that approach to avoid code duplication. Patching the name attribute afterwards feels "dirty" to me (but is of course a legitimate approach in python, see for example pyvisa usage)
> 

I think this is a good objective, in my opinion.
I share that it's a bit "dirty", but having a class attribute is the proper way (of course, again, in my opinion)
> > the way I understand the name attribute (the name identifiying the instrument) is that it should be more appropriately a class attribute and not an instance attribute
> 
> Sorry, I have to disagree. If you want a class-attached means of identifying an instrument, you should use the class type imo. AFAIK (but have not investigated), in pymeasure the name was always attached to an instance.
>
I think this make sense and here we need to understand what was the original intent of instrument `name`. I am sorry, here I can give you only my opinion based on code review and my common sense.
If I make an analogy with cars:
- class can be `Fiat500` which could have a class attribute describing the car name, eg: "Fiat 500" (i.e. something that will never change with instances, and class type is not necessarily descriptive enough)
- I can have a class derived from `Fiat500`, eg. `Fiat500XL` and here I would redefine the class attribute `name`.
- `mycar = Fiat500(...)`, it is my car and it can have a nickname, but this will not change the car name and can be managed in different way, independently from driver. It will have also reg no to identify it.

Coming back to the lab instrument, it's a common need to log instruments used during an experiment, but in such case I would recommend to use the serial number retrieved with `id` property (at least for SCPI instruments).

I understand other people can have different interpretations and argue that if my interpretation is right, why then the instrument name is a runtime parameter. Here, I could say that the implementation can be not inline with intent, but this is totally arbitrary interpretation from my side :(

bmoneke:
Interesting ideas.

I don't know how often you do not know which instrument is which one (as you call them with different variable names). But for those cases Bilderbuchi talked about, the user can always add at runtime a parameter at runtime like `mycar.nickname = "Fred"` and for another one `myOtherCar.nickname = "Bob"`.

bilderbuchi:
So, you're rich and have two Fiat 500 XL, one red ("Il lampo rosso") and one white ("Innocenza"), and you want to take the red lightning out for a spin. When grabbing the correct car key, would you prefer the car keys to have nametags, or tags with the Vehicle Identification Number?

bilderbuchi:
And yes, I had a situation/setup where there were 3 (I think) power supplies for 3 different aspects of the experiment, I did not have individual variable names available (they were all in a list because their relevant properties were queried every time tick), and I used the `name` to great effect when storing those data traces/displaying data/etc.

bmoneke:
The current implementation does both, in the end: It gives more information about the current device than the class name itself (e.g. `Anc300` has the name `'Anc 300 power meter'`) and gives the user the possibility to name the device individually.

With that, I'm for merging and keeping `name` as an instance property.

bilderbuchi:
Side remark: class or instance property does not make much of a difference. If `Instrument.name` were a class property, as soon as we change it (e.g. in an initialiser) it becomes an instance attribute of that instance. So _that_ distinction, in my mind, seems not that significant.

msmttchr:
> The current implementation does both, in the end: It gives more information about the current device than the class name itself (e.g. `Anc300` has the name `'Anc 300 power meter'`) and gives the user the possibility to name the device individually.
> 
> With that, I'm for merging and keeping `name` as an instance property.

Indeed I agree with your statement, my original point is that:
1) We achieve this target making the code more obscure and difficult to read
2) I interpret `name` as a read only parameter that describe the instrument model for which the class is designed for, and it should not be changed by users

My code score:
```
# Best
class Agilent34410A(Instrument):
    """
    Represent the HP/Agilent/Keysight 34410A and related multimeters.
    Implemented measurements: voltage_dc, voltage_ac, current_dc, current_ac, resistance,
    resistance_4w
    """
    name = "HP/Agilent/Keysight 34410A Multimeter" # class attribute redefined from Instrument class where it could be set to "Generic Instrument", can be also redefined in subclasses of Agilent34410A
        super().__init__(
            adapter, **kwargs
        )

# Acceptable
class Agilent34410A(Instrument):
    """
    Represent the HP/Agilent/Keysight 34410A and related multimeters.
    Implemented measurements: voltage_dc, voltage_ac, current_dc, current_ac, resistance,
    resistance_4w
    """
        super().__init__(
            name = "HP/Agilent/Keysight 34410A Multimeter"
            adapter, **kwargs
        )

# Worst
class Agilent34410A(Instrument):
    """
    Represent the HP/Agilent/Keysight 34410A and related multimeters.
    Implemented measurements: voltage_dc, voltage_ac, current_dc, current_ac, resistance,
    resistance_4w
    """

        kwargs.setdefault('name', "HP/Agilent/Keysight 34410A Multimeter")
        super().__init__(
            adapter, **kwargs
        )
```

In all cases, when you have an instance of the instrument (let's call it `my_inst`), you can do (but this is just python):
- `my_inst.name = "Multimeter1" # Lost original instrument name, but still accessible with Agilent34410A.name in case 1.`
- `my_inst.nick_name = "Multmeter1" # New attribute in class instance, it could be dangerous since it can overwrite an attribute used by the instrument driver (unlikely though)`
- `my_inst.write = None # destroy instrument functionality`

The other point is that we could split features that are associated with the instrument model with features that are associated with the instrument instance, eg:
- `adapter` (instrument instance, since instrument address or adapter can change, but instrument stay the same)
- `name` (instrument class, name does not change)
- `includeSCPI` (instrument class, obviously if a particular instrument support SCPI is a feature of the instrument model, 
  not something that depends on particular instrument).

Given the above, I have to say that I am not really super-happy (obviously I could be wrong) to approve these changes, but I need to respect other people opinion (especially if they are working till late on the project for free :) ) and so feel free to merge them in the master.

bilderbuchi:
I think on the point of name intended to be fixed, we just have to disagree.

> making the code more obscure and difficult to read

Concerning this, is this opinion mainly coming from the fact that we use the `setdefault` method now (and you consider that one obscure)? 
Because, your "acceptable" and "worst" variants have the same length of the initialiser. I'll grant that it's a little less discoverable for an instrument implementor because `name` gets put into the kwargs, not directly into the super() signature. 

However, the setdefault approach has the advantage to remove (_by default_) `name` from the signature of the derived class! 
I thought that worth it because 
* the name _typically_ is not changed, so its presence in the signature adds clutter (e.g. in autocompletion)
* the approach preserves the _current ability_, if desired, to assign a custom name during instantiation.
* the instrument driver is (hopefully) more often _used_ than _written_, so clarity of external API seems relatively important. This is similar reasoning to having "simply" (as much as possible) structured kwargs for connection configuration, and hiding the logic complexity inside `Instrument`, where instrument contributors rarely tread.

One argument concerning the last point that can of course be made: is the API clearer by having name an acceptable element in kwargs, which needs to be documented separately (@bmoneke by the way, it should -- can you add name to the docstring for kwargs in `Instrument`, please`).
I guess in this case I have to plead that clarity is a trade-off (life is full of them :D) -- yes we're not showing upfront that `name` exists, but on the other hand signatures don't get cluttered with sometimes quite long docstrings.

bilderbuchi:
The only other approaches I saw would have been
* The current state, with longer signature
```
    def __init__(self, adapter, name="HP/Agilent/Keysight 34410A Multimeter", **kwargs):
        super().__init__(
            name=name,
            adapter, **kwargs
        )
```
* Shorter signature (but the None feels weird)
```
    def __init__(self, adapter, name=None, **kwargs):
        super().__init__(
            name=name or "HP/Agilent/Keysight 34410A Multimeter",
            adapter, **kwargs
        )
```
Did I miss a better way?

bilderbuchi:
> Given the above, I have to say that I am not really super-happy (obviously I could be wrong) to approve these changes, but I need to respect other people opinion (especially if they are working till late on the project for free :) ) and so feel free to merge them in the master.

Thank you for being gracious about this. On my side, I want to understand other perspectives (I could be wrong, too), and find out if maybe I didn't think of something. So, before merging I first want to understand as well as possible why you're unhappy with the change -- maybe we could use that to improve the PR. 
Also, you're one of the maintainers, too -- we should all be as happy as possible with the codebase, so working on pymeasure remains fun -- after all, we're all not getting paid for this. :-)

bilderbuchi:
> includeSCPI (instrument class, obviously if a particular instrument support SCPI is a feature of the instrument model,
not something that depends on particular instrument).

Thank you for your intution about configuration associated with the instrument _type_, not the _instance_ -- I had not thought of this before in this way! 
I agree concerning `includeSCPI` -- this actually has no place at all in the signature (it does not make sense to change it when instantiating). 
I hope/guess at some point (maybe when resolving #416) we will rectify this, e.g. by having two classes `Instrument` and `SCPIInstrument`, or the previously discussed mixins, to define the SCPI command set.

bmoneke:
Another option (without the possibility to name different instances of the same class) would be, to put the name in `__repr__`. This allows to give a longer name (`str(Fiat500) == "Fiat 500 small four seater car"`) without introducing any variable etc.
Additionally we could state, that the variable name remains reserved for giving individual names to instrument instances?

bilderbuchi:
> Another option (without the possibility to name different instances of the same class) would be, to put the name in `__repr__`. This allows to give a longer name (`str(Fiat500) == "Fiat 500 small four seater car"`) without introducing any variable etc. Additionally we could state, that the variable name remains reserved for giving individual names to instrument instances?

I'd really like to preserve the "possibility to name different instances of the same class", to be honest.
I think you meant `__str__`, not `__repr__`, as `__repr__` typically should (try to) "return a string that would yield an object with the same value when passed to eval()".
Also, doesn't that make the code even more obscure, as instrument implementors now have to implement `__str__`?

bmoneke:
It was just an idea I got, while seeing this code in the adapters. As we're discussing what to do, I thought it good to mention, even though we might not implement it.

msmttchr:
> 
>     * the approach preserves the _current ability_, if desired, to assign a custom name during instantiation.

`PyMeasure` documentation and instrument implementation, as of today, show that this ability is not currently available.
Sorry to be pedantic, but in the vast majority of instruments you do no have the ability to change name during instantiation.
The vast majority of `Instrument` writers intended the parameter as hard coded and non customizable by users.

msmttchr:
> > includeSCPI (instrument class, obviously if a particular instrument support SCPI is a feature of the instrument model,
> > not something that depends on particular instrument).
> 
> Thank you for your intution about configuration associated with the instrument _type_, not the _instance_ -- I had not thought of this before in this way! I agree concerning `includeSCPI` -- this actually has no place at all in the signature (it does not make sense to change it when instantiating). I hope/guess at some point (maybe when resolving #416) we will rectify this, e.g. by having two classes `Instrument` and `SCPIInstrument`, or the previously discussed mixins, to define the SCPI command set.

Yes, I agree with your view.

bilderbuchi:
> > ```
> > * the approach preserves the _current ability_, if desired, to assign a custom name during instantiation.
> > ```
> 
> `PyMeasure` documentation and instrument implementation, as of today, show that this ability is not currently available. Sorry to be pedantic, but in the vast majority of instruments you do no have the ability to change name during instantiation. The vast majority of `Instrument` writers intended the parameter as hard coded and non customizable by users.

Pedantry/precision is fine :-)
Regarding instrument implementations in current master that _do have_ this ability: 
* hcp tc038d
* anritsu AnritsuMS9710C
* andeenhagerling AH2500A and AH2700A (inheriting variant) 
* (and the virtualbench instrument, but that's not really in the style of `Instrument` right now)

Yes, the documentation is lacking here, but that's also what this PR is for. :-)
Another part is the fix that this is made use of consistently across all instruments.

bmoneke:
> bmoneke can you add name to the docstring for kwargs in `Instrument`, please

Instrument contains already a named argument `name` (already in docstring), therefore it should not be added to the kwargs docstring. If you want to add it to the docstring, it has to be added to every single instrument's docstring.

msmttchr:
Il Dom 7 Ago 2022, 19:34 Christoph Buchner ***@***.***> ha
scritto:

> * the approach preserves the _current ability_, if desired, to assign a custom name during instantiation.
>
> PyMeasure documentation and instrument implementation, as of today, show
> that this ability is not currently available. Sorry to be pedantic, but in
> the vast majority of instruments you do no have the ability to change name
> during instantiation. The vast majority of Instrument writers intended
> the parameter as hard coded and non customizable by users.
>
> Pedantry/precision is fine :-)
> Regarding instrument implementations in current master that *do have*
> this ability:
>
>    - hcp tc038d
>    - anritsu AnritsuMS9710C
>    - andeenhagerling AH2500A and AH2700A (inheriting variant)
>    - (and the virtualbench instrument, but that's not really in the style
>    of Instrument right now)
>
> Yes, the documentation is lacking here, but that's also what this PR is
> for. :-)
> Another part is the fix that this is made use of consistently across all
> instruments.
>
Quoting myself "vast majority" doesn't mean all.
So, if I understand well, your argument is that you consider the fact that
few instrument driver use what you think is right, then all the other
implementation should be changed to adapt.

In summary, almost none of the instruments offer this feature. This
feature, when present, is used in some rare special cases.

I think your investigation confirm what I proposed initially, both
documentation and vast majority of instrument implementation consider name
as read-only.

I consider this not only a weak argument to support the change, but indeed
a motivation to avoid the change and keep documentation as it is.

The saga continues.... :)

bilderbuchi:
> Quoting myself "vast majority" doesn't mean all.

Point taken. However, quoting you,
> ...instrument implementation, as of today, show that this ability is not currently available.

means none have it, which is incorrect, as I tried to show. :-)

Obviously, you feel strongly about this, so I propose we hit the "pause" button on this merge for now, and maybe revisit this at some later point, maybe after we can get feedback from @CasperSchippers or @cjermain.

I'll see if I can dig up how I did this earlier, maybe by adjusting the `name` attribute directly? If a user is really interested in changing the name, maybe we can suggest/document to just change the attribute directly.

bilderbuchi:
@bmoneke can you easily split out the unrelated and uncontentious change to `resourceName` in a separate PR, so that we can get that merged?

bmoneke:
Unfortunately I did the commits along the instrument lines because I commited after I did a bunch of files.
I'd have to go through all files (but from this PR I know, which files to change) to do it, resulting in a new PR.
My fear is, that a possible later merge of this PR will cause a lot of conflicts with this previous one. Or does git recognize that the line is already such, as this PR proposes?

CasperSchippers:
> maybe we can get input from @CasperSchippers ...

Sorry for the long time before I got involved. I don't have a very strong opinion about it but since asked:

For me, I think the only real use of the `name` property is that it can be used in error messages to more easily know which instrument raises an error (which, actually, is not or not consistently used in error messages I think, and maybe for some communication issues that you'd also like this information with, this is a bit more difficult to include). For other uses, I either use the class-type (in cases I have to check which type of instrument it is, I most often just use `isinstance`) or the variable name of the object (or if I would have a bunch of identical ones in a loop or something, I guess I'd rather use a dict or something). 

However, if you have a name property and want to use it (also with the clear error-messages in mind), I would be inclined to think that you also need the option to possibly change it, and know clearly what the default value is. Therefore it would make most sense to me to either have the `name` kwarg (with the default value) in the call signature of the init (i.e. `def __init__(self, adapter, name="HP/Agilent/Keysight 34410A Multimeter", **kwargs):`), since that seems to be the most clear version of what is happening (I think it is not really cluttered yet, because there are typically not so many arguments for the init of an instrument class (especially if things like `includeSCPI` are removed from the arguments, with which I agree by the way). Users can then change it upon initialization (which seems most natural for me, but I do not have any very strong arguments as to why), or not bother with it at al. Encouraging the users to (manually) create another property (e.g. `nickname`) I find not a very good idea, as this brings about risks of accidentally overwriting an already existing property.

Having said that, I can see (and agree) with the pro's and con's of both sides, but lean just a tad bit more to the 'changeable-name' side of the discussion. Finally, I wish to repeat that I do not have a strong opinion on the `name` discussion (except for the remark about the nickname-property, about which I do have a more strong opinion); but maybe my view on this can help a bit.

bilderbuchi:
I'm just dropping this here because it's related: If we should go ahead with a naming definition pattern, here is a doc snippet from @bklebel about it (#578):
```
Naming derived instruments
**************************

When deriving an instrument from an existing implementation, the derived instrument typically has a different name (e.g. another model number).
In addition, a user might want to give instantiated instruments setup-specific names (e.g. `Helmholtz_coil_driver` instead of `GenericPSU123`), especially if there is more than one device of the same make in a setup.
To accomodate these needs, there are two similar ways: 

.. code-block:: python
    class BaseClass:
        def __init__(self, adapter, name=None, **kwargs):
            #...
            super().__init__(
                adapter,
                name or "Base instrument name",
                **kwargs,
            )
    class DerivedClass(BaseClass):
        # or do the name handling like this
        def __init__(self, adapter, name="Derived instrument name", **kwargs):
            #...
            super().__init__(
                adapter,
                name=name,
                **kwargs,
            )
```

msmttchr:
Another example of issues with the name, see https://github.com/pymeasure/pymeasure/pull/698#discussion_r1018859477

bklebel:
Since I have been mentioned here, I will put in my two cents (even if it became a bit long now, in writing it...): 
I can understand the side where the `name` of an instrument which is implemented in a certain way should be a class attribute, as it describes exactly this instrument, and not some derived one. At the same time, I do not see why we would want/need such an attribute in the first place, if it was only used for such a purpose. We already have a "proper" name, i.e. the class name itself, which should suffice as identifier. 
My own intuition was that the `name` of an instrument is more like a nickname, as described above, to be able to identify a certain physical instrument - after all, afai understand it, we try to represent our instruments in the code, and I personally call my some of my instruments "Keithley nanovolt X", where X goes from 1 to 3, and "Keithley current Y", where Y can be 1 or 2. 
The only real use for a `name` attribute which I see is exactly this one, however I am surely open to learn more. 

I therefore still stand by what @bilderbuchi has quoted, where I personally prefer the second case, as I think it is more easily understood (and implemented) by our average person who wants to implement a device. 

The problem which @msmttchr has referenced just now would be circumvented if this was properly implemented in all devices. In the Keithley2260B apparently it was done as it is currently for many others, which hinders an easy renaming. 

The version with `kwargs.setdefault("name", "proper instrument name")` looks to be most elegant "under the hood", but might also not be understood as easily, and I agree, it might be unnecessarily much code. I therefore agree that it kind of obfuscates (to a not very experienced programmer) what happens, I am not sure whether the `dictionary.setdefault()` method is a standard method people tend to know. 

My personal preference for simplicity, usability, and readability for someone who might just be beginning to work with python is the following: 
```
class Agilent34410A(Instrument):
    """
    Represent the HP/Agilent/Keysight 34410A and related multimeters.
    Implemented measurements: voltage_dc, voltage_ac, current_dc, current_ac, resistance,
    resistance_4w
    """
    def __init__(self, adapter, name="HP/Agilent/Keysight 34410A Multimeter", **kwargs):
        super().__init__(
            adapter, 
            name,
            **kwargs,
        )

class DerivedFrom_Agilent34410A(Agilent34410A):
    """
    Represent a device for which we derive the class from the HP/Agilent/Keysight 34410A.
    Implemented measurements: voltage_dc, voltage_ac, current_dc, current_ac, resistance,
    resistance_4w
    """
    def __init__(self, adapter, name="Multimeter XXYY", **kwargs):
        super().__init__(
            adapter, 
            name,
            **kwargs,
        )
```

If I understand it correctly, this should not make troubles with inheritance, is reasonably simple, sets an instrument-related name for any class and instance, which can be easily overwritten. However, it would be again quite a lot of work to change again what has been done here with (I believe) some significant investment of time - since I cannot myself invest the time to do such a second change I would not ask it of anyone else, and would be quite fine with keeping `kwargs.setdefault...` as it is now.

bmoneke:
As we already merged a part of the code, we have to touch all files manually, independent of the agreed upon implementation.

As Caspar mentioned the removal of `includeSCPI`, (which touches many files as well), we could do them both together in one PR.

bmoneke:
> Another example of issues with the name, see [#698 (comment)](https://github.com/pymeasure/pymeasure/pull/698#discussion_r1018859477)

In order to ensure more easy subclassing, I'm for enabling all instruments to allow a `name` argument. This PR is already done, so I propse to use it.
If we decide to change to `name="Some funny instrument"` argument in the init, we still can do it without breaking anything. The work to do this change will be almost the same with or without this PR merged.
However, if you prefer, I would do the work to change all instruments to `super().__init__(adapter, name="My name", **kwargs)`.

Is that PR #698 enough for you, to see `name` as changeable, @msmttchr?

If not, we can close the PR (the code still remains and could be used later on).
@bilderbuchi

msmttchr:
> > Another example of issues with the name, see [#698 (comment)](https://github.com/pymeasure/pymeasure/pull/698#discu
> 
> Is that PR #698 enough for you, to see `name` as changeable, @msmttchr?
> 

Yes, even though I would do as class attribute as I mentioned previously. On the other hand, pretty much all users prefer to have it as parameter, so for me it's ok to have as parameter.

AidenDawn:
Something I remember from #719 is that if `ATSBase` and `Instrument` now have a `name` during the test, they don't need to be skipped in `test_all_instruments.py` anymore. I've given this a quick test in your branch, but it would be worth double-checking

bilderbuchi:
> Now remains the question of the documentation:
> 
> Do we want only one style (init parameter or `kwargs.setdefault`), do we allow both, but prefer one or do we allow both? In the first two questions: which one?

I have a slight preference for the "init parameter" way (as Casper said we don't have that much clutter in the signature especially when includeSCPI should disappear, and it's a bit clearer that this option is there). 
The prefernce is not strong, so I would also be ok with the `setdefault` way if others prefer that.
Whatever we choose, we should harmonize on that one in the codebase and docs (which will effectively lead by example), but not _forbid_ the other, imo

bmoneke:
I prefer the init parameter as well, but I'm reluctant to touch all the instruments.

What do you think:
1. We reduce this PR to make the documentation only (regarding init parameter).
2. We introduce the SCPIMixin (without yet removing includeSCPI).
3. When we switch all instruments from includeSCPI to a SCPIMixin, we have to touch all of them (without introducing any other code of discussion), such that we can do both (SCPI and init parameter) in one go.

bilderbuchi:
Excellent suggestion, agreed, that's better!

bmoneke:
Divide and conquer 😁

I have learnt my lesson of large PRs with too little talk about the principles first.

bmoneke:
As discussed, I stored the changes in a local branch (if we should need them) and started from scratch:

- Documentation is changed to the new style
- All instruments are tested for accepting a "name" parameter. All currently present instruments are exempt (for future rework). The scope of the tests is to force new instruments to use that scheme.

Should we add a change note? It only affects new instrument creators.

tests/instruments/test_all_instruments.py
```diff
@@ -54,14 +54,97 @@
...
+]
+
+
+@pytest.mark.parametrize("cls", devices)
+def test_name_argument(cls):
+    "Test that every instrument accepts a name argument"
+    if cls.__name__ in (*proper_adapters, *init):
+        pytest.skip(f"{cls.__name__} cannot be tested without communication.")
+    elif cls.__name__ in nameless_instruments:
+        pytest.skip(f"{cls.__name__} does not accept a name argument.")
```
bilderbuchi, line 148:
```suggestion
        pytest.skip(f"{cls.__name__} does not accept a name argument yet.")
```
:D

tests/instruments/test_all_instruments.py
```diff
@@ -54,14 +54,97 @@
 init = ["ThorlabsPM100USB", "Keithley2700", "TC038", "Agilent34450A",
```
bilderbuchi, line 54: Clearer name `need_init_communication`?

tests/instruments/test_all_instruments.py
```diff
@@ -54,14 +54,97 @@
...
         "AWG401x_AWG", "AWG401x_AFG", "VARX", "HP8116A"]
 # Instruments which require more input arguments
-init.extend(["Instrument", "ATSBase"])
+name_required = ["ATSBase"]
```
bilderbuchi, line 57: Clearer name `require_more_inputargs`?
Then the comments also become superfluous. :-)

tests/instruments/test_all_instruments.py
```diff
@@ -54,14 +54,97 @@
...
 @pytest.mark.parametrize("cls", devices)
-def test_args(cls):
+def test_adapter_arg(cls):
     "Test that every instrument has adapter as their input argument"
     if cls.__name__ in proper_adapters:
         pytest.skip(f"{cls.__name__} does not accept an Adapter instance.")
-    elif cls.__name__ in init:
+    elif cls.__name__ in (*init, *name_required, "Instrument"):
         pytest.skip(f"{cls.__name__} requires communication in init.")
     cls(adapter=MagicMock())
```
bilderbuchi, line 67:
```suggestion
    elif cls.__name__ in (*init):
        pytest.skip(f"{cls.__name__} requires communication in init.")
    elif cls.__name__ in (*name_required, "Instrument"):
        pytest.skip(f"{cls.__name__} requires more input arguments.")
    cls(adapter=MagicMock())
```
To keep the skip messages accurate.
Why does plain `Instrument` not work, here?

bilderbuchi (approved):
LGTM, with some restructuring suggestions for the tests.

bilderbuchi:
> Should we add a change note? It only affects new instrument creators.

Why not. A hint about the new best practice could be useful.

When I compile the changelog, I mostly go by PR titles and what I remember from the PRs. If there are details that should or must be in the changelog, it's good to add them with the PR, at least for us maintainers -- there's less potential that something is missed. This primarily concerns "core"-touching PRs, for instruments it's straightforward enough most of the time.

tests/instruments/test_all_instruments.py
bmoneke: Because `Instrument` needs a `name` argument.
If I supply it for all, the other instruments don't like it.

bmoneke:
Thanks for the suggestions.

tests/instruments/test_all_instruments.py
AidenDawn: I've been following this discussion for a while and wanted to quickly jump in on this point specifically. during #61, I was playing around with this part of the `SPDBase` and had originally included it here, but a comment from @bilderbuchi meant I was able to remove the SPDBase from the equivalent of `name_required` at the time.

> My only excuse is that it must have been too late to notice that comment :-P
> 
> What if you add `kwargs.setdefault('name', 'Siglent SPDxxxxX instrument Base Class')` in `SPDBase`'s init, just before you call `super()__init__`? This should enable using it in tests, but still result in the correct names for the child classes?

Can I just quickly ask to understand why the same thing does not apply here for `ATSBase` or `Instrument` with regards to `name` if a `kwargs.setdefault(...)` has been set for both classes? (I did try to ask this before but I think it went unnoticed)

bmoneke: `Instrument` does not have a default name (in order to enforce a name in the instruments, I think).

Atsbase has to be changed, yet.

tests/instruments/test_all_instruments.py
bmoneke: Done

tests/instruments/test_all_instruments.py
bmoneke: I just added the proper code to the ATSBase 😁

tests/instruments/test_all_instruments.py
bmoneke: Done.

tests/instruments/test_all_instruments.py
bmoneke: I removed that category and added an individual test for Instrument name (with a proper message).

Another option would be, to allow the instrument to have a default name (which I do not like).

docs/dev/adding_instruments.rst
```diff
@@ -236,10 +236,10 @@ The simplest version, suitable when the instrument connection needs default sett
 
 .. code-block:: python
 
-    def __init__(self, adapter, **kwargs):
+    def __init__(self, adapter, name="Extreme 5000", **kwargs):
```
msmttchr, line 239: Minor comment on line 214 above: a blank line needs to be added as follow:
```
The status property could look like this:
<blank line here>
.. testcode::

```

tests/instruments/test_all_instruments.py
bilderbuchi: Perfect, thanks!

docs/dev/adding_instruments.rst
bmoneke: Done.

bmoneke:
As you both had only these minor (unrelated) changes, we can merge, I think.

msmttchr:
> As you both had only these minor (unrelated) changes, we can merge, I think.

I agree

msmttchr (approved):
Good for me

bilderbuchi (approved):
LGTM

PR #660, opened 2022-08-02
Adapter rework

bmoneke:
Reworking the instrument/adapter. This pull request just focuses on the official adapters, Instrument class and keeping the code base in a working condition.

Fixes #567, fixes #512, fixes #409, fixes #219.

Roadmap of different parts (PRs)
----------
1. Preliminary tests of instruments/adapters in the current state #665 
2. This pull request with changes to the official adapters and minimum changes to the instruments to use these
3. A test code generator.
4. Pull requests for changes in individual instruments (removing custom adapters, where applicable)

Open questions
----------------
- [x] move read/write binary_values to adapter or instrument?
- [x] query_delay in instrument.ask or should the instrument replace ask adding its own delay method if necessary?
- [x] remove FakeAdapter or is there any use for it, having implemented the Protocol adapter?
- [x] Deprecate the vxi11 adapter in favour of pyvisa?
- [x] make SerialAdapter similar to pyvisa: timeout in ms? Or both in s?

Adapter rework
==========
General
--------
- [x] Write tests for instruments to be changed. (tests using devices are called `test_instrument_with_device.py` : add appropriate comment into documentation)
- [x] Harmonise ask being write+read (i.e., remove from VISAAdapter), use protocol tests to confirm it works correctly
- [x] Implement waiting in Instrument.ask. time.sleep(0) takes 0.5 µs.
- [x] Remove all now-unnecessary ask() implementations, read and write implementations should have all (test this)
- [x] Implement over-the-wire message-traffic logging here (protocol.py), first
- [x] Check timing impact of non-firing logging messages: less then 0.5 µs
- [x] Roll out message-traffic logging to all Adapters, try to DRY
- [x] adjust tests if instruments got changed (especially those with their own adapter.
- [x] check newly added instruments
- [x] Include a test generator which generates tests out of device communication
- [x] Adjust the documentation: All protocol quirks are in read/write. Timing issues in ask (defaults to write + wait + read). Remove the "custom adapters in init" examples from the "connecting.rst". It's better to adjust read than to write your own adapter. If you need a special adapter, check, whether an adapter had been handed over first (for example for protocol_tests).
- [x] Document how Adapters work and why only the the private methods should be overridden.
- [x] Add a remark in changes.rst

Official adapters
----------------
- [x] prologix adapter: remove rw_delay
- [x] Adapter: Add adapter.close
[...]

PR #665, opened 2022-08-05
Instruments and adapter protocol tests for adapter rework

bmoneke:
Tests for instruments / adapters before adapter rework( #660 ), in order to verify correct working afterwards.

Tests which require instruments connected have been renamed.
Tests added and small, obvious mistakes fixed.


pymeasure/instruments/attocube/adapters.py
```diff
@@ -25,7 +25,31 @@
 import re
 import time
 
-from pymeasure.adapters import TelnetAdapter
+from pymeasure.adapters import ProtocolAdapter, TelnetAdapter
+
+
+class Mock_TelnetAdapter(ProtocolAdapter):
```
bilderbuchi, line 31: Please keep test-specific code in the tests/ directory (in this case test_anc300.py), to keep that out of the pymeasure installed package.

pymeasure/instruments/fwbell/fwbell5080.py
```diff
@@ -63,6 +63,7 @@ class FWBell5080(Instrument):
         'amp-meter': 'DC:AM', 'amp-meter ac': 'AC:AM'
     }
     units = Instrument.control(
+        # TODO a newer version expects ':UNIT:FLUX:%s', is this right?
```
bilderbuchi, line 66: possibly a typo - @cjermain could know.

tests/adapters/test_visa.py
```diff
@@ -37,8 +37,10 @@
     pytest.skip('PyVISA tests require the pyvisa-sim library', allow_module_level=True)
 
 
-def make_visa_adapter(**kwargs):
-    return VISAAdapter(SIM_RESOURCE, visa_library='@sim', **kwargs)
+@pytest.fixture
+def adapter():
+    return VISAAdapter(SIM_RESOURCE, visa_library='@sim',
+                       read_termination="\n")
```
bilderbuchi, line 43: The read_termination should be set by the individual SIM_RESOURCE -- 'ASRL2::INSTR is built-in in pyvisa-sim, I think

tests/instruments/attocube/test_anc300.py
```diff
@@ -0,0 +1,61 @@
...
+#
+
+import pytest
+
+from pymeasure.instruments.attocube import ANC300Controller
+from pymeasure.instruments.attocube.adapters import Mock_TelnetAdapter, AttocubeConsoleAdapter
+
+#############################################################################
+# In order to run the tests, you have to change the superclass of the adapter
+# from TelnetAdapter to Mock_TelnetAdapter
```
bilderbuchi, line 32: can't you say here 
```
pymeasure.instruments.attocube.adapters.AttocubeConsoleAdapter = Mock_TelnetAdapter
```
and then instantiate an `ANC300Controller`

bilderbuchi: there's even a mock/monkeypatch context manager in the stdlib to do this "safely" (i.e. undo it after leaving the scope), but the details escape me right now.

bmoneke: I tried it with monkeypatch, but was not able to get it to work, therefore I did this dirty solution.

I'll try it with that suggestion.

tests/adapters/test_visa.py
bmoneke: the SIM_RESOURCE just defines how the virtual device behaves. The default device (ASRL2) writes "\n". That does not define, how pyvisa (the computer end) communicates with the virtual device. Therefore, I have to tell the VISAAdapter what to expect.

tests/instruments/hp/test_hp8116a.py
```diff
@@ -23,244 +23,39 @@
...
 
 import pytest
-import math
-import time
-from pymeasure.instruments.hp import HP8116A, hp8116a
 
-pytest.skip('Only work with connected hardware', allow_module_level=True)
+from pymeasure.test import expected_protocol
 
+from pymeasure.instruments.hp import HP8116A as _HP8116A
```
bilderbuchi, line 29: I'm curious: why this? Can't you do something like
```
HP8116A.status = property(fget=lambda: Status(5))
```

tests/instruments/hp/test_hp8116a.py
```diff
@@ -23,244 +23,39 @@
...
+            HP8116A,
+            [(b"CST", b"x" * 87 + b' ,\r\n'), (b"IDTY", b"00000035")],
+            ) as instr:
+        assert instr.duty_cycle == 35
 
-    FREQUENCIES = [[1, 1], [1.23, 1.23], [1e3, 1e3], [1.23e3, 1.23e3], [1e6, 1e6], [1.23e6, 1.23e6],
-                   [1.234, 1.23], [1.234e3, 1.23e3], [10.234e3, 10.2e3], [1.234e6, 1.23e6]]
 
-    instr = HP8116A(RESOURCE)
-
```
bilderbuchi, line 77: will you convert more of the already existing tests?

tests/instruments/attocube/test_anc300.py
bilderbuchi: I did stuff like that with `print` before 😅 you gotta love Python

bilderbuchi:
This is looking good! Do you agree that it would be interesting to measure [coverage](https://coverage.readthedocs.io/en/6.4.2/) at least of the adapter classes and `Instrument` to see if the current tests still have blind spots that we could cover with some more targeted tests?

tests/instruments/hp/test_hp8116a.py
bmoneke: That's more elegant. Thanks for that suggestion

tests/instruments/attocube/test_anc300.py
bmoneke: Oh. I do not replace the AttocubeAdapter with the Mock_TelnetAdapter, but I replace the TelnetAdapter (parent class of AttocubeAdapter) with the Mock_Telnet. In other words, I want to replace the parent class of the AttocubeAdapter.

That is more difficult to do.

bmoneke: I just kept the (for me successfully run) test parameters and mark them to fail. After the adapter rework, they should pass.

bmoneke:
> This is looking good! Do you agree that it would be interesting to measure [coverage](https://coverage.readthedocs.io/en/6.4.2/) at least of the adapter classes and `Instrument` to see if the current tests still have blind spots that we could cover with some more targeted tests?

I looked a bit into coverage and it seems easier to use than I thought. What are your plans how to use it? It's always good to know, what is covered and what not.

bilderbuchi:
> I looked a bit into coverage and it seems easier to use than I thought. What are your plans how to use it? It's always good to know, what is covered and what not.

In the long run, it would be nice to have it in CI, too, but until now our test coverage was so miserable that it didn't make sense. If it runs in CI it would mainly be useful _if_ we expect tests to accompany pull requests (so that coverage does not decrease), and I think we're far from that (at least for instrument PRs).

What could maybe be useful is that we could use coverage tracking to make sure that the pymeasure core parts (I think most everything except examples and instruments; personally I'm thinking mainly of `Instrument` and adapters) remain fully covered, that would already be a help to ensure quality.

However, right here, I would just run it manually and look at the reports to find out where we still need tests. If you want, we can add a config entry to pyproject.toml reflecting the previous paragraph and any other needed config, if that makes it easier to re-run later.

tests/adapters/test_adapter.py
```diff
@@ -22,25 +22,52 @@
...
-    assert a.values("5,6,7", cast=str) == ['5', '6', '7']
-    assert a.values("X,Y,Z") == ['X', 'Y', 'Z']
-    assert a.values("X,Y,Z", cast=str) == ['X', 'Y', 'Z']
-    assert a.values("X.Y.Z", separator='.') == ['X', 'Y', 'Z']
+    assert a.values(value, **kwargs) == result
 
 
 def test_adapter_preprocess_reply():
     a = FakeAdapter(preprocess_reply=lambda v: v[1:])
+    str(a)
```
bilderbuchi, line 70: ?

tests/instruments/test_instrument.py
```diff
@@ -317,6 +446,11 @@ class GenericInstrument(FakeInstrument):
     )
 
 
+@pytest.fixture(scope="module")
+def generic():
+    return GenericInstrument()
```
bilderbuchi, line 451: Can you move fixtures and such tooling to before the place of first use, please?

tests/instruments/test_instrument.py
```diff
@@ -52,6 +125,28 @@ class Fake(Instrument):
...
 
+def test_control_check_errors_get(generic):
+    generic.fake_ctrl_check_get_errors = True
+
+    def checking():
+        generic.error = True
+    generic.check_errors = checking
+    generic.fake_ctrl
+    assert generic.error is True
+    del generic.error
```
bilderbuchi, line 136: Those dels at the end of tests seem superfluous?

tests/instruments/test_instrument.py
bilderbuchi: ...or the tests using it after that?

tests/instruments/test_instrument.py
bmoneke: The generic object is created once (scope=module), therefore I have to delete that property afterwards, lest the next test still has generic.error == True.

tests/adapters/test_adapter.py
bmoneke: Just verifying, that the representation is called (code coverage...), yeah, I should assert it.

tests/instruments/test_instrument.py
bilderbuchi: Ah, I missed that on the phone. Why the module scope though, should not be expensive to make a new instance for every test, right?

tests/instruments/test_instrument.py
```diff
@@ -23,13 +23,52 @@
...
+    )
+    fake_measurement = Instrument.measurement(
+        "", "docs",
+        values={'X': 1, 'Y': 2, 'Z': 3},
+        map_values=True,
+        dynamic=True,
+    )
+
+
+@pytest.fixture(scope="module")
```
bilderbuchi, line 60: I think we can afford not having this in module scope, so it gets recreated for every test -- this will also help with test isolation, and you don't need those dels anymore..

tests/instruments/test_instrument.py
```diff
@@ -52,6 +159,28 @@ class Fake(Instrument):
     assert Fake.x.__doc__ == expected_doc
 
 
+def test_control_check_errors_get(generic):
+    generic.fake_ctrl_check_get_errors = True
+    generic.error = False
+
+    def checking():
+        generic.error = True
```
bilderbuchi, line 167: shouldn't that be
```suggestion
    def checking(self):
        self.error = True
```
?

tests/instruments/test_instrument.py
```diff
@@ -23,13 +23,52 @@
...
+def generic():
+    return GenericInstrument()
+
+
+class NewRangeInstrument(GenericInstrument):
+    # Choose different properties' values, like you would for another device model
+    fake_ctrl_values = (10, 20)
+    fake_setting_values = (10, 20)
+    fake_measurement_values = {'X': 4, 'Y': 5, 'Z': 6}
+
```
bilderbuchi, line 70: Can you also move ExtendedInstrument and StrictExtendedInstrument up here, so that all the related classes are in one place?

tests/instruments/test_instrument.py
bmoneke: I solved it differently, but yes, we can remove it.

tests/instruments/test_instrument.py
bmoneke: At first I did it as you suggested, but then I moved it back down to their usage...

tests/instruments/test_instrument.py
bmoneke: That is, what I had tried first, but I got an error, that checking expected an argument, but got none.

tests/instruments/test_instrument.py
bilderbuchi: hm, ok, that's also a fair point, and only taste in the end.

tests/instruments/test_instrument.py
bilderbuchi: yes, because you're attaching it to an instance, not a class, so it does not get [bound automatically](https://stackoverflow.com/questions/972/adding-a-method-to-an-existing-object-instance).
Below invoke the "descriptor protocol" and do `generic.check_errors = checking.__get__(generic)`, should do the trick.

tests/instruments/test_instrument.py
bilderbuchi: Good :-)

bmoneke:
I rebased onto the last merge.

pymeasure/instruments/fwbell/fwbell5080.py
bilderbuchi: Let's leave the TODO in and wait for a reply from Colin, or a bug report by someone else.

bilderbuchi:
I think we're done. Agreed?

bmoneke:
Agreed!

bilderbuchi:
:shipit: then!

PR #674, opened 2022-08-08
Harmonize instruments' adapter argument

bmoneke:
Similar to #659 , but only harmonizing the instruments' adapter argument (including documentation).

I ran all the tests: they succeeded.
I also introduced the "all_instruments_test" from #659 verifying, that a keyword argument "adapter" is enough (i.e. if a device desired a 'port' or 'resourceName', it would fail).

As I used regex ('resource(N|_n)ame'), I should not have done a spelling mistake with those (I verified with my eyes, but many files...).


bilderbuchi:
LGTM, thanks!

PR #683, opened 2022-08-23
Add black configuration

bleykauf:
I think there are some people out there using black as a code formatter, so it might be useful to include some configuration for this in `pyproject.toml` and `setup.cfg`.

 In this PR, I added some configurations to make it easier to use it with the custom line length of 100 (flake8 is configured with line length 100 in `setup.cfg`). 

I also added the recommended options from https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html as separate commits.

For people not using black, nothing should change, except maybe flake8 warnings regarding E203 (which according to the link above is not PEP8 compliant anyway).

### DIFF
--- a/pymeasure/instruments/instrument.py
+++ b/pymeasure/instruments/instrument.py
@@ -277,18 +277,18 @@ class Instrument:
     @staticmethod
     def control(  # noqa: C901 accept that this is a complex method
         get_command,
-            set_command,
-            docs,
-            validator=lambda v, vs: v,
-            values=(),
-            map_values=False,
-            get_process=lambda v: v,
-            set_process=lambda v: v,
-            command_process=lambda c: c,
-            check_set_errors=False,
-            check_get_errors=False,
-            dynamic=False,
-            **kwargs
+        set_command,
+        docs,
+        validator=lambda v, vs: v,
+        values=(),
+        map_values=False,
+        get_process=lambda v: v,
+        set_process=lambda v: v,
+        command_process=lambda c: c,
+        check_set_errors=False,
+        check_get_errors=False,
+        dynamic=False,
+        **kwargs
     ):
         """Returns a property for the class based on the supplied
         commands. This property may be set and read from the
--- a/pymeasure/instruments/instrument.py
+++ b/pymeasure/instruments/instrument.py
@@ -275,19 +275,21 @@ class Instrument:
         return self.adapter.binary_values(command, header_bytes, dtype)
 
     @staticmethod
-    def control(get_command,  # noqa: C901 accept that this is a complex method
-                set_command,
-                docs,
-                validator=lambda v, vs: v,
-                values=(),
-                map_values=False,
-                get_process=lambda v: v,
-                set_process=lambda v: v,
-                command_process=lambda c: c,
-                check_set_errors=False,
-                check_get_errors=False,
-                dynamic=False,
-                **kwargs):
+    def control(  # noqa: C901 accept that this is a complex method
+        get_command,
+            set_command,
+            docs,
+            validator=lambda v, vs: v,
+            values=(),
+            map_values=False,
+            get_process=lambda v: v,
+            set_process=lambda v: v,
+            command_process=lambda c: c,
+            check_set_errors=False,
+            check_get_errors=False,
+            dynamic=False,
+            **kwargs
+    ):
         """Returns a property for the class based on the supplied
         commands. This property may be set and read from the
         instrument. See also :meth:`measurement` and :meth:`setting`.
--- a/docs/dev/coding_standards.rst
+++ b/docs/dev/coding_standards.rst
@@ -12,12 +12,17 @@ The `PEP8 style guide`_ and `PEP257 docstring conventions`_ should be followed.
 .. _PEP8 style guide: https://www.python.org/dev/peps/pep-0008/
 .. _PEP257 docstring conventions: https://www.python.org/dev/peps/pep-0257/
 
-Function and variable names should be lower case with underscores as needed to seperate words. CamelCase should only be used for class names, unless working with Qt, where its use is common.
+Function and variable names should be lower case with underscores as needed to separate words. CamelCase should only be used for class names, unless working with Qt, where its use is common.
 
 In addition, there is a configuration for the `flake8`_ linter present. Our codebase should not trigger any warnings.
 Many editors/IDEs can run this tool in the background while you work, showing results inline. Alternatively, you can run ``flake8`` in the repository root to check for problems. In addition, our automation on Github also runs some checkers. As this results in a much slower feedback loop for you, it's not recommended to rely only on this.
 
 .. _flake8: https://flake8.pycqa.org/en/latest/
+
+It is allowed but not required to use the `black`_ code formatter. Do only format files that you are actively working on and make a dedicated commit for the initial formatting when working on an existing file that does not adhere to the black code style to facilitate code review. If there are conflicts between black's output and flake8 (especially related to `E203`_), flake8 takes precedence. Use ``#noqa : E203`` to disable E203 warnings for a specific line if appropriate.
+
+.. _black: https://black.readthedocs.io/en/stable/
+.. _E203: https://www.flake8rules.com/rules/E203.html
 
 There are no plans to support type hinting in PyMeasure code. This adds a lot of additional code to manage, without a clear advantage for this project. 
 Type documentation should be placed in the docstring where not clear from the variable name.
--- a/pymeasure/instruments/agilent/agilentB1500.py
+++ b/pymeasure/instruments/agilent/agilentB1500.py
@@ -1810,7 +1810,7 @@ class QueryLearn():
     def _get_smu(key, smu_references):
         # command without channel
         command = re.findall(r'(?P<command>[A-Z]+)', key)[0]
-        channel = key[len(command):]
+        channel = key[len(command) :]  # noqa: E203
         return smu_references[int(channel)]
 
     # SMU Modes
--- a/pymeasure/instruments/anaheimautomation/dpseriesmotorcontroller.py
+++ b/pymeasure/instruments/anaheimautomation/dpseriesmotorcontroller.py
@@ -274,7 +274,8 @@ class DPSeriesMotorController(Instrument):
         raise NotImplementedError("steps_to_absolute() must be implemented in subclasses!")
 
     def reset_position(self):
-        """ Reset the position as counted by the motor controller and an externally connected encoder to 0.
+        """
+        Reset position as counted by the motor controller and an externally connected encoder to 0.
         """
         # reset encoder recorded position #
         self.write("ET")
--- a/pymeasure/instruments/hp/hp3437A.py
+++ b/pymeasure/instruments/hp/hp3437A.py
@@ -242,7 +242,7 @@ class HP3437A(HPLegacyInstrument):
         else:
             processed_data = []
             for i in range(0, len(read_data), 2):
-                processed_data.append(self._unpack_data(read_data[i:i+2]))
+                processed_data.append(self._unpack_data(read_data[i : i + 2]))  # noqa: E203
             return_value = np.array(processed_data)
         self.adapter.connection.timeout = current_timeout
         return return_value
--- a/pymeasure/instruments/keysight/keysightDSOX1102G.py
+++ b/pymeasure/instruments/keysight/keysightDSOX1102G.py
@@ -104,7 +104,7 @@ class Channel():
 
     scale = Instrument.control(
         "SCALe?", "SCALe %f",
-        """ A float parameter that specifies the vertical scale, or units per division, in Volts."""
+        """A float parameter that specifies the vertical scale, or units per division, in Volts."""
     )
 
     def __init__(self, instrument, number):
--- a/pymeasure/instruments/srs/sr830.py
+++ b/pymeasure/instruments/srs/sr830.py
@@ -473,8 +473,8 @@ class SR830(Instrument):
                 self.pause_buffer()
                 return ch1, ch2
         self.pauseBuffer()
-        ch1[index:count + 1] = self.buffer_data(1, index, count)
-        ch2[index:count + 1] = self.buffer_data(2, index, count)
+        ch1[index : count + 1] = self.buffer_data(1, index, count)  # noqa: E203
+        ch2[index : count + 1] = self.buffer_data(2, index, count)  # noqa: E203
         return ch1, ch2
 
     def buffer_measure(self, count, stopRequest=None, delay=1e-3):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,3 +7,6 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 100
+
+[tool.isort]
+profile = "black"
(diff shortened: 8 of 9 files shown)


bilderbuchi:
Thanks! I know many people use black nowadays, so if that can be added transparently/non-destructively, that's fine by me. Flake8 is the canonical formatter for pymeasure (also run in our CI), so it's maybe a good idea to add a note about this to the black config entry?

I find it a bit :rofl: how black's "compatbility with other tools" page essentially tells you how the _other_ tools have to change so that black config does not have to be adapted. "Opinionated formatter", indeed :-|

Losing the E203 flag (whitespace before ‘,’, ‘;’, or ‘:’) sure is inconvenient, as flake8 will now not catch a number of frequent other style errors (`mydict= {'a' : 3}`). We could instead annotate the false-positive occurrences of slicing operator spacing with ` #noqa: E203` if they are few as I suspect? Hopefully at some point the relevant upstream issues https://github.com/PyCQA/pycodestyle/issues/373 and https://github.com/PyCQA/pycodestyle/pull/914 will be resolved.

On the topic of non-destructiveness: Do I understand you correctly that with those settings, you have formatted all of pymeasure with black, and then run flake8 and there were no deviations flagged?

SBresler:
> I also added the recommended options from https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html as separate commits.
> 
> For people not using black, nothing should change, except maybe flake8 warnings regarding E203 (which according to the link above is not PEP8 compliant anyway).

I had this exact thought and this was brought up coincidentally a few days after I forked and ran black through everything.

the first step I do when cloning ANY package from github is run it through black. It doesn't do everything perfectly but it just frees up so much mental bandwidth when browsing through established packages to check things.

bleykauf:
Sorry for the very late reply, this somehow slipped through my notifications.

@bilderbuchi:
> Flake8 is the canonical formatter for pymeasure (also run in our CI), so it's maybe a good idea to add a note about this to the black config entry?

With the exception of E203, a black formatted file will always satisfy flake8 so I don't think there is any note necessary. Or do you mean something else?

> Losing the E203 flag (whitespace before ‘,’, ‘;’, or ‘:’) sure is inconvenient, as flake8 will now not catch a number of frequent other style errors (mydict= {'a' : 3}). We could instead annotate the false-positive occurrences of slicing operator spacing with #noqa: E203 if they are few as I suspect? Hopefully at some point the relevant upstream issues https://github.com/PyCQA/pycodestyle/issues/373 and https://github.com/PyCQA/pycodestyle/pull/914 will be resolved.

As I understand it, E203 is at least in its current implementation not PEP8 compliant so is thus ignored. I just ran black on all files and checked with flake8:

```
.\pymeasure\instruments\instrument.py:225:101: E501 line too long (105 > 100 characters)
.\pymeasure\instruments\instrument.py:290:5: C901 'Instrument.control' is too complex (17)
.\pymeasure\instruments\agilent\agilentB1500.py:1859:35: E203 whitespace before ':'
.\pymeasure\instruments\anaheimautomation\dpseriesmotorcontroller.py:167:101: E501 line too long (101 > 100 characters)
.\pymeasure\instruments\anaheimautomation\dpseriesmotorcontroller.py:286:101: E501 line too long (109 > 100 characters)
.\pymeasure\instruments\anaheimautomation\dpseriesmotorcontroller.py:368:101: E501 line too long (101 > 100 characters)
.\pymeasure\instruments\hp\hp3437A.py:239:68: E203 whitespace before ':'
.\pymeasure\instruments\hp\hp8116a.py:241:101: E501 line too long (103 > 100 characters)
.\pymeasure\instruments\keysight\keysightDSOX1102G.py:116:101: E501 line too long (101 > 100 characters)
.\pymeasure\instruments\keysight\keysightDSOX1102G.py:439:101: E501 line too long (103 > 100 characters)
.\pymeasure\instruments\srs\sr830.py:541:18: E203 whitespace before ':'
.\pymeasure\instruments\srs\sr830.py:542:18: E203 whitespace before ':'
.\pymeasure\instruments\srs\sr860.py:403:101: E501 line too long (104 > 100 characters)
.\pymeasure\instruments\srs\sr860.py:411:101: E501 line too long (104 > 100 characters)
.\pymeasure\instruments\srs\sr860.py:419:101: E501 line too long (104 > 100 characters)
.\pymeasure\instruments\srs\sr860.py:427:101: E501 line too long (104 > 100 characters)
.\pymeasure\instruments\srs\sr860.py:659:101: E501 line too long (101 > 100 characters)
.\pymeasure\instruments\srs\sr860.py:668:101: E501 line too long (101 > 100 characters)
.\pymeasure\instruments\srs\sr860.py:677:101: E501 line too long (101 > 100 characters)
.\pymeasure\instruments\srs\sr860.py:686:101: E501 line too long (101 > 100 characters)
.\pymeasure\instruments\temptronic\temptronic_ats545.py:72:101: E501 line too long (101 > 100 characters)
.\pymeasure\instruments\temptronic\temptronic_base.py:558:101: E501 line too long (101 > 100 characters)
```

So there is only 4 occurences of a falsly identified E203. The rest are line lengths errors E501 which have existed before the application of black.

> On the topic of non-destructiveness: Do I understand you correctly that with those settings, you have formatted all of pymeasure with black, and then run flake8 . and there were no deviations flagged?

I was unclear about that. I did not intend to run black on all files with this PR. Doing so would currenlty changes 187 files and all line changes will be attributed to the person commiting these formatting changes potentially making git blame less useful.

What I wanted to do is to add the configuration to aid those people who do want to use black. Typically you would tell your IDE to auto-format using black upon saving. I simply added the config files to make it work out of the box (black checks for these configuration files automatically).

So the behaviour right now would be that nothing changes for people not using black locally at all (with the exception of the E203 issue). If somebody has set up black as described above, only files that are being worked on will be formatted with black. 

One way I see to gradually transition to black would be to use the appropriate pre-commit hook and [this bot](https://github.com/apps/pre-commit-ci) to automatically format with black after pushing.

@SBresler 
> It doesn't do everything perfectly but it just frees up so much mental bandwidth when browsing through established packages to check things.

Absolutely agree, its just so nice to never think about formatting again!

bilderbuchi:
> With the exception of E203, a black formatted file will always satisfy flake8 so I don't think there is any note necessary. Or do you mean something else?

I'm not sure about that, actually. I just ran ` flake8 .` (and `flake8 . --count --extend-select=E9,F63,F7,F82 --show-source --statistics` like our CI does, as that gives a bit more information) 
* at the point where your PR branched off (9ebc63) to confirm that it was clean
* Again, at your current PR HEAD (09d56b) to confirm it was clean
* Then I installed black (22.8.0), ran `black .`, it changed 189 files like you observed. I agree that it wouldn't make sense to do that formatting pass now.
* Now I ran flake8 again, and got
```
$ flake8 . --count --extend-select=E9,F63,F7,F82 --show-source --statistics
./pymeasure/instruments/instrument.py:290:5: C901 'Instrument.control' is too complex (17)
    def control(
    ^
./pymeasure/instruments/anaheimautomation/dpseriesmotorcontroller.py:286:101: E501 line too long (109 > 100 characters)
        """Reset the position as counted by the motor controller and an externally connected encoder to 0."""
                                                                                                    ^
./pymeasure/instruments/keysight/keysightDSOX1102G.py:116:101: E501 line too long (101 > 100 characters)
        """ A float parameter that specifies the vertical scale, or units per division, in Volts.""",
                                                                                                    ^
1     C901 'Instrument.control' is too complex (17)
2     E501 line too long (109 > 100 characters)
3
```
The first one is a false positive, as the black reformatting pushes a `noqa` comment to the wrong line -- easily fixed.
The line-too-long things tell me that black apparently does not observe its own configuration settings (`line-length = 100`).
(If I switch E203 on again, I get 2 warnings about that, not 4 like you do. :curious: -- I would lean towards just decorating those two with `#noqa: E203` comments, and leaving E203 in place to catch `mydict= {'a' : 3}`)

So yeah, precisely because black formatted file apparently will **not** always satisfy flake8 (and seems to be weird about its own config), I would like a note about this -- in case there is disagreement in the future, the precedence is clear. :-)

bilderbuchi:
> What I wanted to do is to add the configuration to aid those people who do want to use black. 

Sure thing, as long as it does not interfere/can coexist with the established workflow, why not.

> Typically you would tell your IDE to auto-format using black upon saving. I simply added the config files to make it work out of the box (black checks for these configuration files automatically).

As does flake8, btw.

> So the behaviour right now would be that nothing changes for people not using black locally at all (with the exception of the E203 issue). If somebody has set up black as described above, only files that are being worked on will be formatted with black.

OK, as we have found that black would edit 80% of the relevant files in the codebase, the probability is high that someone working on pymeasure using black will introduce unrelated edits just by saving the file. Do you think those folks will be fine with accepting that, or is there a black setting that only formats/changes touched _lines_ (so, lines with a git diff), not _files_?

bilderbuchi:
As the necessary changes to integrate this smoothly seem to be few, what do you think about
* Adding a note/comment in the black config that flake8 is the canonical formatter, and takes precedence
* Maybe you can find out why black does not honour its own line length setting?
* Leave the flake8 E203 warning in effect, let us exclude the few instances specifically, instead. Luckily there's not much slicing in the pymeasure codebase.
* Run black on the codebase (without staging/committing). Maybe update it first, maybe you having a different version than me could be responsible for the differences we see?
* Then run `flake8 .` and fix & commit any hits until it runs cleanly.
  - For the one in `control()`, commit the changed indentation and move the noqa comment to the line of the method name, that shoul 
  - If you could not find out why black does not honour the line length, reformulating the docstrings should fix those warnings
  - Add a `# noqa: E203` comment to the instances of E203

bleykauf:
>  The line-too-long things tell me that black apparently does not observe its own configuration settings (line-length = 100).

In these both cases the lines where too long due to a string. I find it interesting, that flake8 does not report an error prior to formatting. If you look at e.g. `dpseriesmotorcontroller.py`:

Before formatting with black:

```python
        """ Reset the position as counted by the motor controller and an externally connected encoder to 0.
        """
```
This doesn't raise E501 even though line-length is over 100.

After formatting with black

```python
        """Reset the position as counted by the motor controller and an externally connected encoder to 0."""
```
This does raise E501. 

I am not sure if this is expected behavior of flake8. [PEP8](https://peps.python.org/pep-0008/#maximum-line-length) actually suggests that docstrings should have a maximum length of 72 characters (even if the line-length is increased but I think this is actually pretty limiting. Interestingly, PEP8 allows to increase the line-length to a maximum if 99 characters while pymeasure uses 100.  But PEP8 [also suggests](https://peps.python.org/pep-0008/#documentation-strings) that for one liner docstrings (such as in the example above) the closing `"""` should be on the same line. [PEP257](https://peps.python.org/pep-0257/) also recommends always using triple double quotes for docstrings but using single quotes would not fix the problem here anyway. PEP257 also requires a one-line summary in any case, so this would have to be changed in any case (or suppressed with `#noqa: E501`). 

The other instance can be fixed by removing the space at the beginning of the string. I see these leading white spaces a lot in pymeasure, is there any reason for this? 

> or is there a black setting that only formats/changes touched lines (so, lines with a git diff), not files?
Formatting parts of files is (and will not be) supported by black, see https://github.com/psf/black/issues/134. This means that if for example a new method is added to an existing instrument the whole instrument class / file will be formatted. I personally did this already in past contributions for consistency's sake within the file. From PEP8: "A style guide is about consistency. Consistency with this style guide is important. Consistency within a project is more important. Consistency within one module or function is the most important." Since we agree that formatting the whole project is not desired at this stage, I think having the same style within the whole file is the next best thing consistency wise. (Of course one could argue that the previous style should be kept...)

bleykauf:
> * [x] Adding a note/comment in the black config that flake8 is the canonical formatter, and takes precedence---

Added the following in https://github.com/pymeasure/pymeasure/pull/683/commits/710fcb3b49a84bd9dacb4914627aca4c9d96f373

"It is allowed but not required to use the `black`_ code formatter. Do only format files that you are actively working on and make a dedicated commit for the initial formatting when working on an existing file that does not adhere to the black code style to facilitate code review. If there are conflicts between black's output and flake8 (especially related to `E203`_), flake8 takes precedence. Use ``#noqa : E203`` to disable E203 warnings for a specific line if appropriate."

@bilderbuchi: Let me know if you are happy with the text or if I should make any changes.

> * [x] Maybe you can find out why black does not honour its own line length setting?

See comment above, has to do with the strings. With one occurence its unclear why it did not raise E501 before, the other one is an easy fix (remove leading white space). In principle, those are edge cases where adhering to one style rule automatically leads to raising of E501. Can be prevented by different string formatting or shortening the docstring as suggested by PEP257.

Fixed via https://github.com/pymeasure/pymeasure/pull/683/commits/09e80204c690c9b0d3599a7148486bb5eb948d8a.

> * [x] Leave the flake8 E203 warning in effect, let us exclude the few instances specifically, instead. Luckily there's not much slicing in the pymeasure codebase.

I also already formatted the slicing so it would raise E203 without the `noqa` even before applying black.

Fixed via https://github.com/pymeasure/pymeasure/pull/683/commits/86ecbbf1e451b0a153bb5edffebe798fae5a48ea.

> * [x] Run black on the codebase (without staging/committing). Maybe update black first, maybe you having a different version than me could be responsible for the differences we see?

Updated to version 22.8.0, had to fix 4 instances of E203.

> * [x] Then run flake8 . and fix & commit any hits until it runs cleanly.

Runs cleanly now, also after applying black.

> * [x] For the one in control(), commit the changed indentation and move the noqa comment to the line of the method name, that shoul

Fixed via https://github.com/pymeasure/pymeasure/pull/683/commits/8d2248b90873ea2317f1f37156b4d0601784254c.

bilderbuchi:
> I find it interesting, that flake8 does not report an error prior to formatting. 

Me, too, I was not aware. I think it [might be](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-max-doc-length) that docstring length is not tracked, but why it raises an error when that becomes a single-line docstring, no idea. :shrug: 

> I see these leading white spaces a lot in pymeasure, is there any reason for this?

No, and I don't agree with leading spaces, either. Someone began this early on, and people are mimicking what's there. I'm not pointing out every instance of this when reviewing sometime 500+ line PR diffs, and we only relatively recently got automated linting. 
Activating a pydocstyle check and displaying those is somewhere on my todo list, but not that urgent. You'll see from #540 that switching flake8 on was quite the amount of review work/changes, and we currently have more urgent fish to fry.

bilderbuchi:
> Let me know if you are happy with the text or if I should make any changes.

LGTM, thanks!

pymeasure/instruments/instrument.py
```diff
@@ -275,19 +275,21 @@ def binary_values(self, command, header_bytes=0, dtype=np.float32):
...
+            values=(),
+            map_values=False,
+            get_process=lambda v: v,
+            set_process=lambda v: v,
+            command_process=lambda c: c,
+            check_set_errors=False,
+            check_get_errors=False,
+            dynamic=False,
+            **kwargs
+    ):
```
bilderbuchi, line 292: This indentation looks weird -- all args should be at the same level. What about this?
```suggestion
    def control(  # noqa: C901 accept that this is a complex method
        get_command,
        set_command,
        docs,
        validator=lambda v, vs: v,
        values=(),
        map_values=False,
        get_process=lambda v: v,
        set_process=lambda v: v,
        command_process=lambda c: c,
        check_set_errors=False,
        check_get_errors=False,
        dynamic=False,
        **kwargs
    ):
```

bilderbuchi (changes requested):
LGTM except for this weird indentation. Thanks!

bilderbuchi:
Thanks!

PR #688, opened 2022-09-10
Compatibility with Qt 6

CasperSchippers:
Updates to make pymeasure compatible with Qt 6 (both PyQt6 and PySide6) while maintaining compatibility with Qt 5 (both PyQt5 and PySide2).

The most changes involve changing the shorthand name of the enum-flags by their full name (I hope to have found all of them).
Also, the getting QApplication for QtGui is deprecated, and I've therefore replaced it by QtWidget.QApplication.

The only issue I have found so far is `QtWidget.QApplication.exec_()` (in `sys.exit(app.exec_())`): in PyQt6 `exec_` has been replaced by `exec`, but PySide2 does not support this (PyQt5 and PySide6 support both `exec_` and `exec`, but whenever `exec_` is used, a deprecation warning is issued. I've solved this now by manually adding `exec` to QtWidgets.QApplication if it is not present, but I'm not sure if this is the best solution.

### DIFF
--- a/pymeasure/display/Qt.py
+++ b/pymeasure/display/Qt.py
@@ -31,6 +31,11 @@ log.addHandler(logging.NullHandler())
 
 QtCore.QSignal = QtCore.Signal
 
+# Should be removed when PySide2 provides QtWidgets.QApplication.exec() or when support for PySide2
+# is dropped
+if not hasattr(QtWidgets.QApplication, 'exec'):
+    QtWidgets.QApplication.exec = QtWidgets.QApplication.exec_
+
 
 def fromUi(*args, **kwargs):
     """ Returns a Qt object constructed using loadUiType
--- a/pymeasure/display/Qt.py
+++ b/pymeasure/display/Qt.py
@@ -29,8 +29,6 @@ from pyqtgraph.Qt import QtGui, QtCore, QtWidgets, loadUiType  # noqa: F401
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
-QtCore.QSignal = QtCore.Signal
-
 # Should be removed when PySide2 provides QtWidgets.QApplication.exec() or when support for PySide2
 # is dropped
 if not hasattr(QtWidgets.QApplication, 'exec'):
--- a/pymeasure/display/Qt.py
+++ b/pymeasure/display/Qt.py
@@ -30,7 +30,7 @@ log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 # Should be removed when PySide2 provides QtWidgets.QApplication.exec() or when support for PySide2
-# is dropped
+# is dropped (https://doc.qt.io/qtforpython/porting_from2.html#class-function-deprecations)
 if not hasattr(QtWidgets.QApplication, 'exec'):
     QtWidgets.QApplication.exec = QtWidgets.QApplication.exec_
 if not hasattr(QtWidgets.QMenu, 'exec'):
--- a/pymeasure/display/Qt.py
+++ b/pymeasure/display/Qt.py
@@ -35,6 +35,10 @@ QtCore.QSignal = QtCore.Signal
 # is dropped
 if not hasattr(QtWidgets.QApplication, 'exec'):
     QtWidgets.QApplication.exec = QtWidgets.QApplication.exec_
+if not hasattr(QtWidgets.QMenu, 'exec'):
+    def exec(self, *args, **kwargs):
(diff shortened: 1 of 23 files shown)


pymeasure/display/Qt.py
```diff
@@ -24,13 +24,18 @@
 
 import logging
 
-from pyqtgraph.Qt import QtGui, QtCore, loadUiType
+from pyqtgraph.Qt import QtGui, QtCore, QtWidgets, loadUiType
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 QtCore.QSignal = QtCore.Signal
```
CasperSchippers, line 32: I came across this line, does anyone know what it's use is? Why not replace all usage in our codebase of `QtCore.QSignal` by `QtCore.Signal` and drop this line?

bilderbuchi: Git blame says that this was changed by @msmttchr in #431 to add pyside2 support. I don't know why, though.

pymeasure/display/Qt.py
```diff
@@ -24,13 +24,18 @@
...
 
+# Should be removed when PySide2 provides QtWidgets.QApplication.exec() or when support for PySide2
+# is dropped
+if not hasattr(QtWidgets.QApplication, 'exec'):
+    QtWidgets.QApplication.exec = QtWidgets.QApplication.exec_
```
bilderbuchi, line 37: that looks like a straightforward patch to me, although it would be good to reference an upstream issue and/or documentation on the background of this to the comment. This would probably make it easier to pick this up/recheck in the future.
I don't know about any potential ramifications of this. I gather we cannot have been the first ones running into the issue, though.

pymeasure/display/Qt.py
msmttchr: I think what suggested by @CasperSchippers is doable.
I think I made this change to keep minimal modifications or simply I didn't realize that this line could have been removed. If I get it right, pyqtgraph provides `QtCore.Signal` for both `PyQt5` and `PySide2`

pymeasure/display/Qt.py
msmttchr: Just googling this issue, you find the explanation quite clear here https://stackoverflow.com/questions/22610720/should-i-use-app-exec-or-app-exec-in-my-pyqt-application.
Essentially, this is related to the fact that `exec` was a reserved keyword in python2 and so people was forced to use `exec_`. This constraint is no longer present in python3 and so `PyQt` people added support for `exec`, but `PySide2` people didn't.

bilderbuchi:
Don't worry about that, most probably this is a kine ending conversion. Maybe it was originally committed with the wrong line endings, and your editor converted them.

CasperSchippers:
Ah that makes sense.

CasperSchippers:
> There is a small issue with the completer of the directory-input that I'm still trying to sort out: at this moment, the completer only completes the path towards the current working directory.

Fixed this issue, so the completer works, but when the gui is stared, the completer is only triggered on the second time you try to activate it.

msmttchr:
@CasperSchippers, I have suggested some changes in a PR in your fork, probably you have noticed it, but just in case.
Anyway feel free to accept or cherrypeek what is relevant in your opinion

bicarlsen:
Wanted to comment that I just tested this out with my project and everything seems to be working well with PyQt6. Thanks for all the work.

CasperSchippers:
thanks @msmttchr, I merged your suggestions!

pymeasure/display/Qt.py
CasperSchippers: Thanks, I removed the line

pymeasure/display/Qt.py
CasperSchippers: I've added a link to the comment, though I could not find an upstream issue as I have the impression that this is not something that is considered an "issue" in PySide2

CasperSchippers:
I think it can be merged (when @msmttchr and/or @bilderbuchi agree)

msmttchr:
@CasperSchippers, I have another small changes related to `PyQt6` on a line that you did not change/update and as such I cannot add the comment inline in the code. I am attaching a patch file for it.
[windows-patch.zip](https://github.com/pymeasure/pymeasure/files/9739282/windows-patch.zip)

CasperSchippers:
@msmttchr, thanks, good one!

msmttchr (approved):
LGTM. Thanks.

bilderbuchi:
Agreed! Let's merge it, then. :-)

PR #696, opened 2022-09-29
Fix for gui_foreign_instrument

msmttchr:
I have discovered by chance that this example is not running and I propose these fixes.


bilderbuchi:
As the comment above the imports indicates, this is supposed to demonstrate how pymeasure also works with other instrument driver libraries, like instrumentkit. 😉

msmttchr:
> As the comment above the imports and the docstring indicates, this is supposed to demonstrate how pymeasure also works with other instrument driver libraries, like instrumentkit. wink Did you have instrumentkit installed?

@bilderbuchi Indeed you are right, I was running the script without paying attention to the comment.
Now, I understand the logic. This PR does not make sense and so I will close it.

PR #698, opened 2022-10-03
New instrument: TEXIO PSW-360L30 Power Supply (minimal implementation)

LastStarDust:
I based this power supply on the Keithley2260B one because the SCPI commands are very similar. 
Actually the API is so similar that only minor changes were necessary.

I did not test with the actual hardware because I do not have a device available today. 
But I will test with the device tomorrow. In the meanwhile you can take a look at the driver and its protocol unit tests.


bilderbuchi:
Haven't reviewed yet, but if the commands and API are quite similar, do you want to try to inherit from the Keithley and tweak as needed? You can set properties in the Keithley do `dynamic=True` as needed for tweaks.

bmoneke:
Having written these tests, refactoring the code is more safe, because you know, that it works the same.
Thanks for the tests!

LastStarDust:
Ok I will inherit from the Keithley model and tweak as necessary

LastStarDust:
@bilderbuchi @bmoneke Now the TEXIO PSW-360L30 power supply Inherits from the Keithley 2260B one. It avoids a lot of code duplication. Thank you for the suggestion. I did not need to modify anything in other classes as well.

pymeasure/instruments/texio/texioPSW360L30.py
```diff
@@ -0,0 +1,54 @@
...
+
+        source = TexioPSW360L30("TCPIP::xxx.xxx.xxx.xxx::2268::SOCKET")
+        source.voltage = 1
+        print(source.voltage)
+        print(source.current)
+        print(source.power)
+        print(source.applied)
+    """
+
+    def __init__(self, adapter, read_termination="\n", **kwargs):
```
bmoneke, line 37: According to your docstring, the read termination is fix "\n": why do you have such a parameter?

If you want to allow the user to change the termination, for some edge cases or subclassing, but not for normal use, I recommend to do it via the kwargs: remove it from the init parameters and add as the first line `kwargs.setdefault('read_termination', "\n").
You would have to remove the parameter from your call to Instrument.init as well.

What this code does, is to inject the your default to the kwargs, if such a key is not present, otherwise it keeps the user's choice.

LastStarDust: You are right. The read termination is fixed so there is no point in letting the user fiddle with it. I removed the parameter altogether.

LastStarDust:
I have also written unit tests with hardware attached. The driver is complete and can be merged whenever you like.

bilderbuchi:
> The problem is that I need to comment it to be able to run unit tests from PyCharm

You should actually be able to use `pytest --no-skip -k Texio` to ignore the skipping. ;-)

pymeasure/instruments/texio/texioPSW360L30.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    def __init__(self, adapter, **kwargs):
+        Instrument.__init__(self,
+                            adapter=adapter,
+                            name="TEXIO PSW-360L30 Power Supply",
+                            read_termination="\n",
+                            **kwargs
+                            )
+
```
bilderbuchi, line 44: Ha, that does mean that the instruction set (except error checking) is exactly the same? :-)

bilderbuchi (changes requested):
One small request, otherwise this looks good. Are the instructions really the same?

LastStarDust:
> One small request, otherwise this looks good. Are the instructions really the same?

Yes instructions are really the same. Checked with actual hardware too.

bmoneke:
Nice. I like the addition of a new instrument with a small adjustment.

pymeasure/instruments/texio/texioPSW360L30.py
LastStarDust: Done in commit a95e7fc4a4d96db3ae1cca1c9d4083a8c07c4092

pymeasure/instruments/texio/texioPSW360L30.py
```diff
@@ -0,0 +1,71 @@
...
+    @property
+    def output(self):
+        """Same as "enabled" property"""
+        return self.enabled
+
+    @output.setter
+    def output(self, value):
+        """Same as "enabled" property"""
+        self.enabled = value
```
bilderbuchi, line 71: You know what, I fully agree with you re: the bad naming for `enabled`. An even better one than `output` is `output_enabled`. :D 
I have a proposition:
* In the Keithley2260B implementation, rename `enabled` to `output_enabled`
* Adapt that in the `shutdown` method, too.
* Immediately below the new `output_enabled`, add this
```
    @property
    def enabled(self):
        warn("Deprecated property name "enabled", use the identical "output_enabled", instead.",
             FutureWarning)
        return self.enabled

    @output.setter
    def enabled(self, value):
        warn("Deprecated property name "enabled", use the identical "output_enabled", instead.",
             FutureWarning)
        self.enabled = value
```
* Add a note in the CHANGES.rst in the deprecated features section about the preferred name for Keithley2260B.enabled.
* Drop this here as it is now unnecessary :party:
What do you think?

LastStarDust: @bilderbuchi I fully agree with all your suggestions. I only have one question about the `shutdown` method. There are scenarios where you want to keep the power supply on, even if the software is not running.  How to deal with it?

LastStarDust:
@bilderbuchi I do not see a deprecated features section in the CHANGES.rst file in my branch.

pymeasure/instruments/texio/texioPSW360L30.py
LastStarDust: @bilderbuchi I do not see a deprecated features section in the CHANGES.rst file in my branch.

bmoneke: > I fully agree with all your suggestions. I only have one question about the `shutdown` method. There are scenarios where you want to keep the power supply on, even if the software is not running. How to deal with it?

I understand that reason. Sometimes I want to close the program, but keep the source enabled. You do not need to call `shutdown`. It's only called automatically if you use the instrument in a context manager. Otherwise you have to call `shutdown` manually.


> I do not see a deprecated features section in the CHANGES.rst file in my branch.

Then simply add such a section.

bilderbuchi: Ah, you missed this in the first bullet, I think: "In the Keithley2260B implementation,..." - those changes should go in that instrument's file, not the texio... :-P

LastStarDust: Yes, I misunderstood that. Now everything should fixed in commit 16a9c6fd88635f98a7d191e012ca0825f93a1cd9

CHANGES.rst
```diff
@@ -1,3 +1,18 @@
+Version 0.XX.X (20XX-XX-XX)
+===========================
+
+Instruments
+-----------
+- New TEXIO PSW-360L30 Power Supply (minimal implementation)
+
+Deprecated Features
+-------------------
```
bmoneke, line 9: Now I know, why you did not find a "deprecated" section: It was added between your start of that pull request and the current state. Therefore the merge conflict.

Please merge the master branch and solve the conflict.

pymeasure/instruments/texio/texioPSW360L30.py
```diff
@@ -0,0 +1,61 @@
...
+        print(source.applied)
+    """
+
+    def __init__(self, adapter, **kwargs):
+        Instrument.__init__(self,
+                            adapter=adapter,
+                            name="TEXIO PSW-360L30 Power Supply",
+                            read_termination="\n",
+                            **kwargs
+                            )
```
bmoneke, line 50: If you override a method and call the overriden method, use `super()` to access that one:
```suggestion
        super().__init__(self,
                         adapter=adapter,
                         name="TEXIO PSW-360L30 Power Supply",
                         read_termination="\n",
                         **kwargs
                         )
```

bmoneke (changes requested):
Two small comments.

pymeasure/instruments/texio/texioPSW360L30.py
LastStarDust: That is not going to work because both `Keithley2260B.__init__` and `TexioPSW360L30.__init__` whould try to call `Instrument.__init__` with different arguments due to line 35 of keythley2260B: `**kwargs`

CHANGES.rst
LastStarDust: I merged the master branch from the pymeasure repository and edited the CHANGES.rst file.

pymeasure/instruments/texio/texioPSW360L30.py
bmoneke: The problem is `name`:
super().__init__ calls the Keithley init, which puts "name" into the kwargs (as it does not know a "name" parameter), then it calls Instrument.init with the name argument twice (once from Keithley init code and once from the kwargs), which raises an Exception.
There are a few options to circumvent this problem:

- Add `kwargs.setdefault('name', "Keithley 2260B DC Power Supply")` in Keithley's init before it calls super and remove the `name=Keithley` part from the super().__init parameters. That way the instrument is called Keithley unless you give it a name via kwarg (as you do it here). (My recommendation)
- call super().__init__ in the Texio init, but without a name parameter. Set `self.name = "Texio...."` after calling super().__init__
- You could call `super(Keithley2260B, self).__init__(`, because that way you get the parent of the Keithley class.

LastStarDust: I tried to follow your recommendation. The problem is not only the name attribute but also the positional arguments, like `adapter`.

bilderbuchi: > You could call super(Keithley2260B, self).__init__(, because that way you get the parent of the Keithley class.

In general:  Don't try to manually override the Method Resolution Order like this. If there's something wonky here, we have to fix it properly so the MRO is set up properly, otherwise, we could miss e.g. some needed init code.

@LastStarDust I don't see off the cuff what the problem with positional arguments would be. What erroneous behaviour did you observe?

LastStarDust: I was trying to pass `adapter` as a keyword argument but it did not work because of multiple definitions.
If I pass it as a positional argument it does work (see the latest commit).

bmoneke:
A last check: @LastStarDust , are you ready to merge?

LastStarDust:
> A last check: @LastStarDust , are you ready to merge?

Yes you can go on.

bmoneke:
Thanks for your contribution!

PR #701, opened 2022-10-06
New instrument: Eurotest hpp_120_256_ieee

sansanda:
New Instrument: 
Brand: Eurotest (also known as iSeg)
Type: High Voltage Supply
Model: 120_256_ieee


pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_range
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] %(message)s",
+    handlers=[
+        logging.StreamHandler()
+    ]
+)
```
LongnoseRob, line 38: I am not quite sure if this call is needed in this form,
I usally stick with the form of 
```
```
This seems to be widly used in instrument implementations

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+        "STATUS,MI",
+        """ Measures the actual output current of the power supply in mAmps.
+        This property is a get so will return a string like this:
+        I, RANGE=5000mA, VALUE=1739mA, then measure_current will
+        return a tuple (5000.0, 1739.0) hence the convenience of the get_process.""",
+        get_process=lambda v: (
+            float(re.findall(r'[-+]?([0-9]*\.[0-9]+|[0-9]+)', v[1])[0]),
+            float(re.findall(r'[-+]?([0-9]*\.[0-9]+|[0-9]+)', v[2])[0])
+        )
+    )
```
LongnoseRob, line 174: For me the use of tuples as return values for these properties does not match the idea of using properties.
Why not use more clearer properties like 

- `voltage` (sets the U-setpoint, terunts the actual voltage)
- `current` (sets the I-setpoint, retruns the actual current)
- `current_limit` (..)
- `voltage limit` (..)
- `voltage_ramp` (..)
- `voltage_range` (read-only ?)
- `current_range` (read-only ?)

The doc-strings are intended to give an explanation about the property they describe and do no need to cover details about the implmentation

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+
+    @property
+    def id(self):
+        """ Returns the identification of the instrument """
+        logging.info(f"Requesting instrument identification...")
```
LongnoseRob: Is this logging entry necessary? 
If there is no vaiable data to be put in the log, a normal string would rpobably also work fine.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+
+        response = self.ask("ID")
+        return response.encode(self.response_encoding).decode('utf-8', 'ignore')
+
+    def status(self):
+        """ Returns the unit Status which is a 16bits response where
+        every bit indicates teh state of one subsystem of the HV Source."""
+        logging.info(f"Requesting instrument status...")
```
LongnoseRob: same as above

bmoneke:
Nice, a new instrument, and already including tests, great!

From the formal side, you miss a few points: Unfortunately, there are two authors files, one in docs (which you changed) and authors.txt in the main folder. Second, instruments/__init__.py should load your eurotest/__init__.py.

What's the fuss about the pip-wheel metadata?

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+    def __init__(self, adapter, response_encoding="iso-8859-2", query_delay=0.1, **kwargs):
+        self.response_encoding = response_encoding  # This instrument use this encoding, so we have to take into account
+        self.query_delay = query_delay  # Delay in s to sleep between the write and read occuring in a query
+        adapter.connection.encoding = self.response_encoding  # comment it if you want to run the test
+        adapter.connection.query_delay = self.query_delay  # comment it if you want to run the test
+        super().__init__(
+            adapter,
+            "Euro Test High Voltage DC Source model HPP-120-256",
+            **kwargs
+        )
```
LongnoseRob: By convention this should be the first entry after the doc-string in the class.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+        time.sleep(self.COMMAND_DELAY)
+
+    def disable_kill(self):
+        """ Disables the kill function of the HV source.
+         When Kill is disabled yellow led is not flashing and the output
+         will NOT be shut OFF permanently if Iout > IOUTmax."""
+        logging.info(f"Configuring instrument kill as disable...")
+
+        self.write("KILL,DISable")
+        time.sleep(self.COMMAND_DELAY)
```
bmoneke, line 235: It's ok to use methods in these cases, however, you could use a property as well, e.g. output = True/False and kill_mode = True/False (or ON/OFF...)

LongnoseRob (commented):
Welcome to pymeasure and thank you for the PR.

I took the liberty and added some comments, please do not take them personally.

Is there a public avaiable manualor programming reference for this product?

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Done in commit. Many thanks.

pymeasure/instruments/eurotest/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
```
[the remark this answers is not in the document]
sansanda: Done. Thank you.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+        voltage could response one of the messages from the next list:
+            LAM,ERROR External Inhibit occurred during Kill enable
+            LAM,INHIBIT External Inhibit occurred
+            LAM,TRIP ERROR Software current trip occurred
+            LAM,INPUT ERROR Wrong command received
+            LAM,OK Status OK"""
+        logging.info(f"Requesting instrument LAM status...")
+
+        response = self.ask("STATUS,LAM")
+        return response.encode(self.response_encoding).decode('utf-8', 'ignore')
```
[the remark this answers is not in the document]
sansanda: Yeahhh. Is a good information for debugging level. 
Done in commit.
Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+
+        self.write("HV,ON")
+        time.sleep(self.COMMAND_DELAY)
+
+    def disable_output(self):
+        """ Disables the output of the HV source. """
+        logging.info(f"Disabling the output...")
+
+        self.write("HV,OFF")
+        time.sleep(self.COMMAND_DELAY)
```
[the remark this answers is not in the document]
sansanda: Yes. As bmoneke suggest I've improve the code using properties (Instrument.settting). Many thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+        self.write("KILL,DISable")
+        time.sleep(self.COMMAND_DELAY)
+
+    def emergency_off(self):
+        """ The output of the HV source will be switched OFF permanently and the values
+        of the voltage a current settings set to zero"""
+        logging.info(f"Sending emergency off command to the instrument.")
+
+        self.write("EMCY OFF")
+        time.sleep(self.COMMAND_DELAY)
```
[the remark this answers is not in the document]
sansanda: Yes. Done in commit as the comment above. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes. But I wanted to modify the adapter before passing it to the constructor of the Superclass. What do you suggest I could resolve it?
Many thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Ok. Done

pip-wheel-metadata/PyMeasure-0.10.1.dev159+gd6b1ef4.dist-info/LICENSE.txt
```diff
@@ -0,0 +1,19 @@
...
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
```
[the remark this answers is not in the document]
sansanda: You mean that I should delete LICENSE.txt?

On the other hand, you can find the manual of the HV suplly [here](https://iseg-hv.com/download/AC_DC/HPS/Manual_HPS_300W_800W_eng.pdf). Our HV is older but it works with the same ET command set.
Thanks.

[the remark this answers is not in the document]
LongnoseRob: > You mean that I should delete LICENSE.txt?

Some files have somehow ended up in your pull request that are not needed for adding the instrument.
like this License-file and some other metadata files.

pip-wheel-metadata/PyMeasure-0.10.1.dev159+gd6b1ef4.dist-info/METADATA
```diff
@@ -0,0 +1,387 @@
+Metadata-Version: 2.1
```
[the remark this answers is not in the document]
sansanda: Same answer as above. Yes?
Thanks

tests/instruments/eurotest/test_eurotestHPP120256.py
```diff
@@ -0,0 +1,73 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+import pytest
```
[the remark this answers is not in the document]
sansanda: Many thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
bmoneke: I agree with LongnoseRob in all points.

- If you ask the voltage (inst.voltage), you expect the current voltage and not a tuple.
- The user does not need to know, what string the device sends. You might keep the device responses as a comment for maintainers.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+    def id(self):
+        """ Returns the identification of the instrument """
+        log.info("Requesting instrument identification...")
+
+        response = self.ask("ID")
+        return response.encode(self.response_encoding).decode('utf-8', 'ignore')
+
+    def status(self):
+        """ Returns the unit Status which is a 16bits response where
+        every bit indicates teh state of one subsystem of the HV Source."""
```
bmoneke, line 205:
```suggestion
        every bit indicates the state of one subsystem of the HV Source."""
```

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+        of the voltage a current settings set to zero"""
+        log.info("Sending emergency off command to the instrument.")
+
+        self.write("EMCY OFF")
+        time.sleep(self.COMMAND_DELAY)
+
+    def shutdown(self, ramp):
+        """
+        Ramps the HV source to zero with a determinated ramp and without waiting
+        to the output reaches zero volts dissables it.
```
bmoneke, line 240: I do not understand the comment: "without waiting to the output reaches zero volts disables it" (btw: disables has a single 's' in the middle).

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+        """
+        log.info(f"Executing the shutdown function with ramp: {ramp} V/s.")
+
+        self.ramp_to_zero(ramp)
+        super().shutdown()
+
+    def ramp_to_zero(self, ramp):
+        """
+        Ramps the HV source to zero with a determinated ramp
+        :param ramp: indicates the ramp we want
```
bmoneke, line 251: What ramps do we have? Is it the slope, or are there different types?

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+        log.info("Executing the ramp_to_zero function with ramp: {ramp} V/s.")
+
+        self.voltage_ramp = ramp
+        time.sleep(self.COMMAND_DELAY)
+        self.voltage = 0
+        time.sleep(self.COMMAND_DELAY)
+        self.voltage = 0
+        time.sleep(self.COMMAND_DELAY)
+
+    def wait_for_voltage_output_set(self, check_period, timeout):
```
bmoneke, line 262: I recommend to add some sensible default values for check_period and timeout. That way a user may just call wait_for.. without supplying further values, unless needed.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+    # Constructor
+    def __init__(self, adapter, response_encoding="iso-8859-2", query_delay=0.1, **kwargs):
+        # This instrument use this encoding, so we have to take into account
+        self.response_encoding = response_encoding
+        # Delay in s to sleep between the write and read occuring in a query
+        self.query_delay = query_delay
+        # comment next line if you want to pass the test
+        # adapter.connection.encoding = self.response_encoding
+        # comment next line if you want to pass the test
+        # adapter.connection.query_delay = self.query_delay
```
bmoneke, line 315: You commented them already. Which test do you want to pass?

bmoneke (commented):
Thanks for your contribution and well documented code. Some documentation (about the communication protocol) is not necessary for the user and can be kept a comment instead of the doc-string.

pip-wheel-metadata/PyMeasure-0.10.1.dev159+gd6b1ef4.dist-info/LICENSE.txt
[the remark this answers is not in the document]
sansanda: Ok

pymeasure/instruments/eurotest/eurotestHPP120256.py
LongnoseRob: Here it might be good to use an enum to decode the status bits data

pymeasure/instruments/eurotest/eurotestHPP120256.py
LongnoseRob: As mentioned [here](https://github.com/pymeasure/pymeasure/pull/701#discussion_r991053525), unless you have major work to do with the adapter, I would just put the termination settings as defaults in the  `___init___`block:
```
        kwargs.setdefault('read_termination', '\r\n')
        kwargs.setdefault('write_termination', '')
        kwargs.setdefault('send_end', True)
        super().__init__(
            adapter,
            "Euro Test High Voltage DC Source model HPP-120-256",
            **kwargs,
        )
```

similiar could be done for querydelay etc.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+                          write_termination="\n",
+                          read_termination="",
+                          send_end=True)
+
+    adapter.connection.timeout = 5000
+    # In my case, instrument uses this encoding on response, so take it into account
+    response_encoding = "iso-8859-2"
+    query_delay = 0.4  # Delay in s to sleep between the write and read occuring in a query
+    command_delay = 0.2
+    hpp120256 = EurotestHPP120256(adapter, response_encoding, query_delay, includeSCPI=False)
```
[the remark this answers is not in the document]
bmoneke: See the documentation about [Defining default connection settings](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#defining-default-connection-settings)

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes, you are right.
done in commit. I hope now it is clearer.

bmoneke: Yes, now I get it. It is not necessary to double the information in the general description and in the parameter description, but not I understand how to use the method.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes. But it doesn't work for querydelay and encoding. 

For example, next code doesn't work for enconding:

    def __init__(self, adapter, response_encoding="iso-8859-2", query_delay=0.1, **kwargs):
        # This instrument use encoding iso-8859-2 on response, so we have to take it into account        
        kwargs.setdefault('encoding', response_encoding)
        super().__init__(
            adapter,
            "Euro Test High Voltage DC Source model HPP-120-256",
            **kwargs
        )

But this other works:

    def __init__(self, adapter, response_encoding="iso-8859-2", query_delay=0.1, **kwargs):

        # This instrument uses iso-8859-2 encoding, so we have to take it into account
        # Also query_delay has to be taken into account for slow response instruments
        self.response_encoding = response_encoding
        self.query_delay = query_delay
        adapter.connection.encoding = self.response_encoding
        adapter.connection.query_delay = self.query_delay
        super().__init__(
            adapter,
            "Euro Test High Voltage DC Source model HPP-120-256",
            **kwargs
        )

Any suggestion will be apreciate. Thanks.

bmoneke: According to the pyvisa documentation and source code, the kwargs are executed afterwards and just do, what you did: setting the encoding and query_delay.

You could add it **after** the super.init:
```
super().__init__(...)
if isinstance(adapter, VISAAdapter):
    self.adapter.connection.encoding = xy
    self.adapter.connection.query_delay = query_delay
```

The if question is necessary, as other adapters do not have an encoding or query_delay.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: It is a changing rate of the voltage output (Volts/second). Named as ramp on the instrument manual. I've rewrite the docstring for better explanation. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Done. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: I agree with that LongnoseRob. I will annotate it as TODO for the next version. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
bilderbuchi: `ramp_rate` maybe a better name for that argument?

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,326 @@
...
+
+        # if you are here is because the voltage_setting
+        # has been reaches at the output of the HV Voltage source
+
+    # Constructor
+    def __init__(self, adapter, response_encoding="iso-8859-2", query_delay=0.1, **kwargs):
+
+        # This instrument uses iso-8859-2 encoding, so we have to take it into account
+        # Also query_delay has to be taken into account for slow response instruments
+        self.response_encoding = response_encoding
```
msmttchr, line 318: As mentioned above by @bmoneke, normally you don't need to assume that `adapter` is passed as an instance of `Adapter` class, but you can also pass a string and a `VISAAdapter` is created for you.
As such, you would need to:
- add the parameter `SCPI=False` to `super` call
- pass all `kwargs` + `encoding` and `query_delay` to super

pymeasure/instruments/eurotest/eurotestHPP120256.py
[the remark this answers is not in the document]
sansanda: Adapted to last commits done by bmoneke and done in commit. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Thanks msmttchr. Now is adapted to last commits on pymeasure:master done bmoneke.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+        voltage_output = self.measure_voltage[1]
+        voltage_output_set = (voltage_output > (voltage_output_setting - error)) and \
+                             (voltage_output < (voltage_output_setting + error))
+
+        while not voltage_output_set:
+            actual_time = time.time()
+            log.info(f"\tWaiting for voltage output set. "
+                     f"Reading output voltage every {check_period} seconds.\n"
+                     f"\tTimeout: {timeout} seconds. Elapsed time: "
+                     f"{round(actual_time - ref_time, ndigits=1)} seconds.")
```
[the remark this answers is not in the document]
sansanda: You are right. Done in commit. Thanks

pip-wheel-metadata/PyMeasure-0.10.1.dev159+gd6b1ef4.dist-info/METADATA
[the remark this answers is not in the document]
sansanda: Done in commit. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Ok. Done in commit. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: I think now the code follow your recomendations. Many thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Hi. I wanted to put in log just to dissable info messages in normal operation. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,320 @@
...
+
+    """
+
+    VOLTAGE_RANGE = [0.0, 12.0]  # kVolts
+    CURRENT_RANGE = [0.0, 25.0]  # mAmps
+    VOLTAGE_RAMP_RANGE = [10, 3000]  # V/s
+    COMMAND_DELAY = 0.2  # s
+
+    # ####################################
+    # # ET-Command set. Non SCPI commands.
```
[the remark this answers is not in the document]
sansanda: I don't understand exactly what you mean. Sorry.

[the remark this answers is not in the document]
bmoneke: I guess "ET" means "EuroTest", doesn't it? Therefore:
```suggestion
    # # EuroTest-Command set. Non SCPI commands.
```

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Code have changed. Now those have been removed. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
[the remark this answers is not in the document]
sansanda: Now I see. Donde in commit. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+    # Constructor
+    def __init__(self,
+                 adapter,
+                 response_encoding="iso-8859-2",
+                 query_delay=0.2,
+                 timeout=5000,
+                 write_termination="\n",
+                 read_termination="",
+                 send_end=True,
+                 includeSCPI=False,
```
bmoneke, line 319: Can you change these options (read/write_termination, send_end, includeSCPI) on the device? If the device always uses these configuration options (I guess you cannot activate SCPI, for example), I would remove them from the init parameters and just call them in super init (e.g. ` super().__init__(includeSCPI=False)`).
It seems unnecessary to offer options to the user (e.g. write_termination), which he cannot use, as the device always needs a specific value ("\n" here).

Is it the same with the response_encoding?

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+    # # EuroTest-Command set. Non SCPI commands.
+    # ####################################
+
+    voltage = Instrument.control(
+        "STATUS,U", "U,%.3fkV",
+        """ A floating point property that represents the output voltage
+        setting (in kV) of the HV Source in kVolts. This property can be set.
+        When this property acts as get, the instrument will return a string like this:
+        U, RANGE=3.000kV, VALUE=2.458kV, then voltage will return a tuple
+        2.458 hence, the convenience of the get_process.""",
```
bmoneke, line 101: I'd keep the string as a comment and not in the doc_string, because it offers no additional information to the user, only to developers. Similarly to all the others. That is my personal opinion.
```suggestion
        """ A floating point property that represents the output voltage
        setting (in kV) of the HV Source in kVolts. This property can be set.""",
        # getter device response: "U, RANGE=3.000kV, VALUE=2.458kV"
```

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+        validator=strict_range,
+        values=VOLTAGE_RANGE,
+        get_process=lambda v: float(re.findall(r'[-+]?([0-9]*\.[0-9]+|[0-9]+)', v[2])[0])
```
bmoneke, line 104: why don't you look for `r'VALUE=([+-]?[\d\.]+)'`? That way, you get the right part and the regex is smaller and more concise.

Btw, looking at that expression, you do not include the plus/minus in the float, such that `"VALUES=-123"` returns `123` instead of `-123`.

bmoneke:
The code looks good to me, with exception of the regex dropping a minus sign. And I added two suggestions how to change the code.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes. Ok. Many thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Answering your questions.

read/write_termination and send_end cannot be changed. So As you mentioned, It would be better to remove it from the constructor.  

On the other hand, the instrument has two command sets: The Eurotest command set and the SCPI command set and they are mutually exclusive. For the moment I'm implemented only the ET command set but i would like to leave the door open to switch (in the future) from one to another. 

For the response_encoding I think is totally internal to the instrument, so I will remove it from the constructor.

Many thanks

bmoneke: Good to know about those two command sets. I know the same thing for another instrument.
In those cases, I think it is best to make to different instruments, if you want to have both command sets. Just state it in the device description. 
So I recommend just to do the one, you are going to use (ET in this case).

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes. Pretty good. I will use it. But not in such fashion. Below my explanation:
What I found when the instrument response, for example, to voltage is an array of strings not a string like this "U, RANGE=3.000kV, VALUE=2.458kV". The instrument response this list ["U", "RANGE=3.000kV", "VALUE=2.458kV"]. Thats the reason I use re.findall(r'[-+]?([0-9]*\.[0-9]+|[0-9]+)', v[2]) over v[2]. 
I will use re.findall(r'([+-]?[\d\.]+)', v[2]) instead.
Also I will rewrite the docstring replacing "U, RANGE=3.000kV, VALUE=2.458kV" for ["U", "RANGE=3.000kV", "VALUE=2.458kV"]

Note: Sincerely I'm not sure why I obtain a list ( ["U", "RANGE=3.000kV", "VALUE=2.458kV"]) instead of an string like this "U, RANGE=3.000kV, VALUE=2.458kV". Perhaps it is an internal transformation of pymeasure? What do you think?

Many thanks

bmoneke: The pymeasure properties call `Instrument.values`, which does splitting the string into several (in order to read several values), default is at ",". 
If you give it `separator=None` (also applicable for properties), it won't split at all.

Or you might use it for your own advantage.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Done in commit. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+            "Euro Test High Voltage DC Source model HPP-120-256",
+            write_termination="\n",
+            read_termination="",
+            send_end=True,
+            includeSCPI=includeSCPI,
+            **kwargs
+        )
+
+        self.query_delay = query_delay
+        self.timeout = timeout
```
bmoneke, line 334: For what do you use the property timeout? Do you want to have a connection timeout? If so, you should add it to the super constructor.

As your implementation assumes the instrument to be in non Scpi mode, you should not add this as a parameter, but just add it to the super constructor. (don't forget to adapt the class docstring)

bmoneke:
I did not find any rough issues (just some points with a different taste), besides the minor issue in the constructor.

Afterwards a maintainer like @bilderbuchi or @msmttchr has to merge the code.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+
+    """
+
+    VOLTAGE_RANGE = [0.0, 12.0]  # kVolts
+    CURRENT_RANGE = [0.0, 25.0]  # mAmps
+    VOLTAGE_RAMP_RANGE = [10, 3000]  # V/s
+    COMMAND_DELAY = 0.2  # s
+
+    response_encoding = "iso-8859-2"
+    f_numbers_regex_pattern = r'([+-]?[\d\.]+)'
```
msmttchr, line 91:
```suggestion
    f_numbers_regex_pattern = r'([+-]?([\d]*\.)?[\d]+)'
```
I think this is more accurate matching (not tested though). Do you agree ?

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+        """ Overrides Instrument ask method for including query_delay time on parent call.
+        :param command: Command string to be sent to the instrument.
+        :returns: String returned by the device without read_termination.
+        """
+        return super().ask(command, self.query_delay)
+
+    # Constructor
+    def __init__(self,
+                 adapter,
+                 includeSCPI=False,
```
msmttchr, line 318:
```suggestion
```

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+                 timeout=5000,
+                 **kwargs):
+
+        super().__init__(
+            adapter,
+            "Euro Test High Voltage DC Source model HPP-120-256",
+            write_termination="\n",
+            read_termination="",
+            send_end=True,
+            includeSCPI=includeSCPI,
```
msmttchr, line 329:
```suggestion
            includeSCPI=False,
```
I think includeSCPI can be safely hardcoded to False.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+from pymeasure.instruments.validators import strict_range
+from pymeasure.instruments.validators import strict_discrete_set
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class EurotestHPP120256(Instrument):
+    """ Represents the Euro Test High Voltage DC Source model HPP-120-256
+    and provides a high-level for interacting with the instrument using the
```
msmttchr, line 39:
```suggestion
    and provides a high-level interface for interacting with the instrument using the
```
Not an English expert, but I feel this is more clear.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+    Euro Test command set (Not SCPI command set).
+
+    .. code-block:: python
+
+    hpp120256 = EurotestHPP120256("GPIB0::20::INSTR",includeSCPI=False)
```
msmttchr, line 44:
```suggestion
    hpp120256 = EurotestHPP120256("GPIB0::20::INSTR")
```
See also comment in `__init__`

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+        float(re.findall(EurotestHPP120256.f_numbers_regex_pattern, v[1])[0])
+    )
+
+    enable_kill = Instrument.setting(
+        "KILL,%s",
+        """ Enables or disables the kill function of the HV source.
+         When Kill is enabled yellow led is flashing and the output
+         will be shut OFF permanently without ramp if Iout > IOUTmax.""",
+        validator=strict_discrete_set,
+        values={'ON': 'ENable', 'OFF': 'DISable', 'ENable': 'ENable', 'DISable': 'DISable',
```
msmttchr, line 180: I think it would be easier to leave only "ON"/"OFF" or better True/False choice to users.
All the other combinations are redundant.
Similar comments to other cases below.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+
+    response = hpp120256.id
+    print(response)
+    print(hpp120256.lam_status)
+    print(hpp120256.status)
+
+    hpp120256.ramp_to_zero(100.0)
+
+    hpp120256.voltage_ramp = 50.0  # V/s
+    time.sleep(command_delay)
```
msmttchr, line 54: It seems to me that this delay is needed after sending any command.
A possible solution, more user friendly, could be that you record the time when last `write` was executed and when a new `write` is requested, you check that at least command_delay has expired, otherwise wait before executing the write.
Of course, this would require to override `write` method

msmttchr (changes requested):
Thanks for your contribution. Few comments from me.

pymeasure/instruments/eurotest/eurotestHPP120256.py
bilderbuchi: Agreed. A common pattern is to name such properties `kill_enabled` and use (only) True/False as the interface towards the users.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+        error = 0.02  # error of +-ten volts (0.01kV) to enter into the voltage stable window
+        voltage_output = self.measure_voltage
+        voltage_output_set = (voltage_output > (voltage_output_setting - error)) and \
+                             (voltage_output < (voltage_output_setting + error))
+
+        while not voltage_output_set:
+            actual_time = time.time()
+            log.debug(f"\tWaiting for voltage output set. "
+                      f"Reading output voltage every {check_period} seconds.\n"
+                      f"\tTimeout: {timeout} seconds. Elapsed time: "
```
bilderbuchi, line 291: The first three lines should probably only be emitted once, before the loop starts. Their info does not change.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+
+        ref_time = time.time()
+        future_time = ref_time + timeout
+
+        voltage_output_setting = self.voltage
+        time.sleep(self.COMMAND_DELAY)
+        error = 0.02  # error of +-ten volts (0.01kV) to enter into the voltage stable window
+        voltage_output = self.measure_voltage
+        voltage_output_set = (voltage_output > (voltage_output_setting - error)) and \
+                             (voltage_output < (voltage_output_setting + error))
```
bilderbuchi, line 285:
```suggestion
        error = 0.02  # error of +-ten volts (0.01kV) to enter into the voltage stable window
        voltage_output = self.measure_voltage
        voltage_output_set = math.isclose(voltage_output, voltage_output_setting, reltol=0.0, atol=error)
```
Same below, use Python's batteries :-)
`voltage_output_reached` might be a clearer variable name.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+                      f"\tTimeout: {timeout} seconds. Elapsed time: "
+                      f"{round(actual_time - ref_time, ndigits=1)} seconds.")
+
+            time.sleep(check_period)  # wait for voltage output reaches the voltage output setting
+            voltage_output = self.measure_voltage
+            voltage_output_set = (voltage_output > (voltage_output_setting - error)) and \
+                                 (voltage_output < (voltage_output_setting + error))
+            log.debug("voltage_output_valid_range: "
+                      "[" + str(voltage_output_setting - error) +
+                      ", " + str(voltage_output_setting + error) + "]")
```
bilderbuchi, line 300: as above, why emit this every iteration?

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+            log.debug("voltage_output: " + str(voltage_output))
+            if actual_time > future_time:
+                raise Exception("Timeout for wait_for_voltage_output_set function")
+
+        # if you are here is because the voltage_setting
+        # has been reaches at the output of the HV Voltage source
```
bilderbuchi, line 306:
```suggestion
        return
```

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def ask(self, command):
+        """ Overrides Instrument ask method for including query_delay time on parent call.
+        :param command: Command string to be sent to the instrument.
+        :returns: String returned by the device without read_termination.
+        """
+        return super().ask(command, self.query_delay)
+
+    # Constructor
+    def __init__(self,
```
bilderbuchi, line 316: please move the ctor to the beginning of the class.

bilderbuchi (changes requested):

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes yes. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes. Actual pattern done as good strings like ....1. I agree.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+        # This property is a get so, the instrument will return a string like this:
+        # "U, RANGE=3.000kV, VALUE=2.458kV", then measure_voltage will return 2458.0,
+        # hence the convenience of the get_process.
+        get_process=lambda v:
+        float(re.findall(EurotestHPP120256.f_numbers_regex_pattern, v[2])[0])
+    )
+
+    voltage_range = Instrument.measurement(
+        "STATUS,MU",
+        """ Measures the actual output voltage range of the HV Source (kV).""",
```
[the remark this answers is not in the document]
sansanda: Sure. You are right. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: As in Intrument ask method where is used time.sleep with a query_delay, here in write method could be possible the use of a delay_after_write. For me could be better leave the query_delay=0 and make use of that  new delay as I need both. What do you think? Thanks.

Note: I think remove change query_delay in ask and leave that new delay un write is not an option.

bmoneke: I don't understand, whether you want a delay after write or not.

Nevertheless, if you consider it opportune, you can subclass `write` which calls afterwards `wait_for` (which calls sleep). With a parameter you may distinguish between cases, where you need it and those where not (I'd default to not wait).

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Done in commit. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes.I need it because the instrument (old) miss commands if we do not wait between two consecutive writes. Some occurs in the ask case (write and read without waiting in the middle). I will considers your proposal. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+        :param ramp: Is the changing rate (ramp in V/s) for the ramp setting
+        """
+        log.info("Executing the ramp_to_zero function with ramp: {ramp} V/s.")
+
+        self.voltage_ramp = ramp
+        time.sleep(self.COMMAND_DELAY)
+        self.voltage = 0
+        time.sleep(self.COMMAND_DELAY)
+        self.voltage = 0
+        time.sleep(self.COMMAND_DELAY)
```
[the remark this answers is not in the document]
sansanda: Yes. For ensuring the voltage setpoint sets correctly. Sometimes we found that the instrument miss that command but I will remove the repetition after overriding the write method. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def wait_for_voltage_output_set(self, check_period=1.0, timeout=60.0):
```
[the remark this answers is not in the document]
sansanda: Yes. Better. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def emergency_off(self):
+        """ The output of the HV source will be switched OFF permanently and the values
+        of the voltage a current settings set to zero"""
+        log.info("Sending emergency off command to the instrument.")
+
+        self.write("EMCY OFF")
+        time.sleep(self.COMMAND_DELAY)
+
+    def shutdown(self, ramp):
```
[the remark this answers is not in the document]
sansanda: voltage_decreasing_rate? Too large but... Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Perfect!!! I didn't know that math function. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
msmttchr: I would suggest to override write and add the delay you need.

pymeasure/instruments/eurotest/eurotestHPP120256.py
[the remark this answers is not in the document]
bilderbuchi: `voltage_rate` ? It should be obvious that it's decreasing maybe? :sweat_smile:

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+            time.sleep(check_period)  # wait for voltage output reaches the voltage output setting
+            voltage_output = self.measure_voltage
+            voltage_output_set = (voltage_output > (voltage_output_setting - error)) and \
+                                 (voltage_output < (voltage_output_setting + error))
+            log.debug("voltage_output_valid_range: "
+                      "[" + str(voltage_output_setting - error) +
+                      ", " + str(voltage_output_setting + error) + "]")
+            log.debug("voltage_output: " + str(voltage_output))
+            if actual_time > future_time:
+                raise Exception("Timeout for wait_for_voltage_output_set function")
```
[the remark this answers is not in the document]
sansanda: Done in commit. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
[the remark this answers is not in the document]
sansanda: Sure. Taking into account the context. Jejejeje

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+    COMMAND_DELAY = 0.2  # s
+
+    response_encoding = "iso-8859-2"
+    f_numbers_regex_pattern = r'([+-]?[\d\.]+)'
+
+    # ####################################
+    # # EuroTest-Command set. Non SCPI commands.
+    # ####################################
+
+    voltage = Instrument.control(
```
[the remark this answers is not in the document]
sansanda: Yes. More precise. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+        # When this property acts as get, the instrument will return a string like this:
+        # "RAMP, RANGE=3000V/s, VALUE=1000V/s", then voltage_ramp will return 1000.0,
+        # hence the convenience of the get_process.
+        validator=strict_range,
+        values=VOLTAGE_RAMP_RANGE,
+        get_process=lambda v:
+        float(re.findall(EurotestHPP120256.f_numbers_regex_pattern, v[2])[0])
+    )
+
+    measure_voltage = Instrument.measurement(
```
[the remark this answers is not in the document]
sansanda: As you say, best voltage

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+        "STATUS,MU",
+        """ Measures the actual output voltage range of the HV Source (kV).""",
+        # This property is a get so, the instrument will return a string like this:
+        # "U, RANGE=3.000kV, VALUE=2.458kV", then voltage_range will return 3000.0,
+        # hence the convenience of the get_process.
+        get_process=lambda v:
+        float(re.findall(EurotestHPP120256.f_numbers_regex_pattern, v[1])[0])
+    )
+
+    measure_current = Instrument.measurement(
```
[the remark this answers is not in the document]
sansanda: Done in commit. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: In debug mode is important tracking the the voltage level at the out because sometimes the instrument miss the voltage_setting command. Also is important to show the error window. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes. You are right. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Got it. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Done in commit. thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Got it. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Done in commit. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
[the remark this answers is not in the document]
sansanda: Solved overriding Instrument write method

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,334 @@
...
+        get_process=lambda v:
+        v[1].encode(EurotestHPP120256.response_encoding).decode('utf-8', 'ignore')
+    )
+
+    lam_status = Instrument.measurement(
+        "STATUS,LAM",
+        """ Returns the LAM status which is the status of the unit from the point
+        of view of the process. Fo example, as a response of asking STATUS,LAM, the HV
+        voltage could response one of the messages from the next list:
+        LAM,ERROR External Inhibit occurred during Kill enable
```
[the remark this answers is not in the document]
sansanda: In this case lam_status response is itself a descriptive string.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,417 @@
...
+        return
+
+    # Wrapper functions for the Adapter object
+    def write(self, command, **kwargs):
+        """Overrides Instrument write method for including write_delay time after the parent call.
+
+        :param command: command string to be sent to the instrument
+        """
+        time.sleep(self.write_delay)
+        super().write(command, **kwargs)
```
msmttchr, line 364:
```suggestion
        actual_write_delay = time.time() - self.last_write_timestamp
        time.sleep(max(0, self.write_delay - actual_write_delay))
        super().write(command, **kwargs)
        self.last_write_timestamp = time.time()
```
If I understand correctly, you would like to ensure that two consecutive writes are separated at least by `self.write_delay`, I believe the code changes I proposed, should achieve this without any overhead.
Note that, if this change is accepted, then `self.last_write_timestamp` should be initialized (to zero?) in `__init__`.

msmttchr (commented):
Suggested optimization for `write` method.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,417 @@
...
+    """ Represents the Euro Test High Voltage DC Source model HPP-120-256
+    and provides a high-level interface for interacting with the instrument using the
+    Euro Test command set (Not SCPI command set).
+
+    .. code-block:: python
+
+    hpp120256 = EurotestHPP120256("GPIB0::20::INSTR")
+
+    response = hpp120256.id
+    print(response)
```
bilderbuchi, line 51:
```suggestion
    print(pp120256.id)
```

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,417 @@
...
+    """
+
+    VOLTAGE_RANGE = [0.0, 12.0]  # kVolts
+    CURRENT_RANGE = [0.0, 25.0]  # mAmps
+    VOLTAGE_RAMP_RANGE = [10, 3000]  # V/s
+    COMMAND_DELAY = 0.2  # s
+
+    response_encoding = "iso-8859-2"
+    f_numbers_regex_pattern = r'([+-]?([\d]*\.)?[\d]+)'
+    regex = re.compile(f_numbers_regex_pattern)
```
bilderbuchi, line 98:
```suggestion
    regex = re.compile(r'([+-]?([\d]*\.)?[\d]+)')
```
If you put the string directly into the call, you avoid having a module-level variable lying around that you are actually not using elsewhere (afaict).

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,417 @@
...
+        "KILL,%s",
+        """ Enables or disables (True/False) the kill function of the HV source.
+         When Kill is enabled yellow led is flashing and the output
+         will be shut OFF permanently without ramp if Iout > IOUTmax.""",
+        validator=strict_discrete_set,
+        values={True: 'ENable', False: 'DISable'},
+        map_values=True
+    )
+
+    enable_output = Instrument.setting(
```
bilderbuchi, line 212:
```suggestion
    output_enabled = Instrument.setting(
```
follows a common pattern we try to establish. 
The instrument does not have a way to query that status?

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,417 @@
...
+        LAM,INPUT ERROR Wrong command received
+        LAM,OK Status OK""",
+        get_process=lambda r:
+        r[1].strip().encode(EurotestHPP120256.response_encoding).decode('utf-8', 'ignore')
+    )
+
+    def emergency_off(self):
+        """ The output of the HV source will be switched OFF permanently and the values
+        of the voltage and current settings set to zero"""
+        log.info("Sending emergency off command to the instrument.")
```
bilderbuchi, line 276: Feels like this should be a `warning` instead of an `info`-level message`.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,417 @@
...
+
+        self.write("EMCY OFF")
+
+    def shutdown(self, voltage_rate):
```
bilderbuchi, line 280: Does it make sense to add a default value for the `voltage_rate` (e.g. the maximum of `VOLTAGE_RAMP_RANGE`), so that users only have to call `shutdown()` for it to safely and quickly shut down?

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,417 @@
...
+        # When this property acts as get, the instrument will return a string like this:
+        # "I, RANGE=5000mA, VALUE=1739mA", then current_limit will return 1739.0,
+        # hence the convenience of the get_process.
+        validator=strict_range,
+        values=CURRENT_RANGE,
+        get_process=lambda r:
+        float(EurotestHPP120256.regex.search(r[2].strip()).groups()[0])
+    )
+
+    voltage_ramp = Instrument.control(
```
bilderbuchi, line 149:
```suggestion
    voltage_ramp_rate = Instrument.control(
```
Better name? Also needs to be changed below, then.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,417 @@
...
+                                              abs_tol=abs_output_voltage_error)
+            log.debug("voltage_output_valid_range: "
+                      "[" + str(voltage_setpoint - abs_output_voltage_error) +
+                      ", " + str(voltage_setpoint + abs_output_voltage_error) + "]")
+            log.debug("voltage_output: " + str(voltage_output))
+            log.debug(f"Elapsed time: {round(actual_time - ref_time, ndigits=1)} seconds.")
+
+            if actual_time > future_time:
+                self.shutdown(200.0)  # in case the voltage were applied at the output
+                raise TimeoutError("Timeout for wait_for_output_voltage_reached function")
```
bilderbuchi, line 351: This feels more pythonic to me, just a suggestion:
```suggestion
        def target_reached(voltage_output):
            return math.isclose(voltage_output, voltage_setpoint, rel_tol=0.0,
                                abs_tol=abs_output_voltage_error)

        log.debug(f"\tWaiting for voltage output set. "
                  f"Reading output voltage every {check_period} seconds.\n"
                  f"\tTimeout: {timeout} seconds.")

        while True:
            actual_time = time.time()
            time.sleep(check_period)  # wait for voltage output reaches the voltage output setting
            voltage_output = self.voltage
            if target_reached(voltage_output):
                break
            log.debug("voltage_output_valid_range: "
                      "[" + str(voltage_setpoint - abs_output_voltage_error) +
                      ", " + str(voltage_setpoint + abs_output_voltage_error) + "]")
            log.debug("voltage_output: " + str(voltage_output))
            log.debug(f"Elapsed time: {round(actual_time - ref_time, ndigits=1)} seconds.")

            if actual_time > future_time:
                self.shutdown(200.0)  # in case the voltage were applied at the output
                raise TimeoutError("Timeout for wait_for_output_voltage_reached function")
```

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Got it

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Got it. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: I think voltage_ramp denotes itself a rate (volts/secs)

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes. Now output_enabled and kill_enabled are Instrument.control. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Sure. Done in commit. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: I see, but what you think if we avoid the creation of a new function and change the if as shown next? 

```
if target_reached(voltage_output):
                break
```

to  

```
if math.isclose(voltage_output, voltage_setpoint, rel_tol=0.0,
                                abs_tol=abs_output_voltage_error):
                break
```

Thanks

sansanda (commented):
Thanks for the requests

sansanda (commented):
Thanks for the requests

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Yes. More accurate. Thanks

tests/instruments/eurotest/test_eurotestHPP120256.py
```diff
@@ -0,0 +1,74 @@
...
+        assert inst.voltage == 2.458
+
+
+def test_measure_current():
+    """Verify the communication of the measure_current getter."""
+    with expected_protocol(
+            EurotestHPP120256,
+            [("STATUS,MI", "IM, RANGE=5000mA, VALUE=1739mA")],
+    ) as inst:
+        assert inst.current == 1739.0
```
bmoneke, line 74: As a matter of style, I'd name the tests "test_property", i.e. "test_voltage", similar to above tests. Similarly the docstrings do not match the property name.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -0,0 +1,426 @@
...
+            log.debug("voltage_output: " + str(voltage_output))
+            log.debug(f"Elapsed time: {round(actual_time - ref_time, ndigits=1)} seconds.")
+
+            if actual_time > future_time:
+                self.shutdown()  # in case the voltage were applied at the output
+                raise TimeoutError("Timeout for wait_for_output_voltage_reached function")
+
+        log.info("Waiting for voltage output set done.")
+
+        return
```
bmoneke, line 362: Why do you have return in the function?

bmoneke (changes requested):
Two small comments from my side and a larger request:
We recently expanded the documentation, also regarding the formulation of [docstrings](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings).

Please reformulate the docstings accordingly. Especially start them with "Set", "Measure", or "Control" (depending of the property type), this replaces those additions like "this property can be set and read".

I hope you understand this change request due to our effort to harmonize docstrings and make them better.

tests/instruments/eurotest/test_eurotestHPP120256.py
sansanda: Right. Thanks

pymeasure/instruments/eurotest/eurotestHPP120256.py
sansanda: Apss. Could be an error. Thanks.

pymeasure/instruments/eurotest/eurotestHPP120256.py
```diff
@@ -199,24 +198,34 @@ def __init__(self,
...
+         will be shut OFF permanently without ramp if Iout > IOUTmax. This property can be set.""",
         validator=strict_discrete_set,
         values={True: 'ENable', False: 'DISable'},
-        map_values=True
+        map_values=True,
+        get_process=lambda r:
+        EurotestHPP120256.EurotestHPP120256Status(
+            int(r[1].strip()[:-1].encode(EurotestHPP120256.response_encoding).
+                decode('utf-8', 'ignore'), 2)
+        ) & EurotestHPP120256.EurotestHPP120256Status.KILL_ENABLE
```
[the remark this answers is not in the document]
sansanda: I supposed that the & operator will return a boolean. Thanks.

docs/about/authors.rst
```diff
@@ -57,3 +57,4 @@ The following developers have contributed to the PyMeasure package:
 | Asaf Yagoda
 | Fabio Garzetti
 | Daniel Schmeer
+| David Sanchez Sanchez
```
bmoneke, line 60: Since the last merge, we do not need the name in "authors.rst" anymore.
Just revert this change and the "merge conflict" message should disappear.

sansanda: Done. Thanks

msmttchr (approved):

pymeasure/instruments/eurotest/eurotestHPP120256.py
bilderbuchi: @sansanda Please don't press "resolve" if the discussion is not resolved, especially if you expect feedback! (that button hides the discussion from view by default)

In general, if it's not a very clear change, like committing a suggested change, let the one who opened a discussion resolve it, that way both parties can check that it is indeed resolved.

bilderbuchi: > I think voltage_ramp denotes itself a rate (volts/secs). I prefer to leave it as is, don't you think?

Well, that's of course a matter of taste up to some point. I'll point out that even your docstring refers to "ramp speed" (more accurately a rate, not a speed, but that's beside the main point), so I'd think that `voltage_ramp_rate` would be clearer. :-)

bilderbuchi: However, it's a minor thing, I'll leave it up to you.

pymeasure/instruments/eurotest/eurotestHPP120256.py
bilderbuchi: Yeah, now that you refactored the loop so you don't need to check that in two different places (like before), you don't need the extra method.

bilderbuchi:
@sansanda this looks ready for me too. 
Can you find the time to refactor the docstrings as @bmoneke suggested? This would save us some effort later. :-)

> Two small comments from my side and a larger request: We recently expanded the documentation, also regarding the formulation of [docstrings](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings).
> 
> Please reformulate the docstings accordingly. Especially start them with "Set", "Measure", or "Control" (depending of the property type), this replaces those additions like "this property can be set and read".
> 
> I hope you understand this change request due to our effort to harmonize docstrings and make them better.

bilderbuchi (approved):

bmoneke:
Thanks a lot for that refactoring!

sansanda:
Thanks to you and best regards.
David

PR #706, opened 2022-10-13
HCP tc038d moved to pymeasure properties.

bmoneke:
I changed the tc038d from a custom design to a default pymeasure design (write message, read message and properties).
This might be an example (besides the one in the documentation) on how to implement register based instruments. For that reason, I opened this PR.

Thanks to the tests, the communication did not change during the refactoring. (Test coverage 100%)

### DIFF
--- a/pymeasure/instruments/hcp/tc038d.py
+++ b/pymeasure/instruments/hcp/tc038d.py
@@ -77,20 +77,21 @@ class TC038D(Instrument):
             - or the number of elements to read (defaults to 1).
         """
         function, address, *values = command.split(",")
+        function = Functions[function]
         data = [self.address]  # 1B device address
-        data.append(Functions[function].value)  # 1B function code
+        data.append(function.value)  # 1B function code
         address = int(address, 16) if "x" in address else int(address)
         data.extend(address.to_bytes(2, "big"))  # 2B register address
-        if function.upper() == "W":
+        if function == Functions.W:
             elements = len(values) * self.byteMode // 2
             data.extend(elements.to_bytes(2, "big"))  # 2B number of elements
             data.append(elements * 2)  # 1B number of bytes to write
             for element in values:
                 data.extend(int(element).to_bytes(self.byteMode, "big", signed=True))
-        elif function.upper() == "R":
+        elif function == Functions.R:
             count = int(values[0]) * self.byteMode // 2 if values else self.byteMode // 2
             data.extend(count.to_bytes(2, "big"))  # 2B number of elements to read
-        elif function == "ECHO":
+        elif function == Functions.ECHO:
             data[-2:] = [0, 0]
             if values:
                 data.extend(int(values[0]).to_bytes(2, "big"))  # 2B test data
--- a/pymeasure/instruments/hcp/tc038d.py
+++ b/pymeasure/instruments/hcp/tc038d.py
@@ -22,6 +22,8 @@
 # THE SOFTWARE.
 #
 
+from enum import Enum
+
 from pymeasure.instruments import Instrument
 
 
@@ -38,6 +40,13 @@ def CRC16(data):
     return [CRC & 0xFF, CRC >> 8]
 
 
+class Functions(Enum):
+    R = 0x03
+    WRITESINGLE = 0x06
+    ECHO = 0x08  # register address has to be 0
+    W = 0x10  # writing multiple variables
+
+
 class TC038D(Instrument):
     """
     Communication with the HCP TC038D oven.
@@ -51,12 +60,6 @@ class TC038D(Instrument):
     """
 
     byteMode = 4
-
-    functions = {'R': 0x03,
-                 'writeSingle': 0x06,
-                 'echo': 0x08,  # register address has to be 0
-                 'W': 0x10,  # writing multiple variables
-                 }
 
     def __init__(self, adapter, name="TC038D", address=1, timeout=1000,
                  **kwargs):
@@ -75,19 +78,19 @@ class TC038D(Instrument):
         """
         function, address, *values = command.split(",")
         data = [self.address]  # 1B device address
-        data.append(self.functions[function])  # 1B function code
+        data.append(Functions[function].value)  # 1B function code
         address = int(address, 16) if "x" in address else int(address)
         data.extend(address.to_bytes(2, "big"))  # 2B register address
         if function.upper() == "W":
             elements = len(values) * self.byteMode // 2
-            data += [elements >> 8, elements & 0xFF]  # 2B number of elements
+            data.extend(elements.to_bytes(2, "big"))  # 2B number of elements
             data.append(elements * 2)  # 1B number of bytes to write
             for element in values:
                 data.extend(int(element).to_bytes(self.byteMode, "big", signed=True))
         elif function.upper() == "R":
             count = int(values[0]) * self.byteMode // 2 if values else self.byteMode // 2
             data.extend(count.to_bytes(2, "big"))  # 2B number of elements to read
-        elif function == "echo":
+        elif function == "ECHO":
             data[-2:] = [0, 0]
             if values:
                 data.extend(int(values[0]).to_bytes(2, "big"))  # 2B test data
@@ -98,7 +101,7 @@ class TC038D(Instrument):
         """Read response and interpret the number"""
         # Slave address, function
         got = self.read_bytes(2)
-        if got[1] == self.functions['R']:
+        if got[1] == Functions.R.value:
             # length of data to follow
             length = self.read_bytes(1)
             # data length, 2 Byte CRC
@@ -106,12 +109,12 @@ class TC038D(Instrument):
             if read[-2:] != bytes(CRC16(got + length + read[:-2])):
                 raise ConnectionError("Response CRC does not match.")
             return int.from_bytes(read[:-2], byteorder="big", signed=True)
-        elif got[1] == self.functions['W']:
+        elif got[1] == Functions.W.value:
             # start address, number elements, CRC; each 2 Bytes long
             got += self.read_bytes(2 + 2 + 2)
             if got[-2:] != bytes(CRC16(got[:-2])):
                 raise ConnectionError("Response CRC does not match.")
-        elif got[1] == self.functions['echo']:
+        elif got[1] == Functions.ECHO.value:
             # start address 0, data, CRC; each 2B
             got += self.read_bytes(2 + 2 + 2)
             if got[-2:] != bytes(CRC16(got[:-2])):
@@ -133,8 +136,8 @@ class TC038D(Instrument):
         self.read()
 
     def ping(self, test_data=0):
-        """Test the connection sending an integer up to 65535."""
-        assert self.ask(f"echo,0,{test_data}") == test_data
+        """Test the connection sending an integer up to 65535, checks the response."""
+        assert self.ask(f"ECHO,0,{test_data}") == test_data
 
     setpoint = Instrument.control(
         "R,0x106", "W,0x106,%i",
--- a/pymeasure/instruments/hcp/tc038d.py
+++ b/pymeasure/instruments/hcp/tc038d.py
@@ -22,8 +22,6 @@
 # THE SOFTWARE.
 #
 
-import collections.abc
-
 from pymeasure.instruments import Instrument
 
 
@@ -55,7 +53,9 @@ class TC038D(Instrument):
     byteMode = 4
 
     functions = {'read': 0x03, 'writeMultiple': 0x10,
-                 'writeSingle': 0x06, 'echo': 0x08}
+                 'writeSingle': 0x06, 'echo': 0x08,
+                 'R': 0x03, 'W': 0x10,
+                 }
 
     def __init__(self, adapter, name="TC038D", address=1, timeout=1000,
                  **kwargs):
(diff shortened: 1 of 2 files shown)


pymeasure/instruments/hcp/tc038d.py
```diff
@@ -55,89 +53,83 @@ class TC038D(Instrument):
...
+            - the hexadecimal address to write to (e.g. 0x106),
+            - the values (comma separated) to write
+            - or the number of elements to read (defaults to 1).
+        """
+        function, address, *values = command.split(",")
+        data = [self.address]  # 1B device address
+        data.append(self.functions[function])  # 1B function code
+        address = int(address, 16)  # register address
+        data += [address >> 8, address & 0xFF]  # 2B register address
+        if data[1] == self.functions['writeMultiple']:
```
msmttchr, line 80:
```suggestion
        if function == 'writeMultiple':
```

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -55,89 +53,83 @@ class TC038D(Instrument):
...
+            - the function read (R) or write (W),
+            - the hexadecimal address to write to (e.g. 0x106),
+            - the values (comma separated) to write
+            - or the number of elements to read (defaults to 1).
+        """
+        function, address, *values = command.split(",")
+        data = [self.address]  # 1B device address
+        data.append(self.functions[function])  # 1B function code
+        address = int(address, 16)  # register address
+        data += [address >> 8, address & 0xFF]  # 2B register address
```
msmttchr, line 79:
```suggestion
        data.extend(address.to_bytes(2, 'little')) # 2B register address
```
Possibly defining data as bytes could further simplify code.

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -55,89 +53,83 @@ class TC038D(Instrument):
...
             for element in values:
-                for i in range(self.byteMode - 1, -1, -1):
-                    data.append(element >> i * 8 & 0xFF)
-        else:
-            raise TypeError(("Values has to be an integer or an iterable of "
-                             f"integers. values: {values}"))
+                data.extend(int(element).to_bytes(self.byteMode, byteorder="big", signed=True))
+        elif data[1] == self.functions['read']:
+            count = int(values[0]) * self.byteMode // 2 if values else self.byteMode // 2
+            data += [count >> 8, count & 0xFF]  # 2B number of elements to read
```
msmttchr, line 88:
```suggestion
            data.extend(count.to_bytes(2, 'little')  # 2B number of elements to read
```
Same also for line 82 above

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -55,89 +53,83 @@ class TC038D(Instrument):
...
+            errors = {0x02: "Wrong start address.",
+                      0x03: "Variable data error.",
+                      0x04: "Operation error."}
+            if end[0] in errors.keys():
+                raise ValueError(errors[end[0]])
+            else:
+                raise ConnectionError(f"Unknown read error. Received: {got} {end}")
+
+    def check_errors(self):
+        """Read the response after setting a value."""
```
msmttchr, line 120: Is the docstring aligned with purpose of method?

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -55,89 +53,83 @@ class TC038D(Instrument):
     byteMode = 4
 
     functions = {'read': 0x03, 'writeMultiple': 0x10,
-                 'writeSingle': 0x06, 'echo': 0x08}
+                 'writeSingle': 0x06, 'echo': 0x08,
```
msmttchr, line 56: Can this be replaced by `Enum` ?

msmttchr (commented):
Few minor observations from my side.

pymeasure/instruments/hcp/tc038d.py
bmoneke: The reason for that strange condition test is, that you can define the function (right now, at least) either by the full name 'writeMultiple' or by the short form 'W' (which you will prefer writing properties).
Now I compare the function codes.

If I get rid of the long form, the test would be more simple.

pymeasure/instruments/hcp/tc038d.py
bmoneke: In fact, that is how I improved the documentation example, but I did not implement it here 🤦.
Thanks for pointing out.
I think it is big endian, however.

pymeasure/instruments/hcp/tc038d.py
bmoneke: The device always responds (even to setters), therefore I have to read afterwards.
The only way to do this right now, is to enable `check_set_errors` in the properties, which call check_errors.

So, I want to read, but have to call it 'check_errors'.

pymeasure/instruments/hcp/tc038d.py
bmoneke: I have not implemented enums yet, as dictionaries mostly do the job natively.
I don't see the advantage of enums, when you have dictionaries. I'll do some research.

bmoneke:
@msmttchr thanks for your comments.

pymeasure/instruments/hcp/tc038d.py
msmttchr: Indeed is big endian. Sorry.

pymeasure/instruments/hcp/tc038d.py
msmttchr: Ok. Thanks for the clarification.

pymeasure/instruments/hcp/tc038d.py
msmttchr: Ok. I think, it could be explained in the docstring, since it's a bit cryptic.
BTW, in the past I implemented an instrument with similar behaviour and I think I overridden the `write` method.
I am a bit hesitant, but in SCPI instrument, you can call `check_errors` whenever you like, so it could mislead users.
No strong opinion on which way is better, feel free to choose what you think is more appropriate.

bmoneke: I have the issue in the back of my mind, as I already proposed a small solution (allowing `control`'s setter to ask instead of write).
Maybe we can tackle it together with the split in Scpi an non Scpi instruments or together with the introduction of channels (with dynamic properties).

pymeasure/instruments/hcp/tc038d.py
bilderbuchi: Might be worth it to add a short source comment with that reason.

pymeasure/instruments/hcp/tc038d.py
bilderbuchi: This way is so much more convenient & expressive, it's very easy to get confused when doing this manually, with shifts and ANDs (as demonstrated here :stuck_out_tongue_closed_eyes: )

pymeasure/instruments/hcp/tc038d.py
bilderbuchi: There's a little better/stricter typing, the access pattern is different (e.g. `functions.READ` instead of `functions['read']` here), which means you'd get autocompletion on enum members, and with `IntFlag` enums, you also get bitwise combination possibilities (if you need that), which I've used to great effect in the past.

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -54,90 +52,100 @@ class TC038D(Instrument):
...
+            - or the number of elements to read (defaults to 1).
+        """
+        function, address, *values = command.split(",")
+        data = [self.address]  # 1B device address
+        data.append(self.functions[function])  # 1B function code
+        address = int(address, 16) if "x" in address else int(address)
+        data.extend(address.to_bytes(2, "big"))  # 2B register address
+        if function.upper() == "W":
             elements = len(values) * self.byteMode // 2
             data += [elements >> 8, elements & 0xFF]  # 2B number of elements
```
bilderbuchi, line 83: Maybe that line should also be converted to the `extend` + `to_bytes` approach?

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -54,90 +52,100 @@ class TC038D(Instrument):
...
-            data.append(len(values) * self.byteMode)  # 1B number of write data
+            data.append(elements * 2)  # 1B number of bytes to write
```
bilderbuchi, line 84: Do you want to remove the `byteMode` if it's not used anymore/only one mode? I'd recommend either using it everywhere relevant, or hard-code it everywhere, but not mixing those two.

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -54,90 +52,100 @@ class TC038D(Instrument):
...
             for element in values:
-                for i in range(self.byteMode - 1, -1, -1):
-                    data.append(element >> i * 8 & 0xFF)
-        else:
-            raise TypeError(("Values has to be an integer or an iterable of "
-                             f"integers. values: {values}"))
+                data.extend(int(element).to_bytes(self.byteMode, "big", signed=True))
+        elif function.upper() == "R":
+            count = int(values[0]) * self.byteMode // 2 if values else self.byteMode // 2
```
bilderbuchi, line 88: Afaict, this will fail if the user uses the "defaults to 1" option of "the number of elements to read", as `values` will be an empty list. Add a test for that?

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -54,90 +52,100 @@ class TC038D(Instrument):
...
-        if got[1] == self.functions['writeMultiple']:
+        if got[1] == self.functions['R']:
+            # length of data to follow
+            length = self.read_bytes(1)
+            # data length, 2 Byte CRC
+            read = self.read_bytes(length[0] + 2)
+            if read[-2:] != bytes(CRC16(got + length + read[:-2])):
+                raise ConnectionError("Response CRC does not match.")
+            return int.from_bytes(read[:-2], byteorder="big", signed=True)
+        elif got[1] == self.functions['W']:
```
bilderbuchi, line 109: So, this part is the "ACK" response of this instrument to sent commands? If yes, might be good to call that out in a comment.

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -54,90 +52,100 @@ class TC038D(Instrument):
...
+                raise ValueError(errors[end[0]])
+            else:
+                raise ConnectionError(f"Unknown read error. Received: {got} {end}")
+
+    def check_errors(self):
+        """To be called from the property setters to read the acknowledgment."""
+        self.read()
+
+    def ping(self, test_data=0):
+        """Test the connection sending an integer up to 65535."""
```
bilderbuchi, line 136:
```suggestion
        """Test the connection sending an integer up to 65535, checks the response."""
```

bilderbuchi:
LGTM, I remarked on a couple of things. I agree that it would be worth investigating if an enum fits better than a dict, it seems "cleaner".

pymeasure/instruments/hcp/tc038d.py
bmoneke: I'm not mixing hard coding and the variable:
The number of elements is half of the number of bytes to write (stupid to convey the same content twice, but such is the protocol), as in four byte mode, every value counts as two elements. In two byte mode, only as one element.

It should work in two byte mode as well, but I never tried it, nor does the manual give examples.

pymeasure/instruments/hcp/tc038d.py
bmoneke: Yes, if got[1] is the writing function code, the read message is the acknowledgement.

pymeasure/instruments/hcp/tc038d.py
bmoneke: Oh, thanks

bmoneke:
Thanks for the comments. I'll look into enums.

pymeasure/instruments/hcp/tc038d.py
bilderbuchi: OK, then.

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -54,90 +61,94 @@ class TC038D(Instrument):
...
+
+        :param str command: comma separated string of:
+            - the function: read ('R') or write ('W') or 'echo',
+            - the address to write to (e.g. '0x106' or '262'),
+            - the values (comma separated) to write
+            - or the number of elements to read (defaults to 1).
+        """
+        function, address, *values = command.split(",")
+        data = [self.address]  # 1B device address
+        data.append(Functions[function].value)  # 1B function code
```
bilderbuchi, line 81: IIRC, if you had `class Functions(IntEnum)` instead, I think this becomes easier because you could just do `Functions[function]`, as you then get the integer when you need it?

pymeasure/instruments/hcp/tc038d.py
```diff
@@ -54,90 +61,94 @@ class TC038D(Instrument):
...
+        address = int(address, 16) if "x" in address else int(address)
+        data.extend(address.to_bytes(2, "big"))  # 2B register address
+        if function.upper() == "W":
```
bilderbuchi, line 84: One benefit of using (Int)Enums would be here, I would do something like
```
# above
function = Functions[function]  # generate an enum from your string input
data.append(function)
# ....
if function == Functions.W  # no string comparison, but enum entry...clearer code, stricter type
# etc
```
Just a suggestion.

pymeasure/instruments/hcp/tc038d.py
bmoneke: I never used enums before, but the official documentation is quite good, therefore I noticed IntEnums, but they did not seem to offer any benefit here.
I tried it with an IntEnum, and it behaves the same like Enum (only comparisons work differently).

pymeasure/instruments/hcp/tc038d.py
bmoneke: Thanks for your suggestions. It helps me use enums better.

pymeasure/instruments/hcp/tc038d.py
bilderbuchi: The key difference is that an IntEnum behaves like an int.
> IntEnum is the same as Enum, but its members are also integers and can be used anywhere that an integer can be used. If any integer operation is performed with an IntEnum member, the resulting value loses its enumeration status.

This enables you (mostly) to transparently use them in a context where you are working with integers, but also want what enums offer, like here. :-)

bmoneke: Oh great, I push the changes immediately.

bilderbuchi: Cool, we are done, then! 😊

PR #707, opened 2022-10-13
Added measurement functionality for Keithley 2700

QuantumOfMoose:
Hello,
I have added some code to the keithley2700.py instrument file that gives users the ability to access and control the measurement functions of the instrument, in addition to its switching capabilities that were previously available.
Most of my additions are taken from the keithley2000.py file, with some modifications to account for the differences between the two instruments.

Thank you for making such a useful tool, I hope this contribution will be helpful.

### DIFF
--- a/pymeasure/instruments/keithley/keithley2700.py
+++ b/pymeasure/instruments/keithley/keithley2700.py
@@ -359,3 +359,574 @@ class Keithley2700(Instrument, KeithleyBuffer):
 
         # write the string to the display
         self.display_text = channel_string
+        
+        ###########
+        # MEASURE #
+        ###########
+        # Most of these are quite similar (and copied from) to keithley2000, as they both use SCPI and are very similar instruments.
+        
+        MODES = {
+        'current': 'CURR:DC', 'current ac': 'CURR:AC',
+        'voltage': 'VOLT:DC', 'voltage ac': 'VOLT:AC',
+        'resistance': 'RES', 'resistance 4W': 'FRES',
+        'period': 'PER', 'frequency': 'FREQ',
+        'temperature': 'TEMP', 'continuity': 'CONT'
+    }
+
+    mode = Instrument.control(
+        ":CONF?", ":CONF:%s",
+        """ A string property that controls the configuration mode for measurements,
+        which can take the values: :code:'current' (DC), :code:'current ac',
+        :code:'voltage' (DC),  :code:'voltage ac', :code:'resistance' (2-wire),
+        :code:'resistance 4W' (4-wire), :code:'period', :code:'frequency',
+        :code:'temperature', and :code:'frequency'. """,
+        validator=strict_discrete_set,
+        values=MODES,
+        map_values=True,
+        get_process=lambda v: v.replace('"', '')
+    )
+
+    beep_state = Instrument.control(
+        ":SYST:BEEP:STAT?",
+        ":SYST:BEEP:STAT %g",
+        """ A string property that enables or disables the system status beeper,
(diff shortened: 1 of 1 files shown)


pymeasure/instruments/keithley/keithley2700.py
```diff
@@ -359,3 +363,574 @@ def display_closed_channels(self):
...
+            self.mode = 'current'
+            self.current_range = max_current
+
+    def measure_resistance(self, max_resistance=10e6, wires=2):
+        """ Configures the instrument to measure voltage,
+        based on a maximum voltage to set the range, and
+        a boolean flag to determine if DC or AC is required.
+
+        :param max_voltage: A voltage in Volts to set the voltage range
+        :param ac: False for DC voltage, and True for AC voltage
```
bmoneke, line 851: The docstring does not match to the content of the method. Wondering, what "wires" means, I stumbled upon the discrepancy.

pymeasure/instruments/keithley/keithley2700.py~
```diff
@@ -0,0 +1,937 @@
+#
```
bmoneke, line 1: I guess you did not intend to commit this file (presumably an autosave file).

bmoneke:
Thank you for your contribution. The code looks good in a first, quick glance. I appreciate the different sections of attributes.

If it is your first contribution here, you may add yourself to the authors (in two files: Authors.txt in the main folder and in authors.rst in the documentation).

How close are both instruments? Would it be feasible to make a common "ancestor" and implement just the changes (either in one of them or in both)?
Dynamic properties (see documentation about adding instruments) allow to change property parameters (allowed values, getter command etc.), such that it is possible to implement easily differences between two devices.

Another point: We added the possibility to check the communication protocol, even without the hardware (e.g. According to the examples in the manual). I invite you, to write some tests, especially for the more complicated methods. For more information see the documentation regarding adding instruments, section 'tests'. If you have questions, I'll be glad to assist you.

bilderbuchi:
> How close are both instruments? Would it be feasible to make a common "ancestor" and implement just the changes (either in one of them or in both)?
Dynamic properties (see documentation about adding instruments) allow to change property parameters (allowed values, getter command etc.), such that it is possible to implement easily differences between two devices.

I agree, this should be the way to go, we don't want to copy-paste 500 LOC unnecessarily. 

See [this section](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#family-of-instruments-with-similar-features) of the documentation on how to achieve that. The `Keithley2700` will inherit from the `Keithley2000`, and only the changes you had to make to make your copy-pasted work need to be added to the 2700. Dynamic properties are the way to go to tweak existing properties. If any methods or properties are _not_ available /appropriate for the 2700, those can be deleted in the 2700 implementation. We can assist if you run into problems.

At some point we should probably go take a closer look at the Keithley instrument family, and extract a common base class (similar to `KeithleyBuffer`) for the things that all instruments have, I suspect there was a lot of copypasta going on.

QuantumOfMoose:
Thank you to both of you for your comments. I agree that a common ancestor would be the way to go for these, and probably all keithley instruments. All the ones I've encountered use SCPI commands, so there is a lot of overlap.

BenediktBurger:
@QuantumOfMoose , do you want to continue this pull request or should we close it for the time being?

QuantumOfMoose:
Hi Benedikt,

Yes, you can go ahead and close it. If I end up doing some more work on the Kiethley i'll just open a new one, but I don't anticipate that for the immediate future.

Have a good week.

Issue #705, opened 2022-10-08
Document full command tweaking possibilities


bilderbuchi:
At some point after completing #660, I'd love to come up with an example for one control, which exercises/uses most/all of our processing options, and show in a diagram/flow chart or similar the flow and mutation of the command string from the user-level query (`a = my_instr.prop`) down to the hardware, and back up. Just to show what pymeasure is capable of, in one big picture. :cloud:

bmoneke:
These examples could address documentation issues #530 and #573 as well.

PR #712, opened 2022-10-18
Update and extend documentation

### LINKED ISSUE #102: Improve documentation

bilderbuchi:
This should reduce confusing situations like [this](https://github.com/pymeasure/pymeasure/discussions/710#discussioncomment-3905231).

Closes #102.

### DIFF
--- a/docs/api/instruments/validators.rst
+++ b/docs/api/instruments/validators.rst
@@ -4,7 +4,7 @@
 Validator functions
 ###################
 
-Validators are used in conjunction with the :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function to allow properties with complex restrictions for valid values. They are described in more detail in the :ref:`Advanced properties <advanced-properties>` section.
+Validators are used in conjunction with the :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function to allow properties with complex restrictions for valid values. They are described in more detail in the :ref:`validators` section.
 
 .. automodule:: pymeasure.instruments.validators
     :members:
--- a/docs/api/instruments/validators.rst
+++ b/docs/api/instruments/validators.rst
@@ -4,7 +4,7 @@
 Validator functions
 ###################
 
-Validators are used in conjunction with the :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function to allow properties with complex restrictions for valid values. They are described in more detail in the :ref:`validators` section.
+Validators are used in conjunction with the :func:`Instrument.control <pymeasure.instruments.Instrument.control>` or :func:`Instrument.setting <pymeasure.instruments.Instrument.setting>` functions to allow properties with complex restrictions for valid values. They are described in more detail in the :ref:`validators` section.
 
 .. automodule:: pymeasure.instruments.validators
     :members:
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -203,6 +203,14 @@ As our signal values are often integers, the most appropriate enum types are :co
 
 :code:`IntFlags` are used by many instruments for the purpose just demonstrated.
 
+The status property could look like this:
+.. testcode::
+
+    status = Instrument.measurement(
+        "STB?", 
+        """Measure the status of the device as enum.""",
+        get_process=lambda v: ErrorCode(v), 
+   )
 .. _default_connection_settings:
 
 Defining default connection settings
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -199,9 +199,9 @@ Writing properties
 ==================
 
 In PyMeasure, `Python properties`_ are the preferred method for dealing with variables that are read or set. 
-PyMeasure comes with two convenience functions for making properties for classes. 
-The :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` function returns a property that issues a GPIB/SCPI requests when the value is used. 
-For example, if our "Extreme 5000" has the :code:`*IDN?` command we can write the following property to be added above the :code:`def __init__` line in our above example class, or added to the class after the fact as in the code here:
+PyMeasure comes with three convenience functions for making properties for classes.
+The :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` function returns a property that can only read values from an instrument.
+For example, if our "Extreme 5000" has the :code:`*IDN?` command we can write the following property to be added after the :code:`def __init__` line in our above example class, or added to the class after the fact as in the code here:
 
 .. _Python properties: https://docs.python.org/3/howto/descriptor.html#properties
 
@@ -227,6 +227,8 @@ When we use this property we will get the identification information.
     >>> extreme.id           # Reads "*IDN?"
     'Extreme 5000 identification from instrument'
 
+Note that the :code:`id` property is already defined for SCPI instruments, so you do not need to implement it for your instruments.
+
 The :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function extends this behavior by creating a property that you can read and set. For example, if our "Extreme 5000" has the :code:`:VOLT?` and :code:`:VOLT <float>` commands that are in Volts, we can write the following property.
 
 .. testcode::
@@ -251,7 +253,9 @@ We can use this property to set the voltage to 100 mV, which will execute the co
     >>> extreme.voltage              # Reads ":VOLT?"
     0.1
 
-Using :func:`Instrument.control <pymeasure.instruments.Instrument.control>` and :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` functions, you can create a number of properties for basic measurements and controls. 
+Finally, the :func:`Instrument.setting <pymeasure.instruments.Instrument.setting>` function can only set, but not read values.
+
+Using the :func:`Instrument.control <pymeasure.instruments.Instrument.control>`, :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` and :func:`Instrument.control <pymeasure.instruments.Instrument.control>` functions, you can create a number of properties for basic measurements and controls.
 
 The :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function can be used with multiple values at once, passed as a tuple. Say, we may set voltages and frequencies in our "Extreme 5000", and the the commands for this are :code:`:VOLTFREQ?` and :code:`:VOLTFREQ <float>,<float>`, we could use the following property:
 
@@ -448,6 +452,34 @@ The dictionary now maps the keys to specific values. The values and keys can be 
     'Y'
 
 As you have seen, the :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function can be significantly extended by using validators and maps.
+
+Boolean properties
+******************
+
+The idea of using maps can be leveraged to implement properties where the user-facing values are booleans, so you can interact in a pythonic way using :code:`True` and :code:`False`:
+
+.. testcode::
+
+    Extreme5000.output_enabled = Instrument.control(
+        "OUTP?", "OUTP %d",
+        """A boolean property that turns the output on (True) or off (False).""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, False: 0},  # the dict values could also be "on" and "off", etc.
+    )
+
+
+.. doctest::
+
+    >>> extreme = Extreme5000("GPIB::1")
+    >>> extreme.output_enabled = True
+    >>> extreme.read()
+    '1'
+    >>> extreme.output_enabled = False
+    >>> extreme.output_enabled
+    False
+
+Good names for boolean properties are chosen such that they could also be a yes/no question: "Is the output enabled?" -> :code:`output_enabled`, :code:`display_active`, etc.
 
 Processing of set values
 ************************
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -655,7 +655,7 @@ In this example, the property name was :code:`voltage` and the parameter to adju
 .. _instruments_with_similar_features:
 
 Instruments with similar features
-*********************************
+=================================
 
 When instruments have a similar set of features, it makes sense to use inheritance to obtain most of the functionality from a parent instrument class, instead of copy-pasting code.
 Sometimes one only needs to add additional properties and methods.
@@ -664,7 +664,7 @@ Often, however, only some details need to be changed.
 This can be dealt with efficiently using dynamic properties.
 
 Instrument family with different parameter values
--------------------------------------------------
+*************************************************
 
 A common case is to have a family of similar instruments with some parameter range different for each family member.
 In this case you would update the specific class parameter range without rewriting the entire property:
@@ -695,7 +695,7 @@ In this case you would update the specific class parameter range without rewriti
 Notice how easily you can derive the different family members from a common class, and the fact that the attribute is now defined at class level and not at instance level.
 
 Instruments with similar command syntax
----------------------------------------
+***************************************
 
 Another use case involves maintaining compatibility between instruments with commands having different syntax, like in the following example.
 
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -280,7 +280,7 @@ You will notice that we use the `Python string format`_ :code:`%g` to format pas
 
 .. _Python string format: https://docs.python.org/3/library/string.html#format-specification-mini-language
 
-We can use this property to set the voltage to 100 mV, which will sends the appropriate command and then requests the current voltage.
+We can use this property to set the voltage to 100 mV, which will send the appropriate command and then requests the current voltage.
 
 .. doctest::
 
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -217,8 +217,11 @@ For many instruments, the simple way presented first is enough, but in case you 
 Writing properties
 ==================
 
-In PyMeasure, `Python properties`_ are the preferred method for dealing with variables that are read or set. 
-PyMeasure comes with three convenience functions for making properties for classes: :func:`Instrument.control <pymeasure.instruments.Instrument.control>`, :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` and :func:`Instrument.setting <pymeasure.instruments.Instrument.setting>`.
+In PyMeasure, `Python properties`_ are the preferred method for dealing with variables that are read or set.
+
+The property factories
+**********************
+PyMeasure comes with three central convenience factory functions for making properties for classes: :func:`Instrument.control <pymeasure.instruments.Instrument.control>`, :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` and :func:`Instrument.setting <pymeasure.instruments.Instrument.setting>`.
 
 The :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` function returns a property that can only read values from an instrument.
 For example, if our "Extreme 5000" has the :code:`*IDN?` command we can write the following property to be added after the :code:`def __init__` line in our above example class, or added to the class after the fact as in the code here:
@@ -276,38 +279,7 @@ Finally, the :func:`Instrument.setting <pymeasure.instruments.Instrument.setting
 
 Using the :func:`Instrument.control <pymeasure.instruments.Instrument.control>`, :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` and :func:`Instrument.control <pymeasure.instruments.Instrument.control>` functions, you can create a number of properties for basic measurements and controls.
 
-Using multiple values
-*********************
-Seldomly, you might need to send/receive multiple values in one command.
-The :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function can be used with multiple values at one time, passed as a tuple. Say, we may set voltages and frequencies in our "Extreme 5000", and the the commands for this are :code:`:VOLTFREQ?` and :code:`:VOLTFREQ <float>,<float>`, we could use the following property:
-
-.. testcode::
-
-    Extreme5000.combination = Instrument.control(
-        ":VOLTFREQ?", ":VOLTFREQ %g,%g",
-        """Simultaneously control the voltage in Volts and the frequency in Hertz (both float).
-
-        This property is set by a tuple.
-        """
-    )
-
-In use, we could set the voltage to 200 mV, and the Frequency to 931 Hz, and read both values immediately afterwards. 
-
-.. doctest::
-
-    >>> extreme = Extreme5000("GPIB::1")
-    >>> extreme.combination = (0.2, 931)        # Executes ":VOLTFREQ 0.2,931"
-    >>> extreme.combination                     # Reads ":VOLTFREQ?"
-    [0.2, 931.0]
-
-This interface is not too convenient, but luckily not often needed.
-
-The next section details additional features of :func:`Instrument.control <pymeasure.instruments.Instrument.control>` that allow you to write properties that cover specific ranges, or have to map between a real value to one used in the command. Furthermore it is shown how to perform more complex processing of return values from your device.
-
-.. _advanced-properties:
-
-Advanced properties
-===================
+The next sections detail additional features of :func:`Instrument.control <pymeasure.instruments.Instrument.control>` that allow you to write properties that cover specific ranges, or have to map between a real value to one used in the command. Furthermore it is shown how to perform more complex processing of return values from your device.
 
 .. _validators:
 
@@ -597,9 +569,34 @@ The same can be also achieved by the `preprocess_reply` keyword argument to :fun
         # notice how we don't need to cast to float anymore
     )
 
+Using multiple values
+*********************
+Seldomly, you might need to send/receive multiple values in one command.
+The :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function can be used with multiple values at one time, passed as a tuple. Say, we may set voltages and frequencies in our "Extreme 5000", and the the commands for this are :code:`:VOLTFREQ?` and :code:`:VOLTFREQ <float>,<float>`, we could use the following property:
+
+.. testcode::
+
+    Extreme5000.combination = Instrument.control(
+        ":VOLTFREQ?", ":VOLTFREQ %g,%g",
+        """Simultaneously control the voltage in Volts and the frequency in Hertz (both float).
+
+        This property is set by a tuple.
+        """
+    )
+
+In use, we could set the voltage to 200 mV, and the Frequency to 931 Hz, and read both values immediately afterwards. 
+
+.. doctest::
+
+    >>> extreme = Extreme5000("GPIB::1")
+    >>> extreme.combination = (0.2, 931)        # Executes ":VOLTFREQ 0.2,931"
+    >>> extreme.combination                     # Reads ":VOLTFREQ?"
+    [0.2, 931.0]
+
+This interface is not too convenient, but luckily not often needed.
 
 Dynamic properties
-===================
+******************
 
 As described in previous sections, Python properties are a very powerful tool to easily code an instrument's programming interface.
 One very interesting feature provided in PyMeasure is the ability to adjust properties' behaviour in subclasses or dynamically in instances.
@@ -616,10 +613,8 @@ Pay attention *not* to inadvertently define other class attribute or instance at
 
 The mechanism works for all the parameters in properties, except :code:`dynamic` and :code:`docs` -- see :func:`Instrument.control <pymeasure.instruments.Instrument.control>`, :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>`, :func:`Instrument.setting <pymeasure.instruments.Instrument.setting>`.
 
-Let us now consider a couple of common use cases for this functionality:
-
 Dynamic validity range
-**********************
+----------------------
 Let's assume we have an instrument with a command that accepts a different valid range of values depending on its current state.
 The code below shows how this can be accomplished with dynamic properties.
 
@@ -658,6 +653,10 @@ Instruments with similar features
 =================================
 
 When instruments have a similar set of features, it makes sense to use inheritance to obtain most of the functionality from a parent instrument class, instead of copy-pasting code.
+
+.. note::
+    Don't forget to update the instrument's :code:`name` attribute accordingly, by either supplying an appropriate argument (if available) during the :code:`super().__init__()` call, or by setting it anew below that call.
+
 Sometimes one only needs to add additional properties and methods.
 Often, some of the already present properties/methods need to be completely replaced by defining them again in the derived class.
 Often, however, only some details need to be changed.
--- a/docs/dev/adding_instruments.rst
+++ b/docs/dev/adding_instruments.rst
@@ -112,9 +112,26 @@ There is a certain order of elements in an instrument class that is useful to ad
 * Then properties (pymeasure-specific or generic Python variants). This will be the bulk of the implementation.
 * Finally, any methods.
 
-In principle you are free to write any functions that are necessary for interacting with the instrument. When doing so, make sure to use the :code:`self.ask(command)`, :code:`self.write(command)`, and :code:`self.read()` methods to issue commands instead of calling the adapter directly. If the communication requires changes to the commands sent/received, you can override these methods in your instrument, for further information see advanced_communication_protocols_.
+Your instrument's user interface
+================================
 
-In practice, we have developed a number of convenience functions for making instruments easy to write and maintain. The following sections detail these conveniences and are highly encouraged.
+Your instrument will have a certain set of properties and methods that are available to a user and discoverable via the documentation or their editor's autocomplete function.
+
+In principle you are free to choose how you do this (with the exception of standard SCPI properties like :code:`id`).
+However, there are a couple of practices that have turned out to be useful to follow:
+
+* Naming things is important. Try to choose clear, expressive, unambiguous names for your instrument's elements.
+* If there are already similar instruments in the same "family" (like a power supply) in pymeasure, try to follow their lead where applicable. It's better if, e.g., all power supplies have a :code:`current_limit` instead of an assortment of :code:`current_max`, :code:`Ilim`, :code:`max_curr`, etc.
+* If there is already an instrument with a similar command set, check if you can inherit from that one and just tweak a couple of things. This massively reduces code duplication and maintenance effort. The section :ref:`family_of_instruments_with_similar_features` shows how to achieve that.
+* The bulk of your instrument's interface will probably be made up of properties for quantities to set and/or read out. Our custom properties (see :ref:`writing_properties` ff. below) offer some convenience features and are therefore preferable, but plain Python properties are also fine.
+* "Actions", commands or verbs should typically be methods, not properties: :code:`recall()`, :code:`trigger_scan()`, :code:`prepare_resistance_measurement()`, etc.
+* This separation between properties and methods also naturally helps with observing the `"command-query separation" principle <https://en.wikipedia.org/wiki/Command%E2%80%93query_separation>`__.
+* If your instrument has multiple identical channels, see XXX. TODO: write section on channel implementations
+
+
+In principle you are free to write any methods that are necessary for interacting with the instrument. When doing so, make sure to use the :code:`self.ask(command)`, :code:`self.write(command)`, and :code:`self.read()` methods to issue commands instead of calling the adapter directly. If the communication requires changes to the commands sent/received, you can override these methods in your instrument, for further information see advanced_communication_protocols_.
+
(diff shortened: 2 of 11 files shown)


bilderbuchi:
I'll try to knock out some of the items in #102 while I'm at it.

docs/quick_start.rst
```diff
@@ -42,11 +42,18 @@ PyMeasure can also be installed with :code:`pip`.
  
 Depending on your operating system, using this method may require additional work to install the required dependencies, which include the Qt libaries.
 
-    
+Installing PyVISA
+-----------------
+Typically, communication with your instrument will happen using PyVISA, which is installed automatically.
+However, you will need a `VISA backend <https://pyvisa.readthedocs.io/en/latest/introduction/configuring.html>`__ installed to handle device communication.
+You may already know what this means and have an appropriate backend (e.g. from NI, Keysight, R&S,...) available.
```
msmttchr, line 49: I find this part a little confusing, you don't know which backend to install (pyvisa-py or NI, etc). What about simply redirecting users to pyvisa documentation and remove lines 49-50 ?

msmttchr (commented):
Updating docs is very useful. Thanks.
Few comments from my side.

docs/dev/adding_instruments.rst
```diff
@@ -105,6 +105,13 @@ This is a minimal instrument definition:
 
 Make sure to include the PyMeasure license to each file, and add yourself as an author to the :code:`AUTHORS.txt` file.
```
bmoneke, line 106:
```suggestion
Make sure to include the PyMeasure license to each file, and add yourself as an author to the :code:`AUTHORS.txt` and :code:`docs/about/authors.rst` files.
```
By chance, I saw this one as well.

docs/quick_start.rst
bmoneke: As a user, I'd prefer to not look up another documentation, but we don't want to repeat their documentation.
Therefore, I propose to keep it simple saying:
```suggestion
For example you can use the python package :code:`pyvisa-py`.
```

docs/quick_start.rst
```diff
@@ -42,11 +42,18 @@ PyMeasure can also be installed with :code:`pip`.
...
+If not, you can get off the ground easily by installing the pure-Python :code:`pyvisa-py` backend package (using the same installation you used above).
+
 Checking the version
 --------------------
 
-Now that you have Python and PyMeasure installed, open up a "Jupyter Notebook" to test which version you have installed. Execute the following code into a notebook cell.
+Now that you have Python and PyMeasure installed, open up a REPL or Jupyter notebook to test which version you have installed.
+Execute the following Python code.
```
bmoneke, line 56: Shouldn't we keep it more general?
```suggestion
Now that you have Python and PyMeasure installed, open your python environment (a REPL or Jupyter notebook) to test which version you have installed.
Execute the following Python code.
```

docs/dev/adding_instruments.rst
bilderbuchi: I think I'll see if I can eliminate this nowadays needless duplication in this PR, instead.

bmoneke: That would be even better. Thanks.

docs/dev/adding_instruments.rst
```diff
@@ -105,6 +105,13 @@ This is a minimal instrument definition:
...
 
+There is a certain order of elements in an instrument class that is useful to adhere to:
+
+* First, the initializer (the :code:`__init__()` method), this makes it faster to find when browsing the source code.
+* Then class attributes/variables, if you need them.
+* Then properties (pymeasure-specific or generic Python variants). This will be the bulk of the implementation.
+* Finally, any methods.
```
bmoneke, line 113: Let us mention somewhere (I think it is fitting here or about properties), that it is perfectly fine to define methods. For example:
```suggestion
* Finally, any methods. 'Actions' like :code:`recall()` or :code:`trigger_scan()` may find their place here.
```

bilderbuchi: I have some drafts for dealing with this already locally.

bmoneke, line 113: It is one option, to separate the properties and methods according to their type.

I, on the other hand, prefer to sort thematically, e.g. I'd add a `start_scan()` method to the `scan_speed`, `scan_range`, and `scan_center` properties.

I guess we can leave it, how you wrote it, because these methods (like `start_scan()`) are rare and can be an exception to this rule and stay with their buddy properties.

bilderbuchi: I see your point. I can emphasize that it's a guideline, not a strict policy. If people start thinking about the order then it's already accomplished most of what I want. :-)

docs/dev/adding_instruments.rst
```diff
@@ -105,6 +105,13 @@ This is a minimal instrument definition:
...
+
```
bmoneke, line 114: Maybe this is also the place for hints about good naming patterns.
```suggestion

It's good to name properties self explaining, short, and descriptive. `voltage_setpoint`, for example, would be good for the output voltage of a voltage source, which can measure a voltage as well. Verbs are a bad choice, because a property is no action (even though it may require one to acquire the data), so the measured voltage property should be called `measured_voltage` or just `voltage`, but not `measure_voltage`.
Similarly, if you want the device to perform some action (starting/stopping something etc.), these methods (mostly without parameters and return values) should use verbs for their name, e.g. `start_scan()`.

```

docs/dev/adding_instruments.rst
bmoneke: You wrote, that it is a useful order. Therefore it is marked as a suggestion. Keep it as is.

bmoneke:
I looked at some points in #102 and some of my thoughts (from reviews) and made suggestions.

One more point. Should we mention, that a property docstring does not need (and should not) tell the string returned by the device? I did not find a good point, therefore the comment is here.

Finally, do you want to include #705 as well or leave it for another day? The content does fit in here, but might delay the rework and can be done separately as well. (I propose to keep them separated for faster finishing this PR)

docs/dev/adding_instruments.rst
bilderbuchi: yes, as I said, I have some stuff locally, but it was too late yesterday. :-P
I'll set this PR to WIP or something :D

bmoneke: They're just some suggestions, that you do not have to write everything yourself.

docs/quick_start.rst
bilderbuchi: I think we should give enough context, and for that we should not ignore the proprietary backends. 

I'll try to improve this. I thought it clear, simple and complete enough already. Experienced users will know the deal when they hear VISA backend, and others get simple instructions without having to read elsewhere (but the link is there if desired).

docs/dev/adding_instruments.rst
bilderbuchi: This is now much easier than when I tried this last time. Not super beautiful because it's a literal block instead of a bullet list, but good enough imo.

docs/dev/adding_instruments.rst
bilderbuchi: Done.

bilderbuchi:
> One more point. Should we mention, that a property docstring does not need (and should not) tell the string returned by the device? I did not find a good point, therefore the comment is here.

I fear I don't understand what you mean. Can you rephrase?

I'll leave #705 out for now, we can do it later.

docs/quick_start.rst
bmoneke: It's good to hint at other backends and proposing (as an example) pyvisa-pay.

It's good with me

docs/dev/adding_instruments.rst
bmoneke: Oh, great expansion. Thanks

bmoneke:
> I fear I don't understand what you mean. Can you rephrase?

In one of the current PRs, the docstring said something like "this is some property. The device responds 'here is my response with value 5',therefore get_process is useful".

I think it is not necessary to expose the device communication protocol (the string sent / received) in the corresponding property.

bilderbuchi:
> In one of the current PRs, the docstring said something like "this is some property. The device responds 'here is my response with value 5',therefore get_process is useful".
>
> I think it is not necessary to expose the device communication protocol (the string sent / received) in the corresponding property.

Done.

bilderbuchi:
I'm done now, another review round is appreciated! :smile:

docs/tutorial/connecting.rst
```diff
@@ -46,7 +57,13 @@ When the :code:`with`-block is exited, the :code:`shutdown` method of the instru
...
 ==============
 
-PyMeasure supports a number of adapters, which are responsible for communicating with the underlying hardware. In the example above, we passed the string "GPIB::4" when constructing the instrument. By default this constructs a VISAAdapter class to connect to the instrument using VISA. Instead of passing a string, we could equally pass an adapter object. ::
+PyMeasure supports a number of adapters, which are responsible for communicating with the underlying hardware.
+In the example above, we passed the string "GPIB::4" when constructing the instrument.
+By default this constructs a VISAAdapter (our most popular, default adapter) to connect to the instrument using VISA.
+Passing a string (or integer in case of GPIB) is by far the most typical way to create pymeasure instruments.
+
+Sometimes, you might need to go beyond the usual setup, which is also possible.
+Instead of passing a string, you could equally pass an adapter object. ::
```
msmttchr, line 66: Comment not related to this line, I add here because is the closest point where I can add it.

In my opinion, all discussion below related to `Lakeshore425` should be removed, since we no longer use this approach.
After the adapters restructuring done by @bmoneke, the only use case for using a custom made adapter is when instrument is using a phy layer protocol not currently supported by pyvisa, e.g. PCIexpress or SPI.

docs/api/instruments/validators.rst
```diff
@@ -4,7 +4,7 @@
 Validator functions
 ###################
 
-Validators are used in conjunction with the :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function to allow properties with complex restrictions for valid values. They are described in more detail in the :ref:`Advanced properties <advanced-properties>` section.
+Validators are used in conjunction with the :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function to allow properties with complex restrictions for valid values. They are described in more detail in the :ref:`validators` section.
```
msmttchr, line 7: I think validators can be also used with `setting` function, which is a particular use case of `control`.

docs/quick_start.rst
```diff
@@ -42,11 +42,18 @@ PyMeasure can also be installed with :code:`pip`.
  
 Depending on your operating system, using this method may require additional work to install the required dependencies, which include the Qt libaries.
 
-    
+Installing VISA
```
msmttchr, line 45: Small typo at line 43 libaries

msmttchr (commented):
Few comments from my side.

docs/quick_start.rst
bilderbuchi: Thanks!

docs/tutorial/connecting.rst
bilderbuchi: Done in 0cfc003.

docs/dev/adding_instruments.rst
```diff
@@ -442,6 +447,38 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+    >>> extreme.read()
+    '1'
+    >>> extreme.output_enabled = False
+    >>> extreme.output_enabled
+    False
+    >>> # Invalid input raises an exception
+    >>> extreme.output_enabled = 34
+    Traceback (most recent call last):
+    ...
+    KeyError: 34
```
bilderbuchi, line 478: So, I realised that we don't need the `strict_discrete_set` validator for the boolean property -- if an invalid input is supplied, you get a KeyError from the dict lookup as shown in the example.

Do you think we should leave it in the docs like this, or with the `strict_discrete_set` added, as the error message might be clearer/more useful?

docs/dev/adding_instruments.rst
```diff
@@ -105,9 +106,48 @@ This is a minimal instrument definition:
...
+Note that not all already available instruments follow these, but expect this to be harmonized in the future.
+
+Frequent properties
+-------------------
+If your instrument has an **output** that can be switched on and off, use a :ref:`boolean property <boolean-properties>` called :code:`output_enabled`.
+
+Power supplies
+--------------
+PSUs typically can measure the *actual* current and voltage, as well as have settings for the voltage level and the current limit.
+To keep naming clear and avoid confusion, implement the properties :code:`current`, :code:`voltage`, :code:`voltage_setpoint` and :code:`current_limit`, respectively.
```
bilderbuchi, line 150: Here, I have started to document things I repeat again and again during reviews. I hope with this we can start shaping our instrument APIs toward homogeny before we get around to implement #416. Feel free to extend this in the future.

docs/dev/adding_instruments.rst
bmoneke: I personally prefer cleaner code (with less additions).

However, a validator ensures the right values. 

We could add, that a mapping dictionary works as a validator as well (maybe even transform the exception in `values` from KeyError to Value Error("Invalid value supplied for mapping") or something.

This should enter the docstring and documentation, I'd say.

docs/dev/adding_instruments.rst
bmoneke: Good idea

docs/dev/adding_instruments.rst
```diff
@@ -187,92 +227,78 @@ In addition, any entries in ``**kwargs**`` take precedence, so if they need to,
...
+        """Control the voltage in Volts (float)."""
     )
 
-You will notice that we use the `Python string format`_ :code:`%g` to pass through the floating point.
+You will notice that we use the `Python string format`_ :code:`%g` to format passed-through values as floating point.
 
 .. _Python string format: https://docs.python.org/3/library/string.html#format-specification-mini-language
 
-We can use this property to set the voltage to 100 mV, which will execute the command and then request the current voltage.
+We can use this property to set the voltage to 100 mV, which will sends the appropriate command and then requests the current voltage.
```
bmoneke, line 283:
```suggestion
We can use this property to set the voltage to 100 mV, which will send the appropriate command and then requests the current voltage.
```

docs/dev/coding_standards.rst
```diff
@@ -30,13 +30,33 @@ Type documentation should be placed in the docstring where not clear from the va
...
+It is advisable to follow the `PEP257 <https://peps.python.org/pep-0257/>`_ docstring guidelines.
+Most importantly:
+
+* Use triple-quoted strings (:code:`"""`) to delimit docstrings.
+* One short summary line in imperative voice, with a period at the end.
+* Optionally, after a blank line, include more detailed information.
+* For functions and methods, you can add documentation on their parameters using the `reStructuredText docstring format <https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists>`__.
+
+Specific to our properties, start them with "Control", "Measure" or "Set" to indicate the kind of property (this information is not visible after import).
+In addition, it is useful to add type and information about :ref:`validators` (if applicable) at the end of the summary line, see the docstrings shown in examples throughout the :ref:`adding-instruments` section.
```
bmoneke, line 59: I like it. This prevents the long comment "this property can be read and set".

docs/dev/adding_instruments.rst
```diff
@@ -105,9 +106,48 @@ This is a minimal instrument definition:
...
+* "Actions", commands or verbs should typically be methods, not properties: :code:`recall()`, :code:`trigger_scan()`, :code:`prepare_resistance_measurement()`, etc.
+* This separation between properties and methods also naturally helps with observing the `"command-query separation" principle <https://en.wikipedia.org/wiki/Command%E2%80%93query_separation>`__.
+* If your instrument has multiple identical channels, see XXX. TODO: write section on channel implementations
+
+In principle you are free to write any methods that are necessary for interacting with the instrument. When doing so, make sure to use the :code:`self.ask(command)`, :code:`self.write(command)`, and :code:`self.read()` methods to issue commands instead of calling the adapter directly. If the communication requires changes to the commands sent/received, you can override these methods in your instrument, for further information see advanced_communication_protocols_.
```
bmoneke, line 132:
```suggestion
In principle, you are free to write any methods that are necessary for interacting with the instrument. When doing so, make sure to use the :code:`self.ask(command)`, :code:`self.write(command)`, and :code:`self.read()` methods to issue commands instead of calling the adapter directly. If the communication requires changes to the commands sent/received, you can override these methods in your instrument, for further information see advanced_communication_protocols_.
```

docs/dev/adding_instruments.rst
```diff
@@ -187,92 +227,78 @@ In addition, any entries in ``**kwargs**`` take precedence, so if they need to,
...
 
-In PyMeasure, `Python properties`_ are the preferred method for dealing with variables that are read or set. 
-PyMeasure comes with two convenience functions for making properties for classes. 
-The :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` function returns a property that issues a GPIB/SCPI requests when the value is used. 
-For example, if our "Extreme 5000" has the :code:`*IDN?` command we can write the following property to be added above the :code:`def __init__` line in our above example class, or added to the class after the fact as in the code here:
+In PyMeasure, `Python properties`_ are the preferred method for dealing with variables that are read or set.
+
+The property factories
+**********************
+PyMeasure comes with three central convenience factory functions for making properties for classes: :func:`Instrument.control <pymeasure.instruments.Instrument.control>`, :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` and :func:`Instrument.setting <pymeasure.instruments.Instrument.setting>`.
```
bmoneke, line 239: In English, you have a comma before "and" in an enumeration.
```suggestion
PyMeasure comes with three central convenience factory functions for making properties for classes: :func:`Instrument.control <pymeasure.instruments.Instrument.control>`, :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>`, and :func:`Instrument.setting <pymeasure.instruments.Instrument.setting>`.
```

docs/dev/adding_instruments.rst
```diff
@@ -187,92 +227,78 @@ In addition, any entries in ``**kwargs**`` take precedence, so if they need to,
...
+
+The :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` function returns a property that can only read values from an instrument.
+For example, if our "Extreme 5000" has the :code:`*IDN?` command we can write the following property to be added after the :code:`def __init__` line in our above example class, or added to the class after the fact as in the code here:
```
bmoneke, line 242:
```suggestion
For example, if our "Extreme 5000" has the :code:`*IDN?` command, we can write the following property to be added after the :code:`def __init__` line in our above example class, or added to the class after the fact as in the code here:
```

docs/dev/adding_instruments.rst
```diff
@@ -187,92 +227,78 @@ In addition, any entries in ``**kwargs**`` take precedence, so if they need to,
...
+        """Control the voltage in Volts (float)."""
     )
 
-You will notice that we use the `Python string format`_ :code:`%g` to pass through the floating point.
+You will notice that we use the `Python string format`_ :code:`%g` to format passed-through values as floating point.
 
 .. _Python string format: https://docs.python.org/3/library/string.html#format-specification-mini-language
 
-We can use this property to set the voltage to 100 mV, which will execute the command and then request the current voltage.
+We can use this property to set the voltage to 100 mV, which will send the appropriate command and then requests the current voltage.
```
bmoneke, line 283: Sorry for correcting again this sentence to make it more clear, that you describe two different actions of this `control`:
```suggestion
We can use this property to set the voltage to 100 mV, which will send the appropriate command, and then to request the current voltage:
```

docs/dev/adding_instruments.rst
```diff
@@ -187,92 +227,78 @@ In addition, any entries in ``**kwargs**`` take precedence, so if they need to,
...
 
 .. doctest::
 
     >>> extreme = Extreme5000("GPIB::1")
-    >>> extreme.voltage = 0.1        # Executes ":VOLT 0.1"
-    >>> extreme.voltage              # Reads ":VOLT?"
+    >>> extreme.voltage = 0.1        # Sends ":VOLT 0.1"
```
bmoneke, line 288: Mirroring the second command.
```suggestion
    >>> extreme.voltage = 0.1        # Sends ":VOLT 0.1" to set the voltage to 100 mV
```

docs/dev/adding_instruments.rst
```diff
@@ -187,92 +227,78 @@ In addition, any entries in ``**kwargs**`` take precedence, so if they need to,
...
+    >>> extreme.voltage              # Sends ":VOLT?" to query for the current value
     0.1
 
-Using :func:`Instrument.control <pymeasure.instruments.Instrument.control>` and :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` functions, you can create a number of properties for basic measurements and controls. 
+Finally, the :func:`Instrument.setting <pymeasure.instruments.Instrument.setting>` function can only set, but not read values.
 
-The :func:`Instrument.control <pymeasure.instruments.Instrument.control>` function can be used with multiple values at once, passed as a tuple. Say, we may set voltages and frequencies in our "Extreme 5000", and the the commands for this are :code:`:VOLTFREQ?` and :code:`:VOLTFREQ <float>,<float>`, we could use the following property:
+Using the :func:`Instrument.control <pymeasure.instruments.Instrument.control>`, :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>` and :func:`Instrument.control <pymeasure.instruments.Instrument.control>` functions, you can create a number of properties for basic measurements and controls.
```
bmoneke, line 294:
```suggestion
Using the :func:`Instrument.control <pymeasure.instruments.Instrument.control>`, :func:`Instrument.measurement <pymeasure.instruments.Instrument.measurement>`, and :func:`Instrument.control <pymeasure.instruments.Instrument.control>` functions, you can create a number of properties for basic measurements and controls.
```

docs/dev/adding_instruments.rst
```diff
@@ -187,92 +227,78 @@ In addition, any entries in ``**kwargs**`` take precedence, so if they need to,
...
 
-.. testcode::
+The next sections detail additional features of :func:`Instrument.control <pymeasure.instruments.Instrument.control>` that allow you to write properties that cover specific ranges, or have to map between a real value to one used in the command. Furthermore it is shown how to perform more complex processing of return values from your device.
```
bmoneke, line 296: including the derivatives as well or does the user understand, that `measurement` and `setting` work the same?
```suggestion
The next sections detail additional features of :func:`Instrument.control <pymeasure.instruments.Instrument.control>` and derivatives that allow you to write properties that cover specific ranges, or have to map between a real value to one used in the command. Furthermore it is shown how to perform more complex processing of return values from your device.
```

docs/dev/adding_instruments.rst
```diff
@@ -442,6 +462,40 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+
+The idea of using maps can be leveraged to implement properties where the user-facing values are booleans, so you can interact in a pythonic way using :code:`True` and :code:`False`:
+
+.. testcode::
+
+    Extreme5000.output_enabled = Instrument.control(
+        "OUTP?", "OUTP %d",
+        """Control the instrument output is enabled (boolean).""",
+        map_values=True,
+        values={True: 1, False: 0},  # the dict values could also be "on" and "off", etc.
```
bmoneke, line 478:
```suggestion
        values={True: 1, False: 0},  # the dict values could also be "on" and "off", etc. depending on the device
```

docs/dev/adding_instruments.rst
```diff
@@ -442,6 +462,40 @@ The dictionary now maps the keys to specific values. The values and keys can be
...
+
+.. doctest::
+
+    >>> extreme = Extreme5000("GPIB::1")
+    >>> extreme.output_enabled = True
+    >>> extreme.read()
+    '1'
+    >>> extreme.output_enabled = False
+    >>> extreme.output_enabled
+    False
```
bmoneke, line 490: How can you read without prior writing?
I guess you ment to show the raw communication.
```suggestion
    >>> extreme.write("OUTP?")
    >>> extreme.read()  # read the response of the device prior to mapping.
    '1'
    >>> extreme.output_enabled = False
    >>> extreme.output_enabled  # read the mapped value
    False
```

docs/dev/adding_instruments.rst
```diff
@@ -531,15 +580,56 @@ The same can be also achieved by the `preprocess_reply` keyword argument to :fun
...
+************************
+If you need to tweak
+
+* the :code:`set_command` string immediately before the value to set is inserted via string formatting (:code:`%g` etc.), or
+* the :code:`get_command` string before sending it to the device,
+
+use the :code:`command_process` parameter of :meth:`~pymeasure.instruments.Instrument.control`.
+
+Note that there is only parameter for both setting and getting, so the utility of this is probably limited.
+Note also that for adding e.g. channel identifiers, there are other, more preferable methods.
```
bmoneke, line 598:
```suggestion
Note, that there is only one parameter for both setting and getting, so the utility of this is probably limited.
Note also, that for adding e.g. channel identifiers, there are other, more preferable methods.
```

docs/dev/adding_instruments.rst
```diff
@@ -590,10 +677,26 @@ The code below shows how this can be accomplished with dynamic properties.
 
 
 Now our voltage property has a dynamic validity range, either [-1, 1] or [0, 1].
+A side effect of this is that the property's docstring should be less specific, to avoid it containing dynamically changed information (like the admissible value range).
```
bmoneke, line 680:
```suggestion
A side effect of this is, that the property's docstring should be less specific, to avoid it containing dynamically changed information (like the admissible value range).
```

docs/dev/adding_instruments.rst
```diff
@@ -590,10 +677,26 @@ The code below shows how this can be accomplished with dynamic properties.
...
+
+When instruments have a similar set of features, it makes sense to use inheritance to obtain most of the functionality from a parent instrument class, instead of copy-pasting code.
+
+.. note::
+    Don't forget to update the instrument's :code:`name` attribute accordingly, by either supplying an appropriate argument (if available) during the :code:`super().__init__()` call, or by setting it anew below that call.
+
+Sometimes one only needs to add additional properties and methods.
+Often, some of the already present properties/methods need to be completely replaced by defining them again in the derived class.
+Often, however, only some details need to be changed.
+This can be dealt with efficiently using dynamic properties.
```
bmoneke, line 696: I'm not happy with the duplication of "often"
```suggestion
In some cases, one only needs to add additional properties and methods.
In other cases, some of the already present properties/methods need to be completely replaced by defining them again in the derived class.
Often, however, only some details need to be changed.
This can be dealt with efficiently using dynamic properties.
```

docs/dev/contribute.rst
```diff
@@ -76,22 +76,22 @@ Now you are ready to write your new feature and make changes to the code. To ens
     git commit -m "A short description about what changed"
     git push
 
-Making a pull-request
+Making a pull request
 =====================
 
-While you are working, its helpful to start a pull-request (PR) targeting the :code:`master` branch of :code:`pymeasure/pymeasure`. This will allow you to discuss your feature with other contributors. We encourage you to start this pull-request after your first commit.
+While you are working, its helpful to start a pull request (PR) targeting the :code:`master` branch of :code:`pymeasure/pymeasure`. This will allow you to discuss your feature with other contributors. We encourage you to start this pull-request after your first commit.
```
bmoneke, line 82: Should we mention the possibilities of a draft (besides to typos in this section)?
```suggestion
While you are working, it is helpful to start a pull request (PR) targeting the :code:`master` branch of :code:`pymeasure/pymeasure`. This will allow you to discuss your feature with other contributors. We encourage you to start this pull request already after your first commit.
You may mark a pull request as a draft, if it is in an early state.
```

docs/tutorial/connecting.rst
```diff
@@ -11,10 +13,21 @@ First import the instrument of interest. ::
 
     from pymeasure.instruments.keithley import Keithley2400
 
-Then construct an object by passing the GPIB address. For this example we connect to the instrument over GPIB (using VISA) with an address of 4. See the :ref:`adapters <adapters>` section below for more details. ::
+Then construct an object by passing the GPIB address. For this example we connect to the instrument over GPIB (using VISA) with an address of 4::
```
bmoneke, line 16: The first sentence is valid for all VISA addresses, afterwards adapted to the example.
```suggestion
Then construct an object by passing the VISA address. For this example we connect to the instrument over GPIB (using VISA) with an address of 4::
```

bmoneke (commented):
I went through the texts and marked some missing commas or suggested different wording, where I thought them helpful.

Thanks a lot for this great improvement of the documentation.

bmoneke:
An additional idea:
We could add an example with an IntEnum/StrEnum instead of mapping with a dictionary.

docs/dev/adding_instruments.rst
bilderbuchi: Well, in [Oxford](https://en.wikipedia.org/wiki/Serial_comma), at least. :wink: 
(for the record, I also use the Oxford comma, just missed that one)

docs/dev/adding_instruments.rst
bilderbuchi: No worries, it's clearer.

docs/dev/adding_instruments.rst
bilderbuchi: I don't agree with all comma suggestions, this one for example I'm pretty sure is non appropriate.

docs/dev/adding_instruments.rst
bilderbuchi: Noted, I instead used the term "property factories" from the previous heading.

docs/dev/adding_instruments.rst
bilderbuchi: No comma here, either, "note that the sky is blue" is one phrase. I added the "one", though. :-P

docs/dev/adding_instruments.rst
bilderbuchi: Indeed. I think this is a bit of an artifact of how our faked instrument works like, and occurs a couple of times more.

bilderbuchi: This will need some deeper changes if we want it to be more representative -- the first line `extreme.output_enabled = True` basically puts a (mapped) `1` into a primitive buffer in `FakeAdapter`, which the `read` consumes without mapping.

If we can't easily adapt `FakeAdapter` to give more meaningful (doctestable) interaction, maybe the easiest thing would be to define a pyvisa-sim device for the handful of examples that use that misleading raw `read()` approach.

docs/dev/adding_instruments.rst
bilderbuchi: > We could add, that a mapping dictionary works as a validator as well (maybe even transform the exception in values from KeyError to Value Error("Invalid value supplied for mapping") or something.

That's probably more confusing to users. Let's add the `strict_discrete_set` back, the error message is clearer, and semantically it's also clearer that the inputs are being validated (even if they would also be "validated" in a way by the subsequent dict lookup).

docs/quick_start.rst
bilderbuchi: @msmttchr good for you, too, now?

bilderbuchi:
> An additional idea: We could add an example with an IntEnum/StrEnum instead of mapping with a dictionary.

Sounds good! Do you have the time to formulate and suggest one? If not, I'll 
* [ ] try to come up with something
* [ ] Also, an `IntFlagEnum` example for composing bitmask flags could be useful, that's a pattern that's not that obvious to newcomers.

docs/quick_start.rst
msmttchr: Yes, absolutely.

docs/dev/adding_instruments.rst
bmoneke: You're right.

bmoneke:
> > An additional idea: We could add an example with an IntEnum/StrEnum instead of mapping with a dictionary.
> 
> Sounds good! Do you have the time to formulate and suggest one? If not, I'll
> 
> * [ ]  try to come up with something
> * [ ]  Also, an `IntFlagEnum` example for composing bitmask flags could be useful, that's a pattern that's not that obvious to newcomers.

I got this idea during a review of #732, but did not try it myself.
In `setting`, an e um will work out of the box, but for `measurement`, we have to adjust `control`, I think.
A first implementation idea adds an option to define the Enum, which will interpret the return value. Another idea is, that the control does this check, if the  `values` parameter is an Enum.

I have to fiddle a bit to find a good solution.

Let's keep it for another pr.

docs/dev/adding_instruments.rst
bilderbuchi: I put the validator in now. Agreed to do it that way?
(Note: I know that the error message talks about a set but displays a dict, but that's a battle for another day)

docs/dev/adding_instruments.rst
bilderbuchi: ping @msmttchr @CasperSchippers : only so that you are aware of this existing, now.

docs/dev/adding_instruments.rst
bilderbuchi: I'll bump this to another issue, we have been doing this for a long time that way and nobody noticed, and it will just drag this PR out much longer :D

bilderbuchi: I'll resolve this in favour of #735.

docs/dev/adding_instruments.rst
bmoneke: In the end, it is an example how you can do it. With the validator in the example, the readers will see how to validate correctly.

So, I'm ok with the validator.

docs/dev/adding_instruments.rst
bmoneke: Ok, resolve it. Somebody has to know the code to find the error, but those don't need the example for the basic communication.

bilderbuchi:
Enum example added in [37c99f5](https://github.com/pymeasure/pymeasure/pull/712/commits/37c99f5320d7e699bbc06cb3b465be8eec9823dd). I don't expect this to need another review, and everything else is resolved. I will merge tomorrow if nothing unexpected happens.

docs/dev/adding_instruments.rst
```diff
@@ -149,6 +149,60 @@ Power supplies
...
+    ...     COOLER_FAILURE = 2
+    ...     HEATER_FAILURE = 1
+    ...     OK = 0
+    ...
+    >>> received_from_device = 7
+    >>> print(ErrorCode(received_from_device))
+    ErrorCode.TEMPSENSOR_FAILURE|COOLER_FAILURE|HEATER_FAILURE
+
+:code:`IntFlags` are used by many instruments for the purpose just demonstrated.
+
```
bmoneke, line 205: Let's include a suggestion how to implement it:
```suggestion

The status property could look like this:
.. testcode::

    status = Instrument.measurement(
        "STB?", 
        """Measure the status of the device as enum.""",
        get_process=lambda v: ErrorCode(v), 
   )
```

docs/dev/adding_instruments.rst
```diff
@@ -149,6 +149,60 @@ Power supplies
 PSUs typically can measure the *actual* current and voltage, as well as have settings for the voltage level and the current limit.
 To keep naming clear and avoid confusion, implement the properties :code:`current`, :code:`voltage`, :code:`voltage_setpoint` and :code:`current_limit`, respectively.
 
+Managing status codes or other indicator values
+***********************************************
+Often, an instrument features one or more collections of specific values that signal some status, an instrument mode or a number of possible configuration values.
+Typically, these are collected in mappings of some sort, as you want to provide a clear and understandable value to the user, while abstracting away the raw data, think :code:`ACQUISITION_MODE` instead of :code:`0x04`.
+The mappings normally are kept at module level (i.e. not defined within the instrument class), so that they are available when using the property factories.
```
bmoneke, line 156: Can't you access them at class level (e.g. AwG401.SomeEnum instead of self.SomeEnum)?
Because for the setter a class enum is better.

Or (for a `control`), the class may contain the module level enum.

Anyway, we can adjust the documentation according to upcoming experience.

bmoneke (commented):
Some ideas regarding the enum text.

Maybe we should make some tests and wait for some experience and then expand the enum part (your general introduction is good) with that experience (I. E. Drop my ideas for now).

docs/dev/adding_instruments.rst
bilderbuchi: LGTM!

docs/dev/adding_instruments.rst
bilderbuchi: In my experience this does not work reliably (also for lambdas, etc), but we can sure test in a more systematic way. I propose to do that another time. Changing the docs to an improved approach should always be welcome. :-)

docs/dev/adding_instruments.rst
bmoneke: I just found an even better implementation:

`get_process=ErrorCode`

Sorry for the change, but we do not need a lambda for calling a method 😁

docs/dev/adding_instruments.rst
bilderbuchi: See #736

PR #718, opened 2022-10-22
Channel implementation added.

bmoneke:
Fixes #616 which complains about missing Channel implementation.

I'm done with the basic implementation:
- Creation of some `Base` class (name is open for suggestions), which contains properties and methods which work inside the class itself (values, ask...). 
- Instrument is based on this Base and calls the adapter. 
- The `Channel` (maybe a different name, I made it similar to Instrument) calls the instrument's methods (instead of adapter) and `write` uses `str.format` to replace '{ch}' in a command with the channel name. I figured, this is the most common use case for a channel. If the command does not contain that variable, it is used 'as is'.

Documentation (docstring, and documentation proper, including example) and tests are not yet done.

ToDos after the implementation itself is done:

- [x] Write proper documentation (including protected "channels"). Example how to add "special" channels, how to always prepend Channel name... You can use Channel.control etc.
- [x] You should use the default prefix and "collection" name, unless you have several different ones. Always use asd_child
- [x] Add a remark in changelog about protected "channels" and "ch_{ch}"
- [x] Reorder the tests according to the split file.
- [x] ensure everything from #364 is done
- [x] add remove_child
- [x] put the formatting in its own method
- [x] add channel descriptors in order to define the children at class level.
- [x] protect auto created children. 
- [x] Adjust the AWG to the code.
- [x] Tests for setup methods (if they remain)
- [x] create an issue to change instruments to this new implementation.

Closes #364

PR #719, opened 2022-10-25
New Instrument: Siglent SPDxxxxX series Power Supplies

AidenDawn:
Add new instrument, SPD1168X power supply with base and channel class for the SPD power supply family. #718 would mean the channel implementation here could rely on a different base class but this works for now. The only unimplemented function is currently the ability to set the timer steps of the PSU, this requires setting a list of values in the same command which I couldn't find an example for. Any pointers to one would be great so I can add that last property

### DIFF
--- a/pymeasure/instruments/siglenttechnologies/__init__.py
+++ b/pymeasure/instruments/siglenttechnologies/__init__.py
@@ -22,6 +22,5 @@
 # THE SOFTWARE.
 #
 
-from .siglent_spdbase import SPDBase
 from .siglent_spd1168x import SPD1168X
 from .siglent_spd1305x import SPD1305X
--- a/pymeasure/instruments/siglenttechnologies/__init__.py
+++ b/pymeasure/instruments/siglenttechnologies/__init__.py
@@ -24,3 +24,4 @@
 
 from .siglent_spdbase import SPDBase
 from .siglent_spd1168x import SPD1168X
+from .siglent_spd1305x import SPD1305X
--- a/pymeasure/instruments/siglenttechnologies/__init__.py
+++ b/pymeasure/instruments/siglenttechnologies/__init__.py
@@ -22,43 +22,5 @@
 # THE SOFTWARE.
 #
 
-from ..errors import RangeError, RangeException
-from .instrument import Instrument
-from .resources import list_resources
-from .validators import discreteTruncate
-
-from . import activetechnologies
-from . import advantest
-from . import agilent
-from . import ametek
-from . import ami
-from . import anaheimautomation
-from . import anapico
-from . import andeenhagerling
-from . import anritsu
-from . import attocube
-from . import danfysik
-from . import deltaelektronika
-from . import fluke
-from . import fwbell
-from . import hcp
-from . import heidenhain
-from . import hp
-from . import keithley
-from . import keysight
-from . import lakeshore
-from . import newport
-from . import ni
-from . import oxfordinstruments
-from . import parker
-from . import razorbill
-from . import rohdeschwarz
-from . import siglenttechnologies
-from . import signalrecovery
-from . import srs
-from . import tektronix
-from . import temptronic
-from . import thermotron
-from . import thorlabs
-from . import toptica
-from . import yokogawa
+from .siglent_spdbase import SPDBase
+from .siglent_spd1168x import SPD1168X
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
@@ -21,6 +21,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 #
+from pymeasure.instruments.instrument import Instrument
+from pymeasure.instruments.validators import strict_discrete_set
 from pymeasure.instruments.siglenttechnologies.siglent_spdbase import SPDBase, SPDChannel
 
 
@@ -37,6 +39,19 @@ class SPD1168X(SPDBase):
 
         self.ch = {}
         self.ch[1] = SPDChannel(self, 1)
+
+    enable_4W_mode = Instrument.setting(
+        "MODE:SET %s",
+        """Configure 4-wire mode.
+
+        :type: bool
+            ``True``: enables 4-wire mode
+            ``False``: disables it.
+        """,
+        validator=strict_discrete_set,
+        values={False: "2W", True: "4W"},
+        map_values=True
+    )
 
     def shutdown(self):
         """ Ensure that the voltage is turned to zero
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
@@ -37,7 +37,6 @@ class SPD1168X(SPDBase):
             **kwargs
         )
 
-        self.ch = {}
         self.ch[1] = SPDChannel(self, 1)
 
     enable_4W_mode = Instrument.setting(
@@ -52,9 +51,3 @@ class SPD1168X(SPDBase):
         values={False: "2W", True: "4W"},
         map_values=True
     )
-
-    def shutdown(self):
-        """ Ensure that the voltage is turned to zero
-        and disable the output. """
-        self.ch[1].voltage_setpoint = 0
-        self.ch[1].enable_output(False)
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
@@ -1,0 +1,56 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2022 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+from pymeasure.instruments.siglenttechnologies.siglent_spdbase import SPDBase, SPDChannel
+
+
+class SPD1168X(SPDBase):
+    """Represent the Siglent SPD1168X Power Supply.
+    """
+
+    def __init__(self, adapter, **kwargs):
+        super().__init__(
+            adapter,
+            name="Siglent SPD1168X Power Supply",
+            **kwargs
+        )
+        self.CH1 = SPDChannel(self, 1)
+
+    def shutdown(self):
+        """ Ensures that the voltage is turned to zero
+        and disables the output. """
+        self.CH1.set_voltage(0.0)
+        self.CH1.disable()
+
+
+if __name__ == "__main__":
+    psu = SPD1168X("TCPIP0::10.20.2.245::5025::SOCKET")
+    print(psu.id)
+    print(psu.selected_channel)
+    psu.CH1.set_current = 1
+    psu.CH1.output = False
+
+    print(psu.CH1.set_voltage)
+    print(psu.CH1.set_current)
+    print(psu.CH1.voltage)
+    print(psu.error)
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
@@ -41,18 +41,5 @@ class SPD1168X(SPDBase):
     def shutdown(self):
         """ Ensure that the voltage is turned to zero
         and disable the output. """
-        self.ch[1].set_voltage(0.0)
-        self.ch[1].disable()
-
-
-if __name__ == "__main__":
-    psu = SPD1168X("TCPIP0::10.20.2.245::5025::SOCKET")
-    print(psu.id)
-    print(psu.selected_channel)
-    psu.ch[1].set_current = 1
-    psu.ch[1].disable()
-
-    print(psu.ch[1].set_voltage)
-    print(psu.ch[1].set_current)
-    print(psu.ch[1].voltage)
-    print(psu.error)
+        self.ch[1].voltage_setpoint = 0
+        self.ch[1].enable_output(False)
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
@@ -31,7 +31,7 @@ class SPD1168X(SPDBase):
     def __init__(self, adapter, **kwargs):
         super().__init__(
             adapter,
-            name="Siglent SPD1168X Power Supply",
+            name="Siglent Technologies SPD1168X Power Supply",
             **kwargs
         )
         self.CH1 = SPDChannel(self, 1)
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
@@ -21,12 +21,11 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 #
-from pymeasure.instruments.instrument import Instrument
-from pymeasure.instruments.validators import strict_discrete_set
-from pymeasure.instruments.siglenttechnologies.siglent_spdbase import SPDBase, SPDChannel
+from pymeasure.instruments.siglenttechnologies.siglent_spdbase import (SPDSingleChannelBase,
+                                                                       SPDChannel)
 
 
-class SPD1168X(SPDBase):
+class SPD1168X(SPDSingleChannelBase):
     """Represent the Siglent SPD1168X Power Supply.
     """
 
@@ -38,16 +37,3 @@ class SPD1168X(SPDBase):
         )
 
         self.ch[1] = SPDChannel(self, 1)
-
-    enable_4W_mode = Instrument.setting(
-        "MODE:SET %s",
-        """Configure 4-wire mode.
-
-        :type: bool
-            ``True``: enables 4-wire mode
-            ``False``: disables it.
-        """,
-        validator=strict_discrete_set,
-        values={False: "2W", True: "4W"},
-        map_values=True
-    )
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
@@ -24,14 +24,19 @@
 from pymeasure.instruments.siglenttechnologies.siglent_spdbase import SPDBase, SPDChannel
 
 
-class SPD1168X(SPDBase):
-    """Represent the Siglent SPD1168X Power Supply.
+class SPD1305X(SPDBase):
+    """Represent the Siglent SPD1305X Power Supply.
     """
 
+    set_voltage_values = [0, 30]
+
+    set_current_values = [0, 5]
+
     def __init__(self, adapter, **kwargs):
+
         super().__init__(
             adapter,
-            name="Siglent Technologies SPD1168X Power Supply",
+            name="Siglent Technologies SPD1305X Power Supply",
             **kwargs
         )
 
@@ -43,16 +48,3 @@ class SPD1168X(SPDBase):
         and disables the output. """
         self.ch[1].set_voltage(0.0)
         self.ch[1].disable()
-
-
-if __name__ == "__main__":
-    psu = SPD1168X("TCPIP0::10.20.2.245::5025::SOCKET")
-    print(psu.id)
-    print(psu.selected_channel)
-    psu.ch[1].set_current = 1
-    psu.ch[1].disable()
-
-    print(psu.ch[1].set_voltage)
-    print(psu.ch[1].set_current)
-    print(psu.ch[1].voltage)
-    print(psu.error)
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
@@ -28,10 +28,6 @@ class SPD1305X(SPDBase):
     """Represent the Siglent SPD1305X Power Supply.
     """
 
-    set_voltage_values = [0, 30]
-
-    set_current_values = [0, 5]
-
     def __init__(self, adapter, **kwargs):
 
         super().__init__(
@@ -42,6 +38,9 @@ class SPD1305X(SPDBase):
 
         self.ch = {}
         self.ch[1] = SPDChannel(self, 1)
+
+        self.ch[1].set_voltage_values = [0, 30]
+        self.ch[1].set_current_values = [0, 5]
 
     def shutdown(self):
         """ Ensures that the voltage is turned to zero
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
@@ -35,15 +35,17 @@ class SPD1305X(SPDBase):
             name="Siglent Technologies SPD1305X Power Supply",
             **kwargs
         )
+        voltage_limits = [0, 30]
+        current_limits = [0, 5]
 
         self.ch = {}
         self.ch[1] = SPDChannel(self, 1)
 
-        self.ch[1].set_voltage_values = [0, 30]
-        self.ch[1].set_current_values = [0, 5]
+        self.ch[1].voltage_setpoint_values = voltage_limits
+        self.ch[1].current_limit_values = current_limits
 
     def shutdown(self):
         """ Ensures that the voltage is turned to zero
         and disables the output. """
-        self.ch[1].set_voltage(0.0)
-        self.ch[1].disable()
+        self.ch[1].voltage_setpoint = 0
+        self.ch[1].enable_output(False)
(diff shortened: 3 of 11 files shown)


bmoneke:
> The only unimplemented function is currently the ability to set the timer steps of the PSU, this requires setting a list of values in the same command

Do I understand it correctly, that you have to write a list (instead of a single value) to the device?

Two solutions:
1. You can use either a validator or a set_process parameter for a pymeasure property, because the value you give it, is processed by both: `value = set_process(validator(value, values))`. The set_process could turn your list into a string appropriate for the device and your set_command could look something like `"TIMeSTeP %s"` introducing the string you previously made.
```
time_step = Instrument.setting("TIMeSTeP %s", "The time steps", set_process=lambda v: str(v))
# you can give another method (static or outside the class) as set_process as well.
inst.time_step = [1, 2, 3, 10]
```
2. You can always use normal python properties, if need arises (in this case set_process would do the job):
```
@time_step.setter
def time_step(self, values):
    values = ":".join([str(value) for value in values])
    self.write(f"TIMeSTeP {values}")
```

AidenDawn:
Thanks. I'll have a look into that and have a play around to get it working. Just for a bit more context to my question, the command structure is as follows:

bmoneke:
Another question: Why did you choose to make a "base Instrument"? Are there more instruments, which just differ in the channel number? Maybe you could read the instrument version (via id) and create that much channels (see active technolgies AWG401x).

We recently added [protocol tests](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#protocol-tests), which do not require a connected instrument. Please provide at least some tests, as they allow us to maintain the code (e.g. changing the channel implementation) without having access to the instrument proper.

AidenDawn:
> Another question: Why did you choose to make a "base Instrument"? Are there more instruments, which just differ in the channel number? Maybe you could read the instrument version (via id) and create that much channels (see active technolgies AWG401x).

The family of instruments includes some 3 channel and some single channel instruments which can all have different voltage and current output properties. I was trying to leave the option open in the future to include these other instruments. I prefer how the channels have been created in that AWG. I'll make some edits and update

> We recently added [protocol tests](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#protocol-tests), which do not require a connected instrument. Please provide at least some tests, as they allow us to maintain the code (e.g. changing the channel implementation) without having access to the instrument proper.

Will do, looks like Ive got some more work to do ;)

pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
```diff
@@ -0,0 +1,58 @@
...
+            name="Siglent Technologies SPD1168X Power Supply",
+            **kwargs
+        )
+
+        self.ch = {}
+        self.ch[1] = SPDChannel(self, 1)
+
+    def shutdown(self):
+        """ Ensures that the voltage is turned to zero
+        and disables the output. """
```
bmoneke, line 43: Docstrings should be an imperative, what the method should do:
```suggestion
        """ Ensure that the voltage is turned to zero
        and disable the output. """
```

pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
```diff
@@ -0,0 +1,58 @@
...
+    psu = SPD1168X("TCPIP0::10.20.2.245::5025::SOCKET")
+    print(psu.id)
+    print(psu.selected_channel)
+    psu.ch[1].set_current = 1
+    psu.ch[1].disable()
+
+    print(psu.ch[1].set_voltage)
+    print(psu.ch[1].set_current)
+    print(psu.ch[1].voltage)
+    print(psu.error)
```
bmoneke, line 58: Did you intend to write a test?

For this, we have a test file folder (see documentation). Tests with devices have a different file from tests without a device.

Some of the present tests with devices show, how to do it. They have, e.g. a parameter area, where you can specify the device address.

AidenDawn: I'm just in the process of writing the test now that I've finished the implementation. Should come through as a final commit if there are no issues

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def __init__(self, adapter, **kwargs):
+        super().__init__(
+            adapter,
+            usb=dict(write_termination='\n',
+                     read_termination='\n'),
+            tcpip=dict(write_termination='\n',
+                       read_termination='\n'),
+            **kwargs
+        )
```
bmoneke, line 151: Please move the __init__s to the top of the classes, to find them more easily.

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,291 @@
...
+    def recall_config(self, index):
+        """Recall a config from memory.
+
+        :param index:
+            string: index of the location from which to recall the configuration
+        :returns: self
+        """
+        self.recall_settings = index
+
+        return self
```
bmoneke, line 173: Why do you return the instrument instance?

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,291 @@
...
+class SPDChannel(object):
+    channel = 1
+
+    voltage = Instrument.measurement(
+        f"MEAS:VOLT? CH{channel}",
+        """Measure the channel output voltage.
+
+        :type: float
+        """
+    )
```
bmoneke, line 185: Does this work as expected?

The property creators work at class level, therefore, all channels will be "CH1".

What you can do, is to define a (**not** f-string) `"MEAS:VOLT? CH{channel}"` and in `SPDChannel.write` you do `self.instrument.write(command.format(channel=self.channel))`. This way, you replace at instance level the channel name.

AidenDawn: Thanks for this one. I was slightly worried that was what I had done here. Good to have some confirmation. Changes are in the works

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
AidenDawn: You're right. These returns aren't necessary. They were left over from a copy of another instrument and I missed them

AidenDawn:
This should now be ready pending a final review of the edits made

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,284 @@
...
+
+    system_status_code = Instrument.measurement(
+        "SYST:STAT?",
+        """Read the system status register.
+
+        :type: :class:`.SystemStatusCode`
+        """,
+        validator=truncated_range,
+        values=[0, 1024],
```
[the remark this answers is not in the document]
AidenDawn: Thanks, I can see why that makes sense. I've definitely seen this style included in other instruments before so I inlcuded it here, but Im removing it now

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,278 @@
...
+    )
+
+    output = Instrument.setting(
+        "OUTP CH{channel},%s",
+        """Configure the power supply output.
+
+        :type: bool
+            ``True``: disable local interface
+            ``False``: enables local interface
+
```
bmoneke, line 229: Does the docstring match the values?

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,278 @@
...
+        return self.instrument.ask(cmd.format(channel=self.channel))
+
+    def write(self, cmd):
+        self.instrument.write(cmd.format(channel=self.channel))
+
+    def values(self, cmd, **kwargs):
+        """ Reads a set of values from the instrument through the adapter,
+        passing on any key-word arguments.
+        """
+        return self.instrument.values(cmd.format(channel=self.channel))
```
bmoneke, line 257: You say, you pass on kwargs, but you don't do it.
```suggestion
        """ Reads a set of values from the instrument through the adapter,
        passing on any key-word arguments.
        """
        return self.instrument.values(cmd.format(channel=self.channel), **kwargs)
```

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,278 @@
...
+            tcpip=dict(write_termination='\n',
+                       read_termination='\n'),
+            **kwargs
+        )
+
+    error = Instrument.measurement(
+        "SYST:ERR?",
+        """Read the error code and information of the instrument.
+
+        :type: str
```
bmoneke, line 79: I've rarely seen this type annotation. For a getter (measurement), most of the time, the type does not matter, as you get the variable in the correct type. It's only important for the setter and there not often. Most of the time it is obvious (boolean, string, float, or int).
Additionally, you can write it in the text, if the type might be ambibous, e.g.:
```suggestion
        """Read the error code and information of the instrument as a string.
```

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,278 @@
...
+
+        :type: int
+        """,
+    )
+
+    selected_channel = Instrument.control(
+        "INST?", "INST %s",
+        """Control the selected channel of the instrument.
+
+        :type: :class:`.SPDChannel`
```
bmoneke, line 129: Is the type really a class or just an index? According to "values", I'd say its an integer.

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,278 @@
...
+
+        """,
+        validator=strict_discrete_set,
+        values={1: "CH1"},
+        map_values=True,
+        dynamic=True
```
bmoneke, line 135: I'd add a comment, that instrument implementors have to change the values according to the number of channels. At a first glance, it looks stupid to have a single valid value.

bmoneke (commented):
A few more comments from my side. Some are only suggestions on the docstrings, and some are minor inconsistencies. Thanks for the patience with my comments.

Now the maintainers (@bilderbuchi @msmttchr ) have to look over the code. They will make some more suggestions and requests.

Thank you again for your contribution.

AidenDawn:
> A few more comments from my side. Some are only suggestions on the docstrings, and some are minor inconsistencies. Thanks for the patience with my comments.
> 
> Now the maintainers (@bilderbuchi @msmttchr ) have to look over the code. They will make some more suggestions and requests.
> 
> Thank you again for your contribution.

Thanks for the comments, it's a good chance to see where I can improve what Im sending out/missing throughout this process

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
AidenDawn: In the case of this instrument, the error code (returned as an integer in a string format) is followed by a plain-text description of the error which has occurred. In the default case, the string "0 No Error" is the result which I agree is an unusual response. You're right that your suggestion is probably more readable as a sentence but this version of formatting should show up very clearly in the docs. The docs page for temptronic_base was my example for this

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
AidenDawn: Good spot

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,294 @@
...
+        return self.instrument.ask(cmd.format(channel=self.channel))
+
+    def write(self, cmd):
+        self.instrument.write(cmd.format(channel=self.channel))
+
+    def values(self, cmd, **kwargs):
+        """ Reads a set of values from the instrument through the adapter,
+        passing on any key-word arguments.
+        """
+        return self.instrument.values(cmd.format(channel=self.channel), **kwargs)
```
bilderbuchi, line 273: If you can wait for some time until #718 lands, the whole channel handling will become much easier, with less need to define these methods.

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,294 @@
...
+        return self
+
+    def disable(self):
+        """Disable the channel output
+
+        :returns: self
+        """
+        self.instrument.selected_channel = self.channel
+        self.output = False
+        return self
```
bilderbuchi, line 294: Why does this duplicate what `output` does? 
Is that the only function that needs that channel selection machinery? If yes, would using the channels `.output` alone even work correctly?
Why are you returning `self` here?

pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
```diff
@@ -0,0 +1,58 @@
...
```
bilderbuchi, line 58: This looks like it's leftover from testing?

tests/instruments/siglenttechnologies/test_siglent_spd1168x.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_set_current_trunc():
+    with expected_protocol(
+        SPD1168X,
+        [("CH1:CURR 8", None),
+         ("CH1:CURR?", "0.5")]
+    ) as inst:
+        inst.ch[1].set_current = 10  # too large, gets truncated
+        assert inst.ch[1].set_current == 0.5
```
bilderbuchi, line 53: You give it 10, it truncates to 8, but then returns 0.5? That seems confusing? :smile:

tests/instruments/test_all_instruments.py
```diff
@@ -54,7 +54,7 @@
 init = ["ThorlabsPM100USB", "Keithley2700", "TC038", "Agilent34450A",
         "AWG401x_AWG", "AWG401x_AFG", "VARX", "HP8116A"]
 # Instruments which require more input arguments
-init.extend(["Instrument", "ATSBase"])
+init.extend(["Instrument", "ATSBase", "SPDBase"])
```
bilderbuchi, line 57: What happens if you don't do that? From what I can see, `SPDBase` does _not_ need communication in its init?

bilderbuchi (changes requested):
Thanks @bmoneke for the preparatory work!
Some remarks from my side. If you can wait a little for #718 to land, this contribution can be streamlined a little more.

bilderbuchi:
> Thanks. I'll have a look into that and have a play around to get it working. Just for a bit more context to my question, the command structure is as follows:
> 
> TIMEr Command format TIMEr: SET <SOURce>, <secnum>, <\volt>, <curr>, <time> <SOURce>:={CH1}; <secnum>;=1 to 5; Description Set timing parameters of specified channel Example TIMEr: SET CH1, 2, 3, 0.5, 2

For this, as it's a pretty complicated thing, I recommend implementing a method `configure_timer`. You can add arguments with clear names for all the needed values, and it sounds like an action to begin with. Not everything is a great fit for properties :-)

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,294 @@
...
+        "%s",
+        """Set the local interface lockout.
+
+        :type: bool
+            ``True``: disables the local interface
+            ``False``: enables it.
+
+        """,
+        validator=strict_discrete_set,
+        values={True: "*LOCK", False: "*UNLOCK"},
```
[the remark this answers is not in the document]
AidenDawn: Youre right, the name needs changing. I think disable_local_interface is probably even better

pymeasure/instruments/siglenttechnologies/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from .siglent_spdbase import SPDBase
```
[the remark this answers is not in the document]
AidenDawn: I've been using temptronic instruments as the template here. I think if you were implementing an instrument in the family that hadn't been specifically included already then there is scope for the exposure of the base class to be useful. But your point is also completely valid. Happy to go either way

tests/instruments/siglenttechnologies/test_siglent_spd1168x.py
AidenDawn: Yes, that's odd

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+    )
+
+    set_current = Instrument.control(
+        "CH{channel}:CURR?", "CH{channel}:CURR %g",
+        """Control the output current configuration of the channel.
+
+        :type: float
+
+        """,
```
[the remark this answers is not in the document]
AidenDawn: For some reason, when I try and and build the docs without this space I end up with:
`C:\Users\dschmeer\Projects\pymeasure\pymeasure\instruments\siglenttechnologies\siglent_spdbase.py:docstring of pymeasure.instruments.siglenttechnologies.siglent_spdbase.SPDChannel:1: WARNING: Field list ends without a blank line; unexpected unindent.`
It did this in a few places in SPDBase, is there a known reason for this? If not Ill just leave the newline in

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,294 @@
...
+            ``True``: enables the output
+            ``False``: disables the output
+
+        """,
+        validator=strict_discrete_set,
+        values={True: "ON", False: "OFF"},
+        map_values=True
+    )
+
+    enable_timer = Instrument.setting(
```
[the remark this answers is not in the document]
AidenDawn: We'll go with enable_timer, particularly given I'm writing set_timer as a method too

tests/instruments/test_all_instruments.py
AidenDawn: This is for the same reason as `ATSBase`. Neither of these two base classes provide an instrument name, which consistently causes the __init__ of `Instrument` to fail. This is not needed for the child classes because the name is defined in there. The other option I tried was to set a base name (something like Siglent SPDxxxxX instrument Base Class) but that caused further errors in the child classes with multiple `name` definitions

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
AidenDawn: I've changed the output part to remove property and replace with an `enable_output` method. hopefully this will make more sense overall

tests/instruments/test_all_instruments.py
bmoneke: @bilderbuchi the comment above the list states the reason for the exception: these instruments require input arguments (in this case `name`).

I know, that the skip message is not appropriate, but I did not add a special check just for `Instrument` class.

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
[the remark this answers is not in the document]
AidenDawn: I've done a bit more testing, it seems like there is a sphinx whitespace bug going on here. I have two options which work, one is the one included above. The other option is to replace the line in the docstring as:
`  :type : float
   """,`. Personally my preference will be to keep the newline because it looks tidier

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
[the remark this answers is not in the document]
bilderbuchi: `disable_local_interface` would [suggest a method](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#your-instrument-s-user-interface), not a property, though, as `disable` is a verb.

pymeasure/instruments/siglenttechnologies/__init__.py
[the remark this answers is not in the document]
bilderbuchi: Hm, good point. If you're implementing a new instrument, I think you should preferably directly import from the`siglent_spdbase` subpackage, and the main package namespace would be for completed instruments. 
We don't have a formulated policy for this, yet.  @pymeasure/pymeasure-maintainers thoughts?

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -0,0 +1,294 @@
...
+
+        :type: int
+        """,
+        validator=strict_range,
+        values=[1, 5]
+    )
+
+    selected_channel = Instrument.control(
+        "INST?", "INST %s",
+        """Control the selected channel of the instrument.
```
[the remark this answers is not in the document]
bilderbuchi: Don't "resolve" conversation where you have a question. :-)

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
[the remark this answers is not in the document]
bilderbuchi: Hm, could be a Sphinx/ReST quirk or bug (couldn't find a report in the sphinx repo with a quick search), resolve it as you want. Great that you checked doc generation locally!

tests/instruments/test_all_instruments.py
bilderbuchi: My only excuse is that it must have been too late to notice that comment :-P

What if you add `kwargs.setdefault('name', 'Siglent SPDxxxxX instrument Base Class')` in `SPDBase`'s init, just before you call `super()__init__`? This should enable using it in tests, but still result in the correct names for the child classes?

pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
```diff
@@ -38,21 +40,21 @@ def __init__(self, adapter, **kwargs):
...
-    print(psu.selected_channel)
-    psu.ch[1].set_current = 1
-    psu.ch[1].disable()
-
-    print(psu.ch[1].set_voltage)
-    print(psu.ch[1].set_current)
-    print(psu.ch[1].voltage)
-    print(psu.error)
+        self.ch[1].voltage_setpoint = 0
+        self.ch[1].enable_output(False)
```
bilderbuchi, line 60: also add a call to `super().shutdown()` at the end, so you get the rest of the cleanup, too.

bilderbuchi: Also, since you seem to duplicate this in the other child class, maybe that should move up to `SPCBase`?

pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
```diff
@@ -35,15 +37,30 @@ def __init__(self, adapter, **kwargs):
...
+        map_values=True
+    )
 
     def shutdown(self):
         """ Ensures that the voltage is turned to zero
         and disables the output. """
-        self.ch[1].set_voltage(0.0)
-        self.ch[1].disable()
+        self.ch[1].voltage_setpoint = 0
+        self.ch[1].enable_output(False)
```
bilderbuchi, line 66: as above.

pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
```diff
@@ -38,21 +40,21 @@ def __init__(self, adapter, **kwargs):
         self.ch = {}
         self.ch[1] = SPDChannel(self, 1)
 
+    enable_4W_mode = Instrument.setting(
```
bilderbuchi, line 43: I'm confused, why is this property now duplicated in the two child classes?

bilderbuchi: Also, from the name (having a verb) this should be a method `enable_4W_mode(enabled):`

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -98,92 +98,64 @@ def __init__(self, adapter, **kwargs):
         get_process=lambda v: SystemStatusCode(int(v, base=16)),
     )
 
-    local_lockout = Instrument.setting(
+    disable_local_interface = Instrument.setting(
```
bilderbuchi, line 101: A verb in the name implies an "action", which is most properly a method, not a property, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#your-instrument-s-user-interface.

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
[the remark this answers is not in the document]
bilderbuchi: I'm not sure I fully understand the feature distribution here, but maybe have an really slim class `SPDSingleChannel(SPDBase)` if you want to isolate the single-channel-model-only features?

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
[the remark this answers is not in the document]
bilderbuchi: or did that work out in the end?

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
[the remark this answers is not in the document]
bilderbuchi: You only need this for `enable_output`, then, and only in setting? Might be easier to just add a second write call to that method?

bmoneke:
I agree, that "base instruments" (for some manufacturers etc.) should not be in the main name space. As a user, you won't need them, but they will pop up with autocompletion and might make it more difficult to find the actual instrument. 

Only implementing an instrument, you need them, and then you can use the slightly longer name. 

Am 13. November 2022 11:58:25 MEZ schrieb Christoph Buchner ***@***.***>:
***@***.*** commented on this pull request.

pymeasure/instruments/siglenttechnologies/__init__.py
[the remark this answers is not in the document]
msmttchr: I agree. My preference would be to not include it in the main package namespace.

tests/instruments/test_all_instruments.py
AidenDawn: This works

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
bilderbuchi: I agree, this seems ready pending the channel changes. 
I would say wait until #718 lands, merge a fresh master into your branch, then rework. 
Alternatively, @bmoneke do you want this in master now, pull into #718 and rework this instrument within #718?

bmoneke: There are many instruments with channels, which I do not rework in #718.
We can rework them step by step, as they will continue to work as they are (just the code base is not as nice).

Let's merge this work as is, in order to close the PR

bilderbuchi: Sounds good! I opened #746 for tracking this.

PR #722, opened 2022-10-26
Docking windows

mcdo0486:
This pull request adds a `DockWindow` window class. It utlizes `QtWidgets.QDockWidget` and `DockArea` from pyqtgraph to display multiple plots to be displayed when running a procedure. A list of labels can be passed to x_axis or y_axis to set the default label for each plot. 

Here is an example of DockWindow: https://gist.github.com/mcdo0486/eb2304a6de87c813028e9218ef283b0b

PR #723, opened 2022-10-26
Multiple plots

mcdo0486:
This pull request adds a MultiPlotWindow window class, MultiPlotWidget and MultiPlotResultsDialog widget classes. It allows multiple plots to be displayed in one window that share a x_axis. A list of labels can be passed y_axis to set the y_axis label for each plot.

Here is an example of MultiPlotWindow: https://gist.github.com/mcdo0486/1627785d5c355b1a1b8833334b4d4c16


pymeasure/display/windows/multiplot_window.py
```diff
@@ -0,0 +1,132 @@
...
+
+        if "widget_list" not in kwargs:
+            kwargs["widget_list"] = ()
+        kwargs["widget_list"] = kwargs["widget_list"] + (self.plot_widget, self.log_widget)
+
+        super().__init__(procedure_class, **kwargs)
+
+        # Setup measured_quantities once we know x_axis and y_axis
+        measure_quantities = [self.x_axis, self.y_axis]
+        if type(self.y_axis) == list:
```
CasperSchippers, line 84: for type checking you can better use `isinstance(self.y_axis, list)`

pymeasure/display/windows/multiplot_window.py
```diff
@@ -0,0 +1,132 @@
...
+            # Expand y_axis if it is a list
+            measure_quantities = [self.x_axis, *self.y_axis]
+        self.browser_widget.browser.measured_quantities = measure_quantities
+
+        logging.getLogger().addHandler(self.log_widget.handler)  # needs to be in Qt context?
+        log.setLevel(self.log_level)
+        log.info("MultiPlotWindow connected to logging")
+
+    def open_experiment(self):
```
CasperSchippers, line 93: I'm not a very big fan of having to amend these basic methods (i.e. `open_experiment` and `new_experiment`) of the `ManagedWindowBase` class, since this prohibits users to make there own ManagedWindow variant by simply add different TabWidgets. (additionally, at this moment some things wont work if you add another 'normal' PlotWidget.)

I think that everything should be handled by the widgets (or maybe the curves) itself, or maybe (but only when necessary of course, so I'm typically a bit reluctant on this regard) by generalising the definitions in the ManagedWindowBase class.

pymeasure/display/widgets/multiplot_widget.py
```diff
@@ -0,0 +1,210 @@
...
+    """ Extends the PlotFrame to allow different columns
+    of the data to be dynamically choosen across multple plots
+    """
+
+    def __init__(self, name, columns, x_axis=None, y_axis=None, num_plots=1, *args, **kwargs):
+
+        super().__init__(name, columns, x_axis, y_axis, setup=False, *args)
+        self.num_plots = num_plots
+        self._setup_ui()
+        self._layout()
```
CasperSchippers, line 52: Here, the `setup=False` logic is also not required.

```suggestion
        self.num_plots = num_plots
        super().__init__(name, columns, x_axis=None, y_axis=None, *args)
```

pymeasure/display/widgets/multiplot_widget.py
```diff
@@ -0,0 +1,210 @@
...
+from .results_dialog import ResultsDialog
+from ...experiment.results import Results
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class MultiPlotWidget(PlotWidget):
+    """ Extends the PlotFrame to allow different columns
+    of the data to be dynamically choosen across multple plots
```
CasperSchippers, line 44: Typo: "chosen across multiple"

pymeasure/display/widgets/plot_widget.py
```diff
@@ -41,20 +41,21 @@ class PlotWidget(TabWidget, QtWidgets.QWidget):
     """
 
     def __init__(self, name, columns, x_axis=None, y_axis=None, refresh_time=0.2,
-                 check_status=True, linewidth=1, parent=None):
```
CasperSchippers, line 44: Refer to my comment regarding `setup=False` in multiplot_widget.py: I think the changes in this file are not necessary.

pymeasure/display/manager.py
```diff
@@ -260,7 +260,13 @@ def _finish(self):
...
         for curve in experiment.curve_list:
             if curve:
-                curve.update_data()
+                # For single plots, update_data() on ResultsCurve
+                if type(curve) == ResultsCurve:
+                    curve.update_data()
+                # For multiple plots, update_data() on list of ResultsCurve
+                elif type(curve) == list:
+                    for c in curve:
+                        c.update_data()
```
CasperSchippers, line 269: To me, it would seem nicer if the curve_list would consist of only actual ResultsCurve objects, not a mix of ResultsCurves and Lists of ResultsCurves, either by expanding the list before it is committed (in the `new_experiment` method of the ManagedWindow(Base) class), or alternatively by introducing a new type of Curve, the MultiResultsCurve.

pymeasure/display/widgets/multiplot_widget.py
```diff
@@ -0,0 +1,210 @@
...
+
+        vbox.addLayout(hbox)
+        vbox.setSpacing(10)
+        for i in range(self.num_plots):
+            vbox.addWidget(self.plot_frame[i])
+        self.setLayout(vbox)
+
+    def new_curve(self, results, color=pg.intColor(0), **kwargs):
+        if 'pen' not in kwargs:
+            kwargs['pen'] = pg.mkPen(color=color, width=2)
```
CasperSchippers, line 127: In line with the `PlotWidget` I would suggest to use `width=self.linewidth`.

CasperSchippers (commented):
Thanks a lot for this PR, really appreciated!!! I see a great potential for this idea, but I do have some thoughts and ideas.
In its current for, I do not per se see a lot benefits over your other PR (the docked plots #722), since the graphs are still practically three separated graphs but now with just a singular drop-down box for the x-axis. Feel free though to convince me otherwise (and I of course also don't speak for the other maintainers).

If, however, it would be possible to really merge the graphs (e.g. reduce/remove the space between the graph, only ticks and labels on the lower x-axis to save space, synchronised x-axes such that zooming/shifting in one axis works for all axes) I would see a tremendous advantage! (The holy grail in this sense I think would be to have the option to add or remove graphs on the go, but this might be (I don't know) extremely difficult).
This might be achievable using the  pyqtgraph.GraphicsLayoutWidget or the pyqtgraph.MultiPlotItem.

I do realize that this might be a bit (maybe a lot) more difficult than the current implementation, so I'm not sure if this is the right way to go, but just wanted to write down my thoughts on this.

Regardless of the outcome of the above suggestion, I added some first remarks (some of which might not be relevant if the above suggestion is followed).

mcdo0486:
> In its current for, I do not per se see a lot benefits over your other PR (the docked plots #722), since the graphs are still practically three separated graphs but now with just a singular drop-down box for the x-axis. Feel free though to convince me otherwise (and I of course also don't speak for the other maintainers).

Hey, thanks for all the useful comments, I will start working through them.

I created this pull request as a follow up to Casvharen's original request which I helped work on in pull request https://github.com/pymeasure/pymeasure/pull/461 

The people that use my fork prefer `DockWindow` instead of `MultiPlotWindow` but I can see a use case for some types of experiments.

docs/tutorial/graphical.rst
```diff
@@ -457,4 +457,4 @@ Multiple parameters for grouping can either be passed as a dict of string: condi
 Note that in this example, :code:`param_A` and :code:`param_B` are identically grouped: they're only visible if :code:`iterations` is between 51 and 99 and if the `toggle` checkbox is checked (i.e. True).
 
 .. _pyqtgraph: http://www.pyqtgraph.org/
-.. _PlotItem: http://www.pyqtgraph.org/documentation/graphicsItems/plotitem.html
+.. _PlotItem: http://www.pyqtgraph.org/documentation/graphicsItems/plotitem.html
```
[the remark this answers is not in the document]
mcdo0486: Reverted in last commit

pymeasure/display/widgets/multiplot_widget.py
```diff
@@ -0,0 +1,210 @@
...
+    def clear(self):
+        for i in range(self.num_plots):
+            self.plot_frame[i].plot.clear()
+
+    def set_color(self, curve, color):
+        for i in range(self.num_plots):
+            curve[i].setPen(pg.mkPen(color=color, width=2))
+
+
+class MultiPlotResultsDialog(ResultsDialog):
```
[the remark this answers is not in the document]
mcdo0486: Removed in last commit

pymeasure/display/widgets/results_dialog.py
```diff
@@ -45,12 +45,16 @@ class ResultsDialog(QtWidgets.QFileDialog):
     :class:`ManagedWindowBase<pymeasure.display.windows.managed_window.ManagedWindowBase>` class
     """
 
-    def __init__(self, columns, x_axis=None, y_axis=None, parent=None):
+    def __init__(self, columns, x_axis=None, y_axis=None, parent=None, setup=True):
```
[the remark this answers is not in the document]
mcdo0486: Reverted in last commit

pymeasure/display/widgets/multiplot_widget.py
mcdo0486: Fixed this one

pymeasure/display/widgets/plot_widget.py
mcdo0486: Reverted this change

pymeasure/display/widgets/multiplot_widget.py
mcdo0486: Reverted this change

pymeasure/display/widgets/multiplot_widget.py
mcdo0486: Fixed this issue

pymeasure/display/windows/multiplot_window.py
mcdo0486: Replaced with isinstance()

msmttchr:
> Thanks a lot for this PR, really appreciated!!! I see a great potential for this idea, but I do have some thoughts and ideas. In its current for, I do not per se see a lot benefits over your other PR (the docked plots #722), since the graphs are still practically three separated graphs but now with just a singular drop-down box for the x-axis. Feel free though to convince me otherwise (and I of course also don't speak for the other maintainers).
> 
> If, however, it would be possible to really merge the graphs (e.g. reduce/remove the space between the graph, only ticks and labels on the lower x-axis to save space, synchronised x-axes such that zooming/shifting in one axis works for all axes) I would see a tremendous advantage! (The holy grail in this sense I think would be to have the option to add or remove graphs on the go, but this might be (I don't know) extremely difficult). This might be achievable using the pyqtgraph.GraphicsLayoutWidget or the pyqtgraph.MultiPlotItem.
> 
> I do realize that this might be a bit (maybe a lot) more difficult than the current implementation, so I'm not sure if this is the right way to go, but just wanted to write down my thoughts on this.
> 
> Regardless of the outcome of the above suggestion, I added some first remarks (some of which might not be relevant if the above suggestion is followed).

I agree with @CasperSchippers, if we are able to make a real multiplot, I see great advantages otherwise, I think the other PR #722  is covering this need.
If you checkout https://github.com/pyqtgraph/pyqtgraph/pull/1359, `pyqtgraph` people are trying to add this feature and hopefully this will converge soon.
My opinion is either:
- we close this PR in favor of #722 
- we wait for https://github.com/pyqtgraph/pyqtgraph/pull/1359 to be merged and then base this PR on it.
- we base implementation on https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/examples/MultiplePlotAxes.py

mcdo0486:
I'm fine closing this PR in favor of Docking windows. 

From talking with a couple users today, the use of multiple curves on a single graph doesn't have as much value as separate graphs. The measurements are often independent and aren't correlated so sharing curves on one graph isn't as useful as simplicity of multiple graphs. This is just random feedback I got today. If there is interest from the users then I'll definitely checkout the multiaxis plot implementation for a widget, thanks for linking to the pyqtgraph thread.

CasperSchippers:
Thank you for your thoughts and for your willingness to close this PR. I will close this in favour of #722 then; to keep track of the ideas/thoughts, I've opened issue #734.

Issue #513, opened 2021-11-05
Should Instrument.shutdown() do more?


bilderbuchi:
Right now, we have

I want to open a discussion if there is an expectation that `Instrument.shutdown()` do more. Right now it's ~ a no-op (setting a boolean and emitting a log message), and it's fully up to the instrument developer to make it do more.

Should it at least close the connection to the hardware? 
*If so: Extract an `Adapter.close()` from `Adapter.__del__()` and call that. TBC Do all Adapters observe that?*

Should it log/warn if not `self.complete`?

Any other things?

LongnoseRob:
From my point of view,  `Instrument.shutdown()` should be the place where the instrument is brought to a safe condition before the connection is closed.
With safe condition I mean conditions like this:
- output off for RF-generators and power suppliies
- motors stopped for motor drivers.

But it could also mean that certain settings in the instrument could be recorded before the connection is terminated.

Ideally `Instrument.shutdown()` would be called for all open instruments in a procedure/experiment, when the procedure/experiment is completed.

Also I would like to see it called if there is an interruption by the user (Ctrl+C) or failure exception stopping the execution of the python script, I think this could be doable with defining sth like `Procedure.__exit__`

I know that SCPI (and lower protoclls like GPIB or USB-TMC) has less focus on safety then e.g. industrial field busses like Profibus, DeviceNet or EtherCat. 
Nevertheless we should try to get everything into a safe-state in such cases.

msmttchr:
I would vote for closing the connection and leaving instrument's specific action to subclasses

bilderbuchi:
All good points! 
Yes, device-specific measured would be implemented in the concrete subclass, and we make sure that that is automatically called by the `shutdown()` method. 
Exit/interruption-safe handling could be happen in `__del__` and/or a finally clause in code handling this, as appropriate.

PR #732, opened 2022-11-04
New instrument: HP8657B RF signal generator

LongnoseRob:
Simple RF signal generator implementation, legacy device with listenm-only GPIB/HPIB interface

### DIFF
--- a/docs/api/instruments/hp/hp8657B.rst
+++ b/docs/api/instruments/hp/hp8657B.rst
@@ -1,0 +1,15 @@
+################################
+HP Signal generator HP8657B
+################################
+
+
+*Note:*
+    * This instrument does not support reading back from the instrument, as it is a listen-only 
+      GPIB device
+    
+    * Other instrument of this family could be implemented using the dymaic ranges feature
+    
+    
+.. autoclass:: pymeasure.instruments.hp.HP8657B
+    :members:
+    :show-inheritance:
--- a/docs/api/instruments/hp/hp8657B.rst
+++ b/docs/api/instruments/hp/hp8657B.rst
@@ -4,11 +4,11 @@ HP Signal generator HP8657B
 
 
 *Note:*
-    * This instrument does not support reading back values, as it is a listen-only GPIB device
+    * This instrument does not support reading back values, as it is a listen-only GPIB device.
     
-    * Other instruments of this family could be implemented using the dymaic ranges feature
+    * Other instruments of this family could be implemented using the dynamic ranges feature.
     
-    * Optorional pulse modulation feature is not support yet
+    * Optional pulse modulation feature is not supported yet.
     
 Glossary:
     ============  ===========
--- a/docs/api/instruments/hp/hp8657B.rst
+++ b/docs/api/instruments/hp/hp8657B.rst
@@ -4,12 +4,11 @@ HP Signal generator HP8657B
 
 
 *Note:*
-    * This instrument does not support reading back from the instrument, as it is a listen-only 
-      GPIB device
+    * This instrument does not support reading back from the instrument, as it is listen-only GPIB device
     
     * Other instrument of this family could be implemented using the dymaic ranges feature
     
-    
+
 .. autoclass:: pymeasure.instruments.hp.HP8657B
     :members:
     :show-inheritance:
--- a/docs/api/instruments/hp/hp8657B.rst
+++ b/docs/api/instruments/hp/hp8657B.rst
@@ -4,9 +4,20 @@ HP Signal generator HP8657B
 
 
 *Note:*
-    * This instrument does not support reading back from the instrument, as it is listen-only GPIB device
+    * This instrument does not support reading back values, as it is a listen-only GPIB device
     
-    * Other instrument of this family could be implemented using the dymaic ranges feature
+    * Other instruments of this family could be implemented using the dymaic ranges feature
+    
+    * Optorional pulse modulation feature is not support yet
+    
+Glossary:
+    ============  ===========
+    Abbreviation  Explanation
+    ============  ===========
+    AM            Amplitude Modulation
+    FM            Frequency Modulation
+    dBm           power level in dB referenced to 1mW
+    ============  ===========
     
 
 .. autoclass:: pymeasure.instruments.hp.HP8657B
--- a/pymeasure/instruments/hp/hp8657b.py
+++ b/pymeasure/instruments/hp/hp8657b.py
@@ -1,7 +1,7 @@
 #
 # This file is part of the PyMeasure package.
 #
-# Copyright (c) 2013-2021 PyMeasure Developers
+# Copyright (c) 2013-2022 PyMeasure Developers
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
--- a/pymeasure/instruments/hp/hp8657b.py
+++ b/pymeasure/instruments/hp/hp8657b.py
@@ -152,7 +152,7 @@ class HP8657B(Instrument):
         "FR %9.1f HZ",
         """
         Set the output frequency of the instrument in Hz.
-        
+
         For the 8567B the valid range is 100 kHz to 2060 MHz.
         """,
         validator=strict_range,
--- a/pymeasure/instruments/hp/hp8657b.py
+++ b/pymeasure/instruments/hp/hp8657b.py
@@ -164,7 +164,7 @@ class HP8657B(Instrument):
         """
         Set the output level in dBm.
 
-        For the 8657B the range is -143.5 to +17 dBm
+        For the 8657B the range is -143.5 to +17 dBm/
 
         """,
         validator=strict_range,
@@ -174,7 +174,7 @@ class HP8657B(Instrument):
     level_offset = Instrument.setting(
         "AO %g DB",
         """
-        Sets the output offset in dB, usable range -199 to +199 dB
+        Set the output offset in dB, usable range -199 to +199 dB.
 
         """,
         validator=strict_range,
@@ -184,7 +184,7 @@ class HP8657B(Instrument):
     output_enabled = Instrument.setting(
         "R%d",
         """
-        Controls the output
+        Control whether the output is enabled.
         """,
         validator=strict_discrete_set,
         values={False: 2, True: 3},
--- a/pymeasure/instruments/hp/hp8657b.py
+++ b/pymeasure/instruments/hp/hp8657b.py
@@ -37,12 +37,11 @@ class HP8657B(Instrument):
     """
 
     def __init__(self, adapter, **kwargs):
-        kwargs.setdefault('read_termination', '\r\n')
-        kwargs.setdefault('send_end', True)
         super().__init__(
             adapter,
             "Hewlett-Packard HP8657B",
             includeSCPI=False,
+            send_emd=True,
             **kwargs,
         )
 
@@ -54,7 +53,6 @@ class HP8657B(Instrument):
 
         """
         log.warning("HP8657B Does not support error status readout")
-        return 0
 
     def clear(self):
         """
--- a/pymeasure/instruments/hp/hp8657b.py
+++ b/pymeasure/instruments/hp/hp8657b.py
@@ -31,16 +31,15 @@ from pymeasure.instruments.validators import strict_discrete_set, strict_range
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
-limits = {
-    "HP6632A": {"Volt_lim": 20.475, "OVP_lim": 22.0, "Cur_lim": 5.118},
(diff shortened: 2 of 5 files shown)


pymeasure/instruments/hp/hp8657b.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def __init__(self, adapter, **kwargs):
+        kwargs.setdefault('read_termination', '\r\n')
+        kwargs.setdefault('send_end', True)
+        super().__init__(
+            adapter,
+            "Hewlett-Packard HP8657B",
+            includeSCPI=False,
+            **kwargs,
+        )
```
bmoneke, line 47: If the device does not support a different termination/send_end, I'd put these kwargs direct in the super.init:
```suggestion
    def __init__(self, adapter, **kwargs):
        super().__init__(
            adapter,
            "Hewlett-Packard HP8657B",
            includeSCPI=False,
            read_termination='\r\n',
            send_end=True,
            **kwargs,
        )
```

pymeasure/instruments/hp/hp8657b.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def check_errors(self):
+        """
+        Method to read the error status register
+
+        as the 8657B does not support any readout of values, this will return 0 and log a warning
+
+        """
+        log.warning("HP8657B Does not support error status readout")
+        return 0
```
bmoneke, line 57: Why do you return 0? The user will think, there is no error. Wouldn't it be better to raise an Exception?

pymeasure/instruments/hp/hp8657b.py
```diff
@@ -0,0 +1,190 @@
...
+        self.adapter.connection.clear()
+
+    id = "HP 8657B Signal generator"  #: Manual ID entry, as the instrument does not talk with us
+
+    # List of modulation sources
+    modulation_source = {"External": 1,
+                         "Int_400Hz": 2,
+                         "Int_1000Hz": 3,
+                         "OFF": 4,
+                         "DC_FM": 5}
```
bmoneke, line 73: If you use an enum, the user will have autocompletion in their editor, e.g. `inst.am_source = Modulation.OFF`

bmoneke:
Thanks for your pull request.

Please provide at least a few [protocol tests](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#protocol-tests) for this device.

pymeasure/instruments/hp/hp8657b.py
```diff
@@ -0,0 +1,188 @@
...
+
+    am_depth = Instrument.setting(
+        "AM %2.1f PC",
+        """
+        sets the modulation depth for AM,
+        usable range 0-99.9%
```
bmoneke, line 77: The property docstrings should follow the normal docstring style (according to the not yet merged documentation overhaul). Therefore the first sentence is an imperative (and starts with a capital letter):
```suggestion
        Set the modulation depth for AM,
        usable range 0-99.9%
```

Additionally, I'd add the abbreviation definitions of "AM" and "FM" to the device docstring, just in case. I guess they mean amplitude and frequency modulation.

pymeasure/instruments/hp/hp8657b.py
LongnoseRob: I like the idea, 
however 'Instrument.control()' does not accecpt Enums as inputs.
Or did I miss a change regarding this?

bmoneke: If you use an IntEnum, the enumeration member will be interpreted as an integer, whenever an integer is expected.
I did not test it yet, but it should work, as `"Command %i" % Modulation.OFF` will be `"Command 4"`, if 
```
class Modulation(IntEnum):
    OFF = 4
```

LongnoseRob: Thanks, it was the `map_values=True` which does not like the IntEnum.
But as this unit is listen-only we do not `map_values`

pymeasure/instruments/hp/hp8657b.py
LongnoseRob: added a shot glossary with https://github.com/pymeasure/pymeasure/pull/732/commits/6d61a9ae93fb0335ed54913de60c07202716f59d

pymeasure/instruments/hp/hp8657b.py
```diff
@@ -0,0 +1,200 @@
...
+        log.warning("HP8657B Does not support error status readout")
+
+    def clear(self):
+        """
+        Reset the instrument to power-on default settings
+
+        """
+        self.adapter.connection.clear()
+
+    id = "HP 8657B Signal generator"  #: Manual ID entry, as the instrument does not talk with us
```
msmttchr, line 74: @LongnoseRob, nice idea to have an `id` anyway for uniformity with other scpi instruments.
I would consider to add the string in IEEE 488.2 format.
I report here an excerpt from a documentation for an Anritsu instrument.

`identification query. query only. this query returns an instrument identification string in ieee-488.2 specified <arbitrary ascii> format consisting of four fields separated by commas. the fields are: <manufacturer>, <model>, <serial #>,<firmware revision level>`

So, in your case, it should be something like:
`"HP,8657B,N/A,N/A"`

What do you think about this?
If there is a general agreement, we could put this as suggestion in PR #712.

LongnoseRob: > So, in your case, it should be something like:
"HP,8657B,N/A,N/A"
 What do you think about this?

Thanks for the feedback and the improvement idea.
One question the _implementation_, would t 
```
id = "HP,8657B,N/A,N/A"
```
be sufficient or should it be 

```
def id(self):
    """"FAKE *IDN? return"""
    return "HP,8657B,N/A,N/A"
```

msmttchr: Well, I think that property exists to allow access to a value that requires some calculation (i.e. code execution) like an attribute. For example:
```
@property
    def id(self):
        if self.SCPI:
            return self.ask("*IDN?").strip()
        else:
```
In our case, we don't need to run any code, so we can go back to plain attribute assignment.

docs/api/instruments/hp/hp8657B.rst
```diff
@@ -4,9 +4,20 @@ HP Signal generator HP8657B
 
 
 *Note:*
-    * This instrument does not support reading back from the instrument, as it is listen-only GPIB device
+    * This instrument does not support reading back values, as it is a listen-only GPIB device
     
-    * Other instrument of this family could be implemented using the dymaic ranges feature
+    * Other instruments of this family could be implemented using the dymaic ranges feature
+    
+    * Optorional pulse modulation feature is not support yet
```
bmoneke, line 11:
```suggestion
    * Other instruments of this family could be implemented using the dynamic ranges feature.
    
    * Optorional pulse modulation feature is not supported yet.
```
What does "Optorional" mean?

pymeasure/instruments/hp/hp8657b.py
```diff
@@ -162,9 +172,9 @@ def clear(self):
         )
 
     level_offset = Instrument.setting(
-        "AO%gDB",
+        "AO %g DB",
         """
-        sets the output offset in dB, usable range -199 to +199 dB
+        Sets the output offset in dB, usable range -199 to +199 dB
```
bmoneke, line 177:
```suggestion
        Set the output offset in dB, usable range -199 to +199 dB.
```

pymeasure/instruments/hp/hp8657b.py
```diff
@@ -174,7 +184,7 @@ def clear(self):
     output_enabled = Instrument.setting(
         "R%d",
         """
-        controls if the output is enabled
+        Controls the output
```
bmoneke, line 187:
```suggestion
        Control whether the output is enabled.
```

bmoneke:
A few more comments, mainly regarding docstring formulation.

pymeasure/instruments/hp/hp8657b.py
```diff
@@ -0,0 +1,200 @@
...
+
+        *Note:*
+            * AM & FM can be active at the same time
+            * only one internal source can be active at the time
+            * use "OFF" to deactivate AM
+
+        """,
+        validator=strict_discrete_set,
+        values=Modulation,
+        )
```
bmoneke, line 108: Please give the user a hint, that he might use the `Modulation` enum (similarly below). Otherwise the user will never know.

bmoneke (changes requested):
Having implemented that small doc change, it will look good to me.
@msmttchr do you have anything else?

msmttchr:
> Having implemented that small doc change, it will look good to me. @msmttchr do you have anything else?

I agree. It looks good to me as well. Thanks

msmttchr (approved):
@LongnoseRob thanks for your contribution. It looks good to me.
@bmoneke feel free to merge after it is good for you.

pymeasure/instruments/hp/hp8657b.py
LongnoseRob: I will add a codeblock exmaplew how to use modulations

pymeasure/instruments/hp/hp8657b.py
```diff
@@ -0,0 +1,222 @@
...
+        usage example:
+
+        .. code-block:: python
+
+            sig_gen = HP8657B("GPIB::7")
+            ...
+            sig_gen.am_source = sig_gen.Modulation.Int_400Hz    #  Enable int. 400 Hz source for AM
+            sig_gen.am_depth = 50                               #  Set AM modulation depth to 50%
+            ...
+            sig_gen.am_source = sig_gen.Modulation.OFF          #  Turn AM off
```
bmoneke, line 114:
```suggestion
        Set the source for the AM function with :attr:`Modulation` enumeration.

        ==========  =======
        Value       Meaning
        ==========  =======
        OFF         no modulation active
        INT_400HZ   internal 400 Hz modulation source
        INT_1000HZ  internal 1000 Hz modulation source
        EXTERNAL    External source, AC coupling
        ==========  =======

        *Note:*
            * AM & FM can be active at the same time
            * only one internal source can be active at the time
            * use "OFF" to deactivate AM

        usage example:

        .. code-block:: python

            sig_gen = HP8657B("GPIB::7")
            ...
            sig_gen.am_source = sig_gen.Modulation.INT_400HZ    #  Enable int. 400 Hz source for AM
            sig_gen.am_depth = 50                               #  Set AM modulation depth to 50%
            ...
            sig_gen.am_source = sig_gen.Modulation.OFF          #  Turn AM off
```

bmoneke (commented):
I suggest some changes (made more clear how to set the values and adjusted the docstrings to the Enum).

Thank you for developing the usage of enums together with me. As it is a new usage, treading new paths requires some work, thanks for staying with it.
With your hints, we added it into the documentation and will improve their usage. Thanks for the feedback!

LongnoseRob:
> I suggest some changes (made more clear how to set the values and adjusted the docstrings to the Enum).
> 
> Thank you for developing the usage of enums together with me. As it is a new usage, treading new paths requires some work, thanks for staying with it. With your hints, we added it into the documentation and will improve their usage. Thanks for the feedback!

Thank you for the code-review and the constructive critique!.

pymeasure/instruments/hp/hp8657b.py
```diff
@@ -0,0 +1,222 @@
...
+        map_values=True
+       )
+
+    def reset(self):
+        self.adapter.connection.clear()
+
+    def shutdown(self):
+        self.output_enabled = False
+        self.adapter.connection.clear()
+        self.adapter.connection.close()
```
bilderbuchi, line 222: Arguably, this should/could call `super().shutdown()`, too (although, cf #513). @bmoneke

bmoneke: Oh, yes...

LongnoseRob: Should I add this line `super().shutdown()`  and push again?
O should we  check all non-SCPI implementations and check if `shutdown()` has this line?

bmoneke: The pull request is already merged. I can just add this line directly.

LongnoseRob: Ok, thats also fine with me.
Thanks

bilderbuchi: The PR has been merged, so you'd need to open another PR for that.
Arguably, all instruments that have redefined `shutdown` should call the parent method, too (or have a good/documented-by-comment reason not to). _That_ would be a useful PR, if you want to. :-)

LongnoseRob: > The PR has been merged, so you'd need to open another PR for that. Arguably, all instruments that have redefined `shutdown` should call the parent method, too (or have a good/documented-by-comment reason not to). _That_ would be a useful PR, if you want to. :-)

A quick greo only gave about 20 instruments with `includeSCPI=False`, so Ill open a new PR for that

PR #737, opened 2022-11-08
[WIP] adding proper shutdown sequence to many non-SCPI instruments

LongnoseRob:
after #732 was merged, it was found that the `shutdown()` definition did not contan the recommended `super.shutdown()`
The initial commit includes many instruments with `includeSCPI=False`, however a _proper_ review of the codebase still might be recommended.

### DIFF
--- a/pymeasure/instruments/anaheimautomation/dpseriesmotorcontroller.py
+++ b/pymeasure/instruments/anaheimautomation/dpseriesmotorcontroller.py
@@ -337,3 +337,8 @@ class DPSeriesMotorController(Instrument):
         """
         while self.busy:
             sleep(interval)
+
+    def shutdown(self):
+        self.adapter.connection.clear()
+        self.adapter.connection.close()
+        super.shutdown()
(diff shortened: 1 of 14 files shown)


bilderbuchi:
Sorry, I think you misunderstood me or I expressed myself badly. This only should be added to instruments that have overridden `Instrument.shutdown` with their own implementation, to make sure these also call the super version!

Also, why do you exclude SCPI instruments from consideration? As I see it, all instruments should be the same for this, no?

LongnoseRob:
> Sorry, I think you misunderstood me or I expressed myself badly. This only should be added to instruments that have overridden `Instrument.shutdown` with their own implementation, to make sure these also call the super version!
> 
> Also, why do you exclude SCPI instruments from consideration? As I see it, all instruments should be the same for this, no?

Sorry also from my side, I tough that 'shutdown()' would be one of the items which were only required for non-SCPI implementations.
But looking at 'instrument.py' this also affects SCPI instruments.

LongnoseRob:
Closing this PR for the time being, will keep this topic on my list and have a more structual sound approach

PR #738, opened 2022-11-10
Fix broken documentation example.

bmoneke:
The example for creating a new instrument with dynamic properties missed a "dynamic=True". I added a test to catch that.


msmttchr (approved):
It looks good to me. Thanks this will save some questions from users :)

bilderbuchi:
Thanks!

PR #739, opened 2022-11-12
Shutdown code unification

LongnoseRob:
[As mentioned](https://github.com/pymeasure/pymeasure/pull/732#discussion_r1016650307) #732  when instruments overwrite the `shutdown()` method, that method should contain `super().shutdown()`.

Topics for disucssion:

1. Some [instruments](https://github.com/LongnoseRob/pymeasure/blob/d639d925d41931458d5fc7abd6983919942e3c42/pymeasure/instruments/ametek/ametek7270.py#L212)  use `self.isShutdown = True`, would this still be required if we add `super().shutdown()` ?
2. `pymeasure/instruments/ni/virtualbench.py` has 2 definitions of `shutdown`, the influence of  adding `super().shutdown()` should be properly checked

### DIFF
--- a/pymeasure/instruments/ametek/ametek7270.py
+++ b/pymeasure/instruments/ametek/ametek7270.py
@@ -1,4 +1,4 @@
-#
+/#
 # This file is part of the PyMeasure package.
 #
 # Copyright (c) 2013-2022 PyMeasure Developers
@@ -208,7 +208,6 @@ class Ametek7270(Instrument):
 
     def shutdown(self):
         """ Ensures the instrument in a safe state """
-        self.voltage = 0.
-        self.isShutdown = True
-        super().shutdown()
         log.info("Shutting down %s" % self.name)
+        self.voltage = 0.
+        super().shutdown()
--- a/pymeasure/instruments/ametek/ametek7270.py
+++ b/pymeasure/instruments/ametek/ametek7270.py
@@ -210,4 +210,5 @@ class Ametek7270(Instrument):
         """ Ensures the instrument in a safe state """
         self.voltage = 0.
         self.isShutdown = True
+        super().shutdown()
         log.info("Shutting down %s" % self.name)
--- a/pymeasure/instruments/ametek/ametek7270.py
+++ b/pymeasure/instruments/ametek/ametek7270.py
@@ -1,4 +1,4 @@
-/#
+#
 # This file is part of the PyMeasure package.
 #
 # Copyright (c) 2013-2022 PyMeasure Developers
--- a/pymeasure/instruments/keithley/keithley2400.py
+++ b/pymeasure/instruments/keithley/keithley2400.py
@@ -763,3 +763,4 @@ class Keithley2400(Instrument, KeithleyBuffer):
             self.ramp_to_voltage(0.0)
         self.stop_buffer()
         self.disable_source()
+        super().shutdown()
--- a/pymeasure/instruments/keithley/keithley2450.py
+++ b/pymeasure/instruments/keithley/keithley2450.py
@@ -635,3 +635,4 @@ class Keithley2450(Instrument, KeithleyBuffer):
             self.ramp_to_voltage(0.0)
         self.stop_buffer()
         self.disable_source()
+        super().shutdown()
--- a/pymeasure/instruments/keithley/keithley6221.py
+++ b/pymeasure/instruments/keithley/keithley6221.py
@@ -410,6 +410,7 @@ class Keithley6221(Instrument, KeithleyBuffer):
         """ Disables the output. """
         log.info("Shutting down %s." % self.name)
         self.disable_source()
+        super().shutdown()
 
     ###############
     # Status bits #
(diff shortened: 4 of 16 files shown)


pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -210,4 +210,5 @@ def shutdown(self):
         """ Ensures the instrument in a safe state """
         self.voltage = 0.
         self.isShutdown = True
+        super().shutdown()
```
msmttchr, line 213: @LongnoseRob, as you mentioned in the overall comment, I would remove line 212, but also 214. They are both included in `super`.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -763,3 +763,4 @@ def shutdown(self):
             self.ramp_to_voltage(0.0)
         self.stop_buffer()
         self.disable_source()
+        super().shutdown()
```
msmttchr, line 766: Line 759 could be removed (included in `super`)

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -635,3 +635,4 @@ def shutdown(self):
             self.ramp_to_voltage(0.0)
         self.stop_buffer()
         self.disable_source()
+        super().shutdown()
```
msmttchr, line 638: Log message could be removed

pymeasure/instruments/keithley/keithley6221.py
```diff
@@ -410,6 +410,7 @@ def shutdown(self):
         """ Disables the output. """
         log.info("Shutting down %s." % self.name)
         self.disable_source()
+        super().shutdown()
```
msmttchr, line 413: Log message could be removed

pymeasure/instruments/keithley/keithley6517b.py
```diff
@@ -347,3 +347,4 @@ def shutdown(self):
         self.ramp_to_voltage(0.0)
         self.stop_buffer()
         self.disable_source()
```
msmttchr, line 349: Log message could be removed.
(I will not comment about this for the remaining files)

pymeasure/instruments/ni/virtualbench.py
```diff
@@ -128,6 +128,7 @@ def shutdown(self):
         log.info("Shutting down %s" % self.name)
         self.vb.release()
         self.isShutdown = True
+        super().shutdown()
```
msmttchr, line 131: This instrument surprisingly (to me) does not inherit from `Instrument` class, so super is not needed and it will fail.

msmttchr (changes requested):
@LongnoseRob, please find attached my comments.
I have highlighted that log messages can be removed, since they are included in the parent's call.
I am not sure whether to recommend to call `super` at the beginning or at the end of the method, so I didn't comment about this, but may be other maintainers have some suggestion. Do you have any opinion about this?

LongnoseRob:
> @LongnoseRob, please find attached my comments. I have highlighted that log messages can be removed, since they are included in the parent's call.

@msmttchr Thank you for the review, I will remove the log messages in the next commits

> I am not sure whether to recommend to call `super` at the beginning or at the end of the method, so I didn't comment about this, but may be other maintainers have some suggestion. Do you have any opinion about this?

I am leaning towards the end of the method, in my opinion first we should take the instrument to a safe condition and after that stopp the communication with the super-call.
But let us see what the maintainers and other contributors have for  opinions

bilderbuchi:
> I am not sure whether to recommend to call `super` at the beginning or at the end of the method, so I didn't comment about this, but may be other maintainers have some suggestion. Do you have any opinion about this?

At the end. When the main super will also close down the connection (or do other generic housekeeping), otherwise the connection is dead when the specific instrument wants to do its cleanup.

pymeasure/instruments/keithley/keithley2450.py
bilderbuchi: This is interesting -- the log message is useful as it tells the user that shutdown has been *initiated*.

As the log message in `Instrument.shutdown` comes at the end of that method, when everything is finished and nothing more is expected to happen, we should change the text of _that_ message to `log.info(f"Finished shutting down {self.name)}")`

pymeasure/instruments/ni/virtualbench.py
bilderbuchi: Yes, the VirtualBench is a separate beast. We should probably clean this up at some point and derive from `Instrument`. I don't know off the cuff how "special" it is, though.

LongnoseRob:
Tracking items:

- [ ] `super().shutdown()` should be last entry in the `shutdown` method
- [ ] logging messages will be kept to inform of starting and finishing shutdown of instrument
- [ ] change text of log-message in `Instrument.shutdown()` to `log.info(f"Finished shutting down {self.name)}")`

msmttchr (approved):
Good for me. Thanks.

bilderbuchi (approved):
LGTM, too. Thank you for your contribution!

PR #697, opened 2022-09-30
LeCroy T3DSO1204 Oscilloscope

LastStarDust:
This implements a device of a new vendor: [LeCroy/Teledyne T3DSO1204 Oscilloscope](https://teledynelecroy.com/oscilloscope/oscilloscopemodel.aspx?modelid=11007&capid=102&mid=504).

Connection to the device is made through an TCP or USB connection.

I based the `pymeasure.instruments.lecroy.LeCroyT3DSO1204` on the `pymeasure.instruments.keysight.KeysightDSOX1102G` but I added many functions, in particular related to the trigger.

I strived for 100% test coverage. However, to run the tests you need the hardware.

PR #740, opened 2022-11-13
Adapt behaviour of get_process with multiple values.

bilderbuchi:
This merges #637 with its changes to the behaviour of get_process when receiving multiple values, and makes sure the change is documented accordingly.

As a first step, I codified the behaviour of get_process when receiving multiple values.

### DIFF
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -362,7 +362,7 @@ class CommonBase:
             as to map values if :code:`map_values` is True.
         :param map_values: A boolean flag that determines if the values should be
             interpreted as a map
-        :param get_process: A function that take a value and allows processing
+        :param get_process: A function that takes a value or list of values and allows processing
             before value mapping, returning the processed value
         :param set_process: A function that takes a value and allows processing
             before value mapping, returning the processed value
@@ -494,7 +494,7 @@ class CommonBase:
             as to map values if :code:`map_values` is True.
         :param map_values: A boolean flag that determines if the values should be
             interpreted as a map
-        :param get_process: A function that take a value and allows processing
+        :param get_process: A function that takes a value or list of values and allows processing
             before value mapping, returning the processed value
         :param command_process: A function that take a command and allows processing
             before executing the command, for getting
--- a/tests/instruments/test_common_base.py
+++ b/tests/instruments/test_common_base.py
@@ -511,6 +511,24 @@ def test_control_get_process(dynamic):
 
 
 @pytest.mark.parametrize("dynamic", [False, True])
+def test_control_get_process_multiple_values(dynamic):
+    """get_process acts on the whole list in case we receive multiple values."""
+    class Fake(FakeBase):
+        x = CommonBase.control(
+            "", "JUNK%d, JUNK%d", "",
+            get_process=lambda v: [int(entry.replace('JUNK', '')) for entry in v],
+            # get_process=lambda v: int(v.replace('JUNK', '')),  #desired single-entry application
+            dynamic=dynamic,
+        )
+
+    fake = Fake()
+    fake.x = 5, 6
+    assert fake.read() == 'JUNK5, JUNK6'  # the raw reply from the instrument
+    fake.x = 5, 6
+    assert fake.x == [5, 6]
+
+
+@pytest.mark.parametrize("dynamic", [False, True])
 def test_control_preprocess_reply_property(dynamic):
     # test setting preprocess_reply at property-level
     class Fake(FakeBase):
@@ -615,7 +633,7 @@ def test_setting_process(dynamic):
 
 
 @pytest.mark.parametrize("dynamic", [False, True])
-def test_control_multivalue(dynamic):
+def test_control_set_multiple_value(dynamic):
     class Fake(FakeBase):
         x = CommonBase.control(
             "", "%d,%d", "",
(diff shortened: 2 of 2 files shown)


bilderbuchi:
I went through the codebase, and have compiled a list of potentially affected properties. It is much larger than expected, but there are surely some `get_process` false positives in there that would actually not be affected by this change (i.e. they don't actually expect to process a list of entries). It's hard to tell when just grepping through the codebase - I basically looked for what could be list indexing. `*sigh*`

For these cases, we have to 
* find out if they would be affected by the change in behaviour of `get_process`
*  if in _any_ doubt that we can exclude the property, implement a protocol test to nail down its behaviour before we merge in #637.

* [ ] AgilentE4980.impedance
* [ ] AnritsuMS9710C.trace_marker
* [ ] BKPrecision9130B.channel
* [ ] Fluke7341.id
* [ ] HP8116A.operating_mode, control_mode, trigger_slope, shape
* [ ] Keithley2306.last_readings, readings, measured_voltages, measured_currents, dvm_voltages, pulse_currents, long_integration_currents
* [ ] Keithley2750.closed_channels
* [ ] Keithley6517B.current, voltage, resistance
* [ ] DSOX1102G.label

Help appreciated.

bilderbuchi:
The `get_process` introduced by  @LastStarDust in https://github.com/pymeasure/pymeasure/pull/697/commits/6b928f2a52bd7a4f777232db938fee48d3b8ef67 needs, afaict, to find the position/index of some string elements in the received comma-separated list.
We need to identify a way to implement that behaviour even with the changed `get_process` behaviour (and also offer a good migration strategy) -- probably using an appropriate `preprocess_reply` (something like `lambda reply: reply.split(',').index("NP")+1`).

bmoneke:
> The `get_process` introduced by @LastStarDust in [6b928f2](https://github.com/pymeasure/pymeasure/commit/6b928f2a52bd7a4f777232db938fee48d3b8ef67) needs, afaict, to find the position/index of some string elements in the received comma-separated list, so acts on the list coming out of `values`. We need to identify a way to implement that behaviour even with the changed `get_process` behaviour (and also offer a good migration strategy) -- probably using an appropriate `preprocess_reply` (something like `lambda reply: reply.split(',').index("NP")+1`).

I did not see, where LastStarDust used the changed `get_process` as he moved that logic to the `read` method, it seems.

However, if you need to work on the list as a whole, I propose to use a normal python property instead of our property factories. That keeps the code simple. We never expected to cover all possible use cases with the property creators.

Earlier, `get_process` worked only on single values. Now, it works on every value (also on a single value). Therefore, if only one value is returned, the behavior did not change (tests should confirm this). Using `get_process` with several values was useless before, therefore I do not expect anyone to have used this combination, but we should check it.

bilderbuchi:
> Using get_process with several values was useless before, therefore I do not expect anyone to have used this combination, but we should check it.

My point is, the commit/PR I linked does exactly that, and it's not useless. 
Therefore, we have to assume that others did something similar, so we need a clear action path, and need to check the potential uses I listed above.

bmoneke:
Oh sorry, I made a mistake remembering the code. I had thought, that `get_process` was not used if `len(vals)>1`, but now I see, that it was used on the returned list.

Now I see the point of potential harm.

`preprocess_reply` seems a viable option here, using regex or splitting at "NP" and then at "," (each time only one split and using the last, and first result, respectively).

> I can rebase this branch after we have merged https://github.com/pymeasure/pymeasure/pull/697, add/adapt the tests, and confirm they still pass. Or they adapt the usage beforehand, and test with their device at hand.

As we want to merge #637 anyway, I propose, that #697 should be adjusted beforehand (`preprocess_reply` will work already).

bilderbuchi:
> EDIT: TC038.information talks only about a single value, besides, a protocol test is in place.

Ah, I had realised that later -- I thought I had edited it out afterwards.

bmoneke:
I looked at the first four entries and they all do list indexing, such that they won't work afterwards. Right now, I don't know how they should work after the change. Changing `read` is no option, as they all do specific work (selecting some entries of the values list or even using some chars of one list entry and other chars of another entry - Fluke.id).

Maybe we should do `vals = get_process(vals)`. That way, we have to change all those, who do get_process on a single value, but the change is straight forward. Or something totally different?

Here my other findings:

- The HP uses list slicing as well.
- The Keithley2306 uses `get_process` to always return a list (even if only one value), such that it won't work either
- Keithleiy2750 uses a `get_process` which works for every entry in the list, if it is a list, or on the single value, therefore it can be adapted easily to the new code
- Keithley6517 does regex on the strings: no issues here.

bilderbuchi:
@dkriegner am I correct that **only** the `sensor_typeA/B/C` properties need the multiple return value mapping functionality? I'm thinking if we should separate this from the rest of the instrument contribution, so we can get the instrument merged (with a bit less functionality), and deal with the pymeasure core change separately? 
If you agree, I'd squash your PR and tease apart the contribution into those two separate parts.

dkriegner:
> @dkriegner am I correct that **only** the `sensor_typeA/B/C` properties need the multiple return value mapping functionality? 

yes, you are right.

>I'm thinking if we should separate this from the rest of the instrument contribution, so we can get the instrument merged (with a bit less functionality), and deal with the pymeasure core change separately? If you agree, we can squash your PR and tease apart the contribution into those two separate parts.

I agree this is a sensible thing since it seems this core change is a bit of a bigger action.

bmoneke:
Should we discuss here or elsewhere, what do do with `get_process`?

bilderbuchi:
> I've pushed all remaining code (i.e. the multiple value feature and the sensor property) to a branch in my fork, so nothing's lost. I just want to put complete-ish tests in place, first. :-)

Note (to preserve information): that branch (which will get put onto this PR when the dust has settled) is https://github.com/bilderbuchi/pymeasure/tree/multi_map_values

bmoneke:
In #781 we convened on using `get_process` on the individual values and `get_process_list` on the whole list, but only one of them both may be used at most.

Do you need help implementing the tests/adapting the instruments?

bilderbuchi:
I have no bandwidth right now to tackle this issue (want to get the thread-safe feature done first), so if you want to take this on, you're welcome to! :-)

The way I planned to do this is
* Find the places where we could potentially break things (basically, the list above)
* Implement tests for those places (TDD style, to lock in current behaviour)
* Then pull in/reuse my branch https://github.com/bilderbuchi/pymeasure/tree/multi_map_values, code/complete the feature, and check that the tests still pass

BenediktBurger:
Implemented by #1291

PR #741, opened 2022-11-14
Refactor Results object to be datastore agnostic

physicsninja:
This PR is a rough sketch of a reimplementation of the results class to be data store agnostic so that users can write to different pipelines using a prescribed framework rather than having to hack apart the `Results` class. The `Results` class is changed to `ResultsBase` which has five methods that need to be implement by subclasses. This refactor takes the suggestion of @mivade in #136 of going one step higher in `Logging` objects and using `Logging.Handler`'s as the base interface instead of `Formatter`'s. Thi s allows the use of non file-based handlers. However, file-based handlers are actually super common, so I have only re-implements something based on `Logging.FileHandler`.

As an example of a different choice, I have implemented a JSONResults that inherits from `FileBaseResults` class that stores data in a somewhat silly way using json. If you change the example in gui.py to take JSONResults instead of CSVResults it should be clear the choice being made.

This should be easily extensible to taking full csv or pandas dicts as in #277.

I know this is a pretty big refactor so please let me know if you have any questions or suggestions or if you think it is folly.


physicsninja:
I have a keen interest in pushing something like this across the finish line if it is judged useful so I guess the first pass would be if this is seems like a good direction at all

bmoneke:
Thanks for your suggestion and contribution.

From our recent instrument changes, we learned, that it is better, to discuss the ideas behind an implementation in an issue than in a PR, where conversations are more difficult to find. Therefore, I propose to use one of the aforementioned issues or create a new issue (as you deem best) to discuss the idea and usage philosophy.
That way the PR discussion remains about implementation details.

At a first glance, I like your idea, but I am more familiar with the instruments than with the measurement part of pymeasure, which I do not use (yet).

BenediktBurger:
Closed due to lack of response. It might be reopened whenever there is interest.

PR #756, opened 2022-11-19
Add dialog box for sequences

mcdo0486:
This PR splits out class SequencerTreeWidget GUI an logic around the sequence tree from class SequenceWidget. 

Add SequenceDialog(QtWidgets.QFileDialog) which can load sequences and uses SequencerTreeWidget to preview what sequence will be loaded.


mcdo0486:
Moved development to PR https://github.com/pymeasure/pymeasure/pull/518

Issue #746, opened 2022-11-15
Rework existing instruments with channels


bilderbuchi:
After #718 lands, we will have a better way for implementing instruments with channels. Existing instruments in the codebase should be migrated to the new approach. Let us collect a list of relevant instruments here:
* [ ] Siglent SPDxxxxX (#719)

msmttchr:
I wasn't aware you can edit other people comment. Is this the correct way to populate the list ?

PR #758, opened 2022-11-22
dev/SPDxxxxX_channels

AidenDawn:
Following on from #718 & #719, I thought I'd try converting the SPD PSU channels to see how it would go. In general, it seems to have gone quite smoothly and it should be possible with a few minor tweaks to the current `SPDChannel` class. The most challenging aspect of the change was what to do with the methods inside `SPDChannel`, can anyone else see a better way of writing these methods? @bmoneke, would you mind having a look over this to see if there are any points in the implementation you would suggest doing differently (on a channels specific basis). This should start ticking off things in #746

### DIFF
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
@@ -36,4 +36,4 @@ class SPD1168X(SPDSingleChannelBase):
             **kwargs
         )
 
-        self.ch[1] = SPDChannel(self, 1)
+        self.add_child(SPDChannel, 1)
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
@@ -21,8 +21,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 #
-from pymeasure.instruments.siglenttechnologies.siglent_spdbase import (SPDSingleChannelBase,
-                                                                       SPDChannel)
+from pymeasure.instruments.siglenttechnologies.siglent_spdbase import SPDSingleChannelBase
 
 
 class SPD1168X(SPDSingleChannelBase):
@@ -35,5 +34,3 @@ class SPD1168X(SPDSingleChannelBase):
             name="Siglent Technologies SPD1168X Power Supply",
             **kwargs
         )
-
-        self.add_child(SPDChannel, 1)
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
@@ -21,12 +21,16 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 #
-from pymeasure.instruments.siglenttechnologies.siglent_spdbase import SPDSingleChannelBase
+from pymeasure.instruments.instrument import Instrument
+from pymeasure.instruments.siglenttechnologies.siglent_spdbase import (SPDSingleChannelBase,
+                                                                       SPDChannel)
 
 
 class SPD1168X(SPDSingleChannelBase):
     """Represent the Siglent SPD1168X Power Supply.
     """
+
+    channels = Instrument.ChannelCreator(SPDChannel, 1)
 
     def __init__(self, adapter, **kwargs):
         super().__init__(
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
@@ -39,7 +39,7 @@ class SPD1305X(SPDSingleChannelBase):
         voltage_limits = [0, 30]
         current_limits = [0, 5]
 
-        self.ch[1] = SPDChannel(self, 1, voltage_limits, current_limits)
+        self.add_child(SPDChannel, 1)
 
-        self.ch[1].voltage_setpoint_values = voltage_limits
-        self.ch[1].current_limit_values = current_limits
+        self.ch_1.voltage_setpoint_values = voltage_limits
+        self.ch_1.current_limit_values = current_limits
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
@@ -30,6 +30,8 @@ class SPD1305X(SPDSingleChannelBase):
     """Represent the Siglent SPD1305X Power Supply.
     """
 
+    channels = Instrument.ChannelCreator(SPDChannel, 1)
+
     def __init__(self, adapter, **kwargs):
 
         super().__init__(
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
@@ -30,7 +30,11 @@ class SPD1305X(SPDSingleChannelBase):
     """Represent the Siglent SPD1305X Power Supply.
     """
 
-    channels = Instrument.ChannelCreator(SPDChannel, 1)
+    voltage_range = [0, 30]
+    current_range = [0, 5]
+    channels = Instrument.ChannelCreator(SPDChannel, 1,
+                                         voltage_range=voltage_range,
+                                         current_range=current_range)
 
     def __init__(self, adapter, **kwargs):
 
@@ -39,12 +43,6 @@ class SPD1305X(SPDSingleChannelBase):
             name="Siglent Technologies SPD1305X Power Supply",
             **kwargs
         )
-        voltages = [0, 30]
-        currents = [0, 5]
 
-        self.channels = Instrument.ChannelCreator(SPDChannel, 1,
-                                                  voltage_range=voltages,
-                                                  current_range=currents)
-
-        self.ch_1.voltage_setpoint_values = voltages
-        self.ch_1.current_limit_values = currents
+        self.ch_1.voltage_setpoint_values = self.voltage_range
+        self.ch_1.current_limit_values = self.current_range
--- a/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
@@ -36,10 +36,10 @@ class SPD1305X(SPDSingleChannelBase):
             name="Siglent Technologies SPD1305X Power Supply",
             **kwargs
         )
-        voltage_limits = [0, 30]
-        current_limits = [0, 5]
+        voltages = [0, 30]
+        currents = [0, 5]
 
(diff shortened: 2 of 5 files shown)


pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -221,30 +217,15 @@ def __init__(self, instrument,
...
         """Enable the channel output.
 
         :type: bool
             ``True``: enables the output
             ``False``: disables it
         """
-        self.instrument.selected_channel = self.channel
-        self.write(f'OUTP CH{self.channel},{("OFF","ON")[enable]}')
+        self.parent.selected_channel = self.id
+        self.write('OUTP CH{ch},' + ("OFF", "ON")[enable])
```
bmoneke, line 228: Interesting usage.
Afaik, True is not guaranteed to be equal to 1, therefore I suggest:
```suggestion
        self.write('OUTP CH{ch},' + ("ON" if enable else "OFF"))
```

pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
```diff
@@ -36,10 +36,10 @@ def __init__(self, adapter, **kwargs):
             name="Siglent Technologies SPD1305X Power Supply",
             **kwargs
         )
-        voltage_limits = [0, 30]
-        current_limits = [0, 5]
+        voltages = [0, 30]
+        currents = [0, 5]
 
-        self.ch[1] = SPDChannel(self, 1, voltage_limits, current_limits)
+        self.add_child(SPDChannel, 1, voltage_range=voltages, current_range=currents)
```
bmoneke, line 42: We made a `ChannelCreator`, which works similar to the property creators, such that you do not need to add it to the init.
Instead you use it at class level:
~~~python
class xy(Instrument):
    channels = ChannelCreator(SPDChannel, 1, **kwargs)
~~~

The result will be, as if you called add_child at the beginning. `kwargs` is the dictionary of kwargs for the Channel initialization.

pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
```diff
@@ -36,4 +36,4 @@ def __init__(self, adapter, **kwargs):
             **kwargs
         )
 
-        self.ch[1] = SPDChannel(self, 1)
+        self.add_child(SPDChannel, 1)
```
bmoneke, line 39: See comment at spd1305.

bmoneke:
Thanks for tackling that issue. It already looks good.
I left a few comments.

`add_child` is more used for dynamic creation. Instead we have a `ChannelCreator`. You might even put that creator in the "SingleChannelBase". You will have to adjust the limits in the 1305er instrument manually (in the init) or override the channels object in the 1305er with another ChannelCreator:
~~~python
class SIG1305(Instrument):
    channels = ChannelCreator...
~~~

You could make some channel methods to properties (`enable_output` to `output_enabled`), but that has nothing to do with the channels.

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
AidenDawn: I've been using this construction myself for a while in Python now. I've found it to be quite a neat way of differentiating between 2 states when constructing strings. I understand the confusion though, I spent the other half of my days working in C/C++ where true can often be anything other than 0

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -143,9 +142,9 @@ def enable_local_interface(self, enable: bool = True):
     def shutdown(self):
         """ Ensure that the voltage is turned to zero
         and disable the output. """
-        for channel in self.ch:
-            channel.voltage_setpoint = 0
-            channel.enable_output(False)
+        for ch in self.channels:
```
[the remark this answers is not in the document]
AidenDawn: Done

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
bilderbuchi: I know it works, but I find it clearer if spelled out (`"ON" if enable else "OFF"`).
The other way reminds me to of using implicit casts from bool to int/float in C++ like `somevalue = x + somefunc(y)*(foo >= bar)` - I don't find that very readable, either.

pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
AidenDawn: I've got a couple of questions about how to move forward with this approach. Ive taken on board the suggestions from your last few comments here and have created the channels in SPDSingleChannelBase as so:
```python
class SPDSingleChannelBase(SPDBase):
    channels = Instrument.ChannelCreator(SPDChannel, 1)
```
but am running into a few issues. Firstly, you've included `**kwargs` in your example above. Why? This will be undefined at the class level. It would be possible if the channels object is overloaded in the `__init__` of the child instrument but this is only useful in the example for the 1305 instrument where the voltage and current range parameters are needed.

Secondly, this implementation consistently causes the tests to fail with the following error:
```python
FAILED tests/instruments/siglenttechnologies/test_siglent_spd1168x.py::test_set_current
AttributeError: 'SPD1168X' object has no attribute 'ch_1'
```
obviously this means that the channel init is failing somewhere but I can't seem to find the answer. Any ideas?

bmoneke: Sorry for that `**kwargs`. I meant it as a placeholder for the arguments you want to give to the Channel (e.g. The limits), but not literally.

I have to look at the code in detail.

AidenDawn: Let me commit what I've got so far. It's not working, but at least you might have a chance to see what's going on

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
AidenDawn: I've changed this over now, and also in the other places Id used the same construction

pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
bmoneke: Thanks for pushing.

I found the error: Right now, the automatic recognition only works, if the ChannelCreator is used in the last class, i.e. here in the SPD1168 class, not in the parent class, as we look just in the last class for ChannelCreators.

Therefore, you have to define `channels = ChannelCreator(...` in the SPD1168, not in the SingleClass (for now), then the tests succeed.
I'll open an issue and look into it.

Thank you for helping finding this bug.

pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
```diff
@@ -36,10 +37,12 @@ def __init__(self, adapter, **kwargs):
             name="Siglent Technologies SPD1305X Power Supply",
             **kwargs
         )
-        voltage_limits = [0, 30]
-        current_limits = [0, 5]
+        voltages = [0, 30]
+        currents = [0, 5]
 
-        self.ch[1] = SPDChannel(self, 1, voltage_limits, current_limits)
+        self.channels = Instrument.ChannelCreator(SPDChannel, 1,
```
bmoneke, line 43: That won't work. The channel creator has to be used at class level similarly to `control`. 
- You could move the variables at module level and then move the Creator at class level (similar to the Single Base Class).
- Or you move the creator (without the values) at class level and you set in this position `ch_1.voltage_range = xy`, similarly the current range. 
- Or you use add child in this position.

AidenDawn: Sorted

pymeasure/instruments/siglenttechnologies/siglent_spd1305x.py
```diff
@@ -29,17 +30,19 @@ class SPD1305X(SPDSingleChannelBase):
     """Represent the Siglent SPD1305X Power Supply.
     """
 
+    voltage_range = [0, 30]
+    current_range = [0, 5]
+    channels = Instrument.ChannelCreator(SPDChannel, 1,
+                                         voltage_range=voltage_range,
+                                         current_range=current_range)
```
bmoneke, line 37: At first I thought, that it does not work, but a fast test confirmed that it works.
Thanks for showing it, because that way also enums could work!

AidenDawn: The 1305 class was getting complicated enough that I felt it needed some test visibility. Glad it was useful. Does this mean that there is a more concise or clear way of doing what I've done here, if you didn't think it would work in the first place?

bmoneke: This is the best way!

I previously tried make something similar to work, but failed, because I used `SPD1305.voltage_values`, which did not work, because `SPD1305` is not yet defined. Therefore I told you to put the variables at module level (i.e. outside the class).

With all that `self` of the instances, I forgot, that class variables are "local variables" inside the class.

AidenDawn: Great, thanks

AidenDawn:
I think is ready to go now. Are there any more issues we can see? I don't see any reason for the last CI run to fail other than a failure during run setup. Linting and all linux runs passed.

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -138,137 +232,23 @@ def enable_local_interface(self, enable: bool = True):
             ``True``: enables the local interface
             ``False``: disables it.
         """
-        self.write(("*LOCK", "*UNLOCK")[enable])
+        self.write(("*UNLOCK" if enable else "*LOCK"))
```
bmoneke, line 235: I think you do not need the second pair of parenthesis. Formerly it was necessary to create a tuple.

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
```diff
@@ -138,137 +232,23 @@ def enable_local_interface(self, enable: bool = True):
...
-            float: current limit of the step
-        :param duration:
-            int: duration of the step in seconds
-        """
-        step = strict_discrete_range(step, [1, 5], 1)
-        voltage = truncated_range(voltage, self.voltage_range)
-        current = truncated_range(current, self.current_range)
-        duration = truncated_range(duration, [0, 10000])
-        self.write(f'TIME:SET CH{self.channel},{step:d},{voltage:1.3f},{current:1.3f},{duration:d}')
+        self.write(f'MODE:SET {("4W" if enable else "2W")}')
```
bmoneke, line 254: Neither do you need the parenthesis here.
```suggestion
        self.write(f'MODE:SET {"4W" if enable else "2W"}')
```

In one example it was needed to group the strings `"A" + ("B" if True else "C")` in order to combine AB or AC, but normally it is sufficient to write `print("A" if True else "B")`.

pymeasure/instruments/siglenttechnologies/siglent_spdbase.py
AidenDawn: Yep, onboard with that. I missed these

AidenDawn:
Looks good!

bmoneke:
Thanks for your work.

PR #761, opened 2022-11-25
Fix ChannelCreator resolution broken #759

### LINKED ISSUE #759: ChannelCreator does not work in parent class
In #758, the ChannelCreator did not work, unless used in the last class defined, because the class dict contains only the items defined in that class, not inherited.

Does "inspect" help here?

bmoneke:
Fixes #759 

- adds a test to ensure that inheritance (and overriding as expected) works
- outsources the channel creation to its own method
- fixes the problem

Fix description:

1. Look through the MRO (in reverse order) and collect all `ChannelCreator` objects in a dictionary, the key is the variable name of the object. The reverse order ensures, that the `ChannelCreator` assigned to `channels` in a subclass replaces that of a superclass assigned also to `channels`.
2. Create the children according to that collected dictionary.

Not working possibilites:

- `inspect.getmembers` fails, because it tries to read all the attributes, even `Instrument.setting`, which is not possible.

### DIFF
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -118,9 +118,10 @@ class CommonBase:
     # Prefix used to store reserved variables
     __reserved_prefix = "___"
 
-    def __init__(self):
+    def __init__(self, **kwargs):
         self._special_names = self._setup_special_names()
         self._create_channels()
+        super().__init__(**kwargs)
 
     class ChannelCreator:
         """Add channels to the parent class.
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -22,6 +22,7 @@
 # THE SOFTWARE.
 #
 
+from inspect import getmembers
 import logging
 
 log = logging.getLogger(__name__)
@@ -176,35 +177,23 @@ class CommonBase:
         special_names = []
         dynamic_params = tuple(set(self._fget_params_list + self._fset_params_list))
         # Check whether class variables of DynamicProperty type are present
-        for obj in (self,) + self.__class__.__mro__:
-            for attr_name, attr in obj.__dict__.items():
-                if isinstance(attr, DynamicProperty):
-                    special_names += [attr_name + "_" + key for key in dynamic_params]
+        for attr_name, attr in getmembers(self.__class__):
+            if isinstance(attr, DynamicProperty):
+                special_names += [attr_name + "_" + key for key in dynamic_params]
         # Check if special variables are defined at class level
-        for obj in (self,) + self.__class__.__mro__:
-            for attr in obj.__dict__:
-                if attr in special_names:
-                    # Copy class special variable at instance level, prefixing reserved_prefix
-                    setattr(self, self.__reserved_prefix + attr, obj.__dict__[attr])
+        for attr, value in getmembers(self.__class__):
+            if attr in special_names:
+                # Copy class special variable at instance level, prefixing reserved_prefix
+                setattr(self, self.__reserved_prefix + attr, value)
         return special_names
 
     def _create_channels(self):
-        """Create channels according to the ChannelCreator objects.
-
-        If a subclass overrides a ChannelCreator object with another one, the
-        one of the subclass is used, not that of the superclass.
-        """
-        channel_creators = {}
-        # Collect all ChannelCreators, override in order of subclasses
-        for obj in reversed((self,) + self.__class__.__mro__):
-            for item, value in obj.__dict__.items():
-                if isinstance(value, self.ChannelCreator):
-                    channel_creators[item] = value
-        # Create the channels
-        for name, creator in channel_creators.items():
-            for cls, id in creator.pairs:
-                child = self.add_child(cls, id, collection=name, **creator.kwargs)
-                child._protected = True
+        """Create channels according to the ChannelCreator objects."""
+        for name, creator in getmembers(self.__class__):
+            if isinstance(creator, CommonBase.ChannelCreator):
+                for cls, id in creator.pairs:
+                    child = self.add_child(cls, id, collection=name, **creator.kwargs)
+                    child._protected = True
 
     def __setattr__(self, name, value):
         """ Add reserved_prefix in front of special variables."""
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -119,12 +119,8 @@ class CommonBase:
 
     def __init__(self):
         self._special_names = self._setup_special_names()
-        # Add children from ChildDescriptors.
-        for item, value in self.__class__.__dict__.items():
-            if isinstance(value, self.ChannelCreator):
-                for cls, id in value.pairs:
-                    child = self.add_child(cls, id, collection=item, **value.kwargs)
-                    child._protected = True
+        # Add children from ChannelCreators.
+        self._create_channels()
 
     class ChannelCreator:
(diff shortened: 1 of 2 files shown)


pymeasure/instruments/common_base.py
```diff
@@ -193,6 +188,24 @@ class attributes that are a DynamicProperty. Check also for class variables
...
 
+    def _create_channels(self):
+        """Create channels according to the ChannelCreator objects.
+
+        If a subclass overrides a ChannelCreator object with another one, the
+        one of the subclass is used, not that of the superclass.
+        """
+        channel_creators = {}
+        # Collect all ChannelCreators, override in order of subclasses
+        for obj in reversed((self,) + self.__class__.__mro__):
```
bilderbuchi, line 199: I'm nervous that we need to use `__mro__`, to be honest :sweat_smile:
The override should automatically be solved by a lookup -- since the MRO starts with the bottom-most class, if it finds a wanted attribute, it just stops. Why do we need to emulate this, here?

Second question, why are you prepending `self` to the mro list. On the one had, this seems to mix an instantiated object with the `<class>` objects that should be in the `__mro__` list. On the other hand, the class itself should already be the first entry in the MRO??

tests/instruments/test_common_base.py
```diff
@@ -255,7 +255,26 @@ def test_remove_child_clears_attributes(self, parent_without_children):
...
+        function = CommonBase.ChannelCreator(GenericBase, "overridden", prefix=None)
+
+    @pytest.fixture()
+    def parent(self):
+        return self.InstrumentSubclass(ProtocolAdapter())
+
+    def test_inherited_children_are_present(self, parent):
+        assert isinstance(parent.ch_A, GenericBase)
+
+    def test_ChannelCreator_is_replaced(self, parent):
```
bilderbuchi, line 270:
```suggestion
    def test_ChannelCreator_is_replaced_by_channel_collection(self, parent):
```
Maybe? Or add a comment that this is the usual mechanism -- I was initially confused why the changes from `InstrumentSubclass` would _replace_ anything in `.channels`.

pymeasure/instruments/common_base.py
bilderbuchi: Blind code sketch what I mean:
```
            for item, value in self.__dict__.items():
                if isinstance(value, self.ChannelCreator):
                    channel_creators[item] = self.gettattr(value)  # let the MRO find the correct attribute (most derived class first)
```

bmoneke: If you give me a possibility to use automatic lookup, I'll use it.
The instance (`self`) dir does not contain them, and the class dirs do not inherit either.
Therefore, you have to look in the instance `self` (in case someone added it to the instance) , the class and every superclass (answering question 2).

I could use `dir()` and `getattr`, which would raise errors with `setting `.

tests/instruments/test_common_base.py
bmoneke: Good point.

pymeasure/instruments/common_base.py
bmoneke: Neither `Instrument.setting` nor the special values of dynamic properties like to be read, but we could catch these errors with try clauses.

bmoneke: If you find such a possibility, we Don't need the dictionary anymore, but you can directly create the channels. I use the dictionary to search for the most recent definition before I add them to the instrument.

msmttchr: What about using inspect? It looks cleaner to me.
I think we have 3 places in the code base where such lookup is performed:
- parameters (using inspect)
- dynamic properties (using MRO)
- channel creator

I think once the best approach is selected, we could make the approach uniform.

bmoneke: `inspect.getmembers` fails, if you either have a dynamic property or a setting property, because the first one creates protected names (which cannot be read) and the latter cannot return a value either.

I based the code on that of dynamic properties, but I reversed the resolution order in order, that the most recent definition remains (as expected).

We could do it in the right order (from subclass to superclass), if we used `setdefault`, but I deemed direct assign better.

msmttchr: I think, that `inspect` doesn't fails, because we are reading attribute at class level, something like:
`for attr, value in inspect.getmembers(self.__class__):`
In this case, python properties are not evaluated and value returned is a property object.

bmoneke: Ah, getting the members of the class is great, thanks. I'll try it at home.

Shall I change special names as well to use inspect?

bilderbuchi: Make sense to me.

msmttchr: > Shall I change special names as well to use inspect?

Yes, please. Also take into account that special names uses also `self` in the attribute lookup, which means that take into account also instance, but I believe this is a bug and so search in class only is also fixing this.

bmoneke: Change is implemented (together with additional tests, to verify, that dynamic properties continue to work as expected).

tests/instruments/test_common_base.py
```diff
@@ -114,7 +114,14 @@ def fake():
...
     # Keep values unchanged, just derive another instrument, e.g. to add more properties
-    pass
+    fake_ctrl2 = CommonBase.control(
+        "", "%d", "docs",
+        validator=truncated_range,
+        values=(1, 10),
+        dynamic=True,
+    )
+
+    fake_ctrl2.values = (2, 20)
```
msmttchr, line 124:
```suggestion
    fake_ctrl2_values = (2, 20)
```
Shouldn't be like this ?

msmttchr (changes requested):
@bmoneke looks alright to me, except one comment.

tests/instruments/test_common_base.py
bmoneke: Yes...

I just verified, that all tests pass with this change.

msmttchr (approved):
It is OK for me. Thanks for your work (both quantity and quality)

tests/instruments/test_common_base.py
bilderbuchi: Funny that the tests passed before, too... is that something we should investigate?

tests/instruments/test_common_base.py
```diff
@@ -711,3 +719,18 @@ def test_dynamic_property_values_update_in_one_instance_leaves_other_unchanged()
...
+    with pytest.raises(AttributeError):
+        inst.fake_ctrl_validator
+    # Test for new attribute
+    with pytest.raises(AttributeError):
+        inst.fake_ctrl2_validator
+
+
+def test_dynamic_property_values_defined_at_superclass_level():
+    inst = StrictExtendedBase()
+    inst.fake_ctrl2 = 17  # outside original values (1, 10)
```
bilderbuchi, line 736: This test doesn't assert anything?

bmoneke: The assignment would fail, if the values parameter would not have been changed from the original values (1, 10) to (2, 20).
Not failing is passing.
And I failed in making the comment understandable...

tests/instruments/test_common_base.py
bmoneke: I already thought about it. I used "truncated" instead of "strict".
Now I fixed it and expanded the test.

bilderbuchi (approved):
LGTM. :shipit:

PR #593, opened 2022-02-17
Split windows.py into a folder with separate modules

mcdo0486:
Splits windows.py into a folder with separate modules.
This will make it easier to develop new window classes outside of a monolithic module.

PR #762, opened 2022-11-25
Dockwidget save layout

mcdo0486:
Since the DockWindow PR is approved, figured I'd push this one up. This adds a "Save Dock Layout" button to DockWidget. When pressed it saves a json file containing the current dock layout state. When a GUI using DockWindow is launched, it looks for that file. If it exists then it loads that layout. By default the layout file is saved in the directory the python GUI file is run "./" . The naming scheme of the file is "[procedure name]_dock_layout.json"

This allows users to rearrange their dock windows, like popping out docks or maximizing the window,  then click "Save Dock Layout", and have that layout be restored when they run the procedure again.


pymeasure/display/widgets/dock_widget.py
```diff
@@ -84,11 +99,25 @@ def _setup_ui(self):
             self.docks.append(dock)
 
     def _layout(self):
+        hbox = QtWidgets.QHBoxLayout()
+        hbox.addStretch(5)
+        hbox.addWidget(self.save_layout_button, 1)
```
CasperSchippers, line 104: While I (at this moment) also don't really see another option, I do have to admit that I find the single `save` button at the top right a bit uggly / a waste of space (which I think is rather precious when using the DockWidget)

pymeasure/display/widgets/dock_widget.py
```diff
@@ -48,14 +51,16 @@ class DockWidget(TabWidget, QtWidgets.QWidget):
...
     :param parent: Passed on to QtWidgets.QWidget. Default is None
     """
 
     def __init__(self, name, procedure_class, x_axis_labels=None, y_axis_labels=None, linewidth=1,
-                 parent=None):
+                 layout_path='./', parent=None):
         super().__init__(name, parent)
 
         self.procedure_class = procedure_class
+        self.dock_layout_filename = layout_path + procedure_class.__name__ + '_dock_layout.json'
```
CasperSchippers, line 63: Maybe (see my general comment) an option would be to have  `layout_filename=None` optional kwargs, and only use the default `procedure_class.__name__ + '_dock_layout.json'` if layout_filename is not provided:
```suggestion
        self.dock_layout_filename = path.join(layout_path, layout_filename or
                                              (procedure_class.__name__ + '_dock_layout.json'))
```

Though I'm not entirely sure if this would be all too useful.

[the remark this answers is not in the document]
mcdo0486: Implemented in last commit

pymeasure/display/widgets/dock_widget.py
```diff
@@ -68,7 +73,17 @@ def __init__(self, name, procedure_class, x_axis_labels=None, y_axis_labels=None
         self._setup_ui()
         self._layout()
 
+    def save_dock_layout(self):
+        layout = self.dock_area.saveState()
+        with open(self.dock_layout_filename, 'w') as f:
+            f.write(json.dumps(layout))
+            log.info('Saved dock layout to file %s' % self.dock_layout_filename)
```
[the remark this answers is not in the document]
mcdo0486: Implemented in last commit

pymeasure/display/widgets/dock_widget.py
```diff
@@ -84,11 +99,25 @@ def _setup_ui(self):
...
+        vbox.addLayout(hbox)
         vbox.addWidget(self.dock_area)
         self.setLayout(vbox)
 
+        # Load dock layout file if it exists in the directory of the current procedure
+        if path.exists(self.dock_layout_filename):
+            with open(self.dock_layout_filename, 'r') as f:
+                dock_layout = f.read()
+                # make sure number of docks in the file matches num_plots
+                if dock_layout.count('dock') == self.num_plots:
```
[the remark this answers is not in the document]
mcdo0486: Implemented in last commit

pymeasure/display/widgets/dock_widget.py
mcdo0486: Implemented in last commit

pymeasure/display/widgets/dock_widget.py
mcdo0486: I agree and I figured this PR would prompt a discussion where widget buttons should go in general. Should widget buttons be confined to their QWidget or should there be a standard place in the GUI for them?

If we assume widgets with buttons will be implemented in a `ManagedWindowBase` implementation, like DockWindow, then we may be able to put widget buttons in a standard place.

GUI real-estate-wise, there is space around  the `BrowserWidget` besides "Open", or the base ManagedWindowBase QWidget around "Queue" and "Abort".

pymeasure/display/widgets/dock_widget.py
```diff
@@ -84,11 +99,25 @@ def _setup_ui(self):
...
+                    self.dock_area.restoreState(json.loads(dock_layout))
+                    log.info('Loaded dock layout from file %s' % self.dock_layout_filename)
```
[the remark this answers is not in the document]
mcdo0486: Implemented in last commit

pymeasure/display/widgets/dock_widget.py
CasperSchippers: I'm a bit reluctant to place the buttons at another place than the widget itself; I foresee that it creates confusion as to what button does what and relates to which part of the interface (plus it'd need some logic to switch between buttons if you switch tabs). Hence I'd keep buttons (somehow) within the widget itself.

In #771, the actions are put in context menu (i.e. right-click somewhere), so I was thinking maybe that would be an idea, to put the save action (and maybe then also a load?) in a context menu; either in the widget itself, but seeing that that is a pyqtgraph widget, I'm not sure if that is the easiest option, so an alternative would be a context menu of the tab-bar. What do you think of this idea?

CasperSchippers:
Also make sure to put a note about this feature in the documentation/tutorial

pymeasure/display/widgets/dock_widget.py
mcdo0486: I tried a couple things with the context menu. If we setup a context menu for the overall `DockWidget`, it will override the `PlotWidget.plot_frame` context menu aka `pyqtgraph.PlotWidget` context menu. Each of those context menus are specific to the plot, for instance checking the "Plot Options -> Transforms -> Log X" option. 

I pushed a change to add "Save Dock Layout" to the context menu of each `pyqtgraph.PlotWidget`, and that works, you can pull down the branch again to test that out.

However the context menu is only accessible in these GUI regions, since `DockWidget` creates a number of `PlotWidgets`:
image

I'll see if there is a way for the widget (red regions) to have a context menu without overriding the `pyqtgraph.PlotWidget` context menu.

mcdo0486: Well, that was easier than I thought. There is a helpful Qt function `widget.childAt(position)` which can be used to only create a context menu at positions in `DockWidget` that aren't a `pyqtgraph.PlotWidget`.

Try out the latest commit and let me know if that is more intuitive for users. I'll work on documentation.

mcdo0486: I'm still not sure if the context menu is the best place for this useful functionality. It should be more obvious then a hidden context menu, but I'm not sure of other options. The tutorial and documentation can demonstrate the context menu.

The context menu will work to set a precedent for not allowing button mania in pymeasure UI design.

CasperSchippers: Hmm, I played around with it, and it seems to work nicely.
I do understand your concern though (re. the fact that it is less obvious than a button), I'm incline to lean towards the context menu nevertheless (over the button), especially as I would expect it is not a button you'll use every time (or rather not even very often) you use the software, but rather use it one of the first few times you setup your experiment and leave it (more or less) alone for the remainder of the use. Again, I can be wrong, but even for many uses, I would not be annoyed if the action is located in the context menu.

CasperSchippers: Regarding the "expensive real estate" around the widgets, I made a PR (#822) to move the coordinates label, to save some space around the plots

mcdo0486: Yea the context menu is growing on me. I'll check out that PR

mcdo0486:
> Also make sure to put a note about this feature in the documentation/tutorial

I pushed a commit yesterday that adds a section to the tutorial regarding the context menu usage. Because it isn't a button, I didn't have to redo all the application window screenshots.

CasperSchippers:
Could you check on which qt, pyqt versions you are when working on this? I've been a bit sloppy with checking things on the correct versions, but I just noticed I get a bunch of errors with using the dockswidget when I adhere to the versions that are mentioned in the pymeasure requirements file (pyqt=5.12.3, qt=5.12.9), but not when using newer versions (pyqt=5.15.7, qt=5.15.4).

mcdo0486:
Yea, I can replicate those errors, and from a quick check it looks like PyQt5 version 5.14.2 is the lowest version that doesn't show any errors. I have been developing with PyQt5==5.15.7 through pip.

The problem is that, strictly speaking, pymeasure doesn't require a version of PyQt5 or pyqt (because pyqtgraph doesn't require them on install). As far as I can tell, the [requirements/pymeasure.yml](https://github.com/pymeasure/pymeasure/blob/master/requirements/pymeasure.yml) with pinned library versions is only used for Github actions.

I tested this in a fresh virtual environment in Anaconda (`conda install -c conda-forge pymeasure`) and with pip (`pip install pymeasure`). After installing pymeasure, you can try to run a file and you'll get this exception:
`Exception: PyQtGraph requires one of PyQt5, PyQt6, PySide2 or PySide6; none of these packages could be imported.` It is at that point you'll need to manually install `pip install pyqt5` or `conda install pyqt`, which will pull down the most recent version of PyQt5.

The actual pymeasure requirements for Anaconda installations are in the [conda-forge feedstock repo](https://github.com/conda-forge/pymeasure-feedstock), which has the [conda-forge recipe](https://github.com/conda-forge/pymeasure-feedstock/blob/main/recipe/meta.yaml) for installing pymeasure. That recipe has all requirements setup with `>=`. For pip, the [setup.cfg](https://github.com/pymeasure/pymeasure/blob/master/setup.cfg) has all versions with `>=`. So we aren't pinning a specific version of `PyQT5` or `pyqt` anywhere.

That said, we probably need to update the `pyqtgraph` requirements.

From the [pyqtgraph feedstock recipe](https://github.com/conda-forge/pyqtgraph-feedstock/blob/main/recipe/meta.yaml) it has a `run_constrained` definition of `- pyqt >=5.15`, so if pyqt is installed then it has to meet that requirement, but it doesn't install pyqt itself.

From the [pyqtgraph setup.py](https://github.com/pyqtgraph/pyqtgraph/blob/master/setup.py), the only requirement is numpy. 

What we probably need to do is increase the minimum version of pyqtgraph.
At the moment we are requiring `pyqtgraph >=0.9.10` and that version doesn't include support for PyQt5, only PyQt4.
`pyqtgraph == 0.11.1` is the lowest version that runs without errors, and includes support for DockWidget.

Although we should probably pin to the latest release of pyqtgraph (currently 0.13.1), as pymeasure relies on it so closely. If we increase the `pyqtgraph` requirement, then that should fix the issue for Anaconda installations, since `pyqtgraph >= 0.13.1` will update `pyqt` if it is already installed. I don't know of a similar quick fix for pip installs.

bilderbuchi:
It's fine to bump to `pyqtgraph >= 0.11.1` in our `setup.cfg`. Please add an entry to the changelog that the minimum version increased (I'll have to update the conda feedstock recipe with the next release, this way I won't forget).

It's fine to bump the `pyqt` and `qt` (and `pyqtgraph`) entries in `requirements.yml` to newer, working versions. 

The gist of it is: 
For user installation (i.e. setup.cfg) you prescribe versions as loosely as possible, to ease the pip/conda's job to find a version compatible with all packages in an environment. Generally, this means prescribing minimum versions in case you require some feature or fix.
For CI, you want your dependency versions to be _fixed_, so that you don't end up generating false alarms when dependencies' versions are updated from one run to the next in a way that breaks your build (for an unrelated reason). This is the job of requirements.yml, and we can update that as needed (and fix any potential fallout that occurs _when_ we update).

CasperSchippers:
Ah, that makes sense; thanks for the inquiry and explanation.

Then let's bump the versions of pyqt, qt, and pyqtgraph. I would propose the following:
- As @bilderbuchi suggests bump `pyqtgraph >= 0.11.1` in `setup.cfg` (with an entry to the changelog)
  Especially if everything works on this version, we should keep this as low as possible according to remark of @bilderbuchi
- in `requirements.yml` bump:
  - `pyqt=5.15.7`
  - `qt=5.15.4`
  - `pyqtgraph=0.12.4  # can be bumped to v6 when support for python 3.7 is dropped`
     `pyqtgraph=0.13.1` does not support python 3.7, so I expect using 13.1 would be problematic for the CI, hence the note to bump when support for 3.7 is dropped
  - I'd also bump `pytest-qt=4.2.0` as we're bumping qt-related packages

On an unrelated sidenote:
Maybe also add the DockWidget to the init in the widgets dir.

CasperSchippers:
How low do we need to keep pyqtgraph in `setup.cfg`? I found one thing that doesn't work with `pyqtgraph < 0.12`, namely the colormap in the ImageWidget. So we need either `pyqtgraph >= 0.12` or make the colormap in the imagewidget work with 0.11.1.

CasperSchippers:
> I'm wondering: Can we add a test to the testsuite that catches the breakage/error that Casper is seeing? That would be good, so we see CI failing when we actually have a problem. :-P

I tried making a test that shows shows the error for this case, this was the simplest I could come up with:
```python
from pymeasure.experiment import Procedure
from pymeasure.display.widgets.dock_widget import DockWidget


def test_dock_widget_init(qtbot, ):
    class TestProcedure(Procedure):
        DATA_COLUMNS = ["x", "y"]

    wdg = DockWidget('Test Widget', TestProcedure, ["x"], ["y"])

    qtbot.addWidget(wdg)
    with qtbot.waitExposed(wdg, 1000):
        wdg.show()
```

Would it be useful to add such a test to this/other widgets; it does test if the widget can be initialized and shown, but I'm not sure how many issues this will capture.

bilderbuchi:
> How low do we need to keep pyqtgraph in `setup.cfg`? I found one thing that doesn't work with `pyqtgraph < 0.12`, namely the colormap in the ImageWidget. So we need either `pyqtgraph >= 0.12` or make the colormap in the imagewidget work with 0.11.1.

The lowest that works for us, so >=0.12 will be okay.

bilderbuchi:
I'd say don't add GUI tests that are not useful. Tests should fail without human intervention (so if you have to look at the result to recognise a failure, that's not useful). We have a virtual frame buffer on CI, so we can trigger display of GUI elements even on CI.

mcdo0486:
> How low do we need to keep pyqtgraph in `setup.cfg`? I found one thing that doesn't work with `pyqtgraph < 0.12`, namely the colormap in the ImageWidget. So we need either `pyqtgraph >= 0.12` or make the colormap in the imagewidget work with 0.11.1.

It think  `pyqtgraph >= 0.12` is a good target both because it fixes the colormap issue, and `0.12.0` is when the pyqtgraph conda forge feedstock started to include a `run_constrained` definition for pyqt with [`pyqt >= 5.12` at the time.](https://github.com/conda-forge/pyqtgraph-feedstock/blob/298162a8fa3f2bd3c10ceafd29e402992ee8c73a/recipe/meta.yaml)

mcdo0486:
> On an unrelated sidenote: Maybe also add the DockWidget to the init in the widgets dir.

I just added a commit that bumped the versions as you noted.

For `DockWidget` being added to `pymeasure/display/widgets/__init__.py` , or for that matter adding `ManagedDockWindow` to `pymeasure/display/windows/__init__.py` , I don't think classes should be added to `__init__.py` unless strictly necessary.  This is similar to [the discussion](https://github.com/pymeasure/pymeasure/discussions/801) related to moving away from adding all instruments to `pymeasure/instruments/__init__.py`.

Adding classes to `__init__.py` will mean code will be loaded into the program even if it won't be used. As custom graphic widgets and windows continue to be added, the memory footprint will continue to increase. So, if someone wants to use it they can import it through the full module path. 

The reason why there are classes in both  `pymeasure/display/widgets/__init__.py` and `pymeasure/display/windows/__init__.py`  is because those classes were originally in `widgets.py` or `windows.py`, so I included them to preserve backwards compatibility for existing PyMeasure scripts. That was in pull requests https://github.com/pymeasure/pymeasure/pull/593 and https://github.com/pymeasure/pymeasure/pull/594

setup.cfg
```diff
@@ -32,7 +32,7 @@ install_requires =
     pint
     pyvisa >= 1.8
     pyserial >= 2.7
-    pyqtgraph >= 0.9.10
+    pyqtgraph >= 0.12
```
bilderbuchi, line 35: Please add a changelog note as I requested, so we don't overlook this.

mcdo0486: Ok, added changes to the CHANGES.rst. I added some other GUI changes I know have been merged recently.

CasperSchippers (approved):
Sorry for the slow response. I think it looks good to me.

pymeasure/display/widgets/dock_widget.py
CasperSchippers: Let's leave it as it is then.

PR #765, opened 2022-11-28
Fix query_delay usage in VISAAdapter, closes #760

### LINKED ISSUE #760: Unable to reach Prologix USB GPIB adapter
Dear pymeasure dev team,

I just wanted to raise you an issue concerning the Prologix adapter on version 0.11.0
I can't address it anymore...

I saw it now depends on Visa adapter class, but I cannot find a way to communicate with.
Does anyone can provide an example please ?

Best,
Benjamin.

bmoneke:
Fixes #760 

- Adjusted test to cover this parameter (enabled or not enabled)
- `Adapter` calls `super` (docstring adjusted accordingly).
- Right usage of `query_delay` (store it always as `self.query_delay` in order to "nest" adapters. Hand it via kwargs to PyVISA).

### DIFF
--- a/pymeasure/adapters/adapter.py
+++ b/pymeasure/adapters/adapter.py
@@ -45,10 +45,11 @@ class Adapter:
             Implement it in the instrument's `read` method instead.
 
     :param log: Parent logger of the 'Adapter' logger.
-    :param kwargs: all other keyword arguments are ignored.
+    :param kwargs: Keyword arguments just to be cooperative.
     """
 
     def __init__(self, preprocess_reply=None, log=None, **kwargs):
+        super().__init__(**kwargs)
         self.preprocess_reply = preprocess_reply
         self.connection = None
         if log is None:
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -85,9 +85,11 @@ class VISAAdapter(Adapter):
 
     def __init__(self, resource_name, visa_library='', preprocess_reply=None,
                  query_delay=0, log=None, **kwargs):
-        super().__init__(preprocess_reply=preprocess_reply, log=log, query_delay=query_delay)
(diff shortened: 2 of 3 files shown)


msmttchr (approved):
@bmoneke, ok to merge for me.

bilderbuchi:
@bmoneke @msmttchr thanks for the quick work! Given this fixes a regression, should we backport it onto the `v0.11.0` tag and cut a `v0.11.1`bugfix release (maybe after waiting a while if more regressions appear)?

bmoneke:
Good idea to backport it in order to publish a bugfix release.

I see pros and cons for earlier or later.
In the end, what is the effort to create that bugfix version? If it is not that much, we can do it immediately.

I do not expect more regressions (and if, we can do another one), as we worked diligently and have quite some tests in place.

This regression was a very specific case (calling an adapter from an adapter with `adapter.gpib()`) with some code complications.

bilderbuchi:
> In the end, what is the effort to create that bugfix version? If it is not that much, we can do it immediately.

Effectively, running through all of https://github.com/pymeasure/pymeasure/blob/master/RELEASE.md -- I'd estimate an hour or so, including waiting for builds etc.

> I do not expect more regressions (and if, we can do another one), as we worked diligently and have quite some tests in place.

Still, this one slipped the net. I'm not being hard on us, it just shows how tricky refactoring is without good test coverage. :-)

> This regression was a very specific case (calling an adapter from an adapter with `adapter.gpib()`) with some code complications.

I understand that the user has a workaround in place? I'd say let's target the weekend for putting out a bugfix release. In the meantime, I created https://github.com/pymeasure/pymeasure/tree/hotfix-0.11.0 - backport PR against this is welcome. If you find the time, please start a changelog entry, too.

bmoneke:
I opened a PR with the (cherry-picked) hotfix and a changelog entry.

PR #769, opened 2022-11-29
HP8657B - GHz fix

LongnoseRob:
This PR fixes a small error in the format string for `hp8657b.frequency` made it impossible to set frequencies higger 999.999999 MHz.
Tested for fucntionality with real hardware.


bmoneke:
Due to the minimum frequency of 100 kHz, I estimate it is not necessary to have 100 mHz resolution.

Thanks for the fix and tests.

Issue #752, opened 2022-11-18
Generator for protocol tests


bmoneke:
During the Instrument rework, I wrote an initial version of a protocol test generator, see [my branch](https://github.com/pymeasure/pymeasure/compare/master...bmoneke:pymeasure:Test_Generator).

If you communicate with an instrument through this generator, the generator creates a protocol test file according to that communication. Use case is, that you have a working implementation and want to write tests, such that later code changes still produce the same communication.
~~~
# Example usage testing properties of an instrument
g = Generator("test_tc038.py")
g.instantiate(TC038, "COM5", 'hcp')
g.test_property("information")
g.test_property("monitored_value")
g.test_property_setter("setpoint", 20)
g.test_property("setpoint")
~~~

I did not propose it earlier, due to more pressing issues and shortage of labour.

Now I want to bring it back to attention, in order to see, where we can put it in the task queue. If contributors see it as helpful for writing tests for their instruments, I'll move it up on the task list (which will delay other works and PR reviews).

LongnoseRob:
Looks interesting,
let me find some time to test it with actual hardware

bilderbuchi:
I think this is a useful feature to easily nail down communication into protocol tests. 
Timing-wise, I'd prefer if we first worked down the PR queue a bit before taking up the next (big?) feature PR. I think 50+ open PRs is frustrating for all involved, and would love to get that down as much as feasible (1 page?). This should also improve our future response time, as right now we have to divide our review attention between too many places.

bmoneke:
Ok, I improved the example with some comments. Whoever wants to use the generator, may do so with my code already existing.

bmoneke:
Thanks for testing it.

The generator was part of an already large PR and therefore postponed. As such it is not yet complete

> for 'test_property_setter` it might be nice to have some logic to use iterables as argument

I already planned to collect all tests of a property together and create a `parametrize` test.
Thanks for suggesting. As a setter might accept an iterable anyway, I can introduce `test_property_setter_values` or such which loops through an iterable (or you can loop through it). However, as you should verify, that the code does what it is supposed to do (e.g. test that setting the voltage to 12.3 V, really does that), it is counterproductive to just test many values without verifying. It might be useful for getting additional values, I concede.

> the test-file is opened but not automatically written if following your example. helped with a g.__del__() at the end

The file is written (at least according to the code), but it is not closed until closure of the generator. Depending on the OS you might not see the changes, because the file is blocked.

[...]

PR #777, opened 2022-12-06
HP3478A - Add calibration data related functions.

tomverbeure:
The HP3478A stores its calibration data in SRAM, using a Lithium battery to retain the contents. To avoid calibration data loss, it's advised to back up this data.

This PR adds the following functionality:

* read the calibration data into an array
* write calibration data back to the device
* check the integrity of the calibration data


pymeasure/instruments/hp/hp3478A.py
```diff
@@ -463,3 +463,88 @@ def trigger(self):
...
+
+        The CAL ENABLE switch at the front of the instrument must be set to ON,
+        otherwise an exception is raised.
+
+        IMPORTANT: changing the calibration data results in permanent loss of
+        the previous data. Use with care!
+
+        """
+
+        if not(self.calibration_enabled):
```
bmoneke, line 504: I learnt it to write `not` as a keyword and not as a function:
```suggestion
        if not self.calibration_enabled:
```

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -463,3 +463,88 @@ def trigger(self):
...
+
+        Expects an array of 256 values between 0 and 15.
+
+        The CAL ENABLE switch at the front of the instrument must be set to ON,
+        otherwise an exception is raised.
+
+        IMPORTANT: changing the calibration data results in permanent loss of
+        the previous data. Use with care!
+
+        """
```
bmoneke, line 502: This docstring works well as a method docstring. For a property, however, only the getter docstring is shown in the documentation such that you have to add this text to the getter docstring (obviously combining them in a meaningful way).
In the end, you want to have a single docstring describing the property as getter **and** setter.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -463,3 +463,88 @@ def trigger(self):
...
+        cal_data = []
+        for addr in range(0, 256):
+            # To fetch one nibble: 'W<address>', where address is a raw 8-bit number.
+            cmd = bytes([ord('W'), addr])
+            self.write_bytes(cmd)
+            rvalue = self.read_bytes(1)[0]
+            if rvalue < 64 or rvalue >= 80:
+                raise Exception("calibration nibble out of range")
+            cal_data.append(rvalue-64)
```
bmoneke, line 486: I first got confused, why do you check the range (64, 80) if values are in the range (0,15). A comment could help.
```suggestion
            # rvalue is a hexadecimal ASCII character.
            if rvalue < 64 or rvalue >= 80:
                raise Exception("calibration nibble out of range")
            cal_data.append(rvalue-64)
```

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -463,3 +463,88 @@ def trigger(self):
...
+
+        Returns True when the checksum of all used calibration entries is
+        correct.
+
+        """
+        for entry_nr in range(0, 19):
+            if entry_nr in [5, 16, 18]:
+                continue
+            if not(self.verify_calibration_entry(cal_data, entry_nr)):
+                return False
```
bmoneke, line 549: I'm unsure, whether raising an exception would be better than just returning false. I guess it depends on the use case.

bmoneke (changes requested):
Thanks for your contribution. You might add yourself to `AUTHORS.txt`.

I kindly ask you, to add unit tests for the verification and protocol tests for the data transfer, see [documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#protocol-tests). That ensures that it works as expected. If you have a device at hand, you might use #752 to get a copy of calibration data for the tests.

pymeasure/instruments/hp/hp3478A.py
tomverbeure: An exception would be IMO fine when you add this function as part of a non-explicit sanity check (like some of the other exception that I throw in the write data setter). 

But when used stand-alone the user wants to know the status, so a boolean makes more sense IMO?

bmoneke: That's the reason I was unsure. As you describe it, a boolean might be better.

LongnoseRob:
> The HP3478A stores its calibration data in SRAM, using a Lithium battery to retain the contents. To avoid calibration data loss, it's advised to back up this data.
> 
> This PR adds the following functionality:
> 
>     * read the calibration data into an array
> 
>     * write calibration data back to the device
> 
>     * check the integrity of the calibration data

Thanks for adding the calibration data handling methods, I have a HP3478A avialable for tests and verifications if required.
Will probably test it over the next weekend.

tomverbeure:
> Thanks for adding the calibration data handling methods, I have a HP3478A avialable for tests and verifications if required. Will probably test it over the next weekend.

I did quite a bit of testing with [this test script](https://github.com/tomverbeure/hp3478a_calibration/blob/pymeasure/pymeasure_test.py), as part of writing [a blog post](https://tomverbeure.github.io/2022/12/02/HP3478A-Multimeter-Calibration-Data-Backup-and-Battery-Replacement.html)  (which will be updated to mention `pymeasure` once the PR has been merged.)

But you're more than welcome to give it a try yourself!

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -463,3 +463,88 @@ def trigger(self):
...
+        IMPORTANT: changing the calibration data results in permanent loss of
+        the previous data. Use with care!
+
+        """
+
+        if not(self.calibration_enabled):
+            raise Exception("CAL ENABLE switch not set to ON")
+        if len(cal_data) != 256:
+            raise Exception("cal_data must contain 256 values")
+
```
bilderbuchi, line 508: Should you use the checksum verification from below on `cal_data` before starting the write?

bilderbuchi:
By the way, if you don't want to be annoyed by the "pesky" linter, have your editor use the `flake8` (or `darker`, if you must) configuration we have in the repo, or if that's not possible, just run `flake8 .` before committing. :-)

pymeasure/instruments/hp/hp3478A.py
bmoneke: I also thought about that.
Its another check preventing erroneous change and thus loss of data.

tomverbeure: I'll make the check enabled by default, with a function parameter to override: I definitely want to be able to write corrupt data to check that the write function is really working.

bmoneke: Can't you test the write function with a working calibration data set? Maybe using two different sets to verify, that it changes?

tomverbeure:
> By the way, if you don't want to be annoyed by the "pesky" linter, have your editor use the `flake8` (or `darker`, if you must) configuration we have in the repo, or if that's not possible, just run `flake8 .` before committing. :-)

I'm not annoyed about linting process, but about some linting rules.

For example, the fact that it complains about the first `True`:

```python
        assert instr.calibration_enabled == True
        assert instr.calibration_enabled == False
```

Anyway, it's a distraction not worth going further into.

LongnoseRob:
@tomverbeure just gave this code a test-run with my HP3478A, reading and writing the cal-data works fine.
Thanks for the addition, now I can start looking into replacing the battery..

tomverbeure:
> 

> @tomverbeure just gave this code a test-run with my HP3478A, reading and writing the cal-data works fine. Thanks for the addition, now I can start looking into replacing the battery..

Good to hear!

@bmoneke Is there anything else you'd like to see changed? I don't have a way to check the generated documentation, so maybe there could be some issue left there.

Tom

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -463,3 +463,109 @@ def trigger(self):
...
+        The cal_data parameter format is the same as the 'calibration_data' property.
+
+        Verification of the cal_data array can be bypassed by setting
+        'verify_calibration_data' to False.
+
+        """
+        if verify_calibration_data:
+            if not self.verify_calibration_data(cal_data):
+                raise Exception("cal_data verification fail.")
```
bmoneke, line 527: You can (but don't have to) combine them:
```suggestion
        if verify_calibration_data and not self.verify_calibration_data(cal_data):
            raise Exception("cal_data verification fail.")
```
Only if the first part (`verify_calibration_data`) is True, the second call is done.

tests/instruments/hp/test_hp3478a.py
```diff
@@ -0,0 +1,143 @@
...
+        1, 12, 2, 1, 15, 10, 10, 0, 0, 0, 0, 4, 2, 3, 0, 0,
+        3, 12, 14, 7, 0, 0, 0, 0, 0, 4, 3, 0, 1, 12, 3, 14,
+        8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 15, 9, 9,
+        8, 6, 0, 9, 3, 14, 3, 1, 1, 12, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 15, 15, 0, 0, 0, 0, 0, 0, 0, 0
+    ]
+
+
+def convert_cal_data_to_cal_read_xfers(cal_data):
+    # Convert cal_data into 256 single byte 'W' read operations.
```
bmoneke, line 49: What are "xfers"? Transfer?

tests/instruments/hp/test_hp3478a.py
```diff
@@ -0,0 +1,143 @@
...
+                (b"B", b'\x00\x20\x00\x00\x00'),        # cal_enable bit is set
+                (b"B", b'\x00\x00\x00\x00\x00')
+            ],
+    ) as instr:
+        assert instr.calibration_enabled
+        assert not instr.calibration_enabled
+
+
+def test_calibration_data_getter():
+    cal_read_xfers = convert_cal_data_to_cal_read_xfers(VALID_CAL_DATA)
```
bmoneke, line 83: You use this line several times, why don't you make it a global variable (it is static) or a fixture?

tests/instruments/hp/test_hp3478a.py
```diff
@@ -0,0 +1,143 @@
...
+
+            # setter pass
+            [
+                (b"B", b'\x00\x20\x00\x00\x00'),
+            ] + valid_cal_write_xfers +
+
+            # setter fail due to invalid data
+            [
+                (b"B", b'\x00\x20\x00\x00\x00'),
+            ]
```
bmoneke, line 112: I'd split it in three different tests, that way you know why the setter failed and not only, that it failed. If the first part fails, you do not know, whether the other parts work as expected.
In fact, pytest recommends to do a single assert in a test.

If you want to test several outcomes of some action, you can do the action in a fixture and have several tests, each testing one outcome. You can use classes (starting with "Test") to group tests together.

tests/instruments/hp/test_hp3478a.py
```diff
@@ -0,0 +1,143 @@
...
+
+    ) as instr:
+        # There will be no data writes when cal_enable is false.
+        try:
+            instr.calibration_data = VALID_CAL_DATA
+        except Exception:
+            pass
```
bmoneke, line 119: Do you expect an exception to be raised? Then you can use `pytest.raises`:
```python
        with pytest.raises(SomeFunnyException, match="error message which I want to verify"):
            instr.calibration_data = VALID_CAL_DATA
```
That way the test passes if and only if `SomeFunnyException` was raised with a message containing above string. If the test fails, you know that raising the error did not happen as expected.

tests/instruments/hp/test_hp3478a.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Writing correct data
+        instr.calibration_data = VALID_CAL_DATA
+
+        # Assigning invalid data results in an Exception without data writes
+        try:
+            valid_cal_data_corrupt = VALID_CAL_DATA.copy()
+            valid_cal_data_corrupt[1] = 1
```
bmoneke, line 127: You use it below as well. Why don't you make it as a global variable?

bmoneke (changes requested):
I commented some tests. Thanks for including them. If you have questions regarding pytest, don't hesitate to ask.
For the one suggestion you might have to import pytest.

pymeasure/instruments/hp/hp3478A.py
tomverbeure: Fixed.

tests/instruments/hp/test_hp3478a.py
tomverbeure: Yes.

tests/instruments/hp/test_hp3478a.py
tomverbeure: VALID_CAL_DATA is the root calibration data that belongs in the instrument SRAM.

The 2 convert_* functions are functions that transform it into GPIB transfers. I think it makes sense to have those factored as individual functions. It's something that's useful to have if others want to reuse it at some point for their own code.

After that:
*  test_calibration_data_getter creates a cal_read_xfers array, 
*  test_calibration_data_setter_pass creates a valid_cal_write_xfers array from VALID_CAL_DATA
*  test_write_calibration_data creates invalid_cal_write_xfers from a corrupted VALID_CAL_DATA

So cases of anything being used multiple times.

tests/instruments/hp/test_hp3478a.py
tomverbeure: Done.

tests/instruments/hp/test_hp3478a.py
tomverbeure: Done.

tests/instruments/hp/test_hp3478a.py
tomverbeure: See above. Only used once.

pymeasure/instruments/hp/hp3478A.py
```diff
@@ -522,9 +522,8 @@ def write_calibration_data(self, cal_data, verify_calibration_data=True):
         'verify_calibration_data' to False.
 
         """
-        if verify_calibration_data:
-            if not self.verify_calibration_data(cal_data):
-                raise Exception("cal_data verification fail.")
+        if verify_calibration_data and not self.verify_calibration_data(cal_data):
+            raise Exception("cal_data verification fail.")
```
bmoneke, line 526: As far as I know, you should not use 'Exception', but some subclass. Here I'd use a Value Error (use it in the tests as well).

tomverbeure: Done.

bmoneke:
@tomverbeure may I merge this PR?

tomverbeure:
Of course! :-)

bmoneke:
Thanks for your contribution!

Issue #742, opened 2022-11-14
How to test adapter.connection calls


bmoneke:
Some instruments call the adapter's `connection`property, like #697 which makes ProtocolAdapter fail. Instruments should not be adjusted to make the ProtocolAdapter work.

The problem is, that you are not supposed to use normally the adapter's connection directly, because the `connection` depends on the specific adapter type. If an instrument calls the connection, the ProtocolAdapter fails. Instruments should not be adjusted to make the ProtocolAdapter work.

Two solutions:

- A generic mock connection seems good, because that would care for many use cases. Otherwise we had to add several methods (clear, close, timeout property, get_stb...) individually.
- Or we could add timeout as a property to the `Adapter`, which will call its connection depending on the Adapter type (ProtocolAdapter stores it as a variable). Similarly the "read_stb" and `clear` etc.

bilderbuchi:
Option 1 sounds preferable/easier to get right.

bmoneke:
Also #778 requires the timeout as a property. I guess the best path is a generic mock, but also a timeout property in the `Adapter`.

Sometimes, there are solutions to use VISA methods or general methods (e.g. setting timeout temporarily to 0). The latter is open for more adapters, while the former requires a VISAAdapter.
Therefore, I prefer the more general approach.

bilderbuchi:
> Another advantage of a Adapter.timeout property: That way we can ensure the right timeout setting regardless of the unit of the adapter (see difference VISA and pyserial). If an instrument sets the timeout to 500, it will be half a second, for the VISAAdapter and for the SerialAdapter, because the latter's property transforms it to 0.5 internally.

This is a good application, harmonising that would be useful indeed. Let's use the pyVISA way as a baseline. One thing that concerns me is that this will now make our configuration API more complex, as we need to special-case the `timeout` parameter, and also probably intercept it if people supply it already (but assume the "other" convention), and because it does not land directly in the adapter.connection as we have been saying until now. :-/

Should we use a different name? `timeout_ms` would make the unit clear, allow us to cleanly intercept/deprecate use of `timeout`, and break the correspondence with the respective connection kwarg (pyVISA, serial), so less potential for confusion.

bmoneke:
I did not think about the parameters of the adapter, but how the instrument interacts with the present adapter.
However I see the confusion in starting a serial adapter with a timeout of 0.5 and later setting the timeout to 500 for the same effect.

Maybe we just change the serial adapters interpretation of timeout?

PR #778, opened 2022-12-06
new Instrument: AJA International DC sputtering power supply

dkriegner:
This implements a new Instrument from the vendor AJA International. Its manual can be found here:
https://www.nanophys.kth.se/nanolab/aja/manuals-pdf/DCXS750-Specifications-V108a.pdf

I hope that compared to the other Instruments I submitted, this is rather straight forward. :)
I see only two hiccups:

- The device uses fixed length replies to all commands and does not use any message terminator. I overload `Instrument.ask` and `Instrument.values` to make it work. The Instrument.read method is rather useless since it always times out since no terminator can be found.
- Upon power cycling the device sends some characters which one should disregard upon initialization. Due to the fixed length nature of the replies one can not simply ignore these characters. Since `adapter.flush_read_buffer()` does not work I implemented a workaround where I need to catch two exceptions (one for the test suite to work and the other to catch the read timeout which always occurs).

If you see better ways to handle these issues, please let me know.


bmoneke:
>     * The device uses fixed length replies to all commands and does not use any message terminator. I overload `Instrument.ask` and `Instrument.values` to make it work. The Instrument.read method is rather useless since it always times out since no terminator can be found.

Please override `self.read` and call there `self.read_bytes`, that is the best way to implement it. You can use `read_bytes(-1)` to read the whole buffer, if applicable.

>     * Upon power cycling the device sends some characters which one should disregard upon initialization. Due to the fixed length nature of the replies one can not simply ignore these characters. Since `adapter.flush_read_buffer()` does not work I implemented a workaround where I need to catch two exceptions (one for the test suite to work and the other to catch the read timeout which always occurs).

You could do `self.adapter.connection.flush_read_buffer()`, as the `adapter.connection` is the VISA object. Tests will fail for now as well, until we add a mock connection to the ProtocolAdapter, which we talk about in #742 .

I'd say, leave it as is, until you are done with the rest or #742 is done and the way to go is manifest.

dkriegner:
thanks for the recommendations. I tested this, but there are some problems.

I overload `Instrument.read` now instead of `ask/values`. It makes the new device code easier but I need to change something in the core. The problem is that `read_bytes(-1)` does not work for me: I get a pyvisa.errors.VisaIOError (Timeout) whenever I call `read_bytes(-1)` independent if there is something waiting in the read buffer or not. This is clearly not what I expected when reading the documentation. Not sure if this can be a problem on my end.
But even if `read_bytes(-1)` would work I would need make query_delay long enough so that the full message is ready to be read. I think this is suboptimal to depend on the timing. It would be better to wait for the right amount of characters.
So to keep the reply_length argument without overloading `ask/values` I need to minorly modify these methods in Instrument.

regarding issue PR 742: Yes this will be good once fixed. There is an additional problem for me since `flush_read_buffer` does not work for all connections pyvisa offers. My device is connected via a TCPIP socket and for this type of connection I get a `NotImplementedError` (it works for a real serial connection though). I changed the code a bit to reflect this, but this anyway needs to wait for 742 to be cleaned up. One could think about moving the fallback implementation of the buffer flushing in the core code.

bmoneke:
Sorry for the misunderstanding. You can use `read_bytes(-1)` similarly to the `read` you use right now. As VISA does not know when to stop (as there is no termination char), you have to wait for a timeout.

It's fine with me, if you add `**kwargs` to `comon_base.CommonBase.values` arguments (starting on line 293), which hands them to `ask`:
~~~python
    def values(self, command, separator=',', cast=float, preprocess_reply=None, **kwargs):
        """Write a command to the instrument and return a list of formatted
        values from the result.

        :param separator: A separator character to split the string into a list
        :param cast: A type to cast the result
        :param preprocess_reply: optional callable used to preprocess values
            received from the instrument. The callable returns the processed
            string.
        :param \\**kwargs: Keyword arguments for :meth:`ask`.
        :returns: A list of the desired type, or strings where the casting fails
        """
        results = str(self.ask(command, **kwargs)).strip()
~~~

Afterwards you can override ask and give the byte_length to read.

bmoneke:
> I still don't understand how one could employ read_bytes(-1): Because if a timeout and the resulting exception occurs I am not sure how I can extract the buffer which was read until the timeout? For me this seems no option here anyway.
> 
> edit: I guess I understand now you are just saying that I do not limit myself since read_bytes(-1) is equivalent to the normal read (which both here seem useless).

Exactly: You can use `read_bytes(-1)` for flushing the buffer, but not for normal readout.

pymeasure/instruments/validators.py
```diff
@@ -140,6 +140,20 @@ def truncated_discrete_set(value, values):
     return values[-1]
 
 
+def truncated_string(value, maxlength):
+    """ Provides a validator function that truncates a string to a maximum length.
+    It returns the original value if its length is smaller or equal to the maximum,
+    otherwise only the characters up to the maximum length will be returned.
+
+    :param value: a value to test
+    :param maxlength: maximum length of the string
```
bilderbuchi, line 149:
```suggestion
    :param maxlength: maximum length of the string ("values" arg of the property factories)
```

pymeasure/instruments/validators.py
```diff
@@ -140,6 +140,20 @@ def truncated_discrete_set(value, values):
     return values[-1]
 
 
+def truncated_string(value, maxlength):
```
bilderbuchi, line 143: Is that really such a useful validator to put into core (it's only used once, and there maybe not even necessary)? I'm not sure, I'd probably make a lambda for that one occasion.
If yes, this should also get mentioned in the validator documentation.

bilderbuchi:
@bmoneke: We are accumulating instruments with "weird" behaviour quirks. Should we start a FAQ-style directory in the documentation where we record weird/special instruments, so that contributors can check their list for implementation inspiration?
I imagine something like
>  * TC038, x and y implement binary frame-based communication
* aja.DCXS has replies that are all the same lengths
* aja.DCXS has some garbage in the buffer after startup, that needs to be discarded
* XYZ (I forget which) VNA can dynamically configure itself depending on which hardware modules are present
* In the MKS937B, not all the channels have all capabilities
* ...

What do you think?

pymeasure/instruments/aja/dcxs.py
```diff
@@ -0,0 +1,214 @@
...
+        validator=strict_range,
+        values=range(60),
+    )
+
+    material = Instrument.control(
+        "n", "I%08s", """material name of the sputter target""",
+        cast=str,
+        reply_length=8,
+        validator=truncated_string,
+        values=8,
```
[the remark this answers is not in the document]
dkriegner: As far as I understand, a validator is never used upon reading, but only upon setting a property. So I think it's not redundant.

pymeasure/instruments/validators.py
dkriegner: Certainly this can be solved with a lambda in set_process. But I think the validator provides the user with the best error message. If you don't like it in core it can certainly be moved to the dcxs.py file?

pymeasure/instruments/common_base.py
```diff
@@ -279,7 +279,7 @@ def wait_for(self, query_delay=0):
         """
         raise NotImplementedError("Implement in subclass!")
 
-    def ask(self, command, query_delay=0):
+    def ask(self, command, query_delay=0, **kwargs):
```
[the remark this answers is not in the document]
bmoneke: Yes. A mock call might suffice, or writing our own ask method, which stores the kwargs handed to it.

bmoneke:
> @bmoneke: We are accumulating instruments with "weird" behaviour quirks. Should we start a FAQ-style directory in the documentation

@bilderbuchi good idea to start a list of example solutions to special problems. I'd say at the end of the adding instruments file.

pymeasure/instruments/aja/dcxs.py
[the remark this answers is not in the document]
bilderbuchi: Ha, oops, my bad. Of course.

pymeasure/instruments/validators.py
dkriegner: I moved it into the dcxs code. I personally would prefer to keep it as an extra function because like this it can have a proper docstring

pymeasure/instruments/aja/dcxs.py
```diff
@@ -0,0 +1,229 @@
...
+            includeSCPI=False,
+            write_termination="",
+            read_termination="",
+            **kwargs
+        )
+        # here we want to flush the read buffer since the device upon power up sends some '>'
+        # characters.
+        try:
+            self.adapter.flush_read_buffer()
+        except NotImplementedError:
```
dkriegner, line 72: after merging #836 the except branches below can be removed

pymeasure/instruments/aja/dcxs.py
```diff
@@ -0,0 +1,229 @@
...
+            # occurs in test suite (see #742 -> should be removed before merging)
+            pass
+
+    def read(self, reply_length=-1, **kwargs):
+        return self.read_bytes(reply_length, **kwargs).decode()
+
+    id = Instrument.measurement(
+        "?", """Power supply type identifier""",
+        cast=str,
+        reply_length=9,
```
dkriegner, line 96: I still believe that passing through the `reply_length` to the read method is the best since any read waits for the appropriate amount of received characters and thats how the communication protocol of the device is defined.  

But after #839 one could also remove `reply_length` and it should work also if an appropriate `query_delay` is used. Relying solely on the timing seems, however, dangerous

pymeasure/instruments/common_base.py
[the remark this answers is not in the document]
dkriegner: I have added a test which tests if the kwarg is passed through to `read`. Please have a look if this is ok for you. It of course requires `ask` to stay a `write`/`read` sequence.

dkriegner (commented):
I think this PR should also wait for #836 to simplify the `__init__`

pymeasure/instruments/aja/dcxs.py
```diff
@@ -0,0 +1,229 @@
...
+                except pyvisa.errors.VisaIOError:
+                    # occurs always when calling read and no character is waiting
+                    pass
+                self.adapter.connection.timeout = timeout
+            except AttributeError:
+                # occurs in test suite (see #742 -> should be removed before merging)
+                pass
+        except AttributeError:
+            # occurs in test suite (see #742 -> should be removed before merging)
+            pass
```
dkriegner, line 88: simplification after #836 

```suggestion
        self.adapter.flush_read_buffer()
```

pymeasure/instruments/common_base.py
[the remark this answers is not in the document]
bmoneke: Oh, I just realised, that it is in the pymeasure core.

Why should the kwargs go to the read and not to write (or both)?
Because there is no simple yet general solution for that question, I did not add kwargs. The most practical solution is, that an instrument (like yours, I guess), which requires more arguments in ask, overrides ask. That way you can decide, which arguments go where.

pymeasure/instruments/aja/dcxs.py
```diff
@@ -0,0 +1,229 @@
...
+    It returns the original value if its length is smaller or equal to the maximum,
+    otherwise only the characters up to the maximum length will be returned.
+
+    :param value: a value to test
+    :param maxlength: maximum length of the string ("values" arg of the property factories)
+    """
+    if len(value) <= maxlength:
+        return value
+    else:
+        return value[:maxlength]
```
bmoneke, line 42: You can make it even shorter:
```suggestion
    return value[:maxlength]
```
A slice returns up to that many elements.

pymeasure/instruments/aja/dcxs.py
bmoneke: If we know the length of the answers, we should definitively read that many bytes and not trying to do timing vodoo.

pymeasure/instruments/aja/dcxs.py
```diff
@@ -0,0 +1,229 @@
...
+    )
+
+    software_version = Instrument.measurement(
+        "z", """Software revision of the power supply firmware""",
+        cast=str,
+        reply_length=5,
+    )
+
+    power = Instrument.measurement(
+        "d", """Actual output power in W""",
```
bmoneke, line 106: Would you mind to adjust your property docstrings? They should start with "Measure", "Set", or "Control" in order to indicate the possibilities of that property:
```suggestion
        "d", """Measure the actual output power in W.""",
```

bmoneke:
A few remarks from my side, but the code looks good.

pymeasure/instruments/common_base.py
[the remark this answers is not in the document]
dkriegner: ok, I understand your concern and I will move the `ask` implementation to the Instrument. 
But would you at least agree that values shall forward the `kwargs` to ask in `CommonBase`? 
Otherwise one would also need to reimplement `values`.

pymeasure/instruments/aja/dcxs.py
dkriegner: I will work on that

pymeasure/instruments/common_base.py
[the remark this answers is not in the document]
bmoneke: As there is no other good recipient of kwargs in values than "ask", I'm fine with handing kwargs to ask.

pymeasure/instruments/aja/dcxs.py
dkriegner: I improved the docstring. I hope its now acceptable. Let me know otherwise

pymeasure/instruments/aja/dcxs.py
```diff
@@ -0,0 +1,229 @@
...
+    The communication protocol of the device uses single character commands and fixed length replys,
+    both without any terminator.
+
+    :param adapter: pyvisa resource name of the instrument or adapter instance
+    :param string name: The name of the instrument.
+    :param kwargs: Any valid key-word argument for Instrument
+    """
+
+    def __init__(self, adapter, name="AJA DCXS sputtering power supply", **kwargs):
+        kwargs.setdefault("asrl", dict(baud_rate=38400))
```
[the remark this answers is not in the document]
dkriegner: I moved it into the super call

pymeasure/instruments/aja/dcxs.py
dkriegner: I finally removed the validator and put the single statement in a lambda functions

pymeasure/instruments/common_base.py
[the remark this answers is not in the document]
bmoneke: It is good, how values handles the kwargs, but `ask` should not accept kwargs and drop them silently.

Please remove the **kwargs from the `ask` method.

That way a user sees, that his arguments are not processed (because ask raises an error).

pymeasure/instruments/aja/dcxs.py
bmoneke: They are good. Thanks.

bmoneke:
It looks good on my side, now we have to wait for the other PRs. (#839 and #836)

bmoneke:
Thanks for your work @dkriegner

PR #785, opened 2022-12-09
DRAFT: Adding Keysight E36312A power supply

scandey:
This is a three output power supply with a very standard VISA (over Ethernet or USB) interface. 

I've started with some very basic controls across three channels. Not sure I've done everything right the first try, but I'm opening the pull request in hopes of getting early feedback.

### DIFF
--- a/pymeasure/instruments/keysight/keysightE36312A.py
+++ b/pymeasure/instruments/keysight/keysightE36312A.py
@@ -1,0 +1,98 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2022 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+
+import logging
+
+from pymeasure.instruments import Instrument, Channel
+from pymeasure.instruments.validators import strict_range, strict_discrete_set
+
+from pymeasure.adapters import VISAAdapter
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+class VoltageChannel(Channel):
+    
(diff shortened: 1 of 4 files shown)


bmoneke (commented):
Thanks for your contribution. The code looks tidy.

Could you add the proper documentation links as well, please? See the [documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#adding-documentation).

scandey:
What is the best way to check the dynamic validator values? 
I started with:
```
def test_current_limit_validator():
    """Verify the current limit validator."""
    with expected_protocol(
        KeysightE36312A,
        [(None, None)],
    ) as inst:
        with pytest.raises(ValueError):
            inst.ch_1.current_limit = 7
```
but it doesn't work since the ValueError interrupts the protocol.

pymeasure/instruments/keysight/keysightE36312A.py
```diff
@@ -0,0 +1,93 @@
...
+    VOLTAGE_RANGE = [0, 25]
+    CURRENT_RANGE = [0, 5]
+
+    voltage_setpoint = Channel.control(
+        "VOLT? (@{ch})",
+        "VOLT %g, (@{ch})",
+        """Control the output voltage of this channel, range depends on channel.""",
+        validator=strict_range,
+        values=VOLTAGE_RANGE
+    )
```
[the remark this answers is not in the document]
scandey: Thank you! I added those.

bmoneke:
If you do not expect any communication (because the validator raises the exception **before** any command is sent), you have to leave the communication list empty:
~~~python
def test_current_limit_validator():
    """Verify the current limit validator."""
    with expected_protocol(
        KeysightE36312A,
        [],  # empty list
    ) as inst:
        with pytest.raises(ValueError):
            inst.ch_1.current_limit = 7
~~~

pymeasure/instruments/keysight/keysightE36312A.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    channels = Instrument.ChannelCreator(VoltageChannel, (1, 2, 3))
+
+    def __init__(self, adapter, name="Keysight E36312A", **kwargs):
+        super().__init__(
+            adapter, name, **kwargs
+        )
+        self.channels[1].voltage_setpoint_values = [0, 6]
+        self.channels[1].current_limit_values = [0, 5]
```
bmoneke, line 94: Only the first channel has different limits than the other channels or why do you limit them here?

tests/instruments/keysight/test_keysightE36312A.py
```diff
@@ -0,0 +1,69 @@
...
+        assert inst.ch_3.current_limit == 0.5
+
+# def test_current_limit_validator():
+#     """Verify the current limit validator."""
+#     with expected_protocol(
+#         KeysightE36312A,
+#         [(None, None)],
+#     ) as inst:
+#         with pytest.raises(ValueError):
+#             inst.ch_1.current_limit = 7
```
bmoneke, line 58: You do not expect any communication, so the list has to be empty. You can even verify, that the right Error text is raised:
```suggestion
def test_current_limit_validator():
    """Verify the current limit validator."""
    with expected_protocol(
        KeysightE36312A,
        [],
    ) as inst:
        with pytest.raises(ValueError, match="not in range"):
            inst.ch_1.current_limit = 7
```

bmoneke:
Thanks for the contribution.
Two small conversations and we're good to go, I think.

Please verify, that the indenting spaces in the test changed by me are correct.

pymeasure/instruments/keysight/keysightE36312A.py
scandey: Correct, only the first channel has a different set of limits.

tests/instruments/keysight/test_keysightE36312A.py
scandey: Resolved in separate commit

bmoneke:
Just to confirm: The code can be merged from your side?

scandey:
> Just to confirm: The code can be merged from your side?

I pulled the AUTHORS change from master back into my branch so now I have no conflicts. I don't have permission to merge though, if that's what you're asking?

scandey:
I am ready for merge! Thank you for checking.

bmoneke:
Thanks for your contribution.

PR #787, opened 2022-12-12
Anritsu ms2090a

aruznieto:


pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+class AnritsuMS2090A(Instrument):
+    """Anritsu MS2090A Handheld Spectrum Analyzer."""
+
+    def __init__(self, adapter, **kwargs):
+        """Constructor."""
+        self.analysis_mode = None
+        super().__init__(
+            adapter, name="Anritsu MS2090A Handheld Spectrum Analyzer", **kwargs)
```
bmoneke, line 43:
```suggestion
    def __init__(self, adapter, name="Anritsu MS2090A Handheld Spectrum Analyzer", **kwargs):
        self.analysis_mode = None
        super().__init__(
            adapter, name, **kwargs)
```
First, you do not need a docstring for the constructor, because the class docstring serves also as constructro docstring.
Second, we require a name parameter to ensure easy subclassing.

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,516 @@
...
+    UNITSFREQ_MAP = []
+
+    ####################################
+    #              GPS                 #
+    ####################################
+
+    gps_full = Instrument.measurement(
+        "FETCh:GPS:FULL?",
+        '''
+        Returns the timestamp, latitude, longitude, altitude, and satellite count of the device.
```
bmoneke, line 63: Please look at the documentation regarding [docstrings](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings). You will find examples of good docstrings in the [adding instruments documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html).
Especially, The first sentence of a docstring (Python stlye) should be an imperative. Further, the docstrings of `measurement`, `setting`, and `control` should start with "Measure", "Set", "Control" (pymeasure style) as you cannot see in the documentation, whether a property is only gettable, only settable, or both.

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,516 @@
...
+    )
+
+    ####################################
+    # Spectrum Parameters - Wavelength #
+    ####################################
+
+    frequency_center = Instrument.control(
+        "FREQuency:CENTer?", "FREQuency:CENTer %s",
+        "Sets the center frequency in Hz",
+        validator=double_validation_value_and_freq,
```
bmoneke, line 102: why do you use your double validator, if you state, that you expect the frequency in Hz?

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,516 @@
...
+        values=[[-99999999990, 300000000000], UNITSFREQ],
+    )
+
+    abort = Instrument.setting(
+        "ABOR",
+        '''
+        Abort measure
+        '''
+    )
```
bmoneke, line 156: An "action", like abort, should be a method, not a property, as you do not set anything, see [documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#your-instrument-s-user-interface).

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,516 @@
...
+
+    fet_power = Instrument.measurement(
```
bmoneke, line 158: What do you mean with "fet_power"? Do you measure the power? Why don't you just call it "power"?
Similarly the measurements below.
For naming hints see the [documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#your-instrument-s-user-interface).

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,516 @@
...
+        values=ONOFF,
+        map_values=True
+    )
+
+    init_sweep = Instrument.setting(
+        "INIT",
+        '''
+        Initiates a sweep/measurement.
+        '''
+    )
```
bmoneke, line 356: should be a method, like abort:
```suggestion
    def init_sweep(self):
        '''
        Initiate a sweep/measurement.
        '''
        self.write("INIT")
```

similarly "init_sweep_all"

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,516 @@
...
+
+    init_spa_self = Instrument.measurement(
+        "INIT:SPA:SELF?",
+        docs='''
+        Perform a self-test and return the results.
+        '''
+    )
+
+    inst_active_state = Instrument.control(
```
bmoneke, line 372: You do not have to imitate the communication while naming properties. Would not "active_state" be better?

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,516 @@
...
+    meas_power = Instrument.measurement(
+        "MEASure:CHPower:CHPower?",
+        '''
+        Sets the active measurement to channel power, sets the default measurement
+        parameters, triggers a new measurement and returns channel power as the result. It is a
+        combination of the commands :CONFigure:CHPower; :READ:CHPower:CHPower?
+        '''
+    )
+
+    meas_density = Instrument.measurement(
```
bmoneke, line 410: better name?
```suggestion
    power_density= Instrument.measurement(

```

bmoneke (changes requested):
Thanks for your contribution.

I added a few comments to the code, in the hope, that a fast feedback helps you to implement it already in the upcoming changes.

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,535 @@
...
+                value_1, values_1
+            ))
+
+        if min(values_2) <= value_2 <= max(values_2):
+            pass_range = value_2
+        else:
+            raise ValueError('Value of {:g} is not in range [{:g},{:g}]'.format(
+                value_2, min(values_2), max(values_2)
+            ))
+        return f'{pass_range} {pass_discrete}'
```
bmoneke, line 65: The validator has to be outside the class definition.

Don't you think it is cumbersome to write `inst.frequency = 500, "MHz"` instead of `inst.frequency = 500e6`, especially if you try to access these controls from a program?

I think the best way is, if the methods accept a value in Hz and return the value in Hz, such that nobody has to fiddle with the units string.

If for some reason (accuracy etc.) you have to use the different units, you could make your validator accept a float (frequency in Hz) and then the validator decides what unit (Hz, kHz, MHz, or GHz) is best and hands the (modified) float and the corresponding unit string to the device.

aruznieto: Yes, maybe it is better only introduce it on Hz. I am going to change it

pymeasure/instruments/anritsu/anritsuMS9710C.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, adapter, name="Anritsu MS9710C Optical Spectrum Analyzer", **
     #  Mappings #
     #############
     ONOFF = ["ON", "OFF"]
-    ONOFF_MAPPING = {True: 'ON', False: 'OFF', 1: 'ON', 0: 'OFF'}
+    ONOFF_MAPPING = {True: 'ON', False: 'OFF', 1: 'ON', 0: 'OFF', 'ON': 'ON', 'OFF': 'OFF'}
```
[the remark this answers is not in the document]
bilderbuchi: Agreed, I also prefer a single-key interface (True/False) -- otherwise you just accumulated different key pairs ~endlessly, because why not _also_ allow `yes`/`no`, `on`/`off`,...

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,509 @@
...
+        self.analysis_mode = None
+        super().__init__(
+            adapter, name, **kwargs)
+
+    #############
+    #  Mappings #
+    #############
+
+    ONOFF = ["ON", "OFF"]
+    ONOFF_MAPPING = {True: 'ON', False: 'OFF', 1: 'ON', 0: 'OFF'}
```
bilderbuchi, line 48: As mentioned on the other instrument, can you limit the interface to `True` and `False` as keys? This is what we tend to do in other instruments, too, and it keeps the selection of possibilities short...

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,509 @@
...
+        '''
+        Returns the timestamp, latitude, longitude, altitude, and satellite count of the device.
+        '''
+    )
+
+    gps_all = Instrument.measurement(
+        "FETCh:GPS:ALL?",
+        '''
+        Returns the fix timestamp, latitude, longitude, altitude and information on the sat used.
+        '''
```
bilderbuchi, line 67: Please use PEP257-compatible docstring formatting. In this case, delimit them with three double quotes `"""`, and don't start with a newline
```suggestion
        """Returns the fix timestamp, latitude, longitude, altitude and information on the sat used.
        """

```

pymeasure/instruments/anritsu/anritsuMS2090A.py
aruznieto: I do it but now test throws error.

tests/instruments/anritsu/test_anritsuMS2090A.py
```diff
@@ -0,0 +1,51 @@
...
+            ) as instr:
+        assert instr.frequency_center == 4
+
+
+def test_preamp():
+    with expected_protocol(
+            AnritsuMS2090A,
+            [(b"POWer:RF:GAIN:STATe?", 'ON')],
+            ) as instr:
+        assert instr.preamp == 'ON'
```
bmoneke, line 51: That test fails, because the property returns True or False (map_values works both ways).

aruznieto: I don't know how it works, the instrument needs "ON" or "OFF". If I put this It doesn't work

```python
def test_preamp():
    with expected_protocol(
            AnritsuMS2090A,
            [(b"POWer:RF:GAIN:STATe?", True)],
            ) as instr:
        assert instr.preamp == True
```

aruznieto: Ok. I understand it...

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,513 @@
...
+
+    ####################################
+    # Spectrum Parameters - Wavelength #
+    ####################################
+
+    frequency_center = Instrument.control(
+        "FREQuency:CENTer?", "FREQuency:CENTer %g",
+        "Sets the center frequency in Hz",
+        validator=truncated_range,
+        values=[-99999999995, 299999999995]
```
LongnoseRob, line 98: Would it not technically make more sense to limit this range to [9000, 54E9] per the specifiaction of the instrument? 
Also it might be better to use the `strict_range` type of validator

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -0,0 +1,513 @@
...
+    frequency_span_last = Instrument.setting(
+        "FREQuency:SPAN:LAST",
+        "Sets the frequency span to the previous span value."
+    )
+
+    frequency_start = Instrument.control(
+        "FREQuency:STARt?", "FREQuency:STARt %g",
+        "Sets the start frequency in Hz",
+        validator=truncated_range,
+        values=[-100000000000, 299999999990],
```
LongnoseRob, line 129: as above, does a negative frequency make sense in this case?

LongnoseRob (commented):
Thank you for the addition of this instrument,
I have left a few comments regarding the range of some parameters and the validator useage..

tests/instruments/anritsu/test_anritsuMS2090A.py
bmoneke: The list (argument to expected protocol) contains the communication with the device ("on"), while you verify the result of your code: `assert inst.preamp is True`.

That way you can test whether user input/driver output translates correctly to communication with the device (according to the manual, corroborated by real world tests).

LongnoseRob (approved):
thanks, LGTM

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -25,7 +25,8 @@
 from pymeasure.instruments import Instrument
 from pymeasure.instruments.validators import (
     strict_discrete_set,
-    truncated_range
+    truncated_range,
+    strict_range
```
bmoneke, line 29: Just an information: In such a case it is good to add a comma behind strict_range as well, i.e. just add a comma behind every item, even the last one.
That way you can add or remove elements (by removing their line) in that list without affecting other lines (as you now changed line 28 in order to add line 29).

pymeasure/instruments/anritsu/anritsuMS2090A.py
```diff
@@ -95,21 +95,21 @@ def __init__(self, adapter, name="Anritsu MS2090A Handheld Spectrum Analyzer", *
     frequency_center = Instrument.control(
         "FREQuency:CENTer?", "FREQuency:CENTer %g",
         "Sets the center frequency in Hz",
-        validator=truncated_range,
-        values=[-99999999995, 299999999995]
+        validator=strict_range,
+        values=[9E3, 54E9]
```
bmoneke, line 99: Did you correct an implementation error, because the limits changed.

bmoneke (approved):
Just one informational comment and one question: You changed the limits of some controls. Do they now reflect the manual/data sheet?

pymeasure/instruments/anritsu/anritsuMS2090A.py
aruznieto: I change the limits as [LongnoseRob](https://github.com/LongnoseRob) says me. In the manual appears the limits that I write at the begin

pymeasure/instruments/anritsu/anritsuMS9710C.py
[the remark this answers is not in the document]
bmoneke: Oh, and please revert this change.

pymeasure/instruments/anritsu/anritsuMS2090A.py
aruznieto: Yes, but If you read [Programming Manual](https://dl.cdn-anritsu.com/en-us/test-measurement/files/Manuals/Programming-Manual/10580-00445R.pdf) of the instrument, the limits are other. But I think that It is better your recommendation.

PR #790, opened 2022-12-13
Fix issue with ImageWidget

CasperSchippers:
When playing around with another PR, I came  across an issue with the ImageWidget: when a new measurement is queued, it crashes because the initialization of the `ResultsImage` class calls the `scale` and `translate`, methods, which does (nowadays) only return the scaling or translation values.

### DIFF
--- a/pymeasure/display/curves.py
+++ b/pymeasure/display/curves.py
@@ -26,7 +26,7 @@ import logging
 
 import numpy as np
 import pyqtgraph as pg
-from .Qt import QtCore
+from .Qt import QtCore, QtGui
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
@@ -82,11 +82,13 @@ class ResultsImage(pg.ImageItem):
 
         super().__init__(image=self.img_data)
 
-        # Scale and translate image so that the pixels are in the coorect
+        # Scale and translate image so that the pixels are in the correct
         # position in "data coordinates"
-        self.scale(self.xstep, self.ystep)
-        self.translate(int(self.xstart / self.xstep) - 0.5,
-                       int(self.ystart / self.ystep) - 0.5)  # 0.5 so pixels centered
+        tr = QtGui.QTransform()
+        tr.scale(self.xstep, self.ystep)
+        tr.translate(int(self.xstart / self.xstep) - 0.5,
+                     int(self.ystart / self.ystep) - 0.5)  # 0.5 so pixels centered
+        self.setTransform(tr)
 
     def update_data(self):
         if self.force_reload:


msmttchr:
> When playing around with another PR, I came across an issue with the ImageWidget: when a new measurement is queued, it crashes because the initialization of the `ResultsImage` class calls the `scale` and `translate`, methods, which does (nowadays) only return the scaling or translation values.

What version of pyqtgraph are you using? In my PC, I have 0.12.3 and I see no crashes.
In any case, your fix should be good for future evolution.

CasperSchippers:
Ah that seems to be the source of the issue indeed (I looked for changes or mentions of `scale` in the documentations and changelogs, but was not able to find it).

Could you confirm this fix also works (or rather doesn't break it) on your 0.12.3 installation. (It did for me, but you never know).

bilderbuchi:
This probably means we also need to bump the minimum version of pyqtgraph in setup.cfg, to account for the new api usage, right?

msmttchr:
> Ah that seems to be the source of the issue indeed (I looked for changes or mentions of `scale` in the documentations and changelogs, but was not able to find it).
> 
> Could you confirm this fix also works (or rather doesn't break it) on your 0.12.3 installation. (It did for me, but you never know).

I confirm, this is working on my side without problems also with 0.12.3.

msmttchr (approved):
OK for me. The problem is not shown with `pyqtgraph `version aligned to our current requirement, but it is useful to avoid headaches for future updates of `pyqtgraph `

msmttchr:
> This probably means we also need to bump the minimum version of pyqtgraph in setup.cfg, to account for the new api usage, right?

I don't think so, current version of `pyqtgraph`  and newer version available today are ok with the code of this PR.

bilderbuchi:
> > This probably means we also need to bump the minimum version of pyqtgraph in setup.cfg, to account for the new api usage, right?
> 
> I don't think so, current version of `pyqtgraph` and newer version available today are ok with the code of this PR.

Ah, so with the code in this PR, this is OK both with the current newest version of pyqtgraph and the minimum pyqtgraph version as defined in setup.cfg? Do I understand this correctly?

msmttchr:
Correct.

Il Ven 16 Dic 2022, 20:18 Christoph Buchner ***@***.***> ha
scritto:

bilderbuchi:
Great, thanks!

msmttchr:
I found out that these functions (scale, translate) were already deprecated in 0.12.3 and they have been removed from `pyqtgraph/Qt/__init__.py` in commit https://github.com/pyqtgraph/pyqtgraph/commit/895be36cd7bfc4dbc5d22f288b539085cf789a9f.

So, it is totally safe and appropriate to merge.

PR #791, opened 2022-12-15
Add TeledyneT3AFG waveform generator instrument

scandey:
In progress instrument for the T3AFG signal generator family of instruments. I only have access to the T3AFG80, but I wanted to start with a general implementation and the ranges could be adjusted later in inheritance if other people want to make interfaces for other models. 

My big question for this instrument is how to handle the strings that are returned by the instrument in get_process in a way that is dependent on the current waveform type. For instance, the frequency control doesn't make any sense if the wavetype is DC so the instrument doesn't even return it when the waveform info is query'd. 

https://cdn.teledynelecroy.com/files/manuals/t3afg-programming-guide.pdf

### DIFF
--- a/docs/api/instruments/teledyne/index.rst
+++ b/docs/api/instruments/teledyne/index.rst
@@ -1,0 +1,12 @@
+.. module:: pymeasure.instruments.teledyne
+
+#########
+Teledyne
+#########
+
+This section contains specific documentation on the Teledyne instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments>`.
+
+.. toctree::
+   :maxdepth: 2
+
+   teledyneT3AFG
--- a/pymeasure/instruments/teledyne/teledyneT3AFG.py
+++ b/pymeasure/instruments/teledyne/teledyneT3AFG.py
@@ -25,14 +25,24 @@
 import logging
 
 from pymeasure.instruments import Instrument, Channel
-from pymeasure.instruments.validators import strict_range
+from pymeasure.instruments.validators import strict_range, strict_discrete_set
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 
 class SignalChannel(Channel):
-    pass
+    
+    output_enabled = Channel.control(
+        "C{ch}:OUTPut?",
+        "C{ch}:OUTPut %s",
+        """Control whether the channel output is enabled (boolean).""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 'ON', False: 'OFF'},
+        get_process=lambda x: x[0].split(' ')[1],
+        # Not sure why "OFF" doesn't get transformed into False
+    )
 
 
 class TeledyneT3AFG(Instrument):
@@ -47,7 +57,7 @@ class TeledyneT3AFG(Instrument):
     generator=TeledyneT3AFG(resource)
     """
 
-    channels = Instrument.ChannelCreator(SignalChannel)
+    channels = Instrument.ChannelCreator(SignalChannel, (1,2))
 
     def __init__(self, adapter, name="Teledyne T3AFG", **kwargs):
         super().__init__(
--- a/pymeasure/instruments/teledyne/teledyneT3AFG.py
+++ b/pymeasure/instruments/teledyne/teledyneT3AFG.py
@@ -38,8 +38,11 @@ class SignalChannel(Channel):
 class TeledyneT3AFG(Instrument):
     """Represents the Teledyne T3AFG series of arbitrary waveform
     generator interface for interacting with the instrument.
-    Limits are not accurate for T3AFG5, T3AFG10.
-    
+
+    Intially targeting T3AFG80, some features may not be avalible on
+    lower end models and features from higher end models are not 
+    included here intially. 
+
     .. code-block: python
     generator=TeledyneT3AFG(resource)
     """
--- a/pymeasure/instruments/teledyne/teledyneT3AFG.py
+++ b/pymeasure/instruments/teledyne/teledyneT3AFG.py
@@ -44,6 +44,57 @@ class SignalChannel(Channel):
         # Not sure why "OFF" doesn't get transformed into False
     )
 
+    # TODO: Add other OUTPut related controls like Load and Polarity
+
+    wavetype = Channel.control(
+        "C{ch}:BSWV?",
+        "C{ch}:BSWV WVTP,%s",
+        """Control the type of waveform to be output.
+        Options are: {SINE, SQUARE, RAMP, PULSE, NOISE, ARB, DC, PRBS, IQ}
+        """,
+        validator=strict_discrete_set,
+        # TODO: make ENUM for more descriptive names?
+        values=['SINE', 'SQUARE', 'RAMP', 'PULSE', 'NOISE', 'ARB', 'DC', 'PRBS', 'IQ'],
+        get_process=lambda x: x[1],
+    )
+
+    # TODO: Wavetype dependent processing with errors
+    # For example, DC wavetype does not return frequency
+
+    # TODO: Add frequency ranges per model?
+    frequency = Channel.control(
+        "C{ch}:BSWV?",
+        "C{ch}:BSWV FRQ,%g",
+        """Control the frequency of waveform to be output in Hertz.""",
+        validator=strict_range,
+        values=[0,350e6],
+        get_process=lambda x: x,
+        #get_process=lambda x: x[1].split(',')[2].split('HZ'),
+    )
+
+    # TODO: Add range of outputs per model? Tricky without knowing offset
+    amplitude = Channel.control(
+        "C{ch}:BSWV?",
+        "C{ch}:BSWV AMP,%g",
+        """Control the amplitude of waveform to be output in volts peak-to-peak.""",
+        validator=strict_range,
+        values=[0,5],
+        get_process=lambda x: x[1].split(',')[6].split('V'),
+    )
+
+    # TODO: Add range of offset per model? Tricky without knowing amplitude
+    offset = Channel.control(
+        "C{ch}:BSWV?",
+        "C{ch}:BSWV OFST,%g",
+        """Control the offset of waveform to be output in volts.""",
+        validator=strict_range,
+        values=[0,5],
+        get_process=lambda x: x[1].split(',')[8].split('V'),
+    )
+
+    # TODO: Add other Basic Waveform parameters like period
+
+    # TODO: Add odd parameter like MAX_OUTPUT_AMP
 
 class TeledyneT3AFG(Instrument):
     """Represents the Teledyne T3AFG series of arbitrary waveform
@@ -63,3 +114,5 @@ class TeledyneT3AFG(Instrument):
         super().__init__(
             adapter, name, **kwargs
         )
+
+    # TODO: Add channel coupling control
--- a/pymeasure/instruments/teledyne/teledyneT3AFG.py
+++ b/pymeasure/instruments/teledyne/teledyneT3AFG.py
@@ -1,0 +1,52 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2022 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import logging
+
+from pymeasure.instruments import Instrument, Channel
+from pymeasure.instruments.validators import strict_range
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class SignalChannel(Channel):
+    pass
+
+
+class TeledyneT3AFG(Instrument):
+    """Represents the Teledyne T3AFG series of arbitrary waveform
+    generator interface for interacting with the instrument.
+    Limits are not accurate for T3AFG5, T3AFG10.
+    
+    .. code-block: python
+    generator=TeledyneT3AFG(resource)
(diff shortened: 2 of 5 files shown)


LongnoseRob:
Thank you for the contribution!


> In progress instrument for the T3AFG signal generator family of instruments. I only have access to the T3AFG80, but I wanted to start with a general implementation and the ranges could be adjusted later in inheritance if other people want to make interfaces for other models.

Instruments from the same family with different specifications (e.g. higher frequency range) could be added easily with the [dynamic validity ranges](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#dynamic-properties)


> 
> My big question for this instrument is how to handle the strings that are returned by the instrument in get_process in a way that is dependent on the current waveform type. For instance, the frequency control doesn't make any sense if the wavetype is DC so the instrument doesn't even return it when the waveform info is query'd.
> 

If the frequency settings are just ignired by the instrument without error, I would just put a warning/note in the docstring that this setting has no meaning in DC mode.

If the instrument would trigger an alarm status, it might be worth checking if `check_set_errors` could chatch it and issue an Exception, see also [here](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#checking-the-instrument-for-errors)

> https://cdn.teledynelecroy.com/files/manuals/t3afg-programming-guide.pdf
Thank you for sharing the link to the programmin manual, this is a helpful reference.

One more comment as you created a new directory `teledyne` for this instrument.

1. there seems to be a missing entry for `teledyne/indes.rst `in `docs/api/instruments/index.rst`
2.  the company is offically called `Teledyne Lecroy`, so maybe it might be worth thinking about moving `teledyne` and `lecroy` together.  @bmoneke @CasperSchippers @bilderbuchi  any opinion on this?

Please have a look at the issues reported by the linter during the CI

pymeasure/instruments/teledyne/teledyneT3AFG.py
```diff
@@ -0,0 +1,118 @@
...
+    
+    output_enabled = Channel.control(
+        "C{ch}:OUTPut?",
+        "C{ch}:OUTPut %s",
+        """Control whether the channel output is enabled (boolean).""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 'ON', False: 'OFF'},
+        get_process=lambda x: x[0].split(' ')[1],
+        # Not sure why "OFF" doesn't get transformed into False
```
bmoneke, line 44: Consider writing a test for this, which will allow you to compare expectation to your code.
Verify, that your device responds with "OFF" (in that part of the message).

scandey: The string returned by the instrument is quite long, but I put an example in the test for this control.

I'm not sure that I'm doing the mapping right though. The result of the get_process string parsing lambda does not appear to go through the "map_values" mapping before being returned to the user. I just get "OFF" as opposed to my expected False. Are get_process and map_values mutually exclusive?

bmoneke:

Regarding the name: Earlier it was only "LeCroy" and after its acquisition by "Teledyne" renamed in the combined name. I prefer to keep them under the name "LeCroy", because that way people looking for older models will find them as well.

> My big question for this instrument is how to handle the strings that are returned by the instrument in get_process in a way that is dependent on the current waveform type. For instance, the frequency control doesn't make any sense if the wavetype is DC so the instrument doesn't even return it when the waveform info is query'd.

I just had a look at the manual to understand your question better: First tip, implement quite a few of these examples as tests (it's wonderful how many communication examples there are), that way you know, whether your get_process methods work.

You have to dissect strings, for which you can use different methods:

- regex (more complicated, not necessary here, I'd say)
- you can split at "FREQ" for example and check, whether that is present
- you can split at the "," and then return the next value: `get_process=lambda vals: v[1].split(",")[v[1].split(",").index("FRQ")+1]`.

Regarding the last solution, to include catching errors etc. you can create a function:
~~~python
def get_process_generator(keyword):
    """Generate a `get_process` method searching for `keyword`."
    def selector(values):
        values = values[1].split[","]
        try:
            return values[values.index(keyword) + 1]
        except (ValueError, IndexError):
            return None
    return selector
~~~
In the controls, you define `get_process=get_process_generator("FRQ")`. That way you get the content of the entry after your keyword, if it exists, or None.

pymeasure/instruments/teledyne/teledyneT3AFG.py
bmoneke: Oh yes, right now `map_values` works only, if `values` returns a single value. #740 tries to fix this.

You can do it in the get_process, if you like:
```python
        get_process=lambda x: :True if [0].split(" ")[1] == "ON" else False
```

scandey: I went with that suggestion, thank you!

scandey:
Regarding the company naming, I'm happy with moving these to the Lecroy folder or a new Teledyne-Lecroy folder. However, the actual instrument here has always been a Teledyne instrument and the branding on the actual instrument only says Teledyne, so an end user would probably be a little confused with just the Lecroy folder option.

bmoneke:
If it is branded as Teledyne (and not also as LeCroy elsewhere), I'm for keeping it under the name Teledyne.
We can link in the documentation from one manufacturer to the other (Teledyne acquired LeCroy, for older devices look there, for example).

scandey:
I didn't adjust the documentation to link Teledyne to LeCroy, not sure how to do it with the docs system. That being said, I believe the functionality related to this generator is ready for review/merge.

docs/api/instruments/teledyne/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.teledyne
+
+#########
+Teledyne
+#########
+
+This section contains specific documentation on the Teledyne instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments>`.
+
```
bmoneke, line 8: That is, how you can connect to another documentation file with `:doc:'path_to_file_without_rst'` (you have to change the apostroph according to the example below). I made a suggestion of text:
```suggestion

You may find older devices under the name :doc:`../lecroy/index`.

```

Please add a similar comment to the `lecroy/index.rst` file, linking to this file.

pymeasure/instruments/teledyne/teledyneT3AFG.py
```diff
@@ -0,0 +1,131 @@
...
+        "C{ch}:BSWV WVTP,%s",
+        """Control the type of waveform to be output.
+        Options are: {SINE, SQUARE, RAMP, PULSE, NOISE, ARB, DC, PRBS, IQ}
+        """,
+        validator=strict_discrete_set,
+        values=['SINE', 'SQUARE', 'RAMP', 'PULSE', 'NOISE', 'ARB', 'DC', 'PRBS', 'IQ'],
+        get_process=lambda x: x[1],
+    )
+
+    # TODO: Add frequency ranges per model?
```
bmoneke, line 74: For your information. With [dynamic properties](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#dynamic-properties) you can change the ranges per module afterwards.

pymeasure/instruments/teledyne/teledyneT3AFG.py
```diff
@@ -0,0 +1,131 @@
...
+    # TODO: Add range of offset per model? Tricky without knowing amplitude
+    offset = Channel.control(
+        "C{ch}:BSWV?",
+        "C{ch}:BSWV OFST,%g",
+        """Control the offset of waveform to be output in volts.
+        Has no effect when WVTP is NOISE.""",
+        validator=strict_range,
+        values=[0, 5],
+        get_process=get_process_generator('OFST', 'V', float),
+    )
```
bmoneke, line 105: I'd add a comment to the docstrings, that the maximum value depends on the currently set offset/amplitude.
Does the instrument raise an error if the value is wrong? Could you read that error?

SCPI compliant devices allow to read the error state. You can call the method "check_errors" after a setter has set its value, if you add the option `check_set_errors=True`, see [documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#checking-the-instrument-for-errors).

bmoneke (commented):
A few comments how you could solve your questions.

pymeasure/instruments/teledyne/teledyneT3AFG.py
scandey: I'll go ahead and preemptively make the relevant properties dynamic, but I am not planning on writing instrument definitions for other instruments in the family since I can't test them.

pymeasure/instruments/teledyne/teledyneT3AFG.py
scandey: I've added the check_set_errors flag and that appears to be working, though the -108.0, "Parameter not allowed" error appears after every command. The commands appear to work properly though, so I'm not sure what to make of the error appearing constantly.

pymeasure/instruments/teledyne/teledyneT3AFG.py
```diff
@@ -0,0 +1,153 @@
...
+    generator.ch_1.wavetype='SINE'
+    generator.ch_1.amplitude=2
+    generator.ch_1.output_enabled=True
+    """
+
+    channels = Instrument.ChannelCreator(SignalChannel, (1, 2))
+
+    def __init__(self, adapter, name="Teledyne T3AFG", **kwargs):
+        super().__init__(
+            adapter, name, **kwargs
```
bmoneke, line 150: Does this instrument use default SCPI commands? If so, could you please add it explicitly:
```suggestion
            adapter, name, includeSCPI=True, **kwargs
```

pymeasure/instruments/teledyne/teledyneT3AFG.py
bmoneke: the `check_errors` sends "SYST:ERR?" (default SCPI command) to the device (several times, until no error is returned). Maybe that command is not right for your device.

pymeasure/instruments/teledyne/teledyneT3AFG.py
```diff
@@ -0,0 +1,131 @@
...
+            if keyword in values:
+                try:
+                    return type(values[values.index(keyword)+1].strip(unit))
+                except (ValueError, IndexError):
+                    # Something went quite wrong if the keyword exists but the value doesn't
+                    return None
+            else:
+                # Wrong wavetype for this keyword
+                return None
+        return selector
```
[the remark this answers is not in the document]
scandey: I chose to put the util method outside the channel class in this case because I can imagine it being useful for the larger instrument class if someone wanted to implement instrument level controls

pymeasure/instruments/teledyne/teledyneT3AFG.py
scandey: The programming manual doesn't include anything obviously claiming it has all of the standard SCPI commands and doesn't list most of them, but they all seem to work so I'll add this.

pymeasure/instruments/teledyne/teledyneT3AFG.py
scandey: SYST:ERR? seems to be working fine, just every regular command I've tried seems to cause a parameter error. 

Unfortunately there isn't anything in the programming manual about variations on the default SCPI schema or examples including SYST:ERR? commands. There also isn't any detail on what specifically various errors are related to.

scandey: For context, the debug log looks like this:
```
>>> t.ch_1.amplitude=1
DEBUG:WRITE:C1:BSWV AMP,1
DEBUG:WRITE:SYST:ERR?
DEBUG:TCPIP0::128.32.13.70::5023::SOCKET - reading 20480 bytes (last status <StatusCode.success_max_count_read: 1073676294>)
DEBUG:READ:-108,"Parameter not allowed"
ERROR:Teledyne T3AFG: -108.0, "Parameter not allowed"
DEBUG:WRITE:SYST:ERR?
DEBUG:TCPIP0::128.32.13.70::5023::SOCKET - reading 20480 bytes (last status <StatusCode.success_max_count_read: 1073676294>)
DEBUG:READ:0,"No error"
```
```
>>> t.ch_1.amplitude
DEBUG:WRITE:C1:BSWV?
DEBUG:TCPIP0::128.32.13.70::5023::SOCKET - reading 20480 bytes (last status <StatusCode.success_max_count_read: 1073676294>)
DEBUG:READ:C1:BSWV WVTP,SINE,FRQ,80000000HZ,PERI,1.25e-08S,AMP,1V,AMPVRMS,0.3535Vrms,MAX_OUTPUT_AMP,20V,OFST,5V,HLEV,5.5V,LLEV,4.5V,PHSE,0
1.0
```

bmoneke: Interesting, that it always raises an error regarding invalid parameters.
Please add a note, probably to the instrument docstring, regarding this behaviour.

scandey: Added a note, I'll come back and update if I ever figure out why that's happening

scandey: Aha! I had a write termination of '\n' which worked initially with the USB interface, but does not work for the Ethernet interface that I'm testing with remotely. Removing the write termination solves this issue.

bmoneke: You know how you can define the endings for different connection types?

Here is the corresponding documentation : https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#multiple-interfaces

scandey: Ah, that's very handy! I'll add the ethernet interface that I am sure of for now. When I am back at work I'll test out all of my devices and I can go through and add more interface args for all of them.

bilderbuchi:
This looks reasonable to me.
I agree on the naming, best to go with the branding on the device, it doesn't make sense following manufacturers through their merger&acquisitions cycles :-P (see also Keithley->Tektronix or HP->Agilent->Keysight)

pymeasure/instruments/teledyne/teledyneT3AFG.py
```diff
@@ -0,0 +1,157 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2022 PyMeasure Developers
```
bmoneke, line 4:
```suggestion
# Copyright (c) 2013-2023 PyMeasure Developers
```

bmoneke (approved):
I'm unsure, what to do with the "TODO" comments, whether it is better to put one remark, collecting what could be done, or leaving them as they are right now.

The rest of the code looks good. If you agree, I'll merge it.

bilderbuchi:
> I'm unsure, what to do with the "TODO" comments, whether it is better to put one remark, collecting what could be done, or leaving them as they are right now.

I'd condense them into one remark summarizing what capability is missing/could be added, either as a comment or in the class docstring.
A `TODO`/`FIXME` implies to me that the item is an enhancement/bug that is not urgent enough for an issue/immediate implementation, but was noticed in passing while working on other things. For example, work tracking while implementing a PR -- I'd expect most TODOs to be dealt with at the end of a PR.

Missing instrument functionality basically waits for somebody needing it, but won't be addressed on its own. We don't aim for functionally-complete instrument implementations in pymeasure. If we had TODOs for every missing feature in every instrument, that would drown out other TODO items.

scandey:
> I'd condense them into one remark summarizing what capability is missing/could be added, either as a comment or in the class docstring.

I've removed the TODO comments and added a "Future Improvements (help welcomed)" section to the class docstring for visibility. In the future I will probably implement some of those improvements, but I want to make it obvious that other contributors are welcome to work on the instrument. (I've been nervous to mess with other people's instrument implementations so far, not sure about the etiquette there.)

bilderbuchi:
> (I've been nervous to mess with other people's instrument implementations so far, not sure about the etiquette there.)

Don't worry, everybody is glad about "free" bug fixes or feature extensions for their instruments! Pooling effort is kinda the idea of pymeasure.

bmoneke:
May I merge it @scandey?

This is open source, we're glad, if people improve and expand it. Just keep backward compatibility in mind.

scandey:
Yes, please merge!

PR #793, opened 2022-12-15
Add maxsplit parameter to values.

bmoneke:
I added the `maxsplit` parameter to values to hand it to the split function.

With that, you can deactivate it (`maxsplit=0`) and get a single value, such that mapping works (see #791 ) even with a single result.

It is also good to deactivate splitting, if you expect binary data (i.e. any character).

Tests are expanded accordingly.


msmttchr (approved):
Ok for me

bilderbuchi:
Yeah, that sounds useful in some corner cases. LGTM

PR #794, opened 2022-12-16
Fix checkboxes not working for groups in inputs-widget

CasperSchippers:
I came across an issue that prevented the input-groups from working correctly when they were controlled by a checkbox. This issue only appears when using PyQt6 and PySide6. In this PR this is fixed by changing from the stateChanged to the toggled signal.


bilderbuchi:
Is ot feasible to add a test for the breaking behaviour, which then shows that it's fixed by the implemented change?

bmoneke:
Good idea to expand the tests.
There is a pyqt pytest plug in.
However, some more complex workings are different to test.

CasperSchippers:
That's indeed a good idea. I've had some time to find out how pytest-qt works and implemented some tests.

msmttchr (approved):
OK for me.

bilderbuchi:
@CasperSchippers do you want to merge this?

CasperSchippers:
Ah, right, yes; I'll merge this

PR #795, opened 2022-12-17
Fix keithley2600.py shutdown()

mcdo0486:
The `Channel` is not a subclass of `Instrument`, so calling super().shutdown() doesn't do anything.

Fixed to call instrument.shutdown() instead.

### DIFF
--- a/pymeasure/instruments/keithley/keithley2600.py
+++ b/pymeasure/instruments/keithley/keithley2600.py
@@ -321,4 +321,3 @@ class Channel:
         else:
             self.ramp_to_voltage(0.0)
         self.source_output = 'OFF'
-        self.instrument.shutdown()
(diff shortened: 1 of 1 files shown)


bmoneke:
Thanks for pointing at that mistake. However, your solution is not sound: If you shut down one channel, you do not want to do the instrument shutdown as well.
The line should be removed altogether.

Another point:
I don't know, what the idea of that method is. A proper way to include it in the overall scheme, would be, that the (not yet existing) `instrument.shutdown` calls the `shutdown` method of all channels.

If you want, you can implement this. If you want to do so, please consider changing the `Channel` class (in this file) to be a subclass of our newly created `instrument.Channel` object. This makes it easier to loop over all channels independently of the number of channels present.
If you need help, I can assist.

mcdo0486:
That's a good point. I've removed the line from the instrument driver. This will fix the issue for users that currently have procedures that call `inst.ChA.shutdown() ` I'll look into migrating the instrument to the channel schema in a different PR.

bmoneke:
Ok, sounds like a plan.

You did not push the change yet.

mcdo0486:
Ok, that is pushed now.

Dan

bmoneke:
Thanks for your bug fixes and changes!

PR #796, opened 2022-12-17
Fix results not adding a data column that isn't in Pint unit registry

mcdo0486:
Currently if a data column includes a unit that isn't in the Pint unit registry, then the procedure will quit with `pint.errors.UndefinedUnitError` exception.

Example: https://gist.github.com/mcdo0486/5694edb7256cf2b93cf8e12ebe630941

This PR adds a try-except block in `Results._parse_columns()` which will log if a unit isn't in the registry, similar to `Results.format()`, but will not crash the program.


bilderbuchi:
Well, shouldn't this fail hard if a unit is not recognized? This indicates something's not right, you wouldn't want to continue then, no? In what situation did you encounter this?

mcdo0486:
Why should it hard fail if the text between parenthesis in a data column is not in the unit registry? That seems to go against the flexibility of pymeasure.

People may put other things besides units in the parenthesis, like a note to self `"Random (note)"`. Or they are recording computed data and put the equation in the column header. The latter is the case a researcher ran into.

bmoneke:
You still can put a note or something else into the header, if you use brackets ("[]") or something else. We chose parenthesis (according to Si) as units indicator.

We use parenthesis for units definition (I forgot where we wrote it), therefore an invalid units definition should fail in order to allow fixing it.

Instead of killing the program, we could show an error message, but the meaurement should not start, I think.

bilderbuchi:
@bmoneke agreed! We should check if the documentation on the column name format is clear enough.

bmoneke:
> @bmoneke agreed! We should check if the [documentation](https://pymeasure.readthedocs.io/en/latest/tutorial/procedure.html#modifying-our-script) on the column name format is clear enough. Probably we need to improve on that.

@bilderbuchi I checked the documentation and I think it is well written. However, the place might not be the best. We can add a remark regarding the units of columns in the [first mention of procedures](https://pymeasure.readthedocs.io/en/latest/tutorial/procedure.html#running-procedures) (maybe linking to the paragraph you cited) and, even more important, add a corresponding API documentation entry.

mcdo0486:
I understand that the `()` _should_ contain a Pint unit, but I'm not sure why if `()` doesn't contain a Pint unit.registry we can't continue with a logged error and writing the column. Why should it be a show stopper? How would it negatively effect the data in the results file? 

This PR follows the behavior of `Results.format()` where if a column is not parsed as a valid Pint unit, then we simply just write the value. By including the try-except pattern in `Results._parse_columns()`, the data column is not included in the `Results.units` dictionary, and doesn't get formatted as a Pint unit in `Results.format()`

If we are going to have an opinionated format for data columns then it should be validated and caught as early as possible throughout the framework, for instance adding a validation step for procedure data columns in `ManagedBaseWindow` init.

bilderbuchi:
> If we are going to have an opinionated format for data columns then it should be validated and caught as early as possible throughout the framework, for instance adding a validation step for procedure data columns in `ManagedBaseWindow` init.

This sounds like a good idea!

IMO, our data is (by a large majority) physical quantities, which have a name, a value and (optionally) a unit. The header format we use ensure that we can simply and unambiguously determine the name and a unit, if present: There's a name, and if it's followed by parentheses, these contain the units.

An additional thing is, as you're aware, we want to move in the direction of a more flexible data/communication setup (#676), and for this a simpler specification with fewer moving parts (more "opinionated" as you put it) helps to keep the complexity of designing a generic interface down. 

Note: One could say that we should even go one step further and follow the NIST recommendation of always doing captions for value data as `cell_temperature/(mK)` to categorically avoid misunderstandings, but IIRC we decided against that for practicality reasons.

mcdo0486:
The effect of Pint has been invisible to me because every pymeasure instrument driver I've used has formatted and returned a float for a data point, which [bypasses all of the Pint logic in `CSVFormatter.format()`](https://github.com/pymeasure/pymeasure/blob/20fe65307432ce4156ea115a0a9c43d5c77431fa/pymeasure/experiment/results.py#L179).

It is a good point that following an external standard, like NIST SI units, can help future-proof protocol compatibility to a degree. I'll rework this PR to validate the header format in ManagedWindowBase and raise an appropriate exception that provides more context to the user. I think hitting the "Queue" button, triggering the `Procedure.__init__` is too late. I'll also reorder the conditional clauses in `CSVFormatter.format()` so the overhead of Pint is lessened, as that function gets hit for datapoint in a record.

bmoneke:
Thanks for improving the code base.

mcdo0486:
Just pushed changes to match our discussion:

Moved CSVFormatter.parsed_columns() static method to Procedure class.
Expand docstring for parsed_columns() to outline pymeasure column header format.
Add raise exception with more context for users.
Add validation step in Procedure.get_measurement() and ManagedWindowBase.__init__ to confirm procedure DATA_COLUMNS follow pymeasure column header format.

mcdo0486:
Also optimized ` CSVFormatter.format()` logic to bias floats. From tests of running the example Procedure.

From tests of the tutorial `RandomProcedure` run without the GUI with 100000 iterations, this optmization saves on average .3 seconds.

pymeasure/experiment/procedure.py
```diff
@@ -75,6 +78,33 @@ def __init__(self, **kwargs):
                 log.info(f'Setting parameter {key} to {kwargs[key]}')
         self.gen_measurement()
 
+    @staticmethod
+    def parse_columns(columns):
```
bmoneke, line 82: I don't know whether I wrote tests for that method or not.
Please write tests (or adjust them) for this method.
As a static method, writing tests is straight forward: verify that a certain input produces the expected output. (pytest.mark.parametrize is helpful)

mcdo0486: Yea, I just caught that and pushed a change, I should have run pytest first.

Updated tests based on changes.
Modify numeric check in `Procedure.parsed_columns()` from `isinstance()` to `type()` check, because `isinstance(bool, int)` returns `True`. Did a quick test and getting the same time improvements.

tests/experiment/test_procedure.py
```diff
@@ -88,8 +93,18 @@ class TestProcedure(Procedure):
...
     assert p.prop == 5
+
+
+@pytest.mark.parametrize("header, units", (
+        ("x (m)", ureg.m),
+        ("x (m/s)", ureg.m / ureg.s),
+        ("x (V/(m*s))", ureg.V / ureg.m / ureg.s),
+))
+def test_csv_formatter_parse_columns(header, units):
+    assert Procedure.parse_columns([header])[header] == ureg.Quantity(1, units)
```
bmoneke, line 110: I don't know, whether you plan to do it: You can also test, that wrong inputs raise an error, for example
~~~python
with pytest.raises(DivideByZeroError):
    1/0
~~~
Will pass the test. The test will fail, if the exception is not raised!

A "match" parameter can verify, that a certain string is part of the error message.

mcdo0486: Just added two tests, one for valid headers with no units and one for invalid units in header.

tests/experiment/test_procedure.py
```diff
@@ -88,8 +93,34 @@ class TestProcedure(Procedure):
...
+def test_procedure_no_parsed_units(valid_header_no_unit):
+    assert Procedure.parse_columns(valid_header_no_unit) == {}
+
+
+@pytest.mark.parametrize("invalid_header_unit", (
+        ["x (sqrt)"], ["x (x)"], ["x (y)"],
+))
+def test_procedure_invalid_parsed_unit(invalid_header_unit):
+    with pytest.raises(ValueError):
+        Procedure.parse_columns(invalid_header_unit)
```
bilderbuchi, line 126: The fact that `"x ( notes )"` pass (without parsed units) and `"x (sqrt)"` fail/raise could be useful for your researcher wanting to put non-unit content into parentheses?

bilderbuchi:
What I've seen so far looks good to me. Thank you for adding tests!

tests/experiment/test_procedure.py
bmoneke: Thanks

tests/experiment/test_procedure.py
mcdo0486: Yup exactly, that's why I put them in as tests to call out how notes or equations can be put into a DATA_COLUMNS header.

pymeasure/experiment/results.py
```diff
@@ -142,10 +144,9 @@ def format(self, record):
         :return: a string
         """
         line = []
-        numeric_types = (float, int, Decimal)
         for x in self.columns:
             value = record.get(x, float("nan"))
-            if type(value) in numeric_types:
+            if type(value) in self.numeric_types:
```
bilderbuchi, line 149: Why not `isinstance(value, numbers.Number)` instead of dragging around a class attribute?

mcdo0486: Good idea, fixed in the last commit.

mcdo0486: @bilderbuchi it's been a while since this PR, I just rediscovered why I use `type(value) in self.numeric_types`.

`isinstance(True, numbers.Number)` evaluates to `True`, which means if a boolean is passed it would be interpreted as a data point 1. [This test]( https://github.com/mcdo0486/pymeasure/blob/6b33d02d69ae6cbe526456b928caa1324132581a/tests/experiment/test_results.py#L95) catches that case.

bilderbuchi: Ok, sorry, I didn't pick that up.

bmoneke:
If you're ready to merge, just tell me.

bmoneke:
Ok

PR #798, opened 2022-12-18
Support for Racal-Dana 1992 universal counter

tomverbeure:
This PR adds fairly extensive (but not yet complete) support for the old but somewhat popular Racal-Dana 1992 frequency counter.

While it's functional (and being used by myself) at this time, it's a draft, not ready to be merged.

There are a few topics I'd like to bring up:

* The 1992 has 3 inputs, but they are not independent channels: there's only 1 measurement head. Sometimes, a measurement uses 1 input (e.g. freq counting, period measurement), sometimes it uses 2 (phase calculation, ratio calculation etc.). The pymeasure Channel infrastructure is IMO too heavy for this use case, so I'm not using it. The way I handle channels in this driver is lightweight and clean.
* Some config parameters can be read, most can not. I'm not using the standard `.control` and `.measurement` way to access them because I didn't find a way to use them in combination with `read_bytes`, which is needed for the 1992. See next:
* The 1992 doesn't use EOI when it returns data. Because of this, the only way with pyvisa to get data back is to use `read_bytes(21)`. `read` and `read_raw` doesn't work: it will always result in a timeout.
* The 1992 also doesn't use a read-return mechanism to return measurement data. Instead it continuously fills up ring buffer when it has new data. 
* I had to import `read_stb()` from `pyvisa` into the Adapter class. This is because you have to poll the `STB` register to check that there's data available (otherwise you get a timeout on `read_bytes`). (You could also use `wait_for_srq()`, but the Linux version of PyVisa doesn't support that.)
* The code is already flake8 clean, except for this:

    ```python
    channel_params = {
        'A' : {
            'coupling'      : { 'AC'   : 'AAC',  'DC'     : 'ADC' },
            'attenuation'   : { 'X1'   : 'AAD',  'X10'    : 'AAE' },
            'trigger'       : { 'auto' : 'AAU',  'manual' : 'AMN' },
            'impedance'     : { '50'   : 'ALI',  '1M'     : 'AHI' },
            'slope'         : { 'pos'  : 'APS',  'neg'    : 'ANS' },
            'filtering'     : { True   : 'AFE',  False    : 'AFD' },
            'trigger_level' : None,   # Special case, see code
            },
        'B' : {
            'coupling'      : { 'AC'   : 'BAC',  'DC'     : 'BDC' },
            'attenuation'   : { 'X1'   : 'BAD',  'X10'    : 'BAE' },
            'trigger'       : { 'auto' : 'BAU',  'manual' : 'BMN' },
            'impedance'     : { '50'   : 'BLI',  '1M'     : 'BHI' },
            'slope'         : { 'pos'  : 'BPS',  'neg'    : 'BNS' },
            'trigger_level' : None,
            },
    }

    operating_modes = {
        'frequency_a'     : 'FA',
        'period_a'        : 'PA',
        'self_check'      : 'CK',
        'phase_a_to_b'    : 'PH',
        'ratio_a_to_b'    : 'RA',
        'ratio_c_to_b'    : 'RC',
        'interval_a_to_b' : 'TI',
        'total_a_by_b'    : 'TA',
    }
    ```
    If that kind of formatting is not allowed, I will not continue to polish this driver for inclusion. Let me know if that's ok, so that we don't waste time on this. But if it's ok, let me know how to disable this the formatting check for this.
* A comprehensive driver has a lot of options. I don't mind writing a bunch of tests, but I hope it's OK to not test some esoteric stuff that I can't really verify on the bench first. Right now, there are tons of instruments in pyinstrument with no tests at all, it should be ok to have a new driver with a good amount of tests even if coverage isn't 100%?

Let me know.

Thanks!
Tom


bmoneke (commented):
Thanks for this new instrument and especially for the reasoning why you did, what you did. I left a few comments, which also address a few of your questions.

First a request: Please add the instrument to the documentation as well, see [adding documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#adding-documentation).



> The 1992 has 3 inputs, but they are not independent channels: there's only 1 measurement head. Sometimes, a measurement uses 1 input (e.g. freq counting, period measurement), sometimes it uses 2 (phase calculation, ratio calculation etc.). The pymeasure Channel infrastructure is IMO too heavy for this use case, so I'm not using it. The way I handle channels in this driver is lightweight and clean.

I did not see yet, how the channels work exactly. But it might be ok.

> Some config parameters can be read, most can not. I'm not using the standard .control and .measurement way to access them because I didn't find a way to use them in combination with read_bytes, which is needed for the 1992. See next:

> The 1992 doesn't use EOI when it returns data. Because of this, the only way with pyvisa to get data back is to use read_bytes(21). read and read_raw doesn't work: it will always result in a timeout.

See my comment regarding `read`. If you need help implementing it, don't hesitate to ask.

> I had to import read_stb() from pyvisa into the Adapter class. This is because you have to poll the STB register to check that there's data available (otherwise you get a timeout on read_bytes). (You could also use wait_for_srq(), but the Linux version of PyVisa doesn't support that.)

see my comment.

> The code is already flake8 clean, except for this:

It seems fine with me, personally. How to mark it to skip, I don't know yet. I have to research it.

> A comprehensive driver has a lot of options. I don't mind writing a bunch of tests, but I hope it's OK to not test some esoteric stuff that I can't really verify on the bench first. Right now, there are tons of instruments in pyinstrument with no tests at all, it should be ok to have a new driver with a good amount of tests even if coverage isn't 100%?

Tests are a new feature and obviously we did not write tests for all the instruments already in the library. I ask you to write at least a few tests, because it helps a lot, if there are at least a few. Better if they cover the most important parts. You do not have to test everything.

If you want, I have a test generator in my fork, see #752 . With that, you can observe your communication and write tests automatically.
That is a great feature to have, before you rework your code with `control`s, as tests are really helpful in finding bugs during reworking of code, which already works.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,269 @@
...
+
+        self.write(' ' + Racal1992.operating_modes[mode])
+        pass
+
+    # ============================================================
+    # Wait for measurement value
+    # ============================================================
+    def wait_for_measurement(self):
+        while (self.adapter.read_stb() & 0x10) == 0:
+            print(".", end='', flush=True)
```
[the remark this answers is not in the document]
tomverbeure: Added logging, as well as a progressDots parameter to enable or disable them.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,269 @@
...
+            raise Exception(f"{mode} is not a valid operating mode")
+
+        self.write(' ' + Racal1992.operating_modes[mode])
+        pass
+
+    # ============================================================
+    # Wait for measurement value
+    # ============================================================
+    def wait_for_measurement(self):
+        while (self.adapter.read_stb() & 0x10) == 0:
```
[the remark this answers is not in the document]
tomverbeure: I've reverted the read_stb() commit.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,269 @@
...
+    )
+
+    def __init__(self, adapter, **kwargs):
+        kwargs.setdefault('write_termination', '\r\n')
+
+        super().__init__(
+            adapter,
+            "Racal-Dana 1992",
+            **kwargs
+        )
```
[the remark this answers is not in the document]
tomverbeure: The Racal 1992 supports a whole bunch of different terminations. But there's also a very weird modal issue that can be worked-around by using a different write termination. See my notes in [this blog post](https://tomverbeure.github.io/2022/12/18/Setting-Up-a-GPIB-Communication-with-a-Racal-Dana-1992-Counter.html#weird-modal-syntax-error-behavior-with-3-character-commands).
So it definitely makes sense to make the write_termination programmable.

I've also switched to the new name convention.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,269 @@
...
+            settings_str += " " + command
+
+        self.write(settings_str + trigger_str)
+
+    # ============================================================
+    # IP - Instrument Preset
+    # ============================================================
+    def preset(self):
+        self.write(' IP')
+        pass
```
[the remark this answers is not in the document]
tomverbeure: `pass` is just a way to explicitly indicate the end of a function that doesn't return a parameter. I'll remove it.

pymeasure/adapters/visa.py
```diff
@@ -218,6 +218,11 @@ def wait_for_srq(self, timeout=25, delay=0.1):
         """
         self.connection.wait_for_srq(timeout * 1000)
 
+    def read_stb(self):
+        """ Read the instrument Servive Request Status register
+        """
+        return self.connection.read_stb()
+
```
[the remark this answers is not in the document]
tomverbeure: Fixed.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,269 @@
...
+        else:
+            raise Exception("Unsupported return type")
+
+    def set_param(self, param, value=None):
+        self.write(f" S{param} {value}")
+
+    def fetch_param(self, param):
+        self.write(' R' + param)
+        self.wait_for()
+        return self.read_and_decode(allowed_types=param)
```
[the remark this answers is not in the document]
tomverbeure: Fetch_param is gone.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,269 @@
...
+
+        if allowed_types and val_type not in allowed_types:
+            raise Exception("Unexpected value type returned")
+
+        if val_type in Racal1992.int_types:
+            return int(val)
+        elif val_type in Racal1992.float_types:
+            return val
+        else:
+            raise Exception("Unsupported return type")
```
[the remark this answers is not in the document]
tomverbeure: I've converted most access methods to control/measurement/setting.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,269 @@
...
+    .. code-block:: python
+
+        from pymeasure.instruments.racal import Racal1992
+        counter = Racal1992("GPIB0::10")
+
+    """
+
+    resolution = Instrument.control(
+            None,
+            " SRS %d",
```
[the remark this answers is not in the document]
tomverbeure: Done.

tomverbeure:
@bmoneke I think all issues have been resolved. Let me know if there's anything that you'd like to see fixed.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,327 @@
...
+    def __init__(self, adapter, name="Racal-Dana 1992", **kwargs):
+        kwargs.setdefault('write_termination', '\r\n')
+
+        super().__init__(
+            adapter,
+            name,
+            **kwargs
+        )
+
+    def read(self):
```
msmttchr, line 226: @tomverbeure, I have read your blog about the weird problem that happened in your instrument, very strange indeed. Although I don't have a possible explanation, I would suggest to re-implement the `write` method such that you can add a space in front of any command to write. This would avoid to disseminate spaces in all the command definition and, if one day the explanation is found, you can easily change back as appropriate

tomverbeure: I've refactored all the leading spaces into a new `write` method.

bmoneke (changes requested):
Thanks again for your contribution and the many tests.
If it is no draft anymore, you can mark this or as "ready for review" (which will be necessary before merging).

The code looks already good, I have only a few minor points. Especially with the usage of `Exception`: Please choose more appropriate Exception types than that most geral base class. I made some suggestions.

I did not fully understand, how that reading of measurement data works, maybe you can add a comment and maybe expand your example in the docstring.

If you celebrate it, I whish you a merry Christmas!

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,336 @@
...
+
+        if allowed_types and val_type not in allowed_types:
+            raise Exception("Unexpected value type returned")
+
+        if val_type in Racal1992.int_types:
+            return int(val)
+        elif val_type in Racal1992.float_types:
+            return val
+        else:
+            raise Exception("Unsupported return type")
```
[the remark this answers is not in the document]
tomverbeure: * I don't really understand the value of adding `@staticmethod`.  I mean, it's a class method either way. But I don't mind adding it either.
* Comment moved.
* Ok.
* The returned string is always formatted as a float, even if the underlying value is an int. I could change it to `int(float(v))`, but I don't think that's any clearer... If I do `int(v))` on the string directly, Python3 bails like this:`invalid literal for int() with base 10: '+00000001992.E+00'`

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,336 @@
...
+            "SRS %d",
+            """ Control the resolution of the counter with an integer from 3 to 10 that
+            specifies the number of significant digits. """,
+            get_process=(lambda v: Racal1992.decode(v, "RS"))
+        )
+
+    delay_enable = Instrument.setting(
+            "D%s",
+            """ Enable or disable delay. True=enable, False=disable""",
+            set_process=(lambda enable: {True: "E", False: "D"}[enable])
```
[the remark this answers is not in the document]
tomverbeure: Done.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,336 @@
...
+            "SF%s",
+            """ Enable or disable special function. True=enable, False=disable""",
+            set_process=(lambda enable: {True: "E", False: "D"}[enable])
+        )
+
+    # FIXME: not tested on real instrument!
+    special_function_number = Instrument.control(
+            "RSF",
+            "S%d",
+            """ Control special function.""",
```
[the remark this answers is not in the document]
tomverbeure: Yes. It's mentioned extensively in the manual and self-evident.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,336 @@
...
+            adapter,
+            name,
+            **kwargs
+        )
+
+    def read(self):
+        return self.read_bytes(21).decode('utf-8')
+
+    def write(self, s):
+        # Space added in front of all commands that are sent to the
```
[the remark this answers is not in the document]
tomverbeure: I choose between docstring and comment based on whether or not I consider the comment to be worthy of the API documentation. But I can change it to all docstring if you wish...

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,336 @@
...
+        """
+        if channel_name not in Racal1992.channel_params:
+            raise Exception("Channel name must by 'A' or 'B'")
+
+        commands = []
+        trigger_str = ""
+
+        for setting, value in settings.items():
+            if setting not in Racal1992.channel_params[channel_name]:
+                raise Exception(f"Channel {channel_name} does not support a {setting} setting")
```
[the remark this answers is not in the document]
tomverbeure: Ok.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,336 @@
...
+                raise Exception("Timeout while waiting for measurement")
+
+        return stb
+
+    # ============================================================
+    # Measured value
+    # ============================================================
+    @property
+    def measured_value(self):
+        """Measured value."""
```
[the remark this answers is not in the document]
tomverbeure: Documentation added.

tests/instruments/racal/test_racal1992.py
```diff
@@ -0,0 +1,339 @@
...
+
+def test_self_check():
+    with expected_protocol(
+            Racal1992,
+            [
+                (' CK', 'CK+010.00000000E+06'),
+            ],
+    ) as instr:
+        instr.operating_mode = 'self_check'
+        assert instr.measured_value == 10000000.0
```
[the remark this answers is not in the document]
tomverbeure: Changing the operating mode doesn't result in a guaranteed response. But once an operating mode is set, the instrument will continuously store measurement data in its response data FIFO. In the case of self_check, the instrument is configured with a 10MHz input that measures frequency. If the resolution is set to 9 digits or less, this will result in a new measurement at least every second, so reading a value will already result in a measured_value.
I don't think anything needs to change here.

tomverbeure:
The PR should be ready for merging now.

Tom

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,360 @@
...
+        This here, for example, is math constant Z: MZ+001.00000000E+00
+        """
+        if len(v) != 19:
+            raise ReturnValueError("Length of instrument response must always be 19 characters")
+
+        val_type = v[0:2]
+        val = float(v[2:19])
+
+        if allowed_types and val_type not in allowed_types:
+            raise ValueError("Unexpected value type returned")
```
bilderbuchi, line 100: Just a remark/suggestion: In general when throwing exceptions, I try to give the user information that could be useful for subsequent troubleshooting, e.g. the value/object that violated some expectation. E.g. here
```suggestion
            raise ValueError(f"Unexpected value type returned: '{val_type}'")
```
or above the response that is not 19 chars long, etc.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,360 @@
...
+        The FIFO buffer gets cleared when a command is received that requires
+        an immediate reply, such reading a setting. It also gets cleared when
+        an operating mode is cleared.
+
+        When there is no measurement data, this property will stall until data
+        is avaiable. It was also timeout after a time that can be set with the standard
+        pyvisa API.
+
+        One can make sure that measurement data is available by furst calling
+        `wait_for_measurement()`.
```
bilderbuchi, line 358:
```suggestion
        When there is no measurement data, this property will stall until data
        is available. It will also timeout after a time that can be set with the standard
        pyvisa API.

        One can make sure that measurement data is available by first calling
        `wait_for_measurement()`.
```

pymeasure/instruments/racal/racal1992.py
```diff
@@ -80,26 +84,27 @@ class Racal1992(Instrument):
...
-    # All values returned follow the same format: 2 letters to indicate
-    # the type of the value returned, followed by a floating point number
-    # (which could be an integer, of course.)
-    # This here, for example, is math constant Z: MZ+001.00000000E+00
     def decode(v, allowed_types=None):
+        """All values returned follow the same format: 2 letters to indicate
+        the type of the value returned, followed by a floating point number
+        (which could be an integer, of course.)
+        This here, for example, is math constant Z: MZ+001.00000000E+00
+        """
```
bmoneke, line 92: To follow python standards, a docstring should start with an imperative first clause:
```suggestion
        """Decode received message.

        All values returned follow the same format: 2 letters to indicate
        the type of the value returned, followed by a floating point number
        (which could be an integer, of course.)
        This here, for example, is math constant Z: MZ+001.00000000E+00
        """
```

pymeasure/instruments/racal/racal1992.py
```diff
@@ -227,10 +235,10 @@ def read(self):
...
 
     def write(self, s):
-        # Space added in front of all commands that are sent to the
-        # instrument to work around weird model issue. It shouldn't
-        # be needed on almost all devices, but it also doesn't hurt. And it
-        # fixes a real issue that's seen on a few device.
+        """Space added in front of all commands that are sent to the
+        instrument to work around weird model issue. It shouldn't
+        be needed on almost all devices, but it also doesn't hurt. And it
+        fixes a real issue that's seen on a few device."""
```
bmoneke, line 241: Also here, a first, imperative clause:
```suggestion
        """Add a space in front of all commands that are sent to the
        instrument to work around weird model issue.

        It shouldn't be needed on almost all devices, but it also doesn't hurt.
        And it fixes a real issue that's seen on a few devices."""
```

bmoneke (approved):
Just two small remarks regarding docstring style.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,363 @@
...
+        'period_a'        : 'PA',    # noqa
+        'phase_a_rel_b'   : 'PH',    # noqa
+        'ratio_a_to_b'    : 'RA',    # noqa
+        'ratio_c_to_b'    : 'RC',    # noqa
+        'interval_a_to_b' : 'TI',    # noqa
+        'total_a_by_b'    : 'TA',    # noqa
+        'frequency_c'     : 'FC',    # noqa
+    }
+
+    def decode(v, allowed_types=None):
```
bmoneke, line 87: Please mark this method as static method, as it does not require instance or class:
```suggestion
    @staticmethod
    def decode(v, allowed_types=None):
```

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,363 @@
...
+        )
+
+    trigger_level_b = Instrument.control(
+            "RLB",
+            "SLB %f",
+            """ Control trigger level for channel B""",
+            get_process=(lambda v: Racal1992.decode(v, "LB"))
+        )
+
+    def __init__(self, adapter, name="Racal-Dana 1992", **kwargs):
```
bmoneke, line 227: Please move the init to the top of the class, in order to find it more easily.

bmoneke (approved):
Sorry for the last minute changes (that happens, when you have to use a phone for code review...)

The code looks good. Thank you for that new instrument (and small typo fix). 

@bilderbuchi and @msmttchr can this be merged? 

@tomverbeure are you finished with the instrument (after the two changes), such that it may be merged after the two changes?

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,363 @@
...
+                'frequency_a',
+                'period_a',
+                'phase_a_rel_b',
+                'ratio_a_to_b',
+                'ratio_c_to_b',
+                'interval_a_to_b',
+                'total_a_by_b',
+                'frequency_c'
+            """,
+            set_process=(lambda mode: Racal1992.operating_modes[mode]),
```
msmttchr, line 126:
```suggestion
            values=Racal1992.operating_modes,
            map_values=True,
```
I think this is a better fit for your needs and it should be equivalent to the `set_process` parameters.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,363 @@
...
+        )
+
+    resolution = Instrument.control(
+            "RRS",
+            "SRS %d",
+            """ Control the resolution of the counter with an integer from 3 to 10 that
+            specifies the number of significant digits. """,
+            get_process=(lambda v: Racal1992.decode(v, "RS"))
```
msmttchr, line 134:
```suggestion
            get_process=Racal1992.decode("RS")
```
I think you could use this syntax, if you change `decode` in a function factory based on the original `decode` function, something like:
```
def decode(allowed_types=None):
   return partial(former_decode(allowed_types=allowed_types)
```
where `former_decode` is your original decode static method.
You need also to include partial as follow:
`from functools import partial`

Anyway, not a strong suggestion, I would leave to your decision whether you like it and it's worth the effort.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,363 @@
...
+        kwargs.setdefault('write_termination', '\r\n')
+
+        super().__init__(
+            adapter,
+            name,
+            **kwargs
+        )
+
+    def read(self):
+        return self.read_bytes(21).decode('utf-8')
```
msmttchr, line 237: `utf-8` decoding will fail for every byte value>=128.
Are you sure all the bytes returned are < 128?
If not `latin-1` decoding should do the job.

pymeasure/instruments/racal/racal1992.py
```diff
@@ -0,0 +1,363 @@
...
+    def write(self, s):
+        """Add a space in front of all commands that are sent to the
+        instrument to work around weird model issue.
+
+        It shouldn't be needed on almost all devices, but it also doesn't hurt.
+        And it fixes a real issue that's seen on a few devices."""
+        super().write(' ' + s)
+
+    def read_and_decode(self, allowed_types=None):
+        v = self.read_bytes(21).decode('utf-8')
```
msmttchr, line 248:
```suggestion
        v = self.read()
```

msmttchr (commented):
@tomverbeure some feedback from my side.

pymeasure/instruments/racal/racal1992.py
tomverbeure: Done.

pymeasure/instruments/racal/racal1992.py
tomverbeure: Done.

pymeasure/instruments/racal/racal1992.py
tomverbeure: I don't understand how that works, but I tried your changes. It resulted in an error. 
When choosing between making this fix or not submitting this driver (I don't want to spend more time on this), I think the first choice is the better option.

pymeasure/instruments/racal/racal1992.py
tomverbeure: Same as the previous comment. The current code works, the suggest change does not. It's an immaterial change.

pymeasure/instruments/racal/racal1992.py
tomverbeure: Yes, I'm sure all bytes values are <128. It's not broken. Let's not fix it.

pymeasure/instruments/racal/racal1992.py
tomverbeure: No. 

As mentioned at the very top of this thread, the Racal1992 doesn't return EOI and always returns 21 bytes. Your suggestion doesn't work.

BenediktBurger (approved):
@tomverbeure may I merge your driver?

tomverbeure:
> @tomverbeure may I merge your driver?

Yes!

BenediktBurger:
Thanks for the driver and patience.

Issue #673, opened 2022-08-07
Drop Python 3.7 support


bilderbuchi:
When Python 3.7 support ends in June 2023, we will drop support, too.
At this point, we can also start using Python 3.8 features. Please collect any changes we should make in this issue.
* [ ] use the `@final` decorator for a stronger hint than a docstring that some methods should not be overridden: https://stackoverflow.com/a/58939735/599884 (Some candidates will probably be in Adapter)

bmoneke:
I got something else for Python>3.8 (not yet for Python3.8!).
Maybe we can mark such parts with "# Python>versionToDrop code", e.g. "# Python>3.7 @final", or in my case "# Python>3.8 str.remove_suffix()".
That way you can search the codebase for this string and find all the places to change.

bilderbuchi:
Agreed, I have also been doing that in the past.

BenediktBurger:
Python 3.7 support is already removed.

Issue #672, opened 2022-08-07
Define a deprecation policy


bilderbuchi:
We should come up with and document a general deprecation policy/timeline, to set expectations for feature deprecations/removal.
When we deprecate features, we should define a reasonable timeframe (wall time or number of versions) for how long deprecation warnings stay in.
For all deprecated features, we need to (should) keep the old code in place, with a `FutureWarning` (preferred as it's not filtered out by default) or a `DeprecationWarning`.

Currently, we should remove all deprecated features before releasing v1.0. 
Later, we will follow semver and need a major release for backwards-incompatible changes.

bmoneke:
@bilderbuchi proposed a "@deprecated" decorator in #717 .

I have some ideas:
That decorator could contain to parameters: Version and the warning message. It could also add the deprecation note (with the warning message as content) to the docstring. How to include it in Sphinx, I do not know.

bmoneke:
A deprecation warning should indicate, what is deprecated and how to fix it.

bmoneke:
Some example deprecation code:
~~~python
from warnings import warn

def deprecated_feature():
    """Do something deprecated.

    .. deprecated:: {next_version}
        Use {replacement} instead.
    """
    warn("{feature} is deprecated, use {replacement} instead", FutureWarning)
~~~

- The Sphinx documentation should contain the next version (from which on it will be deprecated) and should indicate what to do instead. As it is in the documentation of the deprecated feature, the feature does not have to be named.
- Raise a warning with **FutureWarning** as DeprecationWarnings are ignored unless debug options are activated. The warning should contain the deprecated feature as you receive the warning without knowing where it originates from.

bmoneke:
For new Python features (or software version etc.), which can not yet be used, due to support for older python versions,
it is good to mention the last not supported python version.
For example feature X is introduced in Python 3.8, therefore it should be put into a comment `works for python>3.7` or similar, such that it can be found, when python support 3.7 is dropped. See also #673 .

bilderbuchi:
Researching available libraries, https://pypi.org/project/Deprecated/ seems to be a good candidate for an already implemented decorator. It is [possible](https://deprecated.readthedocs.io/en/latest/tutorial.html#deprecation-warning-classes) to change it to emit `FutureWarning` (as we want). It can also automatically add a Sphinx note to the docstring.
TBC how we best decorate attributes/properties with this.

BenediktBurger:
[...]

PR #799, opened 2022-12-19
Update Flake8 version

CasperSchippers:
The current flake8 version is missing line-too-long issues when a line is part of a triple-quoted string.
This is solved for flake8>5.0.0, I propose to update the version to the latest version (6.0.0) and fix the issues that now surface; i.e. the triple-quoted string issues, and "E275 missing whitespace after keyword" issues that are added by the updated pycodestyle dependency of flake8.

### DIFF
--- a/pymeasure/instruments/anaheimautomation/dpseriesmotorcontroller.py
+++ b/pymeasure/instruments/anaheimautomation/dpseriesmotorcontroller.py
@@ -154,8 +154,9 @@ class DPSeriesMotorController(Instrument):
 
     encoder_window = Instrument.control(
         "VEW", "EW%i",
-        """An integer property that represents the allowable error in encoder pulses from the desired
-           position before the encoder auto-correct function runs. This property can be set.""",
+        """An integer property that represents the allowable error in encoder pulses from the
+        desired position before the encoder auto-correct function runs. This property can be set.
+        """,
         validator=truncated_range,
         values=[0, 255],
         cast=int,
@@ -330,7 +331,8 @@ class DPSeriesMotorController(Instrument):
         super().write(cmd_str)
 
     def wait_for_completion(self, interval=0.5):
-        """ Block until the controller is not "busy" (i.e. block until the motor is no longer moving.)
+        """ Block until the controller is not "busy" (i.e. block until the motor is no longer
+        moving.)
 
         :param interval: (float) seconds between queries to the "busy" flag.
         :return: None
--- a/requirements/pymeasure.yml
+++ b/requirements/pymeasure.yml
@@ -16,7 +16,7 @@ dependencies:
   - pytest-qt=3.3.0
   - pytest-runner=5.2
   - pytest=6.2.5
-  - flake8=4.0.1
+  - flake8=6.0.0
   - setuptools_scm # don't pin, to get newest features
   - sphinx=4.3.1
   - sphinx_rtd_theme=1.0.0
--- a/pymeasure/instruments/agilent/agilent34450A.py
+++ b/pymeasure/instruments/agilent/agilent34450A.py
@@ -320,8 +320,8 @@ class Agilent34450A(Instrument):
     ###################
 
     temperature = Instrument.measurement(":READ?",
-                                         """ Reads a temperature measurement in Celsius, based on the
-                                         active :attr:`~.Agilent34450A.mode`. """
+                                         """ Reads a temperature measurement in Celsius, based on
+                                         the active :attr:`~.Agilent34450A.mode`. """
                                          )
 
     #############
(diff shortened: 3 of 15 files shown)


bilderbuchi:
Is there a setting where flake8 can ignore line-too-long if it's in a multiline string? I feel like having to artificially break apart a docstring's _summary line_ because of the line length limitation is weird.

CasperSchippers:
> Is there a setting where flake8 can ignore line-too-long if it's in a _multiline_ string? I feel like having to artificially break apart a docstring's _summary line_ because of the line length limitation is weird.

Hmmm, not as far as I can see at this moment; there is the "max-doc-length" option, but that seems to act on a line in addition to the standard line-length check. Would you prefer not to update the flake8 version?

Apart from that, it is not yet clear to me why it can't install flake8=6.0.0 on the CI (it does use conda-forge, right?)

bilderbuchi:
> Hmmm, not as far as I can see at this moment; there is the "max-doc-length" option, but that seems to act on a line in addition to the standard line-length check. Would you prefer not to update the flake8 version?

We could add `  # noqa` (or even `# noqua <error-code>` to be more specific) after the end of these few strings instead of wrapping the lines?
 
> Apart from that, it is not yet clear to me why it can't install flake8=6.0.0 on the CI (it does use conda-forge, right?)

I don't know either. Maybe a transient problem? I can definitely find 6.0.0 from my Linux laptop.

bilderbuchi:
> Apart from that, it is not yet clear to me why it can't install flake8=6.0.0 on the CI (it does use conda-forge, right?)

I _think_ this could be because of some dependency problem -- flake8 version 6 requires python>=3.8.1, and we still support 3.7, so this would break in any case.

Why the py38 build fails here, I don't know exactly. However, given the first point: do you need flake8 version *6* for this change, or would the last release of the v5 series also be suitable?

CasperSchippers:
> We could add ` # noqa` (or even `# noqua <error-code>` to be more specific) after the end of these few strings instead of wrapping the lines?

Sure, that makes sense; I'll do that on the lines for which it is appropriate.

pymeasure/instruments/anaheimautomation/dpseriesmotorcontroller.py
```diff
@@ -330,7 +331,7 @@ def write(self, command):
         super().write(cmd_str)
 
     def wait_for_completion(self, interval=0.5):
-        """ Block until the controller is not "busy" (i.e. block until the motor is no longer moving.)
+        """ Block until the controller is not "busy" (i.e. block until the motor is no longer moving.)  # noqa: E501
```
bilderbuchi, line 334: AFAICT, this makes `  # noqa: E501` part of the docstring, correct?
Given that `noqa is supposed to be a source _comment_, does it also work if given at the end of the multiline string (i.e. `"""  # noqa: E501`?

requirements/pymeasure.yml
```diff
@@ -16,7 +16,7 @@ dependencies:
   - pytest-qt=3.3.0
   - pytest-runner=5.2
   - pytest=6.2.5
-  - flake8=4.0.1
+  - flake8=5.0.4
```
bilderbuchi, line 19:
```suggestion
  - flake8=5.0.4 # can be bumped to v6 as soon as we drop python 3.7
```
To document the constraint we have, and leave a breadcrumb for things to do when updating the minimum python version.

bmoneke: Let us think, how we put these hints into the code, that we might find them more easily. Silimar to #672
Elsewhere I used "py>3.7" as a code word.

bilderbuchi: TBH, I just "search in all files" for `3.7` and `37`. The first one typically catches all but one or two, false positives should be few.

bmoneke: But still the question remains, whether the last version or the next version is mentioned ("bump version, when we drop 37" or "minimum version 38")

pymeasure/instruments/anaheimautomation/dpseriesmotorcontroller.py
CasperSchippers: That is actually a very good point (that I should've thought about to check, woops), and you're right, the `noqa` does appear in the docstring; it does indeed also work if I place it after the string, so I corrected it.

requirements/pymeasure.yml
CasperSchippers: Good idea. Since it concerns dropping support for a python version, it makes most sense to me to mention the version that is dropped (so 3.7 in this case).

bmoneke: I just saw, that we already talked about that in #673 
Your suggestion is reasonable, I'll add it to #672 

From my side, this conversation can be resolved.

Issue #784, opened 2022-12-08
Channel of Instrument does not use overwritten Instrument methods


dkriegner:
I am implementing another device driver based on pymeasure's `Instrument` and run into an issue.
The device uses messages based on binary code for communication. I have overwritten `Instrument.read/write/values` to work with binary messages. This all works great. For a set of properties, it seemed useful to use the `Channel` framework.

Using `Channel`s, however, seems to not use all the overwritten methods. In particular my custom `values` method is not used. 
I have prepared a minimal working example where the test framework nicely illustrates the problem.

The Instrument code and test file can be found at
https://gist.github.com/dkriegner/7af08dc78fdd1461125e079670675bcf

The test_prop succeeds as expected, while testing the same property via a channel fails because `CommonBase.values` seems to be used (the string conversion causes the problem).

```
    def test_prop_channel():
        """Verify the processing the load capacity property via Channel"""
        with expected_protocol(
            DKTest,
            [(b'A', b'\x01'), ],
        ) as inst:
>           assert inst.ch_1.prop == 1
E           assert "b'\\x01'" == 1
```

Is it wrong to expect that with `Channel`s I can still overwrite the `values` method?

My WIP code for the real device can be found in my [fork](https://github.com/dkriegner/pymeasure/tree/tcpowerconversion_cxn), but its not ready and lots of other things issue the problem there.

bmoneke:
Channels and Instrument both use `values` of their commen superclass `CommonBase`.

Normally all communication problems can be handled in read/write, such that it is not necessary to overload `values`.

Why did you change `values` at all? What do you try to achieve?

dkriegner:
What I am trying to achieve is that `values` does not perform string conversion and split. The string conversion could likely be undone by using `encode` where needed, but I do not see an option to avoid the `split`. 

My instrument is sending numbers (integers) binary encoded. So for certain values the [`split`-function](https://github.com/pymeasure/pymeasure/blob/a33b5188d14a63691f1f5ab1dcb8311a446a77d0/pymeasure/instruments/common_base.py#L308) in `CommonBase` will cause troubles because I can not select a separator which will avoid producing an unwanted split in all cases.

For example communication messages see the [test file ](https://github.com/dkriegner/pymeasure/blob/tcpowerconversion_cxn/tests/instruments/tcpowerconversion/test_cxn.py)

I do not see how this works without overloading `values`?

[...]

Issue #736, opened 2022-11-06
Better way of encapsulating enums


bilderbuchi:
When defining enums that belong to a class (e.g. error codes), it would be clean design if the enums are defined inside the class. However, as our property factories run at import time, not all attributes and scopes are set up yet, so I've often run into problems doing this, which is why I recommend to define them at module level, instead.
This is not a concern when they are used inside methods.

@bmoneke suggested to test if we can actually use them as arguments to the property factories, when they are defined inside the class. Does that work for all aspects (plain args, callables like `cast`, lambdas)? If yes, define some tests confirming the behaviour, and document the improved approach. Consider if we should move existing documents.

_Originally posted by @bmoneke in https://github.com/pymeasure/pymeasure/pull/712#discussion_r1014733992_

bmoneke:
There is another reason for adding them to class:

Let's say we have a control for the waveform of the generator.
In order to set the control, it would be more easy to set it with `inst.waveform = inst.Waveforms.SINE` instead of importing the different enums from the instrument file itself `from pymeasure.instruments.manufacturer.device import *`.

bmoneke:
### Difficulties

I did some tests with enums and found some difficulties:

- Using `cast=SomeEnum` requires the enum to be known before the Instrument is defined. Therefore, it cannot be defined in the Instrument itself, but globally, you can, however use `get_process=lambda v: SomeEnum(v)`.
- `StrEnum` is a python 3.11 feature, therefore `set_process=lambda v: v.value` is necessary for `Enum`s defining strings.
- If you use `set_process=lambda v: v.value` (as necessary for plain `Enum`), you cannot use strings anymore. (if you require an IntFlag, the integers themselves are valid values as well.
- In the getter, IntEnum (and IntFlag) require ints, therefore you have to cast to int before.

### Example

Here is a working example of some enums and an instrument:
~~~
class MODES(enum.Enum):
    VOLTAGE = "MV"
    CURRENT = "MC"

class EnumInstrument(Instrument):

    def __init__(self, adapter, *args, **kwargs):
        super().__init__(adapter, "Test", *args, **kwargs)

    class OPTIONS(enum.IntEnum):
        ONE = 1
        SEVEN = 7

    class STATUS(enum.IntFlag):
        POWER_SHORTAGE = 1
        OVER_CURRENT = 2
        INTERLOCK = 4

    class WAVEFORMS(enum.Enum):
        FUNNY = "FY"
        SINE = "sn"
[...]

Issue #733, opened 2022-11-05
Make properties return enums


bmoneke:
Enumerations (in `enum` module) offer autocompletion for selecting different values (e.g.  Sine or triangle voltage curve) and decoding of a bitwise combined value (e.g. Status byte).

Right now, pymeasure's property factories can accept IntEnum, StrEnum, IntFlag for the setter out of the box, as they are interpreted as Str or int.
The getter, however, cannot natively use an enum to map a return value. You can use `get_process=lambda v: SomeEnum(v)`, but that is cumbersome.

I see different implementations in the getter of control:
1. Leave as is (using get_process). 
2. set the enum as 'values' parameter (we could add an enum validator as well) and add an instance check to the "map_values" routine (right now lists and dictionaries), which is only a small change. This requires to set map_values to True.
4. add an enum parameter which will work like suggestion 2, but without specifying map_values again. Maybe it could do the validator check as well. This requires the most changes to `control`, but gives the smallest invokation (see examples) 

Examples:
```
Status1 = Instrument.measurement("stb?", get_process= lambda v: SomeEnum(v)) 

Status2 = Instrument.measurement("stb?",
    values=SomeEnum, map_values=True, 
    validator=enum_validator) 

Status3 = Instrument.measurement("stb?", enum=SomeEnum) 
```

My favorite is version 2, as it does not introduce a new logic, but uses already known mechanics. Version 3 is the most neat one, but I do not think, that it is worth the new logic with more complexity, especially, as it won't be used that much (and most of the time you have to define several parameters anyway). 

What do you think?

bmoneke:
Implementation for version 2 (extract of control getter and the newly inserted lines) :
```
            if len(vals) == 1:
                value = get_process(vals[0])
                if not map_values:
                    return value
                elif isinstance(values, (list, tuple, range)):
                    return values[int(value)]
                elif isinstance(values, dict):
                    for k, v in values.items():
                        if v == value:
                            return k
                    raise KeyError(f"Value {value} not found in mapped values")
# begin new part
                elif isinstance(values, Enum):
                    return values(value)
# end new part
                else:
                    raise ValueError(
                        'Values of type `{}` are not allowed '
                        'for Instrument.control'.format(type(values))
                    )
```

bilderbuchi:
I think the values/mapping code is already quite complex and hard to reason about (cf. that one instrument PR that wants to add multi-map capability), I'm hesitant to make that even more complex.
[...]

PR #800, opened 2022-12-20
add T&C Power Conversion RF power supply

dkriegner:
This adds a new Instrument. An RF power supply for magnetron sputtering. 

This instrument is largely based on binary communication, which makes overwriting `read/write/values` methods necessary. Every command sent to the instrument is acknowledged, and the response messages are quite structured: they use some header and checksum fields and do not use any terminator. The tests give a flavor of the kind of communication.

Additionally, due to the issue outlined in #784 I split `values` into two functions which allows using the `Channels` with the overwritten `values` method of the parent `Instrument`. This approach was suggested in #784.

### DIFF
--- a/pymeasure/instruments/tcpowerconversion/tccxn.py
+++ b/pymeasure/instruments/tcpowerconversion/tccxn.py
@@ -32,6 +32,11 @@ def values(self, command, cast=int, separator=',', preprocess_reply=None, **kwar
     """Write a command to the instrument and return a list of formatted
     values from the result.
 
+    This is derived from CommonBase.values and adapted here for use with bytes
+    communication messages (no str conversion and strip). It is implemented as a
+    general method to allow using it equally in PresetChannel and CXN. See
+    Github issue #784 for details.
+
     :param command: SCPI command to be sent to the instrument
     :param separator: A separator character to split the string into a list
     :param cast: A type to cast the result
@@ -57,10 +62,7 @@ def values(self, command, cast=int, separator=',', preprocess_reply=None, **kwar
 
 
 def int2char(value):
-    """
-    converts an integer (unsigned-16bit) to a tuple containing the values
-    of two representing characters.
-    """
+    """Convert an 16-bit unsigned integer to a tuple of two representing characters."""
     return tuple(int(b) for b in value.to_bytes(2, "big"))
 
 
@@ -69,7 +71,7 @@ class PresetChannel(Channel):
 
     load_capacity = Instrument.control(
         "GU\x00{ch:c}\x00\x00", "TD{ch:c}\x01\x00%c",
-        """ percentage of full-scale value of the load capacity preset """,
+        """Control the percentage of full-scale value of the load capacity preset.""",
         preprocess_reply=lambda d: struct.unpack(">H", d[2:4]),
         validator=strict_discrete_set,
         values=range(101),
@@ -77,7 +79,7 @@ class PresetChannel(Channel):
 
     tune_capacity = Instrument.control(
         "GU\x00{ch:c}\x00\x00", "TD{ch:c}\x02\x00%c",
-        """ percentage of full-scale value of the tune capacity preset """,
+        """Control the percentage of full-scale value of the tune capacity preset.""",
         preprocess_reply=lambda d: struct.unpack(">H", d[4:6]),
         validator=strict_discrete_set,
         values=range(101),
@@ -114,10 +116,13 @@ class CXN(Instrument):
     :param string name: Name of the instrument.
     :param kwargs: Any valid key-word argument for Instrument
 
-    Note: In order to enable setting any parameters one has to request control
-    and periodically (at least once per 2s) poll any value from the device.
-    Failure to do so will mean loss of control and the device will reset
-    certain parameters (setpoint, disable RF, ...)
+    .. Note::
+
+        In order to enable setting any parameters one has to request control
+        and periodically (at least once per 2s) poll any value from the device.
+        Failure to do so will mean loss of control and the device will reset
+        certain parameters (setpoint, disable RF, ...). If no value should be polled
+        but control should remain active one can also use the ping method.
     """
     # use predefined values method to allow reusing in the Channels
     values = values
@@ -126,38 +131,52 @@ class CXN(Instrument):
 
     def __init__(self, adapter, name="T&C RF sputtering power supply", address=0, **kwargs):
         self.address = address
-        kwargs.setdefault("asrl", dict(baud_rate=38400))
         super().__init__(adapter,
                          name,
                          includeSCPI=False,
                          write_termination="",
                          read_termination="",
+                         asrl=dict(baud_rate=38400),
                          **kwargs)
 
     @staticmethod
     def _checksum(msg):
-        """
-        returns 2 bytes checksum calculated by a bytewise sum
+        """Calculate a 2 bytes checksum calculated by a bytewise sum of the message.
 
-        Parameters
-        ----------
-        msg : bytes object corresponding to the message
+        :param bytes msg: message content
+        :returns: calculated checksum
+        :rtype: bytes
         """
         return struct.pack(">H", sum(msg))
 
     def _prepend_cmdheader(self, cmd):
-        """
-        prepends command start byte and address to the command
+        """Prepends command start byte and address to the command.
 
-        Parameters
-        ----------
-        msg : bytes object corresponding to the command id and parameters
+        :param bytes msg: command message
         """
         return b"C" + self.address.to_bytes(1, "big") + cmd
 
-    def read(self):
+    def _check_acknowledgment(self):
+        """Check reply string for acknowledgement byte.
+
+        :raises ValueError: if an invalid an invalid byte is read from the instrument
         """
-        Reads from the instrument and returns the data fields as bytes
+        ret = super().read_bytes(1)
+        if ret == b"*":
+            return
+        # no valid acknowledgement message found
+        raise ValueError(
+            f"invalid reply '{ret}' found in acknowledgement check")
+
+    def read(self):
+        """Reads a response message from the instrument.
+
+        This method determines the length of the message from the automatically
+        by reading the message header and also checks for a correct checksum.
+
+        :returns: the data fields
+        :rtype: bytes
+        :raises ValueError: if a checksum error is detected
         """
         header = super().read_bytes(4)
         # check valid header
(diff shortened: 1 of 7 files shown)


pymeasure/instruments/tcpowerconversion/tccxn.py
```diff
@@ -0,0 +1,392 @@
...
+        """ ramp rate in watts/second""",
+        preprocess_reply=lambda d: struct.unpack(">H", d[2:]),
+        set_process=int2char,
+        validator=strict_discrete_set,
+        values=range(1, 99),
+    )
+
+    manual_mode = Instrument.control(
+        "GT\x00\x00\x00\x00", "TM\x00%c\x00\x00",
+        """ set manual tuner mode """,
```
bmoneke, line 332: You can read it as well, don`t you?
```suggestion
        """Control manual tuner mode.""",
```

pymeasure/instruments/tcpowerconversion/tccxn.py
```diff
@@ -0,0 +1,392 @@
...
+            print("error(CXN): control request denied!")
+
+    def release_control(self):
+        self.write("BC\x00\x00\x00\x00")
+        status = int(struct.unpack(">H", self.read())[0])
+        if status != 0:
+            print("error(CXN): release of control unsuccessful!")
+
+    def ping(self):
+        self.write("BP\x00\x00\x00\x00")
```
bmoneke, line 392: Please provide docstrings for these methods.

bmoneke (changes requested):
A few comments.

Please have a look at your docstrings (Especially starting with "Control", "Measure"/"Get", and "Set" and ending with a period.

Thanks for your contribution.

pymeasure/instruments/tcpowerconversion/tccxn.py
```diff
@@ -0,0 +1,392 @@
...
+    power = Instrument.measurement(
+        "GP\x00\x00\x00\x00",
+        """ Get power readings for forward/reverse/load power in watts """,
+        preprocess_reply=lambda d: struct.unpack(">HHH", d),
+        get_process=lambda d: (float(d[0])/10, float(d[1])/10, float(d[2])/10),
+    )
+
+    status = Instrument.measurement(
+        "GS\x00\x00\x00\x00",
+        """ Get status field, where used bits correspond to:
```
bilderbuchi, line 243: If you want, you could use an `IntFlag` enum to implement these status fields, to get a more user-readable representation back. Check other uses of `IntFlag` in the codebase to see how that's done.

bilderbuchi (commented):
Some feedback from a first pass. I'll check out the issue with `values` separately.

pymeasure/instruments/tcpowerconversion/tccxn.py
bmoneke: There are two issues regarding possible implementations: #733 #736

pymeasure/instruments/tcpowerconversion/tccxn.py
```diff
@@ -0,0 +1,392 @@
...
+    """
+    return tuple(int(b) for b in value.to_bytes(2, "big"))
+
+
+class PresetChannel(Channel):
+    values = values
+
+    load_capacity = Instrument.control(
+        "GU\x00{ch:c}\x00\x00", "TD{ch:c}\x01\x00%c",
+        """ percentage of full-scale value of the load capacity preset """,
```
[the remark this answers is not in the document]
dkriegner: Thanks for you feedback. I have reworked all docstrings and hope they now correspond more to default style.

pymeasure/instruments/tcpowerconversion/tccxn.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def __init__(self, adapter, name="T&C RF sputtering power supply", address=0, **kwargs):
+        self.address = address
+        kwargs.setdefault("asrl", dict(baud_rate=38400))
+        super().__init__(adapter,
+                         name,
+                         includeSCPI=False,
+                         write_termination="",
+                         read_termination="",
+                         **kwargs)
```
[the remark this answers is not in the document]
dkriegner: I agree that it makes sense to hard code the `baud_rate`, but I propose to use `asrl=dict(baud_rate=38400)`. Although the hardware of this device offers only RS232 I for example use it via an RS232 ethernet converter and therefore access it via `"TCPIP:....:SOCKET"`. I think hard coding `baud_rate` as suggested would cause troubles (#334)

So I guess what I am saying is that I think the use of the [Single Interface](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#single-interface) part of the documentation should be discouraged, because problems will arise if a converter of the hardware interface is used.

pymeasure/instruments/tcpowerconversion/tccxn.py
dkriegner: I agree and implemented this below

pymeasure/instruments/tcpowerconversion/tccxn.py
```diff
@@ -0,0 +1,392 @@
...
+        """
+        Writes to the instrument and includes needed command header/address
+
+        :param command: command string to be sent to the instrument
+        """
+        fullcmd = self._prepend_cmdheader(command.encode())
+        super().write_bytes(fullcmd + self._checksum(fullcmd))
+        self.check_acknowledgment()
+
+    def check_acknowledgment(self):
```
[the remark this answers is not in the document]
dkriegner: Thanks for the feedback. Yes this could/should be private. I changed it accordingly.

pymeasure/instruments/tcpowerconversion/tccxn.py
```diff
@@ -0,0 +1,392 @@
...
+
+import struct
+
+from pymeasure.instruments.validators import strict_discrete_set
+from pymeasure.instruments import Channel, Instrument
+
+
+def values(self, command, cast=int, separator=',', preprocess_reply=None, **kwargs):
+    """Write a command to the instrument and return a list of formatted
+    values from the result.
```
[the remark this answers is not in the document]
dkriegner: I extended the docstring. If its not discouraged I would keep there a link github issue #784 where the idea originates from.

pymeasure/instruments/tcpowerconversion/tccxn.py
```diff
@@ -0,0 +1,392 @@
...
+            else:
+                results[i] = cast(result)
+        except Exception:
+            pass  # Keep as bytes
+    return results
+
+
+def int2char(value):
+    """
+    converts an integer (unsigned-16bit) to a tuple containing the values
```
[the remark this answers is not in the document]
dkriegner: Sorry this is likely my bad habit (but not forbidden by PEP257 where it says: The summary line may be on the same line as the opening quotes or on the next line. see this [part of PEP257](https://peps.python.org/pep-0257/#multi-line-docstrings))

I have changed all docstrings to hopefully better correspond to the style you use here. I am more used to numpydoc style which is somehow different...

pymeasure/instruments/tcpowerconversion/tccxn.py
dkriegner: Thanks for this hint. This looks very useful to me. I tried to implement it.

pymeasure/instruments/tcpowerconversion/tccxn.py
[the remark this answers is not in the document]
bmoneke: You can always override these defaults, if you supply an `Adapter` instance to the instrument.
That is the preferred way if you use a converter (similar to the PrologixAdapter).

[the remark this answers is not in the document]
dkriegner: That's true, but it comes with the overhead that the user needs to also know (set him/herself) `read_termination` and `write_termination` which is all not necessary if one uses the [Multiple interfaces](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#multiple-interfaces) approach. Does the multiple interface approach come with any disadvantage?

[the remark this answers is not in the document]
bmoneke: > Does the multiple interface approach come with any disadvantage?

Not really, only a little more direct (you have a direct assignment instead of the `asrl={'baud_rate': 38400}`). Either way is fine for this PR.

I thank you for pointing out the issue of converters. You may open an issue to discuss the discouragement of the single interface, in order to get more opinions and to implement the documentation change.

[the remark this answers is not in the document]
dkriegner: I opened #815 to discuss this further.
I here implemented it now without setdefault.

pymeasure/instruments/tcpowerconversion/tccxn.py
```diff
@@ -0,0 +1,427 @@
...
+                                        prefix="preset_")
+
+    def __init__(self, adapter, name="T&C RF sputtering power supply", address=0, **kwargs):
+        self.address = address
+        super().__init__(adapter,
+                         name,
+                         includeSCPI=False,
+                         write_termination="",
+                         read_termination="",
+                         asrl=dict(baud_rate=38400),
```
msmttchr, line 140: I see `baud_rate=38400` extremely simple, clear and working.
While I found `asrl=dict(baud_rate=38400)` a bit suspicious to the casual reader.
Why would I need to define a python dictionary and assign to `asrl` parameter when I can be deliciously direct and say `baud_rate=38400` ?

bmoneke: We discussed it in conversation: https://github.com/pymeasure/pymeasure/pull/800#discussion_r1055470564

bilderbuchi (approved):
Looks good to me, let's wait for the others to conclude their feedback.

bmoneke:
From my side, it can be merged. @msmttchr could you take a look at your open conversation, please?
I'd say you can merge it, if you agree.

pymeasure/instruments/tcpowerconversion/tccxn.py
msmttchr: For sure, the code proposed it works, but my personal taste is to avoid making code more complex to address extreme cases. In the specific case of ethernet to serial, I am for using a dedicated adapter, but this discussion can continued on #815.

dkriegner: @bmoneke after discussion in #815 seems to be concluded: Can this now be kept as is? 
I guess with the outcome of #815 it could either stay as is or also use `kwargs.setdefault(baud_rate, 38400)` without  `asrl`. 
what is your preferred way?

bmoneke: kwargs.setdefault will have the same effect as setting it directly. Type specific stuff (like baud_rate) should be in the type dict.

dkriegner: Thanks for merging this and sorry to keep commenting on something closed, but if you do not mind I would like to understand what you mean by this:

> kwargs.setdefault will have the same effect as setting it directly. 

This is not the first time you say this, but it seems not to be true. If you use `setdefault` then one can overwrite the value. If I put it directly in the super call  and try to overwrite it will give a TypeError.

```
class testkwarg(dict):
    def __init__(self, **kwargs):
        kwargs.setdefault("a", 2)
        super().__init__(b=2, **kwargs)

testkwargs(a=3)  # works with result {'b': 2, 'a': 3}
testkwargs(b=3)  # raises TypeError
```

bmoneke: I agree that you can override.

I meant, that you setting the baud rate via setdefault will prevent your Ethernet adapters from working, as you cannot "unset" the baud rate.

In this regard, both options are the same.

PR #803, opened 2022-12-22
Drop str conversion in values

bmoneke:
As discussed in #784 , the str conversion in the beginning of `values` is superfluous, as `read` returns strings on default. If you want to work with bytes, this string conversion raises Exceptions, therefore I removed it.

The TC038D relied on this conversion, therefore I fixed it.


bilderbuchi (approved):
LGTM, this makes sense!

PR #810, opened 2022-12-31
34401A driver improvement

CodingMarco:
The previous implementation of the 34401A driver was very basic there were no controls for essential parameters like range, resolution or NPLC. So I've expanded the driver to support all relevant SCPI commands.

### DIFF
--- a/pymeasure/instruments/hp/hp34401A.py
+++ b/pymeasure/instruments/hp/hp34401A.py
@@ -36,6 +36,22 @@ def _deprecation_warning(property_name):
 
 
 class HP34401A(Instrument):
+    """ Represents the HP / Agilent / Keysight 34401A Multimeter and
+    provides a high-level interface for interacting with the instrument.
+
+    .. code-block:: python
+
+        dmm = HP34401A("GPIB::1")
+        dmm.function_ = "DCV"
+        print(dmm.reading)  # -> Single float reading
+
+        dmm.nplc = 0.02
+        dmm.autozero_enabled = False
+        dmm.trigger_count = 100
+        dmm.trigger_delay = "MIN"
+        print(dmm.reading)  # -> Array of 100 very fast readings
+
+    """
     FUNCTIONS = {"DCV": "VOLT", "DCV_RATIO": "VOLT:RAT",
                  "ACV": "VOLT:AC", "DCI": "CURR", "ACI": "CURR:AC",
                  "R2W": "RES", "R4W": "FRES", "FREQ": "FREQ",
--- a/pymeasure/instruments/hp/hp34401A.py
+++ b/pymeasure/instruments/hp/hp34401A.py
@@ -22,18 +22,15 @@
 # THE SOFTWARE.
 #
 
-import logging
+from warnings import warn
 from pymeasure.instruments import Instrument
 from pymeasure.instruments.validators import strict_discrete_set
-
-log = logging.getLogger(__name__)
-log.addHandler(logging.NullHandler())
 
 
 def _deprecation_warning(property_name):
     def func(x):
-        log.warning(f'Deprecated property name "{property_name}", use the "function_" '
-                    'and "reading" properties instead.')
+        warn(f'Deprecated property name "{property_name}", use the "function_" '
+             'and "reading" properties instead.', DeprecationWarning)
         return x
     return func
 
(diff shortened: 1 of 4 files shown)


tests/instruments/hp/test_hp34401a.py
```diff
@@ -0,0 +1,292 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import pytest
+import time
+import logging
+from pymeasure.instruments.hp import HP34401A
+
+pytest.skip('Only work with connected hardware', allow_module_level=True)
```
bmoneke, line 30: We recently added the possibility to call a test file with the resource string as a parameter instead of hard coding it into the test.
Please see the current documentation version on how to implement it.

tests/instruments/hp/test_hp34401a.py
```diff
@@ -0,0 +1,292 @@
+#
+# This file is part of the PyMeasure package.
```
bmoneke, line 2: This file name is reserved for protocol tests, that means tests, which just test that the right commands are sent / interpreted. Please provide a couple of these tests, as they help us a lot maintaining the code base. 

For tests with a device, the file name has to be different.

See the documentation regarding tests for more information: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#writing-tests

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -49,3 +49,229 @@ def __init__(self, adapter, **kwargs):
             asrl={'baud_rate': 9600, 'data_bits': 7, 'parity': 2},
             **kwargs
         )
+
+    function_ = Instrument.control(
+        "FUNC?", "FUNC \"%s\"",
+        """ A string property that controls the measurement function,
```
bmoneke, line 55: Please follow the current style of code, which is not yet implemented in all instruments: https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings

Especially a docstring (python standard) has to be in imperative Form and we (pymeasure) want to start them with "Control", "Set", or "Measure" to indicate the type of property.

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -23,24 +23,24 @@
...
     """ Represents the HP 34401A instrument.
     """
 
-    voltage_dc = Instrument.measurement("MEAS:VOLT:DC? DEF,DEF", "DC voltage, in Volts")
+    FUNCTIONS = {"DCV": "VOLT", "DCV_RATIO": "VOLT:RAT",
+                 "ACV": "VOLT:AC", "DCI": "CURR", "ACI": "CURR:AC",
+                 "R2W": "RES", "R4W": "FRES", "FREQ": "FREQ",
+                 "PERIOD": "PER", "CONTINUITY": "CONT", "DIODE": "DIOD"}
 
-    voltage_ac = Instrument.measurement("MEAS:VOLT:AC? DEF,DEF", "AC voltage, in Volts")
```
bmoneke, line 34: Why do you remove these properties?

Consider to mark them as deprecated in order to keep the instrument working for users, until they adjust their code.

Feel free to ask how to implement depreciation.

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -49,3 +49,229 @@ def __init__(self, adapter, **kwargs):
...
+        "FUNC?", "FUNC \"%s\"",
+        """ A string property that controls the measurement function,
+        which can take the values: DCV, DCV_RATIO, ACV, DCI, ACI,
+        R2W, R4W, FREQ, PERIOD, CONTINUITY, DIODE. """,
+        validator=strict_discrete_set,
+        values=FUNCTIONS,
+        map_values=True,
+        get_process=lambda v: v.strip('"'))
+
+    @property
```
bmoneke, line 63: I think you should be able to make these properties as pymeasure properties.
Maybe you could use the `command_process` parameter to inject the current range prefix into the Command.

Similarly to the properties below.

bmoneke (changes requested):
Thank you very much for your contribution. See a first round of comments. If you have questions, feel free to ask.

tests/instruments/hp/test_hp34401a.py
bilderbuchi: Those docs are here: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#device-tests

CodingMarco: Thanks, done. Cool new feature!

pymeasure/instruments/hp/hp34401A.py
CodingMarco: I improved the docstrings. Are they fine now from your side?

pymeasure/instruments/hp/hp34401A.py
CodingMarco: I use a get_process now which emits a deprecation warning to minimize code duplication. Is it ok like that or do you have another idea?

pymeasure/instruments/hp/hp34401A.py
CodingMarco: The problem with `command_process` is that it has to be completely static which the currently active function is not. I think I had this problem before...
The only solution I know of for this problem is to use custom properties. Do you know of another one?

Of course I could also add pymeasure properties for each combination of function and (range, autorange, resolution, nplc) like in the 34450A driver, but I think this is nonsense. This would add 44 properties with a lot of duplication and you most certainly will always use the one for the currently active function.

bmoneke: > This would add 44 properties with a lot of duplication and you most certainly will always use the one for the currently active function.

No, don't make it that complicated.

I see the point. The properties are not dynamic.

You could use something similar to the channel ids: you add curly braces in the command string and replace them in the write method.
Just an idea.

If we don't find a better solution than python properties, they are the best 😁

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -22,25 +22,34 @@
...
-    """ Represents the HP 34401A instrument.
-    """
-
-    voltage_dc = Instrument.measurement("MEAS:VOLT:DC? DEF,DEF", "DC voltage, in Volts")
 
-    voltage_ac = Instrument.measurement("MEAS:VOLT:AC? DEF,DEF", "AC voltage, in Volts")
+def _deprecation_warning(property_name):
+    def func(x):
+        log.warning(f'Deprecated property name "{property_name}", use the "function_" '
+                    'and "reading" properties instead.')
```
bmoneke, line 36: We really raise a warning (you have to do `from warnings import warn` first), not just a log entry:
```suggestion
        warn((f'Deprecated property name "{property_name}", use the "function_" '
                    'and "reading" properties instead.'), FutureWarning)
```

pymeasure/instruments/hp/hp34401A.py
bmoneke: Yes, they are good. Thanks.

pymeasure/instruments/hp/hp34401A.py
bmoneke: I like that idea to have a warning as a side effect. I marked, how to raise a proper warning.

bmoneke (commented):
I looked at the comments, thanks for improving the code.

pymeasure/instruments/hp/hp34401A.py
CodingMarco: Ok I will do that.

pymeasure/instruments/hp/hp34401A.py
CodingMarco: Wow, why didn't I think of that... I now added the implementation using a customized `write()` method and this actually makes the code much cleaner and 20 lines shorter. This is in fact exactly what I was searching for. So thanks for this suggestion, I will keep it in mind in the future :)

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -298,24 +296,37 @@ def beep(self):
...
+    )
+
+    def write(self, command):
+        """Write a command to the instrument."""
+        if "<function_prefix_for_range>" in command:
+            command = command.replace("<function_prefix_for_range>",
+                                      self._get_function_prefix_for_range())
+        elif "<function>" in command:
+            command = command.replace("<function>", HP34401A.FUNCTIONS[self.function_])
+        super().write(command)
```
bmoneke, line 327: As a suggestion: You could use the format language with curly braces (you have to change your occurences of `<abc>` to `{abc}`) as they indicate normally that you will format something in a string:
```suggestion
        if "{function_prefix_for_range}" in command:
            command = command.format(function_prefix_for_range=self._get_function_prefix_for_range())
        elif "{function}" in command:
            command = command.format(function=HP34401A.FUNCTIONS[self.function_])
        super().write(command)
```

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -22,23 +22,36 @@
...
-log = logging.getLogger(__name__)
-log.addHandler(logging.NullHandler())
-
 
 def _deprecation_warning(property_name):
     def func(x):
-        log.warning(f'Deprecated property name "{property_name}", use the "function_" '
-                    'and "reading" properties instead.')
+        warn(f'Deprecated property name "{property_name}", use the "function_" '
+             'and "reading" properties instead.', DeprecationWarning)
```
bmoneke, line 33: See #672 on how to do proper deprecation:

It should be FutureWarning, not DeprecationWarnung, because users won't get a DeprecationWarning!

Additionally, please add a deprecated docstring (see #672) to the docstrings (you could write a single text and add it to the docstrings of all the deprecated properties.

Could you verify, that it works (run `make html` in the docs folder and open you local documentation)? For comparison, you might look at the [Adapter api documentation](https://pymeasure.readthedocs.io/en/latest/api/adapters.html#adapter-base-class).
Maybe the additional docstring text has to be a triple quoted one. Something like:
~~~python
deprecated_text = """
.. deprecated:: 0.12
    Use the "function_" and "reading" properties instead.
"""

The "deprecated since 0.11" is created from `.. deprecated:: 0.11`, which is mentioned in #672

Finally, please add a remark to the Changelog.

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -85,97 +98,75 @@ def __init__(self, adapter, **kwargs):
         validator=strict_discrete_set,
         values=FUNCTIONS,
         map_values=True,
-        get_process=lambda v: v.strip('"'))
+        get_process=lambda v: v.strip('"')
```
bmoneke, line 101: As a suggestion: If you put the final parenthesis in a new line, it is good to end this line with a comma. That way you can add more parameters without touching the `get_process` line:
```suggestion
        get_process=lambda v: v.strip('"'),
```
Similarly below.

It is totally fine to end all, even the last, line with a comma.

bmoneke (approved):
One important remark regarding deprecation. And two suggestions/hints regarding coding.

bmoneke:
Looks good to me.

If you @CodingMarco don't want to change anything anymore, I'll merge it.

CodingMarco:
I'm finished. You can merge.

PR #812, opened 2023-01-02
Change copyright from 2022 to 2023

LongnoseRob:
This updates all files with with a copyright header from 2022 (and in some cases older years) to 2023.

it was done with this command line `grep -rFl "(c) 2013-2022" * | xargs sed -i -e  's/2022/2023/' `

- [x] `make html` in 'docs': no failure
- [x] ran test locally: only 4 warnings, no errors

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2013-2021 PyMeasure Developers
+Copyright (c) 2013-2023 PyMeasure Developers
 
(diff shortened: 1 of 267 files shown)


bilderbuchi (approved):
LGTM, thanks!

BenediktBurger:
Here a description on how to use sed: https://unix.stackexchange.com/questions/159367/using-sed-to-find-and-replace#159369

PR #813, opened 2023-01-02
Prologix adapter documentation adjusted to recent changes

### LINKED ISSUE #807: Migrating Prologix / AR488 from 0.10. to 0.11. VI_ERROR_RSRC_NFOUND
Hi, I'm having an issue after updating to 0.11, I understand that the implementation of the PrologixAdapter has changed, previously I was creating an adapter with this code:

`adapter = PrologixAdapter(port='/dev/cu.usbserial-AC008NYI',baudrate=115200,serial_timeout=1)`
now this doesn't work anymore and as far as I understand from the tutorial it should be updated to:
`adapter = PrologixAdapter('/dev/cu.usbserial-AC008NYI')`

This doesn't work and cannot find a solution, this is the error that I get:
`---------------------------------------------------------------------------
VisaIOError                               Traceback (most recent call last)
/var/folders/s_/k_qsdmpn09ng7xgzq1lrkjbm0000gn/T/ipykernel_13441/652802221.py in <module>
      6 import math
      7 import time
----> 8 adapter = PrologixAdapter('/dev/cu.usbserial-AC008NYI')
      9 #adapter = PrologixAdapter(port='/dev/cu.usbserial-AC008NYI',baudrate=115200,serial_timeout=1)
     10 sourcemeter = Keithley2000(adapter.gpib(16))

~/opt/anaconda3/lib/python3.9/site-packages/pymeasure/adapters/prologix.py in __init__(self, resource_name, address, rw_delay, serial_timeout, preprocess_reply, **kwargs)
     82             warn("Use 'timeout' in ms instead", FutureWarning)
     83             kwargs['timeout'] = serial_timeout
---> 84         super().__init__(resource_name,
     85                          asrl={
     86                              'timeout': 500,

~/opt/anaconda3/lib/python3.9/site-packages/pymeasure/adapters/visa.py in __init__(self, resource_name, visa_library, preprocess_reply, query_delay, log, **kwargs)
    106 
    107         # Clean up kwargs considering the interface type matching resource_name
--> 108         if_type = self.manager.resource_info(self.resource_name).interface_type
    109         for key in list(kwargs.keys()):  # iterate over a copy of the keys as we modify kwargs
    110             # Remove all interface-specific kwargs:

~/opt/anaconda3/lib/python3.9/site-packages/pyvisa/highlevel.py in resource_info(self, resource_name, extended)
   3172 
   3173         if extended:
-> 3174             ret, err = self.visalib.parse_resource_extended(self.session, resource_name)
   3175         else:
   3176             ret, err = self.visalib.parse_resource(self.session, resource_name)

~/opt/anaconda3/lib/python3.9/site-packages/pyvisa/ctwrapper/functions.py in parse_resource_extended(library, session, resource_name)
   2075     # [ViSession, ViRsrc, ViPUInt16, ViPUInt16, ViAChar, ViAChar, ViAChar]
   2076     # ViRsrc converts from (str, unicode, bytes) to bytes
-> 2077     ret = library.viParseRsrcEx(
   2078         session,
   2079         resource_name,

~/opt/anaconda3/lib/python3.9/site-packages/pyvisa/ctwrapper/highlevel.py in _return_handler(self, ret_value, func, arguments)
    224                 session = None
    225 
--> 226         return self.handle_return_value(session, ret_value)  # type: ignore
    227 
    228     def list_resources(

~/opt/anaconda3/lib/python3.9/site-packages/pyvisa/highlevel.py in handle_return_value(self, session, status_code)
    249 
    250         if rv < 0:
--> 251             raise errors.VisaIOError(rv)
    252 
    253         if rv in self.issue_warning_on:

VisaIOError: VI_ERROR_RSRC_NFOUND (-1073807343): Insufficient location information or the requested device or resource is not present in the system.`

Thanks for the help!

bmoneke:
The Prologix adapter documentation (and tutorial) is adjusted to the recent change of the base adapter (serial to visa).

Closes #807

### DIFF
--- a/docs/tutorial/connecting.rst
+++ b/docs/tutorial/connecting.rst
@@ -79,8 +79,8 @@ Therefore, new :class:`PrologixAdapter <pymeasure.adapters.PrologixAdapter>` ins
     from pymeasure.adapters import PrologixAdapter
 
     adapter = PrologixAdapter('ASRL/dev/ttyUSB0::INSTR', address=7)
-    sourcemeter1 = Keithley2400(adapter)  # at GPIB address 7
-    sourcemeter2 = Keithley2400(adapter.gpib(9))  # at GPIB address 9
+    sourcemeter = Keithley2400(adapter)  # at GPIB address 7
+    multimeter = Keithley2000(adapter.gpib(9))  # at GPIB address 9
 
 Some equipment may require the vxi-11 protocol for communication. An example would be a Agilent E5810B ethernet to GPIB bridge.
 To use this type equipment the python-vxi11 library has to be installed which is part of the extras package requirements. ::
--- a/docs/tutorial/connecting.rst
+++ b/docs/tutorial/connecting.rst
@@ -72,12 +72,15 @@ Instead of passing a string, you could equally pass an adapter object. ::
     adapter = VISAAdapter("GPIB::4")
     sourcemeter = Keithely2400(adapter)
 
-To instead use a Prologix GPIB device connected on :code:`/dev/ttyUSB0` (proper permissions are needed in Linux, see :class:`PrologixAdapter <pymeasure.adapters.PrologixAdapter>`), the adapter is constructed in a similar way. Unlike the VISA adapter which is specific to each instrument, the Prologix adapter can be shared by many instruments. Therefore, they are addressed separately based on the GPIB address number when passing the adapter into the instrument construction. ::
+To instead use a Prologix GPIB device connected on :code:`/dev/ttyUSB0` (proper permissions are needed in Linux, see :class:`PrologixAdapter <pymeasure.adapters.PrologixAdapter>`), the adapter is constructed in a similar way.
+The Prologix adapter can be shared by many instruments.
+Therefore, new :class:`PrologixAdapter <pymeasure.adapters.PrologixAdapter>` instances with different GPIB addresses can be generated from an already existing instance. ::
 
     from pymeasure.adapters import PrologixAdapter
 
-    adapter = PrologixAdapter('/dev/ttyUSB0')
-    sourcemeter = Keithley2400(adapter.gpib(4))
+    adapter = PrologixAdapter('ASRL/dev/ttyUSB0::INSTR', address=7)
+    sourcemeter1 = Keithley2400(adapter)  # at GPIB address 7
+    sourcemeter2 = Keithley2400(adapter.gpib(9))  # at GPIB address 9
 
 Some equipment may require the vxi-11 protocol for communication. An example would be a Agilent E5810B ethernet to GPIB bridge.
 To use this type equipment the python-vxi11 library has to be installed which is part of the extras package requirements. ::
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -29,17 +29,18 @@ from pymeasure.adapters import VISAAdapter
 
 class PrologixAdapter(VISAAdapter):
     """ Encapsulates the additional commands necessary
-    to communicate over a Prologix GPIB-USB Adapter,
+    to communicate over a Prologix GPIB Adapter,
     using the :class:`VISAAdapter`.
 
-    Each PrologixAdapter is constructed based on a serial port or
-    connection and the GPIB address to be communicated to.
-    Serial connection sharing is achieved by using the :meth:`.gpib`
+    Each PrologixAdapter is constructed based on a connection to the Prologix device
+    itself and the GPIB address of the instrument to be communicated to.
+    Connection sharing is achieved by using the :meth:`.gpib`
     method to spawn new PrologixAdapters for different GPIB addresses.
 
     :param resource_name: A
         `VISA resource string <https://pyvisa.readthedocs.io/en/latest/introduction/names.html>`__
-        that identifies the target of the connection.
+        that identifies the connection to the Prologix device itself, for example
+        "ASRL5" for the 5th COM port.
     :param address: Integer GPIB address of the desired instrument.
     :param rw_delay: An optional delay to set between a write and read call for
         slow to respond instruments.
@@ -63,10 +64,10 @@ class PrologixAdapter(VISAAdapter):
     .. code::
 
         adapter = PrologixAdapter("ASRL5::INSTR", 7)
+        sourcemeter = Keithley2400(adapter)  # at GPIB address 7
         # generate another instance with a different GPIB address:
         adapter2 = adapter.gpib(9)
-        sourcemeter1 = Keithley2400(adapter)  # at GPIB address 7
-        sourcemeter2 = Keithley2400(adapter2)  # at GPIB address 9
+        multimeter = Keithley2000(adapter2)  # at GPIB address 9
 
 
     To allow user access to the Prologix adapter in Linux, create the file:
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -29,7 +29,7 @@ from pymeasure.adapters import VISAAdapter
 
 class PrologixAdapter(VISAAdapter):
     """ Encapsulates the additional commands necessary
-    to communicate over a Prologix GPIB Adapter,
+    to communicate over a Prologix GPIB-USB Adapter,
     using the :class:`VISAAdapter`.
 
     Each PrologixAdapter is constructed based on a connection to the Prologix device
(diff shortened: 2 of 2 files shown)


bilderbuchi (approved):
LGTM, but probably @msmttchr is a better judge.

docs/tutorial/connecting.rst
```diff
@@ -72,12 +72,15 @@ Instead of passing a string, you could equally pass an adapter object. ::
...
-To instead use a Prologix GPIB device connected on :code:`/dev/ttyUSB0` (proper permissions are needed in Linux, see :class:`PrologixAdapter <pymeasure.adapters.PrologixAdapter>`), the adapter is constructed in a similar way. Unlike the VISA adapter which is specific to each instrument, the Prologix adapter can be shared by many instruments. Therefore, they are addressed separately based on the GPIB address number when passing the adapter into the instrument construction. ::
+To instead use a Prologix GPIB device connected on :code:`/dev/ttyUSB0` (proper permissions are needed in Linux, see :class:`PrologixAdapter <pymeasure.adapters.PrologixAdapter>`), the adapter is constructed in a similar way.
+The Prologix adapter can be shared by many instruments.
+Therefore, new :class:`PrologixAdapter <pymeasure.adapters.PrologixAdapter>` instances with different GPIB addresses can be generated from an already existing instance. ::
 
     from pymeasure.adapters import PrologixAdapter
 
-    adapter = PrologixAdapter('/dev/ttyUSB0')
-    sourcemeter = Keithley2400(adapter.gpib(4))
+    adapter = PrologixAdapter('ASRL/dev/ttyUSB0::INSTR', address=7)
```
msmttchr, line 81: It's unfortunate that `PyVISA` addressing name requires this `ASRL` string prefixed which creates incompatibility with the past. `PyVISA` documentation is not clear (at least to me) about this naming scheme. I have checked also in Windows and apparently the form `ASRLCOM5` is not valid. I think only option in Windows is `ASRL5`, even though is not guaranteed that `COMn` will be translated to `ASRLn`. A bit of a mess, but none of your fault.

pymeasure/adapters/prologix.py
```diff
@@ -37,8 +37,10 @@ class PrologixAdapter(VISAAdapter):
     Serial connection sharing is achieved by using the :meth:`.gpib`
     method to spawn new PrologixAdapters for different GPIB addresses.
 
-    :param port: The Serial port name or a connection object
-    :param address: Integer GPIB address of the desired instrument
+    :param resource_name: A
+        `VISA resource string <https://pyvisa.readthedocs.io/en/latest/introduction/names.html>`__
+        that identifies the target of the connection.
```
msmttchr, line 42:
```suggestion
        that identifies the serial port to interface the Prologix adapter. This is typically in the form of `ASRLn` with `n` as a positive integer number.
```

I suggest this change to make clear to users that this VISA address has nothing to do with the address of the instrument you would like to control.

pymeasure/adapters/prologix.py
```diff
@@ -37,8 +37,10 @@ class PrologixAdapter(VISAAdapter):
     Serial connection sharing is achieved by using the :meth:`.gpib`
     method to spawn new PrologixAdapters for different GPIB addresses.
 
-    :param port: The Serial port name or a connection object
-    :param address: Integer GPIB address of the desired instrument
+    :param resource_name: A
```
msmttchr, line 40: I would consider to leave the name as `port` or similar to remove ambiguity with resource name term used in `Instrument` class.

pymeasure/adapters/prologix.py
```diff
@@ -54,7 +56,18 @@ class PrologixAdapter(VISAAdapter):
...
+
+    Usage example:
+
+    .. code::
+
+        adapter = PrologixAdapter("ASRL5::INSTR", 7)
+        # generate another instance with a different GPIB address:
+        adapter2 = adapter.gpib(9)
+        sourcemeter1 = Keithley2400(adapter)  # at GPIB address 7
+        sourcemeter2 = Keithley2400(adapter2)  # at GPIB address 9
```
msmttchr, line 69: I would consider using another instrument class for the second instrument at address 9 to avoid giving the impression, to the casual reader, that you can only use multiple instruments of the same class.

msmttchr (commented):
@bmoneke, few minor comments to your attention. Thanks

docs/tutorial/connecting.rst
bmoneke: To make it more complicated, some backends (NI) also accept "COM5", while others (pyvisa) require "ASRL5".

pymeasure/adapters/prologix.py
bmoneke: I took the documentation from the visa Adapter, but yes, it should be more distinguishable.

I don't want to fix it to the serial connection, as there are ethernet versions of the Prologix, if I interpret correctly the results of my research.

pymeasure/adapters/prologix.py
bmoneke: We don't use "resource_name" in the instrument anymore.
On the other hand, the visa Adapter (superclass) does use it and I tried to keep the name scheme of the superclass.

Besides, the connection could be ethernet etc, where resource name is better than port.

pymeasure/adapters/prologix.py
bmoneke: Good suggestion.
I don't know, which device does have gpib, but in the end, it does not matter for the sake of the example.

bmoneke: I found a multimeter, such that you may measure the output 😁

pymeasure/adapters/prologix.py
msmttchr: There's an ethernet version, but this has never been tested/used with this driver. The original driver was based on `SerialAdapter` and the documentation we provide to the user never refers to the ethernet version. I agree that, in the future, it could be nice to add support to ethernet, after we test it with some real hardware.

bmoneke: Please have a look at my modified texts.

msmttchr: I see the changes and they looks good to me, but I am a bit hesitant to accept the changes that remove the references to the USB version of the Prologix adapter, until we have a proof that also other ethernet version is working.

bmoneke: Should we add a line, that only the USB version is tested?

pymeasure/adapters/prologix.py
msmttchr: This an excerpt of Instrument documentation and we use the term resource name (correctly in my opinion)

> When adapter is a string, this is taken as an appropriate resource name. Depending on your installed VISA library, this can be something simple like COM1 or ASRL2, or a more complicated VISA resource name defining the target of your connection.
>
> When adapter is an integer, a GPIB resource name is created based on that. In either case a VISAAdapter is constructed based on that resource name. Keyword arguments can be used to further configure the connection.

> On the other hand, the visa Adapter (superclass) does use it and I tried to keep the name scheme of the superclass.

Indeed that's was my point, the usage intended in `VISAAdapter` is totally different than the one intended in `PrologixAdapter`, but your latest documentation update clarify this.

Thanks.

pymeasure/adapters/prologix.py
msmttchr: This could be ok, but really this code was designed to work with usb version.
I don't have myself an ethernet version to test.
I see it communicates via port 1234 and apparently the command set is similar, if not identical, but here my preference (not strong) is to be conservative, as I understand this is the approach we normally follow, and avoid to expand the set of supported adapters type.

In any case, not a big deal, in my opinion, feel free to choose your preferred approach:
- mention that only usb is tested
- revert changes that remove explicit reference to usb version.

Thanks

bmoneke: I'll add "USB-GPIB adapter", but leave the parameters as they are.

msmttchr (approved):
Thanks.

docs/tutorial/connecting.rst
bilderbuchi: > To make it more complicated, some backends (NI) also accept "COM5", while others (pyvisa) require "ASRL5".

Indeed that's the case. This inconsistency is tracked in the second part of this report: https://github.com/pyvisa/pyvisa-py/issues/230. I don't know if this something from the VISA standard that's not implemented in pyvisa-py, or if the NI library does something "beyond the specification" when accepting `COM5`.

Issue #764, opened 2022-11-28
Transition deprecated implementations like query_delay to proper implementation


bmoneke:
Some instruments use deprecated features. They should be transitioned to proper implementation.

Maintainers feel free to expand the list, if you encounter an instrument. Everyone else, feel free to comment in order to add instruments to that list, if you encounter a FutureWarning about the usage of a deprecated feature.

### Usage of `query_delay`

- [ ] Temptronic.ATSBase

PR #814, opened 2023-01-03
Test added, that an instrument hands kwargs to the adapter.

bmoneke:
I noticed in a PR, that the implementor forgot to add kwargs to the instrument header.

This test verifies for all instruments, that a specific kwarg is handed to the adapter.

### DIFF
--- a/tests/instruments/test_all_instruments.py
+++ b/tests/instruments/test_all_instruments.py
@@ -22,6 +22,7 @@
 # THE SOFTWARE.
 #
 
+import importlib
 import pytest
 from unittest.mock import MagicMock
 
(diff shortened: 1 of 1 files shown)


msmttchr (approved):
Ok for me. Very efficient approach to make reviews easier.

bilderbuchi:
Good change, I like that we can test our API assumptions across all instruments! 
There are [pytest warnings](https://github.com/pymeasure/pymeasure/actions/runs/3827969233/jobs/6513042647#step:8:82) being emitted, can we avoid those? 
(BTW, a minor thing, I just notice that the deprecation messages don't mention _what_ is being deprecated, one has to look at the source)

bmoneke:
Oh, I see the point, that you cannot determine **what** is deprecated. I added a note to #672 

> There are [pytest warnings](https://github.com/pymeasure/pymeasure/actions/runs/3827969233/jobs/6513042647#step:8:82) being emitted, can we avoid those?

We have to update the codebase from using deprecated features
😁 
I'll add them to #764

PR #818, opened 2023-01-04
ProtocolAdapter can return empty string/bytes.

bmoneke:
Previously, the ProtocolAdapter could not return an empty string/bytes, but continued to return the next message.
Some instruments return an empty line (just the terminators), therefore it is necessary.

This PR changes "no buffer" from `b""` to `None`. And adjusts the tests accordingly.

### DIFF
--- a/tests/adapters/test_protocol.py
+++ b/tests/adapters/test_protocol.py
@@ -38,7 +38,8 @@ def adapter():
 @mark.parametrize("input, output", (("superXY", b"superXY"),
                                     ([1, 2, 3, 4], b"\x01\x02\x03\x04"),
                                     (5, b"5"),
-                                    (4.6, b"4.6")
+                                    (4.6, b"4.6"),
+                                    (None, None),
                                     ))
 def test_to_bytes(input, output):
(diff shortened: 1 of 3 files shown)


tests/adapters/test_protocol.py
```diff
@@ -170,11 +177,16 @@ def test_read_without_write(self):
...
-    def test_read_empties_read_buffer(self):
+    def test_read_all_bytes_empties_read_buffer(self):
         a = ProtocolAdapter()
         a._read_buffer = b"jklasdf"
-        a.read_bytes(10)
-        assert a._read_buffer == b""
+        a.read_bytes(7)
+        assert a._read_buffer is None
+
+    def test_read_all_bytes__from_pairs_empties_read_buffer(self):
```
bilderbuchi, line 186: Two consecutive underscores in the test name, probably a typo.

bilderbuchi (approved):
LGTM, found a typo.

tests/adapters/test_protocol.py
bmoneke: Thanks, fixed.

Issue #815, opened 2023-01-04
Discourage the use of direct setting of interface specific properties in Instruments


dkriegner:
Directly setting interface specific communication parameters as suggested in the documentation in the [Single Interface](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#single-interface) section:

e.g.:
```
def __init__(self, adapter, name="Extreme 5000", baud_rate=2400, **kwargs):
    super().__init__(
        adapter,
        name,
        baud_rate=baud_rate,
        **kwargs
    )
```

leads to the disadvantage that this device can only be used with this particular connection. Even if a device might offer only a specific hardware interface many transparent converters exist (e.g. RS232 to Ethernet) which can make the device visible at the control computer under a different interface. 

With the `VisaAdapter` one would only need to change the resource name and a connection would be possible if the syntax proposed in the [multiple interfaces section](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments.html#multiple-interfaces) of the documentation is used.

For the example above, this would mean:

```
def __init__(self, adapter, name="Extreme 5000", baud_rate=2400, **kwargs):
    super().__init__(
        adapter,
        name,
        asrl={'baud_rate': baud_rate},
        **kwargs
    )
```

I therefore would propose to discourage using the "single interface" way and thereby allow the user to use hardware adapters more easily.

In particular for serial device communication I could anyhow only find the [LakeShore421 device](https://github.com/pymeasure/pymeasure/blob/9013d6f6d2c5666063b30d6b0bbe8d34deae859d/pymeasure/instruments/lakeshore/lakeshore421.py#L63) which uses the "single interface"-way

Other solutions:
@bmoneke pointed out in #800 that a user can always specify an Adapter and therefore also the single interface implementation would allow to be used with communication Adapters. While this is true, it requires the user to reapply all other default parameters like `read/write_termination, timeout, ...` which typically are unchanged if a transparent converter is used.

LongnoseRob:
Comment from the side-line:
`pyvisa` manages certain properties on resource-level such as `read\write_termination` these should not be globally overwritten. --> further detail [here](https://pyvisa.readthedocs.io/en/latest/introduction/communication.html#getting-the-instrument-configuration-right)

bilderbuchi:
[...]

PR #819, opened 2023-01-04
Remove Toptica adapter

### LINKED ISSUE #816: Toptica iBeamSmart issue
While some basic functionality is implemented, there is an issue with writing. The laser sends a prompt, to indicate that it is ready for writing as a handshake:

```
CMD>
```

Currently, read is implemented to wait for just `[OK]`.

I would be happy to make a PR, but first need to better understand the pymeasure philosophy. I see a few options where to put this:

- in a State (e.g., ready for write) + 
- in read
- in write

Does anybody have some guidance?

Thanks
Sam

bmoneke:
Remove the Toptica Adapter and make the ibeamSmart depending on the VISAAdapter alone.
Takes hints from original PR #352 and the new issue #816.

The tests require empty strings, which are added to the ProtocolAdapter in #818 

I added a few tests from the examples @dkriegner gave in #352.

Fixes #816


tests/instruments/toptica/test_ibeamsmart.py
```diff
@@ -0,0 +1,74 @@
...
+import pytest
+
+from pymeasure.test import expected_protocol
+
+
+from pymeasure.instruments.toptica import IBeamSmart
+
+
+# TODO verify the talk usual communication
+init_comm = [("echo off", None), ("prom off", None), ("talk usual", ""), (None, "[OK]")]
```
dkriegner, line 34: I guess this initial communication is somehow valid and likely good enough for the tests. In reality I remember that initially the device still sends back something. The exact message depends on the history of commands sent before. After a power cycle when the echo is turned on, I guess it still echos the first command and also still sends back a prompt. If this is the second time the device is opened (no power cycling in between) then the replies to these commands will be the ones also expected below (i.e. empty string + "[OK]")

If ProtocolAdapter implements the `flush_read_buffer` method I would suggest to add some replies here. Otherwise there should maybe be some extended comment here to warn not to adapt this to the real replies.

Maybe @waveman68 can provide the exact communication upon sending these commands?

bmoneke: Thanks for the comment.

Yes, I expected that there is some response, which gets discarded by that flush. As the protocol does not know flush, I dropped the responses.
It's useless to define communication which is flushed anyway (at least for testing the properties).

Good idea to mention it.

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,192 @@
...
+    def __init__(self, adapter, name="Toptica IBeam Smart laser diode", **kwargs):
+        kwargs.setdefault('read_termination', '\r\n')
+        kwargs.setdefault('write_termination', '\r\n')
         super().__init__(
-            TopticaAdapter(adapter, baud_rate, **kwargs),
-            "toptica IBeam Smart laser diode",
+            adapter,
+            name,
             includeSCPI=False,
+            asrl={'baud_rate': 115200},
```
msmttchr, line 55: `aslr` dictionary is probably redundant.

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,192 @@
...
-            validator=strict_discrete_set,
-            values=[True, False],
-            get_process=lambda s: True if s == 'ON' else False,
-            set_process=lambda v: "on" if v else "off",
+        "sta la", "la %s",
+        """Control emission status of the laser diode driver (bool).""",
+        validator=strict_discrete_set,
+        values=[True, False],
+        get_process=lambda s: True if s == 'ON' else False,
+        set_process=lambda v: "on" if v else "off",
```
msmttchr, line 157:
```suggestion
        values={True: "on", False : "off"},
        map_values=True,
```
Is the suggested change is not possible because the instrument is using upper case in sending and lower case in reception ?

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,192 @@
...
-            get_process=lambda s: True if s == 'ON' else False,
-            set_process=lambda v: "en 1" if v else "di 1",
+        "sta ch 1", "%s",
+        """Control status of Channel 1 of the laser (bool).""",
+        validator=strict_discrete_set,
+        values=[True, False],
+        get_process=lambda s: True if s == 'ON' else False,
+        set_process=lambda v: "en 1" if v else "di 1",
+        check_set_errors=True,
     )
```
msmttchr, line 169: Do you think it is feasible to  implement these as `PyMeasure` channel ? Or it is not worth the effort ?

msmttchr (commented):
Few comment from my side. Thanks

pymeasure/instruments/toptica/ibeamsmart.py
bmoneke: I don't know the instrument itself. I just combined the previous "toptica adapter" and the instrument into one. Therefore I did not touch the properties (yet).

pymeasure/instruments/toptica/ibeamsmart.py
bmoneke: I don't know, whether the channels have more functionality.

I actually thought about the channels as well, but I rejected the idea as too much work for a single property.

pymeasure/instruments/toptica/ibeamsmart.py
bmoneke: You mean, I should give the baud rate directly?
As I don't see, where it is redundant (double).

bmoneke:
Thanks @msmttchr for your Tipps.
My possibilities to refractor the instrument are limited, as I do not know it.

msmttchr:
Yes, I meant this. Thanks

Il Dom 8 Gen 2023, 14:28 Benedikt Moneke ***@***.***> ha
scritto:

pymeasure/instruments/toptica/ibeamsmart.py
msmttchr: Ok. Thanks

pymeasure/instruments/toptica/ibeamsmart.py
msmttchr: Ok. If focus is replacing the adapter, probably my comment, if relevant, is not applicable.
I think it's a good habit to keep the focus to avoid PR too large. I would resolve this conversation then.

bmoneke:
@waveman68 could you please test this implementation with your device?

Does anyone of you, @dkriegner or @waveman68 have a manual and could help me with the protocol tests (just give me some examples)?

pymeasure/instruments/toptica/ibeamsmart.py
dkriegner: This proposed change is affected by this #815. The issue described there made me use the dictionary.

pymeasure/instruments/toptica/ibeamsmart.py
dkriegner: I think the comment that this is not possible because of the specifics of the device is correct! There is a difference in sending and receiving (lower/upper case)

dkriegner:
The manual was delivered as printout. I will ask to get it for you. I'll come back to you.

dkriegner:
Thanks for all the work. If you give sufficient time, I  will instruct someone (remotely) to test the new code with the hardware.

bmoneke:
Thanks @dkriegner , it is not urgent. Whenever you (and your acquantance) have time, is fine.

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,267 @@
...
+            **kwargs
         )
+        # configure communication mode: no repeating and no command prompt
+        self.write('echo off')
+        self.write('prom off')
+        time.sleep(0.04)
+        # clear the initial messages from the controller
+        try:
+            self.adapter.flush_read_buffer()
+        except AttributeError:
```
dkriegner, line 108: Here I see a problem: I guess with `ProtocolAdapter` you get an `AttributeError`. But if I use a translator device (TCPIP to RS232 via VisaAdapter) I get an `NotImplementedError`. So I guess one either should make `ProtocolAdapter` also raise the same exception or catch here both types of Exceptions.

```suggestion
        except (AttributeError, NotImplementedError):
```

In case a translator device is used, one would need here also a fallback implementation of the flush method.
e.g. something like this. But I would find this more elegant to be handled on the Adapter level...
```
def _fallback_flush_read_buffer(self):
    timeout = self.adapter.connection.timeout
    self.adapter.connection.timeout = 1
    try:
        while True:
            super().read()
    except VisaIOError:
        pass
    self.adapter.connction.timeout = timeout
```

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,267 @@
...
+            while True:
+                line = super().read()
+                if line == '[OK]':
+                    break
+                msg.append(line)
+        except VisaIOError:
+            reply = '\n'.join(msg)
+            try:
+                self.adapter.connection.flush_read_buffer()
+            except AttributeError:
```
dkriegner, line 145: Same as in line 108. NotImplementedError should likely be caught and a fallback for the not existing flush method would be needed

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,267 @@
...
+        """Check communication after setting a value.
+
+        Checks if the last reply is only '[OK]', otherwise a ValueError is
+        raised and the read buffer is flushed because one has to assume that
+        some communication is out of sync.
+        """
+        reply = self.read()
+        if reply:
+            # anything else than '[OK]'.
+            self.adapter.connection.flush()
```
dkriegner, line 171: The call signature of this function is wrong. 

See pyvisa documentation (https://pyvisa.readthedocs.io/en/latest/api/visalibrarybase.html#pyvisa.highlevel.VisaLibraryBase.flush). The mandatory argument is missing.

But in fact it would be better to replace to with a call to `self.adapter.flush_read_buffer()` (with the same error handling as above)

tests/instruments/toptica/test_ibeamsmart.py
dkriegner: I think we already reached a conclusion here. thanks for new comment. But I can now confirm (after testing with  the hardware) that after a power cycle the first command ("echo off") still is recorded with an echo.

dkriegner (changes requested):
I can confirm that these changes fix issue #816. 

There are some remaining issues which need to be discussed. Some of them only are important if a translator device (see #815 for a more detailed description) are used. But some are also completely generic.

bmoneke:
Thanks for the information regarding the flush.
I don't know yet, how to do it properly (best would be to not use it).

dkriegner:
I think a flush method is very important to be able to clean up after an error. It's a pity that pyvisa does not implement it for all connection methods.
Since even on the [NI webpage](https://knowledge.ni.com/KnowledgeArticleDetails?id=kA00Z0000019YoCSAU) they recommend using read to flush a TCPIP socket in Labview I assume there is a technical reason for that.

If you would not want to implement a fallback, the colleagues will keep some local code changes. In the end this is again only an issue when using a TCPIP translator. When using via the serial device, the existing flush_read_buffer is sufficient and the error handling is only needed for your unittests. But likely this should then be mentioned as comment. One could also mention that because of this using any Adapter defined outside is unlikely to work (even SerialAdapter would likely fail) since the communication is screwed up after the `__init__`

bmoneke:
What do you think of a flush method based on read?
That would be the most complete instrument implementation. And basically, we strive to just use write and read (or their bytes variants) as these are universal to all connection types (and adapters).

dkriegner:
yes, something based on read (repeated read as shown in my [comment above](https://github.com/pymeasure/pymeasure/pull/819#discussion_r1069595173) or something equivalent) with the timeout set to minimum seems sensible. Not sure if the string conversion in read could make troubles and the byte equivalent would not be better ?!

pymeasure/instruments/toptica/ibeamsmart.py
bilderbuchi: > But if I use a translator device (TCPIP to RS232 via VisaAdapter) I get an ` NotImplementedError`

On what, exactly? Can you paste the error? `VisaAdapter` has `flush_read_buffer()`, so I'd be curious what it's tripping ober?

bilderbuchi: > In case a translator device is used, one would need here also a fallback implementation of the flush method.
e.g. something like this. But I would find this more elegant to be handled on the Adapter level...

Could you elaborate here, too? In my copy of pyvisa, `TCPIPInstrument` should get its `flush` from `MessageBasedResource`, so I'm not sure why we need to have a fallback for that? 
It's a bit hard to debug stuff like this from far away -- exact repros and error messages help :sweat_smile:

dkriegner: > > But if I use a translator device (TCPIP to RS232 via VisaAdapter) I get an ` NotImplementedError`
> 
> On what, exactly? Can you paste the error? `VisaAdapter` has `flush_read_buffer()`, so I'd be curious what it's tripping over?

Since this is independent of the device, you can reproduce the error yourself using:
```
import pymeasure.adapters
test = pymeasure.adapters.VISAAdapter("TCPIP::127.0.0.1::8080::SOCKET")
test.flush_read_buffer()
NotImplementedError                       Traceback (most recent call last)
Cell In[4], line 1
----> 1 test.flush_read_buffer()

File ~/tmp/pymeasure/pymeasure/adapters/visa.py:228, in VISAAdapter.flush_read_buffer(self)
    221 def flush_read_buffer(self):
    222     """ Flush and discard the input buffer
    223 
    224     As detailed by pyvisa, discard the read buffer contents and if data was present
    225     in the read buffer and no END-indicator was present, read from the device until
    226     encountering an END indicator (which causes loss of data).
    227     """
--> 228     self.connection.flush(pyvisa.constants.BufferOperation.discard_read_buffer)

File ~/.local/lib/python3.10/site-packages/pyvisa/resources/messagebased.py:797, in MessageBasedResource.flush(self, mask)
    784 def flush(self, mask: constants.BufferOperation) -> None:
    785     """Manually clears the specified buffers.
    786 
    787     Depending on the value of the mask this can cause the buffer data
   (...)
    795 
    796     """
--> 797     self.visalib.flush(self.session, mask)

File ~/.local/lib/python3.10/site-packages/pyvisa_py/highlevel.py:222, in PyVisaLibrary.flush(self, session, mask)
    220 except KeyError:
    221     return self.handle_return_value(session, StatusCode.error_invalid_object)
--> 222 return self.handle_return_value(session, sess.flush(mask))

File ~/.local/lib/python3.10/site-packages/pyvisa_py/sessions.py:439, in Session.flush(self, mask)
    418 def flush(self, mask: constants.BufferOperation) -> StatusCode:
    419     """Flush the specified buffers.
    420 
    421     The buffers can be associated with formatted I/O operations and/or
   (...)
    437 
    438     """
--> 439     raise NotImplementedError

NotImplementedError: 
```

So somehow (at least when using `pyvisa-py`) the flush method seems not to work as you expect. This is in my opinion also the reason why the fallback would be beneficial.

dkriegner: It seems pyvisa-py just implemented flush for the serial device. https://github.com/pyvisa/pyvisa-py/pull/208

bmoneke: Maybe improving pyvisa-py might be better than creating it ourselves?

dkriegner: I agree that having this in pyvisa-py would be the best, but I am not sure if this is gonna happen quickly for many interface types. I can open an issue there, but it seems to me best would be to have a fallback implementation in pymeasure. It should be used only in case of a `NotImplementedError` (or alike) of the underlying code.

pymeasure/instruments/toptica/ibeamsmart.py
bmoneke: Thanks for pointing it out. I changed it. Probably I mixed them up as I never used any of the both before.

pymeasure/instruments/toptica/ibeamsmart.py
bmoneke: We could (and should) add the error handling in the `VISAAdapter.flush_read_buffer` method, not here, until pyvisa-py is updated. Please feel free to open an issue (or directly a PR).

The instrument will work flawlessly without this fix. I'll resolve.

pymeasure/instruments/toptica/ibeamsmart.py
bmoneke: See above.

bmoneke:
I raised an issue #834 for that missing `flush` method in pyvisa-py as it does not pertain to this instrument itself, but to the VISAAdapter (and more properly to pyvisa-py).

From my side, the code is ready and any maintainer may merge it.

dkriegner (approved):

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,267 @@
...
-            "sh temp sys",
-            """ base plate (heatsink) temperature in degree centigrade.""",
+        "sh temp sys",
+        """Measure base plate (heatsink) temperature in degree centigrade.""",
+    )
+
+    current = Instrument.measurement(
+        "sh cur",
+        """Measure the laser diode current in mA.""",
+        # TODO verify the unit
```
bilderbuchi, line 197: This TODO still open?

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,267 @@
...
+        time.sleep(0.04)
+        # clear the initial messages from the controller
+        try:
+            self.adapter.flush_read_buffer()
+        except AttributeError:
+            log.warning("Adapter does not have 'flush_read_buffer' method.")
+        self.ask('talk usual')
+
+    def read(self):
+        """Read a reply of the instrument and exctract the values, if possible.
```
bilderbuchi, line 113:
```suggestion
        """Read a reply of the instrument and extract the values, if possible.
```

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,267 @@
...
-            """ Status of the laser diode driver.
-                This can be True if the laser is on or False otherwise""",
-            validator=strict_discrete_set,
-            values=[True, False],
-            get_process=lambda s: True if s == 'ON' else False,
-            set_process=lambda v: "on" if v else "off",
+        "sta la", "la %s",
+        """Control emission status of the laser diode driver (bool).
+
+        .. deprecated:: 0.12 Use attr:`emission` instead.
```
bilderbuchi, line 214: does this property need a `command_process` argument like the others?

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,267 @@
...
-            """ Status of Channel 1 of the laser.
-                This can be True if the laser is on or False otherwise""",
-            validator=strict_discrete_set,
-            values=[True, False],
-            get_process=lambda s: True if s == 'ON' else False,
-            set_process=lambda v: "en 1" if v else "di 1",
+        "sta ch 1", "%s",
+        """Control status of Channel 1 of the laser (bool).
+
+        .. deprecated:: 0.12 Use :attr:`ch_1.enabled` instead.
```
bilderbuchi, line 227: Can you please add all the deprecations to a changelog entry?

@bmoneke I feel we should encourage this, otherwise it will be very hard to keep track of all freshly added deprecations when assembling the changelog come release time.

pymeasure/instruments/toptica/ibeamsmart.py
bmoneke: Yes, as I do not have the device and cannot verify it. Maybe just dropping the unit?

pymeasure/instruments/toptica/ibeamsmart.py
bmoneke: Added.

pymeasure/instruments/toptica/ibeamsmart.py
bmoneke: Done.
I agree.

pymeasure/instruments/toptica/ibeamsmart.py
dkriegner: I check again the manual: The unit is not specified there. Other current related parameters use mA (but this does not mean we can certain it is mA). Dropping the unit is the safest. I will try to get hardware access another time and check it.

bmoneke: From these parameters I conferred the mA units.
It would be great, if you could check it. Otherwise we can just drop that units.

pymeasure/instruments/toptica/ibeamsmart.py
```diff
@@ -22,99 +22,267 @@
...
+    def __init__(self, adapter, name="Toptica IBeam Smart laser diode", **kwargs):
         super().__init__(
-            TopticaAdapter(adapter, baud_rate, **kwargs),
-            "toptica IBeam Smart laser diode",
+            adapter,
+            name,
             includeSCPI=False,
+            read_termination='\r\n',
+            write_termination='\r\n',
+            asrl={'baud_rate', 115200},
```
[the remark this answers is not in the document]
bmoneke: I put it in the init.
Yes, you could change it via kwargs, but that way we _show_ that you can change it and that you might have to check the value of your instrument.

pymeasure/instruments/toptica/ibeamsmart.py
dkriegner: We checked with the hardware and the unit seems to be mA as you expected from the manual. Of course one can't be 100% certain, but at full laser power the reported current is just few units below the set limit (which is reported in mA). I think the TODO can be removed and the string kept as is

bmoneke:
@bilderbuchi you might merge at will.

bilderbuchi:
Thanks!

PR #822, opened 2023-01-06
Move coordinates label to the pyqtgraph PlotItem

CasperSchippers:
In order to save room in the plotwidget (especially relevant for widgets such as the DockWidget that houses multiple PlotWidgets) I here propose to make the coordinates label (now positioned just below the PlotItem in a separate, otherwise empty, VBoxLayout) a part of the PlotItem.


msmttchr (approved):
I have tried it and it works fine!
Thanks

bmoneke (approved):
Code looks good.

PR #808, opened 2022-12-29
Dev/anritsu-464xB_drivers

ashokbruno:
Drivers for the Anritsu 474xB series of Vector Network Analyzers

Issue #779, opened 2022-12-06
Add collection of instrument solutions to documentation


bmoneke:
@bilderbuchi proposed in #778 to add a list of solutions for instrument implementations to the documentation:
For different problems there are listed instruments, which solve these problems. 

I think the end of the file might be a good place.

bilderbuchi:
A separate file should also be fine -- the `adding_instruments.rst` is slowly growing huge.

To mirror what's in #778, this should be a FAQ-style directory in the documentation where we record weird/special instruments, so that contributors can check their list for implementation inspiration.
I imagine something like
* TC038, x and y implement binary frame-based communication
* aja.DCXS has replies that are all the same lengths
* aja.DCXS has some garbage in the buffer after startup, that needs to be discarded
* XYZ (I forget which) VNA can dynamically configure itself depending on which hardware modules are present
* In the MKS937B, not all the channels have all capabilities
* ...

bilderbuchi:
* `tcpowerconversion.CXN` (#800) needs overridden `values` for both the instrument and its channels, and needs to take care to use the correct `ask` method. A class-level override `values=values` can be used as long as we don't have a better solution (#811).

bmoneke:
- #802 has different channel names on the front panel (A, B, C) and command ids (1, 2, 3).

bmoneke:
- #827 has different Channel property attributes for different subclassed instruments.

PR #827, opened 2023-01-09
Add implementation of Anritsu MS4644B

CasperSchippers:
An implementation of the Anritsu MS464xB Vector Network Analyzer.


pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,792 @@
...
+        number_of_channels = self.number_of_channels
+
+        if len(self.channels) == number_of_channels:
+            return
+
+        # Remove redant channels
+        while len(self.channels) > number_of_channels:
+            self.remove_child(self.channels[len(self.channels)])
+
+        # Remove create new channels
```
bmoneke, line 83:
```suggestion
        # Create new channels
```

bmoneke (commented):
A small comment regarding a comment.

The code looks good (at a first glance).

What are your plans regarding the other PR?

CasperSchippers:
@bmoneke, I have question regarding Channels and dynamic properties:
This series of instruments contains 4 versions, where the main difference lies in the frequency range it reaches. Normally, I would make a class that represents the family, make the frequency-properties dynamic and adjust the ranges for these dynamic properties within subclasses that correspond to one of the members of the family of instruments. However, these properties that should be dynamic now live within a channel, making this a lot more difficult; I now solved this by moving the adjustment of the dynamic properties to the init method, but I'm not sure if that is the best solution, or if you have any other idea regarding such cases.

ashokbruno:
> > What are your plans regarding the other PR?
> 
> That is a good question. I'd also like to hear if @ashokbruno has any thoughts about this. I'm inclined to think that this implementation (especially since I implemented the methods that missed with respect to #808) is a bit more complete than the other; but I definitely do not want to bump @ashokbruno on the head.

@CasperSchippers This implementation is good. Please feel free to use this. I can check with my use case and modify my implementation(as needed) since it is quite new.

bmoneke:
> @bmoneke, I have question regarding Channels and dynamic properties: This series of instruments contains 4 versions, where the main difference lies in the frequency range it reaches. Normally, I would make a class that represents the family, make the frequency-properties dynamic and adjust the ranges for these dynamic properties within subclasses that correspond to one of the members of the family of instruments. However, these properties that should be dynamic now live within a channel, making this a lot more difficult; I now solved this by moving the adjustment of the dynamic properties to the init method, but I'm not sure if that is the best solution, or if you have any other idea regarding such cases.

Good question and therefore a candidate for #779 

One idea, on how to do it, I'm open for suggestions:

- A "BaseChannel" has dynamic properties
- Different Instruments have their own Channel with changed property attributes
- The instruments have a "channel_definition" (or similar) attribute, which contains a reference to the specific channel class, that `update_channels` might use the right definition. (Just as a thought: In a fixed channel number instrument, you would use `channels = ChannelCreator(MyIndividualChannel, (1,2,3))`)

bmoneke:
I like, that your instrument implements all the traces and ports. It looks quite complete.

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,844 @@
...
+        10, 12, or 16; when the system is in 100000 points mode, the system only supports 1 channel.
+        If a value is provided that is not valid in the present mode, the instrument is set to the
+        next higher channel number.
+        """,
+        values=[1, 16],
+        validator=strict_range,
+        cast=int,
+    )
+
+    display_layout = Channel.control(
```
bilderbuchi, line 271: Should be `Instrument.control`, right?

bmoneke: They are actually the same. For future compatibility (if they might diverge), it is better to use the appropriate prefix.

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,844 @@
...
+
+    def __init__(self, adapter, name="Anritsu MS464xB Vector Network Analyzer", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            timeout=10000,
+            **kwargs,
+        )
+
+        self.update_channels(self.CHANNELS[1])
```
bilderbuchi, line 71: IIUC, the number of channels can be (have to be) chosen at runtime because of the way the specific instrument is connected/equipped with hardware modules or somesuch? In that case, maybe add a `channels` argument to the init signature, and use that?

pymeasure/instruments/anritsu/anritsuMS464xB.py
CasperSchippers: That's a good thing to check that I didn't think of yet; thanks

CasperSchippers:
> @CasperSchippers This implementation is good. Please feel free to use this. I can check with my use case and modify my implementation(as needed) since it is quite new.

Thank you @ashokbruno, that is very kind of you!!

> > @bmoneke, I have question regarding Channels and dynamic properties: This series of instruments contains 4 versions, where the main difference lies in the frequency range it reaches. Normally, I would make a class that represents the family, make the frequency-properties dynamic and adjust the ranges for these dynamic properties within subclasses that correspond to one of the members of the family of instruments. However, these properties that should be dynamic now live within a channel, making this a lot more difficult; I now solved this by moving the adjustment of the dynamic properties to the init method, but I'm not sure if that is the best solution, or if you have any other idea regarding such cases.
> 
> Good question and therefore a candidate for #779
> 
> One idea, on how to do it, I'm open for suggestions:
> 
> * A "BaseChannel" has dynamic properties
> * Different Instruments have their own Channel with changed property attributes
> * The instruments have a "channel_definition" (or similar) attribute, which contains a reference to the specific channel class, that `update_channels` might use the right definition. (Just as a thought: In a fixed channel number instrument, you would use `channels = ChannelCreator(MyIndividualChannel, (1,2,3))`)

That is a good suggestion; However, I tried implementing it as such, but came across the issue that I have locked the order of the class definitions by using some of the attributes of the instrument (e.g. TRACES_MAX, FREQUENCY_RANGE, DISPLAY_LAYOUTS) in the property definitions of the channels, and since I'd like to keep al these "definitions" so to speak at a single place, I'm reluctant to change that order.

> I like, that your instrument implements all the traces and ports. It looks quite complete.

Thanks, but I regret to say that this implementation is (very) far from complete, seeing that that the instrument has two programming manuals with in total a command reference of about 1200 pages, to me it feels more like a solid setup than a complete implementation 😓...

CasperSchippers:
Added question regarding the documentation; I felt that adding the full documentation of each of the actual instruments is a bit cluttering, since they only differ in frequency range. Therefore I thought to add the base-class (AnritsuMS464xB) in full, and only add the specific classes (e.g. AnritsuMS4644B) with a reference to the base-class, but without listing all its members. Is this acceptable, or is it preferred to add each of the documentations in full?

bmoneke:
I think it is good as you propose.
If the other instruments had some more functions etc., it would be different. But just another frequency range: One extensive documentation is enough, in my opinion.

bilderbuchi:
> seeing that that the instrument has two programming manuals with in total a command reference of about 1200 pages, to me it feels more like a solid setup than a complete implementation sweat...

Ha, I feel your pain. I once made a driver for a (modbus-interfaced) industrial _stepper motor_, whose PDF manual was 450+ pages :sweat_smile:

bmoneke:
I opened a PR (#853 ) for the solution of different implementation issues (closing #779).

There you could add the issue with the different channels, once the the documentation is online and you have chosen an implementation path.

bmoneke:
Thanks for adding a note.

CasperSchippers:
What do you think, @bilderbuchi and/or @bmoneke, do you think this is ready to merge or not  yet? (not that I want to push/hurry you or anything, just wasn't completely sure on what the status is)

bmoneke:
I did not know, what the status is. I'll have a look.

tests/instruments/anritsu/test_anritsuMS464xB.py
```diff
@@ -0,0 +1,161 @@
...
+         (":CALC1:PAR:COUN?", "8"),
+         (":CALC2:PAR:COUN?", "12")],
+        active_channels="auto",
+        installed_ports="auto",
+        traces_per_channel="auto",
+    ) as instr:
+        assert len(instr.channels) == 2
+        assert len(instr.ch_1.ports) == 3
+        assert len(instr.ch_1.traces) == 8
+        assert len(instr.ch_2.traces) == 12
```
bmoneke, line 67: I'd split it into several tests, such that you know, which one fails:
```suggestion
def test_init():
    with expected_protocol(
        AnritsuMS464xB,
        [],
    ) as instr:
        assert len(instr.channels) == 16
        assert len(instr.ch_1.ports) == 4
        assert len(instr.ch_1.traces) == 16


def test_init_with_different_channel_numbers():
    # Test init with different number of active channels and installed ports
    with expected_protocol(
        AnritsuMS464xB,
        [],
        active_channels=12,
        installed_ports=2,
        traces_per_channel=10
    ) as instr:
        assert len(instr.channels) == 12
        assert len(instr.ch_1.ports) == 2
        assert len(instr.ch_1.traces) == 10


def test_init_unknown_active_channels():
    # Test init with unknown active channels and traces
    with expected_protocol(
        AnritsuMS464xB,
        [(":SYST:PORT:COUN?", "3"),
         (":DISP:COUN?", "2"),
         (":CALC1:PAR:COUN?", "8"),
         (":CALC2:PAR:COUN?", "12")],
        active_channels="auto",
        installed_ports="auto",
        traces_per_channel="auto",
    ) as instr:
        assert len(instr.channels) == 2
        assert len(instr.ch_1.ports) == 3
        assert len(instr.ch_1.traces) == 8
        assert len(instr.ch_2.traces) == 12
```

tests/instruments/anritsu/test_anritsuMS464xB.py
```diff
@@ -0,0 +1,161 @@
...
+        with pytest.raises(ValueError):
+            instr.ch_1.frequency_stop = 6e10
+
+    with expected_protocol(
+        AnritsuMS4647B,
+        [(":SENS1:FREQ:STOP 7e+10", None)],
+    ) as instr:
+        instr.ch_1.frequency_stop = 7e10
+        with pytest.raises(ValueError):
+            instr.ch_1.frequency_stop = 8e10
```
bmoneke, line 161: Here again, I'd split them into several tests.

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+    """ A class representing the Anritsu MS464xB Vector Network Analyzer (VNA) series.
+
+    This familly consists of the MS4642B, MS4644B, MS4645B, and MS4647B, which are represented in
+    their respective classes (:class:`~.AnritsuMS4642B`, :class:`~.AnritsuMS4644B`,
+    :class:`~.AnritsuMS4645B`, :class:`~.AnritsuMS4647B`), that only differ in the available
+    frequency range.
+
+    Can contain up to 16 instances of :class:`~.MeasurementChannel` (depending on the configuration
+    of the instrument), that are accessible via the `channels` dict or directly via `ch_` + the
+    channel number.
```
bmoneke, line 46:
```suggestion
    They can contain up to 16 instances of :class:`~.MeasurementChannel` (depending on the configuration
    of the instrument), that are accessible via the `channels` dict or directly via `ch_` + the
    channel number.
```

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+        super().__init__(
+            adapter,
+            name,
+            timeout=10000,
+            **kwargs,
+        )
+
+        self.PORTS = installed_ports if installed_ports != "auto" else self.number_of_ports
+
+        number_of_channels = active_channels if active_channels != "auto" else None
```
bmoneke, line 91: For me it is easier to read, if you have an equality comparison:
~~~suggestion
        self.PORTS = self.number_of_ports if installed_ports == "auto" else installed_ports

        number_of_channels = None if active_channels == "auto" else active_channels
~~~

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+
+        self.update_channels(number_of_channels=number_of_channels,
+                             traces=traces_per_channel)
+
+    def update_channels(self, number_of_channels=None, **kwargs):
+        """Create or remove channels to be correct with the actual number of channels.
+
+        :param int number_of_channels: optional, if given defines the desired number of channels.
```
bmoneke, line 99:
```suggestion
        :param int number_of_channels: optional, if given, defines the desired number of channels.
```

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+        =====    ===========================================================
+        """,
+        values=[0, 1, 2],
+        validator=strict_discrete_set,
+        cast=int,
+    )
+
+    datafile_frequency_unit = Instrument.control(
+        ":FORM:SNP:FREQ?", ":FORM:SNP:FREQ %s",
+        """Control the frequency unit displayed in an SNP data file.
```
bmoneke, line 156:
```suggestion
        """Control the frequency unit displayed in a SNP data file.
```

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+        ":DISP:SPL?", ":DISP:SPL %s",
+        """Control the channel display layout in a Row-by-Column format.
+
+        Valid values are: {}. The number
+        following the R indicates the number of rows, following the C the number of columns.
+        """.format(", ".join(DISPLAY_LAYOUTS)),
+        values=DISPLAY_LAYOUTS,
+        validator=strict_discrete_set,
+        cast=int,
+    )
```
bmoneke, line 311: Why don't you use a tuple for the values (ok, validation is more difficult)?
```suggestion
    display_layout = Instrument.control(
        ":DISP:SPL?", ":DISP:SPL R%iC%i",
        """Control the channel display layout in a Row-by-Column format.

        Valid values are: {}. The number
        following the R indicates the number of rows, following the C the number of columns.
        """.format(", ".join(DISPLAY_LAYOUTS)),
        cast=int,
    )
```

You cast to int: what does the returned int indicate?

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+        """,
+        values=TRIGGER_TYPES,
+        validator=strict_discrete_set,
+    )
+
+    external_trigger_delay = Instrument.control(
+        ":TRIG:EXT:DEL?", ":TRIG:EXT:DEL %g",
+        """Control the the delay time of the external trigger.
+
+        Valid values are between 0 [s] and 10 [s] in steps of 1e-9 [s] (i.e. 1 ns).
```
bmoneke, line 362:
```suggestion
        """Control the delay time of the external trigger in seconds.

        Valid values are between 0 [s] and 10 [s] in steps of 1e-9 [s] (i.e. 1 ns).
```

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+        """,
+        values=[0, 10],
+        validator=strict_range,
+    )
+
+    external_trigger_edge = Instrument.control(
+        ":TRIG:EXT:EDG?", ":TRIG:EXT:EDG %s",
+        """Control the which edge is used for triggering of the external trigger.
```
bmoneke, line 370:
```suggestion
        """Control the edge type of the external trigger.
```

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+    def trigger(self):
+        """ Triggers a continuous sweep from the remote interface. """
+        self.write("*TRG")
+
+    def trigger_single(self):
+        """ Triggers a single sweep with synchronization from the remote interface. """
+        self.write(":TRIG:SING")
+
+    def trigger_continuous(self):
+        """ Triggers a continuous sweep from the remote interface. """
```
bmoneke, line 414: In imperative mode 😁 :
```suggestion
        """ Trigger a continuous sweep from the remote interface. """
        self.write("*TRG")

    def trigger_single(self):
        """ Trigger a single sweep with synchronization from the remote interface. """
        self.write(":TRIG:SING")

    def trigger_continuous(self):
        """ Trigger a continuous sweep from the remote interface. """
```

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+        HOLD    Hold the sweep on all channels
+        SING    Perform a single sweep and then hold all channels
+        =====   =================================================
+        """,
+        values=["CONT", "HOLD", "SING"],
+        validator=strict_discrete_set,
+    )
+
+    def load_data_file(self, filename):
+        """Load a data file in VNA HDD.
```
bmoneke, line 436:
```suggestion
        """Load a data file into the VNA HDD.
```

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+        validator=strict_range,
+    )
+
+
+class Trace(Channel):
+    """Represents a trace within a :class:`~.MeasurementChannel` of the Anritsu MS464xB VNA. """
+    placeholder = "tr"
+
+    def activate(self):
+        """ Sets the indicated trace as the active one. """
```
bmoneke, line 533:
```suggestion
        """ Set the indicated trace as the active one. """
```

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+
+        # Remove create new channels
+        while len(self.traces) < number_of_traces:
+            self.add_child(Trace, len(self.traces) + 1, collection="traces", prefix="tr_")
+
+    def check_errors(self):
+        return self.parent.check_errors()
+
+    def activate(self):
+        """ Sets the indicated channel as the active channel. """
```
bmoneke, line 631:
```suggestion
        """ Set the indicated channel as the active channel. """
```

pymeasure/instruments/anritsu/anritsuMS464xB.py
```diff
@@ -0,0 +1,912 @@
...
+
+        Valid values are between 1E7 [Hz] (i.e. 10 MHz) and 4E10 [Hz] (i.e. 40 GHz).
+        """,
+        values=AnritsuMS464xB.FREQUENCY_RANGE,
+        validator=strict_range,
+        dynamic=True,
+    )
+
+    def clear_average_count(self):
+        """ Clears and restarts the averaging sweep count of the indicated channel. """
```
bmoneke, line 778:
```suggestion
        """ Clear and restart the averaging sweep count of the indicated channel. """
```

bmoneke (approved):
I have a few docstring suggestions and one possible code suggestion. A docstring's first sentence should start in imperative mode.

Thanks

tests/instruments/anritsu/test_anritsuMS464xB.py
CasperSchippers: That is a good suggestion, fixed!

tests/instruments/anritsu/test_anritsuMS464xB.py
CasperSchippers: Also fixed

pymeasure/instruments/anritsu/anritsuMS464xB.py
CasperSchippers: Corrected and clarified this.

pymeasure/instruments/anritsu/anritsuMS464xB.py
CasperSchippers: A, yeah, that is indeed clearer

pymeasure/instruments/anritsu/anritsuMS464xB.py
CasperSchippers: Regarding the int: good one, this should just be a string output with the same format as the input.

Regarding the not using a tuple: For most values (e.g. `R2C1`) that would indeed be fine (e.g. `(2, 1)`); but there are a couple of situations where this is a bit more difficult (e.g. `R2C2C1` or `C2R2R1`), where you need a more difficult tuple (e.g. `(2, (2, 1))` and `((2, 1), 2)`. For me it is not worth setting up this conversion/validation for a functionality that I do not expect anyone will very often use (especially not something I expect people to change 'dynamically', but rather always apply a specific certain display layout for a specific measurement type).

bmoneke: Oh, I did not know, that you could use more than one R or C (for example your `R2C2C1`).

In that case, I'd leave the code as is (excecpt for a `cast=str`) and maybe one or two examples in the docstring.

bilderbuchi:
I won't have time for a detailed review, but if the PR looks good for you two, that's good enough for me. :+1: :rocket: :smiley:

pymeasure/instruments/anritsu/anritsuMS464xB.py
CasperSchippers: That is a good idea indeed; I've added the examples (and the cast=str) to both `display_layout` properties.

CasperSchippers:
Thanks!

PR #828, opened 2023-01-10
Instrument has a adapter_kwargs dict for the connection settings.

bmoneke:
In order to facilitate the use of manually created adapters (Prologix or other converters, see #815 ), I propose to move all the kwargs for the adapter in a class dictionary `adapter_kwargs`.
This dictionary is updated with the kwargs handed to the instrument (so modification remains possible, as usual).
The dictionary can be, however, used as kwargs for the manually created adapter, for example:
~~~python
adapter = PrologixAdapter("resource_name", address=5, **MyInstrument.adapter_kwargs)
inst = MyInstrument(adapter)
~~~

If this idea is good, I'll update the documentation accordingly (instruments have to be done one by one).

### DIFF
--- a/tests/instruments/test_instrument.py
+++ b/tests/instruments/test_instrument.py
@@ -113,11 +113,39 @@ def test_init_visa(adapter):
 @pytest.mark.xfail()  # I do not know, when this error is raised
 def test_init_visa_fail():
     with pytest.raises(Exception, match="Invalid adapter"):
-        Instrument("abc", "def", visa_library="@xyz")
+        Instrument('ASRL2::INSTR', "def", visa_library="@xyz")
+
+
+def test_init_visa_comm_kwargs():
+    """Test whether the adapter_kwargs are handed to the visa adapter"""
+    class TestInstrument(Instrument):
+        adapter_kwargs = {'kwarg_test': 5}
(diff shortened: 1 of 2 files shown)


tests/instruments/test_instrument.py
```diff
@@ -113,11 +113,39 @@ def test_init_visa(adapter):
...
+
+def test_init_visa_comm_kwargs_replaced():
+    """Test, whether a kwarg parameter replaces the default definition."""
+    class TestInstrument(Instrument):
+        adapter_kwargs = {'baud_rate': "abc"}
+    with pytest.raises(TypeError, match="str"):
+        # ensure, that the parameter is faulty and raises an error, unless changed
+        TestInstrument('ASRL2::INSTR', "def", visa_library='@sim')
+    TestInstrument('ASRL2::INSTR', "def", visa_library='@sim', baud_rate=500)
+    pass  # Test that no error is raised with corrected baud_rate
```
bilderbuchi, line 136: This line should probably assert that the `baud_rate` is really 500 now.

bmoneke: Thanks. That should work even better.

bmoneke:
Idea scrapped in #815

PR #829, opened 2023-01-10
Deprecation warning texts improved

bmoneke:
The adapters' deprecation warnings did not mention, what is deprecated. I fixed it.
See #672


CasperSchippers (commented):
That is indeed an improvement of the deprecation warnings. I was wondering if it would be possible to also add when these features will be removed?

bmoneke:
We have not yet determined, when they will be removed (at the latest before 1.0, I think)

CasperSchippers (approved):
Fair enough, if that is to be for a later update, I approve.

bilderbuchi:
Indeed, we don't have a formal deprecation schedule, yet. 

Thank you, LGTM.

PR #835, opened 2023-01-19
Remove deprecated telnetlib

### LINKED ISSUE #668: Replace deprecated telnetlib in TelnetAdapter
TelnetAdapter uses [telnetlib](https://docs.python.org/3/library/telnetlib.html) which is deprecated since Python 3.11 (see [PEP 598](https://peps.python.org/pep-0594/#telnetlib)). Python 3.12 will be last version to contain telnetlib.

It is not yet urgent, but I want to raise this issue, as I encountered it by chance.

[PEP 598](https://peps.python.org/pep-0594/#telnetlib) also offers some alternatives. Maybe Pyvisa (with TCP raw) might work as well.

Max-Herbold:
An attempt to remove telnetlib.

This uses a raw socket implementation and replaces the read/write behaviour of the deprecated telnetlib adapter.

Fix #668


bmoneke:
@Max-Herbold thanks for tackling that issue.
Do you think it would be possible to base the telnet adapter on the VISAAdapter?
Or maybe even adjusting that single instrument (Attocube ANC300, I think), which uses the Telnetadapter to use PyVISA, such that the Telnetadapter may be dropped altogether?

Max-Herbold:
Originally I had a device (Newport XPS-D motion controller) that communicated with sockets. I switched to using TelnetAdapter (~2021) because it used the same raw communication. On the instrument, the commands are passed directly to a compiled binary on a remote server and appears to call the function represented by the string sent to port 5001. Super weird. I haven't investigated VISA deeply so please correct me if I'm mistaken, I'm under the impressions VISA mangles the data sent/received in some manner.

Previously I had tried using VISA without success, here is an example script I spun up to demonstrate. Let me know if there is some error with the TCPIP address.

`python
from pymeasure.adapters import VISAAdapter


ip, port = "192.168.254.254", 5001
command = "FirmwareVersionGet(char*)"

address = f"TCPIP0::{ip}::{port}::SOCKET"
adapter = VISAAdapter(address)

try:
    r = adapter.ask(command)
    print(r)
except Exception as e:
    print(e)


import socket


s = socket.socket()
s.connect((ip, port))

s.send(command.encode())
r = s.recv(1024)
print(r)
`

result:
`
C:\Python\Python38\lib\site-packages\pymeasure\adapters\visa.py:173: FutureWarning: Deprecated call `Instrument.ask` instead.
  warn("Deprecated call `Instrument.ask` instead.", FutureWarning)
VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
b'0,XPS-D Standard #12846,EndOfAPI
`

Maybe there is not an obvious mistake why this doesn't work, I can dive into the low level communications if this is desired.

bmoneke:
pyvisas `query` command (which you invoked via `ask`) does the same as `write` and `read`. `read` however, reads until it finds a termination character (defaults to"\r\n", which is never sent by your device).

If you want to read all available data, after you wrote a command with `write` (appending thermination characters) or `write_bytes`, you have to call `VISAAdapter.read_bytes(-1)`. If this does not work, you might have to read one byte after another (giving 1 instead of -1 as a parameter).

dkriegner:
> Or maybe even adjusting that single instrument (Attocube ANC300, I think), which uses the Telnetadapter to use PyVISA, such that the Telnetadapter may be dropped altogether?

Since I wrote the Attocube ANC300 some time back I feel responsible to keep it working. I should have hardware access next week and look into reworking that based on PyVisa. I'll open a separate issue then.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +247,65 @@ def stop_all(self):
...
+            write_termination=self.termination_str,
+            **kwargs
+        )
+
+        super().__init__(
+            adapter,
+            "attocube ANC300 Piezo Controller",
+            includeSCPI=False,
+            **kwargs
+        )
```
bmoneke, line 269: Two points:
1. We require instruments to have a "name" parameter in the init signature.
2. You do not have to create the VISAAdapter manually. If you give a resource string to the super().__init__, a VISAAdapter will be created and kwargs are used for that creation.
3. We deprecated global usage of `preprocess_reply`, instead the corresponding code (here `_extract_reply`) can be put into the `read` method.

Here the implementation (except for moving it to the read method).
```suggestion
    def __init__(self, adapter, name="attocube ANC300 Piezo Controller", axisnames="", passwd="", query_delay=0.05, **kwargs):
        self.query_delay = query_delay
        self.termination_str = "\r\n"

        super().__init__(
            adapter,
            name,
            includeSCPI=False,
            read_termination=self.termination_str,
            write_termination=self.termination_str,
            **kwargs
        )
```

Max-Herbold:
Ah! That is such a nice solution. Ended up using the read_terminator in the VISAAdapter to find `,EndOfAPI`. Works very well. Thanks for pointing that out.

@dkriegner I shouldn't have touched anything, I started to mangle things in your driver. I removed the adapter and moved everything up to the driver. read function accesses the adapter which is probably undesirable. In principle, nothing at the communication level changed. (also, multiple values for preprocess_reply in tests idk what happened lol)

There is maybe one edge case, a socket/telnet instantly knows when a read is complete (how?). Whereas VISA looks for a predefined terminator. If the terminator is not found, the adapter times out. If an instrument doesn't supply some terminator one must wait until a timeout to assume the transmission is complete. Waiting for a timeout seems undesirable. How can this be prevented?

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +247,65 @@ def stop_all(self):
             attribute = getattr(self, attr)
             if isinstance(attribute, Axis):
                 attribute.stop()
+
+    def __init__(self, host, axisnames, passwd, query_delay=0.05, **kwargs):
```
bmoneke, line 251: Please add a note to the CHANGES.rst file regarding the change, that the Attocube uses a resource string and not a hostname any more.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +247,65 @@ def stop_all(self):
...
+        kwargs['query_delay'] = query_delay
```
bmoneke, line 252: Giving `query_delay` to the adapter is deprecated, as we implement that waiting in the instrument itself.
so, store it in `self.query_delay = query_delay` and implement a function (that way you have that delay per default):
~~~python
    def wait_for(self, query_delay=0):
        """Wait for some time. Used by 'ask' to wait before reading.

        :param query_delay: Delay between writing and reading in seconds.
        """
        super().wait_for(query_delay or self.query_delay)
~~~

dkriegner:
thanks for the work. I will only look into it next week. I think it has to be checked with the hardware and I should extend the test cases. 

Also I think `Axis` should be derived from [`Channel`](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html#instruments-with-channels).

tests/instruments/attocube/test_anc300.py
```diff
@@ -22,102 +22,43 @@
...
-            ANC300Controller,
-            [(None, b"xy"),
-             (b"passwd\r\n", b'xy\r\nAuthorization success\r\nOK'),
-             (b"echo off\r\n", b"x\r\nOK"),
-             (b"geto 1\r\n", b"5\r\nOK")],
-            axisnames=["a", "b", "c"],
-            passwd="passwd"
-            ) as instr:
-        assert instr.a.output_voltage == 5
+    monkeypatch.setattr(anc300, "VISAAdapter", FakeAdapter)
```
bmoneke, line 36: Now, that we use a "normal" instrument style, `expected_protocol` should work flawlessly, such that line is superfluous (at least if the init accepts an adapter instance as parameter.

bmoneke (changes requested):
Thanks for tackling the issue.
Here a few comments from my side. Mainly regarding changed aspects of pymeasure workings.

As @dkriegner pointed out, the "Axis" could be made a subclass of `instruments.Channel`, but that can be done in another PR.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +247,65 @@ def stop_all(self):
...
+
+        self._axisnames = axisnames
+        for i, axis in enumerate(axisnames):
+            setattr(self, axis, Axis(self, i + 1))
+
+        time.sleep(self.adapter.query_delay)
+        super().read()  # clear messages sent upon opening the connection
+        # send password and check authorization
+        self.write(passwd)
+        time.sleep(self.adapter.query_delay)
```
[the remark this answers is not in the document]
bilderbuchi: Could also use `self.wait_for()`, which uses `self.query_delay` automatically, no?

pymeasure/adapters/telnet.py
```diff
@@ -1,109 +0,0 @@
...
-#
-
-import telnetlib
-import time
-from warnings import warn
-
-from .adapter import Adapter
-
-
-class TelnetAdapter(Adapter):
```
bilderbuchi, line 32: @bmoneke can we replace the `TelnetAdapter` implementation with a deprecation notice and an equivalently functioning/configured `VISAAdapter` subclass? That way people that currently use the `TelnetAdapter` (outside of the ANC300) would not get broken code immediately?

pymeasure/instruments/attocube/anc300.py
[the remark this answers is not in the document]
bmoneke: `wait_for` requires a parameter (per default we do not have self.query defined), therefore `wait_for` has to be added to this instrument with a global variable (code is in another comment).

Yes, with that method defined `self.wait_for()` would be enough (and better).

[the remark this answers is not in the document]
bilderbuchi: Ah, oops, when I looked it up I found the example in the docs, not the implemented method >.<

pymeasure/adapters/telnet.py
bilderbuchi: I meant transitioning the ANC300 to the "new way" as we are doing here, but in parallel deprecate (not just remove) `TelnetAdapter` somewhat nicely for users that might be using it elsewhere currently, by replacing it with an equivalent implementation using `VisaAdapter` (+ a notice).

bilderbuchi: If that's not feasible, then :shrug: we can just remove it (but it would be nice to leave a better hint than an `AttributeError` or an `ImportError` - maybe a barebones class that raises a FutureWarning with info when used?

tests/instruments/attocube/test_anc300.py
```diff
@@ -22,102 +22,43 @@
...
+#     with expected_protocol(
+#             ANC300Controller,
+#             [(None, b"xy"),
+#              (b"passwd\r\n", b'xy\r\nAuthorization success\r\nOK'),
+#              (b"echo off\r\n", b"x\r\nOK"),
+#              (b"stepu 1 15\r\n", b"OK")],
+#             axisnames=["a", "b", "c"],
+#             passwd="passwd"
+#             ) as instr:
+#         instr.a.stepu = 15
```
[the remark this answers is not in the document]
Max-Herbold: Yeah, not sure how this was possible but the error raised in the GitHub action was, `TypeError: Adapter.__init__() got multiple values for argument 'preprocess_reply'` for both tests.

pymeasure/adapters/telnet.py
Max-Herbold: This is a good idea, I might have jumped the gun on just chucking the whole thing. I think reworking the whole thing so it can become deprecated is maybe not valuable. Maybe add a deprecation message indicating to use the VISAAdapter and maybe something more informative maybe linking to the help I was provided here earlier or some similar article.

tests/instruments/attocube/test_anc300.py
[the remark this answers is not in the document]
Max-Herbold: Got it, these tests are functional but might need to verify the line endings in communication here.

pymeasure/instruments/attocube/anc300.py
Max-Herbold: Added this in, thanks for the method.

tests/instruments/attocube/test_anc300.py
Max-Herbold: Fixed, thanks!

pymeasure/instruments/attocube/anc300.py
Max-Herbold: Thanks, added.

pymeasure/adapters/telnet.py
bmoneke: For the ANC300, we can add a "host" kwarg (and make the adapter kwarg optional). If adapter is None, a deprecation warning will be raised (see #672 on how to implement it) and the host kwarg will be used to create the resource string.

> If that's not feasible, then 🤷 we can just remove it (but it would be nice to leave a better hint than an AttributeError or an ImportError - maybe a barebones class that raises a FutureWarning with info when used?

I don't see a problem keeping the TelnetAdapter in the codebase for some time. Therefore, the TelnetAdapter itself should stay in:  at least with a deprecation warning, better as a subclass of VISAAdapter. I don't know, though, whether the rework of the TelnetAdapter (without testing possibilities) is work the effort.

tests/instruments/attocube/test_anc300.py
```diff
@@ -22,102 +22,43 @@
 # THE SOFTWARE.
 #
 
-from unittest import mock
-
 import pytest
 
 from pymeasure.test import expected_protocol
-from pymeasure.instruments.attocube.adapters import AttocubeConsoleAdapter
+from pymeasure.adapters import FakeAdapter
```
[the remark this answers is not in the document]
Max-Herbold: sorted, thank you.

pymeasure/instruments/attocube/anc300.py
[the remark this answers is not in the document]
Max-Herbold: switched to using `self.query_delay` and `self.wait_for()`.

pymeasure/adapters/telnet.py
Max-Herbold: > For the ANC300, we can add a "host" kwarg (and make the adapter kwarg optional). If adapter is None, a deprecation warning will be raised (see #672 on how to implement it) and the host kwarg will be used to create the resource string.

This is done.

> as a subclass of VISAAdapter.
Calling a connection `telnet (socket)` that is actually `VISA` is misleading and could cause confusion about the arguments available and how transmission terminations are handled. It's probably better to leave the TelnetAdapter how it was and remove it before python 3.13 rolls around. If it is later desired to keep it for any reason, switch to some implementation of `SocketAdapter` using sockets.

Max-Herbold: Added this adapter back. Let me know if this is sufficient to resolve.

pymeasure/adapters/telnet.py
```diff
@@ -0,0 +1,114 @@
...
+from .adapter import Adapter
+
+
+class TelnetAdapter(Adapter):
+    """ Adapter class for using the Python telnetlib package to allow
+    communication to instruments
+
+    .. deprecated:: 0.11.2
+        Preparation for telnetlib deprecation. 
+        TelnetAdapter is deprecated, use VISAAdapter instead.
```
bilderbuchi, line 38: It would be great to tell users _how_ to do that -- what do they have to do to a VISAAdapter so that it behaves like a TelnetAdapter?

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +246,109 @@ def stop_all(self):
             attribute = getattr(self, attr)
             if isinstance(attribute, Axis):
                 attribute.stop()
+
+    def __init__(
```
bmoneke, line 250: The init should be the first method in the class in order to find it easily.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +246,109 @@ def stop_all(self):
...
+
+        .. deprecated:: 0.11.2
+            The 'host' argument is deprecated. Use 'adapter' instead.
+
+        :param adapter: The address of the controller
+        :param name: The name of the device, defaults to attocube ANC300 Piezo Controller (optional)
+        :param axisnames: a string of the axis names, e.g. "xyz"
+        :param passwd: the password for the controller
+        :param query_delay: the time to wait between queries to the controller
+        """
```
bmoneke, line 273: The class docstring is shown for the init as well, please move all relevant information from here to the class docstring and do not add any docstring to the init (such that the init gets the class docstring per default).

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +246,109 @@ def stop_all(self):
...
+    def handle_deprecated_host_arg(self, kwargs: dict):
+        """
+        This pulls the host argument from the kwargs and creates a resource string
+        for the VISAAdapter. This is deprecated and separated out to make it easier
+        to remove in the future. This function should be removed and the adapter
+        argument should be made non-optional in the definition of the __init__ function.
+
+        .. deprecated:: 0.11.2
+            The 'host' argument is deprecated.
+        """
```
bmoneke, line 315: We do not use type hinting in pymeasure. You can put that information in the docstring:
":param dict \\**kwargs: Keyword arguments..."

I'd put the deprecated host argument in the class docstring, as that way it is visible to everyone creating an instance of that class.

bmoneke:
Ah, I see you are a first time contributor @Max-Herbold , you may add yourself to the AUTHORS.txt, if you want.

pymeasure/adapters/telnet.py
Max-Herbold: Thanks, this seems appropriate.

```suggestion
        telnetlib is deprecated in Python 3.11 and will be removed in Python 3.13.
        As a result, TelnetAdapter is deprecated, use VISAAdapter instead.
        The VISAAdapter supports TCPIP socket connections. When using the VISAAdapter,
        the `resource_name` argument should be `TCPIP[board]::<host>::<port>::SOCKET`.
        see here, <https://pyvisa.readthedocs.io/en/latest/introduction/names.html>
```

pymeasure/instruments/attocube/anc300.py
Max-Herbold: Thanks, done

pymeasure/instruments/attocube/anc300.py
Max-Herbold: Understood, and done. Thanks.

pymeasure/adapters/telnet.py
bilderbuchi: Ah, that is all? There's no additional configuration flags or somesuch necessary? That's Great!

Max-Herbold: Yeah, the TelnetAdapter previously implemented everything that the telnetlib was missing from Pymeasures aspect. The python telnet client was always very simple (`class telnetlib.Telnet(host=None, port=0[, timeout])`). All the methods of Telnet were simple reads and writes. VISAAdapter handles these. The TelnetAdapter implemented write_termination and read_termination, pyvisa handles this. It's fairly complete.

pymeasure/adapters/telnet.py
```diff
@@ -33,6 +33,13 @@ class TelnetAdapter(Adapter):
     """ Adapter class for using the Python telnetlib package to allow
     communication to instruments
 
+    .. deprecated:: 0.11.2
+        Preparation for telnetlib deprecation.
```
bmoneke, line 37: I'd say for the deprecated telnetlib package or such, that a user knows, that it is deprecated externally.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -211,12 +212,18 @@ def check_errors(self):
 class ANC300Controller(Instrument):
     """ Attocube ANC300 Piezo stage controller with several axes
 
-    :param host: host address of the instrument
+    :param adapter: The VISA resource name of the controller
+        (e.g. "TCPIP::<address>::<port>::SOCKET") or a created Adapter.
```
bmoneke, line 216: Is the port configurable? Below in "handle_host", you give a fixed port. If the port is fixed, you can still leave it as "<port>", but could add another line, that the port is 7230.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -226,18 +233,49 @@ class ANC300Controller(Instrument):
...
         self._axisnames = axisnames
         for i, axis in enumerate(axisnames):
             setattr(self, axis, Axis(self, i + 1))
 
+        self.wait_for()
+        super().read()  # clear messages sent upon opening the connection
+        # send password and check authorization
+        self.write(passwd)
+        self.wait_for()
+        ret: str = super().read()
```
bmoneke, line 271: In pymeasure we do not use type hints:
```suggestion
        ret = super().read()
```

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +293,54 @@ def stop_all(self):
...
+        """
+        r = self._reg_value.search(reply)
+        if r:
+            return r.groups()[0]
+        else:
+            return reply
+
+    def read(self):
+        """Read after setting a value."""
+        raw: str = super().read()
```
bmoneke, line 332:
```suggestion
        raw = super().read()
```

bmoneke (approved):
A few small comments.

I prefer, if the channel rework is done in another PR (noted down in #746).

Could one of you, @Max-Herbold and @dkriegner , test the code with an actual device?
The read termination is "\r\n" and in the tests we have this sequence in the middle of messages. I fear, that it might not work as expected.

pymeasure/instruments/attocube/anc300.py
Max-Herbold: let me know if this is sufficient

Max-Herbold:
> The read termination is "\r\n" and in the tests we have this sequence in the middle of messages. I fear, that it might not work as expected.
True, maybe the read_termination should be `\r\nOK`, then handle splitting results the same way in `read()` and handling timeout errors if some command failed.
Another option is to leave the terminator as it currently is and repeatedly read until a pyvisa timeout error occurs, "OK" is received or "ERROR" is received. The manual references that the timeout for generating an error is ~30ms so setting the VISAAdapter timeout to 50ms in this scenario is probably reliable and still responsive enough.
I don't have access to one of these instruments. There is only so much if it's behaviour I can simulate based on the documentation.
@dkriegner is this something you could handle solving?

dkriegner:
I look into it with hardware hopefully today.
Putting multiple characters as read_termination I fear is not supported by
pyvisa. And yes the communication protocol of this device with the multiple
lines it returns is a bit of a pain.

bmoneke:
> Putting multiple characters as read_termination I fear is not supported by
pyvisa.

"\r\n" are already multiple characters. Pyvisa reads until the very last character of the termination string and compares then the whole string. Therefore the very last character of the termination must not occur more than once ("aaabaaaac" is ok, but "aaaaacc" is not ok).

dkriegner:
I have reworked the Attocube controller with the hardware. I am afraid it did not work for some time already since with the hardware it did not even work with the current pymeasure:master. Also the existing tests did not reflect the real hardware communication. My changes are included in PR https://github.com/Max-Herbold/pymeasure/pull/3

I think they are in principle complete. I also extended the test code. 

Here few comments I still have:
* I am not sure if the `host` keyword argument you show with the deprecation warning makes a lot of sense. Previously `host` was a regular argument so the chance that someones old code works without adapting are very small.
* In my PR I use `adapter.connection.clear()` which should be replaced by `adapter.flush_read_buffer()` ones #836 is merged.
* `TelnetAdapter` works quite different from all other adapters. You can see that I have to make several changes to make the ANC work now with the VisaAdapter. The `TelnetAdapter` was able to return all the message even if it included multiple read_termination characters. Given these differences I think it should be removed soon.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +299,59 @@ def stop_all(self):
...
+            return reply
+
+    def read(self):
+        """Read after setting a value."""
+        lines = []
+        while True:
+            lines.append(super().read())
+            if lines[-1] in ["OK", "ERROR"]:
+                break
+        print(lines)
```
Max-Herbold, line 343:
```suggestion
```

tests/instruments/attocube/test_anc300.py
```diff
@@ -22,102 +22,99 @@
...
-        pass
+# Note: This communication does not contain the first several device
+# responses, as they are ignored due to `adapter.flush_read_buffer()`.
+passwd = "passwd"
+init_comm = [
+    (passwd, "*" * len(passwd)),
+    (None, "Authorization success"),
+    ("echo off", "> echo off"),
+    (None, "OK"),
+    ]
```
Max-Herbold, line 38:
```suggestion


```

Max-Herbold (commented):
linting

Max-Herbold:
> I think they are in principle complete. I also extended the test code.

I'm very impressed at how clean that was solved. Thanks for fixing this!

pymeasure/instruments/attocube/anc300.py
```diff
@@ -226,18 +237,51 @@ class ANC300Controller(Instrument):
...
+
         self._axisnames = axisnames
         for i, axis in enumerate(axisnames):
             setattr(self, axis, Axis(self, i + 1))
 
+        self.wait_for()
+        # clear messages sent upon opening the connection,
+        # this contains some non-ascii characters!
+        self.adapter.connection.clear()
+        # should be self.adapter.flush_read_buffer() after PR 836
```
bmoneke, line 274: Yes, we should do it in the order #837 , #836 and then this here. Therefore, let us keep all three PRs in a mergeable state:
```suggestion
        self.adapter.flush_read_buffer()
```

bmoneke (approved):
Looks good to me.
Let us prepare this PR for merge, as if the other two were already merged, such that we can merge all three in one go. I marked the change to do.

> I am not sure if the host keyword argument you show with the deprecation warning makes a lot of sense.

Do you have a better suggestion? I agree, that it is not very probable, but with that comment (and code) in place, a user will see the change more easily.

Thanks @dkriegner and @Max-Herbold for that work.

pymeasure/instruments/attocube/anc300.py
bilderbuchi: Please retrigger CI after merging the respective "preceding" PRs, so that we know the tests pass. It won't rerun automatically after merging another PR afaik.

bmoneke: Ok. How can I retrigger CI?

bilderbuchi: Look at the details of one of the jobs below, there should be a "Re-run all jobs" button on the top right.

bilderbuchi: oh wait. I actually don't know if that computes the sha's to use anew, or just runs the same tests with the same merge commit from before...

bilderbuchi: Ah, in principle I think in the PR management section below, there _should_ appear a "Update branch" button if it detects new changes in the base branch.

bmoneke: Thanks

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +299,58 @@ def stop_all(self):
...
+        """Read after setting a value."""
+        lines = []
+        while True:
+            lines.append(super().read())
+            if lines[-1] in ["OK", "ERROR"]:
+                break
+        msg = self.termination_str.join(lines[:-1])
+        if lines[-1] != 'OK':
+            self.adapter.connection.clear()
+            # should be self.adapter.flush_read_buffer() after PR 836
```
dkriegner, line 346:
```suggestion
            self.adapter.flush_read_buffer()
```

dkriegner:
> Do you have a better suggestion? I agree, that it is not very probable, but with that comment (and code) in place, a user will see the change more easily.

I guess the correct way to do this would be to find out if the first argument is a hostname/IP-adress (as opposed to an Adapter-instance or valid VISA resource) and in this case trigger the warning and use the deprecated host handling. In this way user code would still work unchanged but trigger the appropriate warning. No idea if this can be done somewhat elegantly?

bmoneke:
Good idea.
You do have to specify a port in the resource string, don't you?
We could check whether a colon (`":"`) is in the resource string. If not, the warning is raised.

dkriegner:
Yes I think a `::` is needed to specify the port. From the specs I see that the device can also have USB and RS232 connections, but I am not sure if the protocol is the same on these other interfaces... Anyways even for those resource types a double semicolon would be present as far as I understand. The deprecated host handling could be used otherwise

Max-Herbold:
Ok, reimplemented `handle_deprecated_host_arg`. In the table, each case is documented with the implemented behaviour and covers everything mentioned previously. Kept this all in one method so super easy to remove if later desired.

Case #    | adapter (key word) | host (key word)       | Behaviour
-----------------------|--------------------------------------|-------------------|--------------------------------------------------------------------------------------------------------------------
1                     | None                                 | None             | raise TypeError, prompt for "adapter"
2                     | None                                 | Any               | warn, The 'host' argument is deprecated. Use 'adapter' instead. And set `adapter = host` and utilize the cases 3, 4, and 5. |
3                     | TCPIP0::192.168.0.1::7230::SOCKET (str) | Any | :: found so return same string (host ignored)
4                     | 192.168.0.1 (str)                         | Any             | :: not found so display a warning, create and return resource string (with default port) (host ignored)
5                     | PyMeasure Adapter                        | Any              | return adapter (host ignored)
6 | Any other type | Any | raise TypeError, 'adapter' argument must be a string or Adapter (host ignored)

pymeasure/instruments/attocube/anc300.py
```diff
@@ -255,3 +299,76 @@ def stop_all(self):
...
+        current definition of the __init__ function. This is used to support outdated (deprecated)
+        code. and separated out to make it easier to remove in the future. To whoever removes this:
+        This function should be removed and the `adapter` argument in the __init__ method should
+        be made non-optional.
+
+        :param dict kwargs: keyword arguments passed to the __init__ function,
+            including the deprecated `host` argument.
+        :return str: resource string for the VISAAdapter
+        """
+        host = kwargs.pop("host", None)
```
dkriegner, line 315: If I see it correctly then the `host` kwarg did not exist before this PR. It was a positional argument. Can one not remove the `host` kwarg? I think applying the magic below on the adapter argument alone would be sufficient?

dkriegner (commented):
just few minor comments to consider. Although the unit test should now cover the important things I can retest with hardware before merging. just ping me

pymeasure/instruments/attocube/anc300.py
```diff
@@ -300,24 +300,42 @@ def stop_all(self):
...
+            if adapter.find("::") > -1:
+                # adapter is a resource string, so use it
+                return adapter
+            # otherwise, `adapter` can only be an (deprecated) IPv4 address, so display a
+            # deprecation warning and create a resource string
+            warn(
+                "Using an IPv4 address is deprecated. Use a full VISA resource string instead.",
+                FutureWarning,
+            )
+            return f"TCPIP0::{adapter}::7230::SOCKET"
```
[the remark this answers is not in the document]
Max-Herbold: no doesn't change behaviour. Maybe removes some ambiguity

pymeasure/instruments/attocube/anc300.py
Max-Herbold: This was added so that if someone had ANC300Controller(`host=168.169.0.2`, ...) was given. Then requires no update for previous code to continue working without changes.

> I am not sure if the host keyword argument you show with the deprecation warning makes a lot of sense. Previously host was a regular argument so the chance that someones old code works without adapting are very small.

This now works both if previously someone code had ANC300Controller(`host=168.169.0.2`, ...)  or ANC300Controller(`168.169.0.2`, ...). What am I missing here?

pymeasure/instruments/attocube/anc300.py
```diff
@@ -300,24 +300,42 @@ def stop_all(self):
...
-        warn("The 'host' argument is deprecated. Use 'adapter' instead.", FutureWarning)
-        return f"TCPIP0::{host}::7230::SOCKET"
+            warn("The 'host' argument is deprecated. Use 'adapter' instead.", FutureWarning)
+            adapter = host
+
+        if isinstance(adapter, str):
+            if adapter.find("::") > -1:
+                # adapter is a resource string, so use it
+                return adapter
+            # otherwise, `adapter` can only be an (deprecated) IPv4 address, so display a
```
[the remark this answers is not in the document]
Max-Herbold: Hmm, when would the host name contain a "::"?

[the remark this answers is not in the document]
dkriegner: certainly it will not contain `""` , but I would not use "IPv4 address is deprecated" since it could be a hostname as well. I would just rephrase the Warning message and text of the comment

pymeasure/instruments/attocube/anc300.py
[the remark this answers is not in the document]
dkriegner: personally I would use `TCPIP::` and let pyvisa choose the default board.

pymeasure/instruments/attocube/anc300.py
dkriegner: ok, you are right, this is the most backward compatible solution conceivable.

pymeasure/instruments/attocube/anc300.py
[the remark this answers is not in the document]
Max-Herbold: Understood. Thanks for the heads up, didn't occur that this was a feature. :)

dkriegner (approved):
I tested this with the hardware (of course after merging master) and it seems all good. 

Only two remaining comments: 
1) Due to misbehavior of the hardware when asking for the `frequency` a warning is raised by pyvisa. I am afraid this is unavoidable since it occurs only with this property. It nevertheless works correct!

2. Code which used this device before this PR (and did not name every parameter in the constructor call): e.g.
`m = ANC300Controller("192.168.0.1", ["x", "y", "z"], "passwd")`
will still need to be changed, since now there is a new second argument (`name`). The resulting exception is:
`Exception: Attocube authorization failed 'Authorization failure'`
This seems unavailable, but the also shown `FutureWarning` should be a hint for the user that something changed.

Max-Herbold:
>    will still need to be changed, since now there is a new second argument (`name`). The resulting exception is:

That can be fixed multiple ways. The easiest is moving the name to be the last key word argument - please verify that this does not match previous pymeasure structure.

bmoneke:
> The easiest is moving the name to be the last key word argument - please verify that this does not match previous pymeasure structure.

I'm unsure. At one hand, it is good that all instruments start with "adapter, name...", on the other hand, you rarely need to give the name (in an instrument), such that the name might be the last kwarg. I tend to keep name the second kwarg and not the last. @bilderbuchi , what do you think?

dkriegner:
If its a package wide convention to have adapter, name sequence I would keep it. 
The additional warning which was added is likely good to warn users that they have to change their code.

bilderbuchi:
I'd also favour to remain consistent with the rest of the instruments. As long as there is an intelligible warning that is good. Let's also make sure that the changelog calls that out specifically (so that affected folks notice early and can migrate proactively).

bilderbuchi:
> Due to misbehavior of the hardware when asking for the frequency a warning is raised by pyvisa. I am afraid this is unavoidable since it occurs only with this property. It nevertheless works correct!

@dkriegner what's the warning? Anything we can change in the frequency property to work around this (some processing)? Or is this a peculiarity of _your specific_ device, not the model in general?

dkriegner:
> @dkriegner what's the warning? Anything we can change in the frequency property to work around this (some processing)? Or is this a peculiarity of _your specific_ device, not the model in general?

The warning is as follows:
```
.../pymeasure_herbold/pymeasure/adapters/visa.py:168: UserWarning: read string doesn't end with termination characters
  return self.connection.read(**kwargs)
```
It occurs when any of the `Axis.frequency` properties is read. (It does not occur with any other similar property of `Axis`). 
I can't say if this is my specific version. (I guess its likely this specific firmware version). I can not see what we can do in the property to change this since the warning comes from pyvisa internals. It does not affect the functionality nor speed of checking this property and is already documented in the tests.

bmoneke:
In my opinion, we can merge as that seems to be a firmware bug.

You may merge, if you agree, @bilderbuchi

Issue #834, opened 2023-01-18
Pyvisa-py `flush` raises NotImplementedError


bmoneke:
We could (and should) add the error handling in the `VISAAdapter.flush_read_buffer` method, not here, until pyvisa-py is updated. Please feel free to open an issue (or directly a PR).

The instrument will work flawlessly without this fix. I'll resolve.

_Originally posted by @bmoneke in https://github.com/pymeasure/pymeasure/pull/819#discussion_r1073511548_

bilderbuchi:
Related upstream PR: https://github.com/pyvisa/pyvisa-py/pull/208#issue-494956830

dkriegner:
New issue raised upstream: https://github.com/pyvisa/pyvisa-py/issues/348

dkriegner:
small update: pyvisa-py for TCP sockets now has a flush method: https://github.com/pyvisa/pyvisa-py/pull/350
I am afraid for other resource types this won't be so easy.

PR #836, opened 2023-01-19
fallback version of discarding the read buffer

dkriegner:
This PR shall implement a fallback method for `VisaAdapter.flush_read_buffer` which in pyvisa-py currently is supported only by the ASRL resource type and raises a `NotImplementedError` otherwise. 

This is related to issue #834

### DIFF
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -228,7 +228,15 @@ class VISAAdapter(Adapter):
         in the read buffer and no END-indicator was present, read from the device until
         encountering an END indicator (which causes loss of data).
         """
-        self.connection.flush(pyvisa.constants.BufferOperation.discard_read_buffer)
+        try:
+            self.connection.flush(pyvisa.constants.BufferOperation.discard_read_buffer)
+        except NotImplementedError:
+            # fake discarding the read buffer by reading all available messages.
+            try:
+                while True:
+                    self.read()
+            except pyvisa.errors.VisaIOError:
+                pass
 
     def __repr__(self):
         return "<VISAAdapter(resource='%s')>" % self.connection.resource_name
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -231,12 +231,15 @@ class VISAAdapter(Adapter):
         try:
             self.connection.flush(pyvisa.constants.BufferOperation.discard_read_buffer)
         except NotImplementedError:
+            # NotImplementedError is raised when using resource types other than `asrl`
+            # in conjunction with pyvisa-py.
+            # Upstream issue: https://github.com/pyvisa/pyvisa-py/issues/348
             # fake discarding the read buffer by reading all available messages.
             timeout = self.connection.timeout
             self.connection.timeout = 0
             try:
                 while True:
-                    self.read()
+                    self.read_bytes(-1)
             except pyvisa.errors.VisaIOError:
                 pass
             self.connection.timeout = timeout
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -241,7 +241,8 @@ class VISAAdapter(Adapter):
                 self.read_bytes(-1)
             except pyvisa.errors.VisaIOError:
                 pass
-            self.connection.timeout = timeout
+            finally:
+                self.connection.timeout = timeout
 
     def __repr__(self):
         return "<VISAAdapter(resource='%s')>" % self.connection.resource_name
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -232,11 +232,14 @@ class VISAAdapter(Adapter):
             self.connection.flush(pyvisa.constants.BufferOperation.discard_read_buffer)
         except NotImplementedError:
             # fake discarding the read buffer by reading all available messages.
+            timeout = self.connection.timeout
+            self.connection.timeout = 0
             try:
                 while True:
                     self.read()
             except pyvisa.errors.VisaIOError:
                 pass
+            self.connection.timeout = timeout
 
     def __repr__(self):
         return "<VISAAdapter(resource='%s')>" % self.connection.resource_name
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -238,8 +238,7 @@ class VISAAdapter(Adapter):
             timeout = self.connection.timeout
             self.connection.timeout = 0
             try:
-                while True:
-                    self.read_bytes(-1)
+                self.read_bytes(-1)
             except pyvisa.errors.VisaIOError:
                 pass
             self.connection.timeout = timeout
--- a/pymeasure/adapters/protocol.py
+++ b/pymeasure/adapters/protocol.py
@@ -25,6 +25,8 @@
 import logging
 from unittest.mock import MagicMock
 
+import pyvisa
+
 from .adapter import Adapter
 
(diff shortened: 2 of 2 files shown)


dkriegner:
Commit 0100e80 implements the bare minimum working version, but I think it makes sense to reduce the timeout during this operation, which is implemented in 55ae2f4.

I see few issues which should be discussed:
1) one will often call `flush_read_buffer` to clean up after a communication error. In this case it might be clever/required to wait a moment for the miscommunication to settle. e.g. in the `IBeamSmart` Instrument the `flush_read_buffer` is called like this:
```
        time.sleep(0.04)
```
Do we want to include a delay argument to `flush_read_buffer` to include this wait directly in the `VisaAdapter` method?

2) Should we try to get rid of the `AttributeError` generated by attempting to call `flush_read_buffer` from the `ProtocolAdapter`? I think it's strange that we need to change Instruments (see #819) to catch an Exception only generated because of the tests

One could just add to `ProtocolAdapter`
```
def flush_read_buffer(self):
    """Define noop method to allow Instrument tests to pass."""
```
In general I think it makes no sense to guard the `flush_read_buffer` call by `try/except` since if an adapter does not support this function and an Instrument requires it, it is likely better to raise the exception. Follow-up communication is screwed up if flush is required, which renders the Instrument not useable.

pymeasure/adapters/visa.py
```diff
@@ -228,7 +228,18 @@ def flush_read_buffer(self):
...
-        self.connection.flush(pyvisa.constants.BufferOperation.discard_read_buffer)
+        try:
+            self.connection.flush(pyvisa.constants.BufferOperation.discard_read_buffer)
+        except NotImplementedError:
+            # fake discarding the read buffer by reading all available messages.
+            timeout = self.connection.timeout
+            self.connection.timeout = 0
+            try:
+                while True:
+                    self.read()
```
bmoneke, line 239: Read bytes might be better, in case a non ASCII char is sent and the decoding in `read` would get a hickup:
```suggestion
                    self.read_bytes(-1)
```

pymeasure/adapters/visa.py
```diff
@@ -228,7 +228,18 @@ def flush_read_buffer(self):
         in the read buffer and no END-indicator was present, read from the device until
         encountering an END indicator (which causes loss of data).
         """
-        self.connection.flush(pyvisa.constants.BufferOperation.discard_read_buffer)
+        try:
+            self.connection.flush(pyvisa.constants.BufferOperation.discard_read_buffer)
+        except NotImplementedError:
+            # fake discarding the read buffer by reading all available messages.
```
bilderbuchi, line 234: Please add the additional context (that we need this because `flush` is currently only implemented for `asrl`) and a link to the upstream issue in pyvisa-py, so that we can re-check later and remove the workaround when support has been added to more (all?) backends.

dkriegner:
> It should not raise an AttributeError any more, since in https://github.com/pymeasure/pymeasure/pull/782 I faked all adapter.connection calls with a mock.

For me, it still raises the `AttributeError`. When I remove the try/excepts in IBeamSmart.__init__
https://github.com/pymeasure/pymeasure/blob/c04f21157d96e5c73e18b025b4b82c77ff2431e7/pymeasure/instruments/toptica/ibeamsmart.py#L116-L119

I get:
```
FAILED tests/instruments/toptica/test_ibeamsmart.py::test_init - AttributeError: 'ProtocolAdapter' object has no attribute 'flush_read_buffer'. Did you mean: '_read_buffer'?
```

bmoneke:
Sorry, my mistake. I thought flush_read_buffer were an `Adapter` method and its call to the connection failed.

I think it is best, if you copy the method from VISAAdapter (without your falback mechanism).
Then the adapter will call the connection.flush, which can be verified.

dkriegner:
The last two commits should address your comments. 

With these changes IBeamSmart can drop the `AttributeError` catching, but I propose to make this in a separate PR.

Thinking again about my suggestion of a delay argument I would not add it. This is Instrument specific and therefore should be handled in the Instrument.

So from my point of few this is ready.

One somewhat unrelated comment: 
The docstring of `read_bytes` is misleading/wrong. The docstring suggests that all the input buffer is returned with `read_bytes(-1)`, but one only gets one message -> until the termination character!

bmoneke:
> The docstring of read_bytes is misleading/wrong. The docstring suggests that all the input buffer is returned with read_bytes(-1), but one only gets one message -> until the termination character!

Really? I looked (writing the docstring and now) at the pyvisa code and did not see a break at a termchar. For that reason, I used `read_raw` in the VISAAdapter.read_bytes implementation.

dkriegner:
I tested the `read_bytes(-1)` with a real device and there its quite obvious that it does not do what the docstring says.

Why I mention it here is because my implementation could be simplified if `read_bytes(-1)` behaves as documented. 
In this case the `while` loop in `flush_read_buffer` could be dropped.

If you think you will fix the read_bytes code, then one should maybe keep this PR open until this is settled. If you tend to update the docstring than I think this is ok as is.

Another question: Is it ok what I put into `ProtocolAdapter`? It feels awkward that in the `protocol.py` one now needs to `import pyvisa`

bmoneke:
> Another question: Is it ok what I put into ProtocolAdapter? It feels awkward that in the protocol.py one now needs to import pyvisa

I'd put a string in it, instead (maybe the name of the variable). That way you can verify what kind of flush was called and verify that, if you whish.

@bilderbuchi what do you think, how we should put `flush_read_buffer` in the protocolAdapter?

bmoneke:
> I tested the read_bytes(-1) with a real device and there its quite obvious that it does not do what the docstring says.

Thanks

> Why I mention it here is because my implementation could be simplified if read_bytes(-1) behaves as documented.
In this case the while loop in flush_read_buffer could be dropped. If you think you will fix the read_bytes code, then one should maybe keep this PR open until this is settled. If you tend to update the docstring than I think this is ok as is.

Please open an issue and mention there, that also the fallback could be simplified, if the code is change. I don't know yet, how we could change the code in a nice manner, as pyvisa does offer this `read_raw` method, which reads until term_char or buffer is empty, or `read_bytes`, where you have to give a certain number of bytes.

There is no middle thing "Give me all bytes, I don't want to tell how many".

bilderbuchi:
> > Another question: Is it ok what I put into ProtocolAdapter? It feels awkward that in the protocol.py one now needs to import pyvisa
> 
> I'd put a string in it, instead (maybe the name of the variable). That way you can verify what kind of flush was called and verify that, if you whish.
> 
> @bilderbuchi what do you think, how we should put `flush_read_buffer` in the protocolAdapter?

Yes, it's a very short method, so duplicating should be fine in this instance. 
Good call on the pyvisa import, I agree that it seems untidy. Yeah, just put a string as the argument. I don't think it matters much, as in the ProtocolAdapter the `connection` is a mock.

dkriegner:
I think this is in principle ready to merge. A PR fixing #837 can then simplify the code.
Otherwise, keep it open and I look into it once #837 is fixed.

bmoneke:
@dkriegner thanks for the work.
It looks good. I also did fix the adapters in #839 which should be a fast fix.

In that PR, I change the behavior to read all bytes, unless break_on_termchar requested.

If you want you can simplify your code here and we will merge your PR after mine has been merged, or I merge immediately and we change it later. My favourite (as the PR for #837 is ready) is the first option.

bmoneke:
Note to myself:
1. After #839  is merged
2. Merge this
3. Merge #835

PR #838, opened 2023-01-20
Kuhne Electronic KU SG 2.45 250 A instrument implementation added.

jurajjasik:
I'd like to add a support for [Kuhne Electronic 2.45 GHz ISM-Band Microwave Generator](https://shop.kuhne-electronic.com/kuhne/en/shop/professionell/prof-signalquellen/prof-signalgen/prof-power-splitters/KU+SG+2+45-250A+Microwave+Generator/?card=1768).
The communication is implemented according to the manufacturer's documentation but not yet tested with the instrument.


jurajjasik:
It looks to me that the failing of the test was not related to my code. Could you you check it? Thanks.
image

msmttchr:
> It looks to me that the failing of the test was not related to my code. Could you you check it? Thanks. image

I think this was a transient issue on GitHub action. I have re-run failed tests and now they are ok.

pymeasure/instruments/kuhneelectronic/kusg245_250a.py
```diff
@@ -0,0 +1,305 @@
...
+        "r?",
+        "%s",
+        """Control the whether the amplifier enabling is done
+        via external inputs on 8-pin connector
+        or via the serial interface (boolean).
+        """,
+        validator=strict_discrete_set,
+        values=[False, True],
+        set_process=lambda v: {True: "R", False: "r"}[v],
+        get_process=lambda v: bool(v)
```
bilderbuchi, line 101: Why not do this just like `freq_steps_fine_enabled` (replacing the values in the dict accordingly)?

jurajjasik: The reason is in different protocol of setter and getter commands:
- To enable (disable) that function, user must send "R" ("r") commands.
- The status is checked by "r?" command but answer is "1" ("0"). 

Therefore, special `set_process` and `get_process` are used.

The same logic is in `bias_enabled`, `rf_enabled` and `pulse_mode_enabled` properties.

bilderbuchi: oh, for... :cursing_face: 
I see, that makes sense! :-)

jurajjasik:
I have just received device and found out that some of the commands return binary values and another ones return text. 🤦🏻‍♂️
Therefore I switched this PR to draft until I find out some workaround...

bmoneke:
If necessary, you can always write an ordinary python property. For example, if some request returns binary values, you do not use a pymeasure property, but plain python property dealing with that binary data.

By the way, you may add yourselfs to the AUTHORS.txt, if you want.

jurajjasik:
I used ordinary python properties for requests returning binary values.
I found out that configuration requests return acknowledgment character "A" which is now checked using overloaded `write()` method.
The communication was tested using real device.
Tests are also updated.

bilderbuchi:
I probably won't find the time for a detailed review for a while, maybe @bmoneke can fit that in? Otherwise, I ask you for patience.

pymeasure/instruments/kuhneelectronic/kusg245_250a.py
```diff
@@ -0,0 +1,414 @@
...
+    def shutdown(self):
+        """
+        Safe shut-down the generator.
+
+        1. Disable RF output.
+        2. Deactivate biasing.
+        """
+        self.rf_enabled = False
+        self.bias_enabled = False
+
```
bmoneke, line 399: Please add a super call, such that the parents shutdown is called:
```suggestion
        super().shutdown()

```

pymeasure/instruments/kuhneelectronic/kusg245_250a.py
```diff
@@ -0,0 +1,414 @@
...
+        2. Enable RF output.
+        """
+        self.bias_enabled = True
+        time.sleep(0.500)  # not sure if needed
+        self.rf_enabled = True
+
+    def write(self, command, **kwargs):
+        self.adapter.write(command, **kwargs)
+        if _is_expecting_acknowledgement(command) and self.read() != "A":
+            raise ConnectionError("Expected acknowledgment.")
```
bmoneke, line 414: Would it be good to show, what the instrument returned instead?

bmoneke (changes requested):
Thanks for your contribution and welcome to the Contributors @jurajjasik !

The code looks already very good. Thank you for the tests.

I made some comments, mainly regarding small things.

pymeasure/instruments/kuhneelectronic/kusg245_250a.py
```diff
@@ -0,0 +1,414 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import time
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set
+from pymeasure.instruments.validators import truncated_range, truncated_discrete_set
```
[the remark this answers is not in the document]
jurajjasik: Dear @bmoneke, thank you for your code review! I will implement your suggestion in commits step-by-step.
This one is done.

jurajjasik:
Hi Benedikt,

Yes, please.
Thank you!

Best regards,
Juraj

BenediktBurger:
Could you merge the master into this branch, please?

BenediktBurger:
Thanks for your work!

PR #839, opened 2023-01-24
Adapters' read_bytes reworked

### LINKED ISSUE #837: Instrument.read_bytes(-1) does not behave as advertised
The docstring of read_bytes says:

https://github.com/pymeasure/pymeasure/blob/c04f21157d96e5c73e18b025b4b82c77ff2431e7/pymeasure/instruments/instrument.py#L145-L149

Upon testing (see below) it, however, turns out that it only returns the data until a termination character is contained. (The termination character is not stripped). If no termination character is received, `read_bytes(-1)` raises a `pyvisa.errors.VisaIOError`.

I tested this with pyvisa (pyvisa-py) with the code below:

```
from pymeasure.instruments.mksinst import MKS937B
g = MKS937B("TCPIP::10.0.131.254::10001::SOCKET")  # termination_char here is ";"
g.write("SN?")
g.write("SN?")

g.read_bytes(-1) # returns b'@253ACK022869254 ;'
g.read_bytes(-1) # returns b'FF@253ACK022869254 ;'
g.read_bytes(-1) # raises the error below (but the remaining "FF" which is still in the read buffer can not be obtained.
```

```
  File "<stdin>", line 1, in <module>
  File "C:\Users\AJAlab\Documents\GitHub\pymeasure\pymeasure\instruments\instrument.py", line 153, in read_bytes
    return self.adapter.read_bytes(count, **kwargs)
  File "C:\Users\AJAlab\Documents\GitHub\pymeasure\pymeasure\adapters\adapter.py", line 121, in read_bytes
    read = self._read_bytes(count, **kwargs)
  File "C:\Users\AJAlab\Documents\GitHub\pymeasure\pymeasure\adapters\visa.py", line 161, in _read_bytes
    return self.connection.read_raw(**kwargs)
  File "C:\Users\AJAlab\AppData\Roaming\Python\Python310\site-packages\pyvisa\resources\messagebased.py", line 405, in read_raw
    return bytes(self._read_raw(size))
  File "C:\Users\AJAlab\AppData\Roaming\Python\Python310\site-packages\pyvisa\resources\messagebased.py", line 442, in _read_raw
    chunk, status = self.visalib.read(self.session, size)
  File "C:\Users\AJAlab\AppData\Roaming\Python\Python310\site-packages\pyvisa_py\highlevel.py", line 519, in read
    return data, self.handle_return_value(session, status_code)
  File "C:\Users\AJAlab\AppData\Roaming\Python\Python310\site-packages\pyvisa\highlevel.py", line 251, in handle_return_value
    raise errors.VisaIOError(rv)
pyvisa.errors.VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
```

bmoneke:
Closes #837 

As discussed: Default is to **not** break at a term char, but the option remains. Changes are implemented in all adapters, which have `read_bytes`: VISA, Fake, Serial, VXI11.

Tests are added accordingly.

The first commit contains the changes.
The second commit contains some cosmetic docstring changes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ New adapter and instrument mechanics
 ------------------------------------
 - Channel class added. Instrument.channels and Instrument.ch_X (X is any channel name) are reserved for channel implementations.
 - All instruments are required to accept a :code:`name` argument.
+- :code:`read_bytes` of all Adapters does not break on a termination character, unless :code:`break_on_termchar` is set to `True`.
 
 Deprecated features
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ New adapter and instrument mechanics
 ------------------------------------
 - Channel class added. Instrument.channels and Instrument.ch_X (X is any channel name) are reserved for channel implementations.
 - All instruments are required to accept a :code:`name` argument.
-- :code:`read_bytes` of all Adapters does not break on a termination character, unless :code:`break_on_termchar` is set to `True`.
+- Changed: :code:`read_bytes` of all Adapters by default does not stop reading on a termination character, unless the new argument :code:`break_on_termchar` is set to `True`.
 
 Deprecated features
 -------------------
--- a/pymeasure/adapters/protocol.py
+++ b/pymeasure/adapters/protocol.py
@@ -24,6 +24,7 @@
 
 import logging
 from unittest.mock import MagicMock
+from warnings import warn
 
 from .adapter import Adapter
 
@@ -134,6 +135,10 @@ class ProtocolAdapter(Adapter):
 
         :param int count: Number of bytes to read. If -1, return the buffer.
         """
+        if break_on_termchar:
+            warn(("Breaking on termination character in `read_bytes` cannot be tested. "
+                  "You have to separate the message parts in the com_pairs."),
+                 UserWarning)
         if self._read_buffer is not None:
             if count == -1 or count >= len(self._read_buffer):
                 read = self._read_buffer
--- a/pymeasure/adapters/protocol.py
+++ b/pymeasure/adapters/protocol.py
@@ -129,7 +129,7 @@ class ProtocolAdapter(Adapter):
         """Return an already present or freshly fetched read buffer as a string."""
         return self._read_bytes(-1).decode("utf-8")
 
-    def _read_bytes(self, count, **kwargs):
+    def _read_bytes(self, count, break_on_termchar=False, **kwargs):
         """Read `count` number of bytes from the buffer.
 
         :param int count: Number of bytes to read. If -1, return the buffer.
--- a/pymeasure/adapters/serial.py
+++ b/pymeasure/adapters/serial.py
@@ -104,8 +104,8 @@ class SerialAdapter(Adapter):
                                               count if count > 0 else None,
                                               **kwargs)
         else:
-            # At negative count, we read up to one GB, which can be considered the whole buffer.
-            return self.connection.read(count if count >= 0 else 1e9, **kwargs)
+            # At -1 we read a very large number of bytes, which can be considered the whole buffer.
+            return self.connection.read(1e99 if count == -1 else count, **kwargs)
 
     def __repr__(self):
         return "<SerialAdapter(port='%s')>" % self.connection.port
--- a/pymeasure/adapters/serial.py
+++ b/pymeasure/adapters/serial.py
@@ -103,10 +103,9 @@ class SerialAdapter(Adapter):
             return self.connection.read_until(self.read_termination.encode(),
                                               count if count > 0 else None,
                                               **kwargs)
-        elif count == -1:
-            return b"\n".join(self.connection.readlines(**kwargs))
         else:
-            return self.connection.read(count, **kwargs)
+            # At negative count, we read up to one GB, which can be considered the whole buffer.
+            return self.connection.read(count if count >= 0 else 1e9, **kwargs)
 
     def __repr__(self):
         return "<SerialAdapter(port='%s')>" % self.connection.port
--- a/pymeasure/adapters/serial.py
+++ b/pymeasure/adapters/serial.py
@@ -79,10 +79,10 @@ class SerialAdapter(Adapter):
     def _read(self, **kwargs):
         """Read up to (excluding) `read_termination` or the whole read buffer.
 
-        :param kwargs: Keyword arguments for the connection itself.
+        :param \\**kwargs: Keyword arguments for the connection itself.
         :returns str: ASCII response of the instrument (read_termination is removed first).
         """
-        read = self._read_bytes(-1, **kwargs).decode()
+        read = self._read_bytes(-1, break_on_termchar=True, **kwargs).decode()
         # Python>3.8 this shorter form is possible:
         # self._read_bytes(-1).decode().removesuffix(self.read_termination)
         if self.read_termination:
@@ -90,19 +90,21 @@ class SerialAdapter(Adapter):
         else:
             return read
 
-    def _read_bytes(self, count, **kwargs):
+    def _read_bytes(self, count, break_on_termchar, **kwargs):
         """Read a certain number of bytes from the instrument.
 
         :param int count: Number of bytes to read. A value of -1 indicates to
-            read the whole read buffer or until encountering the read_termination.
-        :param kwargs: Keyword arguments for the connection itself.
+            read from the whole read buffer.
+        :param bool break_on_termchar: Stop reading at a termination character.
+        :param \\**kwargs: Keyword arguments for the connection itself.
         :returns bytes: Bytes response of the instrument (including termination).
         """
-        if count == -1:
-            if self.read_termination:
-                return self.connection.read_until(self.read_termination, **kwargs)
-            else:
-                return b"\n".join(self.connection.readlines(**kwargs))
+        if break_on_termchar and self.read_termination:
+            return self.connection.read_until(self.read_termination.encode(),
+                                              count if count > 0 else None,
+                                              **kwargs)
+        elif count == -1:
+            return b"\n".join(self.connection.readlines(**kwargs))
         else:
             return self.connection.read(count, **kwargs)
 
--- a/pymeasure/adapters/serial.py
+++ b/pymeasure/adapters/serial.py
@@ -44,7 +44,7 @@ class SerialAdapter(Adapter):
 
     :param write_termination: String appended to messages before writing them.
     :param read_termination: String expected at end of read message and removed.
-    :param kwargs: Any valid key-word argument for serial.Serial
+    :param \\**kwargs: Any valid key-word argument for serial.Serial
     """
 
     def __init__(self, port, preprocess_reply=None,
@@ -63,7 +63,7 @@ class SerialAdapter(Adapter):
 
         :param str command: Command string to be sent to the instrument
             (without termination).
-        :param kwargs: Keyword arguments for the connection itself.
+        :param \\**kwargs: Keyword arguments for the connection itself.
         """
         command += self.write_termination
         self._write_bytes(command.encode(), **kwargs)
@@ -72,7 +72,7 @@ class SerialAdapter(Adapter):
         """Write the bytes `content` to the instrument.
 
         :param bytes content: The bytes to write to the instrument.
-        :param kwargs: Keyword arguments for the connection itself.
+        :param \\**kwargs: Keyword arguments for the connection itself.
         """
         self.connection.write(content, **kwargs)
 
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -55,7 +55,7 @@ class VISAAdapter(Adapter):
     :param float query_delay: Time in s to wait after writing and before reading.
 
         .. deprecated:: 0.11
-            Implement it in the instrument's `wait_until_read` method instead.
+            Implement it in the instrument's `wait_for` method instead.
 
     :param log: Parent logger of the 'Adapter' logger.
     :param \\**kwargs: Keyword arguments for configuring the PyVISA connection.
@@ -88,7 +88,7 @@ class VISAAdapter(Adapter):
         super().__init__(preprocess_reply=preprocess_reply, log=log)
         if query_delay:
             warn(("Parameter `query_delay` is deprecated. "
-                  "Implement in Instrument's `wait_until_read` instead."),
+                  "Implement in Instrument's `wait_for` instead."),
                  FutureWarning)
             kwargs.setdefault("query_delay", query_delay)
         self.query_delay = query_delay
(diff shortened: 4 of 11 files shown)


pymeasure/adapters/protocol.py
```diff
@@ -129,7 +129,7 @@ def _read(self, **kwargs):
         """Return an already present or freshly fetched read buffer as a string."""
         return self._read_bytes(-1).decode("utf-8")
 
-    def _read_bytes(self, count, **kwargs):
+    def _read_bytes(self, count, break_on_termchar=False, **kwargs):
         """Read `count` number of bytes from the buffer.
```
bilderbuchi, line 133: Should we call out in the docstring that `break_on_termchar` is not honored? (and maybe warn/raise if it's not False)?

pymeasure/adapters/serial.py
```diff
@@ -72,37 +72,39 @@ def _write_bytes(self, content, **kwargs):
...
     def _read(self, **kwargs):
         """Read up to (excluding) `read_termination` or the whole read buffer.
 
-        :param kwargs: Keyword arguments for the connection itself.
+        :param \\**kwargs: Keyword arguments for the connection itself.
         :returns str: ASCII response of the instrument (read_termination is removed first).
         """
-        read = self._read_bytes(-1, **kwargs).decode()
+        read = self._read_bytes(-1, break_on_termchar=True, **kwargs).decode()
         # Python>3.8 this shorter form is possible:
```
bilderbuchi, line 86: One thing that's not 100% clear (for all methods, not specifically here): If it stops reading at the termination char, is the termination char included in the reply, or stripped? If the latter, I think we don't need the if-branch here anymore?

bmoneke: The docstring of `read_bytes` says ":returns:.... (including termination)", makes it clearer?

pymeasure/adapters/protocol.py
bmoneke: The difficulty is, how to implement it, as we do not have termchars in the ProtocolAdapter.
If we raise a NotImplemented Warning at `if break_on_termchar:`, a test could verify, that the correct form is called.

tests/adapters/test_serial.py
```diff
@@ -64,6 +64,17 @@ def test_read_bytes(adapter):
     assert adapter.read_bytes(11) == b"basd\x02fasdf\n"
 
 
+def test_read_bytes_unlimited(adapter):
+    adapter.write_bytes(b"basd\x02fasdf\n")
```
bilderbuchi, line 68: I guess that should have a `\n` (and associated `read_termination` setting) in the middle of the string (like the next test) to confirm that the default indeed is not stopped by a termchar?

tests/adapters/test_visa.py
```diff
@@ -98,6 +98,14 @@ def test_write_read_all_bytes(adapter):
     assert adapter.read_bytes(-1) == b"SCPI,MOCK,VERSION_1.0\n"
 
 
+def test_write_read_break_on_termchar(adapter):
+    """Test read_bytes breaks or does not break on termchar."""
+    adapter.write("*IDN?")
+    adapter.connection.read_termination = ","
+    assert adapter.read_bytes(-1, break_on_termchar=True) == b"SCPI,"
+    assert adapter.read_bytes(-1, break_on_termchar=False) == b"MOCK,VERSION_1.0\n"
```
bilderbuchi, line 106: Can we split this in two tests, where in the second you don't give the `break_on_termchar` argument? That way we would automatically confirm that the default behaviour has not changed.

bilderbuchi (commented):
LGTM, some remarks with some questions and suggestions for improvement.

tests/adapters/test_visa.py
bmoneke: The default behavior (newly introduced in 0.11) has changed, at least according to the implementation.
According to intention and docstring, the behavior remains the same.

The reason I put the second assert into the test is, that after the first assert, the second part of the message remains in the read buffer and the following test fails.

I'll make the default implicit to catch future changes of default behavior. 

Probably I have to use some flush or clear method to do it properly.

tests/adapters/test_serial.py
bmoneke: Basically I have a test for reading the whole string, neglecting the Termchar in middle option, and tests for either option, but yes, we can have one test for the default (testing the not splitting together with just reading all bytes) and one for the option.

pymeasure/adapters/protocol.py
bmoneke: I raise a warning now. If you want to test, that the correct parameter is called, you can verify, that the warning has been raised.

bmoneke:
Having written the tests (now a more complete subset, thanks for that), I had to adjust the adapters:

- pyvisa's read_bytes does not honor "break_on_termchar" parameter (at least in pyvisa-sim), therefore I changed the code.
- I exchanged "readlines" from the SerialAdapter (with some drawbacks, see [PySerial: Readline](https://pyserial.readthedocs.io/en/latest/shortintro.html#readline)) for read_bytes reading 1 GB, as even "readlines" reads until a timeout occurs (maybe several times). We could increase the maximum size, but I figured, that 1 GB is larger than any reasonable chunk size. And if you need more, you can call `read_bytes(1e20)`, for example.

tests/adapters/test_visa.py
bilderbuchi: > The reason I put the second assert into the test is, that after the first assert, the second part of the message remains in the read buffer and the following test fails.

If that happens that is a problem that has to be fixed, not circumvented. We cannot leak state from one test to another, that will lead to weird problems sooner or later.

Is the fixture module scoped? Ill check, I'm just on the phone.

If we change the default behaviour compared to the last release,  and we can't or dont want to avoid thst, that should be documented in the changelog.

bilderbuchi:
> * pyvisa's read_bytes does not honor "break_on_termchar" parameter (at least in pyvisa-sim).

We should find out which of those it is, and report it upstream, if the behaviour is not as documented.

pymeasure/adapters/serial.py
bilderbuchi: Oops, missed that.

tests/adapters/test_visa.py
bmoneke: I added a "read_bytes(-1)" to the fixture, such that the read buffer is emptied after each test.
Each test has its own adapter (I checked the fixture, its the default scope), but somehow pyvisa-sim keeps the state...

I added a changelog entry, that now all adapters do not break at term char. I'd consider the previous behaviour as a bug, not a feature with a transition path.

bmoneke:
> We should find out which of those it is, and report it upstream, if the behaviour is not as documented.

I found my error: I gave break_on_termchar as an ordinary argument, not as a keyword and therefore it landed in the wrong argument...

tests/adapters/test_visa.py
bmoneke: These tests, for example succeed (based on `test_visa.py`):
~~~python
import pytest
from pymeasure.adapters import VISAAdapter

@pytest.fixture
def adapter():
    return VISAAdapter('ASRL2::INSTR', visa_library='@sim',
                       read_termination="\n",
                       timeout=10)


def test_com(adapter):
    adapter.write("*IDN?")
    adapter.connection.read_termination = "x"  # change adapter instance
    assert adapter.connection.read_termination == "x"


def test_com2(adapter):
    assert adapter.connection.read_termination == "\n"  # new adapter instance
    assert adapter.read_bytes(-1) == b"SCPI,MOCK,VERSION_1.0\n"  # old read buffer

bilderbuchi: > I added a changelog entry, that now all adapters do not break at term char. 

Thanks!

> I'd consider the previous behaviour as a bug, not a feature with a transition path.

Fair enough.

bilderbuchi: > Each test has its own adapter (I checked the fixture, its the default scope), but somehow pyvisa-sim keeps the state...

This is messed up. I fear that even if we report a bug it's probably not gonne be addressed, pyvisa-sim is not that well/actively maintained.

pymeasure/adapters/serial.py
```diff
@@ -103,10 +103,9 @@ def _read_bytes(self, count, break_on_termchar, **kwargs):
             return self.connection.read_until(self.read_termination.encode(),
                                               count if count > 0 else None,
                                               **kwargs)
-        elif count == -1:
-            return b"\n".join(self.connection.readlines(**kwargs))
         else:
-            return self.connection.read(count, **kwargs)
+            # At negative count, we read up to one GB, which can be considered the whole buffer.
```
bilderbuchi, line 107: I know it makes the code a bit easier, but why make the implementation different from the docstring (and the other Adapters'), with the "negative count" thing?

CHANGES.rst
```diff
@@ -4,6 +4,7 @@ New adapter and instrument mechanics
 ------------------------------------
 - Channel class added. Instrument.channels and Instrument.ch_X (X is any channel name) are reserved for channel implementations.
 - All instruments are required to accept a :code:`name` argument.
+- :code:`read_bytes` of all Adapters does not break on a termination character, unless :code:`break_on_termchar` is set to `True`.
```
bilderbuchi, line 7:
```suggestion
- Changed: :code:`read_bytes` of all Adapters by default does not stop reading on a termination character, unless the new argument :code:`break_on_termchar` is set to `True`.
```

bilderbuchi: I realise I'm still wondering if people would not by default want to read one "bytes message" delimited by the termination char if they use `adapter.read_bytes()` (mirroring `adapter.read()`). You don't think so?

tests/adapters/test_visa.py
```diff
@@ -88,14 +99,34 @@ def test_write_bytes_read(adapter):
...
+    @pytest.fixture()
+    def adapter(self, adapter):
+        adapter.write("*IDN?")
+        yield adapter
+        # empty the read buffer
+        try:
+            adapter.read_bytes(-1)
+        except pyvisa.errors.VisaIOError as exc:
+            if not exc.args[0].startswith("VI_ERROR_TMO"):
+                raise
```
bilderbuchi, line 112: Why did you not add the shutdown behaviour to the fixture at the top? 
Also, please give it a different name, this overloading is mighty confusing. :D
By the way, it turns out we need to `.close()` the pyvisa resource manager, then the buffer is correctly emptied (as advised in the upstream issue). I also realised that we are not cleaning this up properly, as the `close()` of our `VISAAdapter` does not call `close` on `self.manager`!

pymeasure/adapters/serial.py
bmoneke: > I exchanged "readlines" from the SerialAdapter (with some drawbacks, see PySerial: Readline) for read_bytes reading 1 GB, as even "readlines" reads until a timeout occurs (maybe several times). We could increase the maximum size, but I figured, that 1 GB is larger than any reasonable chunk size. And if you need more, you can call read_bytes(1e20), for example.

I always found that read lines fishy (reading it on the pyserial page), but it seemed to work as expected.
Having written the tests, it did not work as it should (it added additional term chars) according to the documentation.
As the "read line" documentation said, that it is based on (several times) "read", I transitioned to reading a "real large chunk" as reading everything.
We could increase the byte count to 1e18 or even 1e100, such that nobody will ever read that many bytes in one go.

"negative count" includes - 1

bilderbuchi: Sorry, I meant: the docstring says -1 reads everything,  but in fact -42 now _also_ reads everything,  and that is surprising/inconsistent! And all it seems to gain is one less if clause.

CHANGES.rst
bmoneke: We have both use cases for `read_bytes` :

A binary instrument, where you read (certain number of) bytes. Here you do not want to break at a byte which happens to be the (default) term char.

Binary data transfer in am ASCII based instrument: here again you do not want to break.

You have Text based communication, but want to skip the decoding or there are bytes in the message es well...
Here breaking at a term char might be useful.

From these examples I prefer to keep the breaking optional (it is optional in pyvisa as well).

tests/adapters/test_visa.py
bmoneke: `read_bytes` runs into a timeout, if there is no buffer.
To reduce the test time, I only clear the read buffer, where I test reading (which might fail).

Thanks for raising the issue with close. Maybe that solves the problem (and improves our adapter).

pymeasure/adapters/serial.py
bmoneke: Oh, that is what you meant.

I had thought, that it may interpret user input sloppily. I'll change it for strict comparison.

tests/adapters/test_visa.py
bmoneke: I put the cleanup in the top fixture, together with "close".

Even with closing the manager (which I added, inclusive the appropriate tests), I have to use the cleanup procedure.
`Clear` is not implemented in the sim and flush buffer seems not to work (see comments).

tests/adapters/test_visa.py
```diff
@@ -101,32 +133,26 @@ def test_write_bytes_read(adapter):
 
 class TestReadBytes:
     @pytest.fixture()
-    def adapter(self, adapter):
+    def adapterR(self, adapter):
```
bilderbuchi, line 136: :grin: I was thinking more like `adapter_with_filled_buffer`, but fair enough, it's class-local anyway.

pymeasure/adapters/serial.py
bilderbuchi: thanks!

CHANGES.rst
bilderbuchi: Fair enough, thanks!

tests/adapters/test_visa.py
bilderbuchi: thanks, LGTM.

CHANGES.rst
bmoneke: Talking, I forgot the original case (documentation)...

pymeasure/adapters/visa.py
```diff
@@ -126,43 +128,66 @@ def __init__(self, resource_name, visa_library='', preprocess_reply=None,
...
+            return self.connection.read_raw(None, **kwargs)
         else:
-            return self.connection.read_bytes(count, **kwargs)
+            read_termination = self.connection.read_termination
+            self.connection.read_termination = None
+            # Try except allows to set the read_termination even after an error.
+            try:
+                return self.connection.read_raw(**kwargs)
+            finally:
+                self.connection.read_termination = read_termination
```
bmoneke, line 190: This could be solved with a "read_termination_context" more elegantly, but there is an upstream bug in pyvisa: https://github.com/pyvisa/pyvisa/issues/723

bilderbuchi: Good idea! I'd say file an issue here with the context, mention the upstream issue, then it's easy to check when we can start using this.

Issue #792, opened 2022-12-15
Several device tests fail for test_lecroyT3DSO1204_with_device.py


bleykauf:
Just ran `test_lecroyT3DSO1204_with_device.py` and encountered several failed tests. Here is the output from the test:

Here is the short summary:

```
=== short test summary info ======================================================================================
FAILED tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py::TestLeCroyT3DSO1204::test_ch_current_configuration - AssertionError: assert {'attenuation... 'dc 1M', ...} == {'attenuation... 'dc 1M', ...}
FAILED tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py::TestLeCroyT3DSO1204::test_ch_trigger_level - assert 0.12 == 0.125
FAILED tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py::TestLeCroyT3DSO1204::test_ch_trigger_level2 - assert 0.12 == 0.125
FAILED tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py::TestLeCroyT3DSO1204::test_timebase - assert 1e-06 ± 1.0e-07 == 0.0005
FAILED tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py::TestLeCroyT3DSO1204::test_acquisition_status - AssertionError: assert 'auto' == 'triggered'
FAILED tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py::TestLeCroyT3DSO1204::test_waveform_points[100] - AttributeError: 'float' object has no attribute 'index'
FAILED tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py::TestLeCroyT3DSO1204::test_waveform_points[1000] - AttributeError: 'float' object has no attribute 'index'
FAILED tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py::TestLeCroyT3DSO1204::test_waveform_points[10000] - AttributeError: 'float' object has no attribute 'index'
FAILED tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py::TestLeCroyT3DSO1204::test_waveform_preamble - AssertionError: assert {'average': N...000000.0, ...} == {'average': N...000000.0, ...}
[...]

PR #841, opened 2023-01-27
[BUGFIX] LeCroy-Teledyne T3DSO1204: fixed unit tests with actual device

LastStarDust:
As promised I have fixed the unit and integration tests with the actual device connected.

For a more detailed explanation of the issue please refer to https://github.com/pymeasure/pymeasure/issues/792

The changes are mainly in the test routines that were not updated after some changes to the source code.

### DIFF
--- a/tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py
+++ b/tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py
@@ -336,7 +336,7 @@ class TestLeCroyT3DSO1204:
             instrument.autoscale()
             sleep(7)
         instrument.ch(case1).display = True
-        instrument.arm_acquisition()
+        instrument.single()
         sleep(1)
         data, time, preamble = instrument.download_waveform(
             source=case1, requested_points=case2, sparsing=0
@@ -350,7 +350,7 @@ class TestLeCroyT3DSO1204:
     def test_download_single_point(self, instrument):
         instrument.acquisition_type = "normal"
         instrument.ch_1.display = True
-        instrument.arm_acquisition()
+        instrument.single()
         sleep(1)
         data, time, preamble = instrument.download_waveform(source="c1", requested_points=1)
         assert type(data) is np.ndarray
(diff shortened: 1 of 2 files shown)


bmoneke (approved):
Looks good. Thanks for doing it.

bmoneke:
This PR closes #792 , doesn't it?

tests/instruments/lecroy/test_lecroyT3DSO1204_with_device.py
```diff
@@ -389,7 +396,7 @@ def test_download_data_sparsing(self, instrument):
         from matplotlib import pyplot as plt
 
         instrument.ch_1.display = True
-        instrument.single()
+        instrument.arm_acquisition()
```
bilderbuchi, line 399: I'm seeing this change a lot. This method name was used in another instrument already for the same thing (get a single measurement) -- should we maybe harmonize that and use `single()` also for this instrument? 
(afaict we have not included this instrument in a release yet, so we don't even need to deprecate anything as we don't break backwards compatibility)

LastStarDust: The reason I have renamed `single` into `arm_acquisition` is because that is how that function is called in the documentation. But I have nothing against renaming back to `single`.

bilderbuchi: That is a good reason. However, we are slowly starting to nudge towards a more homogeneous API across similar instrument classes, and for that shared names for the same things are preferable.
@bmoneke thoughts?

bmoneke: I'm for unification of names as well.

Is it similar to a "single" button on oscilloscopes?

Then I'd use "single".

LastStarDust: I renamed the `arm_acquisition` method into `single` in commit eff7dfb2adde78103cf6b4825d16ea14ee29daf0

LastStarDust: @bmoneke 

> Is it similar to a "single" button on oscilloscopes?

Yes

bmoneke: @bilderbuchi . If this conversation is resolved for you, please merge.

PR #843, opened 2023-02-01
Update property creator docstrings

### LINKED ISSUE #530: Document kwarg pass-through of property factories
Currently, it's not documented that leftover`**kwargs` of our property factories (control/measurement/setting) get passed on to `Adapter.values`. A prominent use-case is that we can pass e.g. `cast=int` to a factory to transform received values.

bmoneke:
Closes #530

### DIFF
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -524,7 +524,7 @@ class CommonBase:
                 validator=lambda x, y: x, values=(), map_values=False,
                 set_process=lambda v: v,
                 check_set_errors=False, dynamic=False,
-                **kwargs):
+                ):
         """Return a property for the class based on the supplied
         commands. This property may be set, but raises an exception
         when being read from the instrument.
@@ -553,4 +553,4 @@ class CommonBase:
                                   set_process=set_process,
                                   check_set_errors=check_set_errors,
                                   dynamic=dynamic,
-                                  **kwargs)
+                                  )
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -302,7 +302,7 @@ class CommonBase:
             received from the instrument. The callable returns the processed
             string.
         :param maxsplit: At most `maxsplit` splits are done. -1 (default) indicates no limit.
-        :param \\**kwargs: Keyword arguments to be passed to the ask method.
+        :param \\**kwargs: Keyword arguments to be passed to the :meth:`ask` method.
         :returns: A list of the desired type, or strings where the casting fails
         """
         results = self.ask(command, **kwargs).strip()
@@ -375,6 +375,7 @@ class CommonBase:
         :param check_get_errors: Toggles checking errors after getting
         :param dynamic: Specify whether the property parameters are meant to be changed in
             instances or subclasses.
+        :param \\**kwargs: Keyword arguments for :meth:`values`.
 
         Example of usage of dynamic parameter is as follows:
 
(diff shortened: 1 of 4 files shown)


pymeasure/instruments/common_base.py
```diff
@@ -504,6 +505,7 @@ def measurement(get_command, docs, values=(), map_values=None,
         :param check_get_errors: Toggles checking errors after getting
         :param dynamic: Specify whether the property parameters are meant to be changed in
             instances or subclasses. See :meth:`control` for an usage example.
+        :param \\**kwargs: Keyword arguments for :meth:`values`.
```
bilderbuchi, line 508: `setting()` will also need that addition I think.

bmoneke: As far as I know, does setting **set** and not get, therefore it does not invoke `values`.

bilderbuchi: Right. Should we then remove the `kwargs` from the signature of `setting`, or trigger a warning inside `setting` if kwargs is not None?

bmoneke: I'm for removing it from the signature as we are not talking about a specific kwarg, but the general collection of all additionally used ones.

bmoneke: Done.

bilderbuchi: Thanks. Changelog entry "Unused **kwargs argument in Instrument.settings removed" or so)? ([Hyrum's law and all that](https://www.hyrumslaw.com/))

bmoneke: I'll add an entry. Thanks for the link.

bmoneke: Good, that we removed the "kwargs" call, as several tests failed, because the gave "get_process" to a setting etc.
Also the documentation failed with doctests...

bmoneke: I see, that `setting` and `measurement` have different signatures (as they only require subsets of `control`).
If we keep `kwargs` in the signature, you can change more easily from one property creator type to the other, as superfluos kwargs do not stop it working.

So, instead of removing kwargs, we add a note, that they are dropped?

bilderbuchi: > Good, that we removed the "kwargs" call, as several tests failed, because the gave "get_process" to a setting etc.

It's good that these things are uncovered!

> If we keep kwargs in the signature, you can change more easily from one property creator type to the other, as superfluos kwargs do not stop it working.

Is that desirable, though? If you change factory method, you should also adapt the argument to the appropriate ones. Removing **kwargs would highlight the ones that need changing, no?

bmoneke: From a small "docstring and signature PR" to a large, multi file PR 🤦

bilderbuchi: Well, such is life. 😂 you go where the journey leads you...
Also nicely demonstrates how small changes in the core can have amplified impact... 😁

bmoneke: I'll do it. So stay tuned for more changes...

bmoneke:
In the end, there were only a few instruments (three), which needed changes.
At any instruments import, all instruments are imported, therefore all tests failed.
So a fast fix.

@wichers, could you please verify, that the changed `settings` properties of the Advantest RX are indeed settings and not `control` in reality?

bmoneke:
@bilderbuchi I'd say we merge and open an issue (#854), whether the modified `settings` properties of the Advantest should be `control` instead.
This PR does not break functionality, I commented out these useless parameters.

bilderbuchi:
Sounds good! 
I just realised, this caught validator/validators typos, too! :tada:

PR #844, opened 2023-02-01
Turn Oxford adapter into parent class.

### LINKED ISSUE #842: Turn Oxford Adapter into parent class
The last individual Adapter is the Oxford instruments' adapter.

In order to make it homogeneous and improve test possibilites, it should be changed to a parent instrument class.

The adapter is already VISAAdapter based, so the conversion should be straight forward.

bmoneke:
Closes #842

The first commit is just a renaming of a file, in order to see a more meaningful diff from the Adapter to the Instrument subclass.

Note, that I moved some connection settings from the individual classes to this new base class.


bilderbuchi:
Do you want to add some strategic protocol tests as a first step, TDD style, to ensure that the refactoring did not inadvertently change the communication behaviour?

bmoneke:
> Do you want to add some strategic protocol tests as a first step, TDD style, to ensure that the refactoring does not inadvertently change the communication behaviour?

I tried to, but I did not know the special cases, for which the Adapter was written 🤷 

In the end I thought, that reducing it to a renaming with some small changes should work well enough, as I keep all the "special" stuff as is.

bmoneke:
Here again we have the problem, how to write a test, if you do not know, how the communication should look like.

bilderbuchi:
Write a (protocol) test (that will fail) that touches the area you want to change, observe the communication that happened at the current situation, codify that in the test.
Then, change the implementation and observe that the exchanged messages remained the same.

pymeasure/instruments/oxfordinstruments/base.py
```diff
@@ -38,25 +38,47 @@ class OxfordVISAError(Exception):
...
+        kwargs.setdefault('read_termination', '\r')
+
+        super().__init__(adapter,
+                         name=name,
+                         includeSCPI=False,
+                         asrl={
+                             'baud_rate': 9600,
+                             'data_bits': 8,
+                             'parity': 0,
+                             'stop_bits': 20,
```
bklebel, line 69:
```suggestion
                             'parity': vconst.Parity.none,
                             'stop_bits': vconst.StopBits.two,
```
at least for me this is necessary, I am going to check with the latest version of `visa`

pymeasure/instruments/oxfordinstruments/base.py
```diff
@@ -23,7 +23,7 @@
 #
 
 
-from pymeasure.adapters import VISAAdapter
+from pymeasure.instruments import Instrument
 from pyvisa.errors import VisaIOError
```
bklebel, line 28:
```suggestion
from pyvisa import constants as vconst
```
needed for the proper asrl options for parity and stop bits

pymeasure/instruments/oxfordinstruments/base.py
bklebel: newest version of pyvisa and visa, and this works, no idea how it could have worked without it

bklebel (changes requested):
Lucky here, that I have those instruments in the laboratory, and funnily enough, we are now again looking into it. Everything works like a charm, except the things where I suggested changes - I am not sure how it could have worked before, using `parity: 0` instead of the pyvisa constants module, but like this, at least the communication with the `ps120_10` works fine

bklebel:
Ok, I need to go back on my word, I will continue to test this a bit more...

bklebel:
Since @bmoneke introduced the `values` method with directly casting to `float`, this breaks a part of the `PS_custom_get_process` as it is currently implemented, because this still assumes that we do neither use a `preprocess_reply` nor the `cast`. 

for this to continue working with the PS120_10, you need to change
~~~python 
def PS_custom_get_process(v):
    """convert string to proper float value, for working with the PS 120-10 """
    return float(v[1:]) * 1e-2
~~~
in ps120_10.py to 
~~~python 
def PS_custom_get_process(v):
    """convert string to proper float value, for working with the PS 120-10 """
    return v * 1e-2
~~~

Since you're not touching this file in this PR right now, I cannot suggest this change here.

pymeasure/instruments/oxfordinstruments/base.py
bmoneke: Because these constants have the following values (in my pyvisa installation):
~~~
vconst.Parity.none = 0
vconst.StopBits.two = 20
~~~

bmoneke:
Thanks for your testing.
I am not sure, whether I added the float casting. I think it was there already, when I started using pymeasure.

It would be great, if you could provide some example communication pairs (sets of messages exchanged between computer and instrument), such that I may write tests.

bklebel:
No, the float casting itself was always there, but the special decimal point shift for the PS (compared to theIPS) did not reach the `values` method of `Instrument`, since this was handled on the adapter level, and float casting was only applied afterwards. 
Now that we handle this on the `Instrument` level, wee need to take it into account. 

Yes, I can provide such communication pairs, I think, although the numbers in the answer will depend on the internal state of the device, not sure how best to deal with that. Is there already a specified place for such communication protocol data, or do I just comment it here?

pymeasure/instruments/oxfordinstruments/base.py
bklebel: yes, I see that it was so, I am not sure how to deal with that. I am using

PyVISA==1.12.0
visa==1.0.0

But I am not 100% sure whether I am now using the NI visa library or the Keysight one, and even so I am not sure whether that would make a difference here

bmoneke:
There is no test file yet (only locally without much content).
- For full tests, you could open a PR to my branch or comment them here, such that I put them in the test files.
- Alternatively, I could create stub test files and you give "suggestions"
- Or, you just give the comm pairs (and the corresponding command etc.) and I write the tests

pymeasure/instruments/oxfordinstruments/base.py
bklebel: Ok, so, I am confused now. With clean environments that only contain what is needed I can run it with the numbers instead of the `vconst.Parity.none` things. With the production environment which my setup is running on, it does not work -- so this is related to my system, not to visa versions. With the visa constants, it works in all cases for me. Not sure what would be better to keep

bmoneke: We need the visa system anyways, so I do not see any problem using the constants. They are better to read anyway (none compared to 0 and two compared to 20).
If the constants work with any backend, it is one reason more, to use them.

bklebel:
I have a few small debugging changes, which I would like to do a PR to your branch. I wasn't able to base a branch of my own on your fork, could you please sync your branch with pymeasure:master, so I do not bring in the new changes from there? 
For the tests, I will need some more time, but this now is needed for the `PS120-10` to actually control a magnet, something got screwed up there on the way

bmoneke:
I merged the master branch into mine.

bmoneke:
I added a few tests according to a manual I found (https://www.jlab.org/div_dept/physics_division/dsg/technical_documentation/HDice/Manuals_and_Specifications/Oxford%20IPS120-10.pdf)

However, one test fails as I do not know, what the right response is:
- If the device responds "R1.3", then the property returns this string and not a number
- If the device responds "1.3", then the "is_valid_response" complains.

Without knowledge of correct responses, I cannot write more tests.

bklebel:
I cannot fully comment on the IPS, we do only have the PS, and its response might be slightly different. 
the PS always adds a sign, so I would get "R+130" which should be turned into `1.3`. The response will always prepend the requests character, so for sth where you ask with e.g. "R0", you will get back a "R+INTEGER" for the PS, and for the IPS most likely a "R+FLOAT", but I am not 100% sure about the latter.

bmoneke:
"R+130" does not work in the PS120 (I added tests), as it will hand the string "R+130" to the get_process (dividing by 100), which cannot work (dividing string by int).

Good working with your paper!

Thanks for joining in the conversation, despite the paper.

bklebel:
does the test really take the `values` function? There you have the preprocess_reply, which takes away e.g. the "R", and the cast to float. 
for example, I get the following, which I want to see: 
```
>>> ps.field_setpoint
2023-02-28 16:04:01,089 - DEBUG - Adapter - write - WRITE:R8
2023-02-28 16:04:01,106 - DEBUG - Adapter - read - READ:R+00100
1.0
```
Maybe tomorrow I can put more logging messages to track how the answer is transformed from the string to the float, because I definitely get a float back as type

tests/instruments/oxfordinstruments/test_ps120_10.py
```diff
@@ -0,0 +1,80 @@
...
+    with expected_protocol(PS120_10,
+                           [("A0", "A")]
+                           ) as inst:
+        inst.activity = "hold"
+
+
+@pytest.mark.xfail()  # TODO this test fails, the correct response has to be added.
+def test_current_setpoint_getter():
+    with expected_protocol(PS120_10,
+                           [("R0", "R+130")]
```
bklebel, line 57:
```suggestion
                           [("R0", "R+00130")]
```

bmoneke: I already ajusted the test according to your log entries, but that is not the issue.

bmoneke:
I found the problem:

- previously, `preprocess_reply` was given into `kwargs` of `control` and then to `values.
- In #856 , we replaced the `kwargs` in `control` with explicit parameters (and default values), among them `preprocess_reply`
- Consequently, the properties call `values` with `preprocess_reply=None` and thus they ignore the default value in `values`' signature.

bmoneke:
Thank you very much for your log entry and support.

bmoneke:
I fixed the bug (now `ask` removes the leading "R").
Tests are working fine and the code is ready, I think.

I think the tests are already sufficient for catching basic working (you might add more if you whish, could be even in a separate PR).

Finally we can close into completion.

pymeasure/instruments/oxfordinstruments/base.py
```diff
@@ -72,9 +86,15 @@ def ask(self, command):
         """
 
         for attempt in range(self.max_attempts):
-            response = super().ask(command)
+            # Skip the checks in "write"
```
bklebel, line 89:
```suggestion
            # Skip the checks in "write", because we explicitly want to get an answer here
```

bklebel:
Okay great, very nice that you spotted this bug! 
Right now, I am just curious as to where I could see the failing tests in the CI, as far as I saw it, the checks here on GitHub all passed while the problem was still inside. 
I was shortly confused about the introduced skipping of `write` checks in `ask`, I added a longer comment to make that more explicit - essentially, the `if` in `write` would always evaluate to `False` in the case of `ask`ing something from the instrument, except if a user explicitly tried (and use `ask` manually), and prepended the dollar sign to some command, but I guess it is fine to get hiccups when doing something like this. 

Looks mergable to me.

bmoneke:
> Right now, I am just curious as to where I could see the failing tests in the CI, as far as I saw it, the checks here on GitHub all passed while the problem was still inside.

I marked the tests with `xfail`, that I expect them to fail and as such, the whole test passed.

bilderbuchi:
> I found the problem:
> 
>     * previously, `preprocess_reply` was given into `kwargs` of `control` and then to `values.
> 
>     * In [Remove control kwargs #856](https://github.com/pymeasure/pymeasure/pull/856) , we replaced the `kwargs` in `control` with explicit parameters (and default values), among them `preprocess_reply`
> 
>     * Consequently, the properties call `values` with `preprocess_reply=None` and thus they ignore the default value in `values`' signature.

I'm not clear: Is this a bug we caused with #856, and our tests didn't catch this? 
If yes, can it also happen for other instruments? Can we do something to catch the impact of changes in the future?

bilderbuchi (approved):
LGTM, but I think you and @bklebel have this well in hand.

bmoneke:
> 'm not clear: Is this a bug we caused with https://github.com/pymeasure/pymeasure/pull/856, and our tests didn't catch this? 

Yes, we caused it with that PR, but only for this PR, not for the actual instrument, which uses its own adapter with funny methods.

> (I don't see an xfail in the diff in this PR, so I might not be picking up necessary context)

Commit 96d0cfeaf517fa6cf977c89d0ffdb23eb647c6d3 removes xfail in both test files.

> If yes, can it also happen for other instruments?

I checked, no other instrument in our code base overrides the default parameters of `values`, so no other pymeasure instrument is affected.

> Can we do something to catch the impact of changes in the future?

Having tests for instruments. In this case a single test would have been sufficient (for a "normal property", which does not override `get_process`). So it is really important to test all the "special stuff" in an instrument: If you write your own `values`, for example: put a few tests in place, that ensure, that the instrument continues to work as expected.

Issue #172, opened 2019-01-04
Add Prologix gpib to ethernet adaptor 


mbrad092:
Also the VXI-11 adapter isn't present in the pip version of this package. Thanks for your efforts, really enjoying the package!

thosou:
Hi @mbrad092,

I actually also use the ethernet to gpib prologix adapter! I made a personal quick and dirty solution based on what has been done with the serial prologix. I had in mind to open a discussion and see how we could add this contribution.
Regarding the VXI-11 adapter, I think it will be probably available in the next release. until then you can always install from github.

mbrad092:
Oh lol very nice! Any chance I could ask you for to send me your quick and dirty solution in the interim before the contribution is finalized?

thosou:
Hi @mbrad092,

You can find the code here:
https://github.com/thosou/pymeasure/blob/dev/prologix-ethernet/pymeasure/adapters/prologix.py

You can use this adapter the same way as the `PrologixAdapter`
Let me know if you get it to work!

cjermain:
@mbrad092, thanks for reminding me to make a new release! I've published version 0.6 which now has the VXI-11 adapter.

@thosou, it would be great to add your PrologixEthernetAdatper to PyMeasure.

bmoneke:
@msmttchr 
I compared the code @thosou mentioned to our current PrologixAdapter. It should be quite simple to add some good ideas (making auto, eoi, eos configurable and readable).
That way we make the Adapter capable to use the Ethernet version (which seems to use different eoi and eos setting, at least as default) as well.

msmttchr:
I think this is a good idea 😃

bmoneke:
Whoever wants to use the Prologix Adapter with ethernet: the code should work.
Please tell us, if not.

BenediktBurger:
It should work.
I'll close this issue.
Whenever someone has an issue with that one, he might raise a bug issue.

PR #845, opened 2023-02-01
Prologix settings are configurable.

bmoneke:
Incorporate properties from https://github.com/thosou/pymeasure/blob/dev/prologix-ethernet/pymeasure/adapters/prologix.py by @thosou

given in #172 

Only the `read` methods differ (++read vs ++read eoi), could you shed light on this question @msmttchr ?

### DIFF
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -55,9 +55,9 @@ class PrologixAdapter(VISAAdapter):
         .. deprecated:: 0.11
             Implement it in the instrument's `read` method instead.
 
-    :param auto: Default to 0 to turn off read-after-write and address instrument to listen.
-    :param eoi: Enable (1) or disable (0) EOI assertion.
-    :param eos: Set command termination 0 for CR+LF; 2 for LF.
+    :param auto: Enable or disable read-after-write and address instrument to listen.
+    :param eoi: Enable or disable EOI assertion.
+    :param eos: Set command termination string ("\r\n", "\r", "\n", or "")
     :param kwargs: Key-word arguments if constructing a new serial object
 
     :ivar address: Integer GPIB address of the desired instrument.
@@ -90,7 +90,7 @@ class PrologixAdapter(VISAAdapter):
     """
 
     def __init__(self, resource_name, address=None, rw_delay=0, serial_timeout=None,
-                 preprocess_reply=None, auto=0, eoi=1, eos=2, **kwargs):
+                 preprocess_reply=None, auto=False, eoi=True, eos="\n", **kwargs):
         # for legacy rw_delay: prefer new style over old one.
         if rw_delay:
             warn(("Parameter `rw_delay` is deprecated. "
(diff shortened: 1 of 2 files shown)


pymeasure/adapters/prologix.py
```diff
@@ -55,6 +55,9 @@ class PrologixAdapter(VISAAdapter):
         .. deprecated:: 0.11
             Implement it in the instrument's `read` method instead.
 
+    :param auto: Default to 0 to turn off read-after-write and address instrument to listen.
```
msmttchr, line 58:
```suggestion
    :param auto: Enable (1) or disable (0)  read-after-write and address instrument to listen.
```

pymeasure/adapters/prologix.py
```diff
@@ -55,6 +55,9 @@ class PrologixAdapter(VISAAdapter):
...
+    :param eoi: Enable (1) or disable (0) EOI assertion.
+    :param eos: Set command termination 0 for CR+LF; 2 for LF.
```
msmttchr, line 60:
```suggestion
    :param eos: Set command termination 0 for CR+LF; 1 for CR, 2 for LF, 3 for none.
```

pymeasure/adapters/prologix.py
```diff
@@ -107,15 +110,66 @@ def __init__(self, resource_name, address=None, rw_delay=0, serial_timeout=None,
...
+            self.auto = auto
+            self.eoi = eoi
+            self.eos = eos
+
+    @property
+    def auto(self):
+        """Configure Prologix GPIB controller to automatically address instruments
+        to talk after sending them a command in order to read their response. The
+        feature called, Read-After-Write, saves the user from having to issue read commands
+        repeatedly. This property enabled or disabled the Read-After-Write feature.
```
msmttchr, line 122:
```suggestion
        repeatedly. This property enables or disables this feature.
```
I think also valid inputs values and their meaning should be specified in the doc string.

pymeasure/adapters/prologix.py
```diff
@@ -107,15 +110,66 @@ def __init__(self, resource_name, address=None, rw_delay=0, serial_timeout=None,
...
+        """
+        self.write("++auto")
+        return int(self.read())
+
+    @auto.setter
+    def auto(self, value):
+        self.write("++auto {}".format(value))
```
msmttchr, line 129:
```suggestion
        self.write(f"++auto {value}")
```
I think we mostly used this compact form in the code, hence the suggestion. If agreed there are also few other cases below.

pymeasure/adapters/prologix.py
```diff
@@ -107,15 +110,66 @@ def __init__(self, resource_name, address=None, rw_delay=0, serial_timeout=None,
...
+
+    @property
+    def eoi(self):
+        """Enables or Disables the assertion of the EOI signal with the last character
```
msmttchr, line 133: I think also valid inputs values and their meaning should be specified in the doc string.

pymeasure/adapters/prologix.py
```diff
@@ -107,15 +110,66 @@ def __init__(self, resource_name, address=None, rw_delay=0, serial_timeout=None,
...
+    def eoi(self, value):
+        self.write("++eoi {}".format(value))
+
+    @property
+    def eos(self):
+        """Specify GPIB termination characters. When data from host is received,
+        all non-escaped LF, CR and ESC characters are removed and GPIB
+        terminators, as specified by this command, are appended before sending the data to
+        instruments. This command does not affect data from instruments received over GPIB
+        port.
```
msmttchr, line 150: I think also valid inputs values and their meaning should be specified in the doc string.

msmttchr (commented):
Some minor comments on the docstrings.

pymeasure/adapters/prologix.py
```diff
@@ -191,6 +245,7 @@ def _read(self, **kwargs):
         :param kwargs: Keyword arguments for the connection itself.
         :returns str: ASCII response of the instrument (excluding read_termination).
         """
+        # TODO Prologix Ethernet version has only `write("++read")`
```
msmttchr, line 248: You mean in the code of @thosou ?

msmttchr:
> Only the `read` methods differ (++read vs ++read eoi), could you shed light on this question @msmttchr ?

Not sure, I would think is linked to instrument behaviour and so I have no valid explanation.
It is possible that this change was unintentional and left over after debugging.

bmoneke:
Thanks for your review @msmttchr 
I changed the docstrings to pymeasure style ("Control", "Get"...), at the same time, I made the properties accept useful settings (True/False or a termination string).

As I wrote the tests, I noticed, that I have to adjust the `read` method as well: I think, if you try to read the auto setting with `++auto`, you do not have to send `++read eoi`, don't you? Not the read method can read the prologix adapter itself.

bmoneke:
Feel free to merge, if you agree, @msmttchr

msmttchr:
> Thanks for your review @msmttchr I changed the docstrings to pymeasure style ("Control", "Get"...), at the same time, I made the properties accept useful settings (True/False or a termination string).
> 
> As I wrote the tests, I noticed, that I have to adjust the `read` method as well: I think, if you try to read the auto setting with `++auto`, you do not have to send `++read eoi`, don't you? Now the read method can read the prologix adapter itself.

Indeed, you are right. `read` and `write` method are normally dedicated to drive instruments.
From a different perspective, it is possible to define a set of API dedicated to the adapter like:
- `write_adapter`: writes to adapter. In the current implementation, this is achieved using the regular `write` and the `write_adapter` is performed when the data to be written starts with '++'.
- 'read_adapter': reads from adapter. This is achieved with some extra parameter called `prologix`.

I would be inclined to think that this proposal has the drawback to add two more methods, but it would be more clean. In any case not a strong opinion, I just wanted to double check your opinion.

bmoneke:
For logging purposes, read/write_adapter have to pass through the normal read/write methods and we need this `prologix` parameter anyway, so:
~~~python
def read_prologix(self):
    """Read the prologix adapter itself."""
    return self.read(prologix=True)

def write_prologix(self, command):
    """Write a command to the prologix adapter itself, prepending '++'."""
    self.write("++" + command)
~~~

If these additional methods suits you, I'll add the two methods and make all the internal commands (auto, eoi, "++ read"...) use these new methods.

I'm indifferent for their addition.

msmttchr:
> 

Thanks for your kind feedback, I am voting for leaving things as they are, so I will merge it.

msmttchr (approved):
Ok for me. Thanks

Issue #833, opened 2023-01-13
Inconsistent ValueError related to T3AFG instrument check_set_errors


scandey:
It seems like every so often, check_set_errors leads to a portion of the outgoing command being fed back into the error checker. 

More details forthcoming tomorrow, I just wanted to track this here in case it turns out to be a code issue rather than an instrument implementation issue. 

```
  File "<input>", line 1, in <module>
  File "/Users/scottcandey/Documents/TRACERS/tracers_pygse/__pypackages__/3.10/lib/pymeasure/instruments/common_base.py", line 204, in __setattr__
    super().__setattr__(name, value)
  File "/Users/scottcandey/Documents/TRACERS/tracers_pygse/__pypackages__/3.10/lib/pymeasure/instruments/common_base.py", line 88, in __set__
    self.fset(obj, value, **kwargs)
  File "/Users/scottcandey/Documents/TRACERS/tracers_pygse/__pypackages__/3.10/lib/pymeasure/instruments/common_base.py", line 470, in fset
    self.check_errors()
  File "/Users/scottcandey/Documents/TRACERS/tracers_pygse/__pypackages__/3.10/lib/pymeasure/instruments/channel.py", line 110, in check_errors
    return self.parent.check_errors()
  File "/Users/scottcandey/Documents/TRACERS/tracers_pygse/__pypackages__/3.10/lib/pymeasure/instruments/instrument.py", line 207, in check_errors
    if int(err[0]) != 0:
ValueError: invalid literal for int() with base 10: 'C1:BSWV WVTP'
```

scandey:
I realized this is likely due to an automated process running in the same software that regularly requests the status of the function generator approximately every half second. 

If I manually set a control that happens to collide in time with the status check, the error checking function receives the result of the status check query rather than the error query. That probably also explains some of the occasional timeout errors I was seeing.

Probably running into awkward thread issues in my application, rather than anything specific to pymeasure.

bmoneke:
Oh yes, receiving the answer to another request due to multithreading...

bilderbuchi:
Yeah, this is essentially #506

PR #847, opened 2023-02-02
Removing Teledyne T3AFG check_set_errors

scandey:
After testing with the instrument over the past few weeks I've repeatedly run into thread safety issues (see [833](https://github.com/pymeasure/pymeasure/issues/833) and [506](https://github.com/pymeasure/pymeasure/issues/506)). In this case, I don't think check_set_errors really adds anything to pymeasure implementation of this particular instrument. By removing it, I reduce the chance of a user getting an opaque fatal error associated with setting and measuring from the same instrument at the same time.

I also adjusted the tests to match.

### DIFF
--- a/pymeasure/instruments/teledyne/teledyneT3AFG.py
+++ b/pymeasure/instruments/teledyne/teledyneT3AFG.py
@@ -79,7 +79,6 @@ class SignalChannel(Channel):
         values=[0, 350e6],
         get_process=get_process_generator_search('FRQ', 'HZ', float),
         dynamic=True,
-        check_set_errors=True,
     )
 
     amplitude = Channel.control(
@@ -93,7 +92,6 @@ class SignalChannel(Channel):
         values=[-5, 5],
         get_process=get_process_generator_search('AMP', 'V', float),
         dynamic=True,
-        check_set_errors=True,
     )
 
     offset = Channel.control(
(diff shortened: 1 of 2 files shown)


bilderbuchi:
To be honest, I'd rather we fix #506 properly instead of reducing functionality of our instruments. 
We need to find out how to properly write tests for threading issues.

scandey:
I think one conceptual issue is that when I think of a given set action, I think it is atomic. However, the `check_set_errors` makes a set action involve not just one command operation but many query operations. It is hard to be sure a given set action has completed if it then times out on a query (even though the command itself worked fine). 
I personally can't use the instrument implementation the way I wrote it in practice because of the risk of threading issues. I'd really prefer to remove the `check_set_errors` from this instrument until there's a formalized solution to the threading thing.

bmoneke:
Your threading issue is quite special (it depends on the threading), so I'm not sure we have to remove the error checking for all users.
The good thing is, these properties are already dynamic, such that you can drop the check_errors dynamically (either in a subclass or for the created instrument), e.g:
~~~python
inst = Te3AFG(adapter)
for channel in inst.channels:
    channel.frequency_check_set_errors = False
    channel.amplitude_check_set_errors = False
    channel.offset_check_set_errors = False
~~~

bilderbuchi:
I think we should be able to preserve this atomicity you expect (and I agree here) with a good solution for #506. I understand your desire to keep going in the meantime until such a solution lands. @bmoneke's proposed workaround sounds like a viable interim solution, do you agree?

scandey:
The work around works for me, thank you!

PR #850, opened 2023-02-08
Save plot settings in addition to dock layout

mcdo0486:
From talking with users about recently merged save functionality for `DockWidget`, it was clear that saving the plot settings, like log transforms on X axis, was a desired feature. That can be done easily with `pg.PlotWidget.saveState()` and `pg.PlotWidget.restoreState(state)`, so I added that functionality.

Now when saving the dock layout, the settings for each plot are saved as well.


CasperSchippers (approved):
Seems like a nice and simple addition; thanks. I'll merge it right away.

PR #851, opened 2023-02-09
New instrument: IPG Photonics YAR Amplifier series

bmoneke:
A new instrument definition.

### DIFF
--- a/pymeasure/instruments/ipgphotonics/yar.py
+++ b/pymeasure/instruments/ipgphotonics/yar.py
@@ -29,19 +29,15 @@ from pyvisa.constants import Parity, StopBits
 
 
 def emission_validator(value, values):
-    if value in values:
-        return value
-    elif isinstance(value, str):
-        if value.capitalize() in values:
-            return value.capitalize()
-        else:
-            raise ValueError(f"Value is {value}, but 'ON' or 'OFF' required.")
-    elif value is True:
+    if value is True:
         return "ON"
     elif value is False:
         return "OFF"
-    else:
-        raise ValueError(f"Value is {value}, but 'ON' or 'OFF' required.")
+    elif isinstance(value, str):
+        value = value.upper()
+        if value in values:
+            return value
+    raise ValueError(f"Value is {value}, but a boolean or 'ON' or 'OFF' required.")
 
 
 def setpoint_validator(value, values):
--- a/pymeasure/instruments/ipgphotonics/yar.py
+++ b/pymeasure/instruments/ipgphotonics/yar.py
@@ -33,10 +33,6 @@ def emission_validator(value, values):
         return "ON"
     elif value is False:
         return "OFF"
-    elif isinstance(value, str):
-        value = value.upper()
-        if value in values:
-            return value
     raise ValueError(f"Value is {value}, but a boolean or 'ON' or 'OFF' required.")
 
 
@@ -121,7 +117,7 @@ class YAR(Instrument):
         got = int(self.values("STA")[0])
         return self.Status(got)
 
-    emission = Instrument.control(
+    emission_enabled = Instrument.control(
         "STA",
         "EM%s",
         """Control emission of the amplifier (bool).""",
--- a/pymeasure/instruments/ipgphotonics/yar.py
+++ b/pymeasure/instruments/ipgphotonics/yar.py
@@ -124,7 +124,7 @@ class YAR(Instrument):
     emission = Instrument.control(
         "STA",
         "EM%s",
-        """Control emission of the amplifier.""",
+        """Control emission of the amplifier (bool).""",
         cast=int,
         values=("ON", "OFF"),
         validator=emission_validator,
--- a/pymeasure/instruments/ipgphotonics/yar.py
+++ b/pymeasure/instruments/ipgphotonics/yar.py
@@ -1,0 +1,182 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2023 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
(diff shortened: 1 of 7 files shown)


pymeasure/instruments/ipgphotonics/yar.py
```diff
@@ -0,0 +1,182 @@
...
+from pymeasure.instruments import Instrument, validators
+from pyvisa.constants import Parity, StopBits
+
+
+def emission_validator(value, values):
+    if value in values:
+        return value
+    elif isinstance(value, str):
+        if value.capitalize() in values:
+            return value.capitalize()
```
bilderbuchi, line 36: I think the first clause is redundant, as it's fully covered by the second?

bilderbuchi:
Not sure if you are already looking for a review, but I skimmed this and it looks good to me. (what else? :smile: )

bmoneke:
You are invited to review.
I just did not want to pressure you to review.

From my side, the driver is ready for review.

pymeasure/instruments/ipgphotonics/yar.py
bmoneke: I streamlined the validator (first checks for boolean values, as that is the preferred way to interact).
With that change, I noticed a bug with the tests: it should be `upper`, not `capitalize`.

Now I test for a variety of input parameters.

pymeasure/instruments/ipgphotonics/yar.py
```diff
@@ -0,0 +1,178 @@
...
+
+    emission = Instrument.control(
+        "STA",
+        "EM%s",
+        """Control emission of the amplifier.""",
+        cast=int,
+        values=("ON", "OFF"),
+        validator=emission_validator,
+        get_process=lambda v: bool(v & YAR.Status.EMISSION),
+        check_set_errors=True,
```
LongnoseRob, line 132: Would it not make sense to map "ON" and "OFF" to `True` and `False` ?

LongnoseRob (approved):
Hello, 
looks like very much straight forward implementation.
I have only one comment on the `emission` propery,

And a more general question: does this device not have soem kind of interlock-circuit?

pymeasure/instruments/ipgphotonics/yar.py
bmoneke: `lambda v: bool(v & YAR.Status.EMISSION)`

does exactly that:
1. it reads the emission bit from the status byte (there is no dedicated "read emission command")
2. turn the bit to a boolean (True or False

On setting, I use the "emission_validator" to turn a boolean or "ON", "OFF" to the expected strings "ON" or "OFF".
So, you may use a boolean as well.

bmoneke:
> And a more general question: does this device not have soem kind of interlock-circuit?

Yes it has it, but you cannot do anything via remote command with that interlock.

pymeasure/instruments/ipgphotonics/yar.py
LongnoseRob: Oh yes, I did not realize that.

bmoneke: I improved the docstring, stating that it has a boolean value.

LongnoseRob: LGTM

pymeasure/instruments/ipgphotonics/yar.py
bilderbuchi: thanks!

pymeasure/instruments/ipgphotonics/yar.py
```diff
@@ -0,0 +1,178 @@
...
+    @property
+    def status(self):
+        """Get the current status."""
+        got = int(self.values("STA")[0])
+        return self.Status(got)
+
+    emission = Instrument.control(
+        "STA",
+        "EM%s",
+        """Control emission of the amplifier (bool).""",
```
bilderbuchi, line 127: In line with our guidelines, I think `emission_enabled` might be a better name for this boolean property.

tests/instruments/ipgphotonics/test_yar.py
```diff
@@ -0,0 +1,233 @@
...
+def test_clear():
+    with expected_protocol(
+            YAR,
+            init_comm + [("RERR", "RERR")]
+    ) as inst:
+        inst.clear()
+
+
+def test_model():
+    """This test tests already getting min/max output power."""
```
bilderbuchi, line 211: docstring of this and the next test probably copy/pasted?

tests/instruments/ipgphotonics/test_yar.py
```diff
@@ -0,0 +1,233 @@
...
+
+def test_emission_getter():
+    with expected_protocol(
+            YAR,
+            init_comm + [("STA", "STA:1")]
+    ) as inst:
+        assert inst.emission is True
+
+
+@pytest.mark.parametrize("value", (True, "ON", "on", "On", "oN"))
```
bilderbuchi, line 71: In my opinion, we should strive for conciseness/clarity/symmetry in our interfaces. I can see the attraction of being very "general" in what to accept (and in principle that's known [a good thing](https://en.wikipedia.org/wiki/Robustness_principle)). 
However, wouldn't it be preferable to just have `True`/`False` (and the equivalent truthy things that Python accepts) as the interface for this property? 
Accepting all kinds of variations of "on" could be a slippery slope towards [YAML](https://yaml.org/type/bool.html), which accepts  
``` 
y|Y|yes|Yes|YES|n|N|no|No|NO
|true|True|TRUE|false|False|FALSE
|on|On|ON|off|Off|OFF
```
Imo, this is not something we should mimick. Curious to hear your thoughts.

bmoneke: I agree: bool is sufficient.

Looking at other people's code I notice it, too, but changing my own code (predating my pymeasure time), I overlooked it... For that reason we have reviews.

tests/instruments/ipgphotonics/test_yar.py
bmoneke: Yes, Done.

pymeasure/instruments/ipgphotonics/yar.py
bmoneke: Ok, done.

bmoneke:
I resolved your issues, @bilderbuchi , can we merge?

bilderbuchi:
Yes, thank you!

PR #856, opened 2023-02-13
Remove control kwargs

### LINKED ISSUE #855: Do not allow kwargs in property creators
We noticed, that removing kwargs from `setting` caught some errors. ( #843)

We could deprecated the usage of "kwargs" in control and measurement, adding a "v_kwargs" parameter, which is given to values as kwargs.

Additionally, all parameters of "values" have to be added to the signature as well. (spyder already composes the signature, including called methods, too) 

That way we keep flexibility (you can hand on arbitrary arguments to values and then to ask), but we catch many errors in normal cases.

What do you think?

bmoneke:
Closes #855 

This PR deprecates kwargs in the property creators in order to catch wrong used arguments, for example validators in a `measurement`.

The raised warning helped to remove some superfluous code.

This warning (and later TypeError) will reduce review effort.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,7 +13,7 @@ Deprecated features
 - Toptica IBeamSmart: :code:`channel1_enabled`, use :code:`ch_1.enabled` property instead (similar channel2). Also :code:`laser_enabled` is deprecated in favor of :code:`emission`.
 - TelnetAdapter: use :code:`VISAAdapter` instead. VISA supports TCPIP connections. Use the resource_name :code:`TCPIP[board]::<hostname>::<port>::SOCKET` to connect to a server.
 - Attocube ANC300: :code:`host` argument, pass a resource string or adapter as :code:`Adapter` passed to :code:`Instrument`. Now communicates through the :code:`VISAAdapter` rather than deprecated :code:`TelnetAdapter`. The initializer now accepts :code:`name` as its second keyword argument so all previous initialization positional arguments (`axisnames`, `passwd`, `query_delay`) should be switched to keyword arguments.
-- The property creators :code:`control`, :code:`measurement`, and :code:`setting` do not accept arbitrary keyword arguments. Use the :code:`v_kwargs` parameter to give further arguments to :code:`values` method.
+- The property creators :code:`control`, :code:`measurement`, and :code:`setting` do not accept arbitrary keyword arguments anymore. Use the :code:`v_kwargs` parameter to give further arguments to :code:`values` method.
 
 Version 0.11.1 (2022-12-31)
 ===========================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,7 @@ Deprecated features
 - Toptica IBeamSmart: :code:`channel1_enabled`, use :code:`ch_1.enabled` property instead (similar channel2). Also :code:`laser_enabled` is deprecated in favor of :code:`emission`.
 - TelnetAdapter: use :code:`VISAAdapter` instead. VISA supports TCPIP connections. Use the resource_name :code:`TCPIP[board]::<hostname>::<port>::SOCKET` to connect to a server.
 - Attocube ANC300: :code:`host` argument, pass a resource string or adapter as :code:`Adapter` passed to :code:`Instrument`. Now communicates through the :code:`VISAAdapter` rather than deprecated :code:`TelnetAdapter`. The initializer now accepts :code:`name` as its second keyword argument so all previous initialization positional arguments (`axisnames`, `passwd`, `query_delay`) should be switched to keyword arguments.
+- The property creators :code:`control`, :code:`measurement`, and :code:`setting` do not accept arbitrary keyword arguments. Use the :code:`v_kwargs` parameter to give further arguments to :code:`values` method.
 
 Version 0.11.1 (2022-12-31)
 ===========================
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -425,8 +425,9 @@ class CommonBase:
         considered reserved for dynamic property control.
         """
         if kwargs:
-            warn(f"Do not use keyword arguments {kwargs} as `control` parameter for the `values` method, "
-                 f"use `v_kwargs` parameter instead. docs:\n{docs}", FutureWarning)
+            warn(f"Do not use keyword arguments {kwargs} as `control` parameter "
+                 f"for the `values` method, use `v_kwargs` parameter instead. docs:\n{docs}",
+                 FutureWarning)
 
         def fget(self,
                  get_command=get_command,
@@ -548,8 +549,9 @@ class CommonBase:
                 Use `v_kwargs` dictionary parameter instead.
         """
         if kwargs:
-            warn(f"Do not use keyword arguments {kwargs} as `measurement` parameter for the `values` method, "
-                 f"use `v_kwargs` parameter instead. docs:\n{docs}", FutureWarning)
+            warn(f"Do not use keyword arguments {kwargs} as `measurement` parameter "
+                 f"for the `values` method, use `v_kwargs` parameter instead. docs:\n{docs}",
+                 FutureWarning)
             if not v_kwargs:
                 return CommonBase.control(get_command=get_command,
                                           set_command=None,
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -24,6 +24,7 @@
 
 from inspect import getmembers
 import logging
+from warnings import warn
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
@@ -295,15 +296,15 @@ class CommonBase:
         """Write a command to the instrument and return a list of formatted
         values from the result.
 
-        :param command: SCPI command to be sent to the instrument
-        :param separator: A separator character to split the string into a list
-        :param cast: A type to cast the result
+        :param command: SCPI command to be sent to the instrument.
+        :param separator: A separator character to split the string into a list.
+        :param cast: A type to cast the result.
         :param preprocess_reply: optional callable used to preprocess values
             received from the instrument. The callable returns the processed
             string.
         :param maxsplit: At most `maxsplit` splits are done. -1 (default) indicates no limit.
         :param \\**kwargs: Keyword arguments to be passed to the :meth:`ask` method.
-        :returns: A list of the desired type, or strings where the casting fails
+        :returns: A list of the desired type, or strings where the casting fails.
         """
         results = self.ask(command, **kwargs).strip()
         if callable(preprocess_reply):
@@ -348,6 +349,11 @@ class CommonBase:
         check_set_errors=False,
         check_get_errors=False,
         dynamic=False,
+        separator=',',
+        cast=float,
+        preprocess_reply=None,
+        maxsplit=-1,
+        v_kwargs={},
         **kwargs
     ):
         """Return a property for the class based on the supplied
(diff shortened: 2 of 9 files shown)


pymeasure/instruments/common_base.py
```diff
@@ -375,8 +381,18 @@ def control(  # noqa: C901 accept that this is a complex method
         :param check_get_errors: Toggles checking errors after getting
         :param dynamic: Specify whether the property parameters are meant to be changed in
             instances or subclasses.
+        :param separator: A separator character to split the string into a list.
```
bilderbuchi, line 384: which string?

pymeasure/instruments/common_base.py
```diff
@@ -375,8 +381,18 @@ def control(  # noqa: C901 accept that this is a complex method
...
+        :param cast: A type to cast the result.
```
bilderbuchi, line 385: result of what? we could specify at which point the cast happens.

pymeasure/instruments/common_base.py
```diff
@@ -375,8 +381,18 @@ def control(  # noqa: C901 accept that this is a complex method
...
+        :param preprocess_reply: optional callable used to preprocess values
+            received from the instrument. The callable returns the processed
+            string.
+        :param maxsplit: At most `maxsplit` splits are done. -1 (default) indicates no limit.
```
bilderbuchi, line 389: done when?

pymeasure/instruments/common_base.py
```diff
@@ -408,6 +424,10 @@ class SpecificInstrument(GenericInstrument):
         parameters name except `dynamic` and `docs` (e.g. `values` in the example) has to be
         considered reserved for dynamic property control.
         """
+        if kwargs:
+            warn(f"Do not use keyword arguments {kwargs} as `control` parameter "
+                 f"for the `values` method, use `v_kwargs` parameter instead. docs:\n{docs}",
+                 FutureWarning)
```
bilderbuchi, line 431: at this point you could merge/update any present kwargs into `v_kwargs` and continue, then below in the values and other calls you can already use the "new" API without **kwargs, and have less adjustment work later when removing the deprecated code.

pymeasure/instruments/common_base.py
```diff
@@ -505,8 +536,38 @@ def measurement(get_command, docs, values=(), map_values=None,
...
+            received from the instrument. The callable returns the processed
+            string.
+        :param maxsplit: At most `maxsplit` splits are done. -1 (default) indicates no limit.
+        :param dict v_kwargs: Further keyword arguments for :meth:`values`.
         :param \\**kwargs: Keyword arguments for :meth:`values`.
+
+            ..deprecated:: 0.12
+                Use `v_kwargs` dictionary parameter instead.
         """
+        if kwargs:
```
bilderbuchi, line 551: The handling of kwargs I suggested above should enable you here to avoid the duplicated code.

pymeasure/instruments/common_base.py
```diff
@@ -348,6 +349,11 @@ def control(  # noqa: C901 accept that this is a complex method
         check_set_errors=False,
         check_get_errors=False,
         dynamic=False,
+        separator=',',
+        cast=float,
+        preprocess_reply=None,
+        maxsplit=-1,
```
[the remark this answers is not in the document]
bmoneke: Yes, the v_kwargs are given to ask, but an overridden values method might have its own args.

Regarding the four explicit args: They are in pymeasure core defined. All other kwargs are added by an instrument implementor.

pymeasure/instruments/common_base.py
```diff
@@ -348,6 +349,11 @@ def control(  # noqa: C901 accept that this is a complex method
...
+        v_kwargs={},
```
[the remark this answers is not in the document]
bmoneke: Ok.
I was not sure about the name and left a first suggestion.
Values_kwargs (or ask, depending on the decision) is fine

pymeasure/instruments/common_base.py
bmoneke: It does not work (which caused major headaches with the tests... ), as the "kwargs" dict lives in the property creator.
So, when I add something to it in the first property, all other properties have that same argument...

pymeasure/instruments/common_base.py
bmoneke: Done

pymeasure/instruments/common_base.py
bmoneke: I tried to make it more clear.

pymeasure/instruments/common_base.py
bmoneke: I tried to do after the warning:
~~~python
v_kwargs.update(kwargs)
~~~

In the tests, at one point, I verify the kwarg `testing=True`. All following tests complain, that they receive an illegal argument `testing`.

My only idea is, that I change the default v_kwarg for all following usages of that static method.

I could try a temporary variable:
~~~
kw_temp = {}
if kwargs:
    warn()
    kw_temp.update(kwargs)
kw_temp.update(v_kwargs)
...
vals = self.values(..., **kw_temp)
~~~

pymeasure/instruments/common_base.py
```diff
@@ -297,12 +297,14 @@ def values(self, command, separator=',', cast=float, preprocess_reply=None, maxs
...
-            string.
-        :param maxsplit: At most `maxsplit` splits are done. -1 (default) indicates no limit.
+        :param separator: A separator character to split the string returned by
+            the device into a list.
+        :param cast: A type to cast each element of the splitted string.
+        :param preprocess_reply: Optional callable used to preprocess the string
+            received from the instrument, before splitting it.
+            The callable returns the processed string.
+        :param maxsplit: The string returned by the device is splitted at most `maxsplit` times.
+            -1 (default) indicates no limit.
```
bilderbuchi, line 307: Very nice clarifications on the argument descriptions!
To improve this even more, what do you think about adjusting their order in the order they become relevant/applied? I.e. `preprocess_reply`, `separator`, `maxsplit`, `cast`? 
It's a little thing, but could improve clarity/understanding when studying the documentation/docstring?

pymeasure/instruments/common_base.py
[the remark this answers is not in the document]
bilderbuchi: > Yes, the v_kwargs are given to ask, but an overridden values method might have its own kwargs.
> They are in the pymeasure core defined. All other kwargs are added by an instrument implementor.

I see your point/reasoning and agree. However, I still don't like that we have kwargs for values that are _not_ in `values_kwargs` - this is not very "clean". 
How about `custom_kwargs` (preferred) or `additional_kwargs`? That makes it clear that those are _not_ "core" arguments, and part of a custom/overridden values/ask.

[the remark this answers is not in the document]
bilderbuchi: > Any non empty `values_kwargs` dictionary will raise an error, if you call that property, unless an implementor changed either `ask` or `values`.

This has me confused. I thought the purpose of `values_kwargs` should be to pass its contents (e.g. `query_delay`) on towards the standard `ask` within the standard `values`? Why would that raise an error?

pymeasure/instruments/common_base.py
bilderbuchi: This is probably an effect of having these "closures" around, this is useful but can be tricky at times. 
I don't immediately see where the problem comes from, but we should be able to avoid it cleanly. 
I'll probably have to pull the branch to troubleshoot.

bilderbuchi: No wait, you have `v_kwargs={}` in the signature! This is a mutable default argument, and as such a [big no-no](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) in Python. Probably this is causing the problem, and even if not it must be avoided. Can you avoid that and re-check? E.g. use `None` and `if values_kwargs is None: values_kwargs = {}` (or other solution).

pymeasure/instruments/common_base.py
bmoneke: I thought about it, but I kept them in order of the signature.
Probably we can change the signature, but it can be dangerous (if people use args and not kwargs).

pymeasure/instruments/common_base.py
bmoneke: That is the problem (very probably). 
Thanks for finding it. Again I learned something new (or understood it better, why something with lambda failed me in the past).

pymeasure/instruments/common_base.py
bilderbuchi: We could make them keyword-only then any positional use would be an error?
I don't think I have ever seen property factories where those args were positional. For `values`, I'm not sure.

bmoneke: I was thinking about values.

That's the problem with backwards compatibility...

We can order the docs differently than the args.

With "*args" we can prevent normal arguments (or how else do you enforce kwargs?)

pymeasure/instruments/common_base.py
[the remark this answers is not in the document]
bmoneke: The standard ask does not accept any keyword arguments (except query delay).
Here again, it was for overriding reasons.

pymeasure/instruments/common_base.py
bilderbuchi: Mutable default args are not related to lambdas. 
What I think you are thinking of is lambdas and closures: see e.g. https://realpython.com/python-lambda/#closure and https://realpython.com/inner-functions-what-are-they-good-for/#retaining-state-with-inner-functions-closures ?

bmoneke: I know, that they are two different things, but they are kind of related and both appeared in the link you gave me.

pymeasure/instruments/common_base.py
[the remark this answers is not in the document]
bilderbuchi: So, this means an already present `query_delay` cannot remain where it is (as using `**kwargs` is deprecated), but cannot be moved into `custom_kwargs`/`value_kwargs` (as that will raise an error)? :confused:

pymeasure/instruments/common_base.py
bmoneke: I do not remember exactly the problem, but it was something like mutable lambda as function argument.
That did not work out...

Therefore the connection.

pymeasure/instruments/common_base.py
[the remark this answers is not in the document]
bmoneke: It can be moved to custom_args.

Query delay is the only valid kwarg in the pymeasure core.

pymeasure/instruments/common_base.py
bmoneke: > `func(posarg1, posarg2, *, kwarg1=bla, kwarg2=blubb), IIRC.`

That is, what I meant, with "*args".

I did a small test, the star works as you suggest (without assigning it to some list).

CHANGES.rst
```diff
@@ -13,6 +13,7 @@ Deprecated features
 - Toptica IBeamSmart: :code:`channel1_enabled`, use :code:`ch_1.enabled` property instead (similar channel2). Also :code:`laser_enabled` is deprecated in favor of :code:`emission`.
 - TelnetAdapter: use :code:`VISAAdapter` instead. VISA supports TCPIP connections. Use the resource_name :code:`TCPIP[board]::<hostname>::<port>::SOCKET` to connect to a server.
 - Attocube ANC300: :code:`host` argument, pass a resource string or adapter as :code:`Adapter` passed to :code:`Instrument`. Now communicates through the :code:`VISAAdapter` rather than deprecated :code:`TelnetAdapter`. The initializer now accepts :code:`name` as its second keyword argument so all previous initialization positional arguments (`axisnames`, `passwd`, `query_delay`) should be switched to keyword arguments.
+- The property creators :code:`control`, :code:`measurement`, and :code:`setting` do not accept arbitrary keyword arguments. Use the :code:`v_kwargs` parameter to give further arguments to :code:`values` method.
```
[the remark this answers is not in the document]
bmoneke: Done locally.

pymeasure/instruments/common_base.py
bilderbuchi: Yeah, this is https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html#keyword-only-arguments.
What we would be making use of:
> It also becomes possible to reorder keyword-only arguments in a function signature, since keyword-only arguments are only accessible via the name of the argument, not its position in the function signature.

Not sure if the small benefit (better order in the signature) is worth it.

bmoneke: We can do the documentation in logical order and only keep the `values` signature "disordered" (as-is), while `control` has the parameters in logical order (in sync with the documentation)

bmoneke:
I think it is ready: `values` keeps its signature as is.
The parameter is called `values_kwargs` with a default Value of "None".

With that, I hope it can be merged.

bilderbuchi (approved):
LGTM, I think we have improved this sufficiently.

PR #859, opened 2023-02-21
Added Velleman K8090 relay device

### LINKED ISSUE #858: Add support for the Velleman K8090 relay board
I would like to add support for the [Velleman K8090](https://www.velleman.eu/products/view/?id=377966), an 8-channel relay board that communicates using serial.

Like my other issue, I already have old code that I would like to port. This issue is mostly to indicate my work-in-progress.

Please let me know if this device is not popular enough and there is no interest to merge it, which I could understand.

RobertoRoos:
Fixes #858

### DIFF
--- a/docs/api/instruments/velleman/index.rst
+++ b/docs/api/instruments/velleman/index.rst
@@ -4,7 +4,7 @@
 Velleman
 ########
 
-This section contains specific documentation on the Velleman  instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
+This section contains specific documentation on the Velleman instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
 
 .. toctree::
    :maxdepth: 2
--- a/docs/api/instruments/velleman/index.rst
+++ b/docs/api/instruments/velleman/index.rst
@@ -1,0 +1,12 @@
+.. module:: pymeasure.instruments.velleman
+
+########
+Velleman
+########
+
+This section contains specific documentation on the Velleman  instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
+
+.. toctree::
+   :maxdepth: 2
+
+   k8090
--- a/pymeasure/instruments/velleman/velleman_k8090.py
+++ b/pymeasure/instruments/velleman/velleman_k8090.py
@@ -25,6 +25,7 @@
 from enum import IntFlag
 
 from pyvisa import VisaIOError
+from pymeasure.adapters import SerialAdapter, VISAAdapter
 from pymeasure.instruments import Instrument
 
 
@@ -142,11 +143,13 @@ class VellemanK8090(Instrument):
     switch_on = Instrument.setting(
         "0x11,%s",
         """"
-        Switch on a set of channels. Other channels are unaffected.  
+        Switch on a set of channels. Other channels are unaffected.
         Pass either a list or set of channel numbers (starting at 1), or pass a bitmask.
-        
-        After switching this waits for a reply from the device. Expect a blocking
-        time equal to the communication timeout!
+
+        After switching this waits for a reply from the device. This is only send when
+        a relay actually toggles, otherwise expect a blocking time equal to the
+        communication timeout
+        If speed is important, avoid calling `switch_` unnecessarily.
         """,
         set_process=_parse_channels,
         check_set_errors=True,
@@ -182,6 +185,15 @@ class VellemanK8090(Instrument):
         :param command: String like "CMD[, MASK, PARAM1, PARAM2]" - only CMD is mandatory
         :type command: str
         """
+
+        # The device can give status updates when we don't expect it,
+        # drop anything from the buffer first
+        if isinstance(self.adapter, VISAAdapter):
+            self.adapter.flush_read_buffer()
+        elif isinstance(self.adapter, SerialAdapter):
+            # The SerialAdapter does not have `flush_read_buffer` implemented
+            self.adapter.connection.flush()
+
         items_str = command.split(",")
 
         items = [int(it, 16) for it in items_str]
@@ -214,9 +226,8 @@ class VellemanK8090(Instrument):
 
         A read will return a list of CMD, MASK, PARAM1 and PARAM2.
         """
-        # Read all bytes from buffer - until the timeout
-        response = self.read_bytes(-1,break_on_termchar=True)  # `-1` causes a bug: #862
-        # Relying on the ETX byte is not possible since multiple status messages could be queued
+        # Read all bytes from buffer - until the termchar or a timeout
+        response = self.read_bytes(-1, break_on_termchar=True)
 
         if len(response) < 7:
             raise ConnectionError(f"Incoming packet was {len(response)} bytes instead of 7")
--- a/pymeasure/instruments/velleman/velleman_k8090.py
+++ b/pymeasure/instruments/velleman/velleman_k8090.py
@@ -68,10 +68,27 @@ def _ints_to_bool_lists(numbers):
 class VellemanK8090(Instrument):
     """For usage with the K8090 relay board, by Velleman.
 
-    The communication is done by serial USB.
-
     View the "K8090/VM8090 PROTOCOL MANUAL" for the serial command instructions.
+
+    The communication is done by serial USB. The IO settings are fixed:
+
+    ==================  ==================
+    Baud rate           19200
+    Data bits           8
+    Parity              None
+    Stop bits           1
+    Flow control        None
+    ==================  ==================
     """
+
+    def __init__(self, adapter, name="Velleman K8090", timeout=1000, **kwargs):
+        super().__init__(
+            adapter,
+            name=name,
+            asrl={"baud_rate": 19200},
+            timeout=timeout,
+            **kwargs
+        )
 
     BYTE_STX = 0x04
     BYTE_ETX = 0x0F
@@ -91,7 +108,7 @@ class VellemanK8090(Instrument):
         """
         Get current relay status.
         The reply has a different command byte than the request.
-        
+
         Three items (lists of 8 bools) are returned:
         - Previous state: the state of each relay before this event
         - Current state: the state of each relay now
@@ -114,19 +131,12 @@ class VellemanK8090(Instrument):
     switch_off = Instrument.setting(
         "0x12,%s",
         """"
-        Switch off a set of channels. See :prop:`switch_on`.
+        Switch off a set of channels. See :attr:`switch_on`.
         """,
         set_process=_parse_channels,
     )
 
     id = None  # No identification available
-
-    def __init__(
-        self, adapter, name="Velleman K8090", baud_rate=19200, timeout=1000, **kwargs
-    ):
-        super().__init__(
-            adapter, name=name, baud_rate=baud_rate, timeout=timeout, **kwargs
-        )
 
     def _make_checksum(self, command, mask, param1, param2):
         # The formula from the sheet requires twos-complement negation,
@@ -136,12 +146,12 @@ class VellemanK8090(Instrument):
     def switch_on_blocking(self, channels):
         """Switch on a set of channels and wait for the confirmation.
 
-        See :prop:`switch_on`.
+        See :attr:`switch_on`.
 
         The set command is identical, and the processing part of
-        :prop:`status` is called.
+        :attr:`status` is called.
 
-        :returns: See :prop:`status`
+        :returns: See :attr:`status`
         """
         mask = _parse_channels(channels)
         self.write(f"0x11,{mask}")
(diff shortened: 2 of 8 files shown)


pymeasure/instruments/velleman/velleman_k8090.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    if isinstance(numbers, int):
+        numbers = [numbers]
+
+    bool_lists = [[bool(num & (1 << n)) for n in range(8)] for num in numbers]
+
+    if len(bool_lists) == 1:
+        return bool_lists[0]
+
+    return bool_lists
```
bmoneke, line 65: Maybe an [IntFlag enum](https://docs.python.org/3/library/enum.html#intflag) could be useful to set/read which relays are set.

pymeasure/instruments/velleman/velleman_k8090.py
```diff
@@ -0,0 +1,249 @@
...
+        The reply has a different command byte than the request.
+        
+        Three items (lists of 8 bools) are returned:
+        - Previous state: the state of each relay before this event
+        - Current state: the state of each relay now
+        - Timer state: the state of each relay timer
+        """,
+        cast=int,
+        get_process=_get_process_status,
+    )
```
bmoneke, line 102: Here you could take advantage of that intflag.

pymeasure/instruments/velleman/velleman_k8090.py
```diff
@@ -0,0 +1,249 @@
...
+
+    id = None  # No identification available
+
+    def __init__(
+        self, adapter, name="Velleman K8090", baud_rate=19200, timeout=1000, **kwargs
+    ):
+        super().__init__(
+            adapter, name=name, baud_rate=baud_rate, timeout=timeout, **kwargs
+        )
```
bmoneke, line 129: Please move the init to the beginning of the file and please make the baud_rate a "specific parameter" for serial connections (that facilitates other adapters):
```suggestion
    def __init__(
        self, adapter, name="Velleman K8090", baud_rate=19200, timeout=1000, **kwargs
    ):
        super().__init__(
            adapter, name=name, asrl={'baud_rate': baud_rate}, timeout=timeout, **kwargs
        )
```

Is the baud_rate settable? If not, you could replace it with the number 19200.

pymeasure/instruments/velleman/velleman_k8090.py
```diff
@@ -0,0 +1,249 @@
...
+
+        if len(response) < 7:
+            raise ConnectionError(
+                f"Incoming packet was {len(response)} bytes instead of 7"
+            )
+        elif len(response) > 7:
+            response = response[-7:]  # Keep the most recent block only
+
+        stx, command, mask, param1, param2, checksum, etx = list(response)
+
```
bmoneke, line 232: Just a suggestion:
```suggestion
    # Keep the most recent block only
    stx, command, mask, param1, param2, checksum, etx = list(response[-7:])

```

docs/api/instruments/velleman/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.velleman
+
+########
+Velleman
+########
+
+This section contains specific documentation on the Velleman  instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
```
bmoneke, line 7:
```suggestion
This section contains specific documentation on the Velleman instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
```

pymeasure/instruments/velleman/velleman_k8090.py
```diff
@@ -0,0 +1,249 @@
...
+    def _make_checksum(self, command, mask, param1, param2):
+        # The formula from the sheet requires twos-complement negation,
+        # this works
+        return 1 + 0xFF - ((self.BYTE_STX + command + mask + param1 + param2) & 0xFF)
+
+    def switch_on_blocking(self, channels):
+        """Switch on a set of channels and wait for the confirmation.
+
+        See :prop:`switch_on`.
```
bmoneke, line 139:
```suggestion
        See :attr:`switch_on`.
```
For "attribute", that should fix the warning. Also elsewhere in the document.

tests/instruments/velleman/test_velleman_k8090.py
```diff
@@ -0,0 +1,67 @@
+"""
```
bmoneke, line 1: You miss the copyright notice at the top of the file.

bmoneke (commented):
Thanks for contributing, welcome to the Contributors of pymeasure.
I added a few comments/suggestions, how you could improve the code.
Generally, it looks really good.

pymeasure/instruments/velleman/velleman_k8090.py
RobertoRoos: You mean in the API signature or internally? I.e. a user would type something like:

```
from pymeasure.instruments.velleman import VellemanK8090, Channels

instrument.switch_on(Channels.CH1 | Channels.CH2 | Channels.CH3)
instrument.switch_off(Channels.ALL)
```

It's a bit longer, but I do think it's more robust. I'll change it.

pymeasure/instruments/velleman/velleman_k8090.py
RobertoRoos: Ah, I didn't know about the `asrl` argument. Changed and I hardcoded the 19200, because the K8090 does not support a different rate anyway.

Does the instrument now insert the baudrate into the adapter that's passed, regardless of the adapter constructor parameters?

RobertoRoos:
Thank you for the feedback @bmoneke , this is a cool project work on.

Let me know if you'd like more changes or a rebase.

I should also point out not all options of the K8090 are covered yet. But I didn't really want to spend time on the features I don't need at this moment.

pymeasure/instruments/velleman/velleman_k8090.py
bmoneke: Exactly.

bmoneke:
> I should also point out not all options of the K8090 are covered yet. But I didn't really want to spend time on the features I don't need at this moment.

As long as an instrument is usable (most important features covered), it's fine. We all are volunteers with limited time.

pymeasure/instruments/velleman/velleman_k8090.py
```diff
@@ -0,0 +1,272 @@
...
+        The set command is identical, and the processing part of
+        :attr:`status` is called.
+
+        :returns: See :attr:`status`
+        """
+        mask = _parse_channels(channels)
+        self.write(f"0x11,{mask}")
+        data = self.read()
+        items = [int(it) for it in data.split(",")]
+        return _get_process_status(items)
```
bmoneke, line 177: In order to not repeat yourself, I'd access the already defined methods:
```suggestion
        self.switch_on = channels
        return self.status
```

pymeasure/instruments/velleman/velleman_k8090.py
```diff
@@ -0,0 +1,272 @@
...
+
+        A read will return a list of CMD, MASK, PARAM1 and PARAM2.
+        """
+        num_bytes = 7  # Each packet should always be exactly 7 bytes
+
+        if isinstance(self.adapter.connection, serial.Serial):
+            # Because we do not always read the reply to a setting, the buffer
+            # might not have been empty. Try to clear it completely but without waiting
+            # for more
+            num_bytes = max(7, self.adapter.connection.in_waiting)
```
bmoneke, line 247: And what do we do with a VISAAdapter (default for pymeasure)?
You could use `read_bytes(-1)` to read all availabe bytes.

I see a problem here:
1. you switch on/off
2. the device responds with a message (buffer contains 7 bytes)
3. you send another command
4. you read (executed immediately afterwards). As there are already 7 bytes in the buffer you read those (wrong ones): ERROR
5. Now the device response arrives with the correct answer, but nobody reads it.

For that reason, it is better to put a `read_bytes(-1)` as `check_errors` method and set the `check_set_errors` parameter to True in the switch_on/off commands. Similar to this code (taken from hcp.TC038D, the read method does some error checking as well):
~~~python
    def check_errors(self):
        """To be called from the property setters to read the acknowledgment."""
        self.read()

    setpoint = Instrument.control(
        "R,0x106", "W,0x106,%i",
        """The setpoint of the oven in °C.""",
        check_set_errors=True,
        get_process=lambda v: v / 10,
        set_process=lambda v: int(round(v * 10)),
    )
~~~

pymeasure/instruments/velleman/velleman_k8090.py
RobertoRoos: I should indeed use `self.switch_on` here, but we cannot directly use `self.status` as that also requests a status, causing a duplicate request-response couples.
The point of this method is to capture that automated response.

pymeasure/instruments/velleman/velleman_k8090.py
RobertoRoos: Oops, I didn't know the VISAadapter would work with a plain Serial interface. I'll try to test this later today. Kind of address should you submit to connect with e.g. `COM1`?

RobertoRoos: Hm, you are right, that is a valid scenario. I'll implement it as you suggest.
What I dislike about `read_bytes(-1)` is it waits for the full timeout, which seems unnecessary. But maybe I can play with recognizing the STX/ETX bytes.

pymeasure/instruments/velleman/velleman_k8090.py
bmoneke: ah, I see.

pymeasure/instruments/velleman/velleman_k8090.py
bmoneke: > Oops, I didn't know the VISAadapter would work with a plain Serial interface. I'll try to test this later today. Kind of address should you submit to connect with e.g. COM1?

For full description, see https://pyvisa.readthedocs.io/en/latest/introduction/names.html
How you address the device, depends on the backend you use: from NI or keysight or pyvisa-py. The defined way (which should work always) is "ASRL1". The NI backend also understands "COM1". The pyvisa-py backend, however, does not understand "COM1" and requires "ASRL1".

> What I dislike about read_bytes(-1) is it waits for the full timeout, which seems unnecessary. But maybe I can play with recognizing the STX/ETX bytes.

The SerialAdapter waits for the timeout. The VISAAdapter, however, recognizes STX/ETX (if I understand it correctly) out of the box.

RobertoRoos: Yeah. I'm making some progress with a new implementation.

I am done with a version that uses the ETX byte to prevent most unnecessary waits, but guess what: the K8090 will also send the status when the test button on-board are used... So we will always need to drain the buffer fully, we can't even rely on the terminators.
I am getting real tired of this device :')

bmoneke: Ok...

So we're back at handling it in "read".
What do you think: You read all available bytes and then you use the last 7 bytes only?

pymeasure/instruments/velleman/velleman_k8090.py
RobertoRoos: I dropped these blocking versions completely, because the K8090 replies make this a huge mess.

pymeasure/instruments/velleman/velleman_k8090.py
RobertoRoos: Pff, pushed a new update with a new idea:

* before we do a write we just clear the buffer, to make sure no old messages are present.
* Then we write 
* and in `set_check_errors` we read the reply (if any). Reading is assisted by the ETX byte to speed up the process in all normal use cases.

Right now we must read by the ETX byte because #863 .

RobertoRoos: @bmoneke I am resolving this conversation, I think the current implementation is good.

pymeasure/instruments/velleman/velleman_k8090.py
```diff
@@ -0,0 +1,266 @@
...
+
+        See :meth:`write`, replies from the machine use the same format.
+
+        A read will return a list of CMD, MASK, PARAM1 and PARAM2.
+        """
+        # Read all bytes from buffer - until the termchar or a timeout
+        response = self.read_bytes(-1, break_on_termchar=True)
+
+        if len(response) < 7:
+            raise ConnectionError(f"Incoming packet was {len(response)} bytes instead of 7")
```
bmoneke, line 233: Good idea with that `break_on_termchar`.

Only one question: could the other fields (CMD, MASK, PARAM1/2) have by chance the same value as ETX?

RobertoRoos: Omg you are right, they well can be. The ETX byte is 0X0F (0b00001111 in bits), which is a very reasonable mask for e.g. a switch on / off command.

I will indeed add a line to keep reading of the response is too short.

RobertoRoos: I changed my mind. I had the following written:

```
```

bmoneke:
Thanks for your implementation.

Can it be merged from your side?

If you want to verify, that read works as expected, you could add some tests, verifying that wrong messages lead to exceptions, see `with pytest.raises()`.

RobertoRoos:
> If you want to verify, that read works as expected, you could add some tests, verifying that wrong messages lead to exceptions, see `with pytest.raises()`.

Good suggestion, added to the protocol tests.
I'd be ready for a merge now!

PR #866, opened 2023-02-28
Changed behavior of `read_bytes(-1)` for `SerialAdapter`

### LINKED ISSUE #862: `TypeError` on SerialAdapter `read_bytes(-1)`
When I use `self.read_bytes(-1)` from my instrument using a `SerialAdapter`, I get the error:

> C:\Program Files\Python310\lib\ctypes\__init__.py:66: TypeError

With the traceback:

```
..\..\..\pymeasure\instruments\velleman\velleman_k8090.py:247: in read
    response = self.read_bytes(-1)
..\..\..\pymeasure\instruments\instrument.py:153: in read_bytes
    return self.adapter.read_bytes(count, **kwargs)
..\..\..\pymeasure\adapters\adapter.py:123: in read_bytes
    read = self._read_bytes(count, break_on_termchar, **kwargs)
..\..\..\pymeasure\adapters\serial.py:108: in _read_bytes
    return self.connection.read(1e99 if count == -1 else count, **kwargs)
..\..\..\venv\lib\site-packages\serial\serialwin32.py:278: in read
    buf = ctypes.create_string_buffer(n)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

init = 1e+99, size = None

    def create_string_buffer(init, size=None):
        """create_string_buffer(aBytes) -> character array
        create_string_buffer(anInteger) -> character array
        create_string_buffer(aBytes, anInteger) -> character array
        """
        if isinstance(init, bytes):
            if size is None:
                size = len(init)+1
            _sys.audit("ctypes.create_string_buffer", init, size)
            buftype = c_char * size
            buf = buftype()
            buf.value = init
            return buf
        elif isinstance(init, int):
            _sys.audit("ctypes.create_string_buffer", None, init)
            buftype = c_char * init
            buf = buftype()
            return buf
>       raise TypeError(init)
E       TypeError: 1e+99

C:\Program Files\Python310\lib\ctypes\__init__.py:66: TypeError

```

It looks like reading 1e99 bytes no (longer?) works.

RobertoRoos:
Fixes #862 

There are a couple of ways to handle the read-all operation. I don't think it matters too much, as long as the read timeout does not elapse more than once.

### DIFF
--- a/pymeasure/adapters/serial.py
+++ b/pymeasure/adapters/serial.py
@@ -109,12 +109,14 @@ class SerialAdapter(Adapter):
             # For -1 we empty the buffer completely
             return self._read_bytes_until_timeout()
 
-    def _read_bytes_until_timeout(self, **kwargs):
-        """Read from the serial until a timeout occurs, regardless of how many bytes."""
-        # `Serial.readlines()` sounds appropriate instead but it turns out it has an
-        # unpredictable timeout
+    def _read_bytes_until_timeout(self, chunk_size=256, **kwargs):
+        """Read from the serial until a timeout occurs, regardless of the number of bytes.
+
+        :chunk_size: The number of bytes attempted to in a single transaction.
+            Multiple of these transactions will occur.
+        """
+        # `Serial.readlines()` has an unpredictable timeout, see PR #866
         data = bytes()
-        chunk_size = 256
         while True:
             chunk = self.connection.read(chunk_size, **kwargs)
             data += chunk
--- a/pymeasure/adapters/serial.py
+++ b/pymeasure/adapters/serial.py
@@ -94,7 +94,7 @@ class SerialAdapter(Adapter):
         """Read a certain number of bytes from the instrument.
 
         :param int count: Number of bytes to read. A value of -1 indicates to
-            read from the whole read buffer.
+            read from the whole read buffer (waits for timeout).
         :param bool break_on_termchar: Stop reading at a termination character.
         :param \\**kwargs: Keyword arguments for the connection itself.
         :returns bytes: Bytes response of the instrument (including termination).
@@ -103,9 +103,12 @@ class SerialAdapter(Adapter):
             return self.connection.read_until(self.read_termination.encode(),
                                               count if count > 0 else None,
                                               **kwargs)
+        elif count >= 0:
+            return self.connection.read(count, **kwargs)
         else:
-            # At -1 we read a very large number of bytes, which can be considered the whole buffer.
-            return self.connection.read(1e99 if count == -1 else count, **kwargs)
+            # For -1 we empty the buffer completely
+            num_bytes = self.connection.in_waiting + 100  # Read what's already queued + some extra
+            return self.connection.read(num_bytes, **kwargs)
 
     def flush_read_buffer(self):
         """Flush and discard the input buffer."""
--- a/pymeasure/adapters/serial.py
+++ b/pymeasure/adapters/serial.py
@@ -107,8 +107,19 @@ class SerialAdapter(Adapter):
             return self.connection.read(count, **kwargs)
         else:
             # For -1 we empty the buffer completely
-            num_bytes = self.connection.in_waiting + 100  # Read what's already queued + some extra
-            return self.connection.read(num_bytes, **kwargs)
+            return self._read_bytes_until_timeout()
+
+    def _read_bytes_until_timeout(self, **kwargs):
+        """Read from the serial until a timeout occurs, regardless of how many bytes."""
+        # `Serial.readlines()` sounds appropriate instead but it turns out it has an
+        # unpredictable timeout
+        data = bytes()
+        chunk_size = 256
+        while True:
+            chunk = self.connection.read(chunk_size, **kwargs)
+            data += chunk
+            if len(chunk) < chunk_size:  # If fewer bytes got returned, we had a timeout
+                return data
 
     def flush_read_buffer(self):
         """Flush and discard the input buffer."""
--- a/tests/adapters/test_serial_with_loopback.py
+++ b/tests/adapters/test_serial_with_loopback.py
@@ -1,0 +1,80 @@
+#
+# Test the serial adapter using looped COM ports.
+# This looping needs to be done outside this script. One suggestion is
+# using the application `com0com`.
+#
+
+import pytest
+
+from serial import Serial
+from time import time
+from pymeasure.adapters import SerialAdapter
+
+
+ADAPTER_TIMEOUT = 3.0  # Seconds
+
+
+@pytest.fixture()
+def adapter(connected_device_address):
+    """Call with `--device-address="COM3,COM4"` to select two devices."""
+    device = connected_device_address.split(",")[0]
+    return SerialAdapter(
+        device,
+        baudrate=19200,
+        timeout=ADAPTER_TIMEOUT,
+        read_termination=chr(0x0F),
+    )
+
+
+@pytest.fixture()
+def loopback(connected_device_address):
+    """See `adapter()`."""
+    device = connected_device_address.split(",")[1]
+    return Serial(device, baudrate=19200)
+
+
+def test_read(adapter, loopback):
+    """Regular read with fixed number of bytes."""
+    loopback.write(b"abc")
+    result = adapter.read_bytes(3)
+
+    assert len(result) == 3
+
+
+def test_read_all_short(adapter, loopback):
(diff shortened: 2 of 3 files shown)


bmoneke:
Could you please add a test?

RobertoRoos:
Added a test, though I think this was already covered in existing tests.

Actual congruent reading/writing we cannot really test like this I think.

bmoneke:
I did not know by heart, whether we have a test.
As the reading failed for you, I figured, that we cannot have a test.

Does the test pass for you?

I just ran it and for me it passed, which contradicts your experience in #862 .

bmoneke:
A small check:

Using the loopback interface (in the test) environment worked reading that large number of bytes (1e99), but for real communication it does not, so a different implementation in loopback and adapter communication made our tests pass.
Great for tests...

Nevertheless. Thanks, that you found that bug.

pymeasure/adapters/serial.py
```diff
@@ -103,9 +103,12 @@ def _read_bytes(self, count, break_on_termchar, **kwargs):
...
                                               count if count > 0 else None,
                                               **kwargs)
+        elif count >= 0:
+            return self.connection.read(count, **kwargs)
         else:
-            # At -1 we read a very large number of bytes, which can be considered the whole buffer.
-            return self.connection.read(1e99 if count == -1 else count, **kwargs)
+            # For -1 we empty the buffer completely
+            num_bytes = self.connection.in_waiting + 100  # Read what's already queued + some extra
+            return self.connection.read(num_bytes, **kwargs)
```
bmoneke, line 111: Let us return to that original version, which really reads until a timeout happens:
```suggestion
            return b"\n".join(self.connection.readlines(**kwargs))
```

Your proposed solution fails, if a device sends 101 (or more) bytes as we read only the first 100 bytes.

RobertoRoos:
> Does the test pass for you?
> 
> I just ran it and for me it passed, which contradicts your experience in https://github.com/pymeasure/pymeasure/issues/862 .

> A small check:
> 
> Using the loopback interface (in the test) environment worked reading that large number of bytes (1e99), but for real communication it does not, so a different implementation in loopback and adapter communication made our tests pass.
> Great for tests...
> 
> Nevertheless. Thanks, that you found that bug.

Ah, I see. I was testing with another test file that I did not commit, where I used a COM emulator (com0com) to create a real loopback device. There I would get the TypeError.

RobertoRoos:
I was testing with `return b"\n".join(self.connection.readlines(**kwargs))` and it seems the newline character is not removed from the result, so just `"".join(...)` would work.

But also double the timeout occurs before the function returns (which is really an issue inside `pyserial` it seems).

pymeasure/adapters/serial.py
RobertoRoos: I was testing with return b"\n".join(self.connection.readlines(**kwargs)) and it seems the newline character is not removed from the result, so just "".join(...) would work.

But also double the timeout occurs before the function returns (which is really an issue inside pyserial it seems)

RobertoRoos: Pushed a new version, reading by chucks does seem to solve it everything, it gives us the most control. I also included my looped test, let me know if it should be left out instead.

@bmoneke do you concur?

tests/adapters/test_serial_with_loopback.py
```diff
@@ -0,0 +1,80 @@
+#
+# Test the serial adapter using looped COM ports.
+# This looping needs to be done outside this script. One suggestion is
+# using the application `com0com`.
```
bmoneke, line 4: Thanks for that test.

Would you mind to expand the setup description a little bit, such that is more easy to replicate it?

tests/adapters/test_serial_with_loopback.py
```diff
@@ -0,0 +1,80 @@
...
+from serial import Serial
+from time import time
+from pymeasure.adapters import SerialAdapter
+
+
+ADAPTER_TIMEOUT = 3.0  # Seconds
+
+
+@pytest.fixture()
+def adapter(connected_device_address):
```
bmoneke, line 18: Great that you use this device fixture!

pymeasure/adapters/serial.py
bmoneke: The first suggestion (not sent), was reading chunks, but I thought the newline thing shorter.

In the end, your solution is nice and understandable (not that "read lines" magic).

pymeasure/adapters/serial.py
```diff
@@ -103,9 +103,23 @@ def _read_bytes(self, count, break_on_termchar, **kwargs):
...
+            return self.connection.read(count, **kwargs)
         else:
-            # At -1 we read a very large number of bytes, which can be considered the whole buffer.
-            return self.connection.read(1e99 if count == -1 else count, **kwargs)
+            # For -1 we empty the buffer completely
+            return self._read_bytes_until_timeout()
+
+    def _read_bytes_until_timeout(self, **kwargs):
+        """Read from the serial until a timeout occurs, regardless of how many bytes."""
+        # `Serial.readlines()` sounds appropriate instead but it turns out it has an
```
bmoneke, line 114: That comment is in the PR discussion (we could add it to the issue as well), but does not need to be in the source code.

bmoneke:
Only a few small documentation issues, then we can merge.

Thanks for developing this with us.

pymeasure/adapters/serial.py
bilderbuchi: It's useful context, though ("comment why, not what") -- reading without knowing about the issue, the first thing one could be wondering why pymeasure is reimplementing `Serial.readlines`.

pymeasure/adapters/serial.py
```diff
@@ -103,9 +103,23 @@ def _read_bytes(self, count, break_on_termchar, **kwargs):
...
+        # unpredictable timeout
+        data = bytes()
+        chunk_size = 256
```
bilderbuchi, line 117: IMO, this should be a kwarg  of the method -- folks might want to tweak this for performance optimisation/tradeoffs.

tests/adapters/test_serial_with_loopback.py
```diff
@@ -0,0 +1,80 @@
...
+    loopback.write(b"aaaaabbbbbccccceeeee" * 50)
+    start = time()
+    result = adapter.read_bytes(-1)
+    elapsed = time() - start
+
+    assert len(result) == 20 * 50
+    assert ADAPTER_TIMEOUT == pytest.approx(elapsed, abs=0.1)
+
+
+def test_read_all_exact_chunck_size(adapter, loopback):
```
bilderbuchi, line 65:
```suggestion
def test_read_all_exact_chunk_size(adapter, loopback):
```

tests/adapters/test_serial_with_loopback.py
```diff
@@ -0,0 +1,80 @@
...
+    loopback.write(b"a")
+    start = time()
+    result = adapter.read_bytes(-1)
+    elapsed = time() - start
+
+    assert len(result) == 1
+    assert ADAPTER_TIMEOUT == pytest.approx(elapsed, abs=0.1)
+
+
+def test_read_all(adapter, loopback):
```
bilderbuchi, line 55: Might be useful to have the test name or test docstring reflect the fact that we are confirming that reading everything only takes _one_ timeout.

tests/adapters/test_serial_with_loopback.py
RobertoRoos: Sure, you're right. Turned it into a class so the docs description covers it clearly.

pymeasure/adapters/serial.py
RobertoRoos: I agree with @bilderbuchi , the 'why' is leading here. Reduced comment to one line and linked this PR.

pymeasure/adapters/serial.py
RobertoRoos: Sure, no reason not to expose it.

tests/adapters/test_serial_with_loopback.py
RobertoRoos: You're right. I combined the tests with a parametrize and tried to clarify the point of the tests.

RobertoRoos:
Thanks for your input @bilderbuchi !

bmoneke:
@bilderbuchi, if the comments are resolved to your satisfaction, you may merge.

PR #867, opened 2023-02-28
VISAAdapter fix read_bytes(-1)

### LINKED ISSUE #863: `VISAAdapter` still terminating on default term character in `read_bytes(-1)`
Pretty odd and specific issue, not sure if this belong here or on PyVisa.

When I try to read the complete buffer in a serial connection using the `VISAAdapter`, it still breaks on the byte corresponding to `\n`:

```
    def __init__(self, adapter, name="Velleman K8090", timeout=1000, **kwargs):
        super().__init__(
            adapter,
            name=name,
            asrl={"baud_rate": 19200},
            write_termination="",
            read_termination=chr(0x0F),
            timeout=timeout,
            **kwargs,
        )
      
    # ...

    def read(self):
        response = self.read_bytes(-1)

        # `response` will end with "\n", even though there are more bytes in the buffer!
```

Encountered in #859 .

It seems the issue is two fold: in this code any termchar should be ignored and it's even responding to the wrong termchar.

This is with the `pyvisa-py` backend.

bmoneke:
Fixes #863 

Goal: test expected `read_bytes(-1)` behaviour and make it work.

- First commit is a test (which should be added anyways).
- Second commit is a fail proof (though not very elegant) solution to read the whole buffer.

Drawback of the solution: It takes at least one timeout period.

### DIFF
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -171,7 +171,7 @@ class VISAAdapter(Adapter):
         """Read a certain number of bytes from the instrument.
 
         :param int count: Number of bytes to read. A value of -1 indicates to
-            read from the whole read buffer.
+            read from the whole read buffer until timeout.
         :param bool break_on_termchar: Stop reading at a termination character.
         :param \\**kwargs: Keyword arguments for the connection itself.
         :returns bytes: Bytes response of the instrument (including termination).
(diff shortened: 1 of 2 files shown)


pymeasure/adapters/visa.py
```diff
@@ -181,13 +181,14 @@ def _read_bytes(self, count, break_on_termchar=False, **kwargs):
...
-            self.connection.read_termination = None
-            # Try except allows to set the read_termination even after an error.
-            try:
-                return self.connection.read_raw(**kwargs)
-            finally:
-                self.connection.read_termination = read_termination
+            result = bytearray()
+            while True:
+                try:
+                    result.extend(self.connection.read_bytes(1))
```
bilderbuchi, line 187: Might be good to leave a comment (and reference to the upstream issue) about _why_ we are doing this.

bmoneke: Done.

bilderbuchi (approved):
I don't have strong intuition if this is the correct thing to do (e.g. if we should follow upstream behaviour), but LGTM. 
AFAICT, we would undo this approach when/if the upstream https://github.com/pyvisa/pyvisa/issues/728 is fixed, correct?

bmoneke:
> I don't have strong intuition if this is the correct thing to do (e.g. if we should follow upstream behaviour), but LGTM. AFAICT, we would undo this approach when/if the upstream [pyvisa/pyvisa#728](https://github.com/pyvisa/pyvisa/issues/728) is fixed, correct?

Yes, we can always implement it differently.

My rationale was, to deliver, what the docstring promises. As pyvisa (in its current Form) does not support it, I had to change the implementation.
If pyvisa should offer a better solution, we can implement that.

So, we merge?

PR #868, opened 2023-03-03
Added instrument Teledyne HDO6xxx

### LINKED ISSUE #857: Add support for the Teledyne LeCroy HDO6xxx oscilloscope
I would like support for the [LeCroy HDO6054B oscilloscope](https://teledynelecroy.com/oscilloscope/oscilloscopemodel.aspx?modelid=11553). I already have an old private package that controls it, so I hope to be able to port this myself.

Note that that the product is called "Teledyne Lecroy", and I see `pymeasure` has both instrument folders for Teledyne and Lecroy, so I'm not sure where to put it. Any suggestions?

And I hope I can reuse some of the logic from https://pymeasure.readthedocs.io/en/latest/api/instruments/lecroy/lecroyT3DSO1204.html.

I expect that the HDO6xxx family can be controlled the same way, but I am not sure and I cannot test them all. How should I name and describe the device in `pymeasure`?

Connection is made using NI Visa.

RobertoRoos:
Fixes #857

Adding support the Teledyne oscilloscope. I am not done yet, but I would like to hear if this approach is acceptable: I am currently extending the `LeCroyT3DSO1204` class, instead of for example creating a mutual base class.

According to Teledyne, the leading manual for Visa interfacing is: https://cdn.teledynelecroy.com/files/manuals/maui-remote-control-automation_27jul22.pdf

There are some discrepancies between the `LeCroyT3DSO1204` device and my `TeledyneHDO6xxx`, which I can fix by overriding some method/properties. But I was also checking the manual for the T3DSO (https://cdn.teledynelecroy.com/files/manuals/tds031000-2000_programming_manual.pdf) and noticed that some differences in the current code are not supported by the manual.
For example, according to both PDFs the query `BANDWIDTH_LIMIT?` should for both devices return the BWL for all channels, e.g. "CH1,OFF,CH2,ON,CH3,ON,CH4,OFF". But the tests for `LeCroyT3DSO1204` expect a reply per channel.

Unfortunately I don't have a T3DSO to test. I suggest to minimize changes to the existing class, but I thought it'd be good to mention this.


RobertoRoos:
I have worked on an alternative.
The HDO6 class needs some properties to be removed, which is a bit of a pain. So I suggest to go for the more conventional approach and relying more on inheritance.

I haven't pushed new commits into this PR yet, instead I created a new one and created a PR on my own fork just to display the suggested changes:
https://github.com/DEMCON/pymeasure/pull/2/files

The idea is to build on four classes (for now):
UML

I'm fairly sure this [MAUI-group](https://teledynelecroy.com/maui/) of devices are very similar, hence I would add it as an intermediate class.

All logic would be moved upwards as much as possible.

Any opinions?

bmoneke:
Just a quick note: For us maintainers, protocol tests are more useful than tests with an actual device as we can determine whether a code change breaks an implementation without having the device at all.

bilderbuchi:
The approach of layering classes sounds reasonable. However, what seems weird to me is that a "Channel" inherits from a full "oscilloscope"

bilderbuchi:
> But I was also checking the manual for the T3DSO (https://cdn.teledynelecroy.com/files/manuals/tds031000-2000_programming_manual.pdf) and noticed that some differences in the current code are not supported by the manual.
> For example, according to both PDFs the query `BANDWIDTH_LIMIT?` should for both devices return the BWL for all channels, e.g. "CH1,OFF,CH2,ON,CH3,ON,CH4,OFF". But the tests for `LeCroyT3DSO1204` expect a reply per channel.

Maybe the original contributor @LastStarDust can help out clearing this up?

pymeasure/instruments/teledyne/teledyneHDO6xxx.py
```diff
@@ -0,0 +1,256 @@
...
+    if value.endswith(" V"):
+        value = value[:-2]
+
+    return float(value)
+
+
+class ScopeChannel(LeCroyT3DSO1204ScopeChannel):
+    """Extended Channel object for the HDO6xxx."""
+
+    # For some reason "20MHZ" is registered as "ON"
```
bilderbuchi, line 56: 200 or 20?

LastStarDust:
> > But I was also checking the manual for the T3DSO (https://cdn.teledynelecroy.com/files/manuals/tds031000-2000_programming_manual.pdf) and noticed that some differences in the current code are not supported by the manual.
> > For example, according to both PDFs the query `BANDWIDTH_LIMIT?` should for both devices return the BWL for all channels, e.g. "CH1,OFF,CH2,ON,CH3,ON,CH4,OFF". But the tests for `LeCroyT3DSO1204` expect a reply per channel.
> 
> Maybe the original contributor @LastStarDust can help out clearing this up?

I will check it as soon as I am in front of the hardware. I would not be too surprised if the hardware behavior is not consistent with the manual. It has happened before.

pymeasure/instruments/teledyne/teledyneHDO6xxx.py
```diff
@@ -0,0 +1,256 @@
...
+    _BANDWIDTH_LIMITS = ["OFF", "ON", "200MHZ"]
+
+    _TRIGGER_SLOPES = {"negative": "NEG", "positive": "POS"}
+
+    @property
+    def trigger_level2(self):
```
bilderbuchi, line 62: you should be able to delete properties after class definition -- `del ScopeChannel.trigger_level2`.
Although, ideally such properties are pushed down to the class that adds them (i.e. the commonalities are pulled up into a common superclass), as removing class behaviour is considered "bad practice".

RobertoRoos:
> The approach of layering classes sounds reasonable (and is great for code reuse!). However, what seems weird to me is that a "Channel" inherits from a full "oscilloscope"

Whoops, that was just a typo. The channel will actually have their own inheritance, mirroring the device classes.
The duplicate structure is not ideal, but there seems no way around it.

> you should be able to delete properties after class definition -- del ScopeChannel.trigger_level2.
Although, ideally such properties are pushed down to the class that adds them (i.e. the commonalities are pulled up into a common superclass), as removing class behaviour is considered "bad practice".

Of course, this was just a remnant of the old concept with limited inheritance. I have it set-up as you describe now.

RobertoRoos:
Looks like you're on board with the hierarchical structure, so I pulled in my new branch.

pymeasure/instruments/teledyne/teledyneHDO6xxx.py
RobertoRoos: The modes are "OFF", "ON" and "200MHZ", with "ON" meaning 20 MHz.

bmoneke:
I had a quick glance at the code and that is promising.

Do I understand it correctly, that you moved the code from the TS... oscilloscope to the general teledyne_oscilloscope?

LastStarDust:
> Okay, I think I'm about done. Some pretty big refactors already.
> 
> The new HDO6 is not nearly as far implemented as the T3DSO is, but this is as much as I need. (And I don't have experience with oscilloscopes to begin with).
> At least there should be no functions left in the HDO6 class that cause errors.
> 
> @LastStarDust do you think you could run the device tests of this branch on the T3DSO again? 

Yes, I can run the tests but it might take about a week before I can get hold of the hardware.

bilderbuchi:
The test failure comes from the code example in the `TeledyneT3AFG` constructor docstring I think. Try to make it look like this:
``` 
    - Add frequency ranges per model
    - Add channel coupling control

    .. code-block: python
        # Example assumes Ethernet (TCPIP) interface
        generator=TeledyneT3AFG('TCPIP0::xxx.xxx.xxx.xxx::pppp::SOCKET')
        generator.reset()
        generator.ch_1.wavetype='SINE'
        generator.ch_1.amplitude=2
        generator.ch_1.output_enabled=True
        
    """
```

bilderbuchi:
@LastStarDust did you manage to run the tests on the hardware that you intended? Was the possible discrepancy with the manual mentioned [above](https://github.com/pymeasure/pymeasure/pull/868#issuecomment-1455069910) clarified?

ashokbruno:
Sorry I am late to the conversation. I worked at Teledyne LeCroy for a long long time, and I am pretty familiar with the remote operation of the scope. I am onboard with the architecture proposed by @RobertoRoos. 
Here's my comments/responses:
The subsidary that makes scope is Teledyne LeCroy (a small part of Teledyne), Any T3xxx products are under the umbrella Teledyne Test Services.  They are not MAUI based and while it may support SCPI type commands, it may not support the VBS style commands which are the automation framework for LeCroy scopes. Also the automation of these devices are more robust using this framework compared to the SCPI type commands. 
I would seperate the implementation of T3 class based products from the MAUI based products. While I was working at LeCroy I had started a repo to create a high level class most people would be able to use with their scopes. It is by no means complete but it has a fair amount of functionality and code. See  https://github.com/TeledyneLeCroy/lecroydso and https://lecroydso.readthedocs.org/

I need a LeCroy DSO for the Labmaster series of Oscilloscopes which is also a MAUI based high end oscilloscope, and one of my colleagues was looking into integrating it with pymeasure. I will look at your implementation in closer detail to see how we can contribute to this.

Also if you want to test this on different scope models, Teledyne LeCroy offers a product called MAUI Studio(https://www.teledynelecroy.com/support/softwaredownload/mauistudio.aspx) which allows the scope to run on your PC, and simulate various things including remote operation.

bilderbuchi:
@ashokbruno thanks for the additional insight, that's interesting and useful!

RobertoRoos:
I've started the somewhat painful procedure of rebasing my branch back up to `origin/master`.

Aside from that I am not sure what we need to proceed. Do we wait for more feedback from @ashokbruno ? Could we already merge what we have now?

bmoneke:
Normally, it is better for the review process to merge the current master branch into your branch than to rebase and force-push. Force-push has the negative downside, that conversations (gratefully all are already resolved) cannot be linked to code anymore. Additionally you cannot look at the changes from the last review.

RobertoRoos:
> Normally, it is better for the review process to merge the current master branch into your branch than to rebase and force-push. Force-push has the negative downside, that conversations (gratefully all are already resolved) cannot be linked to code anymore. Additionally you cannot look at the changes from the last review.

Ah, I see. I have a local backup of the un-rebased branch. Let me know if you would still like me to force-push it back.

Aside from that, there seems to be a small issue with the CI. Pretty much all the existing commands are failing the doc test because their text does not start with the right words. I guess the original is not failing because the code hasn't been in a diff yet?
Do I actually need to change all the phrasing?

bmoneke:
> Ah, I see. I have a local backup of the un-rebased branch. Let me know if you would still like me to force-push it back.

No, don't redo it again.

> Aside from that, there seems to be a small issue with the CI. Pretty much all the existing commands are failing the doc test because their text does not start with the right words. I guess the original is not failing because the code hasn't been in a diff yet?

The test did not fail before, because the tests were added to master after you branched off. Now you included the tests (via rebasing) into your branch and now they fail.

> Do I actually need to change all the phrasing?

We added these tests, to make sure, that new instruments follow our guidelines. I see, that it is quite some work, but it's straight forward: The first word has to be Get (or Read), Set, or Control for properties. Furthermore, it does not involve changing code which could create more issues.

bmoneke:
> I will check it as soon as I am in front of the hardware. I would not be too surprised if the hardware behavior is not consistent with the manual. It has happened before.

@LastStarDust did you have a chance to check?

bmoneke:
> Aside from that I am not sure what we need to proceed. Do we wait for more feedback from @ashokbruno ? Could we already merge what we have now?

Let us proceed.

The PR is quite large (+ 2k lines, -1k lines). Do you think, it is possible, to split the PR into two?
1. Split T3DSO into T3DSO and LeCroy Oscilloscope (which does not include anything new, right?)
2. Add MAUI and HDO

That split will make the review effort easier as it is smaller. Especially the first step (the separation into two files) creates a lot of changed lines, but is easy to review. If both parts are mangled together, the review is, as if we had 2k new lines and 1k removed original lines.

RobertoRoos:
> We added these tests, to make sure, that new instruments follow our guidelines. I see, that it is quite some work, but it's straight forward: The first word has to be Get (or Read), Set, or Control for properties. Furthermore, it does not involve changing code which could create more issues.

Alright, I'll give it a go.

> The PR is quite large (+ 2k lines, -1k lines). Do you think, it is possible, to split the PR into two?
> 
>     1. Split T3DSO into T3DSO and LeCroy Oscilloscope (which does not include anything new, right?)
> 
>     2. Add MAUI and HDO

Sure, that shouldn't be a problem. Do we care about the commit history up to this point? If not, I'll start a new PR with only the copied changes.

bmoneke:
> Sure, that shouldn't be a problem. Do we care about the commit history up to this point? If not, I'll start a new PR with only the copied changes.

However you like, but as the change is simply (just devide the code into two), a simple, new PR should be enough.

RobertoRoos:
#951 has been made!

LastStarDust:
> > I will check it as soon as I am in front of the hardware. I would not be too surprised if the hardware behavior is not consistent with the manual. It has happened before.
> 
> @LastStarDust did you have a chance to check?

Sorry for the late reply. Unfortunately, I do not have easy access to the hardware anymore, since it is a company oscilloscope and I was moved to a different team/project.

Anyway if I remember correctly, the question was about a small part of the code not being consistent with the User Manual.
I remember that I found some errors and typos in the Manual, so yes the code is obviously not consistent in those cases.

However, I strived for 100% unit test coverage with actual hardware, so I am pretty confident that the code is working with the latest version of the hardware (at the time of the test).

That said, if you find a reproducible bug, I might be able to justify to my employer the additional effort. and let me debug it again.

bmoneke:
Thanks for your offer.

RobertoRoos:
Alright, merged in `origin/master` and fixed the tests. (Thanks @bmoneke for merging the other!)

I think this is ready, what do we think?

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -0,0 +1,211 @@
...
+    # For some reason "20MHZ" is registered as "ON"
+    BANDWIDTH_LIMITS = ["OFF", "ON", "200MHZ"]
+
+    TRIGGER_SLOPES = {"negative": "NEG", "positive": "POS"}
+
+    # Reset listed values for existing commands:
+    bwlimit_values = BANDWIDTH_LIMITS
+
+    def autoscale(self):
+        """Perform auto-setup command for channel."""
```
bmoneke, line 42: name is "autoscale", but docstring (and command) state "auto-setup", would not "autosetup" be a better name?

pymeasure/instruments/teledyne/teledyneHDO6xxx.py
```diff
@@ -0,0 +1,45 @@
...
+    All functionality is inherited from the base class.
+    """
+
+    ch_1 = Instrument.ChannelCreator(TeledyneHDO6xxxChannel, 1)
+
+    ch_2 = Instrument.ChannelCreator(TeledyneHDO6xxxChannel, 2)
+
+    ch_3 = Instrument.ChannelCreator(TeledyneHDO6xxxChannel, 3)
+
+    ch_4 = Instrument.ChannelCreator(TeledyneHDO6xxxChannel, 4)
```
bmoneke, line 45: Isn't that superfluous or will a later implementation differ from the MAUI version?

I see two options: state somewhere a list of oscilloscopes, which can be used by the MAUI class. The list could start with this one.

Another option is the following:
```suggestion
class TeledyneHDO6xxx(TeledyneMAUI):
    """Reference to the Teledyne-Lecroy HDo6xxx class of oscilloscopes.

    All functionality is inherited from the base class.
    """

    pass
```

pymeasure/instruments/teledyne/teledyneT3AFG.py
```diff
@@ -126,11 +126,12 @@ class TeledyneT3AFG(Instrument):
     """Represents the Teledyne T3AFG series of arbitrary waveform
     generator interface for interacting with the instrument.
 
-    Intially targeting T3AFG80, some features may not be available on
+    Initially targeting T3AFG80, some features may not be available on
     lower end models and features from higher end models are not
-    included here intially.
+    included here initially.
```
bmoneke, line 131:
```suggestion
    included here yet.
```

tests/instruments/teledyne/test_teledyneHDO6xxx_with_device.py
```diff
@@ -0,0 +1,316 @@
...
+from pymeasure.instruments.teledyne import TeledyneHDO6xxx
+
+
+class TestTeledyneHDO6xxx:
+    """
+    Unit tests for TeledyneHDO6xxx class.
+
+    This test suite, needs the following setup to work properly:
+        - A TeledyneHDO6xxx device should be connected to the computer;
+        - The device's address must be set in the RESOURCE constant;
```
bmoneke, line 38: That's not true anymore, is it?

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -0,0 +1,211 @@
...
+
+    # Change listed values for existing commands:
+    bwlimit_values = TeledyneMAUIChannel.BANDWIDTH_LIMITS
+
+    ###############
+    #   Trigger   #
+    ###############
+
+    @property
+    def trigger(self):
```
bmoneke, line 125: Would "trigger_setup" or "trigger_config" be better names?

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -0,0 +1,211 @@
...
+            "trigger_slope": self.trigger_slope
+        }
+        return ch_setup
+
+
+class TeledyneMAUI(TeledyneOscilloscope):
+    """A base class for the MAUI-type of Teledyne oscilloscopes.
+
+    This class is not exactly device specific. Nonetheless, it might already work out of the box and
+    therefore this class not abstract.
```
bmoneke, line 100: I would state it differently:
```suggestion
    This base class works out of the box. Some properties, especially the number of channels,
    might have to be adjusted to the actual device.
```

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -0,0 +1,211 @@
...
+        """Perform auto-setup command for channel."""
+        self.write("AUTO_SETUP FIND")
+
+    # noinspection PyIncorrectDocstring
+    def setup(self, **kwargs):
+        """Setup channel. Unspecified settings are not modified.
+
+        Modifying values such as probe attenuation will modify offset, range, etc. Refer to
+        oscilloscope documentation and make multiple consecutive calls to setup() if needed.
+        See property descriptions for more information.
```
bmoneke, line 51: Do you see it useful to include these values (e.g. probe attenuation) here?

Maybe this method should execute the properties in a certain order (e.g. first probe attenuation, then something else...).

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -0,0 +1,211 @@
...
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.teledyne.teledyne_oscilloscope import TeledyneOscilloscope, \
+    TeledyneOscilloscopeChannel, _results_list_to_dict
+
+
+class TeledyneMAUIChannel(TeledyneOscilloscopeChannel):
+    """Base class for channels on a :class:`TeledyneMAUI` device."""
+
+    # For some reason "20MHZ" is registered as "ON"
```
bmoneke, line 33: I guess, that in earlier times, they had only a 20MHz limit and thus they switched the limit on or off. Later they added 200MHz and used the new message for that value.

```suggestion
    # Probably for historic reasons, "20MHZ" is registered as "ON"
```

bmoneke (changes requested):
In general it looks good. I left a few comments, suggestions, and questions.

pymeasure/instruments/teledyne/teledyneMAUI.py
RobertoRoos: Yes, it would be. But now it is inherited from the base class, from the existing T3DSO1204.

I think that for backwards compatibility it is best kept as-is.

pymeasure/instruments/teledyne/teledyneHDO6xxx.py
RobertoRoos: Yeah, you're right. With this approach (being consistent) we would have to add a new empty class for each supported device. Best to keep a list then.

pymeasure/instruments/teledyne/teledyneMAUI.py
RobertoRoos: Probably, but backwards compatibility again.

pymeasure/instruments/teledyne/teledyneMAUI.py
RobertoRoos: Not really, I don't want to expand this further. Personally, I think this function should not exist as users should just use the attributes directly (and rely on their documentation).

But, again, compatibility.

bmoneke: What did you change? Only the docstring? In this case, you could call also `super().setup(**kwargs)`.
If you did not change anything, the method is imported from the parent such that you do not need to copy it.

bilderbuchi:
@bmoneke is this ready to merge?

tests/instruments/teledyne/test_teledyneMAUI_with_device.py
```diff
@@ -26,17 +26,15 @@
...
-class TestTeledyneHDO6xxx:
+class TestTeledyneMAUI:
     """
-    Unit tests for TeledyneHDO6xxx class.
+    Unit tests for TeledyneMAUI class.
 
-    This test suite, needs the following setup to work properly:
-        - A TeledyneHDO6xxx device should be connected to the computer;
-        - The device's address must be set in the RESOURCE constant;
-        - A probe on Channel 1 must be connected to the Demo output of the oscilloscope.
```
bmoneke: Isn't that line (a probe connected) still necessary or do you not use that demo output?

RobertoRoos: No, it seems not necessary as I never connected that probe and all tests succeeded.

bmoneke: Ok

bmoneke:
Thanks for your contribution.

PR #870, opened 2023-03-03
Restructured the Lakeshore instrument drivers to use the new channels feature, and added a new Lakeshore instrument.

samcondon4:
This PR updates the structure of the `lakeshore` temperature monitor/controller instrument drivers to use the `Channels` feature and adds a new driver class: `LakeShore224`. Two channel classes are created in the `lakeshore_base.py` module:

   1. `LakeShoreTemperatureInputChannel`
   2. `LakeShoreHeaterOutputChannel`

The new `LakeShore224` uses the channel creator for 12 independent temperature input channels, and the `LakeShore331` driver is updated to use two temperature input and heater output channels. The `wait_for_temperature()` method previously implemented in the `LakeShore331` driver is moved into the `LakeShoreTemperatureInputChannel` and modified slightly to
wait for the temperature to reach a target temperature that is given as an argument to the function. Previously, the target temperature was obtained by querying the temperature setpoint for the desired channel, but this doesn't allow temperatures to be waited on when using temperature monitoring instruments that have no setpoint property, like the `LakeShore224`.


bilderbuchi:
you can run linting locally: `flake8 .` if you want to avoid Github feeding this to you bit-by-bit :-P

pymeasure/instruments/lakeshore/__init__.py
```diff
@@ -22,6 +22,8 @@
 # THE SOFTWARE.
 #
 
+from .lakeshore_base import LakeShoreTemperatureInputChannel, LakeShoreHeaterOutputChannel
```
bilderbuchi, line 25: Maybe leave out that import, because user won't be able to make any use of these classes directly, so it doesn't make much sense for them to be in the main namespace of the manufacturer.

pymeasure/instruments/lakeshore/lakeshore224.py
```diff
@@ -0,0 +1,52 @@
...
+
+    .. code-block:: python
+
+        monitor = LakeShore224('GPIB::1')
+
+        print(monitor.input_A.kelvin)           # Print the temperature in kelvin on sensor A
+        monitor.input_A.wait_for_temperature()  # Wait for the temperature on sensor A to stabilize.
+    """
+
+    tchannels = ['0', 'A', 'B', 'C1', 'C2', 'C3', 'C4', 'C5', 'D1', 'D2', 'D3', 'D4', 'D5']
```
bilderbuchi, line 46: are you using the `tchannel` attribute elsewhere? If no, maybe pass it to `ChannelCreator` directly, to avoid having a public class attribute lying around that is not useful.

bilderbuchi:
Great that you restructure and promote code reuse!

pymeasure/instruments/lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,117 @@
...
+        'RANGE {ch},%i',
+        """String property controlling heater range, which can take the
+       values: off, low, medium, and high.""",
+        validator=strict_discrete_set,
+        values={'off': 0, 'low': 1, 'medium': 2, 'high': 3},
+        map_values=True)
+    setpoint = Instrument.control(
+        'SETP? {ch}', 'SETP {ch},%f',
+        """A floating point property that control the setpoint temperature
+        in the preferred units of the control loop sensor."""
```
bilderbuchi, line 116: Is there an interface to adjust the preferred unit?

pymeasure/instruments/lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+
+        def percent_difference(temperature):
+            return np.abs(100 * (temperature - target) / target)
+
+        t = time()
+        target_acquired = False
+        while not target_acquired:
+            reading = getattr(self, unit)
+            reading = np.array([reading]) if self.id != '0' else np.array(reading)
```
bilderbuchi, line 77: Why this `if`? Is this channel of the Lakeshore224 formatting replies differently? If so, might be good to call that out in a comment here. (Also in other docstrings, maybe?)

pymeasure/instruments/lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,117 @@
...
+            diff = percent_difference(reading)
+            target_acquired = True if np.all(diff < accuracy) else False
```
bilderbuchi, line 79: There is `np.allclose()` that does the same I think.

pymeasure/instruments/lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,117 @@
...
+
+from pymeasure.instruments import Instrument, Channel
+from pymeasure.instruments.validators import strict_discrete_set
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class LakeShoreTemperatureInputChannel(Channel):
+    """ Temperature input channel on a lakeshore temperature monitor. Reads the temperature in
```
bilderbuchi, line 37: Maybe better `LakeShoreTemperatureMeasurementChannel` or `LakeShoreTemperature(Sensing/Sensor)Channel`? For a moment I was confused that an "input" channel only has measurements (that produce outputs), not controls. :-P

pymeasure/instruments/lakeshore/__init__.py
samcondon4: Good point, this is removed in the next commit.

pymeasure/instruments/lakeshore/lakeshore_base.py
samcondon4: The if is here because for some models, you can use '0' as the channel argument and it returns a list with the values for all channels at once. So instead of having to do queries on each individual channel you can just do a single query on channel '0'.

Since I convert responses to numpy arrays before the `np.all` call, I added this `if` so that when channel 0 was queried you would still end up with `reading` as a 1-dimensional array. Though I am realizing that for the evaluation of `np.all` or `np.allclose` it doesn't really matter, so this if is unnecessary

pymeasure/instruments/lakeshore/lakeshore_base.py
samcondon4: Yeah I think maybe just taking out the `Input` and `Output` from the temperature and heater channel classes is probably a good way to go. So the class names would be:

- `LakeShoreTemperatureChannel`
- `LakeShoreHeaterChannel`

pymeasure/instruments/lakeshore/lakeshore_base.py
samcondon4: As far as I can tell, one can only set the units manually on the device itself and not through a computer interface. I haven't found any command to set the preferred units yet...

bilderbuchi: OK, thanks, just wanted to make sure.

pymeasure/instruments/lakeshore/lakeshore_base.py
bilderbuchi: > so that when channel 0 was queried you would still end up with reading as a 1-dimensional array.

A worthy design goal IMO, getting `sometype` or `list-of-sometype` back depending on circumstances feels annoying to me.

bilderbuchi:
I think we're done here, thanks!

PR #876, opened 2023-03-13
Add (deprecated) global preprocess reply

### LINKED ISSUE #872: The `preprocess_reply` argument in `Instrument.__init__()` is no longer used in a `Instrument.values()` call
Previous to the PR https://github.com/pymeasure/pymeasure/pull/718 rework , `Instrument.values()`  would return `Adapter.values()`.  After the rework `Instrument.values()` is implemented in `CommonBase.values()` and no longer calls `Adapter.values()`.

Previously as part of an Instrument's initialization the argument `preprocess_reply` argument would be passed to the `Adapter` which in turn would be used as `self.preprocess_reply`  in the `Adapter.values()` function:
https://github.com/pymeasure/pymeasure/blob/8bf3c295fa2f977145ea42d633da8f16a7cc4151/pymeasure/adapters/adapter.py#L184

Now `CommonBase` doesn't ingest the `preprocess_reply` argument and so `CommonBase.values()` doesn't create or assign `self.preprocess_reply`:
https://github.com/pymeasure/pymeasure/blob/8bf3c295fa2f977145ea42d633da8f16a7cc4151/pymeasure/instruments/common_base.py#L312

That means when an Instrument driver passes along a `preprocess_reply` argument as part of Instrument initialization, that argument has no effect when `Instrument.values()` is called.

Two instrument drivers in the code base use that functionality:

https://github.com/pymeasure/pymeasure/blob/8bf3c295fa2f977145ea42d633da8f16a7cc4151/pymeasure/instruments/fluke/fluke7341.py#L67

https://github.com/pymeasure/pymeasure/blob/8bf3c295fa2f977145ea42d633da8f16a7cc4151/pymeasure/instruments/signalrecovery/dsp7265.py#L72

bmoneke:
As this PR fixes a regression bug, the deprecation version is the original one (0.11), not the next version.

The `Channel` class does not allow to use the global `preprocess_reply`, as Channels were added after the deprecation.

Closes #872

### DIFF
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -130,8 +130,8 @@ class CommonBase:
         self._special_names = self._setup_special_names()
         self._create_channels()
         if preprocess_reply is not None:
-            warn(("Parameter `preprocess_reply` is deprecated in CommonBase. "
-                 "Implement it in the instrument instead."),
+            warn(("Parameter `preprocess_reply` is deprecated. "
+                 "Implement it in the instrument, e.g. in `read`, instead."),
                  FutureWarning)
         self.preprocess_reply = preprocess_reply
         super().__init__(**kwargs)
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -97,6 +97,13 @@ class CommonBase:
 
     This class contains everything needed for pymeasure's property creator
     :meth:`control` and its derivatives :meth:`measurement` and :meth:`setting`.
+
+    :param preprocess_reply: An optional callable used to preprocess
+        strings received from the instrument. The callable returns the
+        processed string.
+
+        .. deprecated:: 0.11
+            Implement it in the instrument's `read` method instead.
     """
 
     # Variable holding the list of DynamicProperty parameters that are configurable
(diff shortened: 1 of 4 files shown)


pymeasure/instruments/common_base.py
```diff
@@ -119,9 +126,14 @@ class CommonBase:
...
     __reserved_prefix = "___"
 
-    def __init__(self, **kwargs):
+    def __init__(self, preprocess_reply=None, **kwargs):
         self._special_names = self._setup_special_names()
         self._create_channels()
+        if preprocess_reply is not None:
+            warn(("Parameter `preprocess_reply` is deprecated in CommonBase. "
+                 "Implement it in the instrument instead."),
+                 FutureWarning)
```
bilderbuchi, line 135: I have to admit I'm confused by the message. IIUC, if one uses a `preprocess_reply` argument in an Instrument initialisation, this gets passed here, so one gets this message, although `preprocess_reply` was used in `Instrument` already, not in `CommonBase` (which an instrument implementor would not touch anyway).
What did you intend "Implement it in the instrument instead." to mean?

bilderbuchi (changes requested):
One thing is unclear, otherwise this looks good.

pymeasure/instruments/common_base.py
bmoneke: > Should we mention CommonBase here?

I added it, to distinguish it from the similar message in the Adapter. With the additional comment regarding `read`, it will be distinguishable.

> What did you intend "Implement it in the instrument instead." to mean? Implement the functionality in the instrument's read(), right?

`read` is a hot candidate. You could use `values` as well (depending on the circumstances). I offer `read` as a suggestion, below.

What about this
```suggestion
            warn(("Parameter `preprocess_reply` is deprecated. "
                 "Implement it in the instrument, e.g. in `read`, instead."),
                 FutureWarning)
```
?

Issue #667, opened 2022-08-06
check_errors usage inconsistent


bilderbuchi:
Our `Instrument.check_errors()` [implementation](https://github.com/pymeasure/pymeasure/blob/ebf6df97e229c0e916192a495ade65053b234ac3/pymeasure/instruments/instrument.py#L514) returns a list of received errors (as its docstring says).

Interestingly though, where we use them, e.g. in control's fget, https://github.com/pymeasure/pymeasure/blob/ebf6df97e229c0e916192a495ade65053b234ac3/pymeasure/instruments/instrument.py#L361, we simply discard those return values without doing anything!

I think where we use it we should do something with the errors, most probably log them.

msmttchr:
Actually default `check_errors` provided in `PyMeasure` logs errors. If you agree, this issue can be closed.

msmttchr:
I am closing this as per explanation above, @bilderbuchi feel free to reopen if you disagree

bilderbuchi:
No, thanks, it's fine!

bilderbuchi:
I think we should reopen this, even if only to adjust the docstring of `Instrument.check_errors` -- as it is now, the usage does not reflect the docstring, return values are not used (some instrument docstring examples, print them, though!).

We should probably 
* review usage in the codebase
* consider changing the method to return nothing (as we don't use the return values in the core, right now)
* adjust/correct the docstring to advise the correct behaviour (logging errors). In overridden methods we don't have influence on what they will do, and implementors will need guidance on that.

bmoneke:
In #883 the question arose, how to handle errors.

`check_get/set_errors` could
- return a single (maybe aggregated from several errors) error
- return a list of errors
- raise an error.

My proposal:
`check_set_errors` shall raise an error (which is **not** caught by the property, therefore shall be caught in the application) for critical errors. This is at the discretion of the instrument implementor.
For less critical errors it shall return a single error or list of errors (TBD), and the property shall create a corresponding log entry.

bilderbuchi:
Regarding logging: 
I think we should log as close to the source as possible, so in the default `check_errors`. 
This does not have to be the only way to "handle" an error, so I don't think "log and don't return anything" is a good approach. For example, a user of `check_errors` might want to do a retry or something when an error occurs (e.g. in an instrument method) -- we should enable such a workflow without having to intercept/parse logging streams to find out _if_ an error was encountered.

[...]

PR #883, opened 2023-03-30
Split check errors

### LINKED ISSUE #879: Allow property setter to read
Some instruments respond to every command, they even acknowledge setting a value.
Right now, the solution is to set `check_set_errors=True` and to read in `check_errors`, which seems more a hack than a solution.

Some solutions:

1. Previously (in some discussion) I proposed to add a further parameter to `control`, which switches between `write` and `ask` in the setter.
Advantage is, that you can choose it for every property.
Disadvantages are, that we have another parameter and that you have to add it to every single property.
2. `control`'s setter calls `set_property` (name to be determined). This method will just call `write` in the default implementation. An instrument might override this method in order to read afterwards and interpret the read response (which might be an acknowledgment or an error).
Advantages: it is declared only once; no additional parameter for `control`; you can raise an appropriate error "setting property xyz failed" (some instrument stores the last written command in order to be able to raise such an error!).
Disadvantage: You cannot do it for a subset of properties. This should be no problem, as mostly an instrument responds to all setters or to no one.

I like the second solution and would like to implement it. The first solution and current hack are given as alternative solutions.

What are your ideas?

bmoneke:
The property creators call `check_get_errors()`, `check_set_errors()`, which default to `check_errors()`.
Closes #879 

Tests expanded accordingly.
Instruments make usage of that new feature.
In this circumstance: Attocube moved to the new Channels infrastructure (but backwards compatible).

Question: How is the default relationship between these three error checking functions and instrument and Channel?
- Should the Channel functions relate to the instrument's one or should all relate to `check_errors` and that in turn to the instrument's `check_errors` (right now implemented).


bmoneke:
Thinking about the relationship of the Channels to the parent, and considering the experience with the modified instruments, I consider it better if all three check methods of a Channel call their respective counterpart in the instrument.

(right now only check_errors calls it counterpart)

bmoneke:
Now, all three check_errors methods of a channel call the parent's respective method.

It's great to have instruments tests, which ensure the same communication over code changes.

CHANGES.rst
```diff
@@ -5,6 +5,7 @@ New adapter and instrument mechanics
 - Channel class added. Instrument.channels and Instrument.ch_X (X is any channel name) are reserved for channel implementations.
 - All instruments are required to accept a :code:`name` argument.
 - Changed: :code:`read_bytes` of all Adapters by default does not stop reading on a termination character, unless the new argument :code:`break_on_termchar` is set to `True`.
+- The parameters :code:`check_get_errors` and :code:`check_set_errors` enable calling methods of the same name.
```
bilderbuchi, line 8:
```suggestion
- The parameters :code:`check_get_errors` and :code:`check_set_errors` enable calling methods of the same name. This enables more systematically dealing with instruments that acknowledge every "set" command.
```

docs/dev/adding_instruments/properties.rst
```diff
@@ -373,7 +373,10 @@ The same can be also achieved by the `preprocess_reply` keyword argument to :fun
 Checking the instrument for errors
 **********************************
 If you need to separately ask your instrument about its error state after getting/setting, use the parameters :code:`check_get_errors` and :code:`check_set_errors` of :meth:`~pymeasure.instruments.common_base.CommonBase.control`, respectively.
-If those are enabled, the method :meth:`~pymeasure.instruments.Instrument.check_errors` will be called after device communication has concluded.
+If those are enabled, the methods :meth:`~pymeasure.instruments.Instrument.check_get_errors` and :meth:`~pymeasure.instruments.Instrument.check_set_errors`, respectively, will be called be called after device communication has concluded.
+In the default implementation, both methods call :meth:`~pymeasure.instruments.Instrument.check_errors`.
+You can also use :meth:`~pymeasure.instruments.Instrument.check_set_errors` to read an automatic response, if the instrument responds to every set command with an acknowledgment or error.
```
bilderbuchi, line 378:
```suggestion
In the default implementation, for simplicity both methods call :meth:`~pymeasure.instruments.Instrument.check_errors`.
To read the automatic response of instruments that respond to every set command with an acknowledgment or error, override :meth:`~pymeasure.instruments.Instrument.check_set_errors` as needed.
```

pymeasure/instruments/attocube/anc300.py
```diff
@@ -139,34 +139,23 @@ class Axis:
...
-    def ask(self, command, **kwargs):
-        return self.controller.ask(self._add_axis_id(command), **kwargs)
-
-    def write(self, command, **kwargs):
-        return self.controller.write(self._add_axis_id(command), **kwargs)
-
-    def values(self, command, **kwargs):
-        return self.controller.values(self._add_axis_id(command), **kwargs)
+    def _add_axis_id(self, command):
+        self.insert_id(command)
```
bilderbuchi, line 153: This was a private method, so not part of the public API. I think it's fair game to delete `_add_axis_id` without process.

pymeasure/instruments/ipgphotonics/yar.py
```diff
@@ -98,11 +98,9 @@ def read(self):
         else:
             return reply[-1].strip()
 
-    def check_errors(self):
-        """Empty the read buffer."""
-        got = self.read().split(":")
-        if got[0] == "ERR":
-            raise ConnectionError(f"Connection error {got[-1]}.")
```
bilderbuchi, line 105: Where do you deal with the error case now?

pymeasure/instruments/channel.py
```diff
@@ -109,6 +109,20 @@ def check_errors(self):
         """
         return self.parent.check_errors()
```
bilderbuchi, line 110: I think I agree with your reasoning w.r.t. methods calling the parent's counterparts. However, what do we use `Channel.check_errors` for, now? The Channel's get/set variants call the parent's versions directly, so don't need that, also not in the default case.

bilderbuchi (commented):
A first pass, with some questions.

pymeasure/instruments/ipgphotonics/yar.py
bmoneke: `read` does already the error checking.
These lines were a superfluous repetition.

pymeasure/instruments/channel.py
bmoneke: > However, what do we use Channel.check_errors for, now?

Mainly for backwards compatibility, in case someone calls `check_errors` in a method.

Shall we deprecate it? Do we have to deprecate it (Channels are not yet in a release)?

There could be a hypothetical instrument, where you want to ask channel errors with `check_errors`, but that instrument may implement that method anyways.

pymeasure/instruments/attocube/anc300.py
bmoneke: Ok

pymeasure/instruments/attocube/anc300.py
```diff
@@ -139,34 +139,23 @@ class Axis:
...
```
bmoneke, line 154: Remove according to suggestion.
```suggestion
```

pymeasure/instruments/channel.py
bmoneke: I checked the codebase. The following instruments use `check_errors` in a Channel, intended to call the parent's one:

- Keithley2600 (not yet `Channel` class)
- AgilentB1500 (not yet `Channel` class)
- Agilent4156 (not yet `Channel` class)
- KeithleyBuffer (not yet `Channel` class)

In all these occurrences, a method calls `check_errors` in ordert o **ask** for errors. They could use the `check_get/set_errors` methods instead, but it does not seem right to me, as they are intended for getters/setters.

Therefore, I propose to keep all three methods in the Channel.

pymeasure/instruments/ipgphotonics/yar.py
bmoneke: Btw., tests ensure that the error is still raised.

bilderbuchi: Thanks!

pymeasure/instruments/channel.py
bilderbuchi: > Therefore, I propose to keep all three methods in the Channel.

Your reasoning sounds good, I agree.

> Do we have to deprecate it (Channels are not yet in a release)?

Generally, if something hasn't landed in a published release, yet, there is no need to follow a deprecation process, just change it as needed. That is partly the idea with doing a release _before_ merging a big feature -- there is more time to iron out any problems using the master branch.

pymeasure/instruments/common_base.py
```diff
@@ -645,23 +645,16 @@ def setting(set_command, docs,
                                   dynamic=dynamic,
                                   )
 
-    def check_errors(self):
-        """ Read all errors from the instrument.
-
-        :return: list of error entries
-        """
-        raise NotImplementedError("Implement it in a subclass.")
```
bilderbuchi: I agree with keeping check_get/set_errors abstract implementations for the interface only. 
I'm not sure I agree with removing `check_errors` completely -- it's the common parent/"interface definition" of both Instrument.check_errors and Channel.check_errors (as much as Python can do such a thing).

bmoneke: I added it (and a test) to the common_base.

It's good, that it is there, for another reason as well: Once we remove SCPI's `check_errors` from `Instrument`, the inherited version defines the interface, which instrument impelementations should follow.

pymeasure/instruments/instrument.py
```diff
@@ -219,3 +219,17 @@ def check_errors(self):
             return errors
         else:
             raise NotImplementedError("Non SCPI instruments require implementation in subclasses")
+
+    def check_get_errors(self):
+        """Check for errors after having gotten a property.
+
+        Called if :code:`check_get_errors=True` is set for that property.
+        """
+        self.check_errors()
```
bilderbuchi, line 228: Your remark in #667 made me realise that I didn't check the return value part of the interface. I think `check_get_errors` and `check_set_errors` should return what they get from `check_errors` (list of error entries as per the docstring). Otherwise that return just gets dropped on the way to the `check_get|set_errors` caller. One might argue that the only known callers are the properties, which don't make use of that return anyway, but for symmetry/least-surprise reasons it seems neater to me to return the values (and also codify that in the interface/abstract definitions).

bmoneke:
Regarding the return value of `check_get/set_errors` (here for better visibility later on):

It depends on the additional error handling:
1. If the `check_get/set_errors` method should also handle the errors (e.g. log them), they should not return anything.
In this case, they need to know the called command (would be good), which requires a different signature.
2. If `check_get/set_errors` returns the errors, the property may handle the errors (with information only available to the property itself).

I think it is best, if we have a standardized interface between the checking method and the properties, such that the properties may log the errors arising (see #667), which is option 2.

The correct interface definition should be discussed in #667.

pymeasure/instruments/instrument.py
bmoneke: I answered in the general discussion, lest the discussion disappears in a resolved conversation.

bmoneke:
I implemented the logging proposed in #667 and added tests to ensure proper logging.
@bilderbuchi please have a look at the logging message.

mcdo0486:
Tagging this thread with https://github.com/pymeasure/pymeasure/pull/906 again as a reminder to update driver if this PR is merged.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -290,9 +272,9 @@ def __init__(
         # switch console echo off
         self.ask('echo off')
 
-    def check_errors(self):
+    def check_set_errors(self, context=""):
```
bilderbuchi, line 275: This method should return a list of errors, not nothing. Same I think for others below.

bilderbuchi: See the lower part of https://github.com/pymeasure/pymeasure/issues/667#issuecomment-1546647620, too -- e.g. this should probably do a `read()` without context, catch an occurring ValueError, log that (including the `context` known in here, and turn this into an error list to return. Or re-raise after logging, as you prefer.

bmoneke: Done

pymeasure/instruments/common_base.py
```diff
@@ -476,7 +476,13 @@ def fget(self,
...
                                **values_kwargs)
             if check_get_errors:
-                errors = [str(error) for error in self.check_get_errors()]
+                try:
+                    error_list = self.check_get_errors()
+                except Exception as exc:
+                    log.error("Exception raised while getting a property with the command "
+                              f"""'{command_process(get_command)}': '{str(exc)}'.""")
+                    raise
+                errors = [str(error) for error in error_list]
```
bilderbuchi, line 485: Is that conversion necessary? Can't we rely on the expectation that this is already a list-of-strings? Or is that a just-to-be-safe thing?

pymeasure/instruments/common_base.py
```diff
@@ -476,7 +476,13 @@ def fget(self,
...
                 if errors:
                     log.error("Error recieved after trying to get a property with the command "
```
bilderbuchi, line 487: "rec**ei**ved" (also elswhere I think). Sorry, it doesn't let me do a suggestion you can merge straight away.

pymeasure/instruments/attocube/anc300.py
```diff
@@ -273,8 +278,19 @@ def __init__(
...
-        """Read after setting a value."""
-        self.read()
+        """Check for errors after having set a property and log them.
+
+        Called if :code:`check_set_errors=True` is set for that property.
+
+        :return: List of error entries.
+        """
+        try:
+            self.read()
```
bilderbuchi, line 285: So all those methods (in the instruments changed in this commit) never return an error message from `self.read()`? That's what I thought would go into the error list to return.

bmoneke: In these instruments, read already does some error checking (for example checksums or the returned value) as that is needed for getting properties.
In case of some problem, read raises an exception, but does not return an error (which would be indistinguishable from a valid response).

Yes, check_errors could convert the exceptions into a returned error, but then you break symmetry: failed reading of a property is an exception, while failed setting is just a log entry.
Therefore, I decided to continue the exception path (as these instruments were previously) and not the logging path.

pymeasure/instruments/common_base.py
bmoneke: Thanks

pymeasure/instruments/common_base.py
bmoneke: A single error of a Scpi instrument is a list of error number and error text. The "join" command in the next line requires a list of strings. Therefore, the conversion is necessary.

pymeasure/instruments/common_base.py
bmoneke: Done

pymeasure/instruments/common_base.py
bilderbuchi: I see, thanks.

pymeasure/instruments/instrument.py
bilderbuchi: I think this is resolved now?

bmoneke: Yes

bmoneke:
Everything is resolved now, do you want to look over it again, or do we merge, @bilderbuchi ?

bilderbuchi:
I think we're done, good to merge. :-)

PR #888, opened 2023-04-06
Implementation of HP8560A & HP8561B

neuschs:
First non-finished draft for the HP8560A & HP8561B implementation.

*  added files to __init__ file of hp directory
* added implementation and tests for those commands:
- Identification
- Attenuation
- Amplitude Units
- Auto Couple
- Exchange Traces
- Subtract display line from trace b
- Continuous Sweep
- Blank Trace
- Freq Center
- Clear Write Trace
- Test Coupling
- Demodulation Mode
- Demodulation AGC
- Demodulation Time
- Detector Mode Specifically for HP8561B:
- External Mixer
- Conversion Loss

I hope for early feedback before I coded all of the stuff


pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,304 @@
...
+        """
+        Get identification of the device with software and hardware revision (e.g. HP8560A,002,H03)
+        """,
+        maxsplit=0
+    )
+
+    attenuation = Instrument.control(
+        "AT?", "AT %s",
+        """
+        Controls input attenuation in decade steps from 10 to 70db or set to AUTO and MAN(ual)
```
LongnoseRob, line 408: Docsreings should be written in the impterative form, in this case "control" 
Please use a space between the value and the unit desination (dB)

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,304 @@
...
+        frequency / span mode. The span remains constant; the start and stop frequencies change as
+        the center frequency changes.
+        """,
+        validator=joined_validators(strict_range, strict_discrete_set),
+        values=[[0, 1], ["UP", "DN"]],
+        dynamic=True,
+        set_process=lambda v: v if isinstance(v, str) else (
+                '%.11E HZ' % Decimal(v)).replace('+', ''
+                                                 )
+    )
```
LongnoseRob, line 475: 1. Would'nt it sound better to call this property `center_frequency`
2. Spelling of the docstring (`Set` instead of `sets`, **H**ertz)
3. While Ilike the idea of allwoingthe limits also by "UP" & "DN", the whole `set_process` could be dropped if only numerical values are accepted. then the set-string could be written as `"CF %11.0f HZ"`

LongnoseRob (commented):
Hello and thank you for your contribution.

pymeasure/instruments/hp/hp856Xx.py
neuschs: Thanks for the feedback !
Regarding 1) I named the parameter the same as the Rohde & Schwarz FSL implementation, although I agree with your argument.
About 3) I think you are right, I will remove the relative modification parameters.
Would make more sense from an scientific view.

tests/instruments/hp/test_hp856Xx.py
```diff
@@ -0,0 +1,477 @@
...
+        ) as instr:
+            # test set and get of attenuation as integer
+            instr.attenuation = 70
+            assert instr.attenuation == 70
+            # test string parameters
+            instr.attenuation = "AUTO"
+            assert instr.attenuation == 20
+            # test truncation
+            instr.attenuation = 16
+            assert instr.attenuation == 20
```
bmoneke, line 59: It is better to use pytests `parametrize` function if you test different values, that way you notice, if a single value fails or all values.

tests/instruments/hp/test_hp856Xx.py
bmoneke, line 1: Thanks for adding tests!

tests/instruments/hp/test_hp856Xx.py
```diff
@@ -0,0 +1,477 @@
...
+            with pytest.raises(TypeError):
+                instr.blank_trace(0)
+
+    @pytest.mark.parametrize("function,command", [
+        ("start_frequency", "FA"),
+        ("center_frequency", "CF"),
+        ("stop_frequency", "FB"),
+        ("frequency_offset", "FOFFSET")
+    ])
+    @pytest.mark.parametrize("hp_derivat,max_freq", [(HP8560A, 2.9e9), (HP8561B, 6.5e9)])
```
bmoneke, line 119: According to normal python rules, I'd add a space:
```suggestion
    @pytest.mark.parametrize("hp_derivat, max_freq", [(HP8560A, 2.9e9), (HP8561B, 6.5e9)])
```

tests/instruments/hp/test_hp856Xx.py
```diff
@@ -0,0 +1,477 @@
...
+        with expected_protocol(
+                HP856Xx,
+                [("AUNITS " + amplitude_unit, None),
+                 ("AUNITS?", amplitude_unit)],
+        ) as instr:
+            instr.amplitude_unit = amplitude_unit
+            assert instr.amplitude_unit == amplitude_unit
+
+    @pytest.mark.parametrize(
+        "function,command",
```
bmoneke, line 72:
```suggestion
        "function, command",
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,1166 @@
...
+from pymeasure.instruments.validators import strict_discrete_set, truncated_discrete_set, \
+    joined_validators, strict_range
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class StrEnum(str, Enum):
+    def __str__(self):
+        return self.value
```
bmoneke, line 38:
```suggestion
class StrEnum(str, Enum):
    """Enum with str represantation. Native for python > 3.10."""

    def __str__(self):
        return self.value
```

Adding a note, such that it might be replaced, once native `StrEnum` is available in all supported python versions.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,1166 @@
...
+
+    def __repr__(self):
+        return "ErrorCode(\"" + self.short + " - " + self.long + "\")"
+
+    def __eq__(self, other):
+        return self.code == other.code
+
+
+class HP856Xx(Instrument):
+    """
```
bmoneke, line 392: Please provide a short description of the device, typically something like "Represents the HP856Xx series oscilloscopes.

bmoneke (commented):
Thanks for your contribution. It's always good if people contribute new instruments, ideas, and bugfixes.

I left a few comments, but did not review the whole code in detail. It looks good, though, at a first glance.

Please check, whether you could profit from already existing instruments. Maybe there are already drivers for HP similar spectrum analyzers in pymeasure, which could be modified/used for these ones.

neuschs:
Thanks for your feedback, I will refactor the code to your suggestions.

neuschs:
@bmoneke In the last check two warnings arose the temptronic_base uses the sphinx class reference with auomatic resolution :class:`.ErrorCode`. I specify also a class ErrorCode, I would propose to change the docstring in temptronic_base to 
`:class:`temptronic_base.ErrorCode``
So everyone can use his ErrorCode implementation. 

Alternative would be to rename my class to an unique name. 
What do you guys think ?

bmoneke:
> @bmoneke In the last check two warnings arose the temptronic_base uses the sphinx class reference with auomatic resolution :class:'.ErrorCode'. I specify also a class ErrorCode, I would propose to change the docstring in temptronic_base to `:class:'temptronic_base.ErrorCode'` So everyone can use his ErrorCode implementation.
> 
> Alternative would be to rename my class to an unique name. What do you guys think ?

I do not consider unique names as an appropriate solution, so the linking should be tied to the filename.
What do you think, @bilderbuchi?

bilderbuchi:
> > @bmoneke In the last check two warnings arose the temptronic_base uses the sphinx class reference with auomatic resolution :class:'.ErrorCode'. I specify also a class ErrorCode, I would propose to change the docstring in temptronic_base to `:class:'temptronic_base.ErrorCode'` So everyone can use his ErrorCode implementation.
> > Alternative would be to rename my class to an unique name. What do you guys think ?
> 
> I do not consider unique names as an appropriate solution, so the linking should be tied to the filename. What do you think, @bilderbuchi?

I agree, a more specific designation of the docstring should be the proper solution.

neuschs:
Is it okay if I do the change in this PR ?

bilderbuchi:
Yeah, sure. 😃

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,1644 @@
...
+from pymeasure.instruments.validators import strict_discrete_set, truncated_discrete_set, \
+    joined_validators, strict_range
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class StrEnum(str, Enum):
+    """
+    Until StrEnum is broadly available / pymeasure relies on python => 3.10.x
```
bmoneke, line 38:
```suggestion
    Until StrEnum is broadly available / pymeasure relies on python <= 3.10.x
```

neuschs:
I did it a little bit different. If you write `:class:'.ErrorCode'` sphinx tries to auto resolve across multiple files. By writing `:class:'ErrorCode'` without the dot, sphinx only resolves inside the file.

neuschs:
@bmoneke Do you think this PR is in reviewable condition ?

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -498,6 +609,102 @@ def __init__(self, adapter, name="Hewlett-Packard HP856Xx", **kwargs):
...
+        """
+        Control the automatic IF adjustment. This function is normally
+        on. Because the IF is continuously adjusting, executing the IF alignment routine is seldom
+        necessary. When the IF adjustment is not active, an "A" appears on the left side of the
+        display.
+
+        - FULL IF adjustment is done for all IF settings.
+        - CURR IF adjustment is done only for the IF settings currently displayed.
+        - OFF turns the continuous IF adjustment off.
+        - ON reactivates the continuous IF adjustment.
```
bmoneke, line 639:
```suggestion
        - `"FULL"` IF adjustment is done for all IF settings.
        - `"CURR"` IF adjustment is done only for the IF settings currently displayed.
        - `False` turns the continuous IF adjustment off.
        - `True` reactivates the continuous IF adjustment.
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -498,6 +609,102 @@ def __init__(self, adapter, name="Hewlett-Packard HP856Xx", **kwargs):
...
+            Kg of overrange. When adding or subtracting trace data, any results exceeding
+            this limit are clipped at the limit.
+        """,
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: "1", False: "0"},
+        cast=str
+    )
+
+    annotation = Instrument.control(
```
bmoneke, line 695: For boolean properties, it is good to form the property name as a question in order to facilitate to know, what happens:
```suggestion
    annotation_enabled = Instrument.control(
```

now `annotation_enabled = True` is self-evident.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -513,14 +720,15 @@ def __init__(self, adapter, name="Hewlett-Packard HP856Xx", **kwargs):
 
         """,
         validator=joined_validators(strict_discrete_set, truncated_discrete_set),
-        values=[["AUTO", "MAN"], [10, 20, 30, 40, 50, 60, 70]]
+        values=[["AUTO", "MAN"], arange(10, 80, 10)],
+        cast=int
```
bmoneke, line 724: It's good practice to finish the last line with a comma as well. That way, adding more lines (or removing a line) does not change any other line than the added/removed ones.
```suggestion
        cast=int,
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1009,29 +1217,42 @@ def continuous_sweep(self):
...
-        both amplitude uncertainty (where the signal level appears to vary with small changes in
-        frequency) and frequency resolution (due to filter shape factor and sidelobes). Windows
-        are weighting functions that are applied to the input data to force the ends of that
-        data smoothly to zero, thus reducing the step discontinuity and reducing measurement
-        errors.
+        """The FFT command performs a discrete Fourier transform on the source
+        trace array and stores the logarithms of the magnitudes of the results
+        in the destination array. The maximum length of any of the traces is
+        601 points. FFT is designed to be used in transforming zero-span
+        amplitude-modulation information into the frequency domain. Performing
```
bmoneke, line 1224: A docstring should start with a one-line summary (in imperative mode) and than more context:
```suggestion
        """Calculate and show a discrete Fourier transform.

        The FFT command performs a discrete Fourier transform on the source
        trace array and stores the logarithms of the magnitudes of the results
        in the destination array. The maximum length of any of the traces is
        601 points. FFT is designed to be used in transforming zero-span
        amplitude-modulation information into the frequency domain. Performing
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
+            if instr.marker_time == 2:
+                pass
+
+        """
+    )
+
+    marker_signal_tracking = Instrument.control(
+        "MKTRACK?", "MKTRACK %s",
+        """
+        Locate the active marker and sets the center frequency to the
```
bmoneke, line 1759: Please start the docstring with `Control` to indicate the type of property (the API does not show, whether it is a control or measurement or setting).

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
-        an HP 8560A Option 002. Takes frequency band letter as string.
+        self.write("MKSP")
 
-        .. list-table:: Title
-            :widths: 25 25 25 25
-            :header-rows: 1
+    def marker_to_center_frequency_step_size(self):
+        """Set the center frequency step-size equal to the frequency value of
+        the active marker."""
+        self.write("MKSS")
```
bmoneke, line 1735: This is an action (no parameter and no return value), please reformulate the name with a verb, for example `set_marker_to...`.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
+    normalize_trace_data = Instrument.control(
+        "NORMLIZE?", "NORMLIZE %s",
+        """
+        Activate or deactivate the normalization routine for
+        stimulus-response measurements. This function subtracts trace B from trace A, offsets the
+        result by the value of the normalized reference position
+        (:attr:`normalized_reference_level`), and displays the result in trace A.
+        'normalize_trace_data' is intended for use with the TODO STOREOPEN and
+        TODO STORESHORT or
+        TODO STORETHRU commands. These functions are used to store a reference trace into trace B.
```
bmoneke, line 1825: Here are still TODOs

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
+
+    def power_bandwidth(self, trace, percent):
+        """Measure the combined power of all signal responses contained in a
+        trace array. The command then computes the bandwidth equal to a
+        percentage of the total power. For example, if 100% is specified, the
+        power bandwidth equals the current frequency span. If 50% is specified,
+        trace elements are eliminated from either end of the array, until the
+        combined power of the remaining trace elements equals half of the total
+        power computed. The frequency span of these remaining trace elements is
+        the power bandwidth output to the controller.
```
bmoneke, line 1980: It is a method, so please formulate it as an action:
`get_power_bandwidth`.

You could implement the traces with the [Channels class](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html), if you want to.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
+
+    def recall_thru(self):
+        """Recalls the internally stored thru-reference trace into trace B.
+
+        The instrument state is also set to the stored thru-reference
+        state.
+        """
+        self.write("RCLTHRU")
+
+    revision = Instrument.measurement(
```
bmoneke, line 2162: Maybe?
```suggestion
    firmware_revision = Instrument.measurement(
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
+    )
+
+    status = Instrument.measurement(
+        "STB?",
+        """
+        Returns to the controller the decimal equivalent of the bits set in the
+        status byte (see the RQS and SRQ commands). STB is equivalent to a serial poll command.
+        The RQS and associated bits are cleared in the same way that a serial poll command would
+        clear them.
+        """,
```
bmoneke, line 2417: Please start with "Measure" or "Get"
```suggestion
        Get the decimal equivalent of the bits set in the
        status byte (see the RQS and SRQ commands). STB is equivalent to a serial poll command.
        The RQS and associated bits are cleared in the same way that a serial poll command would
        clear them.
        """,
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
+
+        self.write("TWNDOW %s,%s" % (trace, window_mode))
+
+    video_average = Instrument.control(
+        "VAVG?", "VAVG %s",
+        """
+        Activate the video averaging function. Video averaging smooths the
+        displayed trace without using a narrow bandwidth. 'video_average' sets the IF detector to
+        sample mode (see the DET command) and smooths the trace by averaging successive traces
+        with each other. If desired, you can change the detector mode during video averaging.
```
bmoneke, line 2633:
```suggestion
        Control the video averaging function. Video averaging smooths the
        displayed trace without using a narrow bandwidth. 'video_average' sets the IF detector to
        sample mode (see the DET command) and smooths the trace by averaging successive traces
        with each other. If desired, you can change the detector mode during video averaging.
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
+        validator=joined_validators(strict_discrete_set, strict_range),
+        values=[["ON", "OFF"], arange(1, 999)],
+        set_process=lambda v: v if isinstance(v, str) else v,
+        cast=int
+    )
+
+    video_bandwidth = Instrument.control(
+        "VB?", "VB %s",
+        """
+        specifies the video bandwidth. This is normally a coupled function that
```
bmoneke, line 2653:
```suggestion
        Control the video bandwidth. This is normally a coupled function that
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
+        """,
+        validator=joined_validators(strict_discrete_set, strict_range),
+        values=[["AUTO", "MAN"], arange(1, 3e6)],
+        cast=int
+    )
+
+    video_bandwidth_to_resolution_bandwidth = Instrument.control(
+        "VBR?", "VBR %.3E",
+        """
+        Specify the coupling ratio between the video bandwidth and the
```
bmoneke, line 2677:
```suggestion
        Control the coupling ratio between the video bandwidth and the
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1392,131 +1609,1374 @@ def marker_minimum(self):
...
+        self.start_frequency_values = [0, self.MAX_FREQUENCY]
+        self.stop_frequency_values = [0, self.MAX_FREQUENCY]
+        self.frequency_offset_values = [0, self.MAX_FREQUENCY]
+        self.marker_frequency_values = [0, self.MAX_FREQUENCY]
+        self.span_values = [["FULL", "ZERO"], [0, self.MAX_FREQUENCY]]
+
+    source_leveling_control = Instrument.control(
+        "SRCALC?", "SRCALC %s",
+        """
+        Select internal or external leveling for use with the
```
bmoneke, line 2758: Please start the docstrings of all `control`s with "Control". Of all `measurement` with "Measure" or "Get" and all `setting` with "Set".

docs/api/instruments/hp/hp856xx.rst
```diff
@@ -0,0 +1,252 @@
+##################################
+HP 8560A / 8561B Spectrum Analyzer
+##################################
+
+Every unit is the base unit so for time it's seconds, frequency in hertz etc...
```
LongnoseRob, line 5:
```suggestion
Every unit is used in the base unit, so for time it is s (Seconds), frequency in Hz (Hertz) etc...
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,3165 @@
...
+        Select a trigger mode. Selected trigger conditions must be met in order for
+        a sweep to occur. The available trigger modes are listed below. When any trigger mode other
+        than free run is selected, a "T" appears on the left edge of the display.
+        """,
+        validator=strict_discrete_set,
+        values=[e for e in TriggerMode]
+    )
+
+    trace_data_a = Instrument.control(
+        "TRA?", "TRA %s",
```
LongnoseRob, line 2551: Would it notbe handiertohave all trace-methods wortk like a channel implementation?
instead of using 
`output_data = hp8650A.trace_data_a`
make it available as
`output_data = hp8650A.trace.a`

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,3165 @@
...
+        set_process=lambda v: v if isinstance(v, str) else ("%.2E" % v)
+    )
+
+    def title(self, string):
+        """Sets character data in the title area of the display, which is in
+        the upper-right corner.
+
+        A title can be up to two rows of sixteen characters each, Carriage
+        return and line feed characters are not recommended.
+        manual.
```
LongnoseRob, line 2529: The documentation here ends with the word "manual.", this could be worked into a full reference tothe manual, e.g. " See Chapter 3.14 (page 27.1) for more details"

LongnoseRob (commented):
Beside the review from @bmoneke, I had a glance at it and it looks good.
I see some improvement potential:
- How about converttng the trace & marker functionalitiy into a more channel like approach?
- configuration and calibration methods could be grouped as subclasses(e.g. `HP8560A.configuration` or `HP8650A.calibration`
- some docstrings with long list of arguments could be presented as tables.

Overall good work!

neuschs:
I will rework the code according to your feedback.
I like the idea of subclasses to structure the attributes and methods. I will presumably choose a similar structure as in the documentation restructured text.
I also had some thoughts about the trace data functions in general, there's are multiple ways to transfer the data right now it is the easiest to implement but slowest method, I will change that when I have access again to the device.

About splitting the traces into channels:
I am not totally sure if this contributes to the usability because all the trace maths functions are not channel unique but always have hard coded channel entries. So the only benefit would be the few functions/attributes which only modify one channel.

pymeasure/instruments/hp/hp856Xx.py
bmoneke: Today I merged a PR, which checks all properties to have correct docstrings.

Please merge Master into your branch to activate the test.

bmoneke:
Thanks for the suggestion of subclasses for ordering of methods, @LongnoseRob.

I'm not sure, whether we want that in pymeasure (I don't know any instrument which uses such a feature).
One could argue, that is a "calibration" channel. Maybe the best implementation would use Channel subclasses as there we have a defined interaction with the main instrument. 

What do you think @bilderbuchi?

bmoneke:
Channel subclasses (for channels or traces or similar) can help ordering and/or reduce code duplication. However, sometimes it makes it more complicated, for example if you have a single function in that channel.
Think about it. Maybe they are not that useful here (I have to look at the code again).

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1501,8 +1512,8 @@ def marker_to_center_frequency(self):
     marker_frequency = Instrument.control(
         "MKF?", "MKF %.11E",
         """
-        Place an active marker on the chosen frequency or can be queried to
-        return the frequency of the active marker. Default units are in Hertz.
+        Set an active marker on the chosen frequency or get the frequency of the active marker.
+        Default units are in Hertz.
```
bmoneke, line 1516: What about this:
```suggestion
        Control the frequency of the active marker.
        Default units are in Hertz.
```
"Control" means: get or set it.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1729,15 +1761,15 @@ def marker_delta_to_span(self):
...
         """Set the center frequency step-size equal to the frequency value of
         the active marker."""
         self.write("MKSS")
 
     marker_time = Instrument.control(
         "MKT?", "MKT %gS",
         """
-        Place a marker at a position that corresponds to a specified point in
+        Control a marker and set it to a position that corresponds to a specified point in
         time during the sweep. Default units are seconds.
```
bmoneke: What about this?
```suggestion
        Control the marker's time value. Default units are seconds.
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1753,15 +1785,17 @@ def marker_to_center_frequency_step_size(self):
...
     )
 
-    marker_signal_tracking = Instrument.control(
+    marker_signal_tracking_enabled = Instrument.control(
         "MKTRACK?", "MKTRACK %s",
         """
-        Locate the active marker and sets the center frequency to the
-        marker value. This is done after every sweep, thus maintaining the marker value at the
+        Set the active marker activated and sets the center frequency to the
+        marker value.
```
bmoneke, line 1792: As a suggestion:
```suggestion
        Control whether the center frequency follows the active marker.
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -2539,46 +2565,93 @@ def title(self, string):
...
+        When any trigger mode other than free run is selected,
+        a "T" appears on the left edge of the display.
         """,
         validator=strict_discrete_set,
         values=[e for e in TriggerMode]
     )
 
-    trace_data_a = Instrument.control(
-        "TRA?", "TRA %s",
+    def __get_trace_data(self, trace):
```
bmoneke, line 2577: private methods normally start with a single underscore only: `_get_trace_data`.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,3239 @@
...
+
+            # wait for a full sweep and than 'do_something'
+            if instr.done:
+                do_something()
+
+        """,
+        map_values=True,
+        values={True: "1"},
+        cast=str
+    )
```
bmoneke, line 1034: As you do not return anything (or does it return something different than "1"?), this should be better a method:
```suggestion
    def check_done(self):
        """Check whether the device has finished all commands previously sent."""
        answer = self.ask("DONE?")
        # TODO do here some error checking on the answer?
```

bmoneke (changes requested):
I have a few suggestions on how to improve the docstrings.

LongnoseRob (approved):
Had a look and beside the comments from @bmoneke I do not have much to add.
LGTM

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,3239 @@
...
+    DBM = "DBM"
+
+    #: DB over milli Volt
+    DBMV = "DBMV"
+
+    #: DB over micro Volt
+    DBUV = "DBUV"
+
+    #: Volts
+    V = "V"
```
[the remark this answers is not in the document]
neuschs: The way I right now have implemented / used the enumerations, it wouldn't easily work.
The instrument responds with upper case, but the value maps would be mixed case.
I tried shortly implementing a case insensitive StrEnum version, but this would not work with all current supported python versions.

Do you think this change is worth reworking all the attributes get_- ways  ?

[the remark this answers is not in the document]
bmoneke: First, The enum name can be all caps (as enums are that normally), but you can change it, if you like and it is possible. Second, the enum string can be all caps as well, as it is used by the instrument.

You should adjust the comments though (which do not influence the code).


One note: you can "call" an enum, and then it uses the string value (`some_enum("string value")`). And you can use the Index access `somd_enum["enum_name"] `.

Maybe that helps.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -0,0 +1,3257 @@
...
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set, truncated_discrete_set, \
+    joined_validators, strict_range
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class StrEnum(str, Enum):
```
bilderbuchi, line 39: Wait, this is overriding the `StrEnum` name from `enum.StrEnum` (if you are on py 3.11+), do we really want that?
If yes, I'd be more comfortable with a version switch like 
```
if sys.version_info >= (3, 11):
    from enum import StrEnum
else:
    your solution
```
plus a comment to remove it when our minimum version becomes 3.11.
but ideally we'd use packages for that. I have found but not investigated, https://pypi.org/project/backports.strenum/

bilderbuchi:
> Thanks for the suggestion of subclasses for ordering of methods, @LongnoseRob.
> 
> I'm not sure, whether we want that in pymeasure (I don't know any instrument which uses such a feature). One could argue, that is a "calibration" channel. Maybe the best implementation would use Channel subclasses as there we have a defined interaction with the main instrument.
> 
> What do you think @bilderbuchi?

I don't have a strong opinion on this, but right now I think we should not use class hierarchy for grouping config/calibration methods only. 
Also, let's use Channel when it fits well to the thing we model (i.e. an instrument channel or similar) to reduce code duplication, not use it as a code organisation method.

It seems that in the end you didn't go for it anyway, but what was the intent? To introduce an additional name layer to subdivide a large mass of names? `MyInstr.thingA.func1`, `MyInstr.thingB.func2`? I hope we don't often encounter instruments where that is necessary :D

pymeasure/instruments/hp/hp856Xx.py
bmoneke: another idea:
~~~python
try:
    from enum import StrEnum
except ImportError:
   # define it yourself
~~~

bilderbuchi: Yeah, that achieves the same thing as the upper suggestion. I like the first one better because it's more explicit -- it's clear _why_ we are checking and choosing one or the other.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -36,12 +36,15 @@
...
+    from enum import StrEnum
+except ImportError:
+    class StrEnum(str, Enum):
+        """Until StrEnum is broadly available / pymeasure relies on python <=
+        3.10.x."""
 
-    def __str__(self):
-        return self.value
+        def __str__(self):
+            return self.value
```
bmoneke, line 47: Is it necessary?

I did just `pass` in my implementation, but that might be different. Afaik it is sufficient to inherit from str and Enum.

If you need it, it is fine.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -1027,12 +1031,31 @@ def set_continuous_sweep(self):
...
     )
 
+    def check_done(self):
+        """
+        Get back (e.g. return) when all commands in a command string
+        entered before 'done' has been completed. Sending a :meth:`trigger_sweep` command
+        before 'done' ensures that the spectrum analyzer will complete a full sweep before
+        continuing on in a program.
+        Depending on the timeout a timeout error from the adapter will raise before the spectrum
+        analyzer can finish due to an extreme long sweep time.
```
bmoneke, line 1045: As it is not a property anymore, you are free to use other Docstrings, for example:
```suggestion
        Return when all commands in a command string
        entered before :meth:'check_done' has been completed. Sending a :meth:`trigger_sweep` command
        before 'check_done' ensures that the spectrum analyzer will complete a full sweep before
        continuing on in a program.
        Depending on the timeout a timeout error from the adapter will raise before the spectrum
        analyzer can finish due to an extreme long sweep time.
```

bmoneke (approved):
Two small questions.
We're getting really close.

pymeasure/instruments/hp/hp856Xx.py
neuschs: Actually it is kind of "necessary" because the enums string representation would resolve in 'class-name.value' and not just 'value'. I could change all occurances where it is used to accomodate this. What do you think ?

bmoneke: > Actually it is kind of "necessary" because the enums string representation would resolve in 'class-name.value' and not just 'value'.

I did not know this, therefore I asked.

Leave it as it is.

neuschs: Actually it is kind of "necessary" because the enums string representation would resolve in 'class-name.value' and not just 'value'. I could change all occurances where it is used to accomodate this. What do you think ?

PR #890, opened 2023-04-13
Log widget colouring and format option

CasperSchippers:
Triggered by #886, I quickly put in some effort to make available this modification to the logwidget that I had to make recently for a project.

This PR adds colouring to log messages, depending on the log-level (by utilizing the HTML functionality of the QPlainTextView).
Moreover, when a warning, error or critical message is logged, the tab starts blinking to attract attention (this blinking is stopped when the log tab is selected).
Finally, to (hopefully) help #886, I also added a way to pass fmt-string (and datefmt-string) to the formatted of the log-widget.

### DIFF
--- a/pymeasure/display/widgets/log_widget.py
+++ b/pymeasure/display/widgets/log_widget.py
@@ -79,9 +79,9 @@ class LogWidget(TabWidget, QtWidgets.QWidget):
     tab_widget = None
     tab_index = None
 
-    blink_qtimer = QtCore.QTimer()
-    blink_color = None
-    blink_state = False
+    _blink_qtimer = QtCore.QTimer()
+    _blink_color = None
+    _blink_state = False
 
     def __init__(self, name, parent=None, fmt=None, datefmt=None):
         if fmt is not None:
@@ -94,8 +94,8 @@ class LogWidget(TabWidget, QtWidgets.QWidget):
         self._layout()
 
         # Setup blinking
-        self.blink_qtimer.timeout.connect(self.blink)
-        self.handler.connect(self.blinking_start)
+        self._blink_qtimer.timeout.connect(self._blink)
+        self.handler.connect(self._blinking_start)
 
     def _setup_ui(self):
         self.view = QtWidgets.QPlainTextEdit()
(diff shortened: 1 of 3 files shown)


pymeasure/display/widgets/log_widget.py
```diff
@@ -25,38 +25,148 @@
...
+
+    blink_qtimer = QtCore.QTimer()
+    blink_color = None
+    blink_state = False
+
+    def __init__(self, name, parent=None, fmt=None, datefmt=None):
+        if fmt is not None:
+            self.fmt = fmt
+        if datefmt is not None:
+            self.datefmt = datefmt
```
bmoneke, line 90: Regarding the default values: Why don't you put the defaults into the signature?
That way everybody would know, what the defaults are and could orient at that value:
```suggestion
    tab_widget = None
    tab_index = None

    blink_qtimer = QtCore.QTimer()
    blink_color = None
    blink_state = False

    def __init__(self, name, parent=None,
                 fmt='%(asctime)s : %(message)s (%(levelname)s)',
                 datefmt='%m/%d/%Y %I:%M:%S %p',
                 ):
        self.fmt = fmt
        self.datefmt = datefmt
```

bmoneke: Ah, I see the reason know: Below, you want to pass "None" to the constructor, unless the user changed it.

CasperSchippers: Indeed, I wanted to put them as default values in there indeed, but then we need to either place them as default in all of the valling classes, or use some kwargs dictionary, but that seemed to result in relativity much needless code..

pymeasure/display/widgets/log_widget.py
```diff
@@ -25,38 +25,148 @@
...
+        if self.blink_state:
+            self.tab_widget.tabBar().setTabTextColor(
+                self.tab_index,
+                QtGui.QColor("black")
+            )
+        else:
+            self.tab_widget.tabBar().setTabTextColor(
+                self.tab_index,
+                QtGui.QColor(self.blink_color)
+            )
```
msmttchr, line 127:
```suggestion
        self.tab_widget.tabBar().setTabTextColor(
            self.tab_index,
            QtGui.QColor("black" if self.blinking_state else self.blink_color)
        )
```
Minor code optimization, if you like it.

pymeasure/display/widgets/log_widget.py
```diff
@@ -25,38 +25,148 @@
...
+        self.handler.connect(self.view.appendHtml)
 
     def _layout(self):
         vbox = QtWidgets.QVBoxLayout(self)
         vbox.setSpacing(0)
 
         vbox.addWidget(self.view)
         self.setLayout(vbox)
+
+    def blink(self):
```
msmttchr, line 117: I would rename all `blink*` methods prefix them with `_` to indicate by convention that they are internal methods.

msmttchr (approved):
I have tried it and it works fine and it is also very useful.
I left few comments for your attention.

mcdo0486:
> I have tried it and it works fine and it is also very useful.

Echoing that this feature is very useful for end users. I worked on a similar approach but happy to see this one done. A lot of important context is in the log tab and this PR draws attention to that information.

pymeasure/display/widgets/log_widget.py
CasperSchippers: Yes, that's bit more compact, thanks

pymeasure/display/widgets/log_widget.py
CasperSchippers: Yeah, that's a good one actually

pymeasure/display/widgets/log_widget.py
CasperSchippers: Is that okay for you like this? We do have other cases where the default is manually copied into the calling classes (e.g. linewidth for the plot widget), so we could also do that, but this is a relatively long default value (I find it a bit of an ugly thing that there is no nice way to do this...)

msmttchr: @bmoneke, is it ok for you to merge?

bmoneke: Yes it is, as there is no perfect solution.

PR #892, opened 2023-04-15
Keysight e36232a

jwbrooks0:
I've added the Keysight E36232A power supply.  The code is based HEAVILY on other keysight devices already on pymeasure.  

This is my first pull request ever.  Please be gentle.  

I did not use pytest because it would require the physical instrument to be present to whoever runs the test (which is unlikely).  

Once we get this relatively simple instrument successfully merged, I plan to develop a good deal more instruments.

### DIFF
--- a/pymeasure/instruments/keysight/__init__.py
+++ b/pymeasure/instruments/keysight/__init__.py
@@ -26,3 +26,4 @@ from .keysightDSOX1102G import KeysightDSOX1102G
 from .keysightN5767A import KeysightN5767A
 from .keysightN7776C import KeysightN7776C
 from .keysightE36312A import KeysightE36312A
+from .keysightE36232A import KeysightE36232A
--- a/pymeasure/instruments/keysight/keysightE36232A.py
+++ b/pymeasure/instruments/keysight/keysightE36232A.py
@@ -1,0 +1,109 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2022 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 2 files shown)


jwbrooks0:
I don't know what to do about the Python 3.10 error:
`FAILED tests/instruments/test_all_instruments.py::test_name_argument[KeysightE36232A]`
Not sure what it means....

bilderbuchi:
Thank you for your interest!

> The code is based HEAVILY on other keysight devices already on pymeasure

In this case, it's good to _reuse_ as much as possible, instead of copy-pasting code. This reduces the amount of code in this instrument, and also our maintenance burden because we don't have to fix problems N times.
We have a section in the docs on this topic: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#instruments-with-similar-features

> I did not use pytest because it would require the physical instrument to be present to whoever runs the test (which is unlikely).

Actually, we have a way to add tests that don't need a device, and can be used to check if communication is as expected (e.g. from examples in the manual): https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#protocol-tests

> Not sure what it means....

Use this pattern to define the name of your device:
https://github.com/pymeasure/pymeasure/blob/e71107f20d36c2e31489539d77d5f6bf7ecf6a5f/pymeasure/instruments/keithley/keithley2260B.py#L54

jwbrooks0:
Thanks for the replies!  I've got a bunch of reading to do to come up to speed on all this.  I'll make the corrections and re-push when I have time.  I'm looking forward to contributing to this project.  

# In this case, it's good to reuse as much as possible, instead of copy-pasting code. This reduces the amount of code in this instrument, and also our maintenance burden because we don't have to fix problems N times.
I actually have a thought on this.  One option is to do as you suggest.  Write a parent class for all Keysight E362XXA power supplies and then write subclasses for [each specific model](https://www.keysight.com/us/en/catalog/key-34773/e36200-series-autoranging-power-supply.html) (there are 4, each with specific voltage and current limits).   The other option I see is to write a more general purpose driver E362XX that queries the device for the voltage and current limits.  My thought was to do this second option.  In my present pull request, I added the functionality to query the max voltage and current, but I couldn't figure out how to successfully incorporate these values in the values parameter in the relevant Instrument.control definition.  Do you have thoughts on this?

bilderbuchi:
If what is mainly changing is different limit values, follow the first option in the section that I linked above: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#instrument-family-with-different-parameter-values
You would not query dynamically, you'd have a very slim class for each distinct model, that all derive from a base class and just adjust the limits. The key that makes this work is that the properties are `dynamic` (explained in the docs just above that section), which allows this redefinition of the values.

bilderbuchi:
What I actually tried to say, though, is to look at your command set, and at the _already implemented_ Keithley instruments and see if one of them is a close match (in features and commands); then you can base off of that. A generic power supply class is a bit farther out: #416.

jwbrooks0:
> A generic power supply class is a bit farther out: #416.

I agree with you that this is a good long term goal, but this requires a fair amount of work.

jwbrooks0:
I haven't forgotten about this pull request, but I've temporarily lost access to the device.  I should be able to get back to this next week.

bmoneke:
> I haven't forgotten about this pull request, but I've temporarily lost access to the device. I should be able to get back to this next week.

Don't worry. Nobody of us gets paid to maintain this project. If you do not have time or device access at the moment, you may continue later.

BenediktBurger:
Closing due to lack of development.
It can be reopened at any time.

PR #895, opened 2023-04-17
Add Property doc check

bmoneke:
This PR adds an automatic test for property docstrings, that they start with "Measure", "Get", "Control", or "Set".
A list of currently failing instruments is added, in order to check new PRs, but not to require improvements in almost all instruments right now.

ToDo:
- [ ] The current detection is limited to manufacturers which are loaded into `pymeasure.instruments` and to instruments loaded in these `init` files. Instruments which do not satisfy these requirements and all `Channel` classes, are not yet included. moved to another issue: #904


tests/instruments/test_all_instruments.py
```diff
@@ -105,3 +228,15 @@ def test_kwargs_to_adapter(cls):
...
         cls(SIM_RESOURCE, visa_library='@sim', kwarg_test=True)
+
+
+@pytest.mark.parametrize("prop_set", properties)
+def test_property_docstrings(prop_set):
+    device, property_name, prop = prop_set
+    if device.__name__ in legacy_instruments:
+        pytest.skip(f"{device.__name__} is in the codebase and has to be refactored later on.")
+    start = prop.__doc__.split(maxsplit=1)[0]
+    assert start in ("Control", "Measure", "Set", "Get"), (
```
bilderbuchi, line 239: We should think about the choices we want to encourage. It would be easier to just have 3 choices to mirror the control/setting/measurement triad (also thinking of future parsing effort e.g. in pyleco). In that case control/set/read or control/set/get seem most straightforward. 
However, that might feel too restrictive. :shrug: ? The first thing I'd feel like extending is to allow more words for getting ("control" and "set" are already pretty clear and not "overloaded" as terms, I think). "Measure" and "Get" or "Measure" and "Read" (considering we are _reading_ from a device) seem most suitable.
Thoughts?

bilderbuchi: Also, whatever we decide needs to be reflected in the [docs](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings).

tests/instruments/test_all_instruments.py
```diff
@@ -49,6 +49,17 @@
             if b:
                 devices.append(d)
 
+
+# Collect all properties
+# TODO add Channel properties as well: How collect all Channel classes?
```
bilderbuchi, line 54: I guess that will be tricky, considering the generality of our approach towards channel collections.

tests/instruments/test_all_instruments.py
```diff
@@ -49,6 +49,17 @@
...
+properties = []
+for device in devices:
+    for property_name in dir(device):
+        prop = getattr(device, property_name)
+        if isinstance(prop, property):
+            properties.append((device, property_name, prop))
```
bilderbuchi, line 60: Does that result in "nice" test ids in the pytest parametrisation? (Visible when a test fails) Otherwise using a namedtuple might help.

bilderbuchi (commented):
Awesome initiative! I was already thinking how we could gracefully achieve this and bit by bit migrate the existing instruments. Nice solution!

tests/instruments/test_all_instruments.py
bmoneke: > Also, whatever we decide needs to be reflected in the [docs](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings).

I wrote the test to reflect the docs (except for the fact, that we do not verify the correct type, e.g. we do not catch "Set" for a `control`). I thought "Get" would be allowed for `measurement` as well, but now I see, it is not.

For a getting property, I think we should allow a more generic form than "Measure". It seems, for example, strange to use "Measure the serial number".

tests/instruments/test_all_instruments.py
bmoneke: We could collect all `Channel` classes (and in that circumstance all `Instrument` classes) in the files below `pymeasure`.

tests/instruments/test_all_instruments.py
bmoneke: > Does that result in "nice" test ids in the pytest parametrisation? (Visible when a test fails) Otherwise using a namedtuple might help.

I use the values (class and property name) to create a meaningful error message in the test (after `assert`).

Yes, test ids were meaningless, but I used a different option to set the test_ids according to https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids

tests/instruments/test_all_instruments.py
bilderbuchi: "Measure" and "Read", then?

tests/instruments/test_all_instruments.py
bilderbuchi: Nice!

tests/instruments/test_all_instruments.py
bilderbuchi: OK, "Get" and "Measure", then? 
Get/set is nicely symmetric -- for "read" I'd expect "write" as the counterpart.

bmoneke: > OK, "Get" and "Measure", then?

So: Get, Set, Measure, and Control.

bmoneke: I adjusted the documentation accordingly.

tests/instruments/test_all_instruments.py
bmoneke: Shall we leave the collection of all the Channels for another PR?
I suggest so, because it was not that easy, when I first tried it with the instruments themselves.

tests/instruments/test_all_instruments.py
bilderbuchi: Yes, thanks!

bmoneke:
Now we have to make sure, that every PR contains these commits, lest the test fails after merging.

bilderbuchi:
Unfortunately github doesn't have an "update branch" button for that afaik.

PR #896, opened 2023-04-17
Add support for python 3.11

CasperSchippers:


bilderbuchi:
The problem was that some of the packages versions we defined did not have a python 3.11 build available, yet. I bumped them all as required. Unfortunately, both pandas and numpy don't offer a version that has builds from python 3.7 to 3.11, so now the python 3.7 builds fail.

In effect, this means we will have to wait with officially supporting 3.11 until we drop 3.7 when it is EOL in 2 months.
It does _work_ already, but we cannot _comfortably_  (with few changes and duplication) run CI on those 4 versions, and we would want to do that.

CasperSchippers:
Yeah, that seems to corroborate the impression I got; let's leave it as it is for now then and update this when 3.7 reaches EOL.

bilderbuchi:
Just dropping the pymeasure.yml that I came up with when troubleshooting this, for reference/maybe later usefulness:
```yml
name: pymeasure
channels:
  - conda-forge
dependencies:
  - cloudpickle=1.6.0
  - numpy=1.23.4  # does not support py37 anymore
  - pandas=1.5.1  # does not support py37 anymore
  - pint=0.18
  - pyqt=5.15.7
  - pyqtgraph=0.12.4
  - pyserial=3.4
  - pyvisa=1.12.0
  - pyzmq=24.0.1
  - qt=5.15.6
# Development dependencies below
  - pytest-qt=4.2.0
  - pytest-runner=5.2
  - pytest=7.2.0
  - flake8=5.0.4  # can be bumped to v6 when support for python 3.7 is dropped
  - setuptools_scm # don't pin, to get newest features
  - sphinx=4.3.1
  - sphinx_rtd_theme=1.0.0
  - pip  # don't pin, to gain newest conda compatibility fixes
  - pip:
    - pyvisa-sim==0.4.0
```

bilderbuchi:
> In effect, this means we will have to wait with officially supporting 3.11 until we drop 3.7 when it is EOL in 2 months.
> It does _work_ already, but we cannot _comfortably_ (with few changes and duplication) run CI on those 4 versions, and we would want to do that.

OK, with the new release out, now is the time to drop 3.7 and add support for 3.11. :-)
As per the comment above, the versions
* numpy=1.23.4
* pandas=1.5.1
* flake8=6.0.0
should be safe to use from 3.8-3.11

We should definitely grep the codebase for `3.7` and/or `37` to check if we can remove some old things.
As soon as CI runs as expected, depending on how much effort we can invest, we could run a tool like https://github.com/dosisod/refurb/ with appropriate settings to upgrade old usages. That would be fun to try.

bmoneke:
yes, that sounds great!

bilderbuchi:
I changed the sphinx version so that we have a common package version set across all supported python versions, so this installs now. However, now we get weird test failures surrounding Enum representations, but I don't have time right now to troubleshoot why this is occurring now.

CasperSchippers:
> I changed the sphinx version so that we have a common package version set across all supported python versions, so this installs now. However, now we get weird test failures surrounding Enum representations, but I don't have time right now to troubleshoot why this is occurring now.

I will _try_ to debug it this evening.
Regarding the sphinx version: I updated it only from 4.x to 6.x because it resolved the enum issue and is installable accros all python versions on my desktop... but for some reason that wasn't the case on CI....

bilderbuchi:
> Regarding the sphinx version: I updated it only from 4.x to 6.x because it resolved the enum issue and 


So it's the Sphinx version that is causing the enum problem? I find that very surprising for the pytest failures at least -- these are seemingly unrelated to Sphinx at all! :-o

> is installable accros all python versions on my desktop... but for some reason that wasn't the case on CI....

On my Windows desktop at least, `mamba create -n sphinx-test python=3.11 sphinx=6.2.1 sphinx_rtd_theme` fails (can't test on Linux right now).

LongnoseRob:
I see these documentation build problems also with sphinx v 7.0.XWill check which version exactly once I am back from my business trip later today..Robert EckelmannOn Jul 20, 2023, at 16:07, Christoph Buchner ***@***.***> wrote:﻿

Regarding the sphinx version: I updated it only from 4.x to 6.x because it resolved the enum issue and

So it's the Sphinx version that is causing the enum problem? I find that very surprising for the pytest failures at least -- these are seemingly unrelated to Sphinx at all! :-o

is installable accros all python versions on my desktop... but for some reason that wasn't the case on CI....

On my Windows desktop at least, mamba create -n sphinx-test python=3.11 sphinx=6.2.1 sphinx_rtd_theme fails (can't test on Linux right now).

CasperSchippers:
Ok, I found the issue I think: [python 3.11 changed the `__str__` method of the IntEnum](https://docs.python.org/3/library/enum.html#enum.IntEnum):
> Changed in version 3.11: [__str__()](https://docs.python.org/3/reference/datamodel.html#object.__str__) is now int.__str__() to better support the replacement of existing constants use-case. [__format__()](https://docs.python.org/3/reference/datamodel.html#object.__format__) was already int.__format__() for that same reason.

meaning:
```python 
class InstrMode(int, Enum):
    WAITING = 0x00
    HEATING = 0x01
    COOLING = 0x05

current_mode = InstrMode(0x01)
print(current_mode)  #  py <= 3.10 --> InstrMode.HEATING
print(current_mode)  #  py >= 3.11 --> 1
````

This _may_ also be the source of some of the other errors (not sure yet, but I saw some failing that worked with enums).

We can get back the old behaviour if we want by replacing the inheritance from `IntEnum` to `int, Enum`:
```python 
class InstrMode(int, Enum):
    WAITING = 0x00
    HEATING = 0x01
    COOLING = 0x05

current_mode = InstrMode(0x01)
print(current_mode)  #  InstrMode.HEATING
```

I think seeing the way we use the enums (i.e. to assign meaning to a numerical value) i'm inclined to say that its best to retain this old behaviour, so that we e.g. keep the human-readable format when printing an enum valu.

CasperSchippers:
I replaced all the `IntEnums` and `IntFlags` by `int, Enum` and `int, Flag` respectively in https://github.com/pymeasure/pymeasure/pull/896/commits/390691aa32eeb2622b37852fd8e253b7e26ed422 as I propsed.
Feel free to say so if you prefer to keep using the `IntEnum` and `IntFlag` with the new python 3.11 behaviour. I _think_ only the doctest needs fixing in that case.

Alas, the other issues were not solved by changing the Enums (on second look, these were indeed other issues)... but that's something for tomorrow/this weekend.

CasperSchippers:
So much for 'tomorrow', but found the cause for most (not all) of the other errors: an incompatibility between the pyvisa version and the pyvisa-sim version; which could easily be resolved by upgrading pyvisa-sim to the newest version.
Added advantage: it's now available on conda-forge. 🎉

Now nothing is installed via pip anymore, does this mean that we should remove it from the requirements list??

CasperSchippers:
Next error: wrong usage of Flag (should've been an Enum); though I wonder why this didn't show before, since when I just now tested the Flag in py3.8, it also didn't work.....

Next errors are for tomorrow/this weekend.

bmoneke:
> Now nothing is installed via pip anymore, does this mean that we should remove it from the requirements list??

I'd comment pip and the pip installs in the CI environments file, as it might be necessary in the future again and pip is a standard tool. As we do not need it right now, we can comment it.

> I think seeing the way we use the enums (i.e. to assign meaning to a numerical value) i'm inclined to say that its best to retain this 'old' behaviour, so that we e.g. keep the human-readable format when printing an enum value.

Intuitively I would use the new form, as that is the most obvious way (i.e. use `IntEnum`), but I did not look at the details.

mcdo0486:
In this change, I think `requirements/pymeasure.yml` file should be moved to `.github/requirements.yml` as it is only used for CI stuff. It has caused confusion in the past with users.

The actual requirements are in are in `setup.cfg`  for the pip package, and [conda-forge/pymeasure-feedstock](https://github.com/conda-forge/pymeasure-feedstock/) `receipe/meta.yml` for the conda package. (I don't know which requirements feed into building the readthedocs.io site.)

bilderbuchi:
> Ok, I found the issue I think: [python 3.11 changed the `__str__` method of the IntEnum](https://docs.python.org/3/library/enum.html#enum.IntEnum):

What the :angry: ? I mean, I guess the devs are doing all that is possible to avoid a Python 4 situation, but IMO if you change public behaviour like this in a minor release, and kinda glossing it over, something is not right, you might as well say "screw Semver".

> I think seeing the way we use the enums (i.e. to assign meaning to a numerical value) i'm inclined to say that its best to retain this 'old' behaviour, so that we e.g. keep the human-readable format when printing an enum value.

:+1: on this decision!

docs/dev/adding_instruments/instrument.rst
```diff
@@ -152,14 +152,15 @@ To cite the `Python documentation <https://docs.python.org/3.11/howto/enum.html>
 
     An Enum is a set of symbolic names bound to unique values. They are similar to global variables, but they offer a more useful :code:`repr()`, grouping, type-safety, and a few other features.
 
-As our signal values are often integers, the most appropriate enum types are :code:`IntEnum` and :code:`IntFlag`.
+An Enum can be combined with a type of value that you want to bind symbolic names to, by multiple inheritance of the type and :code:`Enum` (or :code:`Flag`).
+As our signal values are often integers, the most appropriate combinations are :code:`(int, Enum)` and :code:`(int, Flag)`.
```
bilderbuchi, line 156:
```suggestion
As our signal values are often integers, the most appropriate combinations are :code:`(int, Enum)` and :code:`(int, Flag)`.
This way, we get meaningful information when printing (e.g. `InstrMode.HEATING`), instead of plain integers (`1`).
```
maybe illuminate _why_ we prefer this?

bilderbuchi:
> In this change, I think `requirements/pymeasure.yml` file should be moved to `.github/requirements.yml` as it is only used for CI stuff. It has caused confusion in the past with users.

Sounds reasonable.

> The actual requirements are in are in `setup.cfg` for the pip package, and [conda-forge/pymeasure-feedstock](https://github.com/conda-forge/pymeasure-feedstock/) `receipe/meta.yml` for the conda package. (I don't know which requirements feed into building the readthedocs.io site.)

The source of truth for packaging is `setup.cfg` (to be moved into `pyproject.toml` at some point), the conda-forge recipe is basically just a copy of that.

RTD also [uses ](https://github.com/pymeasure/pymeasure/blob/master/.readthedocs.yml) the requirements.yml, so you'll have to adjust that path, too. (and to be sure we should check if the builds are still working).

bilderbuchi:
Regarding the current test failure with the HP8116A, " ValueError: 5 is not a valid Status": The problem seems to be triggered by the fact that 5 is a composite of 4 and 1? 

For the record: In all my time with pymeasure, I don't think we ever had that much work with a Python version upgrade. Let's hope the next one is less bad. :crossed_fingers:

bilderbuchi:
> Intuitively I would use the new form, as that is the most obvious way (i.e. use `IntEnum`), but I did not look at the details.

The drawback here is that we don't get nice output from IntEnums/Flags like
```python
>>> print(instr.status)
`Status.connected`
``` 
anymore, as in the new way this just prints `2` or whatever.

bilderbuchi:
> Addendum: The same holds for the `IntFlags`: in Python 3.11 a printing an IntFlag results e.g. in `7` instead of `ErrorCode.TEMPSENSOR_FAILURE|COOLER_FAILURE|HEATER_FAILURE`. Here you can also replace the `(IntFlag)` by `(int, Flag)` to retain the 'old' behaviour.

It might be that this might not be the whole picture. Although also recommended upstream (https://github.com/python/cpython/issues/94763), with this workaround I get problems:
```
>>> class MyFlag(int, Flag):
...     ONE = 1
...     TWO = 2
...     FOUR = 4
... 
>>> MyFlag(5)
ValueError: 5 is not a valid MyFlag
```
:sigh:

CasperSchippers:
> It might be that this might not be the whole picture. Although also recommended upstream ([python/cpython#94763](https://github.com/python/cpython/issues/94763)), with this workaround I get problems:

Well... this sucks.... so replacing IntFlag by (int, Flag) works well for python 3.11, but not for lower versions.... hmppff

CasperSchippers:
I'm afraid I cannot find a good solution..

I'll summarize the options/problems below:

Given the class
```
class MyFlag(IntFlag):
    ONE = 1
    TWO = 2
    FOUR = 4
```
We can
1. Keep using IntFlag & IntEnum (the most straight-forward):
    - `MyFlag(5) == MyFlag(1) | MyFlag(4)` works for all python versions
    - Printing a flag (`print(MyFlag(5))` --> `5`) does no longer show the symbolic names for python>=3.11
      - for python >= 3.11 you can use `MyFlag(5).name`, but this does not work for python<3.11
    - We need to change/drop/dissable the doctests tests the output of printing a flag/enum
    - Really reduces the usefullness of the Enums and Flags though
2. Change `IntEnum` and `IntFlag` to `(int, Enum)`, `(int, Flag)`
    - Retain the printing behaviour for all python versions
    - `MyFlag(5) == MyFlag(1) | MyFlag(4)` does only work for python >= 3.11, so this rather breaks a lot
3. Define the `IntFlag` in PyMeasure itself, depending on th epython version, by either simply importing `IntFlag` (for python < 3.11) or by defining the `IntFlag` class as `(int, Flag)` (and same for `enum`)
    - Behaviour remains unchanged between different python versions
    - Can easily include `boundary=KEEP` as the default behaviour
    - All imports of `IntEnum` and `IntFlag` need to be through PyMeasure
    - Likely be confusing for users/devs as the `IntFlag` and `IntEnum` classes are not consistent with the python definition (could of course use a different name to resolve this confusion)
4. Replace `IntFlag` by `Flag`
    - Retain the printing behaviour for all python versions
    - loses the integer behaviour (e.g. `MyFlag(5) == 5`)


I guess the most viable options are options 3 or 1; so main question I think is whether retaining the nice printing feature is worth having a 'custom' Flag class?

bmoneke:
I'm for option 1, as it sticks with the core python (for better or worse) and does not introduce new dependencies (our own flags), which we have to maintain. 



Am 23. Juli 2023 17:55:43 MESZ schrieb Casper Schippers ***@***.***>:

bilderbuchi:
Thank you for the detailed analysis!
I'd like to see option 3 (with custom names) used to preserve consistent behaviour across python version, and preserve the easy name display, which is a significant benefit of enums. 
I'm disappointed that this apparently has to be a user-side fix for a backwards-incompatible language change. I've also seen custom enums in other projects, although I agree it's a hassle that we'd have to maintain that, and make sure our users use it and are aware.

I think 4 is not viable as we rely on the integer behaviour rather a lot to decompose status codes.

I'm puzzled why 2 is not working correctly. I think I'll post a Stackoverflow question, maybe someone else knows _a better way_ (Raymond Hettinger? :D)

In the end (especially if there is satisfying answer re: 2, the pragmatic choice will probably be 1, and we document/educate that on Python 3.11, you have to add `.name` to get the most useful info out of enums. :-|

bilderbuchi:
Waaaiiiiit, I just realised that the "implicit" printing in the console still works as expected. In a Python 3.11 Ipython REPL, with this code:
```python
from enum import IntFlag
class ErrorCode(IntFlag):
    TEMP_OUT_OF_RANGE = 8
    TEMPSENSOR_FAILURE = 4
    COOLER_FAILURE = 2
    HEATER_FAILURE = 1
    OK = 0

received_from_device = 7

def myfunc(statuscode):
   """Probably representative of internal usage."""
   return ErrorCode(statuscode)
```
I get
```python
>>> ErrorCode(7)
<ErrorCode.TEMPSENSOR_FAILURE|COOLER_FAILURE|HEATER_FAILURE: 7>

>>> print(ErrorCode(7))
7
>>> myfunc(7)
<ErrorCode.TEMPSENSOR_FAILURE|COOLER_FAILURE|HEATER_FAILURE: 7>

>>>repr(ErrorCode(7))
'<ErrorCode.TEMPSENSOR_FAILURE|COOLER_FAILURE|HEATER_FAILURE: 7>'

```
So, maybe our doctests are bad, in that we have an explicit `print` that we shouldn't have? 
A user working interactively would get the status back as an `IntFlag` and it would display as desirable? Some instrument doing some printing/formatting internally would have to take _some_ care and e.g. use a `repr` when processing the status flags.

Could we just continue using plain `IntFlag`, adjust our doctests slightly (to be more representative and avoid the problematic deviation) and option 1 (with some documentation as mentioned above) is suddenly much more attractive?
Am I off-base here?

bmoneke:
That behavior makes sense:

If you print, you want the "real" value.
However a return value in an interactive shell is the enum text, as it is only returned to a human and not to code. 

So it seems useful again. Great! 



Am 23. Juli 2023 22:24:02 MESZ schrieb Christoph Buchner ***@***.***>:

bmoneke:
I just tested it in a normal terminal, the behavior is the same as in ipython

CasperSchippers:
Ah yeah, indeed, that is a good point; stupid that I didn't check this (I tested everything in scripts..., which also does mean that when printing in scripts you do need to explicitly print the repr, so we might want to add that to the documentation).
You are probably right that this is the best option.

> If you print, you want the "real" value.

On this regard I do disagree: I would say that printing is also primarily meant for returning a value to a human (which is why I do not completely understand the change in the Enum library, but ok).

docs/dev/adding_instruments/instrument.rst
CasperSchippers: This is no longer relevant I think.

CasperSchippers:
I changed to using the `IntFlags` and `IntEnums` and updated the doctest to work on all python versions.
(Force pushed to drop the IntFlag to int, Flag commit and related commits)

Am I happy to see all check passing now. Someone should be taught about backwards compatibility....

CasperSchippers:
I changed to using the `IntFlags` and `IntEnums` and updated the doctest to work on all python versions.
(Force pushed to drop the IntFlag to int, Flag commit and related commits)

bilderbuchi:
> Someone should be taught about backwards compatibility....

Indeed, that was a frustrating experience. 😠 
As the python devs don't mark breaking changes as such, it seems we have to go through all *changes* in the what's new document by hand.😔

bmoneke:
And it shows again, that it is important to have tests in place 😁

CHANGES.rst
```diff
@@ -1,5 +1,6 @@
 Upcoming version
 ================
+- Dropped support for Python 3.7, adding support for Python 3.11
```
bmoneke, line 3: Should it not be also `added`?
```suggestion
- Dropped support for Python 3.7, added support for Python 3.11
```

pymeasure/instruments/lakeshore/lakeshore211.py
bmoneke, line 1: Good, that you found that.

bmoneke (approved):
Looks good to me.

Thanks for your work @CasperSchippers and @bilderbuchi with fixing these problems.

pymeasure/instruments/lakeshore/lakeshore211.py
CasperSchippers: Found it thanks to the tests we have in place; they started complaining on python 3.11, but not on the lower python versions.

CasperSchippers:
> As soon as CI runs as expected, depending on how much effort we can invest, we could run a tool like https://github.com/dosisod/refurb/ with appropriate settings to upgrade old usages. That would be fun to try.

I was thinking (and briefly playing around with) such tools ([refurb](https://github.com/dosisod/refurb) and [pyupgrade](https://github.com/asottile/pyupgrade)), but it seems that is a bit more work since it mostly just points you in directions that can be improved (it fixes some issues for you, but still leaves a lot that can be improved). So, seeing that this PR already includes quite some changes, it is probably best to do such code-improvements in a separate PR.

Hence, I'll go ahead and merge this.

PR #897, opened 2023-04-18
Fixes Ametek7270

seb5g:
I'm using this instrument right now and noticed many failures in the code related to:

* not asserting the return value when writing values (dne now by subclassing Instruments with a dedicated check_errors method)
* using right command to translate from index value to float values (Timeconstant and sensitivity)
* sensitivity values are dynamically set depending the IMODE (voltage/current)
* adding other controls/measurements

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -60,5 +60,7 @@ Scott Candey
 Tom Verbeure
 Max Herbold
 Alexander Wichers
+Ashok Bruno
 Robert Roos
 Sebastien Weber
+Sebastian Neusch
--- a/docs/api/instruments/hp/hp856xx.rst
+++ b/docs/api/instruments/hp/hp856xx.rst
@@ -1,0 +1,254 @@
+##################################
+HP 8560A / 8561B Spectrum Analyzer
+##################################
+
+Every unit is used in the base unit, so for time it is s (Seconds), frequency in Hz (Hertz) etc...
+
+-------------------------------------
+Generic Specific Attributes & Methods
+-------------------------------------
+.. contents:: Content
+    :depth: 3
+    :local:
+
+^^^^^^^
+General
+^^^^^^^
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.preset
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.attenuation
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.amplitude_unit
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.trigger_mode
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.detector_mode
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.coupling
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_auto_couple
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_linear_scale
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.logarithmic_scale
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.threshold
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_title
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.status
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.check_done
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.request_service
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.errors
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.save_state
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.recall_state
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.request_service_conditions
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_maximum_hold
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_minimum_hold
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.reference_level_calibration
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.reference_offset
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.reference_level
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.display_line
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.protect_state_enabled
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.mixer_level
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.frequency_counter_mode_enabled
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.frequency_counter_resolution
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.adjust_all
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.adjust_if
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.hold
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.annotation_enabled
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_crt_adjustment_pattern
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.display_parameters
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.firmware_revision
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.graticule_enabled
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.serial_number
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.id
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.elapsed_time
+
+
+^^^^^^^^^^^^
+Demodulation
+^^^^^^^^^^^^
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.demodulation_mode
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.demodulation_agc_enabled
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.demodulation_time
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.squelch
+
+^^^^^^^^^
+Frequency
+^^^^^^^^^
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.start_frequency
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.stop_frequency
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.center_frequency
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.frequency_offset
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.frequency_reference_source
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.span
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_full_span
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.frequency_display_enabled
+
+^^^^^^^^^^^^^^^^^^^^
+Resolution Bandwidth
+^^^^^^^^^^^^^^^^^^^^
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.resolution_bandwidth
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.resolution_bandwidth_to_span_ratio
+
+^^^^^
+Video
+^^^^^
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.video_trigger_level
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.video_bandwidth_to_resolution_bandwidth
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.video_bandwidth
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.video_average
+
+^^^^^^^^^^^^^^^^^^
+FFT & Measurements
+^^^^^^^^^^^^^^^^^^
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.create_fft_trace_window
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.get_power_bandwidth
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.do_fft
+
+^^^^^
+Trace
+^^^^^
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.view_trace
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.get_trace_data_a
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.get_trace_data_b
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_trace_data_a
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_trace_data_b
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.trace_data_format
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.save_trace
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.recall_trace
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.clear_write_trace
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.subtract_display_line_from_trace_b
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.exchange_traces
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.blank_trace
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.trace_a_minus_b_plus_dl_enabled
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.trace_a_minus_b_enabled
+
+^^^^^^
+Marker
+^^^^^^
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.search_peak
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.marker_amplitude
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_marker_to_center_frequency
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.marker_delta
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.marker_frequency
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_marker_minimum
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.marker_noise_mode_enabled
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.deactivate_marker
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.marker_threshold
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.peak_excursion
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_marker_to_reference_level
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_marker_delta_to_span
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_marker_to_center_frequency_step_size
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.marker_time
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.marker_signal_tracking_enabled
+
+^^^^^^^^^^^^^^^^^
+Diagnostic Values
+^^^^^^^^^^^^^^^^^
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.sampling_frequency
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.lo_frequency
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.mroll_frequency
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.oroll_frequency
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.xroll_frequency
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.sampler_harmonic_number
+
+^^^^^
+Sweep
+^^^^^
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.sweep_single
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.sweep_time
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.sweep_couple
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.sweep_output
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.set_continuous_sweep
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.trigger_sweep
+
+^^^^^^^^^^^^^
+Normalization
+^^^^^^^^^^^^^
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.normalize_trace_data_enabled
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.normalized_reference_level
+.. autoattribute:: pymeasure.instruments.hp.hp856Xx.HP856Xx.normalized_reference_position
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Open/Short Calibration (Reflection)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.recall_open_short_average
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.store_open
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.store_short
+
+^^^^^^^^^^^^^^^^
+Thru Calibration
+^^^^^^^^^^^^^^^^
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.store_thru
+.. automethod:: pymeasure.instruments.hp.hp856Xx.HP856Xx.recall_thru
+
+-------------------------------------
+HP8560A Specific Attributes & Methods
+-------------------------------------
+
+.. autoclass:: pymeasure.instruments.hp.HP8560A
+    :members:
+    :show-inheritance:
+
+-------------------------------------
+HP8561B Specific Attributes & Methods
+-------------------------------------
+
+.. autoclass:: pymeasure.instruments.hp.HP8561B
+    :members:
+    :show-inheritance:
+
+------------
+Enumerations
+------------
+
+.. autoclass:: pymeasure.instruments.hp.hp856Xx.AmplitudeUnits
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.hp.hp856Xx.MixerMode
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.hp.hp856Xx.Trace
+    :members:
+    :show-inheritance:
+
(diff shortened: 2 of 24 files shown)


pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -30,6 +31,15 @@
 log.addHandler(logging.NullHandler())
 
 
+class Instrument(Instrument):
+
+    def check_errors(self):
+        errors = []
+        if self.read() != '':
+            errors.append(f'Incorrect return from previously set property')
+        return errors
```
bmoneke, line 40: As you do not do any work with the lists, I'd do the logic differently:
```suggestion
        if self.read() == '':
            return []
        else:
            return ['Incorrect return from previously set property']
```

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -50,31 +66,36 @@ class Ametek7270(Instrument):
...
+        "SEN", "SEN %d",
         """ A floating point property that controls the sensitivity
         range in Volts, which can take discrete values from 2 nV to
         1 V. This property can be set. """,
         validator=truncated_discrete_set,
         values=SENSITIVITIES,
-        map_values=True
+        map_values=True,
+        check_set_errors=True,
+        dynamic=True
```
bmoneke, line 77:
```suggestion
        dynamic=True,

```
It's good practice to finish even the last line with a comma. That way, you do not have to touch lines (see above `map_values=True`, while adding/removing unrelated lines.

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -184,16 +218,25 @@ def __init__(self, adapter, name="Ametek DSP 7270", **kwargs):
...
+        self.sensitivity_values = self.SENSITIVITIES_IMODE[0]
 
     def set_differential_mode(self, lineFiltering=True):
         """ Sets instrument to differential mode -- assuming it is in voltage mode """
-        self.write("VMODE 3")
-        self.write("LF %d 0" % 3 if lineFiltering else 0)
+        self.ask("VMODE 3")
+        self.ask("LF %d 0" % 3 if lineFiltering else 0)
+
+    def set_current_mode(self, low_noise=False):
```
bmoneke, line 229: Please add a docstring describing, what the method does.

bmoneke:
@seb5g thanks for your contribution. It's good if people help keeping the code base clean.
You might add yourself to the Authors.txt.

Would you mind adding a few protocol tests? They help us maintainers to ensure, that the code continues to work as expected without having an actual device.
If you need support writing these tests, don't hesitate to ask.

seb5g:
> Would you mind adding a few protocol tests? They help us maintainers to ensure, that the code continues to work as expected without having an actual device. If you need support writing these tests, don't hesitate to ask.

Hi, I'm not sure to see the point of your protocol tests. As it is (like Mocking the instrument) it looks as if I'm programming twice the communication protocol... I see only one reason to do so, to prevent someone to change the code base as I did just now. But in my case, the code base was not working properly so the tests would not also! It is only by trying with the instrument that i could figure that out, not with the mock tests...

bmoneke:
These mock tests help us to make sure, that the communication generated is the same. That way we can refactor the instrument code or even the pymeasure core code (e.g. `control`).

So yes, they kind of freeze the communication, but that is essential for us maintainers, who cannot test the code with a device.

There are two different development paths of tests:

1. You have a manual with examples. Then you write the tests and you can verify, that you code fulfills these examples.
2. You have a working code and want to make sure, that it continues to work: You write tests for that code. To facilitate test writing, I'm currently working at a test generator (see #882 ), which observes the communication to the device and writes test accordingly. If you have a working code, you can generate the tests easily.

seb5g:
Hello @bmoneke , would you accept this pull request?
Cheers
Seb

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -50,34 +68,39 @@ class Ametek7270(Instrument):
...
         """ A floating point property that controls the sensitivity
         range in Volts, which can take discrete values from 2 nV to
         1 V. This property can be set. """,
         validator=truncated_discrete_set,
         values=SENSITIVITIES,
-        map_values=True
+        map_values=True,
+        check_set_errors=True,
+        dynamic=True,
+
```
bmoneke, line 80: I think that empty line is superfluous
```suggestion

```

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -50,34 +68,39 @@ class Ametek7270(Instrument):
...
     slope = Instrument.control(
         "SLOPE", "SLOPE %d",
         """ A integer property that controls the filter slope in
         dB/octave, which can take the values 6, 12, 18, or 24 dB/octave.
         This property can be set. """,
         validator=truncated_discrete_set,
         values=[6, 12, 18, 24],
-        map_values=True
+        map_values=True,
+        check_set_errors=True
```
bmoneke, line 91: You see, that the line `map_values` got changed, because you add another line, therefore it is good practice, to end every line, even the last one, with a comma:
```suggestion
        map_values=True,
        check_set_errors=True,
```

bmoneke, line 91: You see, that the line `map_values` got changed, because you add another line, therefore it is good practice, to end every line, even the last one, with a comma:
```suggestion
        map_values=True,
        check_set_errors=True,
```

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -172,28 +208,49 @@ class Ametek7270(Instrument):
...
+            self.sensitivity_values = self.SENSITIVITIES_IMODE[1]
 
     def set_channel_A_mode(self):
         """ Sets instrument to channel A mode -- assuming it is in voltage mode """
-        self.write("VMODE 1")
+        self.ask("VMODE 1")
+
+    def identification(self):
+        """Get the instrument ID and firmware version"""
+        return f"{self.id}/{self.ask('VER')}"
```
bmoneke, line 253: First I thought, you ment the "id" property, then I notice, that you use that property and add the firmware version.
Why does the ID + firmware version give another form of identification?

tests/instruments/ametek/generate_tests.py
bmoneke, line 1: @bilderbuchi should we add these test files?
I don't think so.

tests/instruments/ametek/generate_tests.py
```diff
@@ -0,0 +1,108 @@
...
+    dev = Ametek7270('USB0::0x0A2D::0x001B::16216344::RAW')
+    dev.id
+    dev.identification()
+    dev.shutdown()
+
+
+if __name__ == '__main__':
+    # run()
+    # generate()
+    pass
```
bmoneke, line 108: What's that for?

bmoneke:
A first round of comments for your code. Normally it takes a few rounds of rework, but that ensures good and uniform style in the codebase. Your code already looks good, so eventually it will enter pymeasure.
Just ping us (with a message or clicking on "request review"), whenever you want us to have another look at the code.

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -30,6 +31,18 @@
...
+    def check_errors(self):
+        if self.read() == '':
+            return []
+        else:
+            return ['Incorrect return from previously set property']
+
+    def ask(self, command, query_delay=0):
+        return super().ask(command, query_delay).strip()
+
+
```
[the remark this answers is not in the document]
seb5g: That's a good question... It was intended to highlight the change compared to the standard Instrument class. I moved it into my Ametek instrument class directly.

I added some docstrings into the check_errors and ask reimplemented method explaining why, basically:

        """Usually the properties use the `values` method that adds a strip call,
        however several methods use directly the result from ask to be casted into some other types.
        It should therefore also add the strip here
        """

pymeasure/instruments/ametek/ametek7270.py
seb5g: done in next commit

pymeasure/instruments/ametek/ametek7270.py
seb5g: done in next commit

seb5g: done in next commit

pymeasure/instruments/ametek/ametek7270.py
seb5g: well it is just a matter of better identification

tests/instruments/ametek/generate_tests.py
seb5g: it's up to you, now that the test have been generated it doesn't matter

tests/instruments/ametek/test_ametek7270.py
```diff
@@ -0,0 +1,526 @@
...
+    ) as inst:
+        assert inst.x2 == ''
+
+
+def test_y2_getter():
+    with expected_protocol(
+            Ametek7270,
+            [(b'Y2.', b'')],
+    ) as inst:
+        assert inst.y2 == ''
```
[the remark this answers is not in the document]
seb5g: well it depends on the lockin mode (single reference or dual harmonic). In single reference mode, only x, y are read as floats all properties indexed with 1 and 2 are empty string and it is the opposite in dual harmonic mode. 

So how do you expect me to write tests in such a situation?

pymeasure/instruments/ametek/ametek7270.py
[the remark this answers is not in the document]
bmoneke: Do you expect excess spaces, which you have to strip?

At least the conversion to int or float handles spaces well:
`int("  563 ") == 563`

tests/instruments/ametek/generate_tests.py
bmoneke: git rm --cached file1.txt
git commit -m "remove file1.txt"

See also http://stackoverflow.com/questions/2047465/ddg#2047477

pymeasure/instruments/ametek/ametek7270.py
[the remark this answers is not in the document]
seb5g: no it actually adds a new line character \n (which is removed with the strip). i tried to add this new line character in the termination character like '\n\x00' but it does not well.

tests/instruments/ametek/test_ametek7270.py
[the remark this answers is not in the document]
bmoneke: With the tests, I realized the type of the return value, so we are talking about the implementation itself.

I would not expect an empty string as a response value, but 0, None, or nan or that an exception is raised.

As it is an invalid query, an exception, None, or nan can be understood. 

Maybe raising ValueError is the best, as you try to read a property, which is invalid to query right now.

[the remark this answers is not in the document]
bmoneke: What would expect, if a property is not defined?
0 could bot be distinguished from a valid response, so 0 is not a good answer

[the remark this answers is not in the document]
seb5g: The thing is (and I tested it on the instrument right now), no error is generated from the instrument only empty response. I mean the call on the property is not invalid it is just undefined.

 I'm not sure I understand, I mean I'm not the one writing the response in the implementation but the driver (instrument) is. If it replies, in certain circumstances, with an empty string, what should I do?? Should I add a check_get_error thing and depending on the mode raise a ValueError, is that what you mean? But then it is not more coherent with the instrument driver documentation...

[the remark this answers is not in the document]
bmoneke: These drivers, we are writing, hide some details from the user.
Therefore, we introduce changes, we consider appropriate.
Another instrument, for example, returns the temperature times 10. The python code divides by ten and returns an integer, such that the user gets a temperature in degrees.
So there's no problem deviating from the pure instrument communication.

An empty string is not a good response value for an undefined property.
Imagine a user reads that property and tries to calculate something from it.
Then he will get an obscure error, that you can't add Str to int (for example).

If you raise a meaningful exception, the user will notice immediately, what the problem is.

Maybe the best place is the `read` method.

Can you follow me?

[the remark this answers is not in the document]
seb5g: yeah get it, I'll try to write something!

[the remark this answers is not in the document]
seb5g: I looked into the read reimplementation, but how to know where this read comes from and that the empty string response is wrong (in some cases it is not, for instance after a property has been set). Then I thought about the casting, because a float on an empty string raise a ValueError but I saw you discard this error in the values method:

```
 for i, result in enumerate(results):
            try:
                if cast == bool:
                    # Need to cast to float first since results are usually
                    # strings and bool of a non-empty string is always True
                    results[i] = bool(float(result))
                else:
                    results[i] = cast(result)
            except Exception:
                pass  # Keep as string
``` 

and the ckeck_get_error=True will just also call the check_errors

So I tried another thing using the get_process method:

```


    y = Instrument.measurement("Y.",
                               """ Reads the Y value in Volts """,
                               get_process=check_read_not_empty,
                               )
    x1 = Instrument.measurement("X1.",
                                """ Reads the first harmonic X value in Volts """,
                                get_process=check_read_not_empty,
                                )
```

with `check_read_not_empty` is:

```
def check_read_not_empty(value):
    """Called by some properties to check if the reply is not an empty string
    that would mean the properties is currently invalid (probably because the reference mode
    is on single or dual)"""
    if value == '':
        raise ValueError('Invalid response from measurement call')
    else:
        return value
```

I tested this on the instrument, changing dynamically the reference mode, and it works well. What do you think?

pymeasure/instruments/ametek/ametek7270.py
bmoneke: I'm still not convinced, that "identification" (the long form of ID) is a good name for that combination.

Do you need the firmware version (which could basically change) as an additional identification to the serial number?

tests/instruments/ametek/test_ametek7270.py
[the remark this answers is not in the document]
bmoneke: That last code looks good. For reasons like this (or calculating a Value etc.), we have "get_process".

Maybe you could try to make the error message more meaningful, for example hinting at the possibly wrong mode.

It's also good to document in the instrument's docstring this changing between sets of commands.

The idea is, that a user does not have to look at the manual while using your driver.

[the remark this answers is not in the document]
mcdo0486: > alright now how should I modify the tests on these properties that are raising an error? The generator is not catching the error. Do you have a layout for this? I know about pytest.raises(ValueError) but how to write it with the expected protocol?

You can checkout the DSP Signal Recovery lockin amplifier tests:
https://github.com/pymeasure/pymeasure/blob/master/tests/instruments/signalrecovery/test_dspbase.py

Or the TDK lambda tests in my PR branch here:
https://github.com/mcdo0486/pymeasure/tree/dev/tdk_lambda_40-38_80-65/tests/instruments/tdk

The TDK does something kind of similar in the sense that "OK" gets returned by instrument so I needed to use `check_set_errors`. You can see the expected protocol examples there. 

Hope that helps.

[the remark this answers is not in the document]
bmoneke: You can put both with statements into each other:
```python
With pytest.raises():
    With expected_protocol() as inst:
        Pass
```

I'd use a parameterized test (getting all these attributes with "geattr(Inst, Name)") for all the raised exceptions, as they are all the same.

You could make these tests work with returning correct values.


Btw. You said, that the instrument returns a newline char at the end of every message. Why don't you include them in the tests? That way you test, whether your code interprets the messages correctly.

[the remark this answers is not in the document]
mcdo0486: > alright now how should I modify the tests on these properties that are raising an error? The generator is not catching the error. Do you have a layout for this? I know about pytest.raises(ValueError) but how to write it with the expected protocol?

I tried throwing a PR to your branch, but probably easiest to give the example code here:

```
def test_x1_getter():
    with pytest.raises(ValueError):
        with expected_protocol(
                Ametek7270,
                [(b'X1.', b'')],
        ) as inst:
            assert inst.x1 == ''
```

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -98,36 +101,45 @@ class Ametek7270(Instrument):
         validator=truncated_discrete_set,
         values=TIME_CONSTANTS,
         map_values=True,
-        check_set_errors=True
+        check_set_errors=True,
     )
 
     x = Instrument.measurement("X.",
-                               """ Reads the X value in Volts """
+                               """ Reads the X value in Volts """,
```
bmoneke, line 108: I know, that you did not change the docstring itself, so you do not have to change it, but I want to raise your attention, that that docstring is not according to our coding style.

For your information: A merely gettable property should start with "Measure" or "Get", a settable one with "Set" and a property gettable and settable with "Control".
For more details, see https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings

If you want to, you may improve the docstrings. In that case, I'd kindly ask you to remove the Ametek720 from the `grandfathered_docstring_instruments` in "tests/instruments/test_all_instruments.py" such that docstring tests will be performed on this instrument (I just saw, that it is twice in that list).

You might do that docstring rework in a second PR, if you want, to finish this content rework first.

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -137,94 +149,127 @@ class Ametek7270(Instrument):
...
-                                cast=str
+                                cast=str,
                                 )
 
     def __init__(self, adapter, name="Ametek DSP 7270",
                  read_termination='\x00',
                  write_termination='\x00',
-                 **kwargs):
+                 **kwargs,
+                 ):
```
bmoneke, line 235: In this case, you do not need to add the additional comma after kwargs, as `**kwargs` is normally the last entry, collecting all additional keyword arguments. However, adding the comma and new line, is not wrong either.

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -137,94 +149,127 @@ class Ametek7270(Instrument):
...
+        """Usually the properties use the `values` method that adds a strip call,
+        however several methods use directly the result from ask to be casted into some other types.
+        It should therefore also add the strip here
+        """
+        return super().ask(command, query_delay).strip()
+
+    def set_reference_mode(self, mode: int = 0):
+        """Set the instrument in Single, Dual or harmonic modes
+        :param mode: the integer specifying the mode: 0 for Single, 1 for Dual harmonic and 2 for Dual
+        reference
```
bmoneke, line 267: Small improvements:
```suggestion
        """Set the instrument in Single, Dual or harmonic mode.

        :param mode: the integer specifying the mode: 0 for Single, 1 for Dual harmonic, and 2 for
            Dual reference.

```
Reasons:
- The first line is a sentence in imperative mode, which should end with a period
- Between the summary line (first line) and additional text, should be an empty line.

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -41,6 +57,11 @@ class Ametek7270(Instrument):
         200.0e-3, 500.0e-3, 1.0
     ]
 
+    SENSITIVITIES_IMODE = {0: SENSITIVITIES,
+                           1: list(np.array(SENSITIVITIES) * 1e-6),
```
bmoneke, line 61: Is it worth to import the full numpy package for that list of constants?
Isn't it better to generate that list once and then copy the results into the file?

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -50,161 +71,242 @@ class Ametek7270(Instrument):
...
+        if self.read() == '':
+            return []
+        else:
+            return ['Incorrect return from previously set property']
+
+    def ask(self, command, query_delay=0):
+        """Usually the properties use the `values` method that adds a strip call,
+        however several methods use directly the result from ask to be casted into some other types.
+        It should therefore also add the strip here
+        """
```
bmoneke, line 261:
```suggestion
        """Send a command and read the response, stripping white spaces.

        Usually the properties use the `values` method that adds a strip call,
        however several methods use directly the result from ask to be casted into some other types.
        It should therefore also add the strip here, as all responses end with a newline character.
        """
```
I reformatted the docstring according to https://peps.python.org/pep-0257/

First a summary line, then a more extensive description (here I included, why it is necessary).

bmoneke (changes requested):
Hey @seb5g 
thanks for your reformatting. The code looks good, a few questions remain still open (e.g. making the tests pass) and I added a few more. Mainly on docstring style.

pymeasure/instruments/ametek/ametek7270.py
seb5g: Ok I'll see that later

seb5g:
One side question for you guys maintaining Pymeasure:
I'm the main author/developer of PyMoDAQ (pymodaq.cnrs.fr) that among other things has the same purpose as pymeasure (but focused on proposing highly modular GUI for data recording as a function of variable parameters). I think it would be highly  valuable for both our projects to be more aware of what each package do. For this I'd like to invite one of you on the PyMoDAQ's days, a workshop I'm organising next 16/17 october in Paris to present features of pymeasure and discuss of what could be done at the interface of our packages.

Are you interested?

pymeasure/instruments/ametek/ametek7270.py
seb5g: Well I guessed numpy was probably used in pymeasure but you're right, I modified this with list comprehension:

```
1: [sen * 1e-6 for sen in SENSITIVITIES],
```

pymeasure/instruments/ametek/ametek7270.py
seb5g: done

seb5g:
There is something I don't quite understand with tests that should raise a Value Error. I just committed a try but the expected protocol in that case is just returning the value i've put in the protocol. Not raising the ValueError. I feel the ValueError should be in the expected protocol object but not sure one can do that...thanks for your thought on this.

mcdo0486:
> There is something I don't quite understand with tests that should raise a Value Error. I just committed a try but the expected protocol in that case is just returning the value i've put in the protocol. Not raising the ValueError. I feel the ValueError should be in the expected protocol object but not sure one can do that...thanks for your thought on this.

`ValueError` is supposed to happen when the instrument returns `''` right?

So you can change https://github.com/PyMoDAQ/pymeasure/blob/346e57b2daf4a41783f473d236f461c8d474bcb0/tests/instruments/ametek/test_ametek7270.py#L94 to return empty byte string.

```
def test_failing_properties(method, command):
    """in standard single reference mode, these tests should raise a ValueError"""
    with pytest.raises(ValueError):
        with expected_protocol(
                Ametek7270,
                [(f'{command}'.encode(), b'')]
        ) as inst:
           getattr(inst, method) == 0.0
```

pymeasure/instruments/ametek/ametek7270.py
bmoneke: Yes, that is even better!

bmoneke:
> One side question for you guys maintaining Pymeasure: I'm the main author/developer of PyMoDAQ (pymodaq.cnrs.fr) that among other things has the same purpose as pymeasure (but focused on proposing highly modular GUI for data recording as a function of variable parameters). I think it would be highly valuable for both our projects to be more aware of what each package do. For this I'd like to invite one of you on the PyMoDAQ's days, a workshop I'm organising next 16/17 october in Paris to present features of pymeasure and discuss of what could be done at the interface of our packages.
> 
> Are you interested?

Oh, interesting project, thanks for telling about that module. It's always good to connect and learn from each other. Thanks for the invitation. I made my fellow maintainers aware of your offer, let's see, if we can manage an attendance.

By the way, we are developing a communication protocol [LECO](https://github.com/pymeasure/leco-protocol) for connecting instrument drivers and data control/acquisition among different processes and computers. Maybe our thought can help you with your remote control option and maybe you can give some feedback into our design process.

seb5g:
hi @bmoneke , I think this PR is good for another review!

seb5g:
Oh and by the way I think most of the functionalities in this driver is redundant with the one is signalrecovery DSP7265 (actually Ametek bought SignalRecovery hence the different name) but i would in fact suggest to move everything in signalrecovery as it is the brand manufacturer and create some DSP base class with everything shared among the two instruments.

bmoneke:
> Oh and by the way I think most of the functionalities in this driver is redundant with the one is signalrecovery DSP7265 (actually Ametek bought SignalRecovery hence the different name) but i would in fact suggest to move everything in signalrecovery as it is the brand manufacturer and create some DSP base class with everything shared among the two instruments.

You can keep them in two distinct manufacturers, if you deem it best to find the device. I'd go for the name printed on the device, as a potential user will search for that brand name. A similar story is with Teledyne (new name) and LeCroy (old name) or hp and keysight.
What we did in these cases, is a link in the manufacturer documentation (e.g. Teledyne says, that LeCroy contains older models and LeCroy states that Teledyne contains newer models).

Shall I review this instrument first and merge it and afterwards you create that common DSP class, or do you want to do the parent DSP class in this PR?

mcdo0486:
> > Oh and by the way I think most of the functionalities in this driver is redundant with the one is signalrecovery DSP7265 (actually Ametek bought SignalRecovery hence the different name) but i would in fact suggest to move everything in signalrecovery as it is the brand manufacturer and create some DSP base class with everything shared among the two instruments.
> 
> You can keep them in two distinct manufacturers, if you deem it best to find the device. I'd go for the name printed on the device, as a potential user will search for that brand name. A similar story is with Teledyne (new name) and LeCroy (old name) or hp and keysight. What we did in these cases, is a link in the manufacturer documentation (e.g. Teledyne says, that LeCroy contains older models and LeCroy states that Teledyne contains newer models).
> 
> Shall I review this instrument first and merge it and afterwards you create that common DSP class, or do you want to do the parent DSP class in this PR?

I also agree you should keep them as separate manufacturers. 

However I think your `Ametek7270` class can inherit the `DSPBase` class from `pymeasure/instruments/signalrecovery/dsp_base.py` to create a new class, like what is going on with `DSP7265` from `pymeasure/instruments/signalrecovery/dsp7265.py`.

I'm not the original driver author but I refactored the driver to a `DSPBase` for use with `DSP7225`. I have access to 7225 model, and I think I can get access to the 7265 model. It would be good to implement any fixes to `DSPBase`, some of the functionality I haven't tested because it only applies to 7265 and I guess 7270.

seb5g:
I'd like you to accept the PR as it is right now. 

And as @mcdo0486 suggested, I'll work later on using the DSPBase that includes almost everything already in the Ametek7270. 

One question on the way for @mcdo0486 : why modifying the read method to replace the Null character and not use it as termination character as I do?

pymeasure/instruments/ametek/ametek7270.py
```diff
@@ -50,161 +71,244 @@ class Ametek7270(Instrument):
...
             adapter,
             name,
-            **kwargs
-        )
+            read_termination=read_termination,
+            write_termination=write_termination,
+            **kwargs)
+
+    def check_errors(self):
+        """mandatory to be used for property setter
```
bmoneke, line 244: As it is only for the setter, it should be `check_set_errors` (since #883 ):
```suggestion
    def check_set_errors(self):
        """mandatory to be used for property setter
```

pymeasure/instruments/ametek/ametek7270.py
bmoneke: What do you think about `extended_identification` to state, that it is more than the mere identification, as the serial number is already the identification of the device.

tests/instruments/ametek/test_ametek7270.py
```diff
@@ -0,0 +1,508 @@
...
+            [],
+    ):
+        pass  # Verify the expected communication.
+
+
+@pytest.mark.parametrize("comm_pairs, value", (
+    ([(b'SEN', b'27\n')],
+     1.0),
+    ([(b'SEN', b'27\n')],
+     1.0),
```
bmoneke, line 19: What's the difference between those two cases?

tests/instruments/ametek/test_ametek7270.py
```diff
@@ -0,0 +1,508 @@
...
+    ([(b'REFN', b'7\n')],
+     7.0),
+    ([(b'REFN', b'7\n')],
+     7.0),
+    ([(b'REFN', b'7\n')],
+     7.0),
+    ([(b'REFN', b'7\n')],
+     7.0),
+    ([(b'REFN', b'7\n')],
+     7.0),
```
bmoneke, line 111: Why do you run the same test six times? What do you expect to change between these tests?

tests/instruments/ametek/test_ametek7270.py
```diff
@@ -0,0 +1,508 @@
...
+            [(b'VMODE 1', b'')],
+    ) as inst:
+        assert inst.set_channel_A_mode() is None
+
+
+def test_identification():
+    with expected_protocol(
+            Ametek7270,
+            [(b'ID', b'7270\n'),
+             (b'VER', b'2.11\n')],
```
bmoneke, line 498: Ah, now I understand: `id` does not return the serial number but the device type.

bmoneke (approved):
A few comments with only little changes, but it looks good in overall.

tests/instruments/ametek/test_ametek7270.py
seb5g: well it was autogenerated from your auto code...I'll change that

tests/instruments/ametek/test_ametek7270.py
seb5g: well it was autogenerated from your auto code...I'll change that

seb5g:
> > One side question for you guys maintaining Pymeasure: I'm the main author/developer of PyMoDAQ (pymodaq.cnrs.fr) that among other things has the same purpose as pymeasure (but focused on proposing highly modular GUI for data recording as a function of variable parameters). I think it would be highly valuable for both our projects to be more aware of what each package do. For this I'd like to invite one of you on the PyMoDAQ's days, a workshop I'm organising next 16/17 october in Paris to present features of pymeasure and discuss of what could be done at the interface of our packages.
> > Are you interested?
> 
> Oh, interesting project, thanks for telling about that module. It's always good to connect and learn from each other. Thanks for the invitation. I made my fellow maintainers aware of your offer, let's see, if we can manage an attendance.
> 
> By the way, we are developing a communication protocol [LECO](https://github.com/pymeasure/leco-protocol) for connecting instrument drivers and data control/acquisition among different processes and computers. Maybe our thought can help you with your remote control option and maybe you can give some feedback into our design process.

do you think you could reach me by email to discuss about this? My work email address is sebastien.weber@cemes.fr

tests/instruments/ametek/test_ametek7270.py
bmoneke: if you execute one property several times, it accumulates it. The idea is, that you may test different values.

tests/instruments/ametek/test_ametek7270.py
bmoneke: It's not yet in a pymeasure release. Therefore, I think it is acceptable, that the instrument changes a little in behaviour. If someone uses this instrument, they have it in another place or they are on a different branch. So they know, when they switch to a pymeasure release and should do that change consciously.

I feel bad about introducing a deprecated feature into a first release.

seb5g:
I just solved the conflict of authors, so you should be good to go with the merge!

tests/instruments/ametek/test_ametek7270.py
seb5g: Yeah it's fine, people know things are evolving!

tests/instruments/ametek/test_ametek7270.py
```diff
@@ -0,0 +1,480 @@
...
+        assert inst.set_channel_A_mode() is None
+
+
+def test_id():
+    with expected_protocol(
+            Ametek7270,
+            [(b'ID', b'7270\n'),
+             (b'VER', b'2.11\n')],
+    ) as inst:
+        assert inst.identification() == '7270/2.11'
```
bmoneke, line 472: You forgot to change that:
```suggestion
        assert inst.id == '7270/2.11'
```

bmoneke (approved):
It's good except for a failing test, due to not changing identification to id in the test itself.

tests/instruments/ametek/test_ametek7270.py
seb5g: indeed, you have a keen eye! That's done, thanks for the extended review!

bmoneke:
Thanks for your contribution.

PR #899, opened 2023-04-22
NanoVNA

jwbrooks0:
I'm trying to add a USB NanoVNA unit.  Similar to the units shown here: https://www.amazon.com/s?k=nanovna

I found this driver to be difficult in many regards, and I therefore have a number of questions.

1. I found that I had to specify that the adapter was a SerialAdapter instead of just passing the USB port address.  
2. The device has a unique property in that it always echos back any query before transmitting the response.  I wasn't sure how to handled this so I crated an inherited class of SerialAdapter that could strip out the echo from the response.  Not sure if this is a good way of doing this.
3. Related to the previous point.  The _read_echo() function in the modified SerialAdapter class first strips the echo from the reply.  The remaining reply often contains multiple responses.  I don't know if its best to return the responses as a list (there is a .strip() command somewhere that throws an error and effectively prevents me from doing this).  I've settled on joining all the responses back together with a comma between.  Again, not sure if this is the best practice.  
4. I created several Instrument.measurements().  Specially, I created frequencies and data.  These measurements are returned as strings (or possibly lists, see point 3) but should be converted to numpy arrays.  However, I'm not seeing an obvious way to post-process the reply in order to convert the answers.  Note that the frequencies measurement are all floats.  But the data measurements comes in as a somewhat complicated pairing of complex numbers.  
5. Because I couldn't figure out how to postprocess point 4, I ended up creating class functions to do the processing: get_frequencies(), get_S11_data(), etc.  I recognize that this is likely not ideal, but I don't know else how to do this.  
6. Tests.  There isn't really much to test.  I added a test for the sweep control.  Please let me know if I need to do more.

### DIFF
--- a/pymeasure/instruments/nanovna/nanovna.py
+++ b/pymeasure/instruments/nanovna/nanovna.py
@@ -1,0 +1,185 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2023 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from numpy import array
+from pymeasure.adapters import SerialAdapter
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set
+
+
+class SerialAdapterWithEcho(SerialAdapter):
+
+    echo_termination = '\r\n'
+    # write_termination = '\r'
+    # read_termination = '\r\nch> '
+
+    def _read_echo(self, **kwargs):
+        """Read function but includes scrubbing the echo from the reply.
+
+        :param \\**kwargs: Keyword arguments for the connection itself.
+        :returns str: ASCII response of the instrument with echo removed.
+        """
+        read = self._read_bytes(-1, break_on_termchar=True, **kwargs).decode()
+        read = read.replace('ch> ', "").split(self.echo_termination)
+        # echo = read[0]
+        reply = read[1:-1]
+
+        return ",".join(reply)  # TODO maybe don't return as a single string???
+
+
+class NanoVNA(Instrument):
+    """ Represents the NanoVNA interface for interacting with the instrument.
+
+    .. code-block:: python
+
+    """
+
+    def __init__(self, adapter, name="NanoVNA", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            includeSCPI=False,
+            read_termination='\r\nch> ',
+            write_termination='\r',
+            **kwargs
+        )
+
+    def read(self):
+        """
+        Reads from the instrument including the correct termination characters
+        """
+        ret = self.adapter._read_echo()
+        return ret
+
+    def write(self, command):
+        """
+        Writes to the instrument including the device address
+
+        :param command: command string to be sent to the instrument
+        """
+        super().write(command)
+
+    def ask(self, command, query_delay=0):
+        """Write a command to the instrument and return the read response.
+
+        :param command: Command string to be sent to the instrument.
+        :param query_delay: Delay between writing and reading in seconds.
+        :returns: String returned by the device without read_termination.
+        """
+        self.write(command)
+        self.wait_for(query_delay)
+        return self.adapter._read_echo()
+
+    help = Instrument.measurement(
+        'help',
+        """Returns list of available commands. """)
+
+    info = Instrument.measurement(
+        'info',
+        """Returns device information. """)
+
+    pause = Instrument.measurement(
+        'pause',
+        """Pauses acquisition. Returns nothing. """)
+
+    resume = Instrument.measurement(
+        'resume',
+        """Resume acquisition.  Returns nothing. """)
+
+    frequencies = Instrument.measurement("frequencies",
+                                         docs=""" Returns frequencies as a list
+                                         of floats. """,
+                                         # get_process=_process_frequencies,
+                                         )
+
+    data = Instrument.measurement("data",
+                                  docs=""" Returns a list of strings containing
+                                  the complex numbers. """,
+                                  )
+
+    trace = Instrument.measurement("trace",
+                                   docs=""" Gets the trace settings. """,
+                                   # TODO - make this into a control,
+                                   )
+
+    power = Instrument.setting('power %d',
+                               """ Sets the output power. """,
+                               validator=strict_discrete_set,
+                               values=[-1, 0, 1, 2, 3],
+                               )
+
+    sweep = Instrument.control('sweep',
+                               'sweep %s',
+                               """Sets the sweep details.  Input is a string
+                               containing 3 integers separated by spaces.
+                               First int is the start frequency (in Hz).
+                               Second int is the strop frequency (in Hz).
+                               Third int is the number of points in the sweep.
+                               Example:  '13000000 16000000 101'  """,
+                               )
+
+    def _process_frequencies(self, freqs):
+        return array(freqs.split(","), dtype=float)
+
+    def get_frequencies(self):
+        return self._process_frequencies(self.ask("frequencies"))
+
+    def _process_data(self, data):
+        data = array(data.replace(" ", ",").split(","), dtype=float)
+        return data[::2] + data[1::2] * 1j
+
+    def get_S11_data(self):
+        return self._process_data(self.ask("data 0"))
+
+    def get_S21_data(self):
+        return self._process_data(self.ask("data 1"))
+
+    def get_cals(self):
+        cal_load = self._process_data(self.ask("data 2"))
+        cal_open = self._process_data(self.ask("data 3"))
+        cal_short = self._process_data(self.ask("data 4"))
+        cal_thru = self._process_data(self.ask("data 5"))
+        cal_isoln = self._process_data(self.ask("data 6"))
+        return {"load": cal_load, "open": cal_open, "short": cal_short,
+                "thru": cal_thru, "isoln": cal_isoln}
+
+
+if __name__ == "__main__":
+    _cr = '\r'
+    _lf = '\n'
+    _crlf = _cr + _lf
+    _prompt = 'ch> '
+    port = '/dev/ttyACM0'
+    adapter = SerialAdapterWithEcho(port,
+                                    timeout=1,
+                                    write_termination=_cr,
+                                    read_termination=_crlf + _prompt,
+                                    )
+    unit = nanovna(adapter)
+
+    print(unit.help)
+    print(unit.info)
+
+    frequencies = unit.get_frequencies()
+    S11_data = unit.get_S11_data()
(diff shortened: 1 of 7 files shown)


jwbrooks0:
I don't understand why the test is failing =(

bmoneke:
Thank you very much for that new instrument. I did not look at the code, yet.


> 1. I found that I had to specify that the adapter was a SerialAdapter instead of just passing the USB port address.

Did you specify the (virtual) COM port of that USB port? It should work if you give "ASRL9" (or "COM9" with the VISA backend) to the Instrument (or the VISAAdapter) for COM port 9.

> 2. The device has a unique property in that it always echos back any query before transmitting the response. I wasn't sure how to handled this so I crated an inherited class of SerialAdapter that could strip out the echo from the response. Not sure if this is a good way of doing this.

The Adapters should handle only the message over the wire and not any code modification (except term_chars, which are used for message handling). The proper place to implement this stripping of responses is the `read` method of your instrument.
That way you can give the communication (as the device receives/sends it) to the `expected_protocol` and you can verify, that your code does the correct stripping action.

> 3. Related to the previous point. The _read_echo() function in the modified SerialAdapter class first strips the echo from the reply. The remaining reply often contains multiple responses. I don't know if its best to return the responses as a list (there is a .strip() command somewhere that throws an error and effectively prevents me from doing this). I've settled on joining all the responses back together with a comma between. Again, not sure if this is the best practice.

Why does it return several responses at once? Yes, `read` (where you should implement that algorithm) should return a single string.


> 4. I created several Instrument.measurements(). Specially, I created frequencies and data. These measurements are returned as strings (or possibly lists, see point 3) but should be converted to numpy arrays. However, I'm not seeing an obvious way to post-process the reply in order to convert the answers. Note that the frequencies measurement are all floats. But the data measurements comes in as a somewhat complicated pairing of complex numbers.

You found the answer in #898 : You may use `get_process=some_processing_method,` as a parameter.

> 5. Because I couldn't figure out how to postprocess point 4, I ended up creating class functions to do the processing: get_frequencies(), get_S11_data(), etc. I recognize that this is likely not ideal, but I don't know else how to do this.

1. See point 4.
2. Sometimes our property creators (`control`, `measurement`, `setting`) do not fit the bill. In these cases you may use an ordinary [python property](https://docs.python.org/3.8/library/functions.html#property).

> 6. Tests. There isn't really much to test. I added a test for the sweep control. Please let me know if I need to do more.

Tests are good, especially if the instrument returns strange strings (as here). That way you can verify, that your implementation creates / interprets the communication right.
With tests you can also test the post processing (point 4), that the received string is returned as the expected list of floats/complex or numpy array.

> I don't understand why the test is failing =(

`expected_protocol` uses the ProtocolAdapter, which does not have the `read_echo` method you defined in your overridden Serial adapter. See the second point for a fix.

pymeasure/instruments/nanovna/nanovna.py
```diff
@@ -0,0 +1,165 @@
...
+
+    sweep = Instrument.control('sweep',
+                               'sweep %s',
+                               """Sets the sweep details.  Input is a string
+                               containing 3 integers separated by spaces.
+                               First int is the start frequency (in Hz).
+                               Second int is the strop frequency (in Hz).
+                               Third int is the number of points in the sweep.
+                               Example:  '13000000 16000000 101'  """,
+                               )
```
bmoneke, line 140: Why don't you accept a tuple of numbers instead of the string?
```suggestion
    sweep = Instrument.control('sweep',
                               'sweep %i %i %i',
                               """Set the sweep details.  Input is a tuple
                               containing 3 integers.
                               First int is the start frequency (in Hz).
                               Second int is the strop frequency (in Hz).
                               Third int is the number of points in the sweep.
                               Example:  :code:`13000000 16000000 101`.  """,
                               )
```
That way, a user might set `inst.sweep = 130000, 1600000, 101` with integers and not with a string.

By the way, a docstring should be in imperative form, so it is "Set" and not "Sets"

pymeasure/instruments/nanovna/nanovna.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def _process_frequencies(self, freqs):
+        return array(freqs.split(","), dtype=float)
+
+    def get_frequencies(self):
+        return self._process_frequencies(self.ask("frequencies"))
+
+    def _process_data(self, data):
+        data = array(data.replace(" ", ",").split(","), dtype=float)
```
bmoneke, line 149: why don't you split at spaces directly instead of turning spaces to commas and then to split at commas?
```suggestion
        data = array(data.split(), dtype=float)
```

N.B: `split` without any option splits at whitespaces (space, tab, newline, etc.).

pymeasure/instruments/nanovna/nanovna.py
jwbrooks0: Very clever.  Thanks!

pymeasure/instruments/nanovna/nanovna.py
jwbrooks0: Did not know that about split. Thanks!

jwbrooks0:
It's going to take me a while to work through all of this.  Thanks for the feedback.

pymeasure/instruments/nanovna/nanovna.py
jwbrooks0: Arg.  Turns out the order of operations matters so I had to put it back. Good suggestion though.

jwbrooks0:
Ok.  Just pushed updates that address points 1 to 5 above.  I'll deal with the tests (point  6) later.  Presently, the code is still buggy, but I need feedback to fix.  

Here are the three cases I'm testing:
```
if __name__ == "__main__":
    port = '/dev/ttyACM0'

    case = 1
    if case == 1:
        adapter = SerialAdapter(port)
        unit = NanoVNA(adapter)
        print(unit.info)
    elif case == 2:
        adapter = SerialAdapter(port, timeout=1)
        unit = NanoVNA(adapter)
        print(unit.info)
    elif case == 3:
        unit = NanoVNA(port)
        print(unit.info)
```

And here is the output from the 3 cases:

Case 1:  The code hangs and I use Ctrl+C to exit.
```
Python 3.8.10 (default, Mar 13 2023, 10:26:41) 
Type "copyright", "credits" or "license" for more information.

IPython 7.34.0 -- An enhanced Interactive Python.

runfile('/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna/nanovna.py', wdir='/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna')

  File "/home/john/Environments/python38/lib/python3.8/site-packages/spyder_kernels/py3compat.py", line 356, in compat_exec
    exec(code, globals, locals)

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna/nanovna.py", line 248, in <module>
    print(unit.info)

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/common_base.py", line 228, in __getattribute__
    return super().__getattribute__(name)

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/common_base.py", line 472, in fget
    vals = self.values(command_process(get_command),

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/common_base.py", line 323, in values
    results = self.ask(command, **kwargs).strip()

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna/nanovna.py", line 85, in ask
    return self.read()

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna/nanovna.py", line 60, in read
    read = self.adapter._read_bytes(-1, break_on_termchar=True).decode()

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/adapters/serial.py", line 110, in _read_bytes
    return self._read_bytes_until_timeout()

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/adapters/serial.py", line 121, in _read_bytes_until_timeout
    chunk = self.connection.read(chunk_size, **kwargs)

  File "/home/john/Environments/python38/lib/python3.8/site-packages/serial/serialposix.py", line 565, in read
    ready, _, _ = select.select([self.fd, self.pipe_abort_read_r], [], [], timeout.time_left())

KeyboardInterrupt
```



Case 2:  Works perfectly
```
Python 3.8.10 (default, Mar 13 2023, 10:26:41) 
Type "copyright", "credits" or "license" for more information.

IPython 7.34.0 -- An enhanced Interactive Python.

runfile('/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna/nanovna.py', wdir='/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna')
['Kernel:       4.0.0', 'Compiler:     GCC 8.3.1 20190703 (release) [gcc-8-branch revision 273027]', 'Architecture: ARMv6-M', 'Core Variant: Cortex-M0', 'Port Info:    Preemption through NMI', 'Platform:     STM32F072xB Entry Level Medium Density devices', 'Board:        NanoVNA-H', 'Build time:   Dec 26 2019 - 19:32:30']
```




Case 3: Throws the following error
```
Python 3.8.10 (default, Mar 13 2023, 10:26:41) 
Type "copyright", "credits" or "license" for more information.

IPython 7.34.0 -- An enhanced Interactive Python.

runfile('/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna/nanovna.py', wdir='/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna')

  File "/home/john/Environments/python38/lib/python3.8/site-packages/pyvisa_py/highlevel.py", line 156, in open
    parsed = rname.parse_resource_name(resource_name)

  File "/home/john/Environments/python38/lib/python3.8/site-packages/pyvisa/rname.py", line 252, in from_string
    raise InvalidResourceName(

InvalidResourceName: Could not parse /dev/ttyACM0: unknown interface type


During handling of the above exception, another exception occurred:

  File "/home/john/Environments/python38/lib/python3.8/site-packages/spyder_kernels/py3compat.py", line 356, in compat_exec
    exec(code, globals, locals)

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna/nanovna.py", line 254, in <module>
    unit = NanoVNA(port)

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/nanovna/nanovna.py", line 44, in __init__
    super().__init__(

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/instruments/instrument.py", line 80, in __init__
    adapter = VISAAdapter(adapter, **kwargs)

  File "/home/john/Documents/GitHub/pymeasure_master/pymeasure/adapters/visa.py", line 126, in __init__
    self.connection = self.manager.open_resource(

  File "/home/john/Environments/python38/lib/python3.8/site-packages/pyvisa/highlevel.py", line 3284, in open_resource
    res.open(access_mode, open_timeout)

  File "/home/john/Environments/python38/lib/python3.8/site-packages/pyvisa/resources/resource.py", line 282, in open
    self.session, status = self._resource_manager.open_bare_resource(

  File "/home/john/Environments/python38/lib/python3.8/site-packages/pyvisa/highlevel.py", line 3209, in open_bare_resource
    return self.visalib.open(self.session, resource_name, access_mode, open_timeout)

  File "/home/john/Environments/python38/lib/python3.8/site-packages/pyvisa_py/highlevel.py", line 160, in open
    self.handle_return_value(None, StatusCode.error_invalid_resource_name),

  File "/home/john/Environments/python38/lib/python3.8/site-packages/pyvisa/highlevel.py", line 251, in handle_return_value
    raise errors.VisaIOError(rv)

VisaIOError: VI_ERROR_INV_RSRC_NAME (-1073807342): Invalid resource reference specified. Parsing error.
```


Not sure what to make of these three cases:
- Why do I have to provide a short timeout for the code to work (case 1 vs case 2) with I use the Serial adapter?  
- Case 3 seems grabs VISA adapter instead of the Serial.  You mentioned previously that the VISA adapter should be able to do Serial? Is that really true??  If so, it's not natively working.  Maybe I'm confused?

jwbrooks0:
@LongnoseRob , Thanks for the recommendation.  I've looked through their code before, but I find it fairly opaque.   

Here is their serial file: https://github.com/NanoVNA-Saver/nanovna-saver/blob/main/src/NanoVNASaver/Hardware/Serial.py

- They use a very small timeout (0.05).  I'm still not sure why a timeout should be required to operate if the unit is functioning correctly.  
- They also use the command `assert interface_type in {"serial", "usb", "bt", "network"}`, presumably to force the serial comms into using one of these interfaces.  Is there something equivalent for pymeasure?  Meaning, is there a way to force my __init__() function into using the SerialAdapter instead of trying PyVSIA?

pymeasure/instruments/nanovna/nanovna.py
```diff
@@ -0,0 +1,255 @@
...
+
+        :param command: Command string to be sent to the instrument.
+        :param query_delay: Delay between writing and reading in seconds.
+        :returns: String returned by the device without read_termination.
+        """
+        self.write(command)
+        self.wait_for(query_delay)
+        return self.read()
+
+    help = Instrument.measurement(
```
neuschs, line 87: The necessary commands for requesting or controlling should be implemented. You therefore don't need a help regarding the lower layer device commands but just a documentation about your wrapper.

BenediktBurger:
Hey @jwbrooks0 do you intend to continue this PR?

BenediktBurger:
Closing due to lack of development.
It can be reopened at any time.

Issue #894, opened 2023-04-17
VisaAdapter.close should likely not close the pyvisa.ResourceManager


dkriegner:
I am attempting to write code in which an `Instrument` connection can be closed (the instrument potentially be powered off) and later reconnected to the still running code. For this purpose, I call `Instrument.adapter.close()` when I detect that the device is gone.
This, however, causes also other pyvisa connections in the same python session to fail since the `ResourceManager` is invalidated by `VisaAdaper.close` (line 141):

https://github.com/pymeasure/pymeasure/blob/fb08f797247fd9e63103cd394bbbcaaa436c6dfc/pymeasure/adapters/visa.py#L131-L143

Of course one could say this is a pyvisa bug since the `ResourceManager` in the different `VisaAdapter` instances should be independent (see upstream bug: pyvisa/pyvisa#742), but is this behavior in pymeasure intended? Since [`__del__` ](https://github.com/pymeasure/pymeasure/blob/fb08f797247fd9e63103cd394bbbcaaa436c6dfc/pymeasure/adapters/adapter.py#L65) of `Adapter` calls `close()` this means that no Instrument should be garbage collected as long as other instruments are still needed. This seems wrong to me.

For my purpose it seems sufficient to close the device connection which allows later to reopen it and keep other device connections open.

bmoneke:
The goal was, to clear the buffer, as we experienced strange behaviour, see [#839 ](https://github.com/pymeasure/pymeasure/pull/839#discussion_r1087137595).

bilderbuchi:
The relevant upstream issue is https://github.com/pyvisa/pyvisa-sim/issues/68 although [this comment](https://github.com/pyvisa/pyvisa-sim/issues/68#issuecomment-1404191801) suggests that it (or another issue) should still be investigated.

dkriegner:
Do I understand it correctly that the `manager.close()` in `Adapter.close()` was added to work around a bug in pyvisa/pyvisa-sim? (This might have deserved a comment in the code?!)

bilderbuchi:
Yes. And yes, you're probably right.
We might even be able to make this conditional so the RM closing only happens when we are using the `@sim` library (if the RM object exposes that info).

dkriegner:
Yes making this conditional to the `@sim` library would make a lot of sense. `ResourceManager.visalib.library_path` should give some hint on the used library.

[...]

PR #900, opened 2023-04-27
VisaAdapter: close manager only when using pyvisa-sim

dkriegner:
As outlined in #894 closing the `VisaAdapter.manager` has the unwanted side effect of also closing all other pyvisa connections (even those of devices not related to this Visa resource). The reason for this is that the pyvisa ResourceManager shares the hardware session between all its instances (pyvisa/pyvisa#742). 

Closing the resource manager should therefore generally not be done when closing a device connection. It is however required when using pyvisa-sim due to a bug in this package (pyvisa/pyvisa-sim#68). 

I therefore propose these code changes which trigger the close only when using pyvisa-sim.

### DIFF
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -140,7 +140,7 @@ class VISAAdapter(Adapter):
         try:
             if self.manager.visalib.library_path == "unset":
                 # if using the pyvisa-sim library the manager has to be also closed.
-                # this works around pyvisa/pyvisa-sim#68
+                # this works around https://github.com/pyvisa/pyvisa-sim/issues/82
                 self.manager.close()
         except AttributeError:
             # AttributeError can occur during __del__ calling close
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -138,9 +138,13 @@ class VISAAdapter(Adapter):
         """
         super().close()
         try:
-            self.manager.close()
+            if self.manager.visalib.library_path == "unset":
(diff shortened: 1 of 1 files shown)


pymeasure/adapters/visa.py
```diff
@@ -138,9 +138,13 @@ def close(self):
         """
         super().close()
         try:
-            self.manager.close()
+            if self.manager.visalib.library_path == "unset":
```
bilderbuchi, line 141: Is that the "proper" way to detect pyvisa-sim? I guess there's no defined API for the visalib type?

bilderbuchi (approved):
LGTM, except for one small comment/question. 

I pinged the maintainer wr.t. his comment in the upstream issue -- maybe we'll get a better (open) issue to track, instead of this one, so let's wait a little.

pymeasure/adapters/visa.py
```diff
@@ -138,9 +138,13 @@ def close(self):
...
+                # if using the pyvisa-sim library the manager has to be also closed.
+                # this works around pyvisa/pyvisa-sim#68
```
bilderbuchi, line 143:
```suggestion
                # this works around https://github.com/pyvisa/pyvisa-sim/issues/82
```

pymeasure/adapters/visa.py
dkriegner: pyvisa-sim internal uses the same code. I could not identify something better.

dkriegner:
there seems to be no movement upstream. can this be merged?

bilderbuchi:
Yeah why not.

PR #901, opened 2023-04-27
ProtocolAdapter has list in signature

bmoneke:
Writing tests for the Test Generator, I noticed, that the ProtocolAdapter has an empty list in its signature, which caused problems.

This PR generates an empty list and dictionaries in the init/ other method, not in the signature.


bilderbuchi (approved):
Good call! Yes indeed mutable kwargs are a bad idea to have.

Issue #755, opened 2022-11-19
Better API documentation example for instruments


bmoneke:
I just stumbled upon a bad API call, which included all inherited members, such that CI complained.
Therefore I propose to change the line in the documentation from:

"Copy an existing instrument documentation file, which will automatically generate the documentation for the instrument"

to the code, which should be used.

bilderbuchi:
I did not understand [that error message](https://github.com/pymeasure/pymeasure/actions/runs/3503905662/jobs/5869253099), though -- a lot of other instruments use the same way of using `:inherited-members:`. Why did just that one fail?

bmoneke:
Neither did I understand why it happened then and not earlier.
Nevertheless, I think it is good to put our desired code in the documentation.
Do we want to have the inherited methods as well, or not?

bilderbuchi:
> Neither did I understand why it happened then and not earlier.

And also that there remain a lot of instances of what looks like identical usage, but works? I'm puzzled, too :shrug:

bmoneke:
> Yes, IMO, at least we should not forbid it. Think of the TEXIO (#698) - that documentation would be pretty empty without inherited members. 😅 Also, in the future if we get instrument classes defined, we will have a much deeper class hierarchy.

That makes sense, yes.

PR #903, opened 2023-05-02
Fix documentation inheriting Instrument and CommonBase documentation

mcdo0486:
A lot of instrument docs reStructuredText files included `:inherited-members:` to include properties and methods from a parent class, like an instrument base class.

Previously it was sufficient to block the fewer methods and properties inherited from the `Instrument` class with `:exclude-members:` from showing up in the documentation site.  With the addition of `CommonBase`, the docs sites for some instruments now inherit a lot of extraneous properties and methods from `Instrument` via `CommonBase`.

This PR goes through those instrument doc files with various fixes:
1. For instruments that inherit a base class, don't inherit Instrument class, therefore not including the extra code.
2. If not inheriting a class, remove `:inherited-members:` entirely.

Examples of some instruments.
https://pymeasure.readthedocs.io/en/latest/api/instruments/eurotest/eurotestHPP120256.html
https://pymeasure.readthedocs.io/en/latest/api/instruments/texio/texioPSW360L30.html

Include a few minor fixes to instrument docstrings.

At the moment KeithleyBuffer is not being included in the docs as a mixin.


mcdo0486:
Several Keithley instrument drivers use the generic `KeithleyBuffer` mixin as part of their class definition.

Previously the class definition for those instruments would be like `class Keithley2400(Instrument, KeithleyBuffer)`. 

To block the methods and properties from `Instrument` to show up in the docs, `:inherited-members: Instrument` needs to be added to the instruments reStructuredText file.

However the `:inherited-members:` option steps through the class inheritance method resolution order and returns once it finds that class. So after `Instrument` is encountered it returns, and it never reaches `KeithleyBuffer`. 

I've reordered the inheritance order of those instrument drivers so that `KeithleyBuffer` is included in the docs, while `Instrument` is not, for example  `class Keithley2400(KeithleyBuffer, Instrument)`.

Because `KeithleyBuffer` is a generic mixin without an `__init__()`, this doesn't have an effect on the driver's `super().__init__()` calls. That still gets passed to `Instrument.__init__()`.

bmoneke:
> Several Keithley instrument drivers use the generic `KeithleyBuffer` mixin as part of their class definition.
> 
> Previously the class definition for those instruments would be like `class Keithley2400(Instrument, KeithleyBuffer)`.
> 
> To block the methods and properties from `Instrument` to show up in the docs, `:inherited-members: Instrument` needs to be added to the instruments reStructuredText file.
> 
> However the `:inherited-members:` option steps through the class inheritance method resolution order and returns once it finds that class. So after `Instrument` is encountered it returns, and it never reaches `KeithleyBuffer`.
> 
> I've reordered the inheritance order of those instrument drivers so that `KeithleyBuffer` is included in the docs, while `Instrument` is not, for example `class Keithley2400(KeithleyBuffer, Instrument)`.
> 
> Because `KeithleyBuffer` is a generic mixin without an `__init__()`, this doesn't have an effect on the driver's `super().__init__()` calls. That still gets passed to `Instrument.__init__()`.

Thanks for that reordering. Putting `Instrument` further back is even better, as the most basic class should be the last one.
One thing you have to check: Is there any method or property defined in `Instrument` and in `KeithleyBuffer`?
Not only the __init__ would be overridden, but also other methods or properties with the same name. If there aren't any name clashes, reordering does not matter.

mcdo0486:
> Thanks for that reordering. Putting `Instrument` further back is even better, as the most basic class should be the last one. One thing you have to check: Is there any method or property defined in `Instrument` and in `KeithleyBuffer`? Not only the **init** would be overridden, but also other methods or properties with the same name. If there aren't any name clashes, reordering does not matter.

No, there are no other name clashes with `KeithleyBuffer` and inherited classes. All properties or methods in `KeithleyBuffer` include 'buffer' in the name, which doesn't conflict with properties or methods in `Instrument` or `CommonBase`.

bmoneke:
@bilderbuchi should we use this PR to define a policy how to write the API documentation (see #755)?
I'm unsure how to define our policy and whether these changes reflect that policy.

bilderbuchi:
Thank you for tidying this up, and thank you for adding your rationale for what you are doing to the description, that's quite useful!

I'm a bit uncomfortable to tweak the buffer inheritance order _just_ to keep the doc generator happy, this feels a bit like putting the cart before the horse. I think in this case it's fine and the impact should be limited, and I don't know a better way to avoid this (I feel there should be one, though). So, it's fine I think.

bilderbuchi:
It would probably be a good idea to come up with an updated policy, I think so far we mostly copied what usage was already there (since the "dark times" :D)

I haven't thought about what to include and what not. Let me try a first iteration to probe around a bit:

I think there is no question that we should include everything from a specific instrument class and its base classes before `Instrument` (e.g. `BlaBlaBase`).
I feel like what we should definitely include in the API reference from `Instrument` and `CommonBase` is what would be useful to a user of an instrument, so write/read/ask/values/..., and the SCPI properties, shutdown, etc.
What I'm less sure about is stuff that would only be interesting if you have to dig deeper (e.g. to extend an instrument): ChannelCreator, add_child, control/measurement/setting -- these things would typically just clutter the list of things that a user has available in a given class.
OTOH, this is the API _reference_, so I feel like it should be _complete_, and not arbitrarily leave stuff out. :shrug:

In any case, the docs generated for `Instrument` and `CommonBase` themselves should be complete and not leave stuff out.

Definitely loos thoughts and up for discussion.

bmoneke:
I agree, that all properties etc. should be included in the API description.

write, read etc. are mostly used by implementors and could be hidden (they can be reached via the completely described `Instrument` class).

Here again, a "SCPIbaseMixin" class would come handy to distinguish between SCPI propertis/methods and the base `Instrument`.

mcdo0486:
I think a discussion in the pymeasure forum would be a good place to hash some of this out. I'll have another PR about docs coming up, but that one will be more focused.

For instance, should there be mixins? `KeithleyBuffer` could easily have inherited `Instrument` and become the more typical "base class" implementation, and not caused the docs issue. `KeithleyBuffer` is the only class used in multiple class inheritance for drivers. For what it is worth, I think there mixins (which we can define re: pymeasure) should be allowed.

Again for what it is worth, a driver's doc file should only include it's own (and parent driver base class if it exists) documentation. `Instrument` documentation should just be in `Instrument`. As every driver says, it is a high level interface, so including docs for `binary_values` doesn't make sense.

bilderbuchi:
> Again for what it is worth, a driver's doc file should only include it's own (and parent driver "base class" if it exists) documentation. `Instrument` documentation should just be in `Instrument`. As most driver docstrings preface, it is a high level interface for that instruments functionality, so including docs for `write_binary_values` for every driver doesn't make sense.

However, there's SCPI-standard stuff like `id`, and things like `shutdown()`, which should imo *not* be filtered out just because they are in `Instrument`. Part of the issue is that `Instrument` and `CommonBase` include both universally-useful (e.g. for plain users) API and things mainly interesting for instrument implementors.
I'd rather err on the side of more info per instrument page, it's imo prefereable to skip some entries when scrolling/using the *API reference* docs, than having possibly crucial things hidden in the `Instrument` page only.

mcdo0486:
> > Again for what it is worth, a driver's doc file should only include it's own (and parent driver "base class" if it exists) documentation. `Instrument` documentation should just be in `Instrument`. As most driver docstrings preface, it is a high level interface for that instruments functionality, so including docs for `write_binary_values` for every driver doesn't make sense.
> 
> However, there's SCPI-standard stuff like `id`, and things like `shutdown()`, which should imo _not_ be filtered out just because they are in `Instrument`. Part of the issue is that `Instrument` and `CommonBase` include both universally-useful (e.g. for plain users) API and things mainly interesting for instrument implementors. I'd rather err on the side of more info per instrument page, it's imo prefereable to skip some entries when scrolling/using the _API reference_ docs, than having possibly crucial things hidden in the `Instrument` page only.

Ok, I've changed the inheritance blocking to `CommonBase` from `Instrument` so the SCPI commands are included so these instrument pages will show `id` property but will no longer show `ChannelCreator`. 

I'd agree include SCPI commands on the instrument docs page is useful for users, especially with commands like `reset()`. However including all the documentation of `Instrument` and `CommonBase` makes an instrument driver docs page awash with lots of duplicate text, and the commands that the driver class actually has programmed get lost in the mix.

What's useful about the other SCPI Mixin PR  https://github.com/pymeasure/pymeasure/pull/905 is that we'll be able to include those functions easily without including all the other baggage. For instance, we'd be able to do ` class Instrument(SCPIMixin, Instrument)`. to include documentation from `SCPIMixin` without too much baggage of `Instrument` and `CommonBase`.

docs/api/instruments/teledyne/teledyneT3AFG.rst
```diff
@@ -5,4 +5,4 @@ Teledyne T3AFG Arbitrary Waveform Generator
 .. autoclass:: pymeasure.instruments.teledyne.teledyneT3AFG
     :members:
     :show-inheritance:
-    :inherited-members:
+    :inherited-members: Instrument
```
bmoneke, line 8: Why is that still Instrument?
Did I miss anything?

docs/api/instruments/keithley/keithley2000.rst
```diff
@@ -5,5 +5,4 @@ Keithley 2000 Multimeter
 .. autoclass:: pymeasure.instruments.keithley.Keithley2000
     :members:
     :show-inheritance:
-    :inherited-members:
-    :exclude-members: ask, control, clear, measurement, read, setting, values, write
\ No newline at end of file
+    :inherited-members: Instrument
```
bmoneke, line 8: Here it is Instrument

docs/api/instruments/keithley/keithley2260B.rst
```diff
@@ -5,5 +5,4 @@ Keithley 2260B DC Power Supply
 .. autoclass:: pymeasure.instruments.keithley.Keithley2260B
     :members:
     :show-inheritance:
-    :inherited-members:
-    :exclude-members: ask, control, clear, measurement, read, setting, values, write
\ No newline at end of file
+    :inherited-members: Instrument
```
bmoneke, line 8: And here Instrument

bmoneke:
I looked through it. Only a few inconsistencies, but with that, I think, we can merge.

@bilderbuchi?

docs/api/instruments/teledyne/teledyneT3AFG.rst
mcdo0486: I think it would be easier if https://github.com/pymeasure/pymeasure/pull/941 is merged first because it fixes a few files, like this one. After that I can go through and update `:inherited-members:` to be `:inherited-members: CommonBase` throughout the .rst files.

bmoneke: Ok, let's do 941 first.

bmoneke:
Note: wait for #941 to be merged, before continuing.

docs/api/instruments/teledyne/teledyneT3AFG.rst
mcdo0486: Updated that doc

docs/api/instruments/keithley/keithley2260B.rst
mcdo0486: Updated that doc

docs/api/instruments/keithley/keithley2000.rst
mcdo0486: Updated that doc

Issue #571, opened 2022-01-16
Automatic docstring for Instrument properties


msmttchr:
PyMeasure comes with some convenience functions for making properties for `Instrument` sub-classes.
The aim of these properties is to provide basic instrument control and it has an associated docstring.
The docstring often include validity range for parameter or whether a property can be set or not.
All these information are available in the implementation, since the `validator` and `values` parameters specify the applicable range, so the docstring is in part a human readable version of the meaning of these parameters.

We could envisage two problems:

1) Duplication creates extra work
2) Inconsistency may happen if the code is changed but docstring isn't.

@CasperSchippers highlighed some problem of docstring incosistency in PR #376 and
@bilderbuchi hinted the possibility of enhancing the user docstring provided by users during property creation as part of https://github.com/pymeasure/pymeasure/pull/376#issuecomment-1012399933

It could be nice, if we can implement this to remove the burden to mention validity range or whether the property can be set or not as a part of the doc parameter. Also we could have a sort of standardization in providing this information to users.

Some challenge could be that validator may be complex to decode (eg. joined validators).

bmoneke:
As you have to write manually, what the property is about, you can write the validation as well.

The set/get/values can be added easily and reduce the amount of needed writing.

bilderbuchi:
A preliminary format that I have used when updating the documentation in #712 is appending some info to the first/summary line of a property docstring:
`(<type> [if strict validator: "strictly"] [if set: "in <values>", else if range: "from <value0> to <value1>"] [if dynamic: ", dynamic"])`.

It looks like this could be nicely automated.

bmoneke:
One question for discussion: Should we prepend the property type (measurement, setting, control), according to the coding style in the new documentation (measure, set, control) or should the user write it?

Pro: Users won't forget it and maintainers don't have to remind.
Con: User won't know it and the result won't match: "Set" + "Configure the turning rate" does not sound great.

It's an idea I want to raise, even though I'm slightly against it.

LongnoseRob:
> One question for discussion: Should we prepend the property type (measurement, setting, control), according to the coding style in the new documentation (measure, set, control) or should the user write it?

> One question for discussion: Should we prepend the property type (measurement, setting, control), according to the coding style in the new documentation (measure, set, control) or should the user write it?
> 
> Pro: Users won't forget it and maintainers don't have to remind. Con: User won't know it and the result won't match: "Set" + "Configure the turning rate" does not sound great.
> 
[...]

PR #907, opened 2023-05-09
Add valid values automatically to the end of a driver docstring.

mcdo0486:
Automatically add valid values to the end of a driver properties docstring.

Often the actual valid inputs for a property are not obvious from the docstring explanation. This can be for a few reasons, such as lack of detail or complex behavior of the property. Even if the property docstring is very detailed, having a dedicated location to see valid values is useful for a quick reference.

This PR creates a standard location at the end of the docstring and format to list out **Valid Values:**.

Examples from some properties:

**Valid Values:** 6, 12, 18, 24
**Valid Values in range:**  1 - 65535
**Valid Values:** `'internal'`, `'external rear'`, `'external front'`

At the moment one datatype that isn't formatted for valid values is `np.arange` values, for example this `values=np.arange(0, 65.1, 0.1)` assignment in [agilent4156.py](https://github.com/pymeasure/pymeasure/blob/ead6fe5033872a0e9243c2f270ea95b6fe2fded4/pymeasure/instruments/agilent/agilent4156.py#L197). Should that use case be included?

Additionally,  two drivers: [ametek7270.py ](https://github.com/pymeasure/pymeasure/blob/ead6fe5033872a0e9243c2f270ea95b6fe2fded4/pymeasure/instruments/ametek/ametek7270.py#L111) and [keithley2700.py ](https://github.com/pymeasure/pymeasure/blob/ead6fe5033872a0e9243c2f270ea95b6fe2fded4/pymeasure/instruments/keithley/keithley2700.py#L108) used the `values=list(range())`. These were modified to `values=range()` which does the same function without creating a long list object.

### DIFF
--- a/pymeasure/instruments/common_base.py
+++ b/pymeasure/instruments/common_base.py
@@ -530,8 +530,45 @@ class CommonBase:
         # Add the specified document string to the getter
         fget.__doc__ = docs
 
+        # If dynamic, add (dynamic) to the docstring
         if dynamic:
             fget.__doc__ += "(dynamic)"
+
+        def format_docstring_value(value):
+            if isinstance(value, str):
+                return "``'" + value + "'``"
+            elif isinstance(value, bool):
+                return "``" + str(value) + "``"
+            else:
+                return str(value)
+
+        if values != ():
+            valid_values = "\n\n        "
+            if isinstance(values, (list, tuple, dict)):
+                # If the validator has "range" in the function name
+                # format values as a range string: "value - value"
+                if isinstance(values, (list, tuple)) and 'range' in validator.__name__:
+                    valid_values += "**Valid Values in range:** "
+                    valid_values += format_docstring_value(values[0])
+                    valid_values += " - " + format_docstring_value(values[-1])
+                else:
+                    valid_values += "**Valid Values:** "
+                    valid_values += ", ".join(
+                        [format_docstring_value(i) for i in values]).strip()
+            elif isinstance(values, range):
+                valid_values += "**Valid Values in range:** "
+                valid_values += format_docstring_value(values.start)
+                valid_values += " - " + format_docstring_value(values.stop)
+            elif isinstance(values, (int, float)):
+                valid_values += "**Valid Value:** "
+                valid_values += format_docstring_value(values)
+            else:
+                # If values is another type, like np.arange, don't add values into docstring
+                valid_values = ""
+
+            fget.__doc__ += valid_values
+
+        if dynamic:
             return DynamicProperty(fget=fget, fset=fset,
                                    fget_params_list=CommonBase._fget_params_list,
                                    fset_params_list=CommonBase._fset_params_list,
--- a/pymeasure/instruments/ametek/ametek7270.py
+++ b/pymeasure/instruments/ametek/ametek7270.py
@@ -104,11 +104,12 @@ class Ametek7270(Instrument):
                                  )
     harmonic = Instrument.control(
         "REFN", "REFN %d",
-        """ An integer property that represents the reference
+        """An integer property that represents the reference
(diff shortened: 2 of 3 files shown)


tests/instruments/test_instrument.py
bmoneke, line 1: You test code from `common_base.py`, please move the tests to the corresponding `test_common_base.py`.

bmoneke:
One word of caution.
We have to determine, how we want to present the values information.
@bilderbuchi added a possible format in the "adding instruments" documentation.

tests/instruments/test_instrument.py
mcdo0486: Moved to `test_common_base.py`

pymeasure/instruments/common_base.py
```diff
@@ -530,8 +530,46 @@ def fset(self,
         # Add the specified document string to the getter
         fget.__doc__ = docs
 
+        # If dynamic, add (dynamic) to the docstring
```
bilderbuchi, line 533: In general, a good heuristic to follow is
> comment "why", not "what" -- the "what" is visible in the code that is commented.

In this case, the comment is just  plain-english rendition of the two line doc snippet, so does not give additional information beyond the code itself.

bilderbuchi:
> Modified values check to account for deprecated np.array elementwise comparison.

What is that about? is that deprecated in numpy, or in our code? If the former, and it keeps our code cleaner, we might just bump our min numpy version, instead?

bilderbuchi:
Thanks, this will be a useful addition!

> One word of caution. We have to determine, how we want to present the values information. @bilderbuchi added a possible format in the "adding instruments" documentation.

Indeed, we should first agree on a machine-readable format for this type/values information in #571 (we will need this for https://github.com/pymeasure/pyleco), and then implement this here.

pymeasure/instruments/common_base.py
mcdo0486: I agree. I felt compelled to put a placeholder comment since I split the dynamic if statement, but the comment doesn't add anything. Fixed in last commit.

mcdo0486:
> > Modified values check to account for deprecated np.array elementwise comparison.
> 
> What is that about? is that deprecated in numpy, or in our code? If the former, and it keeps our code cleaner, we might just bump our min numpy version, instead?

Deprecated in numpy. `np.arange(10) != ()` yields the following error: ` DeprecationWarning: elementwise comparison failed; this will raise an error in the future.`

[This is the most relevant info](https://github.com/numpy/numpy/issues/11521#issuecomment-525980680) I could find on when that would happen.

https://numpy.org/devdocs/release/1.10.0-notes.html#deprecations

mcdo0486:
> Thanks, this will be a useful addition!
> 
> > One word of caution. We have to determine, how we want to present the values information. @bilderbuchi added a possible format in the "adding instruments" documentation.
> 
> Indeed, we should first agree on a machine-readable format for this type/values information in #571 (we will need this for https://github.com/pymeasure/pyleco), and then implement this here.

I agree a standard format for pymeasure and pyleco makes sense. However this PR is a first step to make the valid values human-readable with a standard location and format, which acts like a quick reference when developing procedures. 

Would it be fine to merge this as a first step, then iterate on this solution with what comes up in pyleco?

bmoneke:
Pyleco won't define how the docstring has to look like. It is, however, necessary, that some information regarding a property are machine readable ( valid values, type etc).
Therefore, we can determine a certain format here in pymeasure ( probably better in an issue first) ourselves.

So, the idea is, to come up with a human and machine readable docstring format.

bilderbuchi:
> Would it be fine to merge this as a first step, then iterate on this solution with what comes up in pyleco?

Please let's design this first (in #571):

> So, the idea is, to come up with a human and machine readable docstring format.

PR #908, opened 2023-05-10
New instrument: Tektronix371A

sansanda:
I'd like to add an instrument. The [Tektronix 371A Curve Tracer](https://www.tek.com/en/datasheet/370b-371b-curve-tracers).
The communication is implemented according to the manufacturer's documentation and tested with the instrument. 
You can see the complete user manual [here](https://w140.com/tekwiki/images/9/91/070-8043-00.pdf).

### DIFF
--- a/pymeasure/instruments/tektronix/tek371A.py
+++ b/pymeasure/instruments/tektronix/tek371A.py
@@ -1,0 +1,837 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2022 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NON INFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+import logging
+import time
+from pyvisa import constants as pyvisa_constants
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_range, truncated_discrete_set
+from pymeasure.instruments.validators import strict_discrete_set
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class WaveformPreamble:
+    """ Represents the waveform preamble for a curve data from the Tektronix model 371A.
+    The preamble contains the information needed for interpreting, scaling,
+    and labeling the numeric information of the curve.
+    This preamble is coded in ASCII characters and is readable by the operator without interpretation by the controller.
+    """
+
+    # The preamble and curve are each a string of eight-bit bytes.
+    # The preamble is a string of ASCII letters, numerals, and punctuation.
+    # Each character is represented by one byte.
+
+    # Preambles are necessary to interpret the numeric information in the curve data that follows them.
+    # Within a preamble, 26 parameters are specified.
+    # The first ten are unique to the 371B curve tracer and are included as a sub-string linked to the WFID: label.
+    # The other 16 parameters include ten that have fixed values and six that vary with the particular data sent.
+    # Within the WFID: sub-string the parameters are separated by slashes,
+    # while the entire sub-string is delimited by a pair of double quote marks.
+    # Most of the WFID: string is rather strictly defined,
+    # with each parameter value being right justified in a fixed length field.
+    # An exception is the BGM value, which may vary in field length.
+    # The remainder of the preamble uses standard punctuation.
+    # A colon links each parameter label with its corresponding value and the individual label and value pairs
+    # are separated with commas.
+
+    # A complete preamble might look like this:
+    # WFMPRE WFID:”INDEX 3/VERT 500MA/ HORIZ 1V/STEP 5V/OFFSET 0.00V/BGM 100mS/VCS 12.3/TEXT /HSNS VCE”,
+    # ENCDG:BIN, NR.PT:3,PT.FMT:XY,XMULT:+1.0E–2,XZERO:0,XOFF: 12,XUNIT:V,YMULT:+5.0E–3,YZERO:0,YOFF:12,
+    # YUNIT:A,BYT/ NR:2,BN.FMT:RP,BIT/NR:10,CRVCHK:CHKSMO,LN.FMT:DOT
+
+    def __init__(self, preamble_array):
+
+        wf_id = preamble_array[0].split("/")
+        self.n_measures_readed = int(preamble_array[2].split(":")[1].replace(" ", ""))
+        self.x_scale_factor = float(preamble_array[4].split(":")[1].replace(" ", ""))
+        self.n_horizontal_resolution_points = 1024  # screen points, like horizontal pixels in a tv screen
+        self.horizontal_range = self.x_scale_factor * self.n_horizontal_resolution_points
+        self.y_scale_factor = float(preamble_array[8].split(":")[1].replace(" ", ""))
+        self.n_vertical_resolution_points = 1024
+        self.vertical_range = self.y_scale_factor * self.n_vertical_resolution_points
+        self.horizontal_offset = int(preamble_array[6].split(":")[1].replace(" ", ""))
+        self.vertical_offset = int(preamble_array[10].split(":")[1].replace(" ", ""))
+        self.horizontal_units = str(preamble_array[7].split(":")[1].replace(" ", ""))
+        self.vertical_units = str(preamble_array[11].split(":")[1].replace(" ", ""))
+
+        self.step_size = wf_id[3].removeprefix("STEP").lstrip(" ")
+        if 'mV' in self.step_size:
+            self.step_size = float(self.step_size.removesuffix("mV")) * 1E-3
+        elif 'V' in self.step_size:
+            self.step_size = float(self.step_size.removesuffix("V"))
+
+        self.step_offset = wf_id[4].removeprefix("OFFSET").lstrip(" ")
+        if 'mV' in self.step_offset:
+            self.step_offset = float(self.step_offset.removesuffix("mV")) * 1E-3
+        elif 'V' in self.step_offset:
+            self.step_offset = float(self.step_offset.removesuffix("V"))
+
+    def __str__(self):
+
+        description = f'The Waveform Preamble is: \n' \
+                      f'Number of Measures Readed from: {self.n_measures_readed} \n' \
+                      f'Horizontal X Scale Factor: {self.x_scale_factor} {self.horizontal_units} \n' \
+                      f'Number Horizontal Range: {self.horizontal_range} \n' \
+                      f'Number Horizontal Resolution Points: {self.n_horizontal_resolution_points} \n' \
+                      f'Horizontal Offset: {self.horizontal_offset} \n' \
+                      f'Horizontal Units: {self.horizontal_units} \n' \
+                      f'Vertical Y Scale Factor: {self.y_scale_factor} {self.vertical_units} \n' \
+                      f'Number Vertical Range: {self.vertical_range} \n' \
+                      f'Number Vertical Resolution Points: {self.n_vertical_resolution_points} \n' \
+                      f'Vertical Offset: {self.vertical_offset} \n' \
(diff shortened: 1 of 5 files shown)


pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+    # sub-string is delimited by a pair of double quote marks. Most of the WFID: string is rather
+    # strictly defined, with each parameter value being right justified in a fixed length field.
+    # An exception is the BGM value, which may vary in field length. The remainder of the
+    # preamble uses standard punctuation. A colon links each parameter label with its
+    # corresponding value and the individual label and value pairs are separated with commas.
+
+    # A complete preamble might look like this: WFMPRE WFID:”INDEX 3/VERT 500MA/ HORIZ 1V/STEP
+    # 5V/OFFSET 0.00V/BGM 100mS/VCS 12.3/TEXT /HSNS VCE”, ENCDG:BIN, NR.PT:3,PT.FMT:XY,
+    # XMULT:+1.0E–2,XZERO:0,XOFF: 12,XUNIT:V,YMULT:+5.0E–3,YZERO:0,YOFF:12, YUNIT:A,BYT/ NR:2,
+    # BN.FMT:RP,BIT/NR:10,CRVCHK:CHKSMO,LN.FMT:DOT
```
bmoneke, line 62: why don't you include this in the device docstring as well?
That way the API contains also this relevant information.

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+    # curve is represented by four bytes, 2 for the 10 bits of the X coordinate and 2 for the 10
+    # bits of the Y coordinate: XXYYXXYY . . . XXYY And finally there is one byte which is the
+    # checksum for the preceding 4098 data bytes.
+
+    # The first XXYY of the 1024 points is the first point taken whe we execute the sweep.
+    # Last point should be the coordinates 0,0.
+    # XXYY is coordinate point. Thus, we have to translate it for getting electrical values.
+
+    def __init__(self,
+                 waveform_preamble: WaveformPreamble,
```
bmoneke, line 152: In pymeasure, we do not use type hinting in this form.
Instead, we add define the parameters in the (class) docstring, where it is possible to state the type as well:
~~~
:param waveform_preample: Some description.
:type waveform_preample: WaveformPreamble
~~~

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+    COLLECTOR_SUPLLY_RANGE = [0.0, 100.0]  # in % of max peak power
+    CURSOR_DOT_SET = list(range(0, 1025, 1))  # 0 is the beginning and 1024 is the end of the curve
+    HORIZONTAL_SOURCE_SET = ["STPgen", "STP", "COLlect", "COL"]
+    VERTICAL_SOURCE_SET = ["COLlect", "COL"]
+    STEP_GENERATOR_SOURCE_SET = ["VOLtage", "CURrent", "VOL", "CUR"]
+    DISPLAY_VALID_STORE_MODES = ["NST", "STO", "NSTore", "STOre"]
+    VALID_MEASUREMENT_MODES = ["REPeat", "REP", "SINgle", "SIN", "SWEep", "SWE", "SSWeep", "SSW"]
+    VALID_CURSOR_MODES = ["OFF", "DOT", "LINE", "WINDOW"]
+    VALID_V_H_COORDINATES_SET = list(range(0, 1025, 1))  # 0 is the beginning
+    # and 1023 is the end of the valid coordinates set
```
bmoneke, line 251: if 1023 is the last valid value, you should use `range(1024)`, which is equal to `range(0, 1024, 1)` as the first 0 and the last 1 are default values.
```suggestion
    VALID_V_H_COORDINATES_SET = list(range(1024))  # 0 is the beginning
    # and 1023 is the end of the valid coordinates set
```

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+        super().__init__(
+            adapter,
+            "Tektronix Curve Tracer model 371A",
+            write_termination="\n",
+            read_termination="",
+            send_end=True,
+            includeSCPI=False,
+            timeout=timeout,
+            **kwargs
+        )
```
bmoneke, line 332: I'm not sure, why the test passes, but the name should be changeable:
```suggestion
    def __init__(self,
                 adapter,
                 name="Tektronix Curve Tracer model 371A"
                 query_delay=0.1,
                 write_delay=0.4,
                 timeout=5000,
                 **kwargs):
        super().__init__(
            adapter,
            name,
            write_termination="\n",
            read_termination="",
            send_end=True,
            includeSCPI=False,
            timeout=timeout,
            **kwargs
        )
```

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+        self.last_write_timestamp = 0.0
+        self.srq_called = False
+
+    # ############################################
+    # # Tektronix Curve Tracer 371A SCPI commands.
+    # ############################################
+
+    id = Instrument.measurement(
+        "ID?",
+        """Return the identification of the instrument (string) """,
```
bmoneke, line 345: For a `measurement`, the docstring should start with `"Measure"` or `"Set"`, not with "Return".
There is a test in the current master, which will fail, unless you comply with this norm.

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+        get_process=lambda r:
+        float(r[0].replace(" ", "").replace("READOUT", ""))
+    )
+
+    crt_readout_v = Instrument.measurement(
+        "REAdout? SCientific",
+        """Return the vertical and horizontal cursor parameters.""",
+        get_process=lambda r:
+        float(r[1])
+    )
```
bmoneke, line 416: The docstrings do not match the content!

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+        """
+        stepgen_step_source_and_size = self.stepgen_step_source_and_size
+        stepgen_step_size = stepgen_step_source_and_size[1]
+        stepgen_step_source = stepgen_step_source_and_size[0]
+        max_offset = 5.0 * stepgen_step_size
+
+        if abs(offset) > max_offset:
+            offset_units = "V" if "VOLTAGE" in stepgen_step_source else "A"
+            raise ValueError(
+                f'Setting the step generator offset....\n'
```
bmoneke, line 657: You can turn it into a plain python property:
~~~python
@property
def stepgen_offset(self):
    """Control the step generator offset."""  # docstring for the property
    return some_value  # getter part

@stepgen_offset.setter
def stepgen_offset(self, value):
    pass  # setter part
~~~

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+        get_process=lambda r:
+        r
+    )
+
+    def curve(self, bytes_count):
+        """
+        Asks the instrument for the curve data for the view curve when in view mode
+        or the curve data for the current display when in store mode.
+        :param bytes_count: The number of bytes to get (read) from the instrument.
+        :return: The binary data readed with the next structure
```
bmoneke, line 712: The past form is (in written english) the same as the present
```suggestion
        :return: The binary data read with the next structure
```

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+
+    waveform_preamble = Instrument.control(
+        "WFMpre?", "WFMpre %s",
+        """Control the waveform preamble data for the currently displayed waveform (string).""",
+        # Response syntax
+        get_process=lambda r:
+        r
+    )
+
+    def curve(self, bytes_count):
```
bmoneke, line 707: If it is a method, it should have an "action" name, i.e. a verb, like "fetch_curve".

bmoneke (changes requested):
Thanks for your contribution.
Here are a few comments of a first round.

pymeasure/instruments/tektronix/tek371A.py
sansanda: got it. Many thanks

pymeasure/instruments/tektronix/tek371A.py
sansanda: You are right. In fact the A/D is 10 bits, so the scale is 0 - 1023. Many thanks

pymeasure/instruments/tektronix/tek371A.py
sansanda: Right. Thanks.

pymeasure/instruments/tektronix/tek371A.py
sansanda: Thanks Benedikt

pymeasure/instruments/tektronix/tek371A.py
sansanda: Solved. Thanks

pymeasure/instruments/tektronix/tek371A.py
sansanda: Right. Thanks

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+        # ['VOLTAGE',2.0]. Parameters out ouf range will have no effect.
+        get_process=lambda r:
+        [r[5].split(":")[0], float(r[5].split(":")[1])] if isinstance(r, list)
+        else r.split(":")[1]
+    )
+
+    def get_stepgen_offset(self):
+        """
+
+        :return:
```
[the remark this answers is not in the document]
sansanda: Done. Thanks

pymeasure/instruments/tektronix/tek371A.py
sansanda: Done. Thanks

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+        values=COLLECTOR_SUPPLY_POLARITY_MODES
+    )
+
+    cs_peakpower = Instrument.control(
+        "PKPower?", "PKPower %s",
+        """Control the collector supply peak power watts settings.""",
+        validator=strict_discrete_set,
+        values=COLLECTOR_SUPPLY_PEAKPOWER_VALUES,
+        get_process=lambda r:
+        float("".join(r.replace(" ", "")).replace("PKPOWER", ""))
```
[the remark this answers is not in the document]
sansanda: Yes, it works. The response syntax is VCSpply <data> where <data> = 0.0 or 0.1 ........ 99.9 or 100.0 I can improve the code changing it with the next: `        float(r[1]) if isinstance(r, list)
        else float("".join(r.replace(" ", "")).replace("PKPOWER", ""))`

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -0,0 +1,859 @@
...
+    )
+
+    cs_collector_supply = Instrument.control(
+        "VCSpply?", "VCSpply %f",
+        """Control the collector supply output level (from 0.0% to 100.0% in 0.1% increments).""",
```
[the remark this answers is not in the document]
sansanda: I have to give it as float with only one decimal and also inside the [0.0, 100.0] range.

pymeasure/instruments/tektronix/tek371A.py
sansanda: got it. Thanks

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -828,32 +716,340 @@ def wait_for_srq(self):
...
+        """
+        # Event code from the instrument has to be interpreted as follows:
+        #
+        # response --> 'EVENT code'
+
+        #  SYSTEM EVENT CODES
+        NO_ERROR = 0
+        POWER_ON = 401
+        MEASURE_SIGNAL_OPERATION_COMPLETE = 402
+        USER_REQUEST = 403
```
bmoneke, line 775: I think an IntEnum is more appropriate here.
IntFlag supposes, that you can have several flags at the same time and the total value is the sum of the individual values. A typical example are status bytes, where each bit signifies something different. A value of 5, for example, would indicate the first (value of 1) and the third flag (value of 4).

sansanda: Done. Thanks

pymeasure/instruments/tektronix/tek371A.py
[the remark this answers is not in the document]
sansanda: Finally I prefer to use the separator = " " option and maxsplit = 1. It becomes into a cleaner code. 
Many thanks @bmoneke

pymeasure/instruments/tektronix/tek371A.py
```diff
@@ -752,6 +752,23 @@ def initialize(self):
...
+        time.sleep(max(0, self.write_delay - actual_write_delay))
+        super().write(command, **kwargs)
+        self.last_write_timestamp = time.time()
+
+    def ask(self, command):
+        """ Overrides Instrument ask method for including query_delay time on parent call.
+        :param command: Command string to be sent to the instrument.
+        :returns: String returned by the device without read_termination.
+        """
+        return super().ask(command, self.query_delay)
```
bilderbuchi, line 770: Do you realise that you will wait `self.query_delay`+`self.write_delay`, as `ask` calls write in the background? Do you really need two different wait intervals, wouldn't the one for `write` be enough?
@bmoneke any feedback? Is there a better approach than this commit (84bdfd16d24d79da7583ae5d9f1621845d373a5d)?

bmoneke: I don't know what the instrument needs. However, in principle both delays can be necessary.
The write delay ensures a certain time between two consecutive write commands, while the query delay ensures some delay between writing and reading the answer.

From that, I do not see any issue in the logic.

However, I do see an issue in signature: if "instrument.ask" accepts a command and a "query_delay", the inheriting method should accept both parameters as well.

You can do "super().ask(command, query_delay or self.query_delay)" with a "query_delay = None" parameter.

PR #909, opened 2023-05-11
Find all instruments and channels for testing

### LINKED ISSUE #904: Check all Channel classes for docstrings
#895 added a property docstring test. It works, however, only for the `Instrument` classes which are publicly available.
Channels (and some base instruments), which are not imported in the init files, are not checked.

This issue is about collecting all `Instrument` and `Channel` subclasses in order to check them for docstring consistencies.

bmoneke:
Closes #904 

This PR improves the finding mechanism, such that all channels (and instrument base classes etc.) are found, even though they are not imported into the `instruments` namespace directly.

However, the module with the class definition has to be imported (i.e. `instruments.__init__.py` has to mention the manufacturer module and that module has to import the device module).

I found a few more grandfathered base classes and channels (regarding docstrings) and another instrument, which uses `Instrument` as a channel base class.
The large diff is regarding the expanded exceptions list.

### DIFF
--- a/pymeasure/instruments/__init__.py
+++ b/pymeasure/instruments/__init__.py
@@ -67,6 +67,7 @@ from . import signalrecovery
 from . import srs
 from . import tcpowerconversion
 from . import tektronix
+from . import teledyne
 from . import temptronic
 from . import texio
 from . import thermotron
--- a/tests/instruments/test_all_instruments.py
+++ b/tests/instruments/test_all_instruments.py
@@ -27,38 +27,53 @@ import pytest
 from unittest.mock import MagicMock
 
 from pymeasure import instruments
-from pymeasure.instruments import Instrument
+from pymeasure.instruments import Instrument, Channel
 
 
 # Collect all instruments
 devices = []
-for manufacturer in dir(instruments):
-    if manufacturer.startswith("__"):
-        continue
-    manu = getattr(instruments, manufacturer)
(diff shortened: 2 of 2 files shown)


bilderbuchi (approved):
LGTM, thanks!

mcdo0486:
I think this PR is great for adding more test coverage. Just adding a comment that hopefully in the future the testing will not rely on `__init__.py` for finding instruments, instead walking over the module directories instead to import instruments.

This discussion https://github.com/pymeasure/pymeasure/discussions/801 talked about moving away from adding instruments from `pymeasure.instruments.__init__.py`, because as more and more drivers get included, which may have their own dependencies, auto importing every instrument when your code touches `pymeasure.instruments` will create a log of undue memory use.

bmoneke:
@mcdo0486 do you have an idea, how to accomplish it?

mcdo0486:
> @mcdo0486 do you have an idea, how to accomplish it?

I did a quick PR on your branch [bmoneke:test_all_channels](https://github.com/bmoneke/pymeasure/tree/test_all_channels) but pretty much use `pathlib.Path` to loop through the folders in `instruments.__path__[0]`.

bmoneke:
@mcdo0486

I'll merge this PR and you may open your code as a PR against master.

The reason is, that this PR already helps to reduce maintenance effort.

Issue #910, opened 2023-05-12
Python 3.8 incompatibility


msmttchr:
I have just found that the line below gives an error with python 3.8.
https://github.com/pymeasure/pymeasure/blob/ead6fe5033872a0e9243c2f270ea95b6fe2fded4/pymeasure/display/widgets/plot_frame.py#L73

Possibly this dictionary 'oring' has been introduced in a subsequent version of python (3.9 is OK for example).

mcdo0486:
Yup, I can confirm that. I ran into that line with a user's anaconda pymeasure install. Should have filed an issue then.

PR #911, opened 2023-05-12
Replace union operator with unpacking

mcdo0486:
Fixes https://github.com/pymeasure/pymeasure/issues/910

Dict union operator was added in 3.9. Replacing the operator by unpacking `self.LABEL_STYLE` makes the statement compatible with python 3.8 and python 3.7.

### DIFF
--- a/pymeasure/display/widgets/plot_frame.py
+++ b/pymeasure/display/widgets/plot_frame.py
@@ -70,7 +70,7 @@ class PlotFrame(QtWidgets.QFrame):
 
         self.plot = self.plot_widget.getPlotItem()
 
-        style = dict(self.LABEL_STYLE | {'justify': 'right'})
+        style = dict({'justify': 'right'}, **self.LABEL_STYLE)
         if "font-size" in style:  # LabelItem wants the size as 'size' rather than 'font-size'
             style["size"] = style.pop("font-size")
         self.coordinates = pg.LabelItem("", parent=self.plot, **style)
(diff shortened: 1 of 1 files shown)


pymeasure/display/widgets/plot_frame.py
```diff
@@ -70,7 +70,7 @@ def _setup_ui(self):
 
         self.plot = self.plot_widget.getPlotItem()
 
-        style = dict(self.LABEL_STYLE | {'justify': 'right'})
+        style = dict({'justify': 'right'}, **self.LABEL_STYLE)
```
CasperSchippers, line 73: This is a very minor thingy: changing the order causes values in `self.LABEL_STYLE` to take precedence over the new values, so maybe this would be a better order:
```suggestion
        style = dict(self.LABEL_STYLE, justify='right')
```

CasperSchippers (changes requested):
Thanks for the quick bugfix! I have only 1 comment

pymeasure/display/widgets/plot_frame.py
mcdo0486: Sounds good, fixed in last commit.

mcdo0486: @CasperSchippers the last runner seems to have borked, can you restart it?

msmttchr: I have re-run failed testa and now it's ok

msmttchr (approved):
Ok for me. Thanks

CasperSchippers (approved):
Thanks, will merge it right away

PR #918, opened 2023-05-19
Increase test_visa.py adapter timeout to 120 seconds to avoid timeout

mcdo0486:
for the `tests/adapter/test_visa.py` test, the simulated visa adapter has a timeout of 10 seconds.
https://github.com/pymeasure/pymeasure/blob/eab361ccfef345a6ffd347868ef68cfa25d21aaf/tests/adapters/test_visa.py#L44

Many CI runners are timing out with this test. I've tested changing the timeout to 60, 1, .2
https://github.com/mcdo0486/pymeasure/commit/dbc1ab778afeefaa8c7b1466a7ca9b5457c50964
https://github.com/mcdo0486/pymeasure/commit/d3eb8666fc6e11b4b5fa4999c4192069060c2cb4
https://github.com/mcdo0486/pymeasure/commit/1294d5af873f784bfc068f90bd0026343093624e

The 60 second timeout github action worked. The 1 second and .2 second both timed out with the same test. I think there is a lot of resource contention on the github runners that can slow the execution of workflows

Increasing the timeout to 120 seconds to account for that.

### DIFF
--- a/tests/adapters/test_visa.py
+++ b/tests/adapters/test_visa.py
@@ -41,7 +41,7 @@ if not is_pyvisa_sim_installed:
 def adapter():
     adapter = VISAAdapter(SIM_RESOURCE, visa_library='@sim',
                           read_termination="\n",
-                          timeout=10,
+                          timeout=120,
                           )
     yield adapter
     # Empty the read buffer, as something might remain there after a test.
--- a/tests/adapters/test_visa.py
+++ b/tests/adapters/test_visa.py
@@ -41,7 +41,8 @@ if not is_pyvisa_sim_installed:
 def adapter():
     adapter = VISAAdapter(SIM_RESOURCE, visa_library='@sim',
                           read_termination="\n",
-                          timeout=120,
+                          # Large timeout allows very slow GitHub action runners to complete.
+                          timeout=60,
                           )
     yield adapter
     # Empty the read buffer, as something might remain there after a test.


bilderbuchi:
Interesting! I hadnt noticed that prpblem - could you point to recent CI failures that were caused by this?
I'm a bit worried about needlessly inflating our CI time. Could the timeout be masking another issue entirely? I am quite surprised that an enormous 10s timeout needs to be increased even more. I have to take a closer look once I'm not on a phone...

mcdo0486:
> Interesting! I hadnt noticed that prpblem - could you point to recent CI failures that were caused by this? I'm a bit worried about needlessly inflating our CI time. Could the timeout be masking another issue entirely? I am quite surprised that an enormous 10s timeout needs to be increased even more. I have to take a closer look once I'm not on a phone...

Here are some example failures:
https://github.com/pymeasure/pymeasure/actions/runs/5025543531/jobs/9012663563
https://github.com/pymeasure/pymeasure/actions/runs/5011378094/jobs/8982142979
https://github.com/pymeasure/pymeasure/actions/runs/4747774758/jobs/8433123218

Increasing the timeout shouldn't make the CI time take longer, it should allow a wider buffer for the test to complete. I think the resource contention or process scheduling on the windows CI runners means the test runs slow, so when it compares `start` and `time.time()` it is past the timeout.

bilderbuchi:
Thanks for collecting those logs. That is quite weird. It does look like a resource problem, indeed.
Your test with 60s seems to work, why double that, then? Maybe a much smaller increase (doubling to 20?) would already have been sufficient?

Finally, in any case, please add an explanation comment to that timeout value. A timeout value that is 3 times the runtime of our whole test suite surely would raise eyebrows otherwise.

bilderbuchi:
Thanks for the update, let's go with this.

PR #919, opened 2023-05-19
Remove auto-importing all instruments in pymeasure/instruments/__init__.py

mcdo0486:
Currently over 90 instruments are auto imported in [pymeasure/instruments/__init__.py](https://github.com/pymeasure/pymeasure/blob/eab361ccfef345a6ffd347868ef68cfa25d21aaf/pymeasure/instruments/__init__.py)

That means if you import an instrument directly, such as `from pymeasure.instruments.hp.hp3437A import HP3437A`, the transversal over `pymeasure.instruments` will import every other instrument as well.

Removing instruments from `pymeasure.instruments.__init__.py` should make pymeasure efficient now and prevent performance problems in the future, especially on resource constrained computers.

Previous discussion on topic: https://github.com/pymeasure/pymeasure/discussions/801


pymeasure/instruments/instrument.py
```diff
@@ -26,7 +26,7 @@
 import time
 
 from .common_base import CommonBase
-from ..adapters import VISAAdapter
+from ..adapters.visa import VISAAdapter
```
bmoneke, line 29: That line is there, in case we did not load all adapters?

Anyways, as it imports explicitly instead of relying on the init.py, I'm fine with that change.

bmoneke (changes requested):
Thanks for the initiative.

The code changes are fine. 
Please add a changelog entry as it is a breaking change!

@bilderbuchi could you give your placet, that only "Channel" and "Instrument" are imported into "pymeasure.instruments".

pymeasure/instruments/instrument.py
mcdo0486: Yup, just a change to import explicitly. Maybe it can futureproof if there are changes to adapters init.py

mcdo0486:
> Thanks for the initiative.
> 
> The code changes are fine. Please add a changelog entry as it is a breaking change!
> 
> @bilderbuchi could you give your placet, that only "Channel" and "Instrument" are imported into "pymeasure.instruments".

Updated CHANGES.rst and "File Structure" section of the "Adding an instrument" documentation.

bmoneke:
Thanks for the change.

@bilderbuchi, you may merge, if you approve the removal.

pymeasure/instruments/srs/sr830.py
```diff
@@ -26,7 +26,8 @@
 import time
 import numpy as np
 from enum import IntFlag
-from pymeasure.instruments import Instrument, discreteTruncate
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import discreteTruncate
```
bilderbuchi, line 30: The `discreteTruncate` import should probably be added to the next line, instead.

pymeasure/instruments/__init__.py
```diff
@@ -22,56 +22,5 @@
 # THE SOFTWARE.
 #
 
-from ..errors import RangeError, RangeException
 from .channel import Channel
 from .instrument import Instrument
-from .resources import list_resources
```
bilderbuchi, line 28: Why remove that one? That seems quite useful to have at the `pymeasure.instruments` layer. I thought that the instruments create the big memory footprint?

CHANGES.rst
```diff
@@ -17,6 +17,10 @@ Deprecated features
 - The property creators :code:`control`, :code:`measurement`, and :code:`setting` do not accept arbitrary keyword arguments anymore. Use the :code:`v_kwargs` parameter to give further arguments to :code:`values` method.
 - The property creators :code:`control`, :code:`measurement`, and :code:`setting` do not accept `command_process` anymore. Use a dynamic property or a `Channel` instead, as appropriate.
 
+Dropped Support
+---------------
+- Instrument manufacture modules are no longer imported in the :code:`pymeasure/instruments/__init__.py` file. Instrument classes will need to be imported from the manufacturer module or explicitly from the instrument driver file. For example, :code:`from pymeasure.instruments import Extreme5000` will need to change to :code:`from pymeasure.instruments.extreme import Extreme5000` or :code:`from pymeasure.instruments.extreme.extreme5000 import Extreme5000`.
```
bilderbuchi, line 22: I think it would be good to illuminate the reasoning behind this change, otherwise it's not really clear why this practical aspect of our API was removed.

bilderbuchi:
> @bilderbuchi could you give your placet, that only "Channel" and "Instrument" are imported into "pymeasure.instruments".

I'm not sure why this removed `list_resources`, too, and remarked on that.

I'm fine with this change in principle, but I would like to get a release out there _before_ merging this, so that we get more time on master to play with this/expose potential problems. 
I want to check the docs, autocomplete in python, ipython, IDE, to make sure this still works as expected (which I find puzzling).

pymeasure/instruments/__init__.py
mcdo0486: In short I removed this import because it isn't required for all instruments and so it shouldn't be included in the init.py. Functionally there is no difference at the moment, because `Instrument` imports `VISAAdapter` thus `pyvisa`, but as per the `SCPIMixin` class that can change https://github.com/pymeasure/pymeasure/pull/905 . 

That's my reasoning, but if you think it should be included for other reasons I can revert this change.

CHANGES.rst
mcdo0486: I added a couple lines explaining the memory footprint issue

pymeasure/instruments/srs/sr830.py
mcdo0486: Added to the other validator imports

pymeasure/instruments/__init__.py
bilderbuchi: >  because it isn't required for all instruments and so it shouldn't be included in the init.py. 

I'm not sure I follow this line of thought. We should then also remove `Channel`, because that's also not required for all instruments.
I'm thinking that this is a useful convenience feature for users in need of troubleshooting, so it's good if it can be easily discovered/accessed. Might be that others disagree? @bmoneke ?

bmoneke: I've never used it, so at least for me, it is not that important, to have it at that position.
But other use cases need it more often.

It might be useful, if you use it more dynamically, plugging devices in and out.

bmoneke: I'm open for keeping it in, if anyone deems it useful. As you like to keep it, let us keep it in the init file.

mcdo0486: Sounds good, I reverted this change in both docs and instruments/__init__.py

bmoneke:
> I'm fine with this change in principle, but I would like to get a release out there before merging this, so that we get more time on master to play with this/expose potential problems.

@bilderbuchi , now that version 0.13.1 is released, could we tackle this issue?

bilderbuchi:
Yeah why not. Note that the changelog entry will be in the wrong place now, and prbably has to be corrected.

If we still need this testing:
> I want to check the docs, autocomplete in python, ipython, IDE, to make sure this still works as expected (which I find puzzling).

I won't find the time for that, but if someone else wants to do that, that would be appreciated (or we don't need it?).

bmoneke:
@mcdo0486 would you mind adjusting the Changelog and fixing the conflict?

If you want, I can help with the checks:
- [ ] IPython
- [ ] Pyhton
- [ ] IDE
- [ ] documentation

mcdo0486:
> @mcdo0486 would you mind adjusting the Changelog and fixing the conflict and ensure that the master branch is merged into your branch (for the documentation)?
> 
> If you want, I can help with the checks:
> 
> * [ ]  IPython
> * [ ]  Pyhton
> * [ ]  IDE
> * [ ]  documentation

Yup I'll work on that and bring this branch up to date.

> I want to check the docs, autocomplete in python, ipython, IDE, to make sure this still works as expected (which I find puzzling).

I'll also work on that.

CHANGES.rst
```diff
@@ -175,6 +175,10 @@ New Contributors
 
 **Full Changelog**: https://github.com/pymeasure/pymeasure/compare/v0.11.1...v0.12.0
 
+Dropped Support
+---------------
+- Instrument manufacturer modules are no longer imported in the :code:`pymeasure/instruments/__init__.py` file. When importing a single instrument into a procedure all instruments would be imported into memory through the manufacturer modules. This change lowers the memory footprint of pymeasure when importing an instrument. Instrument classes will need to be imported from the manufacturer module or explicitly from the instrument driver file. For example, :code:`from pymeasure.instruments import Extreme5000` will need to change to :code:`from pymeasure.instruments.extreme import Extreme5000` or :code:`from pymeasure.instruments.extreme.extreme5000 import Extreme5000`.
+
```
BenediktBurger, line 181: Please move this part to the top of th file (the upcoming version).

BenediktBurger (approved):
One last change (moving the changelog entry to the correct version) and we are good to go.

PR #923, opened 2023-05-25
Dev/feeltech-FY6800_drivers

ashokbruno:
Add Instrument FY6800 AWG


pymeasure/instruments/feeltech/feeltechFY6800.py
```diff
@@ -0,0 +1,122 @@
...
+        else:
+            return cls[input_value.upper()]
+
+
+class AWGChannel(Channel):
+    """Implementation of a AWG channel."""
+
+    waveform_type = Instrument.control(
+        "R<ch>W",
+        "W<ch>W%s",
```
bmoneke, line 73: For channels, we use the "f-string" conversion (see `str.format()`), so you should use `{ch}`:
```suggestion
        "R{ch}W",
        "W{ch}W%s",
```

Obviously, this has to reflect in `insert_id` as well.

pymeasure/instruments/feeltech/feeltechFY6800.py
```diff
@@ -0,0 +1,122 @@
...
+        """ Waveform Type based on the enum.""",
+        validator=strict_discrete_set,
+        values="FeelTechFY6800.WaveformType",
+    )
+
+    def insert_id(self, command):
+        return command.replace("<ch>", "F" if self.id == 2 else "M")
```
bmoneke, line 80: Why don't you call the channels "F" and "M" (in the ChannelCreator, see comment below)? Then you would not have to do this translation here.

In case you stay with 1,2 and M,F, you should use the format string (see above):
```suggestion
        return command.format(ch="F" if self.id == 2 else "M")
```

pymeasure/instruments/feeltech/feeltechFY6800.py
```diff
@@ -0,0 +1,122 @@
...
+        TRAPEZOIDAL_PULSE = "24"
+        SINC_PULSE = "25"
+        NARROW_PULSE = "26"
+        GAUS_WHITE_NOISE = "27"
+        AM = "28"
+        FM = "29"
+        LINEAR_FM = "30"
+
+    def __init__(self, adapter=SerialAdapter, name="FeelTech FY6800", includeSCPI=True, **kwargs):
+        super().__init__(adapter, name, includeSCPI, **kwargs)
```
bmoneke, line 120: It is not necessary, that a user changes `includeSCPI`, therefore, it should not be in the signature, neither should there be the "SerialAdapter" as a default:
```suggestion
    def __init__(self, adapter, name="FeelTech FY6800", **kwargs):
        super().__init__(adapter, name, includeSCPI=True, **kwargs)
```

pymeasure/instruments/feeltech/feeltechFY6800.py
```diff
@@ -0,0 +1,122 @@
...
+
+    channels = Instrument.ChannelCreator(AWGChannel, (1, 2))
```
bmoneke, line 122: If you change this here the channel ids, you do not need `insert_ID` anymore:
```suggestion
    channels = Instrument.ChannelCreator(AWGChannel, ("M", "F"))
```

docs/api/instruments/anritsu/anritsuMS464xB.rst
```diff
@@ -29,3 +29,7 @@ Anritsu MS464xB Vector Network Analyzer
 .. autoclass:: pymeasure.instruments.anritsu.anritsuMS464xB.Port
     :members:
     :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.anritsu.AnritsuMS464xB
+    :members:
+    :show-inheritance:
```
bmoneke, line 35: That instrument is already listed in lines 17-19!
```suggestion
```

pymeasure/instruments/feeltech/feeltechFY6800.py
```diff
@@ -0,0 +1,122 @@
...
+
+class AWGChannel(Channel):
+    """Implementation of a AWG channel."""
+
+    waveform_type = Instrument.control(
+        "R<ch>W",
+        "W<ch>W%s",
+        """ Waveform Type based on the enum.""",
+        validator=strict_discrete_set,
+        values="FeelTechFY6800.WaveformType",
```
bmoneke, line 76: If you want to return the enum member, you can use `cast`:
```suggestion
        """ Waveform Type based on the enum.""",
        validator=strict_discrete_set,
        values="FeelTechFY6800.WaveformType",
        cast=FeelTechFY6800.WaveformType,
```

That way you get the enum member back.

bmoneke (commented):
A few comments. Thanks for the contribution.

PR #927, opened 2023-06-04
Add 'read_tmo_ms' as parameter to Prologix Adapter

neuschs:


bmoneke:
In general, I'm fine with adding this property.
A few minor points though:
- Do we want to give it a more descriptive name (e.g. read_timeout_ms, for example), which deviates from the prologix command?
- The default value could be "None" and only if a user supplies something different than None, the timeout will be changed.

These are design questions.

@msmttchr what do you think?

bmoneke:
> I have mixed opinions about making it not apply in the default case ...

My intention for this is, to not change the current configuration of the adapter, such that you can keep easily a setting which works for you and you do not have to give that parameter for each instrument instantiation.

neuschs:
A fact that I initially not thought about ... I agree

msmttchr:
> In general, I'm fine with adding this property. A few minor points though:
> 
>     * Do we want to give it a more descriptive name (e.g. read_timeout_ms, for example), which deviates from the prologix command?
> 
>     * The default value could be "None" and only if a user supplies something different than None, the timeout will be changed.
> 
> 
> These are design questions.
> 
> @msmttchr what do you think?

I am good with

> In general, I'm fine with adding this property. A few minor points though:
> 
>     * Do we want to give it a more descriptive name (e.g. read_timeout_ms, for example), which deviates from the prologix command?
> 
>     * The default value could be "None" and only if a user supplies something different than None, the timeout will be changed.
> 
> 
> These are design questions.
> 
> @msmttchr what do you think?

I am good with this proposal. I think it is clear and useful.

msmttchr (approved):
LGTM. Thanks.

bmoneke:
Thanks for also adding a test!

PR #928, opened 2023-06-07
New instrument: Wentworth s200 probe table

sansanda:
I'd like to add an instrument. The [Wentworth s200](https://www.wentworthlabs.com/semi-automatic-production-probers/) probe table.
The communication is implemented according to the manufacturer's documentation and tested with the instrument.
You can see the complete operating and service manual [here](https://twiki.cern.ch/twiki/pub/Main/AtlasEdinburghGroupHardwareUpgradeDocumentation/Pegasus_OperatingAndServiceManual_v6.2.pdf).


pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+        self.command_execution_info = S200.ExecutionInfoCode(0)
+
+    chuck_lift = Instrument.setting(
+        "%s",
+        "Control the chuck to the fine down (CDW) or fine up (CUP) position.",
+        # The GDW/GUP command is used to move the chuck to the Gross Up or Gross
+        # Down positions.
+
+        validator=strict_discrete_set,
+        values={True: 'CUP', False: 'CDW'},
```
mcdo0486, line 111: `chuck_lift = True` would seem to indicate the `check_lift` is enabled, not that it will move up or down in a direction.

`values={'up': 'CUP', 'down': 'CDW'}` 
or
`values={'fine_up': 'CUP', 'fine_down': 'CDW'}`

Would one of those make more sense for procedure code?

pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+        )
+        self.write_delay = write_delay
+        self.query_delay = query_delay
+        self.last_write_timestamp = 0.0
+        self.command_execution_info = S200.ExecutionInfoCode(0)
+
+    chuck_lift = Instrument.setting(
+        "%s",
+        "Control the chuck to the fine down (CDW) or fine up (CUP) position.",
+        # The GDW/GUP command is used to move the chuck to the Gross Up or Gross
```
mcdo0486, line 107: Instead of a comment, you can add something like "See `chuck_gross_lift` controls up and down gross movement of the chuck." to the docstring.

pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+        validator=strict_discrete_set,
+        values={True: 'CUP', False: 'CDW'},
+        map_values=True,
+        check_set_errors=True
+    )
+
+    chuck_gross_lift = Instrument.setting(
+        "%s",
+        "Control the chuck to the gross down (CDW) or fine up (CUP) position.",
+        # The CDW/CUP command is used to move the chuck to the Fine Down or
```
mcdo0486, line 119: Similar above both for moving informative information from a comment to the docstring, and changing the values from booleans to more descriptive strings.

pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+        check_set_errors=True
+    )
+
+    x_position = Instrument.control(
+        "PSS X", "GTS X,%d",
+        "Control the x-axis position in microns",
+        validator=strict_range,
+        values=X_POS_VALID_RANGE,
+        get_process=lambda r:
+        float(r[1]),
```
mcdo0486, line 158: It seems the lambdas are auto-formatted in multi line when they  should be on a single line.

What is `r[0]` in the reply message? It may be that a uniform lambda passed to `preprocess_reply` argument would be useful for all these commands that ignore the first element from the response.

For instance, `x_position` and `xy_position` can have the same `preprocess_reply` instead of `xy_position` having a custom `get_process`.

pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+        "Control the z-axis fine lift (in um) of the chuck of the probe station",
+        # AWP compatible: No
+        validator=strict_range,
+        values=Z_FINELIFT_VALID_RANGE,
+        check_set_errors=True,
+        get_process=lambda r:
+        float(r.replace("RKFM ", ""))
+    )
+
+    indexing_mode = Instrument.setting(
```
mcdo0486, line 257: To fit style in [writing boolean properties](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html?highlight=enabled#boolean-properties), I think this should be `indexing_mode_enabled`

pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+        "NX%s",
+        "Control the movement of the chuck to the next die",
+        # If probe_table.next_die = "D" then it will go below the current die
+        # If probe_table.next_die = "U" then it will go above the current die
+        # If probe_table.next_die = "L" then it will go to the left of the current die
+        # If probe_table.next_die = "R" then it will go to the right of the current die
+        # AWP compatible: Yes
+        check_set_errors=True,
+        validator=strict_discrete_set,
+        values=['D', 'U', 'L', 'R']
```
mcdo0486, line 283: Similar to before, the comment should be moved to the docstring, and I think descriptive values would help:
```
values={'down': 'D', 'up': 'U','left': 'L', 'right': 'R'}
map_values=True
```

pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+    )
+
+    status_byte = Instrument.measurement(
+        "STA",
+        "Measure the status byte of the instrument",
+        get_process=lambda r:
+        r
```
mcdo0486, line 290: All of these `lambda r: r` can be removed, as they are the same as the default `get_process` lambda.

pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+        "Set the proble table in local mode. Exits remote mode."
+        # This command does not send a reply. However, when remote mode is re-entered,
+        # the Pegasus unit sends an INF_000 message
+    )
+
+    serial_number = Instrument.measurement(
+        "GSN",
+        "Measure the serial number of the probe table",
+        # AWP compatible: No
+        get_process=lambda r:
```
mcdo0486, line 311: Instead of `get_process=lambda r: str(r)`, I think using `cast=str` should work? [The `CommonBase.values` method hits earlier](https://github.com/pymeasure/pymeasure/blob/3c288c7909acc5cd3c36af425f41b1e891152bae/pymeasure/instruments/common_base.py#L472)  then `get_process`, and it takes in the `cast` argument.

pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+    chuck_override = Instrument.setting(
+        "%s",
+        "Control the chuck override of the main chuck of the probe station.",
+        validator=strict_discrete_set,
+        values={True: 'CO1', False: 'CO0'},
+        map_values=True,
+        check_set_errors=True
+    )
+
+    lamp_on = Instrument.setting(
```
mcdo0486, line 143: Probably should be `lamp_enabled`

pymeasure/instruments/wentworthlabs/s200.py
```diff
@@ -0,0 +1,548 @@
...
+
+
+# log.addHandler(logging.NullHandler())
+
+
+class S200(Instrument):
+    """ Represents the Wenthworth Labs S200 Probe Table
+    and provides a high-level for interacting with the instrument.
+
+    .. code-block:: python
```
mcdo0486, line 42: I think a four space indent is needed to generate the code-block:
```
    .. code-block:: python

        s200 = S200("GPIB0::28::INSTR", query_delay=0.5, write_delay=0.5)
        # test probe table lamp
```

mcdo0486 (commented):
Not familiar with the instrument, but added some general feedback to get the ball rolling.

pymeasure/instruments/wentworthlabs/s200.py
mcdo0486: Just checked out the tests and I see why you ignore the first element.
Yea, I think a `preprocess_reply` may be a better approach than `get_process`.
```
    x_position = Instrument.control(
        "PSS X", "GTS X,%d",
        "Control the x-axis position in microns",
        validator=strict_range,
        values=X_POS_VALID_RANGE,
        preprocess_reply=lambda r: r[r.index(',')+1:],
```

That should `preprocess_reply` would work for all the position properties. Before parsing the reply string for values, ignore characters up to the first comma. It will return single floats or list of floats. You can do something similar for other properties:

```
    y_index = Instrument.control(
        "RYM", "WYM %d",
        "Control the y index of the chuck in units of 10 microns ",
        # AWP compatible: Yes
        validator=strict_range,
        values=X_Y_INDEX_VALID_RANGE,
        check_set_errors=True,
        preprocess_reply=lambda r: r.replace("RYM ", "")
    )
```

pymeasure/instruments/wentworthlabs/s200.py
bmoneke: If you want to move something up or down, methods are better: one for moving upwards, one for downwards.

pymeasure/instruments/wentworthlabs/s200.py
bmoneke: Exactly

sansanda: Hi. Thanks.

pymeasure/instruments/wentworthlabs/s200.py
sansanda: Got it!!! Thanks

pymeasure/instruments/wentworthlabs/s200.py
sansanda: Thanks to both of you. I thing I will create methods.

sansanda: Many thanks to both of you!!!

pymeasure/instruments/wentworthlabs/s200.py
sansanda: many thanks

pymeasure/instruments/wentworthlabs/s200.py
sansanda: thanks

pymeasure/instruments/wentworthlabs/s200.py
sansanda: many thanks

Issue #666, opened 2022-08-05
Extend instrument's property factories to unitful quantities.


bmoneke:
Instrument's property factories do not take into account unitful quantities (pint).

`set_process` and `get_process` can be used as a stop-gap measure, but a proper implementation would be nice.
For example validators cannot work with quantities yet.

bilderbuchi:
Most relevant notes from #248:
> We should first find a good, working design before implementing it. Some thoughts about potential approaches that I had in the back of my head were to
> 
>     * Add a `unit` attribute to the property factories to indicate which unit the property is using. Now that we have dynamic properties, we can even change that at runtime! Very useful e.g. when you can change the unit in the instrument.
> 
>     * Have a global switch (`pymeasure.USE_UNITFUL_QUANTITIES`?) to disable the behaviour by default (at least in a transition period), or find another good way to on-ramp users while eliminating/limiting breakage of existing code.
> 
>     * If `USE_UNITFUL_QUANTITIES` is True _and_ a property has a `unit` defined, interaction (get/set) happens with pint Quantities.
> 
>     * I think it's possible for properties to always _accept_ float, and then assume they are in the property's `unit`.
> 
> 
> > I think it is good to be unit aware, but would not enforce it (but recommend it) for new instruments.
> 
> Agreed!





> Having an option to deactivate units, might be good, because someone might not want to use units (simple program). Maybe a missing pint library switches to that mode? It's only necessary to change the getter, as the setter is always able to handle numbers (see answer below)
> 
> > I think it's possible for properties to always accept float, and then assume they are in the property's unit.
> 
> Yes, that should be done. I already improved IK's `assume_units` method, which takes the input (number or string or quantity) and returns a quantity, to be used by the driver.

bmoneke:
In #929 an instrument with unitful quantities is added, which should be revisited after this issue is resolved.

PR #929, opened 2023-06-07
Add support for Agilent L4534A Digitizer

apaulfox:
This adds support for the 4-channel, 16-bit digitizer from Agilent. It also adds a tst() function for base SCPI devices to support running self-tests with *TST?

Included tests have all been run and passed, including tests with device.

Functionality is not fully tested, but all the basics work as far as I can tell.


pymeasure/adapters/__init__.py
```diff
@@ -39,10 +39,14 @@
...
 try:
     from pymeasure.adapters.serial import SerialAdapter
-    from pymeasure.adapters.prologix import PrologixAdapter
 except ImportError:
     log.warning("PySerial library could not be loaded")
 
+try:
+    from pymeasure.adapters.prologix import PrologixAdapter
+except ImportError:
+    log.warning("Prologix library could not be loaded")
```
bmoneke, line 48: Good catch.
Prologix does not depend on PySerial anymore, but on pyvisa. Therefore it is better to put it with VISAAdapter in the same try clause (a few lines above).

pymeasure/instruments/instrument.py
```diff
@@ -134,6 +134,13 @@ def id(self):
         else:
             raise NotImplementedError("Non SCPI instruments require implementation in subclasses")
 
+    def tst(self, delay=10):
+        """ Run instrument self-test """
+        if self.SCPI:
+            return self.ask("*TST?", delay).strip()
+        else:
+            raise NotImplementedError("Non SCPI instruments require implementation in subclasses")
+
```
bmoneke: Is that a default SCPI command?
Do (almost) all SCPI compliant devices have that command?
If not, I'd put it in the particular instrument, not here.

@msmttchr , do you know more?

pymeasure/instruments/agilent/agilentL4534A.py
bmoneke, line 1: A general note:
Please add links to the documentation as well, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,333 @@
...
+
+from decimal import Decimal
+
+from pymeasure.errors import Error
+from pymeasure.instruments import Instrument, Channel
+from pymeasure.instruments.validators import strict_discrete_set, strict_range, strict_discrete_range
+from pymeasure.units import ureg
+
+
+def sample_count_function(value, values): return strict_discrete_range(value, values, 4)
```
bmoneke, line 38: I have never seen a function definitino in a single line in python. Afaik it is recommended to always start a new line after a colon:
```suggestion
def sample_count_function(value, values):
    return strict_discrete_range(value, values, 4)
```

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,333 @@
...
+SAMPLE_COUNT_VALUES = [8, 128e6]
+PRE_TRIG_VALUES = [0, 128e6]
+RECORD_COUNT_VALUES = [1, 1024]
+
+TRIGGER_HOLDOFF_VALUES = ureg.Quantity(np.asarray([0, 10]), ureg.s)
+TRIGGER_DELAY_VALUES = ureg.Quantity([0, 3600], ureg.s)
+
+def _get_channel_config_process(values):
+    result = {
+        'range': ureg.Quantity(Decimal(values[0]), ureg.V),
```
bmoneke, line 51: Do you get any benefit by using Decimal over a plain float? Especially, after we received a string and turned that into a float.

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,333 @@
...
+    def _validate_acq_config(value, values):
+        strict_discrete_set(value['sample_rate'], AgilentL4534A.SAMPLE_RATE_VALUES)
+        strict_discrete_range(value['samples_per_record'], SAMPLE_COUNT_VALUES, 4)
+        strict_discrete_range(value['pre_trig_samples'], PRE_TRIG_VALUES, 4)
+        strict_range(value['num_records'], RECORD_COUNT_VALUES)
+        strict_range(value['trigger_holdoff'], TRIGGER_HOLDOFF_VALUES)
+        strict_range(value['trigger_delay'], TRIGGER_DELAY_VALUES)
+        return value
+
+    class Channel(Channel):
```
bmoneke, line 122: I think it is better not to replace the (pymeasure) Channel with your new Channel, but to use a new name.

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,333 @@
...
+            strict_discrete_set(value['coupling'], COUPLING_VALUES)
+            strict_discrete_set(value['filter'], FILTER_VALUES)
+            return value
+    
+        config = Instrument.control(
+            "CONF:CHAN:ATTR? (@{ch})",
+            "CONF:CHAN:ATTR (@{ch}),%s",
+            """Control Channel configuration with dict containing range (in V), coupling, and filter""",
+            set_process=lambda v: '{:.3g},{},{}'.format(
+                v['range'].m_as(ureg.V), v['coupling'], v['filter']),
```
bmoneke, line 139: As this is the first instrument using pint quantities:
I'd like, that a mere number (e.g. 5 instead of 5 V) can be accepted as well.

We could use something like "assume_units" from instrument kit:
https://github.com/instrumentkit/InstrumentKit/blob/7352a08b3e613b958b4a20626d778dcdcf5080d7/src/instruments/util_fns.py#L22-L43

I improved that function and instrumentkit is open source, so there should be no problem.

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,333 @@
...
+
+            strict_discrete_set(value['range'], VOLTAGE_RANGE_VALUES)
+            strict_discrete_set(value['coupling'], COUPLING_VALUES)
+            strict_discrete_set(value['filter'], FILTER_VALUES)
+            return value
+    
+        config = Instrument.control(
+            "CONF:CHAN:ATTR? (@{ch})",
+            "CONF:CHAN:ATTR (@{ch}),%s",
+            """Control Channel configuration with dict containing range (in V), coupling, and filter""",
```
bmoneke: It is a more general comment: Docstrings should end with a period (or exclamation mark etc.):
```suggestion
            """Control Channel configuration with dict containing range (in V), coupling, and filter.""",
```

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,333 @@
...
+        @property
+        def voltage(self) -> typing.NDArray[np.float32]:
+            """Get voltage measurements for this channel"""
+            self.write(f'FETC:WAV:VOLT? (@{self.id})')
+            data = self._read_data_block()
+            return ureg.Quantity(np.frombuffer(data, dtype='>f4'), ureg.V)
+
+        @property
+        def adc(self) -> typing.NDArray[np.int16]:
+            """Get raw ADC measurements for this channel in counts (-32,767,+32,767) in current voltage range"""
```
bmoneke: The use of commas is misleading. You could drop them, or use a semicolon in the middle:
```suggestion
            """Get raw ADC measurements for this channel in counts (-32767, +32767) in current voltage range"""
```

tests/instruments/agilent/test_agilentL4534A.py
bmoneke, line 1: Thanks for adding tests.

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def arm(self) -> None:
+        """
+        Trigger ARM condition in software
+        """
+        self.write('ARM')
+
+    def init(self) -> None:
+        """
+        Initialize measurement with current configuration.\n
```
bmoneke, line 344: Why do you use newline chars (\n) in the docstring?

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,333 @@
...
+        Offsets will be cleared when instrument is reset or settings are changed.
+        """
+        return int(self.ask('CAL:ZERO:AUTO? (@{})'.format(','.join(channels)), 5).strip())
+    
+    channels = Instrument.ChannelCreator(Channel, (1, 2, 3, 4))
+
+    def __init__(self, adapter, name="Agilent L4534A Digitizer", **kwargs):
+        super().__init__(
+            adapter, name, **kwargs
+        )
```
bmoneke: It's best to have the init at the top of a class in order to find it more easily.

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,333 @@
...
+        """Control minimum time between triggers, during which trigger inputs are ignored,
+          in decimal seconds (0-10s)""",
+        validator=strict_range,
+        values=TRIGGER_HOLDOFF_VALUES
+    )
+
+    trigger_delay = Instrument.control(
+        "CONF:ACQ:TDEL?",
+        "CONF:ACQ:TDEL %e",
+        """Control delay time after trigger before acquistion starts, in decimal seconds (0-3600s)""",
```
bmoneke: What are "decimal seconds"?

tests/instruments/agilent/test_agilentL4534A_with_device.py
```diff
@@ -0,0 +1,126 @@
...
+    Unit tests for TestAgilentL4534A class.
+
+    This test suite, needs the following setup to work properly:
+        - A TestAgilentL4534A device should be connected to the computer;
+        - The device's address must be set in the RESOURCE constant.
+    """
+
+    ###############################################################
+    # Agilent34450A device goes here:
+    RESOURCE = "TCPIP::169.254.45.32::INSTR"
```
bmoneke, line 45: Please have a look here, how to make these tests skip conditionally on the definition of an adapter (which replaces your "RESOURE" constant): https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#device-tests

bmoneke (changes requested):
Thanks your contribution. You may add yourself to the authors.txt.

I left a few comments, what to improve or questions.
As this is the first instrument with pint units, we might have to decide a few things.

I hope the comments help you.

pymeasure/instruments/instrument.py
apaulfox: According to the SCPI standard (https://www.ivifoundation.org/docs/SCPI-99.PDF), it's mandated command for IEEE 488.2 and therefore also for SCPI. I've looked through a handful of instruments I have on hand, and they all appear to support it.

pymeasure/instruments/agilent/agilentL4534A.py
apaulfox: Makes sense, I assume that should go in pymeasure.units?

pymeasure/instruments/agilent/agilentL4534A.py
apaulfox: Seconds expressed as a decimal, e.g. 1.5 or 0.01, so not just whole seconds. Maybe there's a clearer way to phrase that?

pymeasure/instruments/agilent/agilentL4534A.py
apaulfox: OK, fixed.

pymeasure/instruments/agilent/agilentL4534A.py
bmoneke: yes, I think so.

pymeasure/instruments/agilent/agilentL4534A.py
bmoneke: In our documentation (https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings) we used the example to state the type in parenthesis: "Control something (float).".

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,384 @@
...
+            values=FILTER_VALUES
+        )
+
+        def _read_data_block(self):
+            # Data block has header of form '#<n><n digits>',
+            # where <n>specifies the number of digits used to indicate the size of the block,
+            # and the <n digits> indicate the block size in bytes.
+            digits = self.read_bytes(2).decode()
+            if not digits.startswith('#'):
+                raise Error()
```
mcdo0486, line 183: @bmoneke This is the only place `Error` is used in the codebase, besides being passed directly to [`RangeError`](https://github.com/pymeasure/pymeasure/blob/3c288c7909acc5cd3c36af425f41b1e891152bae/pymeasure/errors.py)

I think `pymeasure/errors.py` would be for defining pymeasure-specific error types, but a generic `Error` type seems too broad. Should `Error` be modified to `InstrumentError` or something more pymeasure-y or be replaced with a builtin `Exception` type?

Also I think there should be a user-friendly message that is passed with this statement if `digits` doesn't start with `#`.

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,384 @@
...
+    Represents the Agilent L4532A/L4534A digitizers.
+    """
+    def __init__(self, adapter, name="Agilent L4534A Digitizer", **kwargs):
+        super().__init__(
+            adapter, name, **kwargs
+        )
+
+    SAMPLE_RATE_VALUES = SAMPLE_RATE_VALUES
+
+    class DigitizerChannel(Channel):
```
mcdo0486, line 126: I'm not familiar with Agilient digitizers, but would other models in that product line be able to use `DigitizerChannel`? Maybe with dynamic properties? If so, I think it'd make more sense to move `DigitizerChannel` out of the `AgilentL4534A` namespace and have it defined somewhere else. If not, it makes sense to leave it there.

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def clear_display(self):
+        return self.write('DISP:TEXT:CLE')
+
+    arm_source = Instrument.control(
+        "CONF:ARM:SOUR?",
+        "CONF:ARM:SOUR %s",
+        """Set the source used to arm the digitizer (IMMediate|EXTernal|SOFTware|TIMer).""",
+        validator=strict_discrete_set,
+        values=['IMM', 'SOFT', 'EXT', 'TIM']
```
mcdo0486, line 224: The way string `values` are used is not consistent across instruments, and there isn't guidance in the pymeasure docs about it, but I think the abbreviated commands should be mapped to the lowercase word for which they stand for e.g.
```
      values={'immediate': 'IMM', 'software': 'SOFT', 'external': 'EXT','timer': 'TIM'],
      map_values=True
```

That allows procedure code that looks like ` inst.arm_source = 'software' ` which brings more clarity at a glance when a user is trying to understand the procedure.

This shouldn't hold up this PR, it can be a post for discussion or an opened issue, just wondering what you think.

apaulfox:
I'm not sure what the issue is with the documentation build. I can't reproduce the issue on my machine,

apaulfox:
> > I'm not sure what the issue is with the documentation build. I can't reproduce the issue on my machine,
> 
> Sphinx is flagging some warnings from the docstrings in the `init` and `auto_zero` methods related to the use of "section title" formatting `-----`, the dashes in those docstrings: https://github.com/pymeasure/pymeasure/blob/409ebeecf9497f73fa77315c3f8dac8ada8a42f8/pymeasure/instruments/agilent/agilentL4534A.py#L392C3-L416
> 
> If you remove those dashes then it should work.

Huh, OK, apparently sphinx only started supporting PEP 287 markdown-style docstrings this year (https://github.com/sphinx-doc/sphinx/pull/10919)? It looks like the workflow is running a fairly old version of sphinx, so that would explain it. That's unfortunate, because VS Code doesn't support formatting for the old sphinx-style docstrings, so it's not going to look good.

mcdo0486:
> > > I'm not sure what the issue is with the documentation build. I can't reproduce the issue on my machine,
> > 
> > 
> > Sphinx is flagging some warnings from the docstrings in the `init` and `auto_zero` methods related to the use of "section title" formatting `-----`, the dashes in those docstrings: https://github.com/pymeasure/pymeasure/blob/409ebeecf9497f73fa77315c3f8dac8ada8a42f8/pymeasure/instruments/agilent/agilentL4534A.py#L392C3-L416
> > If you remove those dashes then it should work.
> 
> Huh, OK, apparently sphinx only started supporting PEP 287 markdown-style docstrings this year ([sphinx-doc/sphinx#10919](https://github.com/sphinx-doc/sphinx/pull/10919))? It looks like the workflow is running a fairly old version of sphinx, so that would explain it. That's unfortunate, because VS Code doesn't support formatting for the old sphinx-style docstrings, so it's not going to look good.

Yea, [`requirements/pymeasure.yml`](https://github.com/pymeasure/pymeasure/blob/3c288c7909acc5cd3c36af425f41b1e891152bae/requirements/pymeasure.yml) is only used by GitHub actions, and it currently installs `sphinx==4.3.1` for running the workflow.

apaulfox:
OK, I think I've addressed all the feedback, so take a look and see if it looks good now.

bmoneke:
I ask you to be patient regarding another review round, I'm not sure, when I can do it, due to a medical issue.

apaulfox:
> I ask you to be patient regarding another review round, I'm not sure, when I can do it, due to a medical issue.

No worries, I can wait.

bilderbuchi:
> I ask you to be patient regarding another review round, I'm not sure, when I can do it, due to a medical issue.

Rest and get well! 👨‍⚕️

bilderbuchi:
> > > I'm not sure what the issue is with the documentation build. I can't reproduce the issue on my machine,
> > 
> > 
> > Sphinx is flagging some warnings from the docstrings in the `init` and `auto_zero` methods related to the use of "section title" formatting `-----`, the dashes in those docstrings: https://github.com/pymeasure/pymeasure/blob/409ebeecf9497f73fa77315c3f8dac8ada8a42f8/pymeasure/instruments/agilent/agilentL4534A.py#L392C3-L416
> > If you remove those dashes then it should work.
> 
> Huh, OK, apparently sphinx only started supporting PEP 287 markdown-style docstrings this year ([sphinx-doc/sphinx#10919](https://github.com/sphinx-doc/sphinx/pull/10919))? It looks like the workflow is running a fairly old version of sphinx, so that would explain it. That's unfortunate, because VS Code doesn't support formatting for the old sphinx-style docstrings, so it's not going to look good.

We can update that to a current stable release no problem, no need to drag along old syntax just because of this.

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+
+        VOLTAGE_RANGE_VALUES = VOLTAGE_RANGE_VALUES
+        COUPLING_VALUES = COUPLING_VALUES
+        FILTER_VALUES = FILTER_VALUES
+
+        config = Instrument.control(
+            "CONF:CHAN:ATTR? (@{ch})",
+            "CONF:CHAN:ATTR (@{ch}),%s",
+            """
+            Control Channel configuration with dict containing range (in V), coupling, and filter.
```
bmoneke, line 155: Are "range", "coupling", "filter" the keys of the dictionary?

Is it necessary to have this "combined" method or would it not be easier (and more descriptive) to use three properties (at least for setting)?

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+            get_process=lambda v: ureg.Quantity(v, ureg.V),  # convert to quantity
+            validator=lambda value, values:
+                strict_discrete_set(assume_units(value, ureg.V), values),
+            values=VOLTAGE_RANGE_VALUES
+        )
+
+        coupling = Instrument.control(
+            "CONF:CHAN:COUP? (@{ch})",
+            "CONF:CHAN:COUP (@{ch}),%s",
+            """Comtrol channel coupling (AC|DC).""",
```
bmoneke, line 180:
```suggestion
            """Control channel coupling (AC|DC).""",
```

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+            :param dtype: Numpy data type expected for this data block
+            :return: numpy array of specified type
+            """
+            # Data block has header of form '#<n><n digits>',
+            # where <n>specifies the number of digits used to indicate the size of the block,
+            # and the <n digits> indicate the block size in bytes.
+            digits = self.read_bytes(2).decode()
+            if digits.startswith('#'):
+                try:
+                    bytes = int(self.read_bytes(int(digits[1])).decode())
```
bmoneke, line 205: As bytes is a python type, I would not use it as a variable name.

bmoneke: "digits[1]" is already an integer

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+                    )
+
+        @property
+        def voltage(self):
+            """Get voltage measurements for this channel (array of float in V)."""
+            self.write(f'FETC:WAV:VOLT? (@{self.id})')
+            return ureg.Quantity(self._read_data_block('>f4').astype(np.float32), ureg.V)
+
+        @property
+        def adc(self):
```
bmoneke, line 231: I'm not happy with "adc", as it is not clear (from the name), what you intend.
Maybe "voltage_raw" (as it is the voltage measured, but returned in raw values).

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+            """
+            Get raw ADC measurements for this channel in counts (int from -32767 to +32767)
+            in current voltage range.
+            """
+            self.write(f'FETC:WAV:ADC? (@{self.id})')
+            return self._read_data_block('>i2').astype(np.int16)
+
+    channels = Instrument.ChannelCreator(DigitizerChannel, (1, 2, 3, 4))
```
bmoneke, line 239: Please move this line up to the beginning of the class, such that you see the important structure of the class at the beginning: Channels and knit method.

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+
+    display = Instrument.control(
+        'DISP:TEXT?',
+        'DISP:TEXT \"%s\"',
+        """Control Display text on screen, up to 12 characters.""",
+        get_process=lambda v: v.strip('\"')
```
bmoneke, line 245: Isn't "Display" on "screen" a duplication? Besides that it should be a lower case d.

pymeasure/instruments/agilent/agilentL4534A.py
bmoneke: Does the device understand the long (lower case) form? And mixed case?

I think it is good, if the property accepts the long form, as it is more expressive than "Arm_source =" IMM"".

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+        "CONF:ARM:SOUR %s",
+        """Set the source used to arm the digitizer (IMMediate|EXTernal|SOFTware|TIMer).""",
+        validator=strict_discrete_set,
+        values=['IMM', 'SOFT', 'EXT', 'TIM']
+    )
+
+    ext_slope = Instrument.control(
+        "CONF:EXT:INP?",
+        "CONF:EXT:INP %s",
+        """Set the edge to be used for the external trigger input (NEGative or POSitive).""",
```
bmoneke, line 262: It is a control, so the docstring should start with "Control", not "Set". Unless it is a "setting" (not readable).

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+        "CONF:TRIG:SOUR %s",
+        """Set the trigger source (IMMediate|SOFTware|EXTernal|CHANnel|OR).""",
+        validator=strict_discrete_set,
+        values=TRIGGER_SOURCE_VALUES
+    )
+
+    sample_rate = Instrument.control(
+        "CONF:ACQ:SRAT?",
+        "CONF:ACQ:SRAT %d",
+        """Set the sample rate in samples/s (Hz):
```
bmoneke, line 296: Also a Control, isn't it?

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+        """,
+        validator=sample_count_function,
+        values=SAMPLE_COUNT_VALUES
+    )
+
+    pre_trigger_samples = Instrument.control(
+        "CONF:ACQ:SPR?",
+        "CONF:ACQ:SPR %d",
+        """
+        Control number of samples that will captured pre-trigger
```
bmoneke, line 333: Suggestion:
```suggestion
        Control number of samples that will captured before the trigger
```

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+        validator=strict_range,
+        values=TRIGGER_HOLDOFF_VALUES
+    )
+
+    trigger_delay = Instrument.control(
+        "CONF:ACQ:TDEL?",
+        "CONF:ACQ:TDEL %e",
+        """
+        Control delay time after trigger before acquistion starts,
+        in seconds (float from 0 to 3600).
```
bmoneke, line 366: Wow, you can delay it by one hour?

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+        validator=_validate_acq_config
+    )
+
+    def arm(self) -> None:
+        """
+        Trigger ARM condition in software
+        """
+        self.write('ARM')
+
+    def init(self) -> None:
```
bmoneke, line 402: What about "initialize"?

pymeasure/instruments/agilent/agilentL4534A.py
```diff
@@ -0,0 +1,425 @@
...
+        Instrument will singal complete once capture is finished.
+        """
+        self.write('INIT')
+
+    def auto_zero(self, channels=[1, 2, 3, 4]):
+        """
+        Auto-zero the inputs, temporarily loading new offsets for the current config.
+
+        :param channels: list of channels to auto-zero
+        :return: 0 if auto-zero completed successfully, Otherwise, an error ocurred
```
bmoneke, line 420: In python, it is better to raise an error, if one happened.
Some idea:
~~~python
If returned_value != 0:
    Raise SomeError("AutoZero failed.")
~~~

pymeasure/units.py
bmoneke, line 1: Would you mind adding tests for the new function?

You could take a look at instrument kit.

pymeasure/errors.py
```diff
@@ -31,5 +31,9 @@ class RangeError(Error):
     pass
 
 
+class InstrumentError(RuntimeError):
```
bmoneke, line 34: Could you use built in Exception types instead?
Does that new Exception help to express an error better?

Using a custom Exception makes Exception handling always more complicated, as the user has to import that Error as well.

bmoneke (changes requested):
So, now I did a thorough review with quite a few comments.

Some advantage, while waiting for the doctor...

pymeasure/errors.py
bilderbuchi: > Does that new Exception help to express an error better?

In that vein, `InstrumentError` seems not very specific?

pymeasure/instruments/agilent/agilentL4534A.py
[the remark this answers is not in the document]
bilderbuchi: Yes, please let's do the unit stuff separately, such a "core" change will just delay merging this PR needlessly, and it's harder to review.

pymeasure/instruments/instrument.py
```diff
@@ -134,6 +134,13 @@ def id(self):
         else:
             raise NotImplementedError("Non SCPI instruments require implementation in subclasses")
 
+    def tst(self, delay=10):
+        """ Run instrument self-test """
```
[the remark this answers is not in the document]
bilderbuchi: Also, this should be called `self_test`; `tst` is just needlessly dense/cryptic.

pymeasure/units.py
```diff
@@ -25,3 +25,50 @@
 import pint
 
 ureg = pint.get_application_registry()
+
+
+def assume_units(value, units):
+    """
+    If units are not provided for ``value`` (that is, if it is a raw
```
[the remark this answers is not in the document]
bilderbuchi: Is there nothing built-in in `pint` for this?

docs/api/instruments/agilent/agilentL4534A.rst
bmoneke, line 1: Please add the Channel in the API documentation as well, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html

Thanks to this PR, I noticed, that it is not written in the part regarding the API documentation, such that we can deal about it in #943 😁

pymeasure/instruments/agilent/agilentL4534A.py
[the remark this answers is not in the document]
bmoneke: Obviously in a new PR, I just mentioned, that it is something to tackle.

The corresponding issue is #666

PR #931, opened 2023-06-09
Add MultiChannelCreator for creating multiple channels

mcdo0486:
Add `MultiChannelCreator` for creating multiple channels, `ChannelCreator` is used for single channel creation.

[Previous behavior of `ChannelCreator`](https://github.com/pymeasure/pymeasure/blob/3c288c7909acc5cd3c36af425f41b1e891152bae/pymeasure/instruments/activetechnologies/AWG401x.py#L433):
```
     ch = Instrument.ChannelCreator(ChannelAFG, (1, 2))
```

[New behavior `ChannelCreator`](https://github.com/mcdo0486/pymeasure/blob/c8f0ddffaa545ed75443479dfcb0bd2276e4e4c5/pymeasure/instruments/activetechnologies/AWG401x.py#L433):
```
    ch_1 = Instrument.ChannelCreator(ChannelAFG, 1)
    ch_2 = Instrument.ChannelCreator(ChannelAFG, 2)
```

`MultiChannelCreator` behavior is the same as the previous `ChannelCreator`:
1. You can generate many channels with a prefix, 'ch_' as default
2. Channels are added to a channel collection, the name of which is the the class attribute `MultiChannelCreator` is assigned to, recommended to be 'channels'.
```
     channels = Instrument.MultiChannelCreator(ScopeChannel, list(range(1, 33)))
```

Updated current instruments that use `ChannelCreator` so all channels are created as they were with factory behavior of `ChannelCreator`, so there won't be backwards incompatibility with people living off the master branch.

However, while the channel interface names will be the same, e.g. generated channel `ch_1` will now be explicitly declared as a `ch_1` property, the default collection to access those channel interfaces will now be `channels` instead of the assigned attribute name.

Once method naming and conventions of use are discussed, I'll update `docs/dev/adding_instruments/channels.rst` to reflect new guidelines for creating channels.


bmoneke:
A first request: Could you read a Channel's docstring and use that for the docstring of the Creator (and documentation)?

Such that the documentation will say: "ch_3: A voltage measurement channel." , while the content is taken from the channel description. 

That will keep it more simple and helps to keep documentation in touch with the code.

mcdo0486:
> A first request: Could you read a Channel's docstring and use that for the docstring of the Creator (and documentation)?
> 
> Such that the documentation will say: "ch_3: A voltage measurement channel." , this content is taken automatically from the channel description.
> 
> That will keep it more simple and helps to keep documentation in touch with the code.

Gotcha, there are a few ways that could happen. Let me know what you think.

1. It could be part of the style guide that people insert the channel's doc. This kind of hurts readability in the source code, but inserts the channel docs directly.
```
    ch_1 = Instrument.ChannelCreator(ChannelAFG, 1,
                 docstring="ch_1: "+ChannelAFG.__doc__)
```

2. If no docstring is passed to `ChannelCreator`, and there is a channel id, it could format the docstring in that manner.
```
    ch_1 = Instrument.ChannelCreator(ChannelAFG, 1)
```
```
    class ChannelCreator:
        def __init__(self, cls, id=None, docstring="", **kwargs):
            self.__doc__ = "Channel "+str(id)+ " " + cls.__doc__ if id and not docstring else docstring
```
3. Same as 2, but remove the docstring argument from `BaseChannelCreator` and auto generate the string every time.

I would lean towards option 1 as it gives greatest flexibility.

For option 2 and especially 3, there would need to be some base format for Channel docstrings, or it can have a lot of redundancy. For example, below is a screenshot of what would be generated for LeCroy T3DSO1204 Oscilloscope:
image

bmoneke:
I prefer automatic documentation (i.e. option 3). Even some redundant information is better than wrong information.

I think the best option, would be, if sphinx showed the class name as a link (e.g. `ch_1 = SomeFunnyChannel()`, linking to `SomeFunnyChannel` class documentation). I don't know, though, whether that is possible.

mcdo0486:
> I prefer automatic documentation (i.e. option 3). Even some redundant information is better than wrong information.
> 
> I think the best option, would be, if sphinx showed the class name as a link (e.g. `ch_1 = SomeFunnyChannel()`, linking to `SomeFunnyChannel` class documentation). I don't know, though, whether that is possible.

I'm not a sphinx lord, the format is weak with this one. But I'll see if we can modify the sphinx conf.py to allow custom format when parsing that object type.

If we're going with auto documentation I'll draft some guidelines for best practices when adding channels.

mcdo0486:
So the issue was that Sphinx wouldn't render references, like links to other classes (`:class:~pymeasure...`), in an attribute name or object representation. [The screenshot at the end of this comment shows how attribute name and `repr` are rendered as literal strings.](https://github.com/pymeasure/pymeasure/pull/931#issuecomment-1589330451)

I couldn't figure out a way to force links to be rendered, html or rst would be rendered as a literal string.

There is another approach, I modified `docs/conf.py` to include a function that hooks into sphinx when it is emits `autodoc-process-docstring` events. For instruments that have channels, we can add additional rows to the documentation that includes the channel information.

At the moment this is the format for the documentation:

Screenshot of docs for `ChannelCreator`:
Screenshot_2023-06-16_14-42-01

Screenshot of docs for `MultiChannelCreator`:
Screenshot_2023-06-16_14-41-50


Other information could be included, but that is the basic setup.

pymeasure/instruments/common_base.py
```diff
@@ -201,13 +243,41 @@ class attributes that are a DynamicProperty. Check also for class variables
                 setattr(self, self.__reserved_prefix + attr, value)
         return special_names
 
+    @staticmethod
+    def get_channels(cls):
+        """Return channels for any BaseChannelCreator channel"""
```
bilderbuchi, line 248: According to the docstrings, those static methods should be part of `BaseChannelCreator`, not `MultiChannelCreator`, right?

pymeasure/instruments/common_base.py
```diff
@@ -353,24 +436,24 @@ def binary_values(self, command, query_delay=0, **kwargs):
...
+            command_process=None,
+            check_set_errors=False,
+            check_get_errors=False,
+            dynamic=False,
+            preprocess_reply=None,
+            separator=',',
+            maxsplit=-1,
+            cast=float,
+            values_kwargs=None,
+            **kwargs
```
bilderbuchi, line 456: I don't understand this indentation change? This should be indented one level from `def control`, and not two? Why is this here?

bilderbuchi:
It's great that you could manage to have Sphinx create useful API documentation. It's inconvenient that we need a hook for this, I'd have imagined we could generate them while settting up the channels, but what's fair is fair. :shrug: 

How do the channel docstrings look outside of Sphinx? E.g. if you are using IPython, and do `myinstr.ch_1?` ?

Finally, what I didn't grasp -- do channels created with the new `ChannelCreator` also get added to a `self.channels` collection?

mcdo0486:
> It's great that you could manage to have Sphinx create useful API documentation. It's inconvenient that we need a hook for this, I'd have imagined we could generate them while settting up the channels, but what's fair is fair. shrug
> 
> How do the channel docstrings look outside of Sphinx? E.g. if you are using IPython, and do `myinstr.ch_1?` ?
> 
> Finally, what I didn't grasp -- do channels created with the new `ChannelCreator` also get added to a `self.channels` collection?

Agreed it isn't ideal to have sphinx-specific workaround for documentation, but it does work.

With the [removal](https://github.com/pymeasure/pymeasure/pull/931/commits/411200a7bb8d31254f51935193447922b6607902#diff-89442c0b4f1f772172dfbd90191347b509c116e99592cf33bc859da65e5e6466L143-L158)  of docstring argument the channel docstrings look like this:

Uninitialized class shows `ChannelCreator` docs:
image

Initialized class shows the passed channel docs:
image

Yes, channels created with new `ChannelCreator` are added to `self.channels` collection by default.

A user could pass the `collection="custom"` as a keyword argument to `ChannelCreator` which would be passed in `kwargs` to `add_child`, which would add it to a custom collection dictionary.

pymeasure/instruments/common_base.py
mcdo0486: That was an IDE auto-format setting, found it and turned it off. Fixed in last commit

pymeasure/instruments/common_base.py
mcdo0486: Actually those static methods are part of `Instrument`, which makes sense because we want to return all channels created by both `ChannelCreator` and `MultiChannelCreator` for an instrument. Do you think the docstrings should be clarified?

bilderbuchi: Yes, please, sounds good.

pymeasure/instruments/siglenttechnologies/siglent_spd1168x.py
```diff
@@ -30,7 +30,7 @@ class SPD1168X(SPDSingleChannelBase):
     """Represent the Siglent SPD1168X Power Supply.
     """
 
-    channels = Instrument.ChannelCreator(SPDChannel, 1)
+    ch_1 = Instrument.ChannelCreator(SPDChannel, 1)
```
bmoneke, line 33: The behaviour is the same, isn't it?

pymeasure/instruments/activetechnologies/AWG401x.py
```diff
@@ -430,7 +430,8 @@ class AWG401x_AFG(AWG401x_base):
         print(wfg.check_errors())       # Get the error queue
 
     """
-    ch = Instrument.ChannelCreator(ChannelAFG, (1, 2))
+    ch_1 = Instrument.ChannelCreator(ChannelAFG, 1)
```
bmoneke, line 433: This change will break the instrument, doesn't it?
If someone used "ch[1]", it won't work anymore.

Please keep it compatible, even if it breaks default pymeasure usage.
Maybe a workaround is "self.ch = self.channels" at the end of the "init".

tests/instruments/tcpowerconversion/test_cxn.py
```diff
@@ -130,4 +130,4 @@ def test_load_capacity_preset(channel):
             (cmd, b"\x2a" + response),
         ],
     ) as inst:
-        assert inst.presets[channel].load_capacity == 50
+        assert inst.channels[channel].load_capacity == 50
```
bmoneke, line 133: Your changes are good, but we have to take care, that we do not break the instruments for the users, probably the original author uses the driver.

bmoneke (commented):
Thanks for your work, a few comments from my side.

We have to take care with breaking changes (renaming channel collections), because the code base has to continue to work in the lab.

bilderbuchi:
> Thanks for your work, a few comments from my side.
> 
> We have to take care with breaking changes (renaming channel collections), because the code base has to continue to work in the lab.

Wait, the Channels feature (intentionally) has not been part of a release, yet! So I don't think we need to "take care with breaking changes", if they have only been on master so far? That was the idea behind merging this _after_ the release, to give us more time to find & eliminate issues like this. If we start doing a deprecation process for all things that landed on master, we generate a lot of effort and needed care. I think we are already quite careful with deprecations (of released functionality), no?

If people are on master, not a release, we are very grateful for the additional testing! I think an expectation to study changes when pulling updates can be expected, no?

pymeasure/instruments/common_base.py
```diff
@@ -131,55 +131,97 @@ def __init__(self, preprocess_reply=None, **kwargs):
...
+    class BaseChannelCreator:
+        """Base class for ChannelCreator and MultiChannelCreator.
+
+        :param cls: Class for all children or tuple/list of classes, one for each child.
+        :param \\**kwargs: Keyword arguments for all children.
+        """
+
+        def __init__(self, cls, **kwargs):
+            try:
+                self.valid_class = issubclass(cls, CommonBase)
```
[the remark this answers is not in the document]
mcdo0486: That would create a circular import, if `from pymeasure.instruments.channel import Channel` was added, unless we would use a workaround. Also, maybe someone needs to create a different `Channel` class for some reason that is still a subclass of `CommonBase`.

pymeasure/instruments/activetechnologies/AWG401x.py
mcdo0486: Yes, when fixing the tests I had the same thought. If the author wrote the tests with the collection, they probably are using it in the procedure. I have a further comment down below on this with regards to breaking changes.

tests/instruments/tcpowerconversion/test_cxn.py
mcdo0486: Same as above re: breaking changes

docs/dev/adding_instruments/channels.rst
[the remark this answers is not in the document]
mcdo0486: Ok, I've pushed a commit with changes to [channels.rst ](
https://github.com/mcdo0486/pymeasure/blob/566ec03d56cd0c1fa9357f47f824976ff15c48a1/docs/dev/adding_instruments/channels.rst)

I included the, somewhat arbitrary, 16 channel cutoff for when people should use `ChannelCreator` vs `MultiChannelCreator`

pymeasure/instruments/common_base.py
mcdo0486: Redid those doc strings in latest commits

docs/dev/adding_instruments/channels.rst
```diff
@@ -34,18 +34,24 @@ Please add the channel to the documentation. In the instrument's documentation f
...
-Without pressing reasons, call the dictionary :code:`channels` and do not change the default prefix in order to keep the code base homogeneous.
+Adding a channel with :class:`~pymeasure.instruments.common_base.CommonBase.ChannelCreator`
+*******************************************************************************************
 
-In order to add or remove programatically channels, use the parent's :meth:`~pymeasure.instruments.common_base.CommonBase.add_child`, :meth:`~pymeasure.instruments.common_base.CommonBase.remove_child` methods.
+For instruments with fewer than 16 channels the class :class:`~pymeasure.instruments.common_base.CommonBase.ChannelCreator`
+should be used to assign each channel interface to a class attribute. :class:`~pymeasure.instruments.common_base.CommonBase.ChannelCreator` constructor accepts two parameters,
+the channel class for this channel interface, and the instrument's channel id for the channel interface.
+
+In this example, we are defining a channel class and an instrument driver class. The ``VoltageChannel`` channel class
```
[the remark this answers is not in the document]
mcdo0486: Alright restructured in last commit

docs/dev/adding_instruments/channels.rst
```diff
@@ -22,10 +22,10 @@ Instruments with channels
...
 For this case, we have :class:`~pymeasure.instruments.Channel`, which is similar to :class:`~pymeasure.instruments.Instrument` and its property factories, but does expect an :class:`~pymeasure.instruments.Instrument` instance (i.e., a parent instrument) instead of an :class:`~pymeasure.adapters.Adapter` as parameter.
 All the channel communication is routed through the instrument's methods (`write`, `read`, etc.).
-However, :meth:`Channel.insert_id <pymeasure.instruments.Channel.insert_id>` uses `str.format` to insert the channel's id at any occurence of the class attribute :attr:`Channel.placeholder`, which defaults to :code:`"ch"`, in the written commands.
-For example :code:`"Ch{ch}:VOLT?"` will be sent as :code:`"Ch3:VOLT?"` to the device, if the channel's id is "3".
+However, :meth:`Channel.insert_id <pymeasure.instruments.Channel.insert_id>` uses ``str.format`` to insert the channel's id at any occurrence of the class attribute :attr:`Channel.placeholder`, which defaults to :code:`"ch"`, in the written commands.
+For example :code:`"Ch{ch}:VOLT?"` will be sent as :code:`"Ch3:VOLT?"` to the instrument, if the channel's id is "3".
```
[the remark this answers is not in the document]
mcdo0486: I put device back in that sentence.

docs/dev/adding_instruments/channels.rst
```diff
@@ -57,26 +60,111 @@ In order to add or remove programatically channels, use the parent's :meth:`~pym
...
-        inst.ch_A.voltage = 1.23
-        assert inst.ch_B.voltage == 4.56
+        inst.output_A.voltage = 1.23
+        assert inst.channels['B'].voltage == 4.56
+
+At instrument class instantiation, the instrument class will create an instance of the channel class and assign it to the class attribute name.
+Additionally the channels will be collected in a dictionary, by default named :code:`channels`.
+We can access the channel interface through that class name:
+
+.. code-block:: python
```
bilderbuchi, line 82: I think it would be good to get these examples under protocol tests, too, so we notice if our behaviour would not match the docs anymore. You can do this with `.. testcode:: with-protocol-tests` sections as demonstrated in the rest of the file.

bilderbuchi:
Except for one small thing looks good to me. @bmoneke what do you think, is this ready after that small fix?

docs/dev/adding_instruments/channels.rst
bmoneke: I had the same thought, looking at the code, but forgot to mention it.

bmoneke:
Yes, it is then ready to be merged.

docs/dev/adding_instruments/channels.rst
mcdo0486: Ok, I've added a protocol tests doctest that accomplishes the same thing as the python example.

bilderbuchi:
Thank you for your work!

PR #934, opened 2023-06-23
Regression: fix CSVFormatter.format not recording np.float types

mcdo0486:
This is my first regression from a PR I submitted, oops. Achievement unlocked?

The https://github.com/pymeasure/pymeasure/pull/796 PR decreased the number of type comparisons before a float is written to the results file, which speeds up recording of a procedure.

If the `value` is of a numeric type, then we should write it straight away. However, this PR used `type(value) in (float, int, Decimal)` to evaluate numeric-ness because `isinstance(True, int)` would evaluate to `True`, which it shouldn't. Boolean values are evaluated later, and for instance if current value's `DATA_COLUMN` has a Pint unit in it, then booleans should not be written to the results file.

However the `type(value) in (float, int, Decimal)` check would miss other numeric types like `np.float64`, causing `np.float64` to be evaluated as `'nan'`.

This PR fixes the issue by reverting back to `isinstance` but adding a check for boolean type.
`if isinstance(value, (float, int, Decimal)) and type(value) is not bool:`

While this adds a logical comparison, the number of comparisons is still fewer then what it was beforehand.

Added a test case for `np.float64`.

### DIFF
--- a/pymeasure/experiment/results.py
+++ b/pymeasure/experiment/results.py
@@ -121,8 +121,6 @@ def unique_filename(directory, prefix='DATA', suffix='', ext='csv',
 class CSVFormatter(logging.Formatter):
     """ Formatter of data results """
 
-    numeric_types = (float, int, Decimal)
-
     def __init__(self, columns, delimiter=','):
         """Creates a csv formatter for a given list of columns (=header).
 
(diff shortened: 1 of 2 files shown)


msmttchr (approved):
It seems ok to me. Thanks.

PR #936, opened 2023-06-28
fix MKS937B check_set_errors functions and improve tests

### LINKED ISSUE #935: recent changes break MKS937B check_set_errors
It was previously ok that `check_errors` returns `None` (or nothing) as for example in [MKS937B](https://github.com/pymeasure/pymeasure/blob/68f5487b0837e9f3c5debd144a7ddfeb7bddf3e5/pymeasure/instruments/mksinst/mks937b.py#L150).

Recently it got mandatory that the `check_set_errors` and `check_get_errors` return lists. This is correctly documented and some existing devices were also changed -> 1214d32.
`MKS937B` it seems was overlooked. 

https://github.com/pymeasure/pymeasure/blob/68f5487b0837e9f3c5debd144a7ddfeb7bddf3e5/pymeasure/instruments/mksinst/mks937b.py#L150-L162

Is the only change needed here to return an empty list upon no error?

dkriegner:
previously the test did not cover any property setter. Also the `check_set_errors` function did not read back the additional termination characters. `check_set_errors` now returns a list upon success as required.

This fixes #935

### DIFF
--- a/pymeasure/instruments/mksinst/mks937b.py
+++ b/pymeasure/instruments/mksinst/mks937b.py
@@ -155,7 +155,8 @@ class MKS937B(Instrument):
         reply = self._re_response.search(ret)
         if reply:
             if reply.group('ack') == 'ACK':
-                return
+                self._check_extra_termination()
+                return []
         # no valid acknowledgement message found
         raise ValueError(f"invalid reply '{ret}' found in check_errors")
 
--- a/tests/instruments/mksinst/test_mks937b.py
+++ b/tests/instruments/mksinst/test_mks937b.py
@@ -57,8 +57,18 @@ def test_ion_gauge_status_invalid_channel():
             inst.ch_2.ion_gauge_status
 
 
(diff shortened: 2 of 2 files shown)


pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -155,7 +155,8 @@ def check_set_errors(self):
         reply = self._re_response.search(ret)
         if reply:
             if reply.group('ack') == 'ACK':
-                return
+                self._check_extra_termination()
```
bmoneke, line 158: Nice, that you found that additional bug.

bmoneke:
Thanks for the fix. It looks good.
Did you test the change with a device, that the additional read does not break the device?

dkriegner:
I checked it with the hardware and the extra terminators are sent by the device as mimicked by the test code. It previously went unnoticed since the extra terminator characters were ignored by the regular expression parsing the reply.

Issue #757, opened 2022-11-22
Instruments announcements


bmoneke:
I have a few instruments in use, which I will publish, once we reduced the number of pull requests and urgent issues. However, in case someone needs these drivers, I'd be happy to open a PR and offer the code. Therefore, I mention these devices here:

- Ophir powermeters: Nova I, Nova II
- Aculight Argos 2400 OPO system
- Laserquantum (Novanta) Finesse laser system
- Thyracont VCR pressure sensor
- IPG Photonics YAR-15K fiber amplifier
- Yokogawa AQ6370 optical spectrum analyzer (very initial state of the driver)

JinchW:
Hi, I wonder if it is convenient to share the control program of YOKOGAWA AQ6370 optical spectrum analyzer now? I am a beginner of Pymeasure and want to get some examples about YOKOGAWA optical spectrum analyzer.

bmoneke:
I added a link to the corresponding branch in my repository.
I have to warn you, though, it is a very initial version and a year old. It does not use the new Channels functionality, so it is not a good example of the newest pymeasure features. Nevertheless, it is a good starting point (just add channels for the traces).

If you need a working and tested code, you can look at instrumentkit: https://github.com/instrumentkit/InstrumentKit/blob/main/src/instruments/yokogawa/yokogawa6370.py

BenediktBurger:
Hey @JinchW , someone else made a good pymeasure driver for the Yokogawa AQ6370. See #1059 .

PR #937, opened 2023-07-05
Add Instrument: Thyracont SmartlineV1 vacuum gauge

dkriegner:
Adds Instrument of a new vendor. A vacuum gauge which communicates via RS485. 

The protocol is described here in detail:
https://wiki.kern.phys.au.dk/Interface_protokoll_Thyracont_eng5.pdf

Note 1: This Instrument covers only older models of this Vendor. New models need to be implemented separately since the protocol differs. (I will do this in the upcoming month in a separate PR.)

Note 2: It seems not all older models follow the protocol outlined in the linked pdf for all commands. I have one where the `unit` property does not work as advertised. Its similar for a lot of other advertised properties which I therefore did not implement since I can't test them. The tests for `unit` implement the behavior outlined in the documentation.


bmoneke:
Interesting, I have another transmitter (VCR) by Thyracont, see #757 , by heart, I do not know, whether the protocol is the same.
At least it uses RS485 as well.

dkriegner:
If its made before ~2017 it might be the same. Newer ones are different,
but both use RS485. Please let me know. I would be very interested in the
code for the newer devices since I did nit code that one yet.

bmoneke:
I added it as a (draft) PR: #940 
The code is working (I've been using it in the lab for more than one year), but requires some cleanup (I started to use Channels after they became available and I started to use Enums).

pymeasure/instruments/thyracont/smartline_v1.py
```diff
@@ -0,0 +1,144 @@
...
+    """Calculate a 1 bytes checksum mapped to a printable ASCII character.
+
+    The checksum is calculated by the sum of the decimal values of each message
+    character modulo 64 + 64.
+
+    :param string msg: message content
+    :returns: calculated checksum
+    :rtype: string
+    """
+    return chr(sum(map(ord, msg)) % 64 + 64)
```
bmoneke, line 39: That is dense (but nice)

pymeasure/instruments/__init__.py
bmoneke, line 1: Don't forget the API documentation links.

pymeasure/instruments/thyracont/smartline_v1.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def check_errors(self):
+        reply = self.read()
+        if len(reply) < 4:
+            raise ValueError(f"Reply of instrument ('{reply}') too short.")
+        if reply[3] in ["N", "X"]:
+            raise ValueError(f"Reply from Instrument indicates an error '{reply}'")
+        return []
+
+    type = Instrument.measurement(
```
bmoneke, line 110: Make it better verbose:
```suggestion
    device_type = Instrument.measurement(
```

pymeasure/instruments/thyracont/smartline_v1.py
```diff
@@ -0,0 +1,144 @@
...
+        "T",
+        """Get the device type.""",
+        cast=str,
+        preprocess_reply=lambda s: s[4:],
+    )
+
+    pressure = Instrument.measurement(
+        "M",
+        """Get the pressure measurement in the set unit.""",
```
bmoneke, line 119: Really in the set unit?
The v2 version always returns mbar. You can only set the display unit, but not the data return unit.

pymeasure/instruments/thyracont/smartline_v1.py
bmoneke, line 39: We could move it into some other file (init? @bilderbuchi ) for use of both instruments.
On the other hand, the method is not too large, such that it can remain in both instruments.

I think a couple of tests would be good.

bmoneke (changes requested):
A few comments, but the code looks good.

pymeasure/instruments/__init__.py
dkriegner: do you want them in the docstring or as comment? The problem is that the vendor does not provide this file anymore. (although they claim to do, but it just links to the new documentation)

pymeasure/instruments/thyracont/smartline_v1.py
dkriegner: you are right, the docstring is wrong. In my model the unit property does not work and the old transmitters can't get a firmware upgrade. But the documentation is clear that its always mbar (as in the new version). I fix the docstring

pymeasure/instruments/thyracont/smartline_v1.py
bmoneke: I like your short version. It was a comment, not a request to change.

Yes, maybe it is good to split it, to increase readability.

pymeasure/instruments/__init__.py
bmoneke: I meant the files in the docs folder: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation

bmoneke:
#940 is merged, thanks to the review of @msmttchr

pymeasure/instruments/__init__.py
dkriegner: Thanks for the reminder. (I should already know that)

pymeasure/instruments/thyracont/smartline_v1.py
dkriegner: I added some tests to check the `calculate_checksum` functionality. Especially since we now have one function which is tested I think even though its a short method it makes sense to reuse it in both v1 and v2 files.

dkriegner:
I hope the recent commits address the made comments. Let me know how you want to handle the `calculate_checksum `function. It might be cleaner to put it in a separate file as you indicated.

pymeasure/instruments/thyracont/smartline_v1.py
bmoneke: Let's keep it in v1, which is a small file (with only a few dependencies) anyways.

bmoneke:
Thanks for your work. From my side, it can be merged. Are you ready to merge?

dkriegner:
From my side its ready to merge. I just still changed rs485 to RS485 in the V2 file, hope this was ok for you. Thanks great feedback and happy to see that we are not the only ones using devices from this vendor.

Issue #913, opened 2023-05-16
Attocube Anc300 uses properties instead of method


bmoneke:
The following two properties should be methods instead, I think
~~~python
    stepu = Instrument.setting(
        "stepu %d",
        """Set the steps upwards for N steps. Mode must be 'stp' and N must be
        positive.""",
        validator=strict_range, values=[0, inf], check_set_errors=True)

    stepd = Instrument.setting(
        "stepd %d",
        """Set the steps downwards for N steps. Mode must be 'stp' and N must be
        positive.""",
        validator=strict_range, values=[0, inf], check_set_errors=True)
~~~

This requires a deprecation etc., therefore is not done in #883

bilderbuchi:
I agree, from what the docstring says, this could probably be a method `step(nr_steps)` or so. One interface/method, hides the detail that the commands are different, which the user does not care about.

bmoneke:
In fact, there is such a method "steps" which uses these properties...

I guess the initial author implemented each command as a property.

bilderbuchi:
Oh, might even be that I reviewed that 👼 😁 
So, close this?

dkriegner:
I am the original author. I think this was my first Instrument and it might be I got too overenthusiastic with the properties

Just removing these properties however seems not like the best option. The `move()` method is not a real replacement of these two properties since move does a bit more. So to keep the flexibility for the user move would need to gain an option to be non-blocking. I can propose something if thats welcome.

bmoneke:
Improvements are always welcome.
If you deem another name better, I'm open for suggestions. Maybe `move_raw` or something.

bilderbuchi:
Seems to be closed by #938

PR #938, opened 2023-07-05
Attocube ANC300: use method instead of property.

dkriegner:
Addressed #913 

I try to fix the mentioned issue by implementing a new method `move_raw` and deprecating the `stepu` / `stepd` properties. To not loose any functionality I introduced a new property which allows to set up a continuous movement in either up or down direction. This was previously possible by `ANC300.<axis>.stepu = 0`. I look forward for your opinion if this should also be a method or can be a property. 

I also wonder if the Deprecation warnings triggered by the tests should be filtered out?


bmoneke:
If you add `with pymeasure.warns(FutureWarning):`, you both suppress the warning **and** make sure, that the warning is raised!

pymeasure/instruments/attocube/anc300.py
```diff
@@ -129,17 +135,40 @@ class Axis(Channel):
...
+    )
+
     stepu = Instrument.setting(
         "stepu %d",
         """Set the steps upwards for N steps. Mode must be 'stp' and N must be
-        positive.""",
-        validator=strict_range, values=[0, inf], check_set_errors=True)
+        positive. 0 causes a continous movement until stop is called.
+
+        .. deprecated:: Use attr:`continuous_move` or meth:`move_raw` instead.
```
bmoneke, line 154: Please add the version, since when it is deprecated, see https://github.com/pymeasure/pymeasure/issues/672#issuecomment-1371900222
```suggestion
        .. deprecated:: 0.13.0 Use attr:`continuous_move` or meth:`move_raw` instead.
```

pymeasure/instruments/attocube/anc300.py
```diff
@@ -129,17 +135,40 @@ class Axis(Channel):
...
+        "%s",
+        """Set a continous movement in 'up' or 'down' direction. Mode must be
+        'stp'. The axis can be halted by the stop method.
+        """,
+        validator=strict_discrete_set,
+        values={"up": "stepu c", "down": "stepd c"},
+        map_values=True,
+        check_set_errors=True,
+    )
+
```
bmoneke, line 148: I'd make it two methods (`move_continuously_up`, `move_continuously_down`) or add it to the `move` method, if the steps are +/- inf.

bmoneke (commented):
Some ideas how to implement it.

pymeasure/instruments/attocube/anc300.py
dkriegner: I like the idea with the +/-inf. It seems unnecessary to have extra methods for this. I'll look into it.

bmoneke:
I just thought about it and noticed, that you cannot block until infinity steps are finished.

Another idea: Have a method "start_moving", which does not block and your "move" method uses the other one.

"start_moving" could support unlimited steps.

Maybe 0 steps could indicate to stop. 

Am 6. Juli 2023 21:31:19 MESZ schrieb Dominik Kriegner ***@***.***>:

dkriegner:
But the non blocking move_raw is basically already what we need with +/-
inf steps. One just has to exclude these argument values in move then.

pymeasure/instruments/attocube/anc300.py
dkriegner: implemented with the other changes below

pymeasure/instruments/attocube/anc300.py
```diff
@@ -132,14 +138,26 @@ class Axis(Channel):
     stepu = Instrument.setting(
         "stepu %d",
         """Set the steps upwards for N steps. Mode must be 'stp' and N must be
-        positive.""",
-        validator=strict_range, values=[0, inf], check_set_errors=True)
+        positive. 0 causes a continous movement until stop is called.
+
+        .. deprecated:: 0.13.0 Use attr:`continuous_move` or meth:`move_raw` instead.
```
bmoneke, line 143: Is `continuous_move` still present? (Similarly below)
```suggestion
        .. deprecated:: 0.13.0 Use meth:`move_raw` instead.
```

pymeasure/instruments/attocube/anc300.py
```diff
@@ -156,26 +174,48 @@ def stop(self):
...
         self.check_set_errors()
 
+    def move_raw(self, steps):
+        """Move 'steps' steps in the direction given by the sign of the
+        argument. This method assumes the mode of the axis is set to 'stp' and
+        it is non-blocking, i.e. it will return immediately after sending the
+        command.
+
+        :param steps: integer value of steps to be performed. A positive
+            sign corresponds to upwards steps, a negative sign to downwards
```
bmoneke, line 184: A user of the ANC300 does know, what "upwards" and "downwards" is?

bmoneke (approved):
A tiny request.

pymeasure/instruments/attocube/anc300.py
dkriegner: I checked the manual and actually the wording used in docstrings matches the Attocube manual. They also write upwards and downwards to specify directions. (This is also the reason the commands are "stepu" and "stepd").

Since I have no rotation stage accessible, and the manual does not mention the rotation sense I would prefer to keep it as is.

bmoneke: Ok. If it is in the manual, a user will understand it, so my original question is answered.

bmoneke:
Is there anything left, you want to include into this PR or may I merge?

dkriegner:
It can be merged from my side since #913 seems addressed. thanks for the helpful comments!

PR #940, opened 2023-07-06
New Thyracont Transmitter (v2 protocol)

bmoneke:
Already announced in #757 
See also #937 for another Thyracont device (v1)

Be aware, that I started to change to Enums and to Channels, such that a few elements are duplicated.

I did not finish cleaning it yet, to push it, but to start cooperation with @dkriegner, I push this not yet cleaned code.

The code works in the lab for quite some time.

### DIFF
--- a/pymeasure/instruments/thyracont/transmitter.py
+++ b/pymeasure/instruments/thyracont/transmitter.py
@@ -1,0 +1,491 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2023 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from enum import IntEnum
+
+from pymeasure.instruments import Instrument, Channel, validators
+from pyvisa.constants import Parity, StopBits
+
+import pyvisa as pv
+from pyvisa import VisaIOError  # noqa: F401 for dependencies
+
+
+def calculateChecksum(message):
+    """Calculate the checksum for string `message`."""
+    value = 0
+    for i in range(len(message)):
+        value += ord(message[i])
+    return chr(value % 64 + 64)
+
+
+def compose_data(value):
+    """Generate a string with the length of `value` and the `value` itself afterwards.
+
+    :param value: Value to send to the device.
+    """
+    value = f"{value}"
+    return f"{len(value):02}{value}"
+
+
+source = {
+    'combination': 0,
+    'pirani': 1,
+    'piezo': 2,
+    'hotCathode': 3,
+    'coldCathode': 4,
+    'ambient': 6,
+    'relative': 7,
+}
+
+
+class Sources(IntEnum):
+    COMBINATION = 0
+    PIRANI = 1
+    PIEZO = 2
+    HOT_CATHODE = 3
+    COLD_CATHODE = 4
+    AMBIENT = 6
+    RELATIVE = 7
+
+
+gas_factor = Channel.control(
+    "0C{ch}00", "0C{ch}%s", "Control the gas correction factor.",
+    values=(0.2, 8),
+    validator=validators.strict_range,
+    set_process=compose_data,
+)
+
+
+class SensorChannel(Channel):
+    """Generic channel for individual pressure sensors of a Transmitter."""
+    _id = -1  # obligatory channel number, define in channel types
+
+    def __init__(self, parent, id):
+        if id is None:
+            id = self._id
+        assert id == self._id, f"Pirani ID has to be {self._id}."
+        super().__init__(parent, id=self._id)
+
+    pressure = Channel.measurement(
+        "0M{ch}00", """Get the current pressure in mbar.""",
+        preprocess_reply=lambda r: r.replace("UR", "0").replace("OR", "inf"),
+    )
+
+
+class Pirani(SensorChannel):
+    """Pirani sensor channel.
+
+    A Pirani sensor measures the heat transfer in the gas. The reading depends on the gas type.
+    The sensor reading is linear to the real pressure below some threshold, around 1 mbar.
+    You may control a gas factor with :attr:`gas_factor`.
+    """
+
+    _id = 1
+
+    gas_factor = gas_factor
+
+    statistics = Channel.measurement(
+        "0PM011",
+        """Get the sensor statistics: wear in percent (negative: corrosion,
+            positive: contamination), time since last adjustment in hours.""",
+        separator="A",
+        cast=int,
+        get_process=lambda vals: [v / 4 for v in vals],
+    )
+
+
+class Piezo(SensorChannel):
+    """Piezo sensor channel. A piezo sensor is independent of the gas present."""
+
+    _id = 2
+
+
+class HotCathode(SensorChannel):
+    """Hot cathode sensor channel."""
+
+    _id = 3
+
+    filament_mode = Instrument.control(
+        "0FC00", "2FC01%i", """The hot cathode filament control setting.""",
+        values={"2 if 1 defect": 0,
+                "Filament1": 1,
+                "Filament2": 2,
+                "toggle>1mbar": 3},
+        validator=validators.strict_discrete_set,
+        map_values=True,
+        cast=int,
+        check_set_errors=True,
+    )
+
+    active_filament = Instrument.measurement(
+        "0FN00", "Get the current filament number."
+    )
+
+    filament_status = Instrument.measurement(
+        "0FS00", """Get the status of the hot cathode filaments.""",
+        values=["Filament 1 and 2 ok",
+                "Filament 1 defective",
+                "Filament 2 defective",
+                "Filament 1 and 2 defective"],
+        map_values=True,
+    )
+
+    gas_factor = gas_factor
+
+    statistics = Channel.measurement(
+        "0PM013",
+        """Get the wear in percent: of filament 1, of filament 2.""",
+        separator="A",
+        cast=int,
+        get_process=lambda vals: [v / 4 for v in vals],
+    )
+
+    # cathode status CA, cathode control CC, cathode control mode CM
+
+
+class ColdCathode(SensorChannel):
+    """Cold cathode sensor channel."""
+
+    _id = 4
+
+    gas_factor = gas_factor
+
+
+class Ambient(SensorChannel):
+
+    _id = 6
+
+
+class Relative(SensorChannel):
+
+    _id = 7
+
+
+class Transmitter(Instrument):
+    """A Thyracont vacuum sensor transmitter.
+
+    Communication Protocol v2 via rs485.
+
+    :param adress: The device address in the range 1-16.
+    """
+
+    source = source
+
+    Sources = Sources
+
+    errors = {'NO_DEF': "Invalid command for this device.",
+              '_LOGIC': "Access Code is invalid or illogical command.",
+              '_RANGE': "Value sent is out of range.",
+              'ERROR1': "Sensor defect or stacked out.",
+              'SYNTAX': "Wrong syntax or mode in data is invalid for this device.",
+              'LENGTH': "Length of data is out of expected range.",
+              '_CD_RE': "Calibration Data Read Error.",
+              '_EP_RE': "EEPROM Read Error.",
+              '_UNSUP': "Unsupported Data for that command.",
+              '_SEDIS': "Sensor element disabled."}
+
+    def __init__(self, adapter, name="Thyracont", baud_rate=115200,
+                 address=1, timeout=250, **kwargs):
+        super().__init__(adapter, name=name, includeSCPI=False,
+                         write_termination="\r",
+                         read_termination="\r",
+                         timeout=timeout,
+                         asrl={'baud_rate': baud_rate,
+                               'parity': Parity.none,
+                               'stop_bits': StopBits.one,
+                               },
+                         **kwargs
+                         )
+        self.address = address  # 1-16
+        """
+        Everything is sent as ASCII characters.
+        Package (bytes and usage): 0-2 address, 3 access code, 4-5 command, 6-7
+            data length.
+            if data: 8-n data to be sent, n+1 checksum, n+2 carriage return
+            if no data: 8 checksum, 9 carriage return
+        Access code for read access: 0 (master->transmitter), 1 response
+            write: 2, 3
+            factory default: 4,5
+            error: -, 7
+            binary 8, 9
+        Data length is number of data in bytes (padding with zeroes on left)
+        Checksum: Add the decimal numbers of the characters before, mod 64, add
+        64, show as ASCII.
+        """
+
+    def write(self, command):
+        """Write a command to the device."""
+        message = f"{self.address:03}{command}"
+        super().write(f"{message}{calculateChecksum(message)}")
+
+    def write_composition(self, accessCode, command, data=""):
+        """Write a command with an accessCode and optional data to the device.
+
+        :param accessCode: How to access the device.
+        :param command: Command to send to the device.
+        :param data: Data for the command.
+        """
+        self.write(f"{accessCode}{command}{compose_data(data)}")
+
+    def ask(self, command_message, query_delay=0):
+        """Ask for some value and check that the response matches the original command.
+
+        :param str command_message: Access code, command, length, and content.
+            The command sent is compared to the response command.
+        """
+        self.write(command_message)
+        self.wait_for(query_delay)
+        return self.read(command_message[1:3])
+
+    def read(self, command=None):
+        """Read from the device and do error checking.
+
+        :param str command: Original command sent to the device to compare it with the response.
+            None deactivates the check.
+        """
+        # Sometimes the answer contains 0x00 or values above 127, such that
+        # decoding fails.
+        answer = b''
+        i = 0
+        while True:
+            try:
+                got = self.adapter.read_bytes(1)
+            except pv.VisaIOError as exc:
+                if exc.abbreviation == 'VI_ERROR_TMO':
+                    raise ConnectionResetError("Timeout.")
+                i += 1
+                if i > 2:
+                    raise ConnectionResetError(f"Timeout at {i} tries.")
+                continue
+            if got == b'\r':
+                break
+            elif int.from_bytes(got, 'little') > 127:
+                pass  # Ignore those values.
+            elif got == b'\x00':
+                pass  # Ignore 0 byte value.
+            else:
+                answer += got
+        got = answer.decode('ascii')
+        # Raises from time to time an UnicodeDecodeError or pyvisa.VisaIOError
+        if got[3] == "7":
+            raise ConnectionError(self.errors[got[8:-1]])
+        if command is not None and got[4:6] != command:
+            raise ConnectionError(f"Wrong response to {command}: '{got}'.")
+        if calculateChecksum(got[:-1]) != got[-1]:
+            raise ConnectionError("Response checksum is wrong.")
+        return got[8:-1]
+
+    def check_set_errors(self):
+        """Check the errors after setting a property."""
+        self.read()
+        return []  # no error happened
+
+    def check_errors(self):
+        # TODO remove after update to pymeasure 0.12
+        return self.check_set_errors()
+
+    def query(self, accessCode, command, data=""):
+        """
+        Send a message to the transmitter and return its answer.
+
+        :param accessCode: How to access the device.
+        :param command: Command to send to the device.
+        :param data: Data for the command.
+        :return str: Response from the device after error checking.
+        """
+        self.write_composition(accessCode, command, data)
+        return self.read(command)
+
+    " Main commands"
+
+    range = Instrument.measurement(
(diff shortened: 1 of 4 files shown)


pymeasure/instruments/thyracont/transmitter.py
```diff
@@ -0,0 +1,491 @@
...
+        values={"2 if 1 defect": 0,
+                "Filament1": 1,
+                "Filament2": 2,
+                "toggle>1mbar": 3},
+        validator=validators.strict_discrete_set,
+        map_values=True,
+        cast=int,
+        check_set_errors=True,
+    )
+
```
dkriegner, line 140:
```suggestion

    degas = Instrument.control(
        "0DG00", "2DG01%i",
        """Control the degas mode.""",
        values={True: 1, False: 0},
        map_values=True,
        validator=validators.strict_discrete_set,
        check_set_errors=True,
    )

    enabled = Instrument.control(
        "0CC00", "2CC01%i",
        """Control the state of the cathode.""",
        values={True: 1, False: 0},
        map_values=True,
        validator=validators.strict_discrete_set,
        check_set_errors=True,
    )
```

pymeasure/instruments/thyracont/transmitter.py
```diff
@@ -0,0 +1,491 @@
...
+    active_filament = Instrument.measurement(
+        "0FN00", "Get the current filament number."
```
dkriegner, line 142:
```suggestion
        "0FN00", "Get the current filament number.",
        cast=int,
```

pymeasure/instruments/thyracont/transmitter.py
```diff
@@ -0,0 +1,491 @@
...
+        values=["Filament 1 and 2 ok",
+                "Filament 1 defective",
+                "Filament 2 defective",
+                "Filament 1 and 2 defective"],
+        map_values=True,
+    )
+
+    gas_factor = gas_factor
+
+    statistics = Channel.measurement(
```
dkriegner, line 156: Note the code here is copy pasted from the Pirani channel but the manual says that the return value is different. For my VSH model this command unfortunately does not work at all. If I send it I get a 
`ConnectionError: Unsupported Data for that command.` (Could be an issue with my model).

pymeasure/instruments/thyracont/transmitter.py
```diff
@@ -0,0 +1,491 @@
...
+        get_process=lambda vals: vals / 4 if isinstance(vals, int) else [v / 4 for v in vals],
+        # TODO simplify once #740 is merged.
+    )
+
+
+class VSR(Transmitter):
+    """Vacuum transmitter of VSR/VCR series with both a piezo and a pirani sensor."""
+
+    piezo = Instrument.ChannelCreator(Piezo, prefix=None)
+    pirani = Instrument.ChannelCreator(Pirani, prefix=None)
```
dkriegner, line 491:
```suggestion
    pirani = Instrument.ChannelCreator(Pirani, prefix=None)


class VSH(Transmitter):
    """Vacuum transmitter of VSH series with both a pirani and a hot cathode sensor."""

    pirani = Instrument.ChannelCreator(Pirani, prefix=None)
    hotcathode = Instrument.ChannelCreator(HotCathode, prefix=None)
```

dkriegner (commented):
thanks for sharing this code. I can confirm it mostly works. 
I have a VSH-type transmitter with a Pirani and hot cathode sensor. I added few suggestions for additions and comments to the code derived from my tests with my hardware.
What I found does not work are the Pirani.statistics and HotCathode.statistics properties. 

But I guess the bigger question is how to handle the V1 (#937) and V2 devices. 
The checksum calculation for example seems to be the same, but the rest so different (V1 protocol is very limited) that likely its ok to keep two different implementations?

pymeasure/instruments/thyracont/transmitter.py
bmoneke: The statistics command is a more recent addition (I year ago?).
I read about it in a blog entry.
You have to update your firmware to get that feature.
The wear is relative to the last calibration **after** the firmware update (I'd say by comparing the calibration data 😁 ).

For the firmware update you can use their software: https://vacugraph.com/

Attention: The firmware update might change the baud_rate (in my case from 9600 to 115200), I had to use a port sniffer to detect that change, as the official software continued to work.

bmoneke:
Thanks for your comments, especially regarding the hot cathode, as I do not have such a device.

> But I guess the bigger question is how to handle the V1 (https://github.com/pymeasure/pymeasure/pull/937) and V2 devices.

If they are different, then we should write two different drivers.

Maybe we could put the checksum calculation in one file and import it in the other one.

pymeasure/instruments/thyracont/transmitter.py
```diff
@@ -0,0 +1,491 @@
...
+
+    gas_factor = gas_factor
+
+    statistics = Channel.measurement(
+        "0PM011",
+        """Get the sensor statistics: wear in percent (negative: corrosion,
+            positive: contamination), time since last adjustment in hours.""",
+        separator="A",
+        cast=int,
+        get_process=lambda vals: [v / 4 for v in vals],
```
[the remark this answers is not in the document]
dkriegner: Let me add that my device seems to support this command and that here really the processing of the reply seems buggy. As far as I see from the response (consistent with the documentation) there are some letters in the reply which separate the different values which somehow will have to be cut away.

pymeasure/instruments/thyracont/transmitter.py
dkriegner: thanks for the hint. It seems I have the latest firmware already. but the hot cathode (in contrast to what is written in the documentation) seems to not support this command. Also the Vacugraph software shows the wear data only for the Pirani channel. For me this feature is not so crucial.

pymeasure/instruments/thyracont/transmitter.py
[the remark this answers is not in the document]
bmoneke: Please give me the response of the device, such that we can use it to write a test.

pymeasure/instruments/thyracont/transmitter.py
bmoneke: I changed it to 'sensor_enabled' in order to comply with pymeasure style.

dkriegner:
I will refactor the smartline_v1 code to make it possible to import the checksum function from there. From my point of view the `transmitter.py` file could be renamed to `smartline_v2.py`.

pymeasure/instruments/thyracont/transmitter.py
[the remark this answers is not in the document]
dkriegner:
```
import pymeasure.instruments
vsh = pymeasure.instruments.thyracont.VSH("...")
vsh.write("0PM011")  # message transmitted "0010PM011p"
vsh.read()  # raw return b'0011PM06W0A382j'
# returns 'W0A382'
```
This corresponds to ~95 hours since zero adjust and 0% wear.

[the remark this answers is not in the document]
dkriegner:
```suggestion
        preprocess_reply=lambda msg: msg.strip("W"),
        get_process=lambda vals: [vals[0], vals[1]/4],
```

This would work for me...

bmoneke:
> renamed to smartline_v2.py.

I agree, thanks for that suggestion.

pymeasure/instruments/thyracont/transmitter.py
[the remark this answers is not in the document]
bmoneke: Thanks for that communication example. I just verified it with my device.

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,525 @@
...
+            - F[float]T[float]
+                switch between low and high value.
+            - D[float]
+                switch at value.
+        """
+        got = self.ask_manually(0, "ST")
+        # VSR/VSL: 0 direct switch at 1 mbar, 1 continuios between 5 to 15 mbar
+        if got == "0":
+            return "direct"
+        elif got == "1":
```
dkriegner, line 422:
```suggestion
        elif got in ["1", "2"]:
```

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,525 @@
...
+        Set the sensor transition to `mode` with optional `values`.
+
+        :param mode: "continuous", "direct"
+            Switch continuously from one sensor to another in the interval of
+            low and high. They can be set in values and default to 5, 15 for
+            VCR. Alternatively switch directly at a single piezo pressure.
+        :param values: The lower transition value and the higher transition value, if applicable.
+        """
+        if mode == "continuous":
+            if not values:
```
dkriegner, line 453: For VSH the command could also be "2".

dkriegner (commented):
Few more small comments from my side. I propose to merge this before the V1 code. I will make some adjustments in the V1 than after this is merged.

tests/instruments/thyracont/test_smartline_v2.py
```diff
@@ -0,0 +1,265 @@
...
+# Piezo tests
+def test_pressure_piezo():
+    with expected_protocol(SmartlineV2, [("0010M200`", "0011M2079.734e2D\r")]) as inst:
+        assert inst.piezo.pressure == 973.4
+
+
+# Hot Cathode tests
+def test_pressure_hot_cathode():
+    with expected_protocol(SmartlineV2, [("0010M300a", "0011M3079.734e2E\r")]) as inst:
+        assert inst.hot_cathode.pressure == 973.4
```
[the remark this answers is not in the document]
bmoneke: I guess I used the same value for all pressure readings, but you're rigth.

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: I'm thinking about removing that method alltogether.

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: then you do not know, what the transition mode is.
I'll name it "continuous mode 2" or such.

pymeasure/instruments/thyracont/transmitter.py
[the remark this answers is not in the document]
bmoneke: done.

pymeasure/instruments/thyracont/smartline_v2.py
dkriegner: fine for me. I would likely never use it. the methods above provide good alternatives

bmoneke:
> Few more small comments from my side. I propose to merge this before the V1 code. I will make some adjustments in the V1 than after this is merged.

I requested a review, but I do not know, when that will happen. I could merge your v1 code before, though.

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,505 @@
...
+from enum import IntEnum
+
+from pymeasure.instruments import Instrument, Channel, validators
+from pyvisa.constants import Parity, StopBits
+
+import pyvisa as pv
+from pyvisa import VisaIOError  # noqa: F401 for dependencies
+
+
+def calculateChecksum(message):
```
msmttchr, line 34:
```suggestion
def calculate_checksum(message):
```
To make style uniform?

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,505 @@
...
+
+class Sources(IntEnum):
+    COMBINATION = 0
+    PIRANI = 1
+    PIEZO = 2
+    HOT_CATHODE = 3
+    COLD_CATHODE = 4
+    AMBIENT = 6
+    RELATIVE = 7
+
```
msmttchr, line 70: I see a repetition between `class Sources` and variable `source`. Could this be avoided ?

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,505 @@
...
+class Pirani(SensorChannel):
+    """Pirani sensor channel.
+
+    A Pirani sensor measures the heat transfer in the gas. The reading depends on the gas type.
+    The sensor reading is linear to the real pressure below some threshold, around 1 mbar.
+    You may control a gas factor with :attr:`gas_factor`.
+    """
+
+    _id = 1
```
msmttchr, line 105:
```suggestion
    _id = source ['pirani'] # or use enum ?
```
You defined symbolic names above and here you use the literal number. I don't understand or I missed your point. Is it intentional?

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,505 @@
...
+                "Filament 2 defective",
+                "Filament 1 and 2 defective"],
+        map_values=True,
+    )
+
+    gas_factor = gas_factor
+
+    statistics = Channel.measurement(
+        "0PM013",
+        """Get the wear in percent: of filament 1, of filament 2.""",
```
msmttchr, line 181:
```suggestion
        """Get the filaments wear level in percent. Return a tuple of two values (wear of filament 1,  wear of filament 2).""",
```
I am not sure whether this is better, just to make sure I understood your docstring. If correct, feel free to leave your comment as it is or change it

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,505 @@
...
+        :param command: Command to send to the device.
+        :param data: Data for the command.
+        :param int query_delay: Time to wait between writing and reading.
+        :return str: Response from the device after error checking.
+        """
+        self.write(f"{accessCode}{command}{compose_data(data)}")
+        self.wait_for(query_delay)
+        return self.read(command)
+
+    def read(self, command=None):
```
msmttchr, line 309: I think (not sure though), you can rewrite the `read` to call `super` reading up to termination character and then filter out unwanted values or use ignore errors in `decode`.

msmttchr (commented):
First pass of my comments. Thanks.

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: Oh yes, I forgot to harmonize that (it predates my pymeasure involvement).

bmoneke: Done.

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: Yes, I'll remove `source`.

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: You're right, that we have the enum. Thanks for hinting at it.

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: I improved it (also the other statistics docstring)

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: Thanks for the hint at `errors="ignore"`.
It does not solve the zero byte (b"\x00") problem, but that can be fixed with a simple replace.

In order to verify correct transition of the code, I installed a test.

bmoneke:
Hey @msmttchr thanks for your review. I'm so glad to be in a team, which helps each other writing good code.

I addressed all of your issues.

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,485 @@
...
+            None deactivates the check.
+        """
+        # Sometimes the answer contains 0x00 or values above 127, such that
+        # decoding fails.
+        response = self.read_bytes(-1, break_on_termchar=True)
+        if b"\x00" in response:
+            response = response.replace(b"\x00", b"")
+
+        # b"\r" is the termination character
+        got = response.rstrip(b"\r").decode('ascii', errors="ignore")
```
msmttchr, line 318:
```suggestion
        got = super().read().replace('\x00', '')
```
I propose this suggestion, it would do the conversion to string for you, but I am not sure what `pyvisa` does when a char > 127 is received. I have seen tests failing, for two reasons:

1. Errors in decoding are not ignored (https://github.com/pymeasure/pymeasure/blob/55e453cc33fe4aa2410389cc8bef779fc8d2222f/pymeasure/adapters/protocol.py#L135). Should be ignored ?
2. `ProtocolAdapter` does not implement `read_termination`. Is this correct ?

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,485 @@
...
+    display_data = Instrument.control(
+        "0DD00", "2DD01%i",
+        """Control the display data source (strict SOURCES).""",
+        values=Sources,
+        cast=str_to_source,
+        validator=validators.strict_discrete_set,
+        check_set_errors=True,
+    )
+
+    def setHigh(self, high=""):
```
msmttchr, line 374:
```suggestion
    def set_high(self, high=""):
```
Minor consistency change.

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,485 @@
...
+        """Set the high pressure to `high` pressure in mbar."""
+        self.ask_manually(2, "AH", high)
+
+    def setLow(self, low=""):
```
msmttchr, line 378:
```suggestion
    def set_low(self, low=""):
```
Minor consistency change.

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,485 @@
...
+        # VSH: 0 direct switch at 4e-4 mbar, 1 continuous between 1e-3 to 2e-3 mbar,
+        #      2 continuous between 2e-3 to 5e-3 mbar
+        if got == "0":
+            return "direct"
+        elif got == "1":
+            return "continuous"
+        elif got == "2":
+            return "continuous 2"
+        else:
+            return got
```
msmttchr, line 409:
```suggestion
        mapping = { 
                   "0" : "direct",
                   "1": "continuous",
                   "2" : "continuous 2",
        }
        try:
            return mapping[got]
        except KeyError:
            return got
```
Minor change, feel free to ignore it (as you probably guess, I hate if statement :) )

msmttchr (commented):
Last bunch of comments from my side. Thanks.

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,485 @@
...
+
+class SensorChannel(Channel):
+    """Generic channel for individual pressure sensors of a Transmitter."""
+
+    _id = -1  # obligatory channel number, define in channel types
+
+    def __init__(self, parent, id):
+        if id is None:
+            id = self._id
+        assert id == self._id, f"Pirani ID has to be {self._id}."
```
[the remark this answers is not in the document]
bmoneke: It serves a purpose:
The `Channel` class requires an `id` parameter and defines such an attribute in its init
In this case, the `id` is tied to the specific class (e.g. `PiraniSensor`).

This check ensures, that nobody defines another id in the ChannelCreator.

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: > but I am not sure what pyvisa does when a char > 127 is received

that is exactly the issue, why I implemented that complicated method: pyvisa uses decode, which hangs with byte values above 127.

> ProtocolAdapter does not implement read_termination. Is this correct ?

The ProtocolAdapter will not break at the TermChar, that is correct.

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: with python 3.10, we can use `match case` :grin:

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: Reasons for the elements:
1. `rstrip`: If we use read_bytes until a term_char, the term_char will be returned as well, so we have to strip it (any adapter, be it serial, VISA or protocol)
2. ignore errors: to skip those extra chars, which happen to end up in the communication.


Btw, it could be a problem with my RS485 adapter, but I don't have an alternative to test.

pymeasure/instruments/thyracont/smartline_v2.py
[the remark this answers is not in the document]
msmttchr: Ok, but as I understand your code, users should not use `SensorChannel` directly, but only via subclasses. If a subclass is used, the ID is automatically defined, so why requesting users to specify the ID?
If the `id` parameter is needed anyway, I would specify it in the signature as `None` and ignore its value, if passed by user. Do you agree?

pymeasure/instruments/thyracont/smartline_v2.py
msmttchr: Ok, I agree this is weird, but I am sorry I have no experience on RS485 either, so I am not able to comment.
At least, the `if` in line 324 can be removed and do always the replace.

Off topic: I am not familiar about the `ProtocolAdapter`, but the behaviour is not consistent with `SerialAdapter`, but I understand this is intentional. Probably I need to read the documentation before commenting further.

pymeasure/instruments/thyracont/smartline_v2.py
```diff
@@ -0,0 +1,485 @@
...
+# THE SOFTWARE.
+#
+
+from enum import IntEnum
+
+from pymeasure.instruments import Instrument, Channel, validators
+from pyvisa.constants import Parity, StopBits
+
+
+def generate_enum_validator(validator):
```
msmttchr, line 31: This function seems to be unused.

pymeasure/instruments/thyracont/smartline_v2.py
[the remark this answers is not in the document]
bmoneke: What about just dropping the `id` from the signature alltogether?

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: > Off topic: I am not familiar about the ProtocolAdapter, but the behaviour is not consistent with SerialAdapter, but I understand this is intentional. Probably I need to read the documentation before commenting further.

The behaviour is consistent, except for the "break_on_termchar". The reason is, that for the `ProtocolAdapter` you do not specify term_chars (to be agnostic to term_chars), therefore, the Adapter cannot recognize a termchar and cannot stop there.
However, you give the communication to the Adapter, so reading two times up to a termchar would require the following `comm__pairs = [(None, b"part one"), (None, b"part two")]` for an instrument returning `b"part one\npart two\n"` with newline as term_char.

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: Done manually (as it is also in the tests).

pymeasure/instruments/thyracont/smartline_v2.py
bmoneke: I changed it (with `mapping.get(got, got)` 😁 )

pymeasure/instruments/thyracont/smartline_v2.py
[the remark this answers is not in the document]
msmttchr: Ok. Thanks for the clarification

msmttchr (approved):
Thanks for your contribution. Excellent job, in my opinion.

PR #944, opened 2023-07-11
Modify AWG401x driver to use 'channels'

mcdo0486:
#931 changed the behavior for `ChannelCreator` so that channels are added to a default collection `channels`.

Because `pymeasure/instruments/activetechnologies/AWG401x.py` does some advanced channel management, namely using [`add_child()` ](https://github.com/pymeasure/pymeasure/blob/6d54511f42bdc000a6779942316f52ebb6542ccf/pymeasure/instruments/activetechnologies/AWG401x.py#L459-L461) to add channels based on model number, [`SequenceEntry`](https://github.com/pymeasure/pymeasure/blob/6d54511f42bdc000a6779942316f52ebb6542ccf/pymeasure/instruments/activetechnologies/AWG401x.py#L855-L868) also using `ch` collection, and [`WaveformsLazyDict`](https://github.com/pymeasure/pymeasure/blob/6d54511f42bdc000a6779942316f52ebb6542ccf/pymeasure/instruments/activetechnologies/AWG401x.py#L766-L776) expecting `ch` key, this PR reverts the `channels` collection back to `ch`.


mcdo0486:
@bmoneke , going through history, it looks like you re-implemented the driver to channels implementation of the first `ChannelCreator`. It could be this can be moved over to the `channels` collection. For instance, is the `SequenceEntry` class now extraneous? Would you have a AWG401 to validate some of these changes?

bmoneke:
I do not have such a device, but did changes according to source code and tests.

mcdo0486:
Gotcha, I can follow it through to see where `channels` should be replacing `ch` in the driver, but b/c `SequenceEntry` is not being used in the driver but uses the 'ch' collection makes me unsure on making those changes.

mcdo0486:
> I do not have such a device, but did changes according to source code and tests.
> 
> What is now the default collection? `channels`? In this case, I'd make the code accordingly to use `channels`, such that our code base remains consistent.

Ok sounds good. Should I close this PR?

bmoneke:
Leave it open for now. I'll investigate, whether we have to take action or not.

bmoneke:
I don't see, why the collection name matters, so you could change it to `channels`.

May I ask you another change regarding `SequenceEntry`?:
change line https://github.com/pymeasure/pymeasure/blob/f6e2c13fd07c4432cdd58dffff18278dd77d74d2/pymeasure/instruments/activetechnologies/AWG401x.py#L849 to
~~~python
return SequenceEntry(self.instrument, self.num_ch, key)
~~~

In a8fbfc3cb1c8e3b2fa89c9c840f7d3a01ad0be70 I changed it, but I do not see any reason, why. It seems, that I removed it by accident.

BenediktBurger:
Hey @mcdo0486 , would you mind merging in the most recent master and give me your ok to merge?

mcdo0486:
Merged master and ok to merged

BenediktBurger:
Thanks for your work.

BenediktBurger:
Thanks for your work.

Issue #783, opened 2022-12-07
Timeout property for adapters


bmoneke:
In #742 the question was raised to harmonise adapters with a uniform timeout unit as VISA uses milliseconds and pyserial seconds.

As pyVISA is our default, its ms setting should be default as well.

msmttchr:
I would also consider whether we want to keep `SerialAdapter` at all. After you managed to switch Prologix to `pyvisa`, I don't see any use of it, except that it requires way less effort to install compared to `pyvisa`.

bmoneke:
With pyvisa-py, there should be no reason against using VISA, unless the space and computation requirement differ that much.

I'm not sure, whether it is good to keep the SerialAdapter as a minimum requirements alternative. Maybe a few instruments (calling VISA methods) won't work, but most instruments will work regardless. For this edge case, a warning about the different timeout units might suffice.

LongnoseRob:
> With pyvisa-py, there should be no reason against using VISA, unless the space and computation requirement differ that much.
> 
> I'm not sure, whether it is good to keep the SerialAdapter as a minimum requirements alternative. Maybe a few instruments (calling VISA methods) won't work, but most instruments will work regardless. For this edge case, a warning about the different timeout units might suffice.

How about a rather longterm plan to deprecatate the SerialAdapter and a short term modification to move the the timing topics (it is only timeout?)to the more used ms?

bilderbuchi:
It's easier with serial, compared to pyvisa, to do spy/loopback inspection (although we currently don't use that, we could to inspect more deeply than with protocoladapter). It's also a bit more "bare-bones"/less complex to reason about. Installation is also more "immediate" (no need to point people to pyvisa-py).
These are just ancillary points, I'm not saying they are sufficient to keep serialadapter around long-term, I just want to bring them to your attention.

PR #948, opened 2023-07-15
Adding NI GPIB-232CT adapter (RS232 - GPIB)

LongnoseRob:
This PR is as of now still WIP, but I would already call for comments

**Todo:**

- [ ] test-file for adapter
- [x] test-file for adapter and some GPIB-device
- [x] test implementation with several GPIB devices
- [x] documentation


pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def _write(self, command, **kwargs):
+        """Write a string to the instrument appending `write_termination`.
+
+        :param str command: Command string to be sent to the instrument
+            (without termination).
+        :param kwargs: Keyword arguments for the connection itself.
+        """
+        self.flush_read_buffer()
+        self.connection.write(f"wrt {self.address} \n  {command}", **kwargs)
```
bmoneke, line 214: You want to write two spaces between newline and the command itself?

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        Initiate a GPIB trigger-event
+        """
+        self.connection.write(f"trg {self.address}")
+
+    def clear(self):
+        """
+        Clear specified device.
+
+        """
+        self.connection.write(f"clr  {self.address}")
```
bmoneke, line 242: Do you intend to send several spaces between "clr" and the address?

The same with other commands.

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+
+        :param address: Integer GPIB address of the desired instrument
+        :param kwargs: Arguments for the initialization
+        :returns: NI_GPIB_232 for specific GPIB address
+        """
+        return NI_GPIB_232(self, address, **kwargs)
+
+    def pass_control(self, primary_address: int, secondary_address: int):
+        """
+        Pass control to drevice with primary_address and optional secondary_address
```
bmoneke, line 272:
```suggestion
        Pass control to device with primary_address and optional secondary_address
```

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        self.connection.write("rsc  1")
+
+    def send_ifc(self):
+        """Pulse the interface clear line (IFC) for at least 200 microseconds."""
+        self.connection.write("sic 0.0002")
+
+    @property
+    def time_out(self):
+        """Control control the GPIB timeout.
```
bmoneke, line 298:
```suggestion
        """Control the GPIB timeout.
```

I think, it is normally (VISA & Co.) called "timeout", not "time_out".

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        # log.debug("reading")
+        self.flush_read_buffer()
+        self.connection.write(f"rd #255 {self.address}")
+        time.sleep(0.050)
+        ret_val = self.connection.read()
+        if ret_val != "0":
+            ret_len = self.connection.read()
+            log.debug(f"length of read {ret_len}")
+            self._check_errors()
```
bmoneke, line 180: What's that for?
The adapter responds with a new line indicating how many signs have been read?

For what do you log the reported length, if the  `Adapter` class already logs the full content?

Would it not be better to verify, that the length of `ret_val` equals the value of  `ret_len`?
I guess that this is the idea, why the adapter sends that number.

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+            count = self.connection.chunk_size - 1
+        self.flush_read_buffer()
+        self.connection.write(f"rd #{count} {self.address}")
+        time.sleep(0.050)
+        ret_val = self.connection.read_bytes(count, break_on_termchar, **kwargs)
+        log.debug(f"read: {ret_val}")
+        time.sleep(0.050)
+        ret_len = self.connection.read()
+        log.debug(f"length of bytes read {ret_len}")
+        self._check_errors()
```
bmoneke, line 203: See above.

bmoneke (commented):
A few comments/questions.

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: Thanks fot point this out,  that was an  oversight..

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: Will change to the more common `timeout` and adjust the docstring as proposed

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: this logging was for debugging purposes,..
Will think about the proposal to compare the len of ret_val vs ret_len.

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+# pylint: disable=W1203
+
+
+class NI_GPIB_232(VISAAdapter):
+    """Encapsulates the additional commands necessary
+       to communicate over a National Instruments GPIB-232CT Adapter,
```
bilderbuchi, line 38: Why leave out the `CT` part of the model name in the class name? Is that some variant indicator? Or is that `CT-A`?

Also, following the CamelCase convention for class names, this should probably be `NIGPIB232CT`, but that might be unwieldy? Thoughts?

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        SRQI = 4096  # SRQ detected while CIC
+        # 2048, 1024,  & 512 are reserved
+        CMPL = 256  # Operation completed
+        LOK = 128  # Lockout state
+        REM = 64  # remote status
+        CIC = 32  # CIC (Controller in Charge) status
+        ATN = 16  # ATN asserted
+        TACS = 8  # Talker active
+        LACS = 4  # Listener active
+        DTAS = 2  # Device triggeer active state
```
bilderbuchi, line 104:
```suggestion
        DTAS = 2  # Device trigger active state
```

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        SCAS = 1  # Device Clear active status
+
+    class GPIBError(Flag):
+        """Enum element for GIBP error bit decoding
+
+        """
+        ECMD = 17  # unregcognized command
+        # 15-16 servered
+        EBUS = 14  # Command bytes could not be sent
+        # 12-13 servered
```
bilderbuchi, line 114:
```suggestion
        # 15-16 reserved
        EBUS = 14  # Command bytes could not be sent
        # 12-13 reserved
```

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        TACS = 8  # Talker active
+        LACS = 4  # Listener active
+        DTAS = 2  # Device triggeer active state
+        SCAS = 1  # Device Clear active status
+
+    class GPIBError(Flag):
+        """Enum element for GIBP error bit decoding
+
+        """
+        ECMD = 17  # unregcognized command
```
bilderbuchi, line 111: "unrecognized"

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        # 15-16 servered
+        EBUS = 14  # Command bytes could not be sent
+        # 12-13 servered
+        ECAP = 11  # No capability for operation
+        # 7-10 reserved
+        EABO = 6  # IO aborted
+        ESAC = 5  # Command requires GPIB-232CT-A to be system controller
+        EARG = 4  # invaild argument(s)
```
bilderbuchi, line 119:
```suggestion
        EARG = 4  # invalid argument(s)
```

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        ATN = 16  # ATN asserted
+        TACS = 8  # Talker active
+        LACS = 4  # Listener active
+        DTAS = 2  # Device triggeer active state
+        SCAS = 1  # Device Clear active status
+
+    class GPIBError(Flag):
+        """Enum element for GIBP error bit decoding
+
+        """
```
bilderbuchi, line 110:
```suggestion
        """Enum element for GPIB error bit decoding"""
```

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        self.connection.write("stat n")
+        if self.connection.bytes_in_buffer == 0:
+            time.sleep(0.125)
+        ret_val = self.connection.read_bytes(self.connection.bytes_in_buffer)
+        log.debug(f"read: {ret_val}")
+        if len(ret_val) <= 3:
+            log.warning(f"only {len(ret_val)} bytes received, content: {ret_val}")
+            return
+        if len(ret_val) < 12:
+            log.warning(f"only {len(ret_val)} bytes received, content: {ret_val}")
```
bilderbuchi, line 149: Why are 3 or 12 bytes (variying degrees of) bad?

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+            [g_s, g_e, s_e, cnt_raw] = ret_val.splitlines()
+            gpib_stat = self.GPIBStatus(int(g_s))
+            gpib_err = self.GPIBError(int(g_e))
+            ser_err = self.SERIALError(int(s_e))
+            count = int(cnt_raw)
+            log.debug(f"count {count}")
+        except ValueError:
+            g_s = ret_val[: ret_val.find(b"\r")]
+            gpib_stat = self.GPIBStatus(int(g_s))
+            log.warning(f"ACHTUNG! {gpib_stat!a} \r\n")
```
bilderbuchi, line 161: Cute. :-D 
Still, we try to stay with English ;-)
Also, a logging message does not need a line termination, right?

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        # Error handling
+        if bool(gpib_stat & self.GPIBStatus.ERR) is True:
+            log.critical(f"Error {self.GPIBError(gpib_err)!a} {self.SERIALError(ser_err)!a}")
```
bilderbuchi, line 164: A critical error is normally something unrecoverable. Should you `raise` an exception at this point?

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        :param bool break_on_termchar: Stop reading at a termination character.
+        :param kwargs: Keyword arguments for the connection itself.
+        :returns bytes: response of the instrument.
+        """
+        # log.debug("read bytes..")
+        if count == -1:
+            count = self.connection.chunk_size - 1
+        self.flush_read_buffer()
+        self.connection.write(f"rd #{count} {self.address}")
+        time.sleep(0.050)
```
bilderbuchi, line 197: are those hard-coded sleeps device-specific and permanent? Or is this something you tweaked?
Is that related/connected to the "GPIB timeout" mentioned below?

In any case, might be good to add a value as a module- or class constant and use that (or multiples of it) in the code. This way this can be changed (if needed) or zeroed (to not slow down tests for example).

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        self.flush_read_buffer()
+
+    def _write_bytes(self, content, **kwargs):
+        """Write byte to the instrument appending `write_termination`.
+
+        :param str content: string to be sent to the instrument (without termination).
+        :param kwargs: Keyword arguments for the connection itself.
+        """
+        self.flush_read_buffer()
+        self.connection.write(f"wrt {self.address} \n {content}", **kwargs)
```
bilderbuchi, line 226: in self.write_termination you have `\r`, but here you use a `\n` as the separator -- is that intentional?

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        time.sleep(0.050)
+        self._check_errors()
+        self.flush_read_buffer()
+
+    def assert_trigger(self):
```
bilderbuchi, line 231:
```suggestion
    def trigger(self):
```
Wouldn't that be simpler?

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+
+        :param address: Integer GPIB address of the desired instrument
+        :param kwargs: Arguments for the initialization
+        :returns: NI_GPIB_232 for specific GPIB address
+        """
+        return NI_GPIB_232(self, address, **kwargs)
+
+    def pass_control(self, primary_address: int, secondary_address: int):
+        """
+        Pass control to device with primary_address and optional secondary_address.
```
bilderbuchi, line 272: secondary_address is not optional here  -- forgot a kwarg default?

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        self.flush_read_buffer()
+        self.connection.write("id \r")
+        time.sleep(0.02)
+        return self.connection.read_bytes(71).decode()
+
+    def wait_for_srq(self, timeout=20, delay=0.1):
+        """Blocks until a SRQ, and leaves the bit high
+
+        :param timeout: Timeout duration in seconds.
+        :raises TimeoutError: "Waiting for SRQ timed out."
```
bilderbuchi, line 325: What's the meaning of the return value (`None` or integer)?

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,346 @@
...
+        while time.perf_counter() < stop:
+            self.connection.write(f"rsp {self.address}")
+            if self.connection.bytes_in_buffer == 0:
+                time.sleep(delay)
+            if self.connection.bytes_in_buffer >= 1:
+                ret_val = self.connection.read_bytes(self.connection.bytes_in_buffer)
+                if int(ret_val) > 0:
+                    return int(ret_val)
+                if int(ret_val) == -1:
+                    raise TimeoutError(f"Waiting for SRQ timed out after {timeout}")
```
bilderbuchi, line 337: Is this a different timeout than the one controlled by the `timeout` argument for this method? If yes, this could probably use some clarification.
I would have expected the TimeoutError to be raised when the while loop exited because the wall time went above the `timeout` cutoff (i.e. that is currently `return None`).

bilderbuchi (commented):
Some questions from a passing read.

Originally, I intended to ask if this could not be derived from `PrologixAdapter` as, as far as I can tell, it has the same purpose. However, the differences are probably too large for this to be useful.

Should we maybe still try to extract a base class, even if only for interface purposes, to make sure we are as universal as possible? @bmoneke @msmttchr , thoughts?

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: As far as I oversee the list of N GPIB-RS232/485 adapters, there are:
- the NI GPIB-RS232 (which is still listed for sale on the NI website)
- the NI GPIB-232/485CT-A and
- the NI GPIB-232CV-A
- the NI GPIB-232CT (older device)

The 232/485CT-A can be used as GPIB master and slave device,
while the CV-A-models are only for converting a device with RS232 to a GPIB-slave or connecting to a single GPIB-device.
Detail can be found [here](https://www.apexwaves.com/pdfs/gpib-232ct-a-user-manual.pdf) 

The `-A` suffix could be omitted, but is printed on the device itself. 
I have found in the meantime a manual for the oolder 232CT version but still will have to compare the commands

For the naming, at the moment I would be ok with` NIGPIB232CT`.
But let me give you a more precive feedback, once I investigated the differewnces between CT-A and CT.

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: That was just to catch my attention during debugging, the `\r\n` are not needed, but sometimes a clean line-break is more readable..
will be remove in the next commit

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: Yes, that should be done properly.
WOuld it be ok to rais a `ValueError` or is there a btter "error-mode"?

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: at the moment these values are selected somehat by tests with my setup at 38.4 kbps comm-speed, I should change this to a global setting.
These delays are inplace, as I found that sometimes the data would not be available if I have no delay at all (at leat under Linux).

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: This is intentional, as the GPIB-232CT-A separates the adress from the message by the `\n` and detects end of data by `\r`

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: - pylint was complaining without the `return None`
- it could happen that no SRQ signal is caught (case None) or the status byte is returned

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: Yes, it is a dedicate timeout paramter for the SRQ "scan"

LongnoseRob:
> Some questions from a passing read.
> 
> Originally, I intended to ask if this could not be derived from `PrologixAdapter` as, as far as I can tell, it has the same purpose. However, the differences are probably too large for this to be useful.
> 
> Should we maybe still try to extract a base class, even if only for interface purposes, to make sure we are as universal as possible? @bmoneke @msmttchr , thoughts?

@bilderbuchi thank you for the comments and questions,
some I have already adressed and pushed solutions, further updates will follow

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: spelling fixed

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: spelling fixed

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: spelling fixed

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: spelling fixed

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: Fixed

bmoneke:
> Originally, I intended to ask if this could not be derived from PrologixAdapter as, as far as I can tell, it has the same purpose. However, the differences are probably too large for this to be useful.

> Should we maybe still try to extract a base class, even if only for interface purposes, to make sure we are as universal as possible? @bmoneke @msmttchr , thoughts?

THe universal interface is the `Adapter`. I have to compare both adapters to find out, whether they could be united. I'll doubt it though, as all the general logic resides in Adapter and VISAAdapter.

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: The "stat n" command to the GPIB-232CT-A returns normally 4 numeric value representing teh GPIB status, the GPIB error detail, the serial error detail and the count of bytes trensferred by the read/write command before.
These two Ifs were from my debugging sessions with serial comm interruptions.
Will be reomoved

pymeasure/adapters/ni_gpib_232ct.py
bilderbuchi: thanks!

pymeasure/adapters/ni_gpib_232ct.py
bilderbuchi: There is the `ConnectionError` exception in the standard lib.

pymeasure/adapters/ni_gpib_232ct.py
bilderbuchi: OK, then.

pymeasure/adapters/ni_gpib_232ct.py
bilderbuchi: OK; you could document that in the docstring.

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,345 @@
...
+         adapter = NI_GPIB_232("ASRL5::INSTR", 7)
+         sourcemeter = Keithley2400(adapter)  # at GPIB address 7
+         # generate another instance with a different GPIB address:
+         adapter2 = adapter.gpib(9)
+         multimeter = Keithley2000(adapter2)  # at GPIB address 9
+
+
+    """
+
+    def __init__(self, resource_name, address=None, serial_timeout=1000, rw_delay=50,
```
bilderbuchi, line 69: We intend to harmonize at least timeouts toward milliseconds (#783), so that is the correct choice. Units should be documented in the docstring, though, for additional clarity.

LongnoseRob: docstrings will be updated in a commit soon

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -89,7 +89,7 @@ def __init__(self, resource_name, address=None, serial_timeout=1000, rw_delay=50
             self.eoi = eoi
         self.flush_read_buffer()
 
-    class GPIBStatus(IntFlag):
+    class GPIBStatus(int, Flag):
```
bilderbuchi, line 92: don't do that yet, we are not yet clear if this is the good way to fix this python 3.11 problem (#896)

LongnoseRob: > don't do that yet, we are not yet clear if this is the good way to fix this python 3.11 problem (#896)

Thanks for the wrning, reverted the last commit

pymeasure/adapters/ni_gpib_232ct.py
bilderbuchi: > For the naming, at the moment I would be ok with` NIGPIB232CT`.
> But let me give you a more precive feedback, once I investigated the differewnces between CT-A and CT.

Sounds good. It might be good to add that useful info you found on the model differences to the source, either in a docstring or code comment (just so it's not lost).

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,328 @@
...
+    Each adapter is constructed based on a connection to the device
+    itself and the GPIB address of the instrument to be communicated with.
+    Connection sharing is achieved by using the `.gpib` method
+    to spawn new NI_GPIB_232s for different GPIB addresses.
+
+    :param resource_name: A VISA resource string that identifies the\
+    connection to the device itself, for example "ASRL5" for the 5th COM port.
+    :param address: Integer GPIB address of the desired instrument.
+    :param serial_timeout: Timeout value for the serial communication, unit: ms
+    :param rw_delay: internal delay, unit: ms, default 50 ms
```
bilderbuchi, line 49: "Internal delay" for what? When is that important? Adding this info makes the parameter more useful and gives needed context for users changing it.

pymeasure/adapters/ni_gpib_232ct.py
```diff
@@ -0,0 +1,328 @@
...
+        self.connection.write("id \r")
+        time.sleep(self.rw_delay/1000)
+        return self.connection.read_bytes(71).decode()
+
+    def wait_for_srq(self, timeout=20, delay=0.1):
+        """Blocks until a SRQ and return status byte(s) if applicable.
+
+        :param timeout: Timeout (for SRQ) in seconds.
+        :raises TimeoutError: "Waiting for SRQ timed out."
+        """
```
bilderbuchi, line 308:
```suggestion
        """Blocks until a SRQ and return status byte(s) if applicable.

        :param timeout: Timeout (for SRQ) in seconds.
        :raises TimeoutError: "Waiting for SRQ timed out."
        
        :returns Integer status byte or None
        """
```
Also, `delay` is not documented.

bilderbuchi: also, @bmoneke timeout should be in ms (as we strive for consistency where possible), right?

pymeasure/adapters/ni_gpib_232ct.py
bilderbuchi: From what I see in the logic of the method, a `ret_val` of `0` as received from the hardware is not possible or invalid.
You could return `0` instead of `None` if no signal is caught, then the return type remains homogeneous (which makes it easier to deal with downstream). Does that make sense in the context of SRQ values?

bilderbuchi: Another tack: what do you think about using the same interface/behaviour as https://pyvisa.readthedocs.io/en/latest/api/resources.html#pyvisa.resources.GPIBInstrument.wait_for_srq ? It does the same thing afaict. Do you need the additional code flow handling (exceptions, return None, getting the received value, etc)?

pymeasure/adapters/ni_gpib_232ct.py
bilderbuchi: I think it's a bit confusing that there are different timeout conditions. Does the difference matter between a timeout from the loop and a timeout from the scan?

pymeasure/adapters/ni_gpib_232ct.py
bilderbuchi: So, we normally implement this delay between read and write on the instrument level as a `query_delay` parameter, e.g.  https://pymeasure.readthedocs.io/en/latest/api/instruments/instruments.html#pymeasure.instruments.common_base.CommonBase.ask
(unfortunately in seconds :cry: )

I'm not sure we should duplicate this feature at different levels of the API. 
Is this delay needed for a specific instrument you are using, or is this _always_ needed in this adapter device?

bilderbuchi:
I'll try to find out what's going on with the failing docs build.

LongnoseRob:
> I'll try to find out what's going on with the failing docs build.

Thanks

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: this delay seems always needed for the communication with the GPIB devices I have here accessible to me right now

bilderbuchi:
So, as an intermediate result, the problem is apparently with a nested class that is an `IntFlag`.
It's not related to docstrings themselves (if I remove them all, it still occurs).
If I move `GPBStatus` to module scope, the problem disappears. (we should not do that)
If I inherit `GPIBStatus` from `Flag`, not `IntFlag`, the problem disappears. (we should not do that)
If I run with Sphinx 6, the problem disappears.

I'll check the Sphinx issue tracker.

LongnoseRob:
> So, as an intermediate result, the problem is apparently with a nested class that is an `IntFlag`. It's not related to docstrings themselves (if I remove them all, it still occurs). If I move `GPBStatus` to module scope, the problem disappears. (we should not do that) If I inherit `GPIBStatus` from `Flag`, not `IntFlag`, the problem disappears. (we should not do that) If I run with Sphinx 6, the problem disappears.
> 
> I'll check the Sphinx issue tracker.

Thanks for the clarification, 
on my system with Sphinx 7.2.5, sphinx-rtd-theme  1.3.0rc1 and Python 3.11.5
the problem still exists.

bilderbuchi:
So, the problem stems from docstring ultimately pulled in from `int`, which end up in the documentation for `GPIBStatus` (which we don't intend).

The minimum change to make the problem go away is to add `IntFlag` to `:inherited-members:` in the adapters.rst entry:
```
=====================
NI GPIB-232CT adapter
=====================

.. autoclass:: pymeasure.adapters.NI_GPIB_232
    :members:
    :undoc-members:
    :inherited-members: IntFlag
    :no-inherited-members: pymeasure.adapters.NI_GPIB_232.GPIBStatus, pymeasure.adapters.NI_GPIB_232.GPIBError, pymeasure.adapters.NI_GPIB_232.SERIALError
    :show-inheritance:
    :private-members: _format_binary_values
```

However, I don't know the intent behind this entry, and cannot find documentation for the option `:no-inherited-members:`, so I'd be curious what you want this to do? In any case, the documentation for the enum classes is shown in the docs I generate.

Removing that and only having `:inherited-members: pymeasure.adapters.NI_GPIB_232.GPIBStatus, pymeasure.adapters.NI_GPIB_232.GPIBError, pymeasure.adapters.NI_GPIB_232.SERIALError, IntFlag` fixes the problem, too.

LongnoseRob:
> So, the problem stems from docstring ultimately pulled in from `int`, which end up in the documentation for `GPIBStatus` (which we don't intend).
> 

Yes that was also my feeling, but I could not think of a solution.

> However, I don't know the intent behind this entry, and cannot find documentation for the option `:no-inherited-members:`, so I'd be curious what you want this to do? In any case, the documentation for the enum classes is shown in the docs I generate.

My intenion was to use the `:no-inherited-members:' as an inverse operation, but that did not work out.
> 
> Removing that and only having `:inherited-members: pymeasure.adapters.NI_GPIB_232.GPIBStatus, pymeasure.adapters.NI_GPIB_232.GPIBError, pymeasure.adapters.NI_GPIB_232.SERIALError, IntFlag` fixes the problem, too.

Done, and tested on my systems.

Thank you for looking into that!

bilderbuchi:
Great, good to hear! You're welcome!

pymeasure/adapters/ni_gpib_232ct.py
bilderbuchi: @bmoneke what do you think, how should we optimally deal with this? Also name the kwarg `query_delay` to avoid having lots of similar sounding delay names floating around? Intentionally choose a different one? We intentionally deprecated an adapter-wide query_delay in VISAAdapter https://pymeasure.readthedocs.io/en/latest/api/adapters.html#pymeasure.adapters.VISAAdapter ...

tests/adapters/test_232ct_with_device.py
```diff
@@ -79,4 +81,4 @@ def test_current_reading(self, make_clean_instrument):
         dmm.resolution = 3
         dmm.mode = "DCI"
         value = dmm.measure_DCI
-        assert type(value) is float
+        assert type(value) is float
```
bmoneke, line 84: it is better to use `isinstance(value, float)` than doing type checks.

LongnoseRob: Thanks for the recommenation, adopted

tests/adapters/test_232ct_with_devices.py
```diff
@@ -0,0 +1,103 @@
...
+
+from pymeasure.adapters import NI_GPIB_232
+from pymeasure.instruments.hp import HP3478A
+from pymeasure.instruments.hp.hpsystempsu import HP6632A
+
+pytest.skip('Only works with connected hardware', allow_module_level=True)
+
+
+class Test_NI232CT_3478A_6632A:
+    """Unit test class for NI GPIB-232CT and deivecs connected on GPIB (HP 3478A, HP6332A).
```
bmoneke, line 37:
```suggestion
    """Unit test class for NI GPIB-232CT and devices connected on GPIB (HP 3478A, HP6332A).
```

LongnoseRob: Thanks fixed

pymeasure/adapters/ni_gpib_232ct.py
LongnoseRob: > But let me give you a more precive feedback, once I investigated the differewnces between CT-A and CT.

I went trought the S-mode description of the 232CT and 232/485CT-A, for the scope of the impelmentation both devices use the same commands and syntax.

updated docstring

LongnoseRob:
Closing this PR due to timecontrains on my side

PR #952, opened 2023-07-22
Threaded access

### LINKED ISSUE #506: Thread safe communication

bilderbuchi:
Closes #506, supersedes #608.


bilderbuchi:
@bmoneke I think I need your help -- my tests uncovered that the Keithley2306 `Relay` (https://github.com/pymeasure/pymeasure/blob/1500e37af25b70a0a50f6a9125ec0acdf4794136/pymeasure/instruments/keithley/keithley2306.py#L577) should be a `Channel`, and also probably that custom `Channel` (https://github.com/pymeasure/pymeasure/blob/1500e37af25b70a0a50f6a9125ec0acdf4794136/pymeasure/instruments/keithley/keithley2306.py#L33) should inherit from the pymeasure one.
I guess we overlooked that one during the Channel transition. Can I fix this quickly without getting bogged down, or is that more restructuring work, in your opinion?

bmoneke:
> I guess we overlooked that one during the Channel transition.

We added channels, but never moved all instruments to channels, see #746

BenediktBurger:
Hey @dkriegner , this PR is not worked on actively, right now.
You may use the work. In #506 , bilderbuchi wrote some comments.
I'll add there a little description of LECO, which might solve your problem via micro-service architecture.

bilderbuchi:
Yes, this is currently not being worked on. However, the implementation is pretty minimal (introducing a ReentrantLock and using it at a couple of spots) and was solid IIRC. 
It's the testing that held up things -- reliably testing for these multithreading problems is really tricky, and I did not cover all the aspects in test_threaded_access yet before I had to step away.

PR #959, opened 2023-08-31
Update ibeamsmart.py

waveman68:
There is a mistake in the asrl= assignment. This should be a dict, not a set, since the `visa.py` adapter expects this. Other instruments implement this correctly.

Otherwise the visa adapter throws an exception:

```python
  File ".../envs/seschet-device-env/lib/python3.11/site-packages/pymeasure/adapters/visa.py", line 122, in __init__
    for k, v in kwargs[key].items():
                ^^^^^^^^^^^^^^^^^
AttributeError: 'set' object has no attribute 'items'
```


waveman68:
This should be labelled as a bug, but I assume that I cannot set labels as a non-contributor.

msmttchr (approved):
Ok. Thanks for your fix

PR #960, opened 2023-09-01
Fix typos

omahs:
Fix typos

### DIFF
--- a/docs/tutorial/graphical.rst
+++ b/docs/tutorial/graphical.rst
@@ -393,7 +393,7 @@ As an example, :code:`arange(0, 10, 1)` generates a list increasing with steps o
 This way complex sequences can be entered easily.
 
 The sequences can be extended and shortened using the buttons :code:`Add root item`, :code:`Add item`, and :code:`Remove item`.
-The later two either add an item as a child of the currently selected item or remove the selected item, respectively.
+The latter two either add an item as a child of the currently selected item or remove the selected item, respectively.
 To queue the entered sequence the button :code:`Queue` sequence can be used.
 If an error occurs in evaluating the sequence text-boxes, this is mentioned in the logger, and nothing is queued.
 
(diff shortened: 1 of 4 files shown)


docs/tutorial/graphical.rst
```diff
@@ -2,7 +2,7 @@
 Using a graphical interface
 ###########################
 
-In the previous tutorial we measured the IV characteristic of a sample to show how we can set up a simple experiment in PyMeasure. The real power of PyMeasure comes when we also use the graphical tools that are included to turn our simple example into a full-flegged user interface.
+In the previous tutorial we measured the IV characteristic of a sample to show how we can set up a simple experiment in PyMeasure. The real power of PyMeasure comes when we also use the graphical tools that are included to turn our simple example into a full-fleged user interface.
```
msmttchr, line 5:
```suggestion
In the previous tutorial we measured the IV characteristic of a sample to show how we can set up a simple experiment in PyMeasure. The real power of PyMeasure comes when we also use the graphical tools that are included to turn our simple example into a full-fledged user interface.
```
I think the fix proposed by you is not correct. Do you agree ?

docs/tutorial/graphical.rst
```diff
@@ -393,7 +393,7 @@ As an example, :code:`arange(0, 10, 1)` generates a list increasing with steps o
 This way complex sequences can be entered easily.
 
 The sequences can be extended and shortened using the buttons :code:`Add root item`, :code:`Add item`, and :code:`Remove item`.
-The later two either add a item as a child of the currently selected item or remove the selected item, respectively.
+The later two either add an item as a child of the currently selected item or remove the selected item, respectively.
```
msmttchr, line 396:
```suggestion
The latter two either add an item as a child of the currently selected item or remove the selected item, respectively.
```
Do you agree ?

msmttchr (changes requested):
Thanks for your contribution. Few other spell issues found.

msmttchr (approved):
Thanks 👍

PR #964, opened 2023-09-11
Dedicated filename input and default queue method

CasperSchippers:
Something I've been wanting to add for quite a while now (initial idea based on #317 and #423).

This PR adds
- a dedicated filename input field (similar to the directory input field that has been added in #317).
- The input field also has a toggle which allows to easily switch between storing the measurement in a file, or to a temporary file.
- I propose to add a default queue method: together with the directory input we have everything we need for a default queue method, reducing the boilerplate code needed to get a working GUI. The queue method doesn't work if the filename and directory input fields are disabled (which can be done via keyword arguments in the ManagedWindows). I hence also removed the queue methods in the examples (whenever applicable).
- I propose to enable the filename and directory input-widgets by default.

I'd like to hear what you think of this?

### DIFF
--- a/pymeasure/display/widgets/filename_widget.py
+++ b/pymeasure/display/widgets/filename_widget.py
@@ -24,7 +24,7 @@
 
 import logging
 
-from ..Qt import QtCore, QtGui, QtWidgets
+from ..Qt import QtCore, QtWidgets
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
--- a/pymeasure/display/widgets/filename_widget.py
+++ b/pymeasure/display/widgets/filename_widget.py
@@ -38,3 +38,30 @@ class FilenameLineEdit(QtWidgets.QLineEdit):
 
     def __init__(self, parent=None):
         super().__init__(parent=parent)
+
+        placeholders = parent.procedure_class.placeholder_names()
+        completer = PlaceholderCompleter(placeholders)
+        self.setCompleter(completer)
+
+
+class PlaceholderCompleter(QtWidgets.QCompleter):
+
+    def __init__(self, placeholders):
+        super().__init__()
+        self.placeholders = placeholders
+
+        self.setCompletionMode(QtWidgets.QCompleter.CompletionMode.PopupCompletion)
+        self.setModelSorting(QtWidgets.QCompleter.ModelSorting.CaseInsensitivelySortedModel)
+        self.setCaseSensitivity(QtCore.Qt.CaseInsensitive)
+        self.setFilterMode(QtCore.Qt.MatchContains)
+
+    def splitPath(self, path):
+        if path.endswith('{'):
+            options = [path + placeholder for placeholder in self.placeholders]
+            model = QtCore.QStringListModel(options)
+            self.setModel(model)
+        elif path.count("{") == path.count("}"):
+            # Clear the autocomplete options
+            self.setModel(QtCore.QStringListModel())
+
+        return [path]
--- a/pymeasure/display/widgets/filename_widget.py
+++ b/pymeasure/display/widgets/filename_widget.py
@@ -22,17 +22,19 @@
 # THE SOFTWARE.
 #
 
-from .browser_widget import BrowserWidget
-from .directory_widget import DirectoryLineEdit
-from .filename_widget import FilenameLineEdit
-from .estimator_widget import EstimatorWidget, EstimatorThread
-from .image_frame import ImageFrame
-from .image_widget import ImageWidget
-from .inputs_widget import InputsWidget
-from .log_widget import LogWidget
-from .plot_frame import PlotFrame
-from .plot_widget import PlotWidget
-from .results_dialog import ResultsDialog
-from .sequencer_widget import SequencerWidget
-from .tab_widget import TabWidget
-from .table_widget import TableWidget
+import logging
+
+from ..Qt import QtCore, QtGui, QtWidgets
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class FilenameLineEdit(QtWidgets.QLineEdit):
+    """
+    Widget that allows to choose a filename.
+    A completer is implemented for quick completion of placeholders
+    """
+
+    def __init__(self, parent=None):
+        super().__init__(parent=parent)
--- a/pymeasure/display/windows/managed_window.py
+++ b/pymeasure/display/windows/managed_window.py
@@ -44,7 +44,7 @@ from ..widgets import (
     FilenameLineEdit,
     EstimatorWidget,
 )
-from ...experiment import Results, Procedure
+from ...experiment import Results, Procedure, unique_filename
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
@@ -510,9 +510,11 @@ class ManagedWindowBase(QtWidgets.QMainWindow):
         self.queue()
 
     def queue(self, procedure=None):
-        """
+        """ Queue a measurement based on the parameters in the input-widget.
 
-        Abstract method, which must be overridden by the child class.
+        Semi-abstract method, which must be overridden by the child class if the filename- and
+        directory-inputs are disabled. when filename- and directory inputs are enabled, overwriting
+        is not required, but can be done for custom naming, input processing, or other features.
 
         Implementations must call ``self.manager.queue(experiment)`` and pass
         an ``experiment``
@@ -539,8 +541,26 @@ class ManagedWindowBase(QtWidgets.QMainWindow):
                 self.manager.queue(experiment)
 
         """
-        raise NotImplementedError(
-            "Abstract method ManagedWindow.queue not implemented")
+
+        # Check if the filename and the directory inputs are available
+        if not (self.filename_input and self.directory_input):
+            raise NotImplementedError("Queue method must be overwritten if the filename- and"
+                                      "directory-inputs are disabled.")
+
+        if procedure is None:
+            procedure = self.make_procedure()
+
+        # TODO: use toggle for saving data
+        filename = unique_filename(
+            self.directory,
+            prefix=self.filename,
+            procedure=procedure
+        )
+
+        results = Results(procedure, filename)
+
+        experiment = self.new_experiment(results)
+        self.manager.queue(experiment)
 
     def abort(self):
         self.abort_button.setEnabled(False)
--- a/pymeasure/display/windows/managed_window.py
+++ b/pymeasure/display/windows/managed_window.py
@@ -27,6 +27,7 @@ import logging
 import os
 import platform
 import subprocess
+import tempfile
 
 import pyqtgraph as pg
 
@@ -550,12 +551,14 @@ class ManagedWindowBase(QtWidgets.QMainWindow):
         if procedure is None:
             procedure = self.make_procedure()
 
-        # TODO: use toggle for saving data
-        filename = unique_filename(
-            self.directory,
-            prefix=self.filename,
-            procedure=procedure
-        )
+        if self.store_measurement:
+            filename = unique_filename(
+                self.directory,
+                prefix=self.filename,
+                procedure=procedure
+            )
+        else:
+            filename = tempfile.mktemp(prefix='TempFile_', suffix='.csv')
 
         results = Results(procedure, filename)
 
--- a/pymeasure/display/windows/managed_window.py
+++ b/pymeasure/display/windows/managed_window.py
@@ -161,7 +161,7 @@ class ManagedWindowBase(QtWidgets.QMainWindow):
 
     def _setup_ui(self):
         if self.filename_input:
-            self.filename_label = QtGui.QLabel(self)
+            self.filename_label = QtWidgets.QLabel(self)
             self.filename_label.setText('Filename')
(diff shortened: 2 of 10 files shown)


bmoneke:
Good initiative, sounds good.

pymeasure/display/windows/managed_window.py
```diff
@@ -155,6 +163,15 @@ def __init__(self,
         self._layout()
 
     def _setup_ui(self):
+        if self.filename_input:
+            self.writefile_toggle = QtWidgets.QCheckBox('Write file', self)
+            self.writefile_toggle.setLayoutDirection(QtCore.Qt.RightToLeft)
+            self.writefile_toggle.setChecked(True)
+            self.writefile_toggle.stateChanged.connect(self.toggle_file_dir_input_active)
```
bmoneke, line 170: I like tooltips, which describe, what an option does. Maybe you could add one.
I guess this toggle switches between writing a file into the directory and a temporary file, doesn`t it?

pymeasure/display/windows/managed_window.py
```diff
@@ -107,8 +110,11 @@ class ManagedWindowBase(QtWidgets.QMainWindow):
         code:`Load sequence` button
     :param inputs_in_scrollarea: boolean that display or hide a scrollbar to the input area
     :param directory_input: specify, if present, where the experiment's result will be saved.
+    :param filename_input: specify, if present, the base of the filename where the results will be
+        saved.
```
bmoneke, line 114: How will the full name of the filename look like?

pymeasure/display/widgets/filename_widget.py
```diff
@@ -0,0 +1,67 @@
...
+
+class FilenameLineEdit(QtWidgets.QLineEdit):
+    """
+    Widget that allows to choose a filename.
+    A completer is implemented for quick completion of placeholders
+    """
+
+    def __init__(self, parent=None):
+        super().__init__(parent=parent)
+
```
bmoneke, line 41: The content of this widget does not describe the full filename, does it?
It is just the base, which will be made unique and will be expanded by the file type extension, isn't it?

I'd add a tooltip, which states, that this is the base of the filename.

pymeasure/experiment/results.py
```diff
@@ -69,8 +69,9 @@ def replace_placeholders(string, procedure, date_format="%Y-%m-%d", time_format=
     """
     now = datetime.now()
 
-    parameters = procedure.parameter_objects()
+    parameters = procedure.placeholder_objects()
     placeholders = {param.name: param.value for param in parameters.values()}
+    placeholders.update({varname: param.value for varname, param in parameters.items()})
```
bmoneke, line 74: does not placeholder_objects already contain the parameters, such that line 73 is superfluous?

bmoneke (approved):
Generally good suggestion. I left a few questions/suggestions, especially regarding documentation (I like tooltips to explain more, what a button/checkbox etc. does).

pymeasure/experiment/results.py
CasperSchippers: You are absolutely right; I had removed that but missed it in a commit.

As some context, I was hoping to use some of those nice 'in-string' calculations, like you can do with f-strings (`f"{delay * 10:.2f}"` for example, such that you can format you units nicely for example); for this i would need to add the variable-names of a parameter as a valid placeholder (since the names containing spaces won't work). However, I quickly came to realize that this wont work at all when using the `format` method, rather than f-strings.... unfortunately.

CasperSchippers:
I've restructured some things a bit: I moved the directory and filename input to a new widget (FileInputWidget) and moved the scroll-area optionality to the InputsWidget (and removed some margins)

pymeasure/display/widgets/filename_widget.py
```diff
@@ -0,0 +1,152 @@
...
+        self.placeholders = procedure_class.placeholder_names()
+        self.placeholders.extend(["date", "time"])
+
+        self.setToolTip(
+            "The filename of the file to which the measurement will be stored. Placeholders (in \n"
+            "standard python format, i.e.: '{variable name:formatspec}') will be replaced by \n"
+            "the respective value. The extension '.csv' will be appended, unless an extension\n"
+            "(one of '.txt', or '.csv') is recognized. Additionally, an index number ('_#') is\n"
+            "added to ensure the uniqueness of the filename.\n"
```
CasperSchippers, line 51: Note to self: do not forget to implement stripping the extension.

pymeasure/display/widgets/filename_widget.py
```diff
@@ -0,0 +1,152 @@
...
+        super().__init__(parent=parent)
+
+        self.placeholders = procedure_class.placeholder_names()
+        self.placeholders.extend(["date", "time"])
+
+        self.setToolTip(
+            "The filename of the file to which the measurement will be stored. Placeholders (in \n"
+            "standard python format, i.e.: '{variable name:formatspec}') will be replaced by \n"
+            "the respective value. The extension '.csv' will be appended, unless an extension\n"
+            "(one of '.txt', or '.csv') is recognized. Additionally, an index number ('_#') is\n"
```
CasperSchippers, line 50: Note to self: add attributes for default and recognized extensions.

docs/tutorial/graphical.rst
```diff
@@ -408,16 +378,15 @@ An example of such a sequence file is given below, resulting in the sequence sho
 
 This file can also be automatically loaded at the start of the program by adding the key-word argument :code:`sequence_file="filename.txt"` to the :code:`super().__init__` call, as was done in the example.
 
-Using the directory input
-~~~~~~~~~~~~~~~~~~~~~~~~~
+The filename and directory input
```
CasperSchippers, line 381: Note to self: modify this section.

CasperSchippers:
I think it's ready now.

@bmoneke, since I did a bit of reworking (e.g. moving the filename and directory input fields to a seperate widget, and some small visual changes to the inputswidget), could you have another look (can be quick I think)?

docs/tutorial/graphical.rst
```diff
@@ -208,6 +203,65 @@ Now that you have learned about the ManagedWindow, you have all of the basics to
...
+    :width: 24%
+.. image:: pymeasure-fileinput_complete_filename.png
+    :alt: The filename and directory input widget showing the auto-complete for the filename
+    :width: 24%
+.. image:: pymeasure-fileinput_complete_directory.png
+    :alt: The filename and directory input widget showing the auto-complete for the directory
+    :width: 24%
+
+With the filename input, additionally a toggle (named :guilabel:`Write file` controlling whether the measurement is save at all, or should be stored in a temporary file, is displayed.
+When the toggle is ticked, the measurement is stored to a temporary file, and the other input fields are disabled.
```
bmoneke, line 225: A more concise description
```suggestion
If the checkbox named :guilabel:`Write file` is enabled, the measurement is written to a file.
Otherwise, it is stored in a temporary file.
```

From the code below, I noticed an inconsistency:
You **un**check "write file" in order to write to a file?
For me it is clearer to tick "write file" to write it actually.
Alternatively, it could be labelled "temp file".

docs/tutorial/graphical.rst
```diff
@@ -208,6 +203,65 @@ Now that you have learned about the ManagedWindow, you have all of the basics to
...
+
+When the measurement is to be stored in to a certain file, ensure that the :guilabel:`Write file` is disabled.
```
bmoneke, line 227: It doesn't seem right to me to check "write file" in order to **not** write to a file!

docs/tutorial/graphical.rst
```diff
@@ -208,6 +203,65 @@ Now that you have learned about the ManagedWindow, you have all of the basics to
...
+The filename in the designated field can be entered with or without extension; if the entered extension is recognized (by default :code:`.csv` and :code:`.txt`) are recognized, that extension is used.
```
bmoneke, line 228:
```suggestion
The filename in the designated field can be entered with or without extension.
If the entered extension is recognized (by default :code:`.csv` and :code:`.txt` are recognized), that extension is used.
```

Parenthesis was misleading.

docs/tutorial/graphical.rst
```diff
@@ -208,6 +203,65 @@ Now that you have learned about the ManagedWindow, you have all of the basics to
...
+            self.setWindowTitle('GUI Example')
+
+            self.filename = r'default_filename_delay{Delay Time:4f}s'   # Sets default filename
+            self.directory = r'C:/Path/to/default/directory'            # Sets default directory
+            self.store_measurement = False                              # Controls the 'write-file' togge
+            self.file_input.extensions = ["csv", "txt", "data"]         # Sets recognized extensions, first entry is the default extension
+
+
+The fields for the filename and the directory can be disabled using the boolean keyword arguments :code:`filename_input` and :code:`directory_input`, respectively, in the :class:`~pymeasure.display.windows.managed_window.ManagedWindow` init.
+Note that when a either of these values is set to :code:`False`, the default :meth:`~pymeasure.display.windows.managed_window.ManagedWindowBase.queue` method of the :class:`~pymeasure.display.windows.managed_window.ManagedWindow` class will no longer work, and a new, custom, method needs to be implemented; a basic implementation is shown in the documentation of the :meth:`~pymeasure.display.windows.managed_window.ManagedWindowBase.queue` method.
```
bmoneke, line 263:
```suggestion
Note that when either of these values is set to :code:`False`, the default :meth:`~pymeasure.display.windows.managed_window.ManagedWindowBase.queue` method of the :class:`~pymeasure.display.windows.managed_window.ManagedWindow` class will no longer work, and a new, custom, method needs to be implemented; a basic implementation is shown in the documentation of the :meth:`~pymeasure.display.windows.managed_window.ManagedWindowBase.queue` method.
```

bmoneke (approved):
A few suggestions on wording and a suggestion of coding.

docs/tutorial/graphical.rst
CasperSchippers: Ah yeah, that is in inconsistency in the documentation; it is exactly as you say: you write to file when 'write file' is ticked

docs/tutorial/graphical.rst
CasperSchippers: Yes, you are right, this should reald "enabled". With your clarified sentence above, this one is unnecessary.

pymeasure/display/widgets/fileinput_widget.py
```diff
@@ -0,0 +1,153 @@
...
+from .directory_widget import DirectoryLineEdit
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class FileInputWidget(QtWidgets.QWidget):
+    _extensions = ["csv", "txt"]
+
+    def __init__(self, filename_input=True, directory_input=True, parent=None):
```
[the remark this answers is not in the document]
CasperSchippers: You are right it is a bit of an uggly construction right now. The MixIn classes is an option, but I think you are right that is is more complicated.

When rethinking this, I _think_ that it is not really necessary to be able to select only the filename or directory input field. So now I went for:
- Just a single class, no option to remove one of the fields. So also only a single boolean toggle in the main-window `enable_file_input` instead of the two filename and directory toggles.
- I did add a flag `filename_fixed` that allows for 'freezing' the filename. In that case, the filename field will be disabled, but still present; such that you clearly see where the file will be saved, but are not able to change it.

What do you think?

pymeasure/display/widgets/fileinput_widget.py
```diff
@@ -0,0 +1,153 @@
...
+        super().__init__(parent)
+
+        self._setup_ui(filename_input, directory_input)
+        self._layout()
```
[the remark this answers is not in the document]
CasperSchippers: Yeah, I understand your line of thought, especially with the two `hasattr` checks that were in there.
The idea of this separation is that you define the widgets (that you need/can access later on in the class) in the `setup_ui` method, and have the `layout` method (which can sometimes be quite full and cluttered) to handle where the widgets are placed.

SInce I've removed the `hasattr` check, i'm inclined to keep the structure like this.

pymeasure/display/windows/managed_window.py
```diff
@@ -123,7 +134,8 @@ def __init__(self,
                  sequencer_inputs=None,
                  sequence_file=None,
                  inputs_in_scrollarea=False,
-                 directory_input=False,
+                 directory_input=True,
+                 filename_input=True,
```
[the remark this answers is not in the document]
CasperSchippers: Yeah, I agree. changed.

pymeasure/display/widgets/fileinput_widget.py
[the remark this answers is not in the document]
bmoneke: > SInce I've removed the hasattr check, i'm inclined to keep the structure like this.

It's fine with me to keep the structure, which makes sense to me.

Only the "hasattr" checks made it seem better to unite the methods.

pymeasure/display/widgets/fileinput_widget.py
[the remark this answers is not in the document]
bmoneke: I like it.

CasperSchippers:
Thank you, @bmoneke, for the reviewing and thinking along!

bmoneke:
You're welcome. It show, that review are a great tool for good code (I notice it with pyleco a lot as well).

Issue #965, opened 2023-09-15
Connection Issue with ANC300 device


ileu:
Hello,

I have a problem using the ANC300 Controller. The communication over telnet works fine over the console. But when I try to connect with it in a python script I get the following error:

```
Traceback (most recent call last):
  File "C:\Users\ONG_C54_01\AppData\Roaming\JetBrains\PyCharmCE2022.2\scratches\scratch.py", line 3, in <module>
    controller = ANC300Controller(adapter="TCPIP::192.168.1.2::7230::SOCKET", axisnames=["RZ", "RY", "RX", "LZ", "LY", "LX"], passwd="123456", query_delay=2)
  File "C:\Users\ONG_C54_01\Documents\CryoGUI\venv\lib\site-packages\pymeasure\instruments\attocube\anc300.py", line 282, in __init__
    super().read()
  File "C:\Users\ONG_C54_01\Documents\CryoGUI\venv\lib\site-packages\pymeasure\instruments\instrument.py", line 152, in read
    return self.adapter.read(**kwargs)
  File "C:\Users\ONG_C54_01\Documents\CryoGUI\venv\lib\site-packages\pymeasure\adapters\adapter.py", line 108, in read
    read = self._read(**kwargs)
  File "C:\Users\ONG_C54_01\Documents\CryoGUI\venv\lib\site-packages\pymeasure\adapters\visa.py", line 172, in _read
    return self.connection.read(**kwargs)
  File "C:\Users\ONG_C54_01\Documents\CryoGUI\venv\lib\site-packages\pyvisa\resources\messagebased.py", line 486, in read
    message = self._read_raw().decode(enco)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xff in position 0: ordinal not in range(128)
```

and the corresponding script I run is:

```
from pymeasure.instruments.attocube import ANC300Controller

controller = ANC300Controller(adapter="TCPIP::192.168.1.2::7230::SOCKET", axisnames=["RZ", "RY", "RX", "LZ", "LY", "LX"], passwd="123456", query_delay=2)

print(controller.status)
```

It seems as ``self.adapter.flush_read_buffer()`` doesn't actually flush the buffer and then the adapter tries to read the non-ascii telnet control characters and crashes.

Any help is greatly appreciated.

bmoneke:
What version of pymeasure do you use? I think such an issue got already addressed, but I do not know, whether it is already released.

ileu:
[...]

PR #968, opened 2023-09-22
Fix VISAAdapter flush_read_buffer

ileu:
Added `discard_receive_buffer` after `discard_read_buffer` to make sure buffer is empty.

See discussion #965.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -64,3 +64,4 @@ Ashok Bruno
 Robert Roos
 Sebastien Weber
 Sebastian Neusch
+Ulrich Sauter
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -257,12 +257,13 @@ class VISAAdapter(Adapter):
     def flush_read_buffer(self):
         """ Flush and discard the input buffer
 
-        As detailed by pyvisa, discard the read buffer contents and if data was present
-        in the read buffer and no END-indicator was present, read from the device until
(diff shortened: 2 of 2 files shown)


bmoneke:
Thanks for the PR.

You may add yourself to the authors.txt, if you like.

Could you please try the modified flush method with a serial connection (ASRL/COM), to make sure, that it does not raise any error.

ileu:
Seems to work fine with a serial connection.

I verified with the following code:
```
adapter = VISAAdapter("ASRL5::INSTR")

print(adapter.read_bytes(-1))

adapter.write("ver")

time.sleep(0.5)

adapter.flush_read_buffer()

print(adapter.read_bytes(-1))

adapter.close()
```

resulting in the following 

```
b''
b''
```
and without the `adapter.flush_read_buffer()`

```
b''
b'ver\r\nattocube ANC300 controller version 1.2.2-3 2019-01-25 11:02\r\nANC150 compatibillity console\r\nOK\r\n> '
```

bmoneke:
Thank you for your PR.

bilderbuchi:
@bmoneke do we also need to adapt the ProtocolAdapter's `flush_read_buffer` ? It contains a copy of the code in the VISAAdapter.

bmoneke:
Thanks for remembering that part.

You're right, that it does not mirror the visa adapter anymore. On the other hand, it's there for checking, whether it is called. In this case, it is easier, if "flush" is called once for every "flush read buffer".

Let's keep as is.

PR #970, opened 2023-10-02
Add HP11713A Switch & Attenuator Driver 

neuschs:


pymeasure/instruments/hp/hp11713a.py
```diff
@@ -0,0 +1,200 @@
...
+    4: (False, False, False, True),
+    5: (True, False, False, True),
+    6: (False, True, False, True),
+    7: (True, True, False, True),
+    8: (False, False, True, True),
+    9: (True, False, True, True),
+    10: (False, True, True, True),
+    11: (True, True, True, True),
+}
+""" Mapping of logical values for use with 0 - 11 dB attenuators """
```
bmoneke, line 45: why don't you put that comment in the first line of the dict:
```suggestion
Attenuator_11dB = {
    """ Mapping of logical values for use with 0 - 11 dB attenuators """
    0: (False, False, False, False),
    1: (True, False, False, False),
    2: (False, True, False, False),
    3: (True, True, False, False),
    4: (False, False, False, True),
    5: (True, False, False, True),
    6: (False, True, False, True),
    7: (True, True, False, True),
    8: (False, False, True, True),
    9: (True, False, True, True),
    10: (False, True, True, True),
    11: (True, True, True, True),
}
```

bmoneke (commented):
Thanks for the contribution, I added a few questions and suggestions.

pymeasure/instruments/hp/hp11713a.py
```diff
@@ -0,0 +1,200 @@
...
+        cmd_str += "1" if one else ""
+        cmd_str += "2" if two else ""
+        cmd_str += "3" if three else ""
+        cmd_str += "4" if four else ""
+        cmd_str += "B" if not all([one, two, three, four]) else ""
+        cmd_str += "1" if not one else ""
+        cmd_str += "2" if not two else ""
+        cmd_str += "3" if not three else ""
+        cmd_str += "4" if not four else ""
+        self.write(cmd_str)
```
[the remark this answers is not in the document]
neuschs: The day I coded all of that I was simply a little bit stuck with my head. You are completely right. I will rearrange the code to accomodate channels.

pymeasure/instruments/hp/hp11713a.py
neuschs: It actually isn't possible with python to have docstrings inside dictionarys.

pymeasure/instruments/hp/hp11713a.py
```diff
@@ -0,0 +1,200 @@
...
+
+class HP11713A(Instrument):
+    """
+    A switch driver
+    """
+
+    ATTENUATOR_X = {}
+    ATTENUATOR_Y = {}
+
+    def __init__(self, adapter, name="Hewlett-Packard HP11713A", deactivate_string="B1234567890",
```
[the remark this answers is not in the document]
neuschs: I actually removed it now because I found no plausible reason, while using the device with this branch of pymeasure, to keep it.

docs/api/instruments/hp/hp11713A.rst
```diff
@@ -0,0 +1,9 @@
+###########################################
+HP 11713A Attenuator/Switch Driver
+###########################################
+
+.. autoclass:: pymeasure.instruments.hp.HP11713A
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.hp.hp11713a.Attenuator_11dB
```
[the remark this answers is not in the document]
neuschs: So they show up in the documentation, they are supposed to be used by the user.
Generally this device is used to hook up variable attenuators and these are the default mappings to get a certain attenuation. These mappings are fixed due to the special connectors they use.

tests/instruments/hp/test_hp11713a.py
```diff
@@ -0,0 +1,96 @@
...
+
+import pytest
+
+
+class TestHP11713A:
+    @pytest.mark.parametrize("channel", list(range(0, 9)))
+    def test_channels(self, channel):
+        with expected_protocol(
+                HP11713A,
+                [("A{channel}".format(channel=channel), None),
```
bmoneke, line 36: Just for your information:
`"A{channel}".format(channel=channel)` is equal to the f-string `f"A{channel}"`.
These f-strings make it easier to understand (if there are several replacements), which one is which, you can even add calculations: `f"{3+4}" == "7"`

pymeasure/instruments/hp/hp11713a.py
```diff
@@ -0,0 +1,173 @@
...
+        map_values=True,
+        values={True: "A", False: "B"}
+    )
+
+
+class HP11713A(Instrument):
+    """
+    Represents the HP 11713A Switch and Attenuator Driver and provides a high-level
+    interface for interacting with the instrument.
+
```
bmoneke, line 102: Do you consider it opportun to describe the attenuator definition via dictionaries, such that users understand, that they have to do it?

bmoneke:
Maybe you could expand the documentation a little bit, such that it is easier for the users to understand the use of the attenuator dictionaries.

The code itself, however, is good.

tests/instruments/hp/test_hp11713a.py
neuschs: Thanks for the info !

bmoneke:
Thanks for your contribution.

Issue #973, opened 2023-10-05
Ensure correct python version limitations


bilderbuchi:
It seems that at least pip installations don't pick up Python version limitations correctly. On Python=3.7, `pip install pymeasure` picks pymeasure 0.13, but this does not support 3.7 anymore and importing it fails with
```
    from importlib.metadata import version, PackageNotFoundError
ModuleNotFoundError: No module named 'importlib.metadata'
```
We should asap:
* [x] Find out if this affects pip _and_ conda installations
* [x] Determine what is missing in pyproject.toml or setup.cfg
* [x] Properly constrain minimum Python versions
* [x] Test this in pip and conda environments (should be fair to create conda envs only with python, and use pip for the rest)
* [x] Create an updated release. Probably, as this is a build meta change, it's best to work from tag `v0.13` and just bump the build number and leave "0.13" in place. This also avoids that we have a broken version online.

I could use some help with research on what the appropriate fix is.

bilderbuchi:
While we are at it, we should also 
* [ ] check the `setuptools_scm` dependency in pip installations.
a user reported in #972 that they got
```
import pymeasure
Traceback (most recent call last):
File "C:\ProgramData\Miniconda3\envs\CD_ENV\lib\site-packages\pymeasure_init_.py", line 34, in
import setuptools_scm
ModuleNotFoundError: No module named 'setuptools_scm
```
but I'm not sure if this is not a side effect of the other problem -- the build requirement is [specified](https://github.com/pymeasure/pymeasure/blob/master/pyproject.toml#L2). 
* [ ] We should check the setuptools_scm if that is the correct place for our usage.

bmoneke:
This entry describes how to add the [python_requires option](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires) to a pyproject.toml file:
https://stackoverflow.com/questions/72373093/how-to-define-python-requires-in-pyproject-toml-using-setuptools

In a quick glance, I did not find, where we specify the valid python version.

bilderbuchi:
> In a quick glance, I did not find, where we specify the valid python version.

only in the "classifiers" section https://github.com/pymeasure/pymeasure/blob/master/setup.cfg#L20-L24, we overlooked that that is not enough.

bmoneke:
Yeah, I found the "classifiers" section, but considered it (it seems correctly) as irrelevant for pip.

[...]

PR #975, opened 2023-10-05
Limit the Python version to >=3.8

bilderbuchi:
See #973 -- we did not limit the python version appropriately.


bilderbuchi:
Note: conda package is not affected, as it already had the correct version restriction, I just did not realise that the main package did not >.<.

PR #977, opened 2023-10-06
Bugfix: Append amplitude units

neuschs:
Fixed an issue where the spectrum analyzer reports errors for every set value which has no unit appended.

### DIFF
--- a/pymeasure/instruments/hp/hp856Xx.py
+++ b/pymeasure/instruments/hp/hp856Xx.py
@@ -2562,7 +2562,7 @@ class HP856Xx(Instrument):
             When a trace is in max-hold mode, if the threshold is raised above any of the
             trace data, the data below the threshold will be permanently lost.
         """,
-        validator=strict_range,
+        validator=strict_discrete_set,
         values=arange(-200, 30),
     )
 
--- a/pymeasure/instruments/hp/hp856Xx.py
+++ b/pymeasure/instruments/hp/hp856Xx.py
@@ -751,7 +751,7 @@ class HP856Xx(Instrument):
     def write(self, command, **kwargs):
         if "{amplitude_unit}" in command:
             command = command.format(amplitude_unit=self.amplitude_unit)
-        super(HP856Xx, self).write(command, **kwargs)
+        super().write(command, **kwargs)
 
     def set_auto_couple(self):
         """Set the video bandwidth, resolution bandwidth, input attenuation,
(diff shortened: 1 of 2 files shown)


pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -748,6 +748,11 @@ def set_crt_adjustment_pattern(self):
         set_process=lambda v: str(v).upper()
     )
 
+    def write(self, command, **kwargs):
+        if "{amplitude_unit}" in command:
+            command = command.format(amplitude_unit=self.amplitude_unit)
+        super(HP856Xx, self).write(command, **kwargs)
```
bmoneke, line 754: If you use `super`, you should not mention the current class in order to be more flexible:
```suggestion
        super().write(command, **kwargs)
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -2076,7 +2081,7 @@ def get_power_bandwidth(self, trace, percent):
         """,
         validator=joined_validators(strict_discrete_set, truncated_discrete_set),
         values=[["AUTO", "MAN"], arange(10, 2e6)],
-        set_process=lambda v: v if isinstance(v, str) else int(v),
+        set_process=lambda v: v if isinstance(v, str) else str(int(v)) + " Hz",
```
bmoneke, line 2084: You could use an f-string as an alternative:
```suggestion
        set_process=lambda v: v if isinstance(v, str) else f"{int(v)} Hz",
```

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -2452,7 +2457,7 @@ def request_service(self, input):
         """,
         validator=joined_validators(strict_discrete_set, strict_range),
         values=[["AUTO", "MAN"], arange(50E-6, 100)],
-        set_process=lambda v: v if isinstance(v, str) else ("%.3E" % v)
+        set_process=lambda v: v if isinstance(v, str) else ("%.3E S" % v)
```
bmoneke, line 2460: f-strings are the more modern style. You do not have to change it, I just want to make you aware of these possibilities:
```suggestion
        set_process=lambda v: v if isinstance(v, str) else ("{v:.3E} S")
```

For details of the formatting see https://docs.python.org/3/library/string.html#format-specification-mini-language .
Basic idea: the colon separates the variable from the formatting options.

bmoneke (approved):
A few minor requests/suggestions, especially regarding f-strings.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -995,27 +995,40 @@ def set_continuous_sweep(self):
...
-            if instr.detector_mode == 0:
+            if instr.display_line == 0:
                 pass
 
         """,
-        validator=joined_validators(strict_range, strict_discrete_set),
-        values=[[float("-inf"), float("inf")], ["ON", "OFF"]],
-        set_process=lambda v: v if isinstance(v, str) else '%.11E {amplitude_unit}' % v
+        validator=strict_range,
+        values=[float("-inf"), float("inf")]
```
bmoneke, line 1016: Is it useful to _restrict_ the range to -infinity to infinity?
The documentation won`t show this, only the source code.

In this case, I think, the validator and the values could be removed.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -3158,11 +3190,21 @@ def set_fullband(self, band):
         'set_full_span' command is activated, the span is limited to the frequency band of the
         selected harmonic.
         """,
-        validator=joined_validators(strict_range, strict_discrete_set),
-        values=[[int(1), int(54)], ["ON", "OFF"]],
+        validator=strict_range,
+        values=[int(1), int(54)],
```
bmoneke, line 3194: `1`  is already an integer, equally `54`, so `int(1)` is redundant.

I understand, that you are touching the validators, not the values. Feel free to refactor the code or not.

pymeasure/instruments/hp/hp856Xx.py
```diff
@@ -2568,9 +2564,18 @@ def store_thru(self):
             When a trace is in max-hold mode, if the threshold is raised above any of the
             trace data, the data below the threshold will be permanently lost.
         """,
-        validator=joined_validators(strict_discrete_set, strict_range),
-        values=[["ON", "OFF"], arange(-200, 30)],
-        set_process=lambda v: v if isinstance(v, str) else ("%.2E {amplitude_unit}" % v)
+        validator=strict_range,
+        values=arange(-200, 30),
```
bmoneke, line 2568: I'm not sure, whether that's the right combination of validator and values, but I understand, that it is not the scope of your PR.
AFAIK, strict_range just needs two values (`values=[1, 999]`) and not a list or iterator as `arange` is.

bmoneke (approved):
A few small comments regarding validators etc.

I know, that it is not your scope.

You do not have to refactor the code as it is not scope of your PR, but you may do it while you implement your own changes.
Feel free to refactor or not.

pymeasure/instruments/hp/hp856Xx.py
neuschs: It more or less is the right way, because it is needed that you can't set 100.5

bmoneke: in that case, you need `strict_discrete_set` as a validator, because the `strict_range` does allow anything between min and max of the list

bmoneke:
Is it ready to merge from your side?

neuschs:
> Is it ready to merge from your side?

Yes. Thanks !

PR #979, opened 2023-10-14
Integration of Power Meter HP437B

neuschs:


pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,436 @@
...
+
+
+class HP437B(Instrument):
+    """Represents the HP856XX series spectrum analyzers.
+
+    Don't use this class directly - use their derivative classes
+
+    .. note::
+        Most command descriptions are taken from the document:
+        'HP 8560A, 8561B Operating & Programming'
```
LongnoseRob, line 103: This docstring still needs adoption to the 437

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,436 @@
...
+        """
+        The power meter divides each sensor’s power range into 5 ranges of
+        10 dB each. Range 1 is the most sensitive (lowest power levels), and
+        Range 5 is the least sensitive (highest power levels). Range 5 can be
+        less than 10 dB if the sensor’s power range is less than 50 dB. The
+        range can be set either automatically or manually.
+        'activate_auto_range' automatically selects the correct range for the current
+        measurement.
+        """
+        self.write("RA")
```
LongnoseRob, line 158: Please keep the docstrings in the format proposed in the [style guide](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings)

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,436 @@
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import *
```
LongnoseRob, line 2: It is somehow good practice only to import the actual used validators.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,436 @@
...
+        validator=lambda v, vs: strict_discrete_range(v, vs, 0.001),
+        values=[0.001, 99.999],
+        check_set_errors=True
+    )
+
+    # status_message = Instrument.measurement(
+    #     "SM", "",
+    #     """
+    #     """
+    # )
```
LongnoseRob, line 240: leftover from debugging?

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,436 @@
...
+        check_set_errors=True
+    )
+
+    limit_low = Instrument.setting(
+        "LL%5.2fEN",
+        """
+        Set the lower limit for the builtin limit checking.
+        """,
+        check_set_errors=True
+    )
```
LongnoseRob, line 305: Two comments on these limit functions;
1. without reading the documentation it might be difficult for the user to understand the value range which could be applied.
2. would it make sense to have validators in these functions?

LongnoseRob (commented):
Thank you for the contribution,

i left some comments in addition to the points found by the CI.

On the actual device, have you had a look inot it how much difference there is to the HP43 (with HPIB/GPIB option) and or the dual-channel 438?

neuschs:
> Thank you for the contribution,
> 
> i left some comments in addition to the points found by the CI.
> 
> On the actual device, have you had a look inot it how much difference there is to the HP43 (with HPIB/GPIB option) and or the dual-channel 438?

I had a short look over the HP438A because it got mentioned in the manual of the 437B. 
And indeed they do share a lot of common commands, but I so far have no good idea on how to combine these two.
The HP438A is two-channel where you have to send a message in before every other command to determine which channel you want to configure or measure.
But on the 437B this issues doesn't exist.

What do you mean with "HP43 (with HPIB/GPIB)" ? The 437B as well as the 438A have GPIB options.

LongnoseRob:
> > Thank you for the contribution,
> > i left some comments in addition to the points found by the CI.
> > On the actual device, have you had a look inot it how much difference there is to the HP43 (with HPIB/GPIB option) and or the dual-channel 438?
> 
> I had a short look over the HP438A because it got mentioned in the manual of the 437B. And indeed they do share a lot of common commands, but I so far have no good idea on how to combine these two. The HP438A is two-channel where you have to send a message in before every other command to determine which channel you want to configure or measure. But on the 437B this issues doesn't exist. And additionally implementing a device without being able to test it even once feels wrong.
> 

Yes, without access to the hardware for testing it makes little sense.

> What do you mean with "HP43 (with HPIB/GPIB)" ? The 437B as well as the 438A have GPIB options.
Sorry, I ment the HP436..

neuschs:
> Yes, without access to the hardware for testing it makes little sense.
> 

I can pack the common commands into a base class so someone can later reuse them for the HP438A. But that would only make sense if it makes sense to inherit from the Channel and the BaseClass for the 438A because that one has the two channels.


I checked, the HP436A shares nothing in common with the HP437B.

LongnoseRob:
> > Yes, without access to the hardware for testing it makes little sense.
> 
> I can pack the common commands into a base class so someone can later reuse them for the HP438A. But that would only make sense if it makes sense to inherit from the Channel and the BaseClass for the 438A because that one has the two channels.
> 
> I checked, the HP436A shares nothing in common with the HP437B.

It was just an idea but if it is too much work for the 438, then continue with your initial approach,
If someone then wants to interface the 438, they can then think abiut the channel stuff at that time.

neuschs:
@LongnoseRob 

I created a mockup up of how a 438A combination could be implemented. What do you think ? The multi inherence is kind of awkward.

`
from pymeasure.instruments import Instrument, Channel
from pymeasure.adapters import ProtocolAdapter
 

class HP43XXGeneralStuff:

    @property
    def test_general(self):
        return "value"

 
    @test_general.setter
    def test_general(self, value):
        self.write(value)


class HP43XXChannelStuff:

    @property
    def test_property(self):
        return "value"

 
    @test_property.setter
    def test_property(self, value):
        self.write(value)
 

class HP438AChannel(Channel, HP43XXChannelStuff):


    @HP43XXChannelStuff.test_property.setter
    def test_property(self, value):
        self.write(f"SET{self.id}")
        HP43XXChannelStuff.test_property.fset(self, value)


class HP438A(Instrument, HP43XXGeneralStuff):
    a = Instrument.ChannelCreator(HP438AChannel, "A")
    b = Instrument.ChannelCreator(HP438AChannel, "B")
 

class HP437B(Instrument, HP43XXGeneralStuff, HP43XXChannelStuff):
    pass


hp438a = HP438A(ProtocolAdapter(), name="Test")
hp438a.a.test_property = "Value"
hp438a.test_general = "other value"
 
hp437b = HP437B(ProtocolAdapter(), name="Test")
hp437b.test_property = "Value"
hp437b.test_general = "other value"

`

bmoneke:
A few, quick comments regarding inheritance without looking at the code in detail:
In your mockup, the xx channel / instrument could already inherit from Channel / Instrument, which would make multiple inheritance superfluous.
Would it be feasible to make the 437 (which has those additional commands) inherit from the 438?

A comment regarding testing without a device: for that reason we have protocol tests.
I do not know whether the instrument has tests. If not, you could write tests, before you touch the instrument. After changing it, the tests have to pass again. That ensures, that the communication did not change during code refactoring.

neuschs:
> In your mockup, the xx channel / instrument could already inherit from Channel / Instrument, which would make multiple inheritance superfluous.

Sadly no, because the 437B is no multiple channel device. It has only one power sensor connector and therefore shouldn't offer the channel namespace.
But the 438A has two channels.

> Would it be feasible to make the 437 (which has those additional commands) inherit from the 438?
The 438 has additional commands and has two channs.


> A comment regarding testing without a device: for that reason we have protocol tests.
> I do not know whether the instrument has tests. If not, you could write tests, before you touch the instrument. After changing it, the tests have to pass again. That ensures, that the communication did not change during code refactoring. 
I agree !

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+        self.write("EX")
+        return float(display_content[6:12])
+
+    @duty_cycle.setter
+    def duty_cycle(self, duty_cycle):
+        values = [0.001, 99.999]
+        strict_range(float(duty_cycle), values)
+
+        self.write("DY%02.3fPCT" % float(duty_cycle))
+        self.check_errors()
```
LongnoseRob, line 327: While the documentation talks of a duty cycle in %, wouldn't it make more sense to represent the value of the dutycycle in python as an absolute figure?
Instead of 5 (%) --> 0.05 or 23.45 (%) --> 0.2345

LongnoseRob (approved):
Thank you for the contribution, in gerneral it looks good to me.
I left some comments on certain points where iimprovement could be discussed.

pymeasure/instruments/hp/hp437b.py
neuschs: I agree

LongnoseRob (approved):
Thanks for the latest changes, LGTM

tests/instruments/hp/test_hp437b.py
```diff
@@ -0,0 +1,121 @@
...
+    with expected_protocol(
+            HP437B,
+            [("SE%dEN" % int(sensor_type.value), None),
+             ("ERR?", "000")],
+    ) as instr:
+        instr.sensor_type = sensor_type
+
+
+def test_range():
+    "000000050010001A1002000001"
```
bmoneke, line 121: What does this test test?

Is it unfinished?

I suggest to either remove or finish it.

tests/instruments/hp/test_hp437b_with_device.py
```diff
@@ -0,0 +1,220 @@
...
+
+    This test suite, needs the following setup to work properly:
+        - A HP437B device should be connected to a gpib capable adapter;
+        - The device's address must be set in the RESOURCE constant.
+        - Some power sensor connected to the power reference reading around 0 dBm (+/- 1 dBm)
+    """
+
+    RESOURCE = 'USB0::0x03EB::0x2065::HEWLETT-PACKARD_437B__2.0::0::INSTR'
+
+    instr = HP437B(RESOURCE)
```
bmoneke, line 47: See the [documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#device-tests) regarding how to give a device address to the test environment.
This additional property will also ensure, that the tests are skipped without a connected device.

tests/instruments/hp/test_hp437b.py
bmoneke, line 1: Thanks for adding protocol tests.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+class MeasurementUnit(Enum):
+    """Enumeration to represent the measurement unit the power meter will measure in"""
+
+    Watts = 0
+
+    dBm = 1
+
+    Percent = 2
+
+    dB = 3
```
bmoneke, line 20: Typically, enum members are written with all caps: "WATTS", "DBM", but you might leave them as they are.

Regarding the empty lines: I suggest you to remove them, as the members are all related.

```suggestion
    Watts = 0
    dBm = 1
    Percent = 2
    dB = 3
```

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+from enum import Enum, IntFlag
+
+import logging
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class MeasurementUnit(Enum):
```
bmoneke, line 11: Did you consider using an `IntEnum`?

I have not yet seen your code, but an IntEnum behaves like an `int`, which might become handy.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+    70: "Keyboard/Display controller failure",
+    71: "Keyboard data failure",
+    72: "Data line to A3U26 is open.",
+    73: "Keyboard/Display controller self-test failure",
+    74: "Display not responding",
+    75: "Digital failure",
+}
+
+
+class StatusMessage:
```
bmoneke, line 136: you meant to make it an Enum, didn't you?

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+    def __init__(self, adapter, name="Hewlett-Packard HP437B", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            includeSCPI=False,
+            send_end=True,
+            **kwargs,
+        )
+
+    def __getstatus(status_type: StatusMessage, modifier=lambda v: v):
```
bmoneke, line 177: in pymeasure we're not using type hints (yet).
status_type is a tuple, isn't it?

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+        "*ESR?",
+        """
+        Get the status byte and Master Summary Status bit.
+
+        .. code-block:: python
+
+            print(instr.request_service_conditions)
+            StatusRegister.PowerOn|CommandError
+        """,
+        get_process=lambda v: EventStatusRegister(int(v))
```
bmoneke, line 203: Just that you know: you can use the `cast` attribute to make the integer conversion:
`cast=int`.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+        values={True: "DA", False: "DE"}
+    )
+
+    display_user_message = Instrument.setting(
+        "DU %s",
+        """
+        Set a custom user message up to 12 alpha-numerical chars. If the string is empty or None
+        the user message gets disabled.
+        """,
+        validator=lambda x, y: x if str(x).isalnum() and len(str(x)) <= 12 else str(x)[0:11],
```
bmoneke, line 271: `str(x)[0:11]` will return the first 11 chars, not the twelfth (at index 11) one. So you should use `[0:12]` or `[:12]` (a not written start index defaults to 0) if you want to use the first 12 chars.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+            -0.23  dB REL
+
+        """,
+        cast=str
+    )
+
+    duty_cycle_enabled = Instrument.control(
+        "SM", "DC%d",
+        """
+        Set duty cycle active or inactive. See :attr:`duty_cycle`
```
bmoneke, line 292:
```suggestion
        Control whether the duty cycle is active or inactive. See :attr:`duty_cycle`
```
You can read it as well, therefor "Control"

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+        Get if the lower limit check got triggered.
+        """,
+        map_values=True,
+        values={True: 2, False: 0},
+        cast=int,
+        get_process=__getstatus(StatusMessage.LimitsStatus),
+    )
+
+    power = Instrument.measurement(
+        "",
```
bmoneke, line 462: Maybe add a comment, that an empty string queries for the measurement, because the code looks as if you had forgotten to specify the command string.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+        """
+        Measure the power at the power sensor attached to the power meter in the corresponding unit.
+        In case a measurement would be invalid the power meter responds with the value +9.0200e+40.
```
bmoneke, line 465: You could add a get_process, which replaces that number with `float("nan")`, if you like.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+    def reset(self):
+        self.write("*RST")
+
+    def clear_status_registers(self):
+        self.write("*CLS")
+
+    def preset(self):
+        """
+        Sets the power meter to a known state. Preset
+        conditions are shown in the following table.
```
bmoneke, line 526: known state? Not better a "factory defined state"?

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+
+            instr.relative_mode_enabled = False
+            instr.log_lin = LogLin.Linear
+        """,
+        validator=strict_discrete_set,
+        values=[e for e in LogLin],
+        set_process=lambda v: "LN" if int(v.value) == 0 else "LG",
+        get_process=__getstatus(StatusMessage.LinearLogStatus, lambda v: {int(f.value): e for (
+            e, f) in zip(LogLin, LogLin)}[v])
+    )
```
bmoneke, line 615: what about using `log_enabled` instead?
That way you have linear as the default and you can enable/disable logarithmic options. That makes it more clearer for the code: `log_enabled = True`, for example.

You could name it also `logarithmic_display_enabled` or anything similar.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,949 @@
...
+
+            self.write("EN")
+            self.check_errors()
+
+        self.write("EX")
+        return frequency_data, cal_fac_data
+
+    def sensor_data_write_id_label(self, sensor_id, label):
+        """
+        Set a particular power sensor’s ID label. table to be modified. The sensor ID label must not
```
bmoneke, line 794:
```suggestion
        Set a particular power sensor’s ID label table to be modified. The sensor ID label must not
```

bmoneke (changes requested):
Thanks for your contribution, I commented on a few points to improve.

pymeasure/instruments/hp/hp437b.py
neuschs: The question probably boils down to if the user shall be able to specify a integer for the measurement unit. What do you think ?

pymeasure/instruments/hp/hp437b.py
neuschs: That's not exactly the case, some settings remain while others get actually cleared

pymeasure/instruments/hp/hp437b.py
neuschs: I like the idea to adhere to the common code style in this repository. But I will do it the opposite and call if linear_display_enabled in the RF world logarithmic is way more common used.

pymeasure/instruments/hp/hp437b.py
bmoneke: What I meant is, that you simplify checks in your code.

For example, below you do `e: int(f.value)`, but instead, you can reduce it (for an IntEnum) to `e: e for e in MeasurementUnit` as the comparison will be an integer comparison automatically.

For the user, nothing changes.

pymeasure/instruments/hp/hp437b.py
neuschs: Not this one. That one is only for internal use.

pymeasure/instruments/hp/hp437b.py
neuschs: Thanks

pymeasure/instruments/hp/hp437b.py
```diff
@@ -670,9 +669,8 @@ def resolution(self, resolution):
 
         """,
         validator=strict_discrete_set,
-        values={e: int(f.value) for (e, f) in zip(SensorType, SensorType)},
-        set_process=lambda v: SensorType(v) if isinstance(v, int) else v,
-        map_values=True,
+        values=[e for e in SensorType],
+        set_process=lambda v: int(v),
```
bmoneke, line 673: That's the nice thing about `IntEnum`:
`"SE%dEN" % SensorType.SOME_SENSOR_TYPE` will use exactly the (int) value, so you do not need the `set_process` at all!

neuschs:
@BenediktBurger Can you have a look why that Python 3.8 macos-latest test fails ?

BenediktBurger:
> @BenediktBurger Can you have a look why that Python 3.8 macos-latest test fails ?

The test succeeded.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,945 @@
...
+        inserts a settling-time delay before taking the requested measurement.
+        This settling time allows the internal digital filter to be updated with new values to
+        produce valid, accurate measurement results. The trigger with delay command allows time
+        for settling of the internal amplifiers and filters. It does not allow time for power
+        sensor delay. In cases of large power changes, the delay may not be sufficient for
+        complete settling. Accurate readings can be assured by taking two successive measurements
+        for comparison. Once the measurement results are displayed and read onto the bus,
+        the power meter reverts to standby mode.
+        """
+        self.trigger_delay()
```
BenediktBurger, line 932: That is an infinite recursion, as it calls itself again!.

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,945 @@
...
+    def __init__(self, adapter, name="Hewlett-Packard HP437B", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            includeSCPI=False,
+            send_end=True,
+            **kwargs,
+        )
+
+    def __getstatus(status_type, modifier=lambda v: v):
```
BenediktBurger, line 170: Methods with double underscore are special, just use a single underscore.

It's a staticmethod (does not depend on class or instance), add a `@staticmethod` decorator, please.

BenediktBurger (approved):
Two more small changes, please.

pymeasure/instruments/hp/hp437b.py
neuschs: Yeah thanks, that was a mistake ...

pymeasure/instruments/hp/hp437b.py
```diff
@@ -0,0 +1,944 @@
...
+    def __init__(self, adapter, name="Hewlett-Packard HP437B", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            includeSCPI=False,
+            send_end=True,
+            **kwargs,
+        )
+
+    def _getstatus(status_type, modifier=lambda v: v):
```
BenediktBurger, line 170: > THought about that already, that wouldn't work:
  get_process=_getstatus(StatusMessage.DutyCycleStatus), E   TypeError: 'staticmethod' object is not callable

The following does not work?
```suggestion
    @staticmethod
    def _getstatus(status_type, modifier=lambda v: v):
```

I made a small minimum working example and it works for me:
~~~python
from pymeasure.adapters.protocol import ProtocolAdapter

class Whatever(Instrument):

    @staticmethod
    def static(f):
        def xy(v):
            return f * v
        return xy

    prop = Instrument.control("G", "S", "docs", get_process=static(2))


inst = Whatever(ProtocolAdapter([("G", "1")]), name="abc")
p = inst.prop
assert p == 2
~~~

neuschs: I will fiddle around later, but the issue occured during pytest execution.

BenediktBurger: An alternative is to move that method outside of the class.

BenediktBurger:
May I merge?

neuschs:
Sure go on

PR #982, opened 2023-10-23
Add find_serial_port to find a serial port by providing USB information

### LINKED ISSUE #950: Find USB device from hardware information
The COM port of virtual Com ports via USB may change from time to time.
I propose a simple helper method, which retrieves the COM port of a USB virtual port according to hardware information (vendor id, product id, serial number).
This method allows to specify the hardware information and then to connect always to the correct device, independent of the COM port number assigned at that time.

Rough idea:

~~~python
from serial.tools import list_ports

def findPort(vid: int, pid: int, sn: str) -> str:
    """Find the port name ('COM5') of the first device with the given USB information.

    :param int vid: Vendor ID.
    :param int pid: Product ID.
    :param str sn: Serial number.
    :return: Port as a string as `device` attribute returned from `list_ports.comports`.
    """
    for port in sorted(list_ports.comports()):
        if port.vid == vid and port.pid == pid and port.serial_number == str(sn):
            return port.device
    raise AttributeError("No device found for the given data.")
~~~

bmoneke:
Closes #950 

I did not add "::INSTR" at the end of the resource string, which is optional.

Reasoning for adding it to instruments.__init__: There is potential to use it often, therefore import should be shorter.
Additionally, the module `resources` is already imported and the additional dependency `serial` was already loaded by pyvisa (if pyvisa-py is used), therefore, it should not increase import time.

### DIFF
--- a/docs/tutorial/connecting.rst
+++ b/docs/tutorial/connecting.rst
@@ -28,6 +28,11 @@ Then construct an object by passing the VISA address. For this example we connec
         from pymeasure.instruments import list_resources
         list_resources()
 
+    If you know the USB properties (vendor id, product id, serial numer) of the serial device, you can query for the VISA resource string::
+
+        from pymeasure.instruments.resources import find_serial_port
+        resource_name = find_serial_port(vendor_id=15, product_id=0x12e5, serial_number="sn56X")
+
 For instruments with standard SCPI commands, an :code:`id` property will return the results of a :code:`*IDN?` SCPI command, identifying the instrument. ::
 
     sourcemeter.id
--- a/docs/tutorial/connecting.rst
+++ b/docs/tutorial/connecting.rst
@@ -30,7 +30,7 @@ Then construct an object by passing the VISA address. For this example we connec
 
     If you know the USB properties (vendor id, product id, serial numer) of the serial device, you can query for the VISA resource string::
 
-        from pymeasure.instruments.resources import find_serial_port
+        from pymeasure.instruments import find_serial_port
         resource_name = find_serial_port(vendor_id=15, product_id=0x12e5, serial_number="sn56X")
 
 For instruments with standard SCPI commands, an :code:`id` property will return the results of a :code:`*IDN?` SCPI command, identifying the instrument. ::
--- a/pymeasure/instruments/resources.py
+++ b/pymeasure/instruments/resources.py
@@ -69,10 +69,15 @@ def find_serial_port(vendor_id: Optional[int] = None, product_id: Optional[int] 
 
     Use `None` if you do not want to check for that property.
 
+    .. code-block:: python
+
+        resource_name = find_serial_port(vendor_id=1256, serial_number="SN12345")
+        dmm = Agilent34410(resource_name)
+
     :param int vid: Vendor ID.
     :param int pid: Product ID.
     :param str sn: Serial number.
-    :return: Port as a VISA string for a serial device (e.g. "ASRL5" or "ASRL/dev/ACMO5").
+    :return: Port as a VISA string for a serial device (e.g. "ASRL5" or "ASRL/dev/ttyACM5").
     """
     for port in sorted(list_ports.comports()):
         if ((vendor_id is None or port.vid == vendor_id)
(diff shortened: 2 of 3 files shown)


docs/tutorial/connecting.rst
```diff
@@ -28,6 +28,11 @@ Then construct an object by passing the VISA address. For this example we connec
         from pymeasure.instruments import list_resources
         list_resources()
 
+    If you know the USB properties (vendor id, product id, serial numer) of the serial device, you can query for the VISA resource string::
+
+        from pymeasure.instruments import find_serial_port
+        resource_name = find_serial_port(vendor_id=15, product_id=0x12e5, serial_number="sn56X")
+
```
msmttchr, line 35: Regarding USB, I understand vid and pid, but I am not sure whether serial number is a generic concept. Do you know what it means in usb terms?

pymeasure/instruments/resources.py
```diff
@@ -58,3 +61,24 @@ def list_resources():
...
+    :return: Port as a VISA string for a serial device (e.g. "ASRL5" or "ASRL/dev/ACMO5").
+    """
+    for port in sorted(list_ports.comports()):
+        if ((vendor_id is None or port.vid == vendor_id)
+                and (product_id is None or port.pid == product_id)
+                and (serial_number is None or port.serial_number == str(serial_number))):
+            # remove "COM" from windows serial port names.
+            port_name = port.device.replace("COM", "")
+            return "ASRL" + port_name
+    raise AttributeError("No device found for the given data.")
```
msmttchr, line 84: I am not 100% sure, but I think that the association between the COM number and ASRL number is not always reliable. I mean that COM4 can be ASRL5.

pymeasure/instruments/resources.py
```diff
@@ -58,3 +61,24 @@ def list_resources():
             print(e)
     rm.close()
     return instrs
+
+
+def find_serial_port(vendor_id: Optional[int] = None, product_id: Optional[int] = None,
+                     serial_number: Optional[str] = None) -> str:
```
msmttchr, line 67: From old discussion, I think we are not using type hinting in pymeasure, but @bilderbuchi may know more about this.

msmttchr (commented):
Few comments from my side. Thanks

docs/tutorial/connecting.rst
bmoneke: Almost all USB devices have these three hardware properties: a vendor id, a product id and a serial number. For USB-serial port adapters (as you find them for example in an Arduino or many other devices connected via USB but talking a serial protocol), the vendor is the producer of that adapter and these chips have a serial number (different of the device you talk to, for example a laser).

I know only one company not providing serial numbers for their devices. all other devices in our lab have serial_number.

My procedure is: connect the device, recognize the com port, then I read the hardware info (the ids and the serial number). Afterwards I can recognize that device by the hardware info and not the comport anymore.

pymeasure/instruments/resources.py
bmoneke: Then how do you know the VISA name of a device?
Afaik, for each serial port, there is an ASRLport with that name (so ASRL5 for the **windows** COM5 or ASRL/dev/ACM0 for /dev/ACM0), but that the backend can introduce additional aliases, such that NI has the **VISA** COM5 linked to ASRL5.

bmoneke: You can give a VISA name "COM4" to the VISA port "ASRL5" which is linked to the fifth windows COM port (COM5), but the connection between the com-Port and ASRL5 is fixed.
At least in NI-MAX it was shown to be linked, but an alias (defaulting to "COM5") can be set.

pymeasure/instruments/resources.py
bmoneke: I stripped the hints.

pymeasure/instruments/resources.py
msmttchr: @bmoneke,thanks for the clarification. I remember having problems with this association in the past, but it could have been something wrong on my side.

bmoneke:
@msmttchr that means, we can merge?

msmttchr (approved):
OK to merge for me. Thanks.

PR #988, opened 2023-11-21
Improvements to pendulum CNT-91

bleykauf:
This PR implements several improvements for the pendulum CNT-91 counter:

* much faster readout by increased baud rate
* add back-to-back measurements
* replaced "measurement time" with "gate time" and using "gate time" instead of "sample rate" to be closer to what is written in den manual
* simplified readout procedures (removed unneccesary loops)
* added some constants min/max gate times and buffer sizes from the maual


bleykauf:
Looks like the way I implemented the passing of `baud_rate` fails some tests. I actually do not understand what the correct way to do this is.

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -35,20 +36,26 @@
...
     def __init__(self, adapter, name="Pendulum CNT-91", **kwargs):
-        kwargs.setdefault('timeout', 120000)
-        kwargs.setdefault('read_termination', '\n')
+        # allow long-term measurements, add 30 s for data transfer
+        kwargs.setdefault("timeout", 24 * 60 * 60 * 1000 + 30)
+        kwargs.setdefault("read_termination", "\n")
+        adapter = VISAAdapter(adapter, **kwargs)
+        # could not pass this to VISAAdapter, raises ValueError: 'baud_rate' is not a valid
+        # attribute for type USBInstrument
+        adapter.connection.baud_rate = 256000
```
bmoneke, line 58: You have to use connection (as it belongs only to serial, not to USB etc. connection) specific arguments:

~~~python
super().__init__(
    adapter,
    name,
    asrl={'baud_rate': 256000}),
    **kwargs)
~~~

bmoneke: Alternatively, you could just use `baud_rate=256000` in the super.__init__ call.
You should not create an adapter instance in the instrument's init, that is the problem.

bleykauf: Thanks, that was the piece of documentation I could not find.

bmoneke: I'm glad I could help

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -119,12 +109,12 @@ def read_buffer(self, expected_length=0):
         map_values=True,
     )
 
-    measurement_time = Instrument.control(
+    gate_time = Instrument.control(
```
BenediktBurger, line 112: If you change a property, please keep the old property and add a deprecation note. We do not want that someone's script breaks without notice.

For example:
~~~python
from warnings import warn

@property
def measurement_time(self):
    warn("measurement_time is deprecated, use `gate_time` instead.", FutureWarning)
    return self.gate_time
~~~

This should be noted in the changelog.md as well.

BenediktBurger (changes requested):
Please ensure backward compatibility (I added an example) deprecating features before removal.

In order to test the new methods, I suggest to add protocol tests, see our [documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#protocol-tests). You could use the Test generator, in order to make it easier, as it will record a test with live data.

BenediktBurger:
As you improve that instrument, you could rework the property docstrings according to https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings .
In this case, you can remove the instrument from the "grandfathered" list in the `test_all_instruments.py` test file.

We'll merge the PR even with the old docstrings, though. So don't worry if you don't have time

pymeasure/instruments/pendulum/cnt91.py
bleykauf: I added deprecation warnings for setter/getter of `measurement_time` and for use of `sample_rate` in `buffer_frequency_time_series`. For the latter, I had to turn some arguments keyword-only by the use of `*`. This is might not be ideal for the use of either `gate_time` or `sample_rate` (users might want to specify this as a positional argument) but it guarantees that old scripts will accidentally use incorrect  gate times or sample rates (an exception is thrown in the case that `sample_rate` was provided as a positional argument and the user can change the script accordingly).

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -144,45 +145,46 @@ def read_buffer(self, expected_length=0):
...
         Record a time series to the buffer and read it out after completion.
 
         :param channel: Channel that should be used
         :param n_samples: The number of samples
-        :param sample_rate: Sample rate in Hz
-        :param trigger_source: Optionally specify a trigger source to start the
-            measurement
+        :param gate_time: Gate time in s
+        :param trigger_source: Optionally specify a trigger source to start the measurement
+        :param back_to_back: If True, the buffer measurement is performed back-to-back.
```
BenediktBurger, line 178: You should keep the deprecated parameter in the list, but with a deprecation notice:
```suggestion
        :param back_to_back: If True, the buffer measurement is performed back-to-back.
        :param sample_rate: Sample rate in Hz

            .. deprecated:: 0.14
                Use parameter `gate_time` instead.
```

pymeasure/instruments/pendulum/cnt91.py
BenediktBurger: Thanks.

BenediktBurger (approved):
One suggestion and a small addition (deprecation in the documentation).

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -144,45 +145,46 @@ def read_buffer(self, expected_length=0):
...
     def buffer_frequency_time_series(
-        self, channel, n_samples, sample_rate, trigger_source=None
+        self,
+        channel,
+        n_samples,
+        *,
+        gate_time,
+        trigger_source=None,
+        back_to_back=True,
+        sample_rate=None,
```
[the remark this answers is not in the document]
BenediktBurger: It's up to your discretion.

pymeasure/instruments/pendulum/cnt91.py
```diff
@@ -105,84 +93,107 @@ def read_buffer(self, expected_length=0):
...
         :param channel: Channel that should be used
         :param n_samples: The number of samples
+        :param gate_time: Gate time in s
+        :param trigger_source: Optionally specify a trigger source to start the measurement
+        :param back_to_back: If True, the buffer measurement is performed back-to-back.
         :param sample_rate: Sample rate in Hz
-        :param trigger_source: Optionally specify a trigger source to start the
-            measurement
+            .. deprecated:: 0.14
+                Use parameter `gate_time` instead.
```
BenediktBurger: I think you have to add an empty line for sphinx (that is the linting error):

```suggestion

            .. deprecated:: 0.14
                Use parameter `gate_time` instead.
```

bleykauf: Indeed! Passes the tests now.

pymeasure/instruments/pendulum/cnt91.py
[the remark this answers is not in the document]
bleykauf: Changed it to be fully backwards compatible.

bleykauf:
Docstrings are now updated, backwards-compabability and deprecation warnings are established.

I will set up an instrument and create some tests as suggested in https://github.com/pymeasure/pymeasure/pull/988#pullrequestreview-1821933206 as I find time and will report back as soon as this is done.

bleykauf:
The test generator is a real time saver! Protocol tests are generated and added. I did discover two bugs this way (wrong minimum gate time, missing mapping of `format`) that have already been fixed. 

Should I add the script to generate the tests as well somehow?

BenediktBurger:
Good to hear, that the TestGenerator does work.

No need to add the script.

Are you done?

bleykauf:
Yes, ready from my side :)

BenediktBurger:
Thanks.

PR #992, opened 2023-12-06
[instruments] HP34401A adding remote control command.

Rybok:
* Checking stop bit configuration, should be two according to spec.
    * Added parameter: remote_local_state, it is set to "REM" during init
      by the driver.
      Should be set to "LOC" to return to local front display control.

### DIFF
--- a/pymeasure/instruments/hp/hp34401A.py
+++ b/pymeasure/instruments/hp/hp34401A.py
@@ -25,7 +25,6 @@
 from warnings import warn
 from pymeasure.instruments import Instrument
 from pymeasure.instruments.validators import strict_discrete_set
-from pymeasure.adapters import SerialAdapter
 
 
 deprecated_text = """
@@ -75,11 +74,9 @@ class HP34401A(Instrument):
         super().__init__(
             adapter,
             name,
-            asrl={'baud_rate': 9600, 'data_bits': 7, 'parity': 2},
+            asrl={'baud_rate': 9600, 'data_bits': 8, 'parity': 0},
             **kwargs
         )
-        if type(adapter) == SerialAdapter:
-            self.init_serial()
 
     # Log a deprecated warning for the old function property
     voltage_ac = Instrument.measurement("MEAS:VOLT:AC? DEF,DEF",
@@ -312,15 +309,6 @@ class HP34401A(Instrument):
         values=["LOC", "LOCAL", "REM", "REMOTE", "RWL", "RWLOCK"],
         map_values=False,
     )
-
-    def init_serial(self):
-        self.adapter.write_termination = "\n"
-        self.adapter.read_termination = "\n"
-        self.remote_local_state = "REM"
-
-        serial_settings = self.adapter.connection.get_settings()
-        if serial_settings["stopbits"] != 2:
-            print(f"Serial connection should use 2 stopbits: {serial_settings}")
 
     def beep(self):
         """This command causes the multimeter to beep once."""
--- a/pymeasure/instruments/hp/hp34401A.py
+++ b/pymeasure/instruments/hp/hp34401A.py
@@ -304,7 +304,7 @@ class HP34401A(Instrument):
     # System related commands
     remote_control_enabled = Instrument.control(
         "SYST: ", "SYST:%s",
-        """Control whether remote is enabled.""",
+        """Control whether remote control is enabled.""",
         validator=strict_discrete_set,
         values={True: "REM", False: "LOC"},
         map_values=True,
--- a/pymeasure/instruments/hp/hp34401A.py
+++ b/pymeasure/instruments/hp/hp34401A.py
@@ -25,6 +25,7 @@
 from warnings import warn
 from pymeasure.instruments import Instrument
 from pymeasure.instruments.validators import strict_discrete_set
+from pymeasure.adapters import SerialAdapter
 
 
 deprecated_text = """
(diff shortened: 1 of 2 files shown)


pymeasure/instruments/hp/hp34401A.py
```diff
@@ -300,6 +303,25 @@ def init_trigger(self):
...
     # System related commands
+    remote_local_state = Instrument.setting(
+        "SYST:%s",
+        """ A string property that controls the remote/local state of the
+        function generator. Valid values are: LOC<AL>, REM<OTE>, RWL<OCK>.
+        This setting can only be set. """,
+        validator=strict_discrete_set,
+        values=["LOC", "LOCAL", "REM", "REMOTE", "RWL", "RWLOCK"],
+        map_values=False,
+    )
```
bmoneke, line 314: Is "RWL" relevant? If not, you could change it to a boolean "remote_control_enabled", which makes it a lot easier to understand: `remote_control_enabled = True`.

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -77,6 +78,8 @@ def __init__(self, adapter, name="HP 34401A", **kwargs):
             asrl={'baud_rate': 9600, 'data_bits': 7, 'parity': 2},
             **kwargs
         )
+        if type(adapter) == SerialAdapter:
+            self.init_serial()
```
bmoneke, line 82: The normal procedure is, that a user uses the default VISAAdapter or they create their own adapter instance. In the latter case, the user has to provide the correct settings to the adapter instance (for example the `SerialAdapter`).
Therefore, we do not add any other adapter specific things to an instrument except for the VISAAdapter via the `super().__init__()` call.

If the read/write termination is wrong, you might add it to the super init into the `asrl` dictionary.

Rybok: Will fix it this week.

pymeasure/instruments/hp/hp34401A.py
Rybok: Yes, for long measurements I find RWL very useful to provide protection against accidental reset of the remote control mode.

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, adapter, name="HP 34401A", **kwargs):
         super().__init__(
             adapter,
             name,
-            asrl={'baud_rate': 9600, 'data_bits': 7, 'parity': 2},
+            asrl={'baud_rate': 9600, 'data_bits': 8, 'parity': 0},
```
Rybok, line 77: According to [hp34401manual.pdf](https://engineering.purdue.edu/~aae520/hp34401manual.pdf) these are the default settings. Actually it should be 2 stop bits but using pyvisa one should use the enum provided.

Rybok:
# Test code

```
from pymeasure.instruments.hp import HP34401A
import time

# Create an instance of the HP34401A instrument
dmm = HP34401A("asrl/dev/ttyUSB0")

time.sleep(0.5)

#Important, without the REM no measurements will be read.
dmm.remote_local_state = "REM"

time.sleep(0.5)
print(f"ID: {dmm.id}")



dmm.function_ = "DCV"

for i in range(3):
    time.sleep(0.3)
    print(f"{i} Voltage: {dmm.reading}")  # -> Single float reading

time.sleep(0.3)
dmm.remote_local_state = "LOC"





```

pymeasure/instruments/hp/hp34401A.py
BenediktBurger: Another idea as a **suggestion**: You have two boolean properties: `remote_control_enabled`, `remote_lock_enabled`.

Btw. you do not have to specify `map_value=True`.

pymeasure/instruments/hp/hp34401A.py
BenediktBurger: you could specify the stop bits with `stop_bits=20` (afaik 20 is the integer value of the stop bits 2 enum in pyvisa).

pymeasure/instruments/hp/hp34401A.py
Rybok: Ok, fixed to use boolean properties. But, wouldn't work without map_values set to true. But I guess you meant it for the first case.

pymeasure/instruments/hp/hp34401A.py
Rybok: Added the stop bits plus a link to where this magic "20" number comes from.

pymeasure/instruments/hp/hp34401A.py
BenediktBurger: Sorry, I meant, `map_values=False` is the default value, therefore you do not have to specify it in the (outdated) code above.

pymeasure/instruments/hp/hp34401A.py
```diff
@@ -300,6 +302,22 @@ def init_trigger(self):
     )
 
     # System related commands
+    remote_control_enabled = Instrument.control(
+        "SYST: ", "SYST:%s",
+        """Control whether remote is enabled.""",
```
BenediktBurger, line 307:
```suggestion
        """Control whether remote control is enabled.""",
```

BenediktBurger (approved):
One tiny change left.
As a contributor, you can add yourself to the authors.txt file (already in this PR).

Thanks for your contribution.

Rybok:
> May I merge your code?

Yes, please do! :)

BenediktBurger:
Thanks for your contribution and patience.

PR #993, opened 2023-12-12
add example HP 34401A with RS232

d-el:
Add example with HP 34401A by RS232

### DIFF
--- a/examples/hp34401a_rs232.py
+++ b/examples/hp34401a_rs232.py
@@ -30,17 +30,11 @@ Run the program by changing to the directory containing this file and calling:
 python hp34401a_rs232.py /dev/ttyUSB0
 """
 
-import time
 import logging
-import sys
-from time import sleep
-import numpy as np
+import time
 import serial
 from pymeasure.adapters import SerialAdapter
 from pymeasure.instruments.hp import HP34401A
-from pymeasure.experiment import (
-    Procedure, FloatParameter, unique_filename, Results
-)
 import argparse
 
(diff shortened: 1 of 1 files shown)


examples/hp34401a_rs232.py
```diff
@@ -0,0 +1,81 @@
...
+                        parity=serial.PARITY_NONE,
+                        stopbits=serial.STOPBITS_TWO,
+                        bytesize=serial.EIGHTBITS,
+                        xonxoff=True,
+                        dsrdtr=True,
+                        rtscts=False,
+                        write_termination='\n',
+                        read_termination='\n',
+                        )
+meter = HP34401A(adapter)
```
BenediktBurger, line 59: why don't you just use the default connection (VISA)?
```suggestion
meter = HP34401A("ASRL" + args.serial)

```

BenediktBurger (commented):

What do you want to show with your example?

You can add yourself to the authors.txt.

examples/hp34401a_rs232.py
d-el: I'm facing a problem where the "SYSTem:REMOte" command is missing.
Only now I see the solution in the previous PR.

```
meter = HP34401A("ASRL/dev/ttyUSB0", timeout=5000, baud_rate=1200, data_bits=8, parity=0)
meter.write('SYSTem:REMote')
...
```

msmttchr: This could be probably added in the `__init__` method.

PR #994, opened 2023-12-13
Dev/princetoninstruments

NitramC:
Driver for the Princeton Instruments SP2300i spectrograph/monochromator.

It's a weirdly designed instrument... when connected via RS232 the commands are echoed back. Also, the instrument will not return the termination character until the command has been fully processed (sometimes 10's of seconds). The routes I took around these issues are a bit hacky, but hopefully it's still a useful starting point.

### DIFF
--- a/pymeasure/instruments/princetoninstruments/princetonSP2300i.py
+++ b/pymeasure/instruments/princetoninstruments/princetonSP2300i.py
@@ -1,0 +1,690 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2023 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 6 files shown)


pymeasure/instruments/princetoninstruments/princetonSP2300i.py
```diff
@@ -0,0 +1,697 @@
...
+    # NM/MIN
+    # ?NM/MIN
+    scan_rate = Instrument.control(
+        "?NM/MIN", "%0.3f NM/MIN",
+        """Control scan rate in wavelength/time.
+
+            Set as a pint compatible value.
+        """,
+        get_process=lambda v: ureg.Quantity(v),  # convert to quantity
+        set_process=lambda v: ureg.Quantity(v).m_as('nm/min'),
```
BenediktBurger, line 282: For ease of use, could you add some method, which assumes the correct unit, for example something like:
~~~python
def generate_magnitude_extractor(units)
    def get_quantity_magnitude(value):
        if isinstance(value, ureg.Quantity):
            return value.m_as(units)
        else:
            return value
    return get_quantity_magnitude


scan_rate = Instrument.control(...
    set_process=generate_magnitude_extractor("nm/min")
~~~

That will make it easier to use for people who do not use pint (yet).

For another instrument with pint, see #929

pymeasure/instruments/princetoninstruments/princetonSP2300i.py
```diff
@@ -0,0 +1,697 @@
...
+        check_set_errors=True,
+    )
+
+    # GRATING
+    # ?GRATING
```
BenediktBurger, line 287: That's already in the `control` defined, you do not need to repeat it in front.

pymeasure/instruments/princetoninstruments/princetonSP2300i.py
```diff
@@ -0,0 +1,697 @@
...
+    grating = Instrument.control(
+        "?GRATING", "%d GRATING",
+        """Control grating being used.
```
BenediktBurger, line 290:
```suggestion
        """Control which grating to use.
```

BenediktBurger (commented):
A few comments for the first round.

It is the second instrument to use pint, see #929 for the first one.

Some solutions might have to be discussed.

PR #995, opened 2023-12-14
Making DAQmx a subclass of Instrument

### LINKED ISSUE #662: DAQmx is not an Instrument
The DAQmx [class](https://github.com/pymeasure/pymeasure/blob/8086835d8a211268a26451c28a054770b3907d40/pymeasure/instruments/ni/daqmx.py#L60) is not an Instrument subclass, but a plain `class`.
We should investigate why that is, and subclass from Instrument if possible, as all our instruments are/should be `Instrument`s.

NitramC:
Pushing one possible way to make DAQmx a subclass of Instrument. Fixes #662 .

Pros:
1) DAQmx is a subclass of instrument
2) An Adapter class is used to make a connection to the device.

Cons:
1) There is a new custom NIAdapter
2) Additional dependency on NI-DAQmx Python package instead of simply using ctypes

### DIFF
--- a/pymeasure/adapters/ni.py
+++ b/pymeasure/adapters/ni.py
@@ -1,0 +1,119 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2023 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


NitramC:
Sorry, this pulle request was a bit prematrue.

PR #999, opened 2023-12-19
Fix SR830 missing get_buffer method

seb5g:
and change the misspelled pauseBuffer method
start_buffer was with the wrong boolean
start buffer with false should not add the STRD call

Overall, I wanted to try for fast acquisition and reading from the instrument but somehow the transfer is very slow. I will go for an external analog reading of the outputs. The function to start a scan and read data back are nonetheless working doing something like:

```
srs = SR830('GPIB0::8::INSTR')
srs.reset_buffer()
srs.start_scan()
ch1, ch2 = srs.fill_buffer(100)
```

but it takes about 3s to 20s to get the reading out even with the sample_rate set to 512 S/s ...

### DIFF
--- a/pymeasure/instruments/srs/sr830.py
+++ b/pymeasure/instruments/srs/sr830.py
@@ -503,7 +503,8 @@ class SR830(Instrument):
         return ch1, ch2
 
     def buffer_measure(self, count, stopRequest=None, delay=1e-3):
-        """ Start a fast measurement mode and transfers data from buffer to extract mean and std measurements
+        """ Start a fast measurement mode and transfers data from buffer to extract mean
+        and std measurements
 
         Return the mean and std from both channels
         """
@@ -589,7 +590,8 @@ class SR830(Instrument):
         return self.values(command)
 
     def save_setup(self, setup_number: int):
-        """Save the current instrument configuration (all parameters) in a memory referred to by an integer
+        """Save the current instrument configuration (all parameters) in a memory
+        referred to by an integer
 
         The integer must be comprised between 1 and 9 (included)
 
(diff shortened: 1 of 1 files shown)


pymeasure/instruments/srs/sr830.py
```diff
@@ -583,3 +588,39 @@ def snap(self, val1="X", val2="Y", *vals):
 
         command = "SNAP? " + ",".join(vals_idx)
         return self.values(command)
+
+    def save_setup(self, setup_number: int):
+        """Save the current instrument configuration (all parameters) in a memory
+        referred to by an integer
+
+        The integer must be comprised between 1 and 9 (included)
```
BenediktBurger, line 596: That is already written in the option below, I'd remove it.

pymeasure/instruments/srs/sr830.py
```diff
@@ -583,3 +588,39 @@ def snap(self, val1="X", val2="Y", *vals):
...
+        :param setup_number: the integer referring to the memory (between 1 and 9 (included))
+        """
+        if 1 <= setup_number <= 9:
+            self.write(f'RSET{setup_number:d};')
+
+    def start_scan(self):
+        """
+
+        :return:
+        """
```
BenediktBurger, line 618: If you do not return anything, just remove that line. On the other hand, it is good to write something into the docstring, if you have one.

Also below.

BenediktBurger (changes requested):
Thanks for fixing it.

Two small comments.

seb5g:
Alright I've simplified the docstrings and added some others

BenediktBurger:
Thanks for your fix.

PR #1001, opened 2023-12-19
Improve parameters.

BenediktBurger:
A small improvement in the `Parameters`.


BenediktBurger:
Does not work as expected

PR #112, opened 2017-09-17
Update graphical.rst

JuMaD:

PR #1004, opened 2023-12-19
Add support for Hewlett Packard 8753E VNA

Sionwage:
# I would like to add a basic HP8753E VNA support to pymeasure.

The functionality is not ready but I would like feedback as I go and there are some specifics for coding I would like assistance with finding documentation on how to handle these.

1. How do I handle default Prologix GPIB-USB adapter parameters to work with this device?

      I found these pages that address how to handle read/write terminations and other parameters for the various adapters but I am unsure how to apply this to the prologix gpib-usb adapter I utilize.
      
      * This issue states I shouldn't set specific adapter settings to make the instrument work:
        - https://github.com/pymeasure/pymeasure/issues/815
      * These pages makes it look like I should be configuring adapter settings:
        - https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#defining-default-connection-settings
        - https://pyvisa.readthedocs.io/en/latest/api/constants.html#pyvisa.constants.InterfaceType
      
      I am not sure if I should be using `asrl` or `gpib` for `pyvisa.constants.InterfaceType` since its a serial device connecting to GPIB.
      
      Unfortunately, it seems like if I am not defining the Prologix settings and commanding them when using the VNA, there will be unexpected behavior or it will timeout/error out.
      
      It looks like it is addressed here but not implemented here https://github.com/pyvisa/pyvisa/issues/112

2. If I utilize default Prologix GPIB-USB adapter parameters for one instrument, is the functionality to change these parameters on the fly for other instruments in place? IE, if I need '++auto 1' for one instrument but not another, with pymeasure send the command to the adapter if I communicate to alternating instruments and remember when it needs to change it?

      The #112 issue makes it look like different parameters needed with the Prologix adapter for multiple instruments are not handled yet with pymeasure or pyvisa.

3. Are we allowed to utilize cached properties to save time on communicating?

## My to-do list of actions still needed to polish this for others to use is: 

- [x] Need unit tests for this class that don't rely on having the instrument. Currently my unit tests rely on having the instrument on the bench and connected. I know there is some functionality at [Adding Instruments->Writing tests->Protocol tests->Test generator](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html) to create tests that don't demand the instrument be attached to the computer.
- [x] Remove adapter specific commands like my workarounds using the Prologix adapter from the class or be able to toggle the Prologix specific functions on/off when initializing the class.
- [x] Write comments for each of the functions clarifying them 
- [x] Start the comments for properties with ['Get', 'Measure', 'Set', or 'Control']
- [x] May need to move the `__init__` block to `__new__`?
- [x] Exceptions may need to use more specific exceptions from the pymeasure library
- [x] Probably need to remove the `__version` and `if __name__=="__main__":` blocks
- [x] Address other code style issues I am undoubtedly not adhering to.


pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+__version__ = "0.1.1"
+
```
BenediktBurger, line 26: As you said, we do not use file version strings.
```suggestion
```

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+from pymeasure.adapters import PrologixAdapter
+from pymeasure.instruments import Instrument, RangeException, discreteTruncate
+
+
+class HP8753E(Instrument):
+    """Represents the HP8753E Vector Network Analyzer
+    and provides a high-level interface for taking scans of the
+    scattering parameters.
+
+    Template copied from Agilent 8722 class in pymeasure
```
BenediktBurger, line 46: What do you want to tell with this line?

If the code is quite similar to the Agilent8722, could you inherit from the other one and just implement the changes?

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+
+    SCAN_POINT_VALUES = [3, 11, 21, 26, 51, 101, 201, 401, 801, 1601]
+    SCATTERING_PARAMETERS = ["S11", "S12", "S21", "S22"]
+    S11, S12, S21, S22 = SCATTERING_PARAMETERS
+
+    start_frequency = Instrument.control(
+        "STAR?",
+        "STAR %e Hz",
+        """ A floating point property that represents the start frequency
+        in Hz. This property can be set.
```
BenediktBurger, line 57: Regarding the docstrings see the [documentation](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings), especially:

Specific to properties, start them with “Control”, “Get”, “Measure”, or “Set” to indicate the kind of property, as it is not visible after import, whether a property is gettable (“Get” or “Measure”), settable (“Set”), or both (“Control”). In addition, it is useful to add type and information about [Restricting values with validators](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#validators) (if applicable) at the end of the summary line, see the docstrings shown in examples throughout the [Adding instruments](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/index.html#adding-instruments) section. For example a docstring could be `"""Control the voltage in Volts (float strictly from -1 to 1)."""`.

Please note, that not all instruments in our library adhere to that rule yet.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+        can be set.""",
+        cast=bool,
+    )
+
+    power = Instrument.control(
+        "POWE?",
+        "POWE%.2e",
+        """ A float that can be set from -70 to +10dBm to control output power from the VNA ports. 
+        This property can be set.""",
+        cast=lambda x: float(x),
```
BenediktBurger, line 103: `lambda x: float(x)` is exactly the same as `float`, which is the default value for the `cast` parameter, such that you do not have to use it at all.
```suggestion
```

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+    )
+
+    def __init__(self, adapter, name="Hewlett Packard 8753E Vector Network Analyzer", **kwargs):
```
BenediktBurger, line 106: Please place the init at the top of the class.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+        # this should probably be in __new__
```
BenediktBurger, line 107: No, normally you do not use `__new__`, init is correct.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+        super().__init__(adapter, name, **kwargs)
```
BenediktBurger, line 108: Please provide an information, whether the device uses SCPI or not (`includeSCPI` parameter).

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+            self.adapter.gpib_read_timeout = 500
+            self.adapter.connection.timeout = 700
+            self.adapter.eoi = 1
+            self.adapter.eos = "\n"
+            self.adapter.connection.read_termination = self.adapter.eos
+            self.adapter.connection.read_termination = "\r\n"
+            self.adapter.write("++eot_enable 0")
+            self.adapter.flush_read_buffer()
+            self.write("FORM4")
+            self.adapter.flush_read_buffer()
```
BenediktBurger, line 121: You asked, how to define the connection settings:

- First, the driver (this class) should define the communication expected from the computer without any Adapter in between, therefore you should define the GPIB parameters in the `super().__init__` call, see the [documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#defining-default-connection-settings).
- If you use a non default adapter (for example Prologix instead of the default VISAAdapter), you have to set up that adapter outside of this file, that is, in your personal code.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+
+    @cached_property
+    def id(self):
+        """Get the identification of the instrument."""
+        if self.SCPI:
+            result = self.ask("*IDN?").strip("\n")
+            self._manu, self._model, _, self._fw = result.strip("\r\n").split(",")
+            return result
+        else:
+            raise NotImplementedError("Non SCPI instruments require implementation in subclasses")
```
BenediktBurger, line 131: you should know, whether your instrument implements SCPI or not.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+            self.adapter.eoi = 1
+            self.adapter.eos = "\n"
+            self.adapter.connection.read_termination = self.adapter.eos
+            self.adapter.connection.read_termination = "\r\n"
+            self.adapter.write("++eot_enable 0")
+            self.adapter.flush_read_buffer()
+            self.write("FORM4")
+            self.adapter.flush_read_buffer()
+
+    @cached_property
```
BenediktBurger, line 123: What is your goal here? Making readout of the id faster? But how often do you read the id?

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+            if int(self.ask(f"{parameter}?")) == 1:
+                return parameter
+        return None
+
+    @parameter.setter
+    def parameter(self, value):
+        if value in HP8753E.SCATTERING_PARAMETERS:
+            self.write("%s" % value)
+        else:
+            raise Exception("Invalid scattering parameter requested" " for Hewlett Packard 8753E")
```
BenediktBurger, line 196: Can't you do that with our property creators (`Instrument.control`)?

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+        self.averaging_enabled = True
+
+    def is_averaging(self):
+        """Returns True if averaging is enabled"""
+        warnings.warn("Don't use is_averaging(), use averaging_enabled instead", FutureWarning)
+        return self.averaging_enabled
+
+    def restart_averaging(self, averages):
+        warnings.warn("Don't use restart_averaging(), use scan_single() instead", FutureWarning)
+        self.scan_single()
```
BenediktBurger, line 271: I figure you copied these methods from the other driver.

As they are deprecated, remove them, please.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+if __name__ == "__main__":
+    adapter = PrologixAdapter("ASRL4::INSTR", address=16, visa_library="@py")
+    # adapter = PrologixAdapter("ASRL4::INSTR", address=16)
+
+    vna = HP8753E(adapter)
+
+    vna.adapter.auto = 1
+    print(vna.id)
```
BenediktBurger, line 394: As you said, we do not use this code block, as everyone has his own configuration.

BenediktBurger (commented):
Thank you very much for your contribution. I commented on your code.

A few general points to consider:

- You might add yourself to the `authors.txt` file
- Please add documentation files (see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation) for the API documentation

Answers to your questions:

- "How do I handle default Prologix parameters": You have to define them in your personal code, not in pymeasure, as pymeasure defines the parameters for the instrument communication, such that anyone might use the device independent of the adapter, see the init part.
- "I am not sure if I should be using asrl or gpib for pyvisa.constants.InterfaceType since its a serial device connecting to GPIB.": `gpib` as that is the device connection.
- "Unfortunately, it seems like if I am not defining the Prologix settings and commanding them when using the VNA, there will be unexpected behavior or it will timeout/error out.": You have to create your `PrologixAdapter` with the correct settings beforehand.
- "If I utilize default Prologix GPIB-USB adapter parameters for one instrument, is the functionality to change these parameters on the fly for other instruments in place? IE, if I need '++auto 1' for one instrument but not another, with pymeasure send the command to the adapter if I communicate to alternating instruments and remember when it needs to change it?": You have to manage your PrologixAdapter yourself.
- "Are we allowed to utilize cached properties to save time on communicating?" Right now we do not use them. In case they might be good, we could discuss it. I've seen you cache the id/serial_number, but I don't see the point of caching, as you read these data rarely.

If you have more questions, don't hesitate to ask.

pymeasure/instruments/hp/hp8753e.py
Sionwage: Thank you so much for the quick feedback! I'll work on implementing your feedback in addition to my other to-do items.

pymeasure/instruments/hp/hp8753e.py
Sionwage: I could look to see if the commands are compatible or even close.

I think I need to import hplegacyinstrument at least as well.

pymeasure/instruments/hp/hp8753e.py
Sionwage: Removed `__version__`

pymeasure/instruments/hp/hp8753e.py
Sionwage: Honestly the Agilent8722 class could have been imported but all of the exceptions would need to be rewritten to reflect the proper VNA model being controlled.

BenediktBurger: What exceptions would have to be changed?
Wouldn't that be a lot less work now (creating the new driver and tests) and later on (maintaining one instead of two drivers)?

BenediktBurger: You have put a lot of limits in your code.
How confident are you, that your code (and the limits) apply to that other driver as well?

One approach would be to compare the settings in both manuals and decide on that base to use your driver for both instruments (maintaining the old, deprecated code in the agilent, though, for backward compatibility). 

I don't know whether you have the time for that work.

The backup solution is to add some line in each docstring, stating that the other device is quite similar.
That way someone, who knows the agilent could check whether your driver fits to the agilent and update the agilent one.
If you choose this way, you should probably open an issue as well.

Sionwage: Actually I first tried to set those limits on initialization but it would error out. I now realize I would need to redefine those properties in the `__init__` function or initialize them to default values and override them in the `__init__` function if not using the default values.

I want the doc strings and exceptions to spell out the limits for what gets defined when the class is run and not the hard-coded values I currently have written. I know for a fact that both the 8722 and 8753 have a plethora of frequency ranges they can cover based on options purchased for the instrument.

I would prefer if there was a GPIB command to ask the instrument what the Min/Max value limits are but I don't know if a command exists for the 8722 or 8753 to do that like more modern instruments can with various SCPI commands. I might be able to program the limits based on the options reported using a lookup table instead.

To start, I'll just work on allowing for configuring the limits to all of the properties I have limits defined.

Since I don't have a 8722 to test against, I would rather not depreciate that code since it probably works for those that do and I can't ensure my code would too. If I can figure out how to make the Bytesio function the 8722ES class uses in `data_complex` work with my Prologix I would probably be a little less hesitant.

BenediktBurger: Ok, let us not mix those two instruments (for this PR), so this PR touches only your new instrument.

> Actually I first tried to set those limits on initialization but it would error out. I now realize I would need to redefine those properties in the __init__ function or initialize them to default values and override them in the __init__ function if not using the default values.

For setting values later on, you could use `dynamic=True` in the properties, see [dynamic-properties documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#dynamic-properties).

> I want the doc strings and exceptions to spell out the limits for what gets defined when the class is run and not the hard-coded values I currently have written. I know for a fact that both the 8722 and 8753 have a plethora of frequency ranges they can cover based on options purchased for the instrument.

Our properties do not allow to change the documentation dynamically (see aforementioned documentation), however, if you use a `strict_range` validator, the error message will return the currently active limits.
In fact, in most cases a `strict` validator is better than a `truncated` one, as the latter does not give a feedback about the truncation, such that the user might erroneously think, that the value is valid.

> Since I don't have a 8722 to test against, I would rather not depreciate that code since it probably works for those that do and I can't ensure my code would too.

I did not mean to deprecate the whole 8722, but keep the methods of the 8722, which are already deprecated. As we do not combine those two instruments (at the moment). Don't worry about it anymore.

pymeasure/instruments/hp/__init__.py
BenediktBurger, line 1: You reordered them by numbers, didn't you?

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,390 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from time import sleep
+from time import time as now
+
+from io import BytesIO
```
BenediktBurger, line 28: You do not use that line.

If you look in the github code page (here) you see the linting issues marked.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,390 @@
...
+    SCAN_POINT_VALUES = [3, 11, 21, 26, 51, 101, 201, 401, 801, 1601]
+    SCATTERING_PARAMETERS = ["S11", "S12", "S21", "S22"]
+
+    start_frequency = Instrument.control(
+        "STAR?",
+        "STAR %e Hz",
+        f"""Control the start frequency in Hz. (float truncated from 30_000.0
+        to 6_000_000_000.0).
+        """,
+        validator=truncated_range,
```
BenediktBurger, line 78: As said in the other comment, `strict` is better than truncated, as the user will know, when the result differs from the input. With truncation, a user might give `start_frequency=10_000` and believe, that it worked.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,390 @@
...
+        return self._fw
+
+    def set_fixed_frequency(self, frequency):
+        """Set the sweep to be a fixed frequency in Hz. (float truncated from 30_000.0 to 6_000_000_000.0)"""
+        self.start_frequency = frequency
+        self.stop_frequency = frequency
+        self.scan_points = 3
+
+    @property
+    def parameter(self):
```
BenediktBurger, line 255: maybe "scattering_parameter" or even "active_scattering_parameter".

Could you do it with a property creator (`Instrument.control`)?

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,390 @@
...
+        return np.linspace(self.start_frequency, self.stop_frequency, num=self.scan_points)
+
+    @property
+    def data_complex(self, timeout=10):
+        """Gets the complex s-parameter measurements from the last scan.
+        This function is blocking until it is completed. (returns nparray
+        sized by number of points in sweep)
+        
+        Args:
+        timeout - a value in seconds to timeout the function
```
BenediktBurger, line 347: We use sphinx style to indicate parameters:
```suggestion
        :param timeout: timeout of the function in seconds
```
between ":param" and the parameter name, you can add the type, if you wish: ":param float timeout:".

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+class HP8753E(Instrument):
+    """Represents the HP8753E Vector Network Analyzer
+    and provides a high-level interface for taking scans of the
+    scattering parameters.
+
+    Keyword arguements:
+    adapter -- <placeholder for description of pymeasure.adapter>
+    name -- <str> describing the instrument
```
BenediktBurger, line 44: We use sphinx notation for the parameters (for the documentation page):

```suggestion
    :param adapter: A string, integer, or :py:class:`~pymeasure.adapters.Adapter` subclass object
    :param str name: Name of the instrument
```

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,390 @@
...
+    """
+
+    def __init__(
+        self,
+        adapter=None,
+        name=None,
```
BenediktBurger, line 50: Normally, the default name is the model of the instrument, in this case for example "HP8753E VNA" or just "HP8753E", but I see your idea here.

The idea is, that a user can give a name to all his instruments, even though they might be of the same model (therefore instruments must have a "name" parameter).

If you wish to know the serial number etc. you can always use the `id` method.

Therefore the `name` is user defined, while the `id` is device defined.

pyproject.toml
BenediktBurger, line 1: We haven't used poetry.

I'd rather not have another list of dependencies in another file. Therefore I'd like to exclude this file from your PR.

What do you think @bilderbuchi , @msmttchr , @CasperSchippers ?

pymeasure/instruments/ni/daqmx.py
BenediktBurger, line 1: Why do you change this file?

If it does not work, we should tackle it in another issue and PR.

tests/instruments/hp/test_hp8753e.py
BenediktBurger, line 1: These [protocol tests](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#protocol-tests) are quite easy:

base structure is:
~~~python
import pytest

from pymeasure.instruments.extreme5000 import Extreme5000

def test_voltage():
    """Verify the communication of the voltage getter."""
    with expected_protocol(
        Extreme5000,
        [(":VOLT 0.345", None),
         (":VOLT?", "0.3000")],
    ) as inst:
        inst.voltage = 0.345
        assert inst.voltage == 0.3
~~~

Here you send first ":VOLT 0.345" to the device, without reading (None), afterwards you send ":VOLT?", and read "0.3000" from the device. These strings might come from the manual as examples for setting the voltage to 0.345 V and reading the voltage of 0.3 V.

tests/instruments/hp/test_hp8753e.py
```diff
@@ -0,0 +1,32 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import numpy as np
+
+from pymeasure.adapters import PrologixAdapter
+from pymeasure.instruments.hp import HP8753E
+
```
BenediktBurger, line 29: Here an example, which should work:
```suggestion
from pymeasure.test import expected_protocol

def test_start_frequency_setter():
    with expected_protocol(
        HP8753E,
        [("STAR 1.2345e6 Hz", None)],
        name="test",  # in order to not trigger reading the id.
    ) as inst:
        inst.start_frequency = 123_456


def test_start_frequency_getter():
    with expected_protocol(
        HP8753E,
        [("STAR?", "123456")],
        name="test",  # in order to not trigger reading the id.
    ) as inst:
        assert inst.start_frequency == 123_456
```

tests/instruments/hp/test_hp8753e_with_device.py
```diff
@@ -0,0 +1,384 @@
...
+from pymeasure.instruments.hp import HP8753E
+
+# pytest.skip("Only works with connected hardware", allow_module_level=True)
+
+
+@contextmanager
+def init_prologix_adapter():
+    try:
+        prologix = PrologixAdapter(
+            resource_name="ASRL4::INSTR", address=16, visa_library="@py", auto=1
```
BenediktBurger, line 43: Instead of hardcoding the resource_name, we have a fixture ready, see the [device-tests documentation](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#device-tests).

BenediktBurger (changes requested):
A second round of comments.

Thanks for your work, which is already in a good state.

pyproject.toml
bilderbuchi: I agree with dropping that change. 

If poetry cannot pull the project metadata from `setup.cfg`, where it currently lives, we could (finally) migrate these to `pyproject.toml` -- we ned to make sure that all our supported Python and (especially) pip versions can deal with that, first. The reason we have not switched a long time ago was compatibility with old pip versions in use out in the "wild".
In any case, such a change should be a separate issue/PR.

Sionwage: I only uploaded the poetry dependencies in the `pyproject.toml` because I was sloppy with git and poetry is so easy to use.

pymeasure/instruments/ni/daqmx.py
Sionwage: I think this is Python 3.12 issue when importing which is why I changed it for testing my PR. I created https://github.com/pymeasure/pymeasure/issues/1008 identifying this issue.

BenediktBurger: It is best to keep different issues separate. Therefore, this change should not enter this PR.

pymeasure/instruments/hp/hp8753e.py
Sionwage: I can switch to "measurement_parameter" to allow both scattering parameters and absolute measurements like A, B, C, D, R1, R2, R3, R4 as used in other VNAs like the E5061/E5071.

BenediktBurger: Measurement parameter sounds fine too.

Sionwage:
Is there a way to change the branch to merge from? I have somehow completely hosed my git clone and cannot update/sync my copied repo with the changes made in pymeasure:master since opening this.

Otherwise I'll just copy out my code, nuke my fork repo, refork the repo and create a branch before putting my code back in so I can sync upstream changes.

BenediktBurger:
You cannot change the branch **name** to be merged from.

You can, however use different commits in the branch:
1. create a temporary branch, where you are at right now (and maybe the files)
2. delete your current branch
3. create a new branch with the name of the current branch (for example on the current master)
4. Do the changes you want (e.g. copying the files)
5. Push your changes to your github repo (using push force)

tests/instruments/hp/test_hp8753e_with_device.py
Sionwage: I'm not sure how to use that to pass both the PrologixAdapter address as well as the GPIB address.

Working on creating unit tests for my Agilent E4418B I kludged this workaround:

    # use this command to run the test without editing it
    # for prologix adapters
    # pytest '.\tests\instruments\agilent\test_agilentE4418B_with_device.py' --device-address="PrologixAdapter,ASRL4::INSTR,13"
    # everyone else, the `--device-address` flag works normally
    # pytest '.\tests\instruments\agilent\test_agilentE4418B_with_device.py' --device-address="GPIB::whatever::your::visa::says"
    
    
    @pytest.fixture(scope="module")
    def power_meter(connected_device_address):
        try:
            if "PrologixAdapter" not in connected_device_address:
                power_meter = AgilentE4418B(connected_device_address)
            else:
                _, prologix_address, gpib_address, *other_address_info = connected_device_address.split(",")
                prologix = PrologixAdapter(resource_name=prologix_address, visa_library="@py", auto=1)
                # need to ensure `eot_enable` is set to zero otherwise you will have to read twice to
                # get rid of the extra new line character
                prologix.write("++eot_enable 0")
                power_meter = AgilentE4418B(adapter=prologix.gpib(int(gpib_address)))
        except IOError:
            print("Not able to connect to power meter")
            assert False
    
        yield power_meter

Sionwage:
As I see it, I think there are two items still needing to get addressed before you would accept this pull request.

* PrologixAdapter fixture for unit-testing the HP8753E needs to be resolved
* May need more unit tests to cover the 40% or so of instrument code that isn't in a unit-test.

BenediktBurger:
> May need more unit tests to cover the 40% or so of instrument code that isn't in a unit-test.

Don't worry about less than 100% patch coverage.
Tests are good, but we do not have any limit for merging pull requests.

pymeasure/instruments/hp/hp8753e.py
Sionwage: `BytesIO` import line removed.

pymeasure/instruments/hp/__init__.py
Sionwage: I did but now the `__init__.py` is based on the state of the master branch with my insertion.

pymeasure/instruments/hp/hp8753e.py
Sionwage: Switched validator to use `strict_range` instead.

tests/instruments/hp/test_hp8753e_with_device.py
Sionwage: I'll wait for feedback on how to read the value from `pytest --device-address` and see if there is an appropriate way to pass all of the necessary information to use the `PrologixAdapter` in the fixture.

tests/instruments/hp/test_hp8753e.py
Sionwage: I used this method to get _some_ code coverage but I will have to work on improving it in the future. One of the biggest issues I have encountered is the instrument having an apparent non-deterministic delay on completing the same commands in the same sequence.

tests/instruments/hp/test_hp8753e.py
Sionwage: I used this method to get some code coverage.

pymeasure/instruments/ni/daqmx.py
Sionwage: File is left as is but the NI devices do cause trouble when trying to run pytest without having their drivers or python module installed which causes me to need to comment out NI instruments when trying to test offline.

pyproject.toml
Sionwage: I finally figured how to setup the test environment without poetry while creating this PR. It might be of value for me to type it up and add a section for more obtuse programmers like me to get up to speed on contributing.

Poetry and pyenv are so convenient for creating the necessary virtual environments and I feel like the barrier to entry are lower with those.

pymeasure/instruments/hp/hp8753e.py
Sionwage: Fixed doc string to use the sphinx style.

pymeasure/instruments/hp/hp8753e.py
Sionwage: Marking this one as resolved since there is another request higher in the comments requesting the implimentation gets changed to `Instrument.control`

pymeasure/instruments/hp/hp8753e.py
Sionwage: I added dynamic properties where appropriate. Otherwise I think everything in this conversation has been resolved.

pymeasure/instruments/hp/hp8753e.py
Sionwage: I need to add this to the class docstring still!

tests/instruments/hp/test_hp8753e_with_device.py
BenediktBurger: There is no support (yet) to use a prologix adapter in these tests.

We could modify the fixture to generate a prologix adapter if needed.

pymeasure/instruments/hp/hp8753e.py
Sionwage: I don't know how to rewrite this property to query the instrument for each possible measurement parameter and break when one of the properties finally responds with a '1'.

So if the instrument was set to "S22", I would need to ask it "S11?" which it will reply with "0". Same for "S12" and "S21". Once it gets asked "S22" and gets the '1' it should break. I wouldn't want to ask it "AR?" "BR?", "AB?", "MEASA?" etc for the other measurement parameters since that will just add time to complete the test.

It looks like `Instrument.control` doesn't have the hooks to iterate through asking the potentially 10 queries.

tests/instruments/hp/test_hp8753e_with_device.py
Sionwage: Let me know if this can stay as is or what to change it to. Since I have the Prologix adapter, I can't use the test fixture to code it.

pymeasure/instruments/hp/hp8753e.py
Sionwage: I can remove the extra fluff of querying the device to name it if it is desired for consistency. Let me know if you want to keep it or toss it.

I'm inclined to keep it unless it would cause something to break by having two devices with the same name or something.

Sionwage:
I think I am close to done with all the requests made and just need answers to the last 3 unresolved conversations to finish.

pymeasure/instruments/hp/hp8753e.py
BenediktBurger: Yes the general "Control" does not do that. Therefore, it is fine to use your custom property.

BenediktBurger:
Hey @Sionwage,
I'm in the process of reviewing, but I'll need some time, due to time constraints.
I ask for your patience.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+    )
+
+    def set_sweep_time_fastest(self):
+        """Set instrument scan sweep time to select fastest possible time."""
+        self.write("SWEA")
+
+    averages = Instrument.control(
+        "AVERFACT?",
+        "AVERFACT%d",
+        """Control the number of averages for a scan sweep. (int truncated from 1 to 999).
```
BenediktBurger, line 176:
```suggestion
        """Control the number of averages for a scan sweep. (int strict from 1 to 999).
```

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+
+    id = Instrument.measurement(
+        "*IDN?",
+        """Get the identification of the instrument (list of strings)""",
+        cast=str,
+    )
+
+    sn = Instrument.measurement(
+        "OUTPSERN",
+        """Get the serial number for the instrument""",
```
BenediktBurger, line 269:
```suggestion
        """Get the serial number of the instrument""",
```

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+    )
+
+    options = Instrument.measurement(
+        "OUTPOPTS",
+        """Get the installed options for the instrument""",
```
BenediktBurger, line 274:
```suggestion
        """Get the installed options of the instrument""",
```

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+        """Get the firmware of the instrument."""
+        if self._fw == "":
+            self._manu, self._model, _, self._fw = self.id
+        return self._fw
+
+    def set_fixed_frequency(self, frequency):
+        """Set the sweep to be a fixed frequency in Hz."""
+        self.start_frequency = frequency
+        self.stop_frequency = frequency
+        self.scan_points = 3
```
BenediktBurger, line 302: What about "set_single_frequency_scan"?

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+
+    @property
+    def measuring_parameter(self):
+        """Get the active Scattering or Measuring Parameter being measured.
```
BenediktBurger, line 306: Beware, the getter's docstring will be used for the whole property!

Therefore, you should use "Control" here and add also the content of the setter's docstring, which will be invisible later (and you can remove it).

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+        (int from the set {ALLOWED_BANDWIDTH}).""",
+        cast=lambda x: int(float(x)),
+        validator=strict_discrete_set,
+        values=ALLOWED_BANDWIDTH,
+    )
+
+    def reset(self):
+        """Reset the instrument. May cause RF Output power to be enabled!"""
+        self.write("*RST")
+        sleep(0.25)
```
BenediktBurger, line 338: Why sleep? Is the device unresponsive? If that's the case, please add a comment.

Shouldn't the user consider it?

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def scan(self, timeout=10):
+        """Initiates a scan with the number of averages specified and
+        blocks until the operation is complete.
```
BenediktBurger, line 342: Docstrings should be in imperative mode:
```suggestion
        """Initiate a scan with the number of averages specified and
        block until the operation is complete.
```

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+                    f"VNA Scan took longer than {sweep_time} seconds to complete and timed out."
+                )
+
+        sleep(0.5)
+        while self.adapter.connection.bytes_in_buffer > 0:
+            _ = self.read()
+            sleep(0.25)
+
+    def scan_single(self):
+        """Initiates a single scan or N scans averaged based on averaging
```
BenediktBurger, line 384: Docstrings should also end with a period:
```suggestion
        """Initiate a single scan or N scans averaged based on averaging.
```

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+                pass
+
+        # process string data into complex numbers
+        preformat_data = [float(point[0]) + float(point[1]) * 1j for point in temp_data]
+        data = np.array(preformat_data)
+
+        return data
+
+    def shutdown(self):
+        """Shutdown - Disables RF Output."""
```
BenediktBurger, line 471:
```suggestion
        """Shutdown - Disable RF Output."""
```

BenediktBurger (changes requested):
Thanks for your contribution.

I mentioned some incosistencies and mainly documentation issues. The instrument looks already good.

I noticed quite a lot dynamic properties. Do you intend to use them as dynamic properties or is it "just in case"?

tests/instruments/hp/test_hp8753e.py
```diff
@@ -0,0 +1,149 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2023 PyMeasure Developers
```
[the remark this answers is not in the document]
Sionwage: Corrected

tests/instruments/test_all_instruments.py
```diff
@@ -70,6 +70,8 @@ def find_devices_in_module(module):
             # On Windows instruments.ni.daqmx can raise an OSError before ModuleNotFoundError
             # when checking installed driver files
             pass
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
Sionwage: This was missed as I have to add this because of the NI / nicau.dll issue for not having National Instruments drivers or visa installed on my PC.

This at least allows me to verify my git clone locally is running all of the tests correctly including the linting before I push the committed changes. I wish the NI instruments modules didn't punish people that don't use them and I think its a Python 3.11->3.12 change that causes the issue.

These two lines are removed.

pymeasure/instruments/hp/hp8753e.py
Sionwage: Changed the docstring for measuring_parameter getter to use 'Control' and removed the docstring to the setter.

pymeasure/instruments/hp/hp8753e.py
Sionwage: Fixed

pymeasure/instruments/hp/hp8753e.py
Sionwage: Added the `delay` parameter to the reset function with a default value of 0.25 seconds.

tests/instruments/hp/test_hp8753e_with_device.py
```diff
@@ -0,0 +1,373 @@
...
+        del prologix
+
+
+@contextmanager
+def init_HP8753E(**kwargs):
+    with init_prologix_adapter() as prologix:
+        vna = HP8753E(adapter=prologix, **kwargs)
+        yield vna
+        del vna
+
```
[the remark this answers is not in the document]
Sionwage: switched to calling `vna.shutdown()`

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+            # *IDN? to populate the name of the VNA
+            try:
+                self._manu, self._model, _, self._fw = self.id
+            except ValueError:
+                self._manu = "Hewlett Packard"
+                self._model = "8753E"
+            self._desc = "Vector Network Analyzer"
+            name = self.name = f"{self._manu} {self._model} {self._desc}"
+        else:
+            self.name = name
```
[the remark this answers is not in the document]
Sionwage: dropped the else clause

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+    )
+
+    averaging_enabled = Instrument.control(
+        "AVERO?",
+        "AVERO%d",
+        """Control whether or not averaging is enabled. (boolean)""",
+        cast=bool,
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, False: 0, "ON": 1, "OFF": 0, "1": True, "0": False},
```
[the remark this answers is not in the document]
Sionwage: Using `values={True: 1, False: 0},`

pymeasure/instruments/hp/hp8753e.py
Sionwage: changed it to `set_single_frequency_scan`

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,472 @@
...
+    )
+
+    correction_enabled = Instrument.control(
+        "CORR?",
+        "CORR%d",
+        """Control whether or not correction is enabled. (boolean)""",
+        cast=bool,
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, False: 0, "ON": 1, "OFF": 0},
```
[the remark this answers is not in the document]
Sionwage: Using `values={True: 1, False: 0},`

Sionwage:
Not sure how to resolve `WARNING: Field list ends without a blank line; unexpected unindent.`

Otherwise I think everything has been addressed now.

tests/instruments/hp/test_hp8753e_with_device.py
```diff
@@ -0,0 +1,338 @@
...
+    assert hp8753e.averages == 32
+    hp8753e.averages = 18
+    assert hp8753e.averages == 18
+    hp8753e.averages = 1
+    assert hp8753e.averages == 1
+    hp8753e.averaging_enabled = False
+
+
+# this test will wear out the mechanical relays and should be run only as necessary
+@pytest.mark.skip
```
BenediktBurger, line 282: You can add a reason to the skip:
```suggestion
@pytest.mark.skip(reason="This test will wear out the mechanical relays and should be run only as necessary")
```

tests/instruments/hp/test_hp8753e_with_device.py
BenediktBurger: Cool idea to use the Prologix in the resource-name!

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,484 @@
...
+        `HP8753E.stop_frequency`, `HP8753E.span_frequency`, `HP8753E.center_frequency`
+        on initialization (float in Hz).
+    :param max_frequency: Optional set the maximum validator for `HP8753E.start_frequency`,
+        `HP8753E.stop_frequency`, `HP8753E.span_frequency`, `HP8753E.center_frequency`
+        on initialization (float in Hz).
+    :param min_power: Optional set the minimum validator for `HP8753E.power` on initialization
+        (float in dBm).
+    :param max_power: Optional set the maximum validator for `HP8753E.power` on initialization
+        (float in dBm).
+    """
```
BenediktBurger, line 55: I think that is the culprit for the warning.
For such a list, sphinx expects the last line to start with `:param` or an empty line, so you could do:
- either move `(float ind dBm).` to the end of the previous line 53 (such that the last line starts with `:param...`
- or add an empty line before the end of the docstring.

We have a line length limit of 100, so the first option should work.

CHANGES.rst
BenediktBurger, line 1: Please revert the changes in this file.

PRs should be based on the master branch and merge there.

The v0.14 branch is there to write the changelog for the next release.

BenediktBurger (changes requested):
Some comments.

Yes, it comes good together, we're almost done.

tests/instruments/hp/test_hp8753e_with_device.py
```diff
@@ -40,17 +40,23 @@ def hp8753e(connected_device_address):
...
         else:
             _, prologix_address, gpib_address, *other_address_info = connected_device_address.split(
                 ","
             )
-            prologix = PrologixAdapter(resource_name=prologix_address, visa_library="@py", auto=1)
+
+            temp_prologix = PrologixAdapter(
+                resource_name=prologix_address, visa_library="@py", auto=1
+            )
+            prologix = temp_prologix.gpib(gpib_address)
```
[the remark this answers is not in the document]
Sionwage: Fixed. Thank you for being patient with me fixing the issues identified in my code. Hopefully the next instrument (Keysight E5071C) will go a bit smoother.

I think the functionality you exposed with this PR will help me to utilize inheritance for the multitude of shared functionality of the 5061 and 5071 VNAs since they can be configured with 2 or 4 ports and a plethora of different frequency ranges.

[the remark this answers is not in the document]
BenediktBurger: Thanks for considering more instruments and thanks for your patience and working along with this review process.

tests/instruments/hp/test_hp8753e_with_device.py
```diff
@@ -0,0 +1,349 @@
...
+            # get rid of the extra new line character
+            prologix.write("++eot_enable 0")
+            vna = HP8753E(adapter=prologix)
+
+    except IOError:
+        print("Not able to connect to vna")
+        assert False
+
+    yield vna
+    del vna
```
BenediktBurger, line 68: This is an interesting solution on how to use a prologix adapter for device tests.

BenediktBurger: This comment is for creating an additional issue.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,476 @@
...
+        "A": "MEASA",
+        "B": "MEASB",
+        "R": "MEASR",
+    }
+
+    start_frequency = Instrument.control(
+        "STAR?",
+        "STAR %e Hz",
+        """Control the start frequency in Hz. (float).
+        Property is dynamic and allows remapping min/max limits.
```
BenediktBurger, line 113: One last comment:
Looking at the dynamic properties, I noticed, that they add `(dynamic)` to the docstring, so you do not have to do it manually.

Sionwage: Do you want me to remove the `Property is dynamic and allows remapping min/max limits.` lines from the dynamic properties?

BenediktBurger: Yes, please.

pymeasure/instruments/hp/hp8753e.py
```diff
@@ -0,0 +1,476 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2023 PyMeasure Developers
```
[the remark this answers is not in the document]
Sionwage: Fixed

pymeasure/instruments/hp/hp8753e.py
Sionwage: Removed `Property is dynamic and allows remapping min/max limits.` Should be go now.

BenediktBurger:
Thank you for contributing!

Sionwage:
> Thank you for contributing!

Thank you for your patience!

PR #1005, opened 2023-12-20
Extract init communication

BenediktBurger:
In order to expand _all instrument tests_, I moved all the communication in the `__init__` methods into their own `_init_communication` method.
That allows to replace that method for certain tests.

The name of that method could be different, if required.

### DIFF
--- a/pymeasure/instruments/activetechnologies/AWG401x.py
+++ b/pymeasure/instruments/activetechnologies/AWG401x.py
@@ -445,6 +445,9 @@ class AWG401x_AFG(AWG401x_base):
     def __init__(self, adapter, **kwargs):
         super().__init__(adapter, **kwargs)
 
+        self._init_communication()
+
+    def _init_communication(self):
         model = self.id.split(",")[1]
         if model == "AWG4012":
             num_ch = 2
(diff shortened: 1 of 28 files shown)


BenediktBurger:
I got a new idea:
1. instead of defining a special method just for tests (which is bad practice to write the code for tests).
2. all instruments must have a parameter `_init_comm_pairs` which contains the communication pairs for tests

Reasoning:
- It is a small memory burden to carry a list.
- It does not require additional design considerations for the coding of the `__init__`
- The same list can be used for the protocol tests

ToDos, if this idea gets accepted:
- [ ] write changelog
- [ ] add documentation how to treat the instruments

BenediktBurger:
What do you think, @bilderbuchi ?

bilderbuchi:
> 2\. all instruments must have a parameter `_init_comm_pairs` which contains the init communication pairs for tests (defaults to an empty list)

I think test-related code should _not_ leak into the instrument codebase. With this change, now part of the protocol test definition resides in the test files, and part in the instrument _implementation_ (`_init_comm_pairs`) -- I find this split quite confusing/suboptimal. 

I'm not sure why this is necessary in the first place because I could only give this a skim-read? Especially, because in the test files you now extract it back to a module-level constant again, anyway? (e.g. `AFG_init_comm = AWG401x_AFG._init_comm_pairs`)

Is [this pattern](https://github.com/pymeasure/pymeasure/pull/1005/files#diff-b5a297187171979fe40d575c461421190192f3726e83e7a3bbf3df724da93d0bR204) the reason? Is this purely for test_all_instruments being able to test instruments that have some communication in `__init__`?

Can't we inject that `_init_comm_pairs` attribute in the test code and get an identical effect?
E.g. maintain a Defaultdict(None) of (cls: _init_comm_pairs) items somewhere in `tests/` and inject those if available? 
`cls(adapter=ProtocolAdapter(INITCOMMAP[cls]))`

A drawback of this is that the init comm pairs reside not with the test files of the relevant instrument. 
Instead, we could require to place those in a tuple with a specified name, and a lookup method tries to find if such a tuple exists in the `tests/instruments/manufacturer/instrument` module. It's a bit "spooky action at a distance" but pytest is full of this magic, anyway. :-)
`cls(adapter=ProtocolAdapter(look_up_initcomm(cls)))`

BenediktBurger:
> Is [this pattern](https://github.com/pymeasure/pymeasure/pull/1005/files#diff-b5a297187171979fe40d575c461421190192f3726e83e7a3bbf3df724da93d0bR204) the reason? Is this purely for test_all_instruments being able to test instruments that have some communication in `__init__`?

Exactly, the change is purely to be able to test instruments that communicate in their `__init__`.

Yeah, maybe we can find a way to extract the init_comm_pairs from the test files.

BenediktBurger:
@coderabbitai review

Issue #804, opened 2022-12-23
SCPIbaseMixin class instead of includeSCPI


BenediktBurger:
Ideas regarding instruments with common controls include Mixin classes which contain controls.

The first element are the base SCPI commands (right now included in `Instrument`)

I propose the following transition path:

1. Create a Mixin class and write the proper documentation (right now, the includeSCPI is not in the documentation!). I propose `SCPIbaseMixin` as class name, as it contains the SCPI base commands.
2. Deprecate `includeSCPI`
3. Add the SCPIbaseMixin to all instruments with `includeSCPI=True`, add `SCPIbaseUnknownMixin` to all instruments without an explicit definition of `includeSCPI`. The latter class is a subclass of the former one and raises a deprecation warning, such that we get information, whether the instrument is SCPI conform (changing to the SCPIbaseMixin) or not (removing the Mixin).

During step 3, we can tackle #658 (already closed) as well, as both require touching the init of all instruments.

msmttchr:
I think this is a very good proposal.
I would consider simply removing `includeSCPI` (and associated code) rather than deprecating it.
I am not sure also how useful is `SCPIbaseUnknownMixin` since the default has always been `includeSCPI=True` and there was no recommendation to explicitly set it to `True`.

Finally, I am not clear how the Mixin class would be defined. Is it derived from `object` or other class?
Are we going to use control/measurement/setting properties?

If you have time, could you please sketch some implementation in the thread ?

bmoneke:
> I am not sure also how useful is SCPIbaseUnknownMixin since the default has always been includeSCPI=True and there was no recommendation to explicitly set it to True.

There are two reasons, why an instrument does not have `includeSCPI` defined: Either the implementor wanted to use SCPI and left the default, or the implementor did not know, that it exists and therefore could not set it to True.
As we do not know, which one is True, I want to have this UnknownMixin (only for our current codebase, not for future additions) in order to get feedback of users and to sort the codebase.

> Finally, I am not clear how the Mixin class would be defined. Is it derived from object or other class?

The Mixin just contains properties and methods (both suppose that an Instrument as "mainclass" exists), so it does not need a superclass:

~~~python
class SCPIbaseMixin:

    @property
    def id(self):
        """ Requests and returns the identification of the instrument. """
        return self.ask("*IDN?").strip()

    def clear(self):
        """ Clears the instrument status byte
        """
        self.write("*CLS")

[...]

PR #1007, opened 2023-12-20
Deprecate implicit includeSCPI declaration

### LINKED ISSUE #1000: Deprecate implicit includeSCPI definition for mixin based instruments
The transition from our current setup with `Instrument` including SCPI commands per default to a mixin based process (#804 , #905 ) is difficult as `includeSCPI` is True on default, but the future default will be without SCPI.

In order to facilitate the transition, I propose the following:

1. Change the default value to `None`
2. In the init, `None` will be converted to `True`, but will raise a deprecation warning.
3. We could add a test for new instruments that they have `includeSCPI` defined.

That change should prompt people to include a definition into their driver (those, which are in pymeasure and those which they use privately).

What do you think @bilderbuchi , @CasperSchippers , @msmttchr ?

BenediktBurger:
In preparation for #804, #905, this PR deprecates not declaring `includeSCPI` for enabling SCPI functions.
That allows later to switch to SCPI mixin class. Closes #1000 

Does not break working code: If it is  not declared, only a warning is raised and SCPI is enabled (as previously).

An "all_instruments" test checks for new instruments to declare `includeSCPI`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Upcoming version
 Deprecated features
 -------------------
 - Replaced :code:`directory_input` keyword-argument of :code:`ManagedWindowBase` by :code:`enable_file_input` (@CasperSchippers, #964)
-- Deprecated to not specify :code:`incldueSCPI` for an instrument which uses SCPI (@BenediktBurger, #1007)
+- Make parameter :code:`incldueSCPI` obligarory for all instruments, even those which use SCPI (@BenediktBurger, #1007)
 - Replaced :code:`celcius` attribute of :code:`LakeShoreTemperatureChannel` by :code:`celsius` (@afuetterer, #1003)
 
 GUI
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Upcoming version
 Deprecated features
 -------------------
 - Replaced :code:`directory_input` keyword-argument of :code:`ManagedWindowBase` by :code:`enable_file_input` (@CasperSchippers, #964)
-- Make parameter :code:`incldueSCPI` obligarory for all instruments, even those which use SCPI (@BenediktBurger, #1007)
+- Make parameter :code:`incldueSCPI` obligatory for all instruments, even those which use SCPI (@BenediktBurger, #1007)
 - Replaced :code:`celcius` attribute of :code:`LakeShoreTemperatureChannel` by :code:`celsius` (@afuetterer, #1003)
 
 GUI
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Upcoming version
 Deprecated features
 -------------------
 - Replaced :code:`directory_input` keyword-argument of :code:`ManagedWindowBase` by :code:`enable_file_input` (@CasperSchippers, #964)
-- Make parameter :code:`incldueSCPI` obligatory for all instruments, even those which use SCPI (@BenediktBurger, #1007)
+- Make parameter :code:`includeSCPI` obligatory for all instruments, even those which use SCPI (@BenediktBurger, #1007)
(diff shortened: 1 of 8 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,7 @@ Upcoming version
 Deprecated features
 -------------------
 - Replaced :code:`directory_input` keyword-argument of :code:`ManagedWindowBase` by :code:`enable_file_input` (@CasperSchippers, #964)
+- Deprecated to not specify :code:`incldueSCPI` for an instrument which uses SCPI (@BenediktBurger, #1007)
```
bilderbuchi, line 7:
```suggestion
- Deprecated to not specify :code:`includeSCPI` for an instrument which uses SCPI (@BenediktBurger, #1007)
```

tests/instruments/test_all_instruments.py
```diff
@@ -215,6 +215,91 @@ def find_devices_in_module(module):
     "ChannelAWG",
     "ChannelAFG",
 ]
+# Instruments which do not YET define `includeSCPI` explicitly
+grandfathered_includeSCPI_instruments = [
```
bilderbuchi, line 219: do you plan a follow-up PR that converts those instruments to the new usage? That would be good!

BenediktBurger: I cannot do that work alone, as I do not know, which instrument is Scpi compliant.

We have to rely on users submitting that information (they will get a warning).
Later (in the restructuring), we can add a "scpiUnknownMixin" (see my other pr) for these instruments.
That will keep the code base in a working state, but will also mark these instruments as maybe not Scpi compliant.

Tdlr: no large pr planned, but in small pieces over time.

bilderbuchi: I see (I think). So far, if not specified, we assume that SCPI inclusion is true, so assuming that the instruments in the codebase are implemented correctly, we could just carry that current behaviour over into explicit settings? Or do you want to take this opportunity to confirm for all those instruments if SCPI compatibility is _really_ appropriate?

BenediktBurger: I don't mind shipping both PRs together.

We should raise some warning, though, that Scpi compliance is not known in order to inform users.
That will trigger people to add the information, but won't stop code from working.

CHANGES.rst
BenediktBurger: I don't understand your message, therefore, I made a new suggestion, which is, hopefully, easier to understand.

bilderbuchi: No problem! However
```
:code:`incldueSCPI` obligarory for
```
-> 
```
:code:`includeSCPI` obligatory for
```

BenediktBurger: Done. Thanks.

tests/instruments/test_all_instruments.py
BenediktBurger: > We should raise some warning, though, that Scpi compliance is not known in order to inform users.
That will trigger people to add the information, but won't stop code from working.

The deprecation warning is that warning.

CHANGES.rst
```diff
@@ -4,7 +4,7 @@ Upcoming version
 Deprecated features
 -------------------
 - Replaced :code:`directory_input` keyword-argument of :code:`ManagedWindowBase` by :code:`enable_file_input` (@CasperSchippers, #964)
-- Make parameter :code:`incldueSCPI` obligarory for all instruments, even those which use SCPI (@BenediktBurger, #1007)
+- Make parameter :code:`incldueSCPI` obligatory for all instruments, even those which use SCPI (@BenediktBurger, #1007)
```
bilderbuchi, line 7:
```suggestion
- Make parameter :code:`includeSCPI` obligatory for all instruments, even those which use SCPI (@BenediktBurger, #1007)
```

BenediktBurger:
I did all the changes, anything left, @bilderbuchi?

bilderbuchi (approved):
LGTM, thanks!

PR #1010, opened 2023-12-21
Add redpitaya board

seb5g:
Hi there,
no need to review this yet. I just wanted to notify I was working on a driver for the redpitaya board using their SCPI server


pymeasure/instruments/redpitaya/redpitaya_scpi.py
```diff
@@ -0,0 +1,329 @@
...
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, False: 0},
+    )
+
+
+class DigitalChannelN(Channel):
+    """ A digital line of the N type"""
+
+    direction = Channel.control(
```
BenediktBurger, line 36: If you rename the property `direction_in`, and change it to accept booleans, it could be easier to handle than to give the proper string: `direction_in = True`, for example.

pymeasure/instruments/redpitaya/redpitaya_scpi.py
```diff
@@ -0,0 +1,329 @@
...
+        "DIG:PIN:DIR? DIO{ch}_N", "DIG:PIN:DIR %s,DIO{ch}_N",
+        """ Control a digital line to the given direction (str either 'IN' or 'OUT')""",
+        validator=truncated_discrete_set,
+        values=['IN', 'OUT'],
+    )
+
+    state = Channel.control(
```
BenediktBurger, line 43: I estimate, you mean the **enabled** state. Maybe `enabled` or `state_enabled` is better to understand.

pymeasure/instruments/redpitaya/redpitaya_scpi.py
```diff
@@ -0,0 +1,329 @@
...
+
+    def get_data_from_ascii(self) -> np.ndarray:
+        self.write(f"ACQ:SOUR{self.id}:DATA?")
+        data_str = self.read()
+        return np.fromstring(data_str.strip('{}').encode(), sep=',')
+
+    def get_data_from_binary(self) -> np.ndarray:
+        """ Get the data in """
+        self.write(f"ACQ:SOUR{self.id}:DATA?")
+        self.parent.adapter.read_bytes(1)
```
BenediktBurger, line 104: You can just call `read_bytes`, which calls the parent's `read_bytes`.
That way the parent does not have to be the instrument, which has an adapter, but could be another channel.
More importantly, your channel does not have to know how the parent is designed (with an adapter or not).

pymeasure/instruments/redpitaya/redpitaya_scpi.py
```diff
@@ -0,0 +1,329 @@
...
+    )
+
+    def get_data_from_ascii(self) -> np.ndarray:
+        self.write(f"ACQ:SOUR{self.id}:DATA?")
+        data_str = self.read()
+        return np.fromstring(data_str.strip('{}').encode(), sep=',')
+
+    def get_data_from_binary(self) -> np.ndarray:
+        """ Get the data in """
+        self.write(f"ACQ:SOUR{self.id}:DATA?")
```
BenediktBurger, line 103: You do not have to use a format string, as we already have an id replacement (using the `placeholder` attribute, which defaults to `"ch"`) in the `write` method (designed for our property creators):
```suggestion
        self.write("ACQ:SOUR{ch}:DATA?")
```

pymeasure/instruments/redpitaya/redpitaya_scpi.py
```diff
@@ -0,0 +1,329 @@
...
+    LV_MAX = 1
+    HV_MAX = 20
+    CLOCK = 125e6  # Hz
+
+    def __init__(self, ip_address: str, port: int = 5000, name="Redpitaya SCPI",
+                 read_termination='\r\n',
+                 write_termination='\r\n',
+                 **kwargs):
+
+        adapter = f"TCPIP::{ip_address}::{port}::SOCKET"
```
BenediktBurger, line 139: That makes it more difficult to use other adapters (home-made or the Protocol Adapter for tests).

Best way is to give an example how to create the resource name.
You could make the `adapter` keyword optional (defaulting to None) and keep the ip_address and port.
```suggestion
    def __init__(self,
                 adapter=None,
                 ip_address: str, port: int = 5000, name="Redpitaya SCPI",
                 read_termination='\r\n',
                 write_termination='\r\n',
                 **kwargs):

        if adapter is None:
            adapter = f"TCPIP::{ip_address}::{port}::SOCKET"
```

BenediktBurger (commented):
I know, that you did not call for comments, looking at your code, I thought some general comments might be useful.

pymeasure/instruments/redpitaya/redpitaya_scpi.py
seb5g: done

pymeasure/instruments/redpitaya/redpitaya_scpi.py
seb5g: done

pymeasure/instruments/redpitaya/redpitaya_scpi.py
seb5g: done

pymeasure/instruments/redpitaya/redpitaya_scpi.py
seb5g: done

pymeasure/instruments/redpitaya/redpitaya_scpi.py
seb5g: done as suggested

seb5g:
Hi guys, the PR is ready for review. I didn't code the analog ouput signal generation yet as I'm not planning to use it soon, but it should be fairly easy to do so based on this initial work. Thanks for reviewing it

pymeasure/instruments/redpitaya/__init__.py
BenediktBurger, line 1: Please add also the documentation entries (those small files in the docs folder), see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation

pymeasure/instruments/redpitaya/redpitaya_scpi.py
```diff
@@ -0,0 +1,338 @@
...
+    The instrument is accessed using a TCP/IP Socket communication, that is an adapter in the form:
+    "TCPIP::x.y.z.k::port::SOCKET" where x.y.z.k is the IP address of the SCPI server
+    (that should be activated on the board) and port is the TCP/IP port number, usually 5000
+
+    To activate the SCPI server, you have to connect first the redpitaya to your computer/network
+    and enter the url address written on the network plug (on the redpitaya). It should be something
+    like "RP-F06432.LOCAL/" then browse the menu, open the Development application and activate the
+    SCPI server. When activating the server, you'll be notified with the IP/port address to use
+    with this Instrument.
+    """
```
BenediktBurger, line 155: It is good to mention the creation of the address in the instrument's docstring:
```suggestion

    :param ip_address: IP address to use, if `adapter` is None.
    :param port: Port number to use, if `adapter` is None.
    """
```

pymeasure/instruments/redpitaya/redpitaya_scpi.py
```diff
@@ -0,0 +1,338 @@
...
+    board_name = Instrument.measurement("SYST:BRD:Name?",
+                                        """Get the RedPitaya board name""")
+
+    def digital_reset(self):
+        """Reset the state of all digital lines"""
+        self.write("DIG:RST")
+
+    dioN = Instrument.MultiChannelCreator(DigitalChannelN, list(range(7)), prefix='dioN')
+    dioP = Instrument.MultiChannelCreator(DigitalChannelP, list(range(7)), prefix='dioP')
+    led = Instrument.MultiChannelCreator(DigitalChannelLed, list(range(8)), prefix='led')
```
BenediktBurger, line 209: It is better to put the channels close to the init, in order to see the instrument structure more easily.

tests/instruments/redpitaya/generate_redpitaya.py
BenediktBurger, line 1: We have not included these generating files into other instruments. I'm unsure, whether we should do it or not.

In the end, it is similar to a `test_with_device`.

Would you mind, transforming it into a [device test](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#device-tests)?

That could be quite simple.
In the mos simple form, make a single test out of all the code.

Maybe it is even possible to do the generator part, whenever you call the script directly?

BenediktBurger (commented):
Thanks for your contribution. It looks good.

You did not yet add the device documentation.
I left a few comments.

tests/instruments/redpitaya/test_redpitaya.py
BenediktBurger, line 1: The test with a device is called `test_DEVICE_with_device`, while the protocol test ist just called `test_DEVICE`.
Therefore, this file should contain the protocol tests.

seb5g:
yes sure!!

PR #1013, opened 2024-01-02
Feature proposal: Possible to not save all parameters

poje42:
I'm new to this with GitHub, PR etc, so if I do it wrong, please advice me!
I've been using pymeasure several years with several local changes, so I will try to see if I can share some changes which I hope can be of benefit for others. 
First out is a feature that has been on my mind and when I saw the implementation of Metadata, I think it will make even more sense.

The idea is to have input parameters that are not saved as Parameters in the result files. I see at least two uses for this:
1. Some of my GUIs I've made have too many parameters to fit nicely on the screen, so I've used the 'group_by'-parameter. The parameter used to show the different groups needs not be saved.
2. An ID of the DUT that the user enters and should be saved as metadata, has I don't want the parameter to be overwritten if I load parameters from a file.

I've made a proposal in this PR to add a Boolean 'save' parameter to the Parameter class and an example. Maybe this is not the right way to go, I welcome better ideas!

### DIFF
--- a/examples/Basic/gui_no_save_param.py
+++ b/examples/Basic/gui_no_save_param.py
@@ -36,8 +36,10 @@ python gui.py
 import sys
 import random
 from time import sleep
+from PyQt5.QtCore import QLocale
+from pyqtgraph.Qt import QtCore
 
-from pymeasure.experiment import Procedure, IntegerParameter, Parameter, FloatParameter
+from pymeasure.experiment import Procedure, IntegerParameter, Parameter, FloatParameter, ListParameter, unique_filename, Metadata, Results
 from pymeasure.display.Qt import QtWidgets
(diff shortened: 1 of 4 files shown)


BenediktBurger:
Hey @poje42, welcome on our Github page.
We appreciate your wish to share your developments with the community. 

In general, it is good to have an "issue" to discuss a feature proposal and start implementing it, once it is considered good.

As your code is already written, you can can open a PR and link it in your issue.
If it is a PR with little possible discussion (big fix, typos...), no additional issue is necessary.

In your case, an additional issue might be good, as the preliminary discussion is better done in an issue than a pr (which is awful for discussing).

@msmttchr and @CasperSchippers, what do you think about this proposal?

poje42:
Thanks for the respons, made issue #1014.

Issue #1015, opened 2024-01-02
Bug: Exception if loading result file with an empty parameter


poje42:
Steps to Reproduce:
* Run the gui.py example
* Clear the "Random Seed" field
* Run the test
* Remove graph
* Load the generated file
* Exception in results.py

An empty string parameter will be saved as `#<Tab>Random Seed:<Space><LF>`
Results.load() will do a strip() that will remove the trailing space so parse_header() will fail to find `":<Space>"`

CasperSchippers:
Good catch!

This warrants the following consideration (relating also to what is mentioned in the PR about changing the separator): in principle all valid strings can be used as parameter name; while convenient, this does result in issues if you include the separator (i.e., `:<space>`) into the parameter name (or the value) (you can no longer load the value from file).
So whatever we decide regarding the separator, i'd be in favour of not allowing (and checking/enforcing) the parameter name to contain the separator.

BenediktBurger:
> So whatever we decide regarding the separator, i'd be in favour of not allowing (and checking/enforcing) the parameter name to contain the separator.

I agree.

poje42:
So what about this?
* Keep the current separator `:<space>` -> It will not break existing scripts
* Add checking of parameter names ->  Better robustness
* Remove strip in load() -> Parameters can have any content (almost, some characters like LF will not work)

BenediktBurger:
Sounds good to me.

CasperSchippers:
Sounds good indeed.

Seems like you found the strip that causes the issue (sorry didn't have time at the moment to try and find it). One thing that comes to mind is considering whether we should strip specifically for "\r\n" (instead of the standard strip which removes, from top of my head, "<space>\r\n") to prevent any windows vs Unix issues (not sure if such an issue would arise, but just wanted to mention it while I thought of it).

poje42:
The strip is done in load()
```
        with open(data_filename) as f:
            while not header_read:
                line = f.readline()
                if line.startswith(Results.COMMENT):
                    header += line.strip() + Results.LINE_BREAK
                    header_count += 1
                else:
                    header_read = True

```
Good comment @CasperSchippers , the strip must remain to remove other whitespace characters.
The new commit uses:
`                    header += line.strip('\t\v\n\r\f') + Results.LINE_BREAK
`

PR #1016, opened 2024-01-02
Fix exception if loading result file with an empty parameter

poje42:
Fix proposal for #1015

### DIFF
--- a/pymeasure/experiment/parameters.py
+++ b/pymeasure/experiment/parameters.py
@@ -46,6 +46,11 @@ class Parameter:
 
     def __init__(self, name, default=None, ui_class=None, group_by=None, group_condition=True):
         self.name = name
+        separator = ": "
+        if separator in name:
+            raise ValueError("The provided name argument '%s' contains the sparator '%s')."
+                             % (name, separator))
+
         self._value = None
         if default is not None:
             self.value = default
--- a/pymeasure/experiment/parameters.py
+++ b/pymeasure/experiment/parameters.py
@@ -49,7 +49,7 @@ class Parameter:
         separator = ": "
         if separator in name:
             raise ValueError(f"The provided name argument '{name}' contains the "
-                             "separator '{separator}'.")
+                             f"separator '{separator}'.")
 
         self._value = None
         if default is not None:
--- a/pymeasure/experiment/parameters.py
+++ b/pymeasure/experiment/parameters.py
@@ -48,7 +48,7 @@ class Parameter:
         self.name = name
         separator = ": "
         if separator in name:
-            raise ValueError("The provided name argument '%s' contains the sparator '%s')."
+            raise ValueError("The provided name argument '%s' contains the sparator '%s'."
                              % (name, separator))
 
         self._value = None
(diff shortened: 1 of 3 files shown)


BenediktBurger (commented):
Thanks for your suggestion.

You may add yourself to the `authors.txt`, as this is your first PR.

I propose a different solution:
make the separator only the colon (`":"`) and then do an `lstrip()` on the partitioned_line (or a strip?).
That would make the added space superfluous.

poje42:
@BenediktBurger I was considering that solution also, but that may break old scripts that have parameter names such as "Para:meter" or "Parameter:". This is maybe only a theoretical problem and I don't know how strict you see on backward compatibility?

BenediktBurger:
We try to keep backward compatibility.

An alternative change just the exception handling:
~~~python
if part... != separator:
    if part[0].endswith(separator.split()):
        parameters[partitioned_line[0].split(separator]
    else:
        raise
~~~

CasperSchippers:
I'm wondering if stripping is the best way to go, as you'd also remove any intential whitespaces in the value. We know exactly what we add when writing to file, don't we? So if we can prevent the stripping (at any point) of whitespaces, you'd be able to get exactly the input value back.

pymeasure/experiment/parameters.py
```diff
@@ -46,6 +46,11 @@ class Parameter:
 
     def __init__(self, name, default=None, ui_class=None, group_by=None, group_condition=True):
         self.name = name
+        separator = ": "
+        if separator in name:
+            raise ValueError("The provided name argument '%s' contains the sparator '%s')."
+                             % (name, separator))
```
BenediktBurger: You could use the more modern f-string:
```suggestion
            raise ValueError(f"The provided name argument '{name}' contains the separator '{separator}'.")
```

Btw. you spelled it wrong: "sparator" (without "e"). And you have an additional parenthesis in the string (second to last character).

poje42: Careless of me... Also, I like to use f-strings. Thanks for your comments.

pymeasure/experiment/parameters.py
```diff
@@ -46,6 +46,11 @@ class Parameter:
 
     def __init__(self, name, default=None, ui_class=None, group_by=None, group_condition=True):
         self.name = name
+        separator = ": "
+        if separator in name:
+            raise ValueError(f"The provided name argument '{name}' contains the "
+                             "separator '{separator}'.")
```
CasperSchippers, line 52: Because the string is broken in two, you need an additional `f` in front of the second part of the string:

```suggestion
                             f"separator '{separator}'.")
```

CasperSchippers (approved):
Thanks for the bugfix!! Apart from a minor remark on the exception message, it looks good to me!

One thing that this made me realize is that the separator is hard-coded.... that's fairly ugly. But that has nothing to do with your work, and is something to fix in another PR I think (will likely open one after this has been merged).

poje42:
@BenediktBurger As I said before, I'm new to this so is there something more I should do? I think this fix is ready.

BenediktBurger:
Thanks for your work.

For merging, a pr has to be marked "ready for review" (I did it for you).
In fact, un most cases you want to use a "normal" pr. The draft PRs are for cases where you are not yet done with your code, for example, if you write an instrument driver but have not yet written all properties.

PR #1017, opened 2024-01-03
Adjust hashes.

tomverbeure:


BenediktBurger:
Should be together with that other pr.

PR #1018, opened 2024-01-04
Fix property docstrings

BenediktBurger:
This PR fixes the docstrings of the properties of some instruments in the codebase.

Each commit is one instrument (or sometimes a few instruments of one manufacturer).


msmttchr (approved):
Lgtm. Thanks

PR #1026, opened 2024-01-13
Support for numpy 2.0

CasperSchippers:
Numpy 2.0 will soon (few months) be released with some (minor) breaking changes.

This PR fixes the things that break between numpy 1.x and 2.x (while keeping it compatible with both, assuming this is possible).

Also, I've tried standardizing the numpy imports a bit, such that all numpy imports in the repo read `import numpy as np`.

### DIFF
--- a/.github/pymeasure_np2.yml
+++ b/.github/pymeasure_np2.yml
@@ -3,8 +3,8 @@ channels:
   - conda-forge
 dependencies:
   - cloudpickle=1.6.0
-  - pandas>=2.2.2
-#  - pint=0.18 a pint version that has not yet been released (>0.23) is required to work with numpy2
+#  - pandas>=2.2.2  # The version on conda-forge seems not be built with numpy==2, but the pip version has been
+#  - pint=0.18  # A pint version that has not yet been released (>0.23) is required to work with numpy2
   - pyqt=5.15.7
   - pyqtgraph=0.12.4
   - pyserial=3.4
@@ -25,4 +25,5 @@ dependencies:
   - pip  # don't pin, to gain newest conda compatibility fixes
   - pip:
       - numpy==2.0.0rc1
+      - pandas==2.2.2
       - -e git+https://github.com/hgrecco/pint.git#egg=pint
--- a/.github/pymeasure_np2.yml
+++ b/.github/pymeasure_np2.yml
@@ -1,0 +1,28 @@
+name: pymeasure_np2
+channels:
+  - conda-forge
+dependencies:
+  - cloudpickle=1.6.0
+  - pandas>=2.2.2
+#  - pint=0.18 a pint version that has not yet been released (>0.23) is required to work with numpy2
+  - pyqt=5.15.7
+  - pyqtgraph=0.12.4
+  - pyserial=3.4
+  - pyvisa=1.12.0
+  - pyzmq=24.0.1
+  - qt=5.15.6
+# Development dependencies below
+  - pytest-qt=4.2.0
+  - pytest-runner=5.2
+  - pytest=7.2.0
+  - pytest-cov=4.1.0
+  - pyvisa-sim==0.5.1
+  - flake8=6.0.0
+  - setuptools_scm # don't pin, to get newest features
+  - sphinx=5.3.0
+  - sphinx_rtd_theme=1.2.2
+#  pip is currently not needed, but the recommended tool when packages that are unavailable on conda are required
+  - pip  # don't pin, to gain newest conda compatibility fixes
+  - pip:
+      - numpy==2.0.0rc1
+      - -e git+https://github.com/hgrecco/pint.git#egg=pint
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Upcoming version
 ================
 Main items of this new release:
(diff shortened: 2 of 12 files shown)


pymeasure/instruments/fwbell/fwbell5080.py
```diff
@@ -24,9 +24,10 @@
 
 from pymeasure.instruments import Instrument
 from pymeasure.instruments.validators import strict_discrete_set
-from numpy import array, float64
 from time import sleep
```
BenediktBurger, line 27: As you are restructuring the imports, you could move the time import to the top (standard library).

BenediktBurger (commented):
Thanks for your work. I like it to have a consistend import structure of numpy.

One small rework, while you are at it.

BenediktBurger:
Regarding the code coverage check failing, don't worry, as that is a known bug in #1090

BenediktBurger:
> * When support for python 3.8 is removed, we can also increase the minimum pandas version (as pandas 1.0.0 is the first version to support python 3.8)

As that is a nice version number, we can make pandas 1.0.0 the minimum requirement (contrary what I wrote in the issue).

CasperSchippers:
I think that when this is aproved and merged we should consider not waiting too long with a new release.

.github/pymeasure_np2.yml
```diff
@@ -0,0 +1,29 @@
...
+  - flake8=6.0.0
+  - setuptools_scm # don't pin, to get newest features
+  - sphinx=5.3.0
+  - sphinx_rtd_theme=1.2.2
+#  pip is currently not needed, but the recommended tool when packages that are unavailable on conda are required
+  - pip  # don't pin, to gain newest conda compatibility fixes
+  - pip:
+      - numpy==2.0.0rc1
+      - pandas==2.2.2
+      - -e git+https://github.com/hgrecco/pint.git#egg=pint
```
BenediktBurger, line 29: Is the "-e" for an editable install? Is it needed?

.github/pymeasure_np2.yml
```diff
@@ -0,0 +1,29 @@
+name: pymeasure_np2
```
BenediktBurger, line 1: maybe
```suggestion
name: pymeasure_numpy2
```
?

CHANGES.rst
```diff
@@ -1,5 +1,7 @@
 Upcoming version
 ================
+Main items of this new release:
+- Added support for numpy 2.0; limited the allowed numpy version to <3.
```
BenediktBurger, line 4: Should we mention the upper limit of numpy?

BenediktBurger (approved):
A few comments.

Thanks for tackling this issue!

docs/quick_start.rst
```diff
@@ -61,3 +62,9 @@ Execute the following Python code.
     pymeasure.__version__
 
 You should see the version of PyMeasure printed out. At this point you have PyMeasure installed, and you are ready to start using it! Are you ready to :doc:`connect to an instrument <./tutorial/connecting>`?
+
+.. warning::
+    The release of numpy version 2 caused some compatibility issues.
+    PyMeasure is compatible with version 2 as of PyMeasure version 0.14.
+    To use PyMeasure with numpy>=2, you also ensure the dependencies are compatible.
```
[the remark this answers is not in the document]
bilderbuchi: Why does the user _have to_ do this? Should not our package metadata make sure of this automatically?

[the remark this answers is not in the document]
bilderbuchi: I'm not sure if we should even mention this whole warning as long as our package specification is correct? 
This whole paragraph will become outdated really quickly, so we need to remember to remove it again. Finding pandas and pint versions compatible with both pymeasure _and_ numpy 2 is the responsibility of the package manager (pip/conda). If that fails, it's ultimately the issue of a dependency, not ours, right?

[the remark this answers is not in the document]
BenediktBurger: You're right, we should not add this warning.

If numpy 2.0 is already installed, it will fail until the dependencies are met. Otherwise, it won't install numpy 2.0, so it won't be an issue either.

BenediktBurger:
@CasperSchippers , would you mind merging in master in order to check, whether the codecov will work? Thanks!

.github/pymeasure_np2.yml
CasperSchippers: Oh, I thought it was based on some testing and the documentation, but it might not be strictly necesary

docs/quick_start.rst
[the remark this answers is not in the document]
CasperSchippers: Sure, I'll remove the warning.

Note that we need to update the requirements.yml file as well once the final version of both numpy=2.0.0 and pint>0.23 are released.

CHANGES.rst
CasperSchippers: Maybe not, I removed the remarks.

CasperSchippers:
> @CasperSchippers , would you mind merging in master in order to check, whether the codecov will work? Thanks!

Good job, it seems to be working now!

My test doesn't anymore though 😞

BenediktBurger:
It looks good to me. Thanks for your work!

Issue #1027, opened 2024-01-15
Rework Keithleys' next_error


BenediktBurger:
Actually, looking at the code, it seems to me that removes character `“` from string associated to error code.
It looks like something related to personal taste, on the other hand the ‘retry’ when reply length < 2, it seems important, if really needed.

It would be nice to remove completely this custom next_error to simplify code, but perhaps it is not possible. In any case, I am good to merge.

_Originally posted by @msmttchr in https://github.com/pymeasure/pymeasure/pull/1024#discussion_r1450822843_
            

It is about the Keithleys' `next_error` property, which removes double quotes and does re-read whenever there is only one part in the response message.
~~~python
    @property
    def next_error(self):
        """ Returns a tuple of an error code and message from a
        single error. """
        err = super().next_error
        if len(err) < 2:
            err = self.read()  # Try reading again
        code = err[0]
        message = err[1].replace('"', '')
        return (code, message)
~~~

- Do the Keithleys require this re-read?
- Is it good to remove the double quotes?

msmttchr:
In my opinion, it’s difficult to find a reasonable motivation to justify the double quote removaI.
In addition, also custom `check_errors` could be probably removed.
Really, we need some volunteers with same instrument.

BenediktBurger:
> Really, we need some volunteers with same instrument.

For that reason I opened this issue.

msmttchr:
Found Keithkey 2420 in lab and did some testing with GPIB interface:
- Current style of error logging (customized): `INFO:pymeasure.instruments.keithley.keithley2400:Keithley 2400 reported error: -113, Undefined header`
- Standard style for error logging: `ERROR:pymeasure.instruments.instrument:Keithley 2400 SourceMeter: -113.0, "Undefined header"`

Generated in a loop 10000 errors and checked that empty error message or incorrect error are never detected.

I will try to test with serial interface and then I will submit a PR for removing the exception.

BenediktBurger:
Thanks for looking into it.

PR #1030, opened 2024-01-22
Rework Keithleys' next_error

msmttchr:
Address #1027

I think some initial code was developed a long time ago and then copied in all subsequent drivers.

This PR removes exception for `check_errors` method, aligning error logging to the main method.
In addition to this, it removes also customisation of `next_error`, since this is standard SCPI.
The changes have been tested in Keithley2400 class using Keithley 2420 connected via GPIB.

### DIFF
--- a/pymeasure/instruments/keithley/keithley2450.py
+++ b/pymeasure/instruments/keithley/keithley2450.py
@@ -498,31 +498,9 @@ class Keithley2450(KeithleyBuffer, Instrument):
         self.beep(base_frequency * 6.0 / 4.0, duration)
 
     @property
-    def next_error(self):
-        """ Returns a tuple of an error code and message from a
-        single error. """
-        err = super().next_error
-        if len(err) < 2:
-            err = self.read()  # Try reading again
-        code = err[0]
-        message = err[1].replace('"', '')
-        return (code, message)
-
-    @property
     def error(self):
         warn("Deprecated to use `error`, use `next_error` instead.", FutureWarning)
         return self.next_error
-
-    def check_errors(self):
-        """ Logs any system errors reported by the instrument.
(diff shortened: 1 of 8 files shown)


tests/instruments/test_all_instruments.py
```diff
@@ -156,6 +156,7 @@ def find_devices_in_module(module):
     "Keithley2400",
     "Keithley2450",
     "Keithley2600",
+    "Keithley2260B",
```
BenediktBurger, line 159: How did the tests succeed previously without that line?

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -23,7 +23,6 @@
 #
 
 import logging
-import time
```
BenediktBurger, line 26: time is required for other methods as well.

BenediktBurger (approved):
Looks good (besides already linted things).

tests/instruments/test_all_instruments.py
msmttchr: I think because I added the deprecation on the `error` properties, following what has been done in the other files.

BenediktBurger: Adding an instrument to that list, removes the checks for all properties.

The idea is, to eventually remove all instruments from the list.

Therefore, it is better to fix the code (docstring) to pass the test.

msmttchr: Also other files does not have documentation in `error`. I was confused about this. So, this file was the only one with check enabled while the other had the check disabled.

msmttchr: Changes done

Issue #1029, opened 2024-01-19
UDP not supported?


samcondon4:
I have an instrument that communicates over ethernet via the UDP protocol rather than TCP/IP. As far as I can tell, UDP is not supported by pyvisa and hence neither by the `VISAAdapter`. I am considering writing a new adapter class to address this but want to make sure that this wouldn't duplicate any already existing functionality to communicate with UDP instruments.

Can anyone more familiar with pyvisa and the pymeasure adapter ecosystem confirm that the UDP protocol is not currently supported? Thank you!

msmttchr:
I think you're correct in assuming that udp is not supported. You could also try to ask in [https://github.com/pyvisa/pyvisa](pyvisa)

If you need to implement your own adapter, you should create a subclass of `Adapter` and also have a look at existing files under the adapter directory.

PR #1031, opened 2024-01-23
CI-Systems SR200N Blackbody

samcondon4:
New instrument driver added. The communication protocol for this one is a bit more involved with a checksum and bytes communication- so the base `Instrument.write` and `Instrument.read` methods have been overridden. 

This instrument also has an ethernet interface supporting the UDP protocol. Related to issue #1029 I am considering opening another PR with a new UDP adapter class. Will update this driver to support the new adapter at a later date.

### DIFF
--- a/pymeasure/instruments/cisystems/sr200n.py
+++ b/pymeasure/instruments/cisystems/sr200n.py
@@ -1,0 +1,181 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2023 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import struct
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.common_base import CommonBase
+
+
+# - PARAMETER STRINGS - #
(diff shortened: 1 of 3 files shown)


pymeasure/instruments/cisystems/sr200n.py
```diff
@@ -0,0 +1,181 @@
...
+SET_POINT_SET = '0617be %f'
+CHOPPER_FREQ_GET = '081939'
+CHOPPER_FREQ_SET = '06193a %f'
+CHOPPER_POS = '06193b %i'
+
+
+class SR200N(Instrument):
+
+    # - temperature parameters ------------------------------ #
+    temp = CommonBase.measurement(
```
BenediktBurger, line 45: Please use the Instrument's measurement etc.
```suggestion
    temp = Instrument.measurement(
```

pymeasure/instruments/cisystems/sr200n.py
```diff
@@ -0,0 +1,181 @@
...
+        TEMP,
+        """Float representing the current temperature of the blackbody source.""",
```
BenediktBurger, line 47: We made a code style standard regarding docstrings: https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings

Your current docstring does not comply with that style (which is similar to older style in the codebase). Therefore, the tests fail with your docstrings.

Here's the text regarding property docstrings:
> Specific to properties, start them with “Control”, “Get”, “Measure”, or “Set” to indicate the kind of property, as it is not visible after import, whether a property is gettable (“Get” or “Measure”), settable (“Set”), or both (“Control”). In addition, it is useful to add type and information about [Restricting values with validators](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#validators) (if applicable) at the end of the summary line, see the docstrings shown in examples throughout the [Adding instruments](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/index.html#adding-instruments) section. For example a docstring could be """Control the voltage in Volts (float strictly from -1 to 1).""".

pymeasure/instruments/cisystems/sr200n.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def __init__(self, adapter, name='SR200N Blackbody', **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            asrl={
+                'baud_rate': 115200,
+            },
+            **kwargs
+        )
```
BenediktBurger, line 103: Please move the init at the top of the class definition.
And please add `includeSCPI=False`.

pymeasure/instruments/cisystems/sr200n.py
```diff
@@ -0,0 +1,181 @@
...
+        
+        return cs.to_bytes(length=4, byteorder='big')
+
+    def write(self, command, **kwargs):
+        """ Override the base write method to construct the checksum and use `write_bytes()`.
+
+        :param command: hexadecimal command as a string.
+        """
+        # - split the command string on spaces, get the command value if present - #
+        cmd_split = command.split(' ')
```
BenediktBurger, line 137: In the hcp TC038D I used a comma as separator, which is more common, I'd think.

Here's the code, maybe it serves you as inspiration:

https://github.com/pymeasure/pymeasure/blob/f3d8f84c3e84b9de368f392a43accaac6423a602/pymeasure/instruments/hcp/tc038d.py#L76-L105

BenediktBurger (commented):
Thanks for your contribution. I made a few remarks.

Please add also the API documentation (see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation), and please add [protocol tests](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#protocol-tests), which are especially useful for this kind of more complex instruments.

PR #1034, opened 2024-01-26
add MKS 974B vacuum pressure transducer

dkriegner:
I suggest to add a second device from MKS. Since the communication protocol it uses is equal to the one of the other included MKS device I created an `MKSInstrument` class which is reused by both device implementations.

The manual for this device can be found at 
https://www.idealvac.com/files/manuals/MKS-974-MAN-RH.pdf

### DIFF
--- a/docs/api/instruments/mksinst/mksinst.rst
+++ b/docs/api/instruments/mksinst/mksinst.rst
@@ -1,0 +1,7 @@
+###################################
+MKS Instruments abstract Instrument
+###################################
+
+.. autoclass:: pymeasure.instruments.mksinst.mksinst.MKSInstrument
+    :members:
+    :show-inheritance:
--- a/docs/api/instruments/mksinst/mksinst.rst
+++ b/docs/api/instruments/mksinst/mksinst.rst
@@ -1,5 +1,5 @@
 ###################################
-MKS Instruments abstract Instrument
+MKS Instruments Abstract Instrument
 ###################################
 
 .. autoclass:: pymeasure.instruments.mksinst.mksinst.MKSInstrument
--- a/docs/api/instruments/mksinst/mksinst.rst
+++ b/docs/api/instruments/mksinst/mksinst.rst
@@ -5,3 +5,7 @@ MKS Instruments Abstract Instrument
 .. autoclass:: pymeasure.instruments.mksinst.mksinst.MKSInstrument
     :members:
     :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.mksinst.mksinst.RelayChannel
+    :members:
+    :show-inheritance:
--- a/pymeasure/instruments/mksinst/mks937b.py
+++ b/pymeasure/instruments/mksinst/mks937b.py
@@ -26,7 +26,7 @@ from enum import StrEnum
 from pymeasure.instruments import Channel, Instrument
 from pymeasure.instruments.validators import strict_discrete_set
 
-from .mksinst import MKSInstrument
+from .mksinst import MKSInstrument, RelayChannel
 
 
 _ion_gauge_status = {"Wait": "W",
@@ -48,6 +48,24 @@ class Unit(StrEnum):
     mbar = "mBAR"
     Pa = "PASCAL"
     uHg = "MICRON"
+
+
+class Relay(RelayChannel):
+
+    enabled = Channel.control(
+        "EN{ch}?", "EN{ch}!%s",
+        """Control the relay function or disable the setpoint relay.
+
+        Possible values are True/False to enable/disable pressure based
+        activation of the relay and 'SET' to permanently energize the relay.""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={False: "CLEAR",
+                True: "ENABLE",
+                "SET": "SET",
+                },
+        check_set_errors=True,
+    )
 
 
 class PressureChannel(Channel):
@@ -80,13 +98,13 @@ class MKS937B(MKSInstrument):
     """ MKS 937B vacuum gauge controller
 
     Connection to the device is made through an RS232/RS485 serial connection.
-    The communication protocol of this device is as follows:
 
-    Query: '@<aaa><Command>?;FF' with the response '@<aaa>ACK<Response>;FF'
-    Set command: '@<aaa><Command>!<parameter>;FF' with the response '@<aaa>ACK<Response>;FF'
-    Above <aaa> is an address from 001 to 254 which can be specified upon
-    initialization. Since ';FF' is not supported by pyvisa as terminator this
-    class overloads the device communication methods.
+    The 937B gauge controller can connect up to 6 pressure measurement channels
+    for gauges of various types which includes ionization gauges, Pirani and
+    piezo gauges. Based on the pressure values of the gauges twelve setpoint
+    relays can be energized when certain pressure values are reached. The
+    assignment of the relays to measurement channels is fixed to have two
+    relays for each pressure channel.
 
     :param adapter: pyvisa resource name of the instrument or adapter instance
     :param string name: The name of the instrument.
@@ -107,6 +125,30 @@ class MKS937B(MKSInstrument):
     ch_5 = Instrument.ChannelCreator(IonGaugeAndPressureChannel, 5)
 
     ch_6 = Instrument.ChannelCreator(PressureChannel, 6)
+
+    relay_1 = Instrument.ChannelCreator(Relay, 1)
+
+    relay_2 = Instrument.ChannelCreator(Relay, 2)
+
+    relay_3 = Instrument.ChannelCreator(Relay, 3)
+
+    relay_4 = Instrument.ChannelCreator(Relay, 4)
+
+    relay_5 = Instrument.ChannelCreator(Relay, 5)
+
+    relay_6 = Instrument.ChannelCreator(Relay, 6)
+
+    relay_7 = Instrument.ChannelCreator(Relay, 7)
+
+    relay_8 = Instrument.ChannelCreator(Relay, 8)
+
+    relay_9 = Instrument.ChannelCreator(Relay, 9)
+
+    relay_10 = Instrument.ChannelCreator(Relay, 10)
+
+    relay_11 = Instrument.ChannelCreator(Relay, 11)
+
+    relay_12 = Instrument.ChannelCreator(Relay, 12)
 
     def __init__(self, adapter, name="MKS 937B vacuum gauge controller", address=253, **kwargs):
         super().__init__(
--- a/pymeasure/instruments/mksinst/mks937b.py
+++ b/pymeasure/instruments/mksinst/mks937b.py
@@ -192,6 +192,11 @@ class MKS937B(MKSInstrument):
         Unit.uHg/'Micron'.""",
         validator=strict_discrete_set,
         map_values=True,
-        values={u: u.value for u in Unit},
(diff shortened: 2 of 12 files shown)


docs/api/instruments/mksinst/mksinst.rst
```diff
@@ -0,0 +1,7 @@
+###################################
+MKS Instruments abstract Instrument
```
BenediktBurger, line 2: For consistency:
```suggestion
MKS Instruments Abstract Instrument
```

pymeasure/instruments/mksinst/mksinst.py
```diff
@@ -0,0 +1,116 @@
...
+    initialization. Since ';FF' is not supported by pyvisa as terminator this
+    class overloads the device communication methods.
+
+    :param adapter: pyvisa resource name of the instrument or adapter instance
+    :param string name: The name of the instrument.
+    :param address: device address included in every message to the instrument
+                    (default=253)
+    :param kwargs: Any valid key-word argument for Instrument
+    """
+    def __init__(self, adapter, name="MKS Instrument", address=253, **kwargs):
```
BenediktBurger, line 48:
```suggestion

    def __init__(self, adapter, name="MKS Instrument", address=253, **kwargs):
```

pymeasure/instruments/mksinst/mks974b.py
```diff
@@ -0,0 +1,224 @@
...
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Channel, Instrument
+from pymeasure.instruments.validators import strict_discrete_set
+
+from .mksinst import MKSInstrument
+
+
+class SetpointChannel(Channel):
```
BenediktBurger, line 31: I think `Relay` or `RelayChannel` would be better.

pymeasure/instruments/mksinst/mks974b.py
```diff
@@ -0,0 +1,224 @@
...
+    class overloads the device communication methods.
+
+    :param adapter: pyvisa resource name of the instrument or adapter instance
+    :param string name: The name of the instrument.
+    :param address: device address included in every message to the instrument
+                    (default=253)
+    :param kwargs: Any valid key-word argument for Instrument
+    """
+
+    setpoint_1 = Instrument.ChannelCreator(SetpointChannel, 1)
```
BenediktBurger, line 105: Here again, I think you should call it `relay_1`...

pymeasure/instruments/mksinst/mks974b.py
```diff
@@ -0,0 +1,224 @@
...
+        "U?", "U!%s",
+        """Control pressure unit used for all pressure readings from the instrument""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={"Torr": "TORR",
+                "mBar": "MBAR",
+                "Pascal": "PASCAL",
+                },
+        check_set_errors=True,
+    )
```
BenediktBurger, line 206: You could use an enum for the values in order to make it easier to set the values.
At least you should mention the possible units in the docstring.
you could change the dictionary keys (and therefore possible user values) to the units abbreviation: Torr, mbar, Pa

pymeasure/instruments/mksinst/mks974b.py
```diff
@@ -0,0 +1,224 @@
...
+        cast=str,
+    )
+
+    model = Instrument.measurement(
+        "MD?", """Get the transducer model number""",
+        cast=str,
+    )
+
+    part_number = Instrument.measurement(
+        "MD?", """Get the transducer part number""",
```
BenediktBurger, line 154: Are these commands common to all instruments?
At least the serial number command is common.

If it is just the serial number, I like the current separation between communication logic in the parent and everything else in the child.

If it is a certain set of shared properties, they could be in the parent class.

dkriegner: I now checked and for the MKS937B and 974B only the serial number is common. There is some minor overlap with respect to the RelayChannels/unit. I will look into it if there are options to still put few things to parent classes (and at the same time improve the 937B instrument -> does currently not have relay channels which it also supports)

pymeasure/instruments/mksinst/mks974b.py
dkriegner: Ok, lets go with RelayChannel. I will revise the code accordingly

dkriegner: I now moved the `RelayChannel` to the common MKS file. `Relay` is used in the respective device files to complete the implementation. I hope the docstring of `RelayChannel` is inherited to `Relay`

pymeasure/instruments/mksinst/mks974b.py
dkriegner: I now use an `Enum`. Since I did not see some hints in the documentation on how the Enum should be used I hope I found the intended way. Note that I introduced the use of Enum also in the MKS937B. This is a change with respect to v.0.13.1 so maybe needs to marked somehow?

pymeasure/instruments/mksinst/mks974b.py
dkriegner: I kept the properties separate since I think the overlap is small. Even the unit property which one might assume to be the same is different in two minor details and I see no benefit in putting it to the central code.

BenediktBurger: As you say: If the overlap is minimal, it is more work and more difficult to understand to put this stuff in a common superclass.

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -127,13 +177,11 @@ def __init__(self, adapter, name="MKS 937B vacuum gauge controller", address=253
...
+
+        Allowed units are Unit.Torr, Unit.mbar, Unit.Pa, Unit.uHg""",
         validator=strict_discrete_set,
         map_values=True,
-        values={"Torr": "TORR",
-                "mBar": "mBAR",
-                "Pascal": "PASCAL",
-                "Micron": "MICRON",
-                },
+        values={u: u.value for u in Unit},
```
BenediktBurger, line 185: I think, that we should allow the old values as well in order to be backward compatible.

pymeasure/instruments/mksinst/mks974b.py
```diff
@@ -195,13 +163,12 @@ def id(self):
...
+        """Control pressure unit used for all pressure readings from the instrument.
+
+        Allowed units are Unit.Torr, Unit.mbar, Unit.Pa.""",
         validator=strict_discrete_set,
         map_values=True,
-        values={"Torr": "TORR",
-                "mBar": "MBAR",
-                "Pascal": "PASCAL",
-                },
+        values={u: u.value for u in Unit},
```
BenediktBurger, line 171: Interesting solution.
In https://github.com/pymeasure/pymeasure/issues/736#issuecomment-1318453826 I collected several methods to use enums with some comments.
I added your solution there.

BenediktBurger (commented):
Thanks for your changes.

pymeasure/instruments/mksinst/mks937b.py
dkriegner: I added back the old values in addition to the new ones. This solves setting the unit with the old string option. Still upon reading the new `Unit` enum is returned since I kept those at the top of the dictionary. It could be swapped to keep the old as default, but honestly I consider the enum the better solution and it just shifts the breaking change to a later point. Any ideas?

pymeasure/instruments/mksinst/mks974b.py
dkriegner: ok, i did not see this issue. Would be nice to have the best/recommended (if there is any) way in the documentation. Let me know if here you want it any different.

BenediktBurger: > Would be nice to have the best/recommended (if there is any) way in the documentation.

That was exactly the intend of that issue, to write a documentation, but we did not find a perfect solution.

Keep it as is.

pymeasure/instruments/mksinst/mks937b.py
BenediktBurger: I see. This instrument has been released, already, didn't it?

If you make the enum an `enum.StrEnum` (or `class StrEnum(str, Enum)` for python < 3.11), it will behave like a string, such that it will be backward compatible.

BenediktBurger: you can do
~~~python
try:
    from enum import StrEnum
except ImportError:
    from enum import Enum
    class StrEnum(str, Enum):
        pass

dkriegner: Yes the MKS937B was already part of the last published pymeasure version.

I now use StrEnum for both devices, but it is not backward compatible. I believe full backward compatibility can't be maintained while also moving to the enum. Personally I consider it worth breaking the backward compatibility in favor of moving to the enum which is more easy to use.

(Why is it not backward compatible: previously the user could set `dev.unit = 'Torr'`, even with `StrEnum` it is now `dev.unit = "TORR"`. Its even worse with `mbar` where the device requires a mix small/captital spelling of `mBAR`)

I am not sure what your policy is with respect to backward compatibility. Let me know how to proceed.

pymeasure/instruments/mksinst/mks937b.py
```diff
@@ -21,14 +21,22 @@
...
 from .mksinst import MKSInstrument, RelayChannel
 
 
+try:
+    from enum import StrEnum
+except ImportError:
+    from enum import Enum
+
+    class StrEnum(str, Enum):
+        """Until StrEnum is broadly available from the standard library"""
```
BenediktBurger, line 37: It's good to indicate, when this crutch can be removed, see #1038 (end of article)
```suggestion
        """Until StrEnum is broadly available from the standard library"""
        # Python>3.10 remove it
```

pymeasure/instruments/mksinst/mks974b.py
```diff
@@ -21,15 +21,23 @@
...
 
 
-class Unit(Enum):
+try:
+    from enum import StrEnum
+except ImportError:
+    from enum import Enum
+
+    class StrEnum(str, Enum):
+        """Until StrEnum is broadly available from the standard library"""
```
BenediktBurger, line 37:
```suggestion
        """Until StrEnum is broadly available from the standard library"""
        # Python>3.10 remove it.
```

pymeasure/instruments/mksinst/mks937b.py
BenediktBurger: The StrEnum makes the return value backward compatible.

For the accepted values, however, we should do it this way to include both possibilities:
~~~python
unit_dict = {u: u.value for u in Units}
unit_dict.update({
    "Torr": "TORR",
    # ...
})
~~~

dkriegner: I believe I am missing something. The return value is still `Unit.Torr` which also as StrEnum equals only "TORR" but not "Torr". So the backward compatibility issue which is difficult to solve is for reading. What you suggest and what I had before solves only the writing. To have it backward compatible for both reading and writing one would need to use the old values first. Then however the enum can only be used for writing. See [1dad20a](https://github.com/pymeasure/pymeasure/pull/1034/commits/1dad20ab85ca80e53dcc1658e140418d00190544)

BenediktBurger:
Are you ready to merge?

dkriegner:
yes please merge. thank you for all the feedback.

PR #1035, opened 2024-01-29
Redpitaya patch

seb5g:
This is a patch of the initial PR #1010 modifying the method to fetch data (easier if triggering) and renamed a setting that didn't make sense

### DIFF
--- a/pymeasure/instruments/redpitaya/redpitaya_scpi.py
+++ b/pymeasure/instruments/redpitaya/redpitaya_scpi.py
@@ -119,29 +119,33 @@ class AnalogInputFastChannel(Channel):
         values=['LV', 'HV'],
     )
 
-    def get_data_from_ascii(self, start=0, npts: int = None) -> np.ndarray:
-        """ Read data from the buffer from ascii format, see :meth:acq_format
+    def get_data(self, npts: int = None, format='ASCII') -> np.ndarray:
+        """ Read data from the buffer
 
-        :param start: starting point in the buffer
         :param npts: number of points to be read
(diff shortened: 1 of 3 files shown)


seb5g:
I've build a pymodaq plugin using it which is now fully functional so the acquisition part of the driver should not change much than this

pymeasure/instruments/redpitaya/redpitaya_scpi.py
```diff
@@ -119,14 +119,33 @@ class AnalogInputFastChannel(Channel):
...
     )
 
-    def get_data_from_ascii(self) -> np.ndarray:
-        self.write("ACQ:SOUR{ch}:DATA?")
+    def get_data(self, npts: int = None, format='ASCII') -> np.ndarray:
+        """ Read data from the buffer
+
+        :param npts: number of points to be read
+        :param format: either 'ASCII' or 'BIN', see :meth:acq_format
+        :param center_trigger: if True data are symmetric around the trigger event position
```
BenediktBurger, line 127: There is no center trigger parameter!

BenediktBurger (approved):
Thanks for the improvement.

We keep code backward compatible and raise deprecation warnings before removal, if there is a removal at all.

pymeasure/instruments/redpitaya/redpitaya_scpi.py
seb5g: done

BenediktBurger:
May I merge the changes?

seb5g:
Yes please

Issue #1014, opened 2024-01-02
Proposal: Possibility to not save all parameters.


poje42:
The idea is to have input parameters that are not saved as Parameters in the result files. I see at least two uses for this:

1. To fit a GUI with many parameters nicely on the screen, the 'group_by'-parameter can be used. The parameter used to show the different groups needs not be saved.
2. An ID of the DUT that the user enters and will be saved as metadata, as you don't want the parameter to be overwritten if you load parameters from a result file.

I've made a proposal in PR #1013 to add a Boolean 'save' parameter to the Parameter class and an example. 
Maybe this is not the right way to go, I welcome better ideas!

BenediktBurger:
@msmttchr , @CasperSchippers , what do you think, as it touches what you mostly use?

CasperSchippers:
I think this is a good idea.

I do see a conflict between your two usecases with a single `save` boolean: you do not want to write the group-by parameter in your example to the file if you specify `save=False`; however, for the ID you do want to write it to file, but not load it upon using the parameters of the file. So that means that the `save` boolean has two different meanings, if I understand you correctly?

poje42:
If you look in the example I've made you can see that I assign the ID parameter value to metadata (gui_no_save_param.py row 74) for use case 2.
I would say that use case 1 is "nice to have" and use case 2 is an actual issue I had.

Of course, it would be possible to add a metadata boolean to the Parameter class, but I saw this proposal as simpler and still make it possible to solve two use cases.

CasperSchippers:
Ahh yes, missed that line; thanks for the explanation!

Due to the fact that I missed that, I actually started thinking that maybe explicit is better than implicit/obfuscated.
What I'm thinking now (and don't yet know how difficult/easy this would be) is that it could be most clear if the metadata class has some sort of "user input" toggle, or indeed a "metadata" toggle for a parameter; not sure what the best approach is het, so feel let me know what you think.

poje42:
Well, how to define "best"? :)
My initial thoughts was make it simple. The parameter already has the user interface (input fields, visibility control etc), types (float, list, integer etc), it was only missing that I don't want it to be overwritten.  I can do any type of operation with input information before setting the metadata field(s) in startup(). Also, even if I don't see it today, a general input function may have other uses than those mentioned above.
As I understand your thoughts, the intention is to tighter connect metadata with the user input, which may be cleaner?

In that case I think we first should which structure we think is best:
Structure A
[...]

PR #1036, opened 2024-01-29
Feature: restructure Parameter and Metadata and create non-saved input-fields

CasperSchippers:
A suggested implemented for #1014 (and PR #1013).

### DIFF
--- a/pymeasure/experiment/parameters.py
+++ b/pymeasure/experiment/parameters.py
@@ -505,47 +505,6 @@ class PhysicalParameter(Parameter, PhysicalInputField):
     pass
 
 
-class Measurable:
-    """ Encapsulates the information for a measurable experiment parameter
-    with information about the name, fget function and units if supplied.
-    The value property is called when the procedure retrieves a datapoint
-    and calls the fget function. If no fget function is specified, the value
-    property will return the latest set value of the parameter (or default
-    if never set).
-
-    :var value: The value of the parameter
-
-    :param name: The parameter name
-    :param fget: The parameter fget function (e.g. an instrument parameter)
-    :param default: The default value
-    """
-    DATA_COLUMNS = []
-
-    def __init__(self, name, fget=None, units=None, measure=True, default=None, **kwargs):
-        self.name = name
-        self.units = units
-        self.measure = measure
-        if fget is not None:
-            self.fget = fget
-            self._value = fget()
-        else:
-            self._value = default
-        Measurable.DATA_COLUMNS.append(name)
-
-    def fget(self):
-        return self._value
-
-    @property
-    def value(self):
-        if hasattr(self, 'fget'):
-            self._value = self.fget()
-        return self._value
-
-    @value.setter
-    def value(self, value):
-        self._value = value
-
-
 class Metadata(InputField):
     """ Encapsulates the information for metadata of the experiment with
     information about the name, the fget function and the units, if supplied.
@@ -658,3 +617,44 @@ class ListMetadata(Metadata, ListInputField):
 
 class PhysicalMetadata(Metadata, PhysicalInputField):
     pass
+
+
+class Measurable:
+    """ Encapsulates the information for a measurable experiment parameter
+    with information about the name, fget function and units if supplied.
+    The value property is called when the procedure retrieves a datapoint
+    and calls the fget function. If no fget function is specified, the value
+    property will return the latest set value of the parameter (or default
+    if never set).
+
+    :var value: The value of the parameter
+
+    :param name: The parameter name
+    :param fget: The parameter fget function (e.g. an instrument parameter)
+    :param default: The default value
+    """
+    DATA_COLUMNS = []
+
+    def __init__(self, name, fget=None, units=None, measure=True, default=None, **kwargs):
+        self.name = name
+        self.units = units
+        self.measure = measure
+        if fget is not None:
+            self.fget = fget
+            self._value = fget()
+        else:
+            self._value = default
+        Measurable.DATA_COLUMNS.append(name)
+
+    def fget(self):
+        return self._value
+
+    @property
+    def value(self):
+        if hasattr(self, 'fget'):
+            self._value = self.fget()
+        return self._value
+
+    @value.setter
+    def value(self, value):
+        self._value = value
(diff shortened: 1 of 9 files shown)


CasperSchippers:
@BenediktBurger, @bilderbuchi, and/or @msmttchr: what do you think of a structure like what I proposed here?

I'm not sure now what I think of having all these classes (a lot of which are just inheritances without any changes, just to be able to 'dispatch' on the class-type). Maybe it's fine like this, but am not sure yet, so I'm curious what you think of this structure and, if you don't like it, if you have any other/better suggestions?

pymeasure/experiment/parameters.py
```diff
@@ -21,23 +21,36 @@
...
-class Parameter:
-    """ Encapsulates the information for an experiment parameter
-    with information about the name, and units if supplied.
+class InputField:
+    """ Encapsulates the information for an input of an experiment with information about the name.
 
-    :var value: The value of the parameter
+    :class:`.InputField` serves as a parent-class to :class:`.Parameter` and :class:`.Metadata` (and
+    their respective subclasses). These classes differ in whether (and if yes, when) their value is
+    writen to the data-file, and whether the value is applied when getting parameters from a file.
```
BenediktBurger, line 32:
```suggestion
    written to the data-file, and whether the value is applied when getting parameters from a file.
```

pymeasure/experiment/parameters.py
```diff
@@ -90,15 +103,15 @@ def cli_args(self):
...
+        - index 1: List of value to format a help string, that is either,
         the name of the fields to be documented or a tuple with (helps_string,
         field)
         - index 2: type
         """
         return (self.default, self._help_fields, self.convert)
 
     def is_set(self):
-        """ Returns True if the Parameter value is set
+        """ Returns True if the InputField value is set
```
BenediktBurger, line 114:
```suggestion
        """ Return True if the InputField value is set
```

pymeasure/experiment/parameters.py
```diff
@@ -468,92 +472,165 @@ def __repr__(self):
...
-    The value property is called when the procedure retrieves a datapoint
-    and calls the fget function. If no fget function is specified, the value
-    property will return the latest set value of the parameter (or default
-    if never set).
+class Parameter(InputField):
+    """ Encapsulates the information for a parameter of an experiment with information about the
+    name.
 
-    :var value: The value of the parameter
+    The data dat is stored in a parameter is written to the parameters section of the data file
```
BenediktBurger, line 479:
```suggestion
    The data that is stored in a parameter is written to the parameters section of the data file
```

pymeasure/experiment/parameters.py
```diff
@@ -468,92 +472,165 @@ def __repr__(self):
...
+    """
+    pass
+
+
+class Metadata(InputField):
     """ Encapsulates the information for metadata of the experiment with
     information about the name, the fget function and the units, if supplied.
     If no fget function is specified, the value property will return the
     latest set value of the parameter (or default if never set).
+    Subclass of :class:`.InputField`
```
BenediktBurger, line 588: Is it necessary to state the parent class?

In the api documentation you can toggle "show_inheritance" to show that information.

That makes the actual inheritance the single point of truth.

pymeasure/experiment/parameters.py
```diff
@@ -581,9 +658,79 @@ def eval_fget(self, parent):
...
 
-        if self.units is not None:
+        if hasattr(self, 'units') and self.units is not None:
             result += " %s" % self.units
 
         return result
+
+
+class IntegerMetadata(Metadata, IntegerInputField):
+    pass
```
BenediktBurger, line 675: Why do you have lengthy docstrings for the parameters, but not for the Metadata?

Naive, I would not add docstrings to either one.

BenediktBurger (commented):
A few comments (mainly regarding documentation typos).

I see the introduction of many very similar classes, just different in type.

Another idea: what about a "type" parameter for the Input Field?
Valid arguments could be native types (float, int...) or an enum member (especially for those additional types).
So it would be `steps = InputField(type=int)` (or Parameter or Metadata instead of InputField) or `type=Types.INT` using an enum.

That is easy, but reduces the number of classes drastically.

pymeasure/experiment/parameters.py
CasperSchippers: I simply did not get around to doing that. I was not yet completely finished with this, but mainly wondered about the structure with all the classes.

CasperSchippers:
> Another idea: what about a \"type\" parameter for the Input Field?

I chose the classes approach for symmetry to the existing set of Parameter classes.

Hmmm, that is maybe cleaner. However, I do not want to break backwards compatibility, so I'm not sure if we (ever) want to remove the parameter classes.

BenediktBurger:
> An option would be to rewrite it with the types as you suggest, and add the set of Parameter classes for backwards compatibility?

Reading the email, I wanted to suggest exactly that, just to note, that you added it yourself 😁

CasperSchippers:
Mainly as a note to myself (though feel free to criticize)

One challenge I saw with having a single parameter class was the fact that all the different subclasses now have custom `convert`, `__str__`, and `__repr__` methods.
One way to solve this is to create gigantic versions of that method with some if-else switch statement (match-case could be a better fit  but is nog introduced in 3.10) to effectively dispatch between different versions; seems ugly though.

A solution might be composition: have a Parameter class, which is effectively a type agnostic container, but holds a Value class that serves the different types. The `convert`, `__str__` and `__repr__` of Parameter would simply pass through to this Value class. Parameter would also have some flags, like `write_to_file`, `load_from_file` and `is_metadata`.

If going this way, I'm also inclined to not make `string` the 'default'/base class but rather have an abstract base, and string just one of the specific instances.

bilderbuchi:
Composition over inheritance is often recommended in principle, so the sketched approach sounds reasonable to me. :+1:

I can't judge the relative merits of the other proposal, I've never been deeply immersed in that part of the codebase.

BenediktBurger:
> A solution might be composition: have a Parameter class, which is effectively a type agnostic container, but holds a Value class that serves the different types. The `convert`, `__str__` and `__repr__` of Parameter would simply pass through to this Value class. Parameter would also have some flags, like `write_to_file`, `load_from_file` and `is_metadata`.

I like that solution in favor of complicated methods.

> If going this way, I'm also inclined to not make `string` the 'default'/base class but rather have an abstract base, and string just one of the specific instances.

Sounds good.

> I can't judge the relative merits of the other proposal, I've never been deeply immersed in that part of the codebase.

The final decision on the approach you have to do yourself, maybe with @msmttchr, as neither I am an expert in that part of the codebase

PR #1039, opened 2024-02-01
Make CI publish on release creation.

BenediktBurger:
Requires also the following steps:

- [x] Create environment `release`
- [x] Add this configuration to trusted publishers on pypi
- [x] Adjust RELEASE.md

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,31 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Build and Upload Python Package
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+    environment: release
+    permissions:
+      id-token: write  # IMPORTANT: this permission is mandatory for trusted publishing
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install build twine
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@v1.8.11
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: '3.x'
     - name: Install dependencies
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -28,4 +28,4 @@ jobs:
     - name: Build package
       run: python -m build
     - name: Publish package
-      uses: pypa/gh-action-pypi-publish@v1.8.11
+      uses: pypa/gh-action-pypi-publish@release/v1
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -16,20 +16,18 @@
 5. Push the changes up as a PR
 6. Verify that the builds complete
 7. Merge the PR
-8. Fetch `master`, build and check the source packages
-    - `python -m pip install --upgrade build twine`
-    - `python -m build`
-    - Check the distributions (`twine check dist/*`, version will not yet be correct)
-9. Create a git tag in the format "vX.Y.Z"
-10. Build final packages and confirm the correct version number is being used
-     - `python -m build`
-     - Check the distributions (`twine check dist/*`)
-11. Push the Git tag
-12. Create a tagged [release on GitHub](https://github.com/pymeasure/pymeasure/releases). You'll have to paste in the changelog entry and probably edit it a bit as that form expects Markdown, not ReST (probably just removing `:code:` tags will be sufficient).
+8. Create a new [release on GitHub](https://github.com/pymeasure/pymeasure/releases)
+    * Add a tag name in the format "vX.Y.Z" and select "create tag on publish"
+    * You'll have to paste in the changelog entry and probably edit it a bit as that form expects Markdown, not ReST (probably just removing `:code:` tags will be sufficient).
+    * Publish the release
+8. Approve the _build and upload_ run under Actions.
+   This will create the wheel and upload it to PyPI.
 
-## PyPI release
+## PyPI release - manually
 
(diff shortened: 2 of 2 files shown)


BenediktBurger:
This PR adds the workflow, however a few steps with admin rights in this repo / at pypi are still necessary. @cjermain , @bilderbuchi 
If these steps are done, we can merge.

BenediktBurger:
Bump setup-python version to most recent version.

bilderbuchi:
One thing I am not sure about is the change to the release workflow -- what happens if there is a problem/issue with creating the Python _package_? 
Currently, this is caught because the packager does a package  build and test upload to pypi/test from the local machine, and only tags _afterwards_.
With the new workflow, IIUC, one would finish the changelog, create a tag, push the tag, create a GH release on the web interface, and _then_ the package is autocreated, which means problems would be discovered only after the tag has been created, which would necessitate another immediate bugfix release.

Or is this really an edge-case concern? I don't remember encountering a problem in the local package-building step before, so maybe it's not worth taking this into account, and the benefit of packaging automation by far outweighs that.

BTW, this PR should also update RELEASE.md with the updated workflow.

BenediktBurger:
> BTW, this PR should also update RELEASE.md with the updated workflow.

That is, what I meant with the gist, as I thought, the realease workflow is in a gist.
I updated the ToDos in the beginning.

> With the new workflow, IIUC, one would finish the changelog, create a tag, push the tag, create a GH release on the web interface, and then the package is autocreated, which means problems would be discovered only after the tag has been created, which would necessitate another immediate bugfix release.

The created release is only a _github release_. You can delete it (and the tag), as it is unlikely, that someone pulled the tag in exactly the moment between creating a github release and the push to pypi.

I encountered it in pyleco, when I made a wrong tag name (dot between v and version), such that the creation failed. I deleted the github release and made a new one with the same commit.

bilderbuchi:
> You can delete it (and the tag), as it is unlikely, that someone pulled the tag in exactly the moment between creating a github release and the push to pypi.

Yeah, although this violates the general git best practice that public history should be immutable (which we mainly practice with main/develop branches anyway as we allow force-pushing PR branches). So, having this as the formal/defined way is something I would avoid. Personally, I've always shied away from recreating an already pushed _tag_. 
Btw, it could be that a CI, bot or other automated process watching our repo would act on the fresh tag.

I'm wondering how other projects using that action see that concern, or if there's a FAQ or best practice defined?

bilderbuchi:
Out of curiosity, did you see that you could also trigger the publish workflow on new tags? https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#defining-a-workflow-job-environment
Then the GH release is independent of the pypi stuff, and this manual part could be done afterwards. 
Not sure what's better, I'm not particularly fond of the "Releases" thing which kinda mirrors/duplicates what Git tags already are for.

BenediktBurger:
For pyleco, I did 
* the draft release (keeping it open)
* created the changelog from there,
* pushed the changelog
* published the release

which worked well for me.

> Not sure what's better, I'm not particularly fond of the "Releases" thing which kinda mirrors/duplicates what Git tags already are for.

I see it as a special form of tags: You have that prominent window, showing you, the current version and the last versions (you can have more tags than just versions).

> Out of curiosity, did you see that you could also trigger the publish workflow on new tags?

I did not look for that.

BenediktBurger:
> Yeah, although this violates the general git best practice that public history should be immutable (which we mainly practice with main/develop branches anyway as we allow force-pushing PR branches). So, having this as the formal/defined way is something I would avoid. Personally, I've always shied away from recreating an already pushed tag.
Btw, it could be that a CI, bot or other automated process watching our repo would act on the fresh tag.

You have to create a tag for setuptools to succeed.
Therefore, if a build fails and you have to do a change (new commit), you have to create a second tag.
In this case you have to increase the patch number by one, I'd say.

We could do as a workflow either:
- mark a release only as "latest", after the publishing to PyPI succeeded (maybe that could be automated too?)
- publish to PyPI on tag creation and create a github release afterwards

Anyways, as you said, github releases are (for python packages on PyPI) not very important.

BenediktBurger:
@bilderbuchi do you want to think about this, as a preparation for the 0.14 release?

RELEASE.md
```diff
@@ -16,21 +16,19 @@
...
+    * Publish the release
+8. Approve the _build and upload_ run under Actions.
+   This will create the wheel and upload it to PyPI.
+
+## PyPI release - manually
 
 Official guide [here](https://packaging.python.org/en/latest/tutorials/packaging-projects/).
 
+If the upload action does not work, you can create a PyPI release manually:
+
```
bilderbuchi, line 31: this is missing some parts of the steps 8/9/10 above that now are gone... if the upload action does not work, you don't have a built wheel locally or on the server.

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,31 @@
...
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install build twine
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@v1.8.11
```
bilderbuchi, line 31: this is a very strict version, normally I'd expect to see `@v1` to make use of subsequent (patch) releases. Any particular reason why you pin so hard here?

bilderbuchi:
> a few steps with admin rights in this repo / at pypi are still necessary.

not sure what you are referring to here, but AFAIK only Colin has the admin bit on this repo (but I could not confirm that, my settings view is quite limited). We should probably take this up with him by email, if we can extend that circle. Otherwise we'd have to again put him into the critical path for getting releases out.

BenediktBurger:
>> a few steps with admin rights in this repo / at pypi are still necessary.

I referred to the tasks in the first post: creating the `release` environment and adding the release file to the trusted publishers on pypi.

> Otherwise we'd have to again put him into the critical path for getting releases out.

No, once he has set it up, we don't need the admin rights anymore.

RELEASE.md
BenediktBurger: I added them.

.github/workflows/python-publish.yml
BenediktBurger: You have to use a specific version. The author does not provide a v1 tag, but only a v1 release branch, I'll switch to that one.

RELEASE.md
bilderbuchi: thanks!

bilderbuchi:
OK then, fine with me :shipit:, also to try this out with the upcoming release. :-)

BenediktBurger:
As Colin did the requirements, we can merge, I think.
@bilderbuchi?

bilderbuchi:
> As Colin did the requirements, we can merge, I think. @bilderbuchi?

LGTM!

PR #1043, opened 2024-02-14
dev/keithley 2182

ConnorGCarr:
- Create a driver for the Keithley-2182 nanovoltmeter
- Small bug fix on the keithley buffer (bitwise comparison with status bit)

### DIFF
--- a/pymeasure/instruments/keithley/keithley2182.py
+++ b/pymeasure/instruments/keithley/keithley2182.py
@@ -26,7 +26,9 @@ import logging
 import time
 
 from pymeasure.instruments import Instrument, Channel
-from pymeasure.instruments.validators import truncated_range, strict_discrete_set, strict_range
+from pymeasure.instruments.validators import (truncated_range,
+                                              strict_discrete_set,
+                                              strict_range)
 
 from .buffer import KeithleyBuffer
 
@@ -50,7 +52,7 @@ class Keithley2182Channel(Channel):
 
     def __init__(self, instrument, id_):
         """Set max voltage depending on channel."""
-        if id_==1:
+        if id_ == 1:
             self.voltage_range_values = [0, 120]
             self.voltage_offset_values = [-120, 120]
         else:
@@ -60,7 +62,7 @@ class Keithley2182Channel(Channel):
 
     voltage_range = Channel.control(
         ":SENS:VOLT:CHAN{ch}:RANG?", ":SENS:VOLT:CHAN{ch}:RANG %g",
-        """ A floating point property that controls the positive full-scale 
+        """ A floating point property that controls the positive full-scale
         measurement voltage range in Volts, which can take values from 0 to
         120 V (Ch. 1) or 12 V (Ch. 2). Auto-range is automatically disabled
         when this property is set. """,
@@ -110,40 +112,40 @@ class Keithley2182Channel(Channel):
         values=['OFF', 'ON'],
     )
 
-    def measure_voltage(self, auto_range=True, nplc=1):
-        """ Configures and acquires voltage measurement.
-        
-        :param auto_range: Enables auto_range if True, else uses set voltage range
+    def setup_voltage(self, auto_range=True, nplc=1):
+        """Changes active channel and configures channel for voltage
+            measurement.
+
+        :param auto_range: Enables auto_range if True, else uses set voltage
+            range
         :param nplc: Number of power line cycles (NPLC) from 0.01 to 50/60
         """
-        log.info("%s is measuring voltage.", self.name)
         self.write(":SENS:CHAN {ch};"
                    ":SENS:FUNC 'VOLT';"
                    f":SENS:VOLT:NPLC {nplc};")
         if auto_range:
-            self.write(":SENS:VOLT:RANG:AUTO 1;")
-        self.write(":READ?")
+            self.write(":SENS:VOLT:RANG:AUTO 1")
         self.check_errors()
 
-    def measure_temperature(self, nplc=1):
-        """ Configures and acquires temperature measurement.
+    def setup_temperature(self, nplc=1):
+        """Changes active channel and configures channel for temperature
+            measurement.
 
         :param channel: Temperature measurement channel
         :param nplc: Number of power line cycles (NPLC) from 0.01 to 50/60
         """
-        log.info("%s is measuring temperature.", self.name)
         self.write(":SENS:CHAN {ch};"
                    ":SENS:FUNC 'TEMP';"
-                   f":SENS:CURR:NPLC {nplc};")
-        self.write(":READ?")
+                   f":SENS:TEMP:NPLC {nplc}")
         self.check_errors()
 
     def acquire_reference(self, func='VOLT'):
         """Acquire a measurement and store it as the relative offset value.
-        
+
         :param func: 'VOLT' or 'TEMP'
         """
         self.write(f":SENS:{func}:""CHAN{ch}:REF:ACQ")
+
 
 class Keithley2182(KeithleyBuffer, Instrument):
     """Represents the Keithley 2182 Nanovoltmeter and provides a
@@ -158,9 +160,9 @@ class Keithley2182(KeithleyBuffer, Instrument):
         keithley.active_channel = 1             # Sets channel 1 for active measurement
         keithley.channel_function = 'voltage'   # Configures active channel for voltage measurement
         print(keithley.voltage)                 # Prints the voltage in volts
-        
-        keithley.ch_1.measure_voltage()         # Measure voltage on channel 1
-        keithley.ch_2.measure_temp()            # Measure temperature on channel 2
+
+        volt = keithley.ch_1.measure_voltage()  # Measure voltage on channel 1
+        temp = keithley.ch_2.measure_temp()     # Measure temperature on channel 2
 
     """
 
@@ -202,8 +204,8 @@ class Keithley2182(KeithleyBuffer, Instrument):
     )
     channel_function = Instrument.control(
         ":SENS:FUNC?", "SENS:FUNC %s",
-        """ A property that controls the measurement mode of the active channel.
-        Valid options are voltage and temperature. """,
+        """ A property that controls the measurement mode of the active
+        channel. Valid options are voltage and temperature. """,
         validator=strict_discrete_set,
         values={'voltage': 'VOLT', 'temperature': 'TEMP'},
         map_values=True
@@ -240,11 +242,11 @@ class Keithley2182(KeithleyBuffer, Instrument):
         for this reading. """
     )
     thermocouple = Instrument.control(
-        ":SENS:TEMP:TC?", ":SENSE:TEMP:TC %s",
+        ":SENS:TEMP:TC?", ":SENS:TEMP:TC %s",
         """ A property that controls the thermocouple type for temperature
         measurements. Valid options are B, E, J, K, N, R, S, and T. """,
         validator=strict_discrete_set,
-        values=['B','E','J','K','N','R','S','T']
+        values=['B', 'E', 'J', 'K', 'N', 'R', 'S', 'T']
     )
     temperature_nplc = Instrument.control(
         ":SENS:TEMP:NPLC?", ":SENS:TEMP:NPLC %g",
@@ -262,7 +264,7 @@ class Keithley2182(KeithleyBuffer, Instrument):
         """ A property that controls the type of thermocouple reference
         junction. Default is internal.""",
         validator=strict_discrete_set,
-        values=['SIM','INT'],
+        values=['SIM', 'INT'],
     )
     temperature_simulated_reference = Instrument.control(
         ":SENS:TEMP:RJUN:SIM?", ":SENS:TEMP:RJUN:SIM %g",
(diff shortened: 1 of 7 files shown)


pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,409 @@
...
+            self.voltage_range_values = [0, 120]
+            self.voltage_offset_values = [-120, 120]
+        else:
+            self.voltage_range_values = [0, 12]
+            self.voltage_offset_values = [-12, 12]
+        super().__init__(instrument, id_)
+
+    voltage_range = Channel.control(
+        ":SENS:VOLT:CHAN{ch}:RANG?", ":SENS:VOLT:CHAN{ch}:RANG %g",
+        """ A floating point property that controls the positive full-scale
```
BenediktBurger, line 65: Docstrings of Proterties should start with "Set", "Get" (or "Measure"), or "Control", depending on the type, see https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings .

Did you copy that code? Maybe you could inherit?

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,409 @@
...
+        in seconds, which can take values from 0 to 999999.999 s. """,
+        validator=truncated_range,
+        values=[0, 999999.999]
+    )
+
+    ###########
+    # Methods #
+    ###########
+
+    def __init__(self, adapter, name="Keithley 2182 Nanovoltmeter",
```
BenediktBurger, line 339: It's good to have the init at the top, to find it easily.

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,409 @@
...
+                 read_termination='\r', **kwargs):
+        super().__init__(
+            adapter, name, **kwargs
+        )
```
BenediktBurger, line 343: Please add, whether it supports SCPI or not, for example if it does:
```suggestion
        super().__init__(
            adapter, name, includeSCPI=True, **kwargs
        )
```

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -62,61 +62,58 @@ def __init__(self, instrument, id_):
...
     voltage_range_auto = Channel.control(
         ":SENS:VOLT:CHAN{ch}:RANG:AUTO?", ":SENS:VOLT:CHAN{ch}:RANG:AUTO %d",
-        """ A boolean property that controls whether auto voltage ranging is
-        enabled. Valid values are True and False.""",
+        """Control whether auto voltage ranging is enabled.
+        Valid values are True and False.""",
         validator=strict_discrete_set,
         values={True: 1, False: 0},
         map_values=True,
     )
```
BenediktBurger, line 79: For boolean properties, it is good to use a name like `voltage_range_auto_enabled`, which makes the code descriptive:
`voltage_range_auto_enabled = True` is clear, while otherwise you never know, what True means: enabled or disabled.

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,399 @@
...
+        Valid ranges are from 0 to 120 V for Ch. 1, and 0 to 12 V for Ch. 2.
+        Auto-range is automatically disabled when this property is set.""",
+        validator=truncated_range,
+        values=[0, 120],
+        dynamic=True,
+    )
+    voltage_range_auto = Channel.control(
+        ":SENS:VOLT:CHAN{ch}:RANG:AUTO?", ":SENS:VOLT:CHAN{ch}:RANG:AUTO %d",
+        """Control whether auto voltage ranging is enabled.
+        Valid values are True and False.""",
```
BenediktBurger, line 75: For boolean properties, it is good to use a name like `voltage_range_auto_enabled`, which makes the code descriptive:
`voltage_range_auto_enabled = True` is clear, while otherwise you never know, what True means: enabled or disabled.
```suggestion
    voltage_range_auto_enabled = Channel.control(
        ":SENS:VOLT:CHAN{ch}:RANG:AUTO?", ":SENS:VOLT:CHAN{ch}:RANG:AUTO %d",
        """Control whether auto voltage ranging is enabled.
        Valid values are True and False.""",
```

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,399 @@
...
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+    voltage_offset = Channel.control(
```
BenediktBurger, line 80: Please add empty lines between two properties:
```suggestion
    )

    voltage_offset = Channel.control(
```

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,399 @@
...
+    )
+    voltage_offset_state = Channel.control(
+        ":SENS:VOLT:CHAN{ch}:REF:STAT?", ":SENS:VOLT:CHAN{ch}:REF:STAT %s",
+        """Control whether voltage is measured as a relative or absolute value.
+        Enabled by default for Ch. 2 voltage, which is measured relative to Ch. 1
+        voltage.""",
+        validator=strict_discrete_set,
+        values={'OFF': 0, 'ON': 1},
+        map_values=True
+    )
```
BenediktBurger, line 105: Why don't you make it boolean (also changing the name)?
```suggestion
    voltage_offset_enabled = Channel.control(
        ":SENS:VOLT:CHAN{ch}:REF:STAT?", ":SENS:VOLT:CHAN{ch}:REF:STAT %s",
        """Control whether voltage is measured as a relative or absolute value.
        Enabled by default for Ch. 2 voltage, which is measured relative to Ch. 1
        voltage (bool).""",
        validator=strict_discrete_set,
        values={False: 0, True: 1},
        map_values=True
    )
```

similarly below.

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,399 @@
...
+        self.check_errors()
+
+    def acquire_reference(self, func='VOLT'):
+        """Acquire a measurement and store it as the relative offset value.
+
+        Only acquires reference if offset state is ON.
+
+        :param func: 'VOLT' or 'TEMP'
+        """
+        self.write(f":SENS:{func}:""CHAN{ch}:REF:ACQ")
```
BenediktBurger, line 147: Consider making two methods, one for voltage and one for temperature.
Using strings for choosing options is not that handy.

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,399 @@
...
+    #################
+
+    auto_zero = Instrument.control(
+        ":SYST:AZER:STAT?", ":SYST:AZER:STAT %d",
+        """Control the auto zero option.
+        Valid values are True (enabled) and False (disabled).""",
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
```
BenediktBurger, line 189: Here again, using `enabled` in the name, makes it self evident, without the need to write the values into the docstring:
```suggestion
    auto_zero_enabled = Instrument.control(
        ":SYST:AZER:STAT?", ":SYST:AZER:STAT %d",
        """Control the auto zero option (bool).""",
        validator=strict_discrete_set,
        values={True: 1, False: 0},
        map_values=True,
    )
```

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,399 @@
...
+
+    @property
+    def error(self):
+        """Get a tuple of an error code and message from a single error."""
+        err = self.values(":system:error?")
+        if len(err) < 2:
+            err = self.read()  # Try reading again
+        code = err[0]
+        message = err[1].replace('"', '')
+        return (code, message)
```
BenediktBurger, line 356: If I remember correctly, this property is called `next_error` and is already included for SCPI devices. See `Instrument` itself.

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,399 @@
...
+
+    def check_errors(self):
+        """Log any system errors reported by the instrument."""
+        code, message = self.error
+        while code != 0:
+            t = time.time()
+            log.info("Keithley 2182 reported error: %d, %s", code, message)
+            code, message = self.error
+            if (time.time() - t) > 10:
+                log.warning("Timed out for Keithley 2182 error retrieval.")
```
BenediktBurger, line 366: Check, whether that code is necessary, or it is sufficient to use the inherited method (from `Instrument`).

BenediktBurger (changes requested):
A second round of comments.

pymeasure/instruments/keithley/keithley2182.py
ConnorGCarr: Thanks for the suggestion, I wasn't aware `next_error` and `check_errors` were already implemented.

pymeasure/instruments/keithley/keithley2182.py
ConnorGCarr: Thanks for the advice on boolean naming, I've updated the properties accordingly

ConnorGCarr:
I'm done with edits on my side, unless there are additional changes that need to be made.

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,390 @@
...
+
+    def __init__(self, instrument, id_):
+        """Set max voltage depending on channel."""
+        if id_ == 1:
+            self.voltage_range_values = [0, 120]
+            self.voltage_offset_values = [-120, 120]
+        else:
+            self.voltage_range_values = [0, 12]
+            self.voltage_offset_values = [-12, 12]
+        super().__init__(instrument, id_)
```
BenediktBurger, line 60: Please follow the parent class name scheme: `parent` and `id`:
```suggestion
    def __init__(self, parent, id):
        """Set max voltage depending on channel."""
        if id == 1:
            self.voltage_range_values = [0, 120]
            self.voltage_offset_values = [-120, 120]
        else:
            self.voltage_range_values = [0, 12]
            self.voltage_offset_values = [-12, 12]
        super().__init__(parent, id)
```

Basically, you could use the Channel as a child of another channel as well, therefore, parent is more generic than instrument.

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,390 @@
...
+
+    voltage_range = Channel.control(
+        ":SENS:VOLT:CHAN{ch}:RANG?", ":SENS:VOLT:CHAN{ch}:RANG %g",
+        """Control the positive full-scale measurement voltage range in Volts.
+        Valid ranges are from 0 to 120 V for Ch. 1, and 0 to 12 V for Ch. 2.
+        Auto-range is automatically disabled when this property is set.""",
+        validator=truncated_range,
```
BenediktBurger, line 67: I don't know, whether I mentioned it:
Use the `truncated` validators with care!

Imagine, someone writes `voltage_range=200`, which will be effectively only `voltage_range=120`, but the user does not know, that the voltage range is different from his value of 200.

A strict validator will raise an exception, which notifies the user of the wrong value.

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,390 @@
...
+                                                  voltage measurement
+        keithley.ch_2.setup_temperature()       # Set channel 2 active and prepare
+                                                  temperature measurement
+
+    """
+
+    def __init__(self, adapter, name="Keithley 2182 Nanovoltmeter",
+                 read_termination='\r', **kwargs):
+        super().__init__(adapter, name, read_termination=read_termination,
+                         includeSCPI=True, **kwargs)
```
BenediktBurger, line 184: Can you change the read_termination?
If not, you should not add it to the signature of the init, but hardcode it in the following way:
```suggestion
    def __init__(self, adapter, name="Keithley 2182 Nanovoltmeter",
                 **kwargs):
        super().__init__(adapter, name, read_termination="\r",
                         includeSCPI=True, **kwargs)
```

BenediktBurger (changes requested):
We're getting close.
Thanks for your patience.

A few comments.

pymeasure/instruments/keithley/keithley2182.py
ConnorGCarr: The terminator can be CR, LF, or a combination of both. The default isn't listed in the instrument documentation, but ours is set to CR, so I've made that the default here.

BenediktBurger: In this case, it is good as you did it.

A considerable amount of instruments have a fixed termination, where it useless, making it configurable. Therefore, I asked.

pymeasure/instruments/keithley/keithley2182.py
```diff
@@ -0,0 +1,390 @@
...
```
BenediktBurger, line 184: We recently changed the way, how to define the Scpi compliance, sorry for that:
Now you have to import "SCPIMixin" from instruments and add that to the inheritance:
"class KeithleyXYZ(SCPIMixin, KeithleyBuffer, Instrument):"
With that, you can remove the "includeSCPI" argument.

BenediktBurger:
May I merge this PR?

BenediktBurger:
Thanks for your contribution.

PR #1044, opened 2024-02-14
Dev/proterial rod4

ConnorGCarr:
Add Proterial (formerly Hitachi Metals) ROD-4 mass flow controller driver

### DIFF
--- a/docs/api/instruments/proterial/index.rst
+++ b/docs/api/instruments/proterial/index.rst
@@ -1,0 +1,12 @@
+.. module:: pymeasure.instruments.proterial
+
+#########
+Proterial
+#########
+
+This section contains specific documentation on the Proterial instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
+
+.. toctree::
+   :maxdepth: 2
+
+   rod4
--- a/docs/api/instruments/proterial/index.rst
+++ b/docs/api/instruments/proterial/index.rst
@@ -4,7 +4,7 @@
 Proterial
 #########
 
-This section contains specific documentation on the Proterial instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
+This section contains specific documentation on the Proterial (formerly Hitachi Metals) instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
 
 .. toctree::
    :maxdepth: 2
--- a/pymeasure/instruments/__init__.py
+++ b/pymeasure/instruments/__init__.py
@@ -63,6 +63,7 @@ from . import oxfordinstruments
 from . import parker
 from . import pendulum
 from . import racal
+from . import proterial
 from . import razorbill
 from . import rohdeschwarz
 from . import siglenttechnologies
--- a/pymeasure/instruments/__init__.py
+++ b/pymeasure/instruments/__init__.py
@@ -63,7 +63,6 @@ from . import oxfordinstruments
 from . import parker
 from . import pendulum
 from . import racal
-from . import proterial
 from . import razorbill
 from . import rohdeschwarz
 from . import siglenttechnologies
--- a/pymeasure/instruments/proterial/rod4.py
+++ b/pymeasure/instruments/proterial/rod4.py
@@ -37,31 +37,31 @@ class ROD4Channel(Channel):
     """ Implementation of a ROD-4 MFC channel """
 
     actual_flow = Channel.measurement(
-        "\020{ch}RFX",
+        "\x020{ch}RFX",
         """ Read the actual flow in % . """
     )
     setpoint = Channel.control(
-        "\020{ch}RFD", "\020{ch}SFD%.1f",
+        "\x020{ch}RFD", "\x020{ch}SFD%.1f",
         """A property that controls the set point in % of control range. """,
         validator=truncated_range,
         values=[0, 100],
     )
     mfc_range = Channel.control(
-        "\020{ch}RFK", "\020{ch}SFK%d",
+        "\x020{ch}RFK", "\x020{ch}SFK%d",
         """An integer property that controls the MFC range in sccm.
         Upper limit is 200 slm. """,
         validator=truncated_range,
         values=[0, 200000]
     )
     ramp_time = Channel.control(
-        "\020{ch}RRT", "\020{ch}SRT%.1f",
+        "\x020{ch}RRT", "\x020{ch}SRT%.1f",
         """A property that controls the MFC set point ramping time in
         seconds.""",
         validator=truncated_range,
         values=[0, 200000]
     )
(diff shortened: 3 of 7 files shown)


docs/api/instruments/proterial/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.proterial
+
+#########
+Proterial
+#########
+
+This section contains specific documentation on the Proterial instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
```
BenediktBurger, line 7: Here you could mention the old name of the manufacturer.

pymeasure/instruments/proterial/rod4.py
```diff
@@ -0,0 +1,144 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2023 PyMeasure Developers
```
BenediktBurger, line 4:
```suggestion
# Copyright (c) 2013-2024 PyMeasure Developers
```

pymeasure/instruments/proterial/rod4.py
```diff
@@ -0,0 +1,144 @@
...
+        """Control the MFC valve mode.
+        Valid options are `flow`, `close`, and `open`. """,
+        validator=strict_discrete_set,
+        values={'flow': 0, 'close': 1, 'open': 2},
+        map_values=True
+    )
+
+    def __init__(self, *argv, **kwargs):
+        super().__init__(*argv, **kwargs)
+        self._flow_unit = None
```
BenediktBurger, line 73: It is good to have the init at the top of the class, to find it more easily.

pymeasure/instruments/proterial/rod4.py
```diff
@@ -0,0 +1,144 @@
...
+            log.info("Flow units have not been set on ROD-4 channel "
+                     f"{self.id}")
+        return self._flow_unit
+
+    @flow_unit.setter
+    def flow_unit(self, units):
+        values = {'%': 0, 'sccm': 1, 'slm': 1}
+        if isinstance(units, str) and units.casefold() in values:
+            units = units.casefold()
+            self.write("\x020{ch}SFU%d" % values[units])
```
BenediktBurger, line 90: You could use normal f-strings:
```suggestion
            self.write(f"\x020{{ch}}SFU{value[units]}")
```
If you use double curly braces, they will be reduced by one, but not evaluated.

pymeasure/instruments/proterial/rod4.py
```diff
@@ -0,0 +1,144 @@
...
+            self._flow_unit = 'sccm'
+        else:
+            log.info('Units must be %, sccm, or slm.')
```
BenediktBurger, line 93: Good to log, but the user will not notice, that something is amiss. Would it not be better to raise a ValueError?

pymeasure/instruments/proterial/rod4.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    ch_1 = Instrument.ChannelCreator(ROD4Channel, 1)
+    ch_2 = Instrument.ChannelCreator(ROD4Channel, 2)
+    ch_3 = Instrument.ChannelCreator(ROD4Channel, 3)
+    ch_4 = Instrument.ChannelCreator(ROD4Channel, 4)
+
+    def __init__(self, adapter, name="ROD-4 MFC Controller", **kwargs):
+        super().__init__(
+            adapter, name, read_termination='\r', write_termination='\r', **kwargs
+        )
```
BenediktBurger, line 120: It is mandatory to specify, whether it supports SCPI or not:
```suggestion
        super().__init__(
            adapter, name, read_termination='\r', write_termination='\r',
            includeSCPI=False,
            **kwargs)
```

tests/instruments/proterial/test_rod4.py
BenediktBurger, line 1: Thanks for writing tests.

pymeasure/instruments/proterial/rod4.py
BenediktBurger, line 90: Why a plain property? To read just once?

Is it too much overhead to read it every time?

We try to make the instrument the single point of truth, therefore normally we read it every time from the instrument.
Someone could, for example, change the units on the instrument display, but your code will still interpret it as the old unit.

pymeasure/instruments/proterial/rod4.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @keyboard.setter
+    def keyboard(self, status):
+        values = {'unlocked': 0, 'locked': 1}
+        if isinstance(status, str) and status.casefold() in values:
+            status = status.casefold()
+            self.write("\x0200SKO%d" % values[status])
+            self._keyboard = values[status]
+        else:
+            log.info('Status must be unlocked or locked.')
```
BenediktBurger, line 144: See above for stored state of instrument.

BenediktBurger (changes requested):
Thanks for your contribution.

A few small requests for changes.

pymeasure/instruments/proterial/rod4.py
ConnorGCarr: Good point. Oddly enough, there's no equivalent read command for the flow_unit or keyboard status. This was my workaround, but maybe it's better to just make this an instrument.setting property instead.

ConnorGCarr: Changed this to a "setting" property, removed the need for f-string

ConnorGCarr: Changed to a setting property

pymeasure/instruments/proterial/rod4.py
ConnorGCarr: Changed to a setting property

pymeasure/instruments/proterial/rod4.py
```diff
@@ -109,36 +95,28 @@ class ROD4(Instrument):
...
-    @keyboard.setter
-    def keyboard(self, status):
-        values = {'unlocked': 0, 'locked': 1}
-        if isinstance(status, str) and status.casefold() in values:
-            status = status.casefold()
-            self.write("\x0200SKO%d" % values[status])
-            self._keyboard = values[status]
-        else:
-            log.info('Status must be unlocked or locked.')
+    keyboard = Instrument.setting(
```
BenediktBurger, line 115: Hey,
What about `keyboard_locked`? That would make it more easy to use: `keyboard_locked = True` without thinking about the strings and what True/False mean.

ConnorGCarr: Thanks for the suggestion, I've changed it.

pymeasure/instruments/__init__.py
BenediktBurger, line 1: This file causes the conflict, as we do not import all manufacturers anymore.

If you remove your added line from this file, there should be no conflict anymore.

ConnorGCarr:
> Are you ready, that I merge this driver?

Had to add set command error checking, everything is good to go now.

PR #1048, opened 2024-02-19
fixing set command of SR860 aux output

wehlgrundspitze:
Between the SR830 and the SR860 the numbering of the aux outputs has changed from 1 to 4 towards 0 to 3. This has been accounted for the aux output read commands, but not for the set commands. This leads to a mismatch (tested on a real device), so the aux_out_1 command in facts set the voltage of channel 2. The commit fixes this issue.


BenediktBurger:
Thanks for finding, testing, and fixing this bug.

Issue #1050, opened 2024-02-21
I think the PrologixAdapter isn't getting inherited correctly when initializing instruments that use channels in python 3.12.


Sionwage:
I am trying to create an instrument class for an Agilent E4418B Power Meter with the following code:

    import logging
    from time import sleep
    from time import time as now
    
    import numpy as np
    from pyvisa import VisaIOError
    
    from pymeasure.instruments import Instrument, Channel
    from pymeasure.instruments.validators import strict_discrete_set, truncated_range, strict_range
    
    log = logging.getLogger(__name__)
    log.addHandler(logging.NullHandler())
    
    class PowerMeterChannel(Channel):
        """Channel implimentation for the Agilent E4418B Power Meter"""
    
        frequency = Channel.control('SENS{ch}:FREQ?',
                                    'SENS{ch}:FREQ %e',
                                    """Control the frequency the power meter corrects its
                                    measurement for in Hz. Value range can be changed based
                                    on the power head used.
    
                                    Type: :code:`float`
    
                                    .. code-block:: python
    
                                        # set the frequency to 1.21GHz
                                        instr.frequency = 1.21e9
    
                                        if instr.frequency == 10e6:
                                            pass
    
                                    """,
                                    values=[10e6, 18e9],
                                    validator=strict_range,
                                    dynamic=True,
                                    cast=float)
    
    
    class AgilentE4418B(Instrument):
        """
        Represents the Agilent E4418B Power Meter for measuring RF/Microwave
        power at a given frequency.
    
        This device has one or two channels which would be displaced in one or two of the
        windows on the display.
    
        :param pymeasure.adapter adapter: Adapter used to connect to instrument
        :param str name: Name of the device or generated from the `AgilentE4418B.id` call on `__init__`
        :param includeSCPI: Keyword arguments for the adapter.
        :param kwargs: Keyword arguments for the adapter.
        """
    
        def __init__(
            self,
            adapter=None,
            name="Agilent E4418B Power Meter",
            **kwargs,
        ):
            super().__init__(adapter=adapter, name=name, includeSCPI=True, **kwargs)
    
    
    
        channels = Instrument.MultiChannelCreator(PowerMeterChannel, [1, 2])
    
        def ask(self, command, query_delay=None):
            self.adapter.write(command)
            if query_delay is not None:
                sleep(query_delay)
            return self.adapter.read()
    
        def read(self):
            return self.adapter.read()
    
        def write(self, command):
            self.adapter.write(command)
[...]

Issue #1008, opened 2023-12-20
Python 3.12 seems to work if NI instruments gets commented out


Sionwage:
I would like to report success in using PyMeasure with Python 3.12 when I comment out line 59 in `pymeasure\instruments\__init__.py`

More specifically, the submodule 'NI' within instruments wants to load `DAQmx` which ends up looking for `ctypes.windll.nicaiu` which if not installed raises an AttributeError.

Running the following:

```python
from pymeasure.instruments import Instrument
```

Full error message
```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
File ~\.pyenv\pyenv-win\versions\3.12.0\Lib\ctypes\__init__.py:450, in LibraryLoader.__getattr__(self, name)
    449 try:
--> 450     dll = self._dlltype(name)
    451 except OSError:

File ~\.pyenv\pyenv-win\versions\3.12.0\Lib\ctypes\__init__.py:379, in CDLL.__init__(self, name, mode, handle, use_errno, use_last_error, winmode)
    378 if handle is None:
--> 379     self._handle = _dlopen(self._name, mode)
    380 else:

FileNotFoundError: Could not find module 'nicaiu' (or one of its dependencies). Try using the full path with constructor syntax.

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
Cell In[1], line 6
      3 from io import BytesIO
      5 import numpy as np
----> 6 from pymeasure.instruments import Instrument, RangeException, discreteTruncate
      7 from pyvisa import VisaIOError
      9 from pymeasure.adapters import PrologixAdapter

File ~\AppData\Local\pypoetry\Cache\virtualenvs\py3-12-pymeasure-env-DpKIcySs-py3.12\Lib\site-packages\pymeasure\instruments\__init__.py:59
     57 from . import mksinst
     58 from . import newport
---> 59 from . import ni
     60 from . import novanta
     61 from . import oxfordinstruments

File ~\AppData\Local\pypoetry\Cache\virtualenvs\py3-12-pymeasure-env-DpKIcySs-py3.12\Lib\site-packages\pymeasure\instruments\ni\__init__.py:26
      1 #
      2 # This file is part of the PyMeasure package.
      3 #
   (...)
     22 # THE SOFTWARE.
     23 #
     25 try:
---> 26     from .daqmx import DAQmx
     27 except OSError:
     28     # Error Logging is handled within package
     29     pass

[...]

Issue #1049, opened 2024-02-19
Allow small drop in code coverage


BenediktBurger:
CI tests fail, if the code coverage is reduced, even only a little bit, maybe due to a deleted file.

We could allow a drop of half a percent (for example) without failing the CI status:

https://docs.codecov.com/docs/commit-status#target

CasperSchippers:
I did not realize before that a drop in coverage would result in a failed test.
Do we really want to _enforce_ the coverage or do we just want to _encourage_ keeping up the coverage?

I would be inclined to suggest to only include the coverage test as an encouragement, so no failing on reduced coverage, to keep contributing accessible.

(I have to admit that I also do very often not add tests for all code in my PRs since I do not think that testing every function or line of code is really necessary, and, especially for the GUI part, the testing is more difficult than the contribution itself.)

BenediktBurger:
> I did not realize before that a drop in coverage would result in a failed test.

There is an additional "check" marked as failed, if the requirements are not met.

> I would be inclined to suggest to only include the coverage test as an encouragement, so no failing on reduced coverage, to keep contributing accessible.

Until now, I ignored that failed test in several pull requests as we do not enforce it.

If we ignore that, we can disable that test altogether.

The pull request comment already encourages tests, so that should be sufficient.

BenediktBurger:
I changed the PR to do no status check, so coverage does not affect the CI pipeline.

PR #1051, opened 2024-02-21
Add Python 3.12 to pymeasure

BenediktBurger:
Make pymeasure fit for python 3.12.

This PR updates first the CI to start checking for problems.

Issues, which arose:
- The environment has to be updated, as several packages are not compatible with Python 3.12
   - Pandas has to be larger than 1.5.3 (which is the last 1.* release) to be compatible with Python 3.12, but pandas>2.1 requires Python 3.9
- #1008 
- #1050

### DIFF
--- a/.github/workflows/pymeasure_CI.yml
+++ b/.github/workflows/pymeasure_CI.yml
@@ -95,7 +95,7 @@ jobs:
       fail-fast: true
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -103,7 +103,8 @@ jobs:
       - name: Install pymeasure requirements
         uses: mamba-org/setup-micromamba@v1
         with:
-          environment-file: .github/pymeasure.yml
+          # ternary operator: if python-version == 3.12 use pymeasure_python312.yml else use pymeasure.yml
+          environment-file: ${{ matrix.python-version == '3.12' && '.github/pymeasure_python312.yml' || '.github/pymeasure.yml'}}
           create-args: python=${{ matrix.python-version }}
           cache-environment-key: py${{ matrix.python-version }}-${{ matrix.os }}-mamba-${{ env.CACHE_NUMBER }}-${{ hashFiles('environment.yml') }}
           cache-downloads: false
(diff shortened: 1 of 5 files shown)


BenediktBurger:
I had to remove Python 3.8 from CI, as there is no pandas version compatible with Python 3.8 and Python 3.12 at the same time.

Now CI works and confirms, that `nicaiu` makes problems (see #1008 ). Additionally flake8 goes rogue and complains about missing whitespaces after a colon **in a string**! 🤦

BenediktBurger:
I try also to allow a  small drop of coverage (here 0.5%), see #1049

Sionwage:
Would it be acceptable to add an exception for the NI driver attribute error "nicaiu" so pytest will run while waiting for NI to fix their driver to work with 3.12?

BenediktBurger:
> Would it be acceptable to add an exception for the NI driver attribute error "nicaiu" so pytest will run while waiting for NI to fix their driver to work with 3.12?

Yes, I think that would be a quick fix until we resolve the issue with the NI drivers.

BenediktBurger:
> Would it be acceptable to add an exception for the NI driver attribute error "nicaiu" so pytest will run while waiting for NI to fix their driver to work with 3.12?

I did this in the last commit.

BenediktBurger:
Codecov upload fails due to rate limit, but that's not important, as there are almost no changes in the code proper.

.github/workflows/pymeasure_CI.yml
```diff
@@ -154,3 +154,43 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install pymeasure requirements
+        uses: mamba-org/setup-micromamba@v1
+        with:
+          environment-file: .github/pymeasure_python312.yml
+          create-args: python=${{ matrix.python-version }}
+          cache-environment-key: py${{ matrix.python-version }}-${{ matrix.os }}-mamba-${{ env.CACHE_NUMBER }}-${{ hashFiles('environment.yml') }}
+          cache-downloads: false
```
CasperSchippers, line 179: This is the only difference right, grabbing a different environmenet file?

As this is only a temporary solution, I'm fine with this duplicate job. But maybe if we want to reduce duplicates, it might also be possible to combine it with the main test job using some conditions (e.g., `if: ${{ matrix.python-version }} == "3.12"` and `if: ${{ matrix.python-version }} != "3.12"`).

CasperSchippers (approved):
I think it is good to indeed add support for 3.12, thanks for this!

I have one remark, regarding the tests.

.github/workflows/pymeasure_CI.yml
BenediktBurger: Good point, I'll try it.

BenediktBurger: It works!

CasperSchippers:
Nice, that's even neater than what I thought possible. Good job!

PR #1052, opened 2024-02-21
Add codecov.yml with small coverage drop allowed.

### LINKED ISSUE #1049: Allow small drop in code coverage

BenediktBurger:
This codecov.yml allows the coverage to drop by a quarter percent:
Sometimes the coverage fluctuates a little bit from test to test (around 0.1 percent), or removing lines reduces coverage, which makes that CI test fail. (see the last pull request #919 and current master)

closes #1049 

We could increase the threshold to half a percent or even one percent.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        # basic
+        target: auto
+        threshold: 0.25%  # allows a small drop of coverage for flaky tests
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,4 @@
 coverage:
   status:
-    project:
-      default:
-        # basic
-        target: auto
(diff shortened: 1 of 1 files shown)


BenediktBurger:
I modified the codecov file such, that there are no additional checks (neither for the changed lines nor for the whole project).

BenediktBurger:
Can we merge this PR? @CasperSchippers 

Currently, a drop in overall coverage makes checks fail. This PR (according to discussion in #1049 , but differently from the title) removes coverage CI checks while keeping the coverage message (like above).

CasperSchippers:
Yes, sounds good to me.

PR #1053, opened 2024-02-26
Add Ophir powermeters

BenediktBurger:
Adds Ophir powermeters.

Several share the same commands, therefore I have a large base class and mixin classes for device groups.

### DIFF
--- a/docs/api/instruments/index.rst
+++ b/docs/api/instruments/index.rst
@@ -57,6 +57,7 @@ Instruments by manufacturer:
    newport/index
    ni/index
    novanta/index
+   ophir/index
    oxfordinstruments/index
    parker/index
    philips/index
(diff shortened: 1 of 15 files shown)


BenediktBurger:
@coderabbitai review

bilderbuchi:
@BenediktBurger wow, these reviews seem quite detailed! :-o Are they useful (read: correct), too, in your experience?

BenediktBurger:
@bilderbuchi , yes they are very useful as most comments are things I should mark as well. Sometimes it even cites our documentation (you should add docstrings, copyright notices, or other coding guidelines).
In two cases I had to reject them, (adding docstrings to the __init__...), but it seems to store these information.

Because of its usefulness and quickness (contributors get feedback after a few minutes and not days), I enabled it automatically for all PRs, reducing the burden of reviewing as code style, syntax, code problems are already commented and I can look more to some other aspects of coding.

PR #1054, opened 2024-02-28
Add Keithley2281S

PfannenHans:
Add support for Keithley2281S-20-6 battery simulator and analyzer.


pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,325 @@
...
+    Common commands, beside `function_mode` and power supply commands should also work for
+    Keithley 2280S power supplies, although this is untested.
+    """
+    _VOLTAGE_RANGE = [0.0, 20]
+    _CURRENT_RANGE_PS = [0.1, 6.1]
+    _CURRENT_RANGE_BT_BS = [0.0, 6.1]
+    _INTERNAL_MEMORY_SLOTS = [*range(1, 10)]
+
+    def __init__(self, adapter, name="Keithley2281S", **kwargs):
+        super().__init__(adapter, name, **kwargs)
```
BenediktBurger, line 43: You need to explicitly state the SCPI functionality, for example:
```suggestion
    def __init__(self, adapter, name="Keithley2281S", **kwargs):
        super().__init__(adapter, name, includeSCPI=True, **kwargs)
```

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,325 @@
...
+
+    # Common commands
+
+    cm_display_text_data = Instrument.setting(
+        ":DISP:USER:TEXT '%s'",
+        """Control text to be displayed(24 characters)."""
```
BenediktBurger, line 49: A property's docstring should start with "Set" for a `setting`, "Control" for a `control` and "Get" or "Measure" for a `measurement`.

BenediktBurger (commented):
Thank you for your contribution, it is always great to see new contributors.

A few quick comments to the code and some general comments:
- You may add yourself to the `authors.txt`, if you wish so.
- Please add proper documentation files and add it to the manufacturer's init, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html
- I encourage you to add protocol tests in order to test your functionality, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,345 @@
...
+    Common commands beside `function_mode` and power supply commands should also work for
+    Keithley 2280S power supplies, although this is untested.
+    """
+    _VOLTAGE_RANGE = [0.0, 20]
+    _CURRENT_RANGE_PS = [0.1, 6.1]
+    _CURRENT_RANGE_BT_BS = [0.0, 6.1]
+    _INTERNAL_MEMORY_SLOTS = [*range(1, 10)]
+
+    def __init__(self, adapter, name="Keithley2281S", **kwargs):
+        super().__init__(adapter, name, includeSCPI=True, **kwargs)
```
BenediktBurger, line 67: I'm sorry for the back and forth:
Recently I told you to specify `includeSCPI` explicitly, which you did, thank you.
That was, however, only in preparation for a larger change (#905).
Now, that other PR has been merged, so the new style to define SCPI compliant instruments works differently:

```suggestion
from pymeasure.instruments import SCPIMixin

class Keithley2281S(SCPIMixin, Instrument, KeithleyBuffer):
    """
    Represents the Keithley 2281S-20-6 power supply and battery simulator / characterizer.
    Common commands beside `function_mode` and power supply commands should also work for
    Keithley 2280S power supplies, although this is untested.
    """
    _VOLTAGE_RANGE = [0.0, 20]
    _CURRENT_RANGE_PS = [0.1, 6.1]
    _CURRENT_RANGE_BT_BS = [0.0, 6.1]
    _INTERNAL_MEMORY_SLOTS = [*range(1, 10)]

    def __init__(self, adapter, name="Keithley2281S", **kwargs):
        super().__init__(adapter, name, **kwargs)
```
Please move the import together with the import of `Instrument`.

This mixin-based implementation is a first step to create more harmonized instruments.

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,347 @@
...
+    )
+
+    # Needs two arguments
+    # bs_load_model_usb = Instrument.setting(
+    #    ":BATT:MOD:LOAD:USB %d, \"%s\"",
+    #    """Loads a battery model from USB to the internal memory.""",
+    #    validator=strict_discrete_set,
+    #    values=_INTERNAL_MEMORY_SLOTS
+    #    cast=int
+    # )
```
PfannenHans, line 266: @BenediktBurger thanks for the review so far. Maybe you can give me a hint on this one.  
I would need two arguments here: The internal memory slot where to store the battery profile and the name of the battery-profile on an USB-Stick.  
Is it possible to have two arguments using pymeasure? Otherwise i would hardcode the internal memory slot to be used.

BenediktBurger: In this case, a method (not a property) is better suited.

PfannenHans:
Unfortunately I could not find a solution for the warnings ins the Documentation generation step.

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,537 @@
...
+        ":BATT:OUTP %s",
+        """Control the output state.""",
+        validator=strict_discrete_set,
+        values={True: "ON", False: "OFF"},
+        map_values=True,
+    )
+
+    bt_test_control = Instrument.setting(
+        ":BATT:TEST:SENS:AH:EXEC %s",
+        """Control the output state.""",
```
BenediktBurger, line 358: It is a `setting`, so please use "Set":
```suggestion
        """Set the output state.""",
```

I do not understand, from the docstring, what do you set and what options do you have. I suggest to put the valid values into the docstring.

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,537 @@
...
+        # Replace with match case, once Python>=3.10 is required
+        function_mode = self.cm_function_mode
+        if function_mode == "POWER":
+            return self.ps_buffer_data
+        elif function_mode == "TEST":
+            return self.bt_buffer_data
+        elif function_mode == "SIMULATOR":
+            return self.bs_buffer_data
+
+    # Common commands (cm_*), these can be used in any function mode
```
BenediktBurger, line 114: Did you consider using channels for the power supply, battery testing, and battery simulation modes?

That would make it easier to understand, I think, as you could drop all the prefixes.

BenediktBurger (changes requested):
Thanks for your contribution. I added a few comments.

BenediktBurger:
> Unfortunately I could not find a solution for the warnings ins the Documentation generation step.

I do not understand them either.
As both enums are broken, you could try a few things: Maybe the comments are the problem or the underscores. Another idea is to check their docstrings.

Finally, if you make the enum values descriptive, you do not have to add them to the `.rst` file, which will make the error disappear (I guess).

mcdo0486:
> Unfortunately I could not find a solution for the warnings ins the Documentation generation step.

This is caused by generating documentation for inherited methods for the `enum.IntFlag` classes (`Keithley2281SOperationEventRegister` & `Keithley2281SMeasurementEventRegister`) in [docs/api/instruments/keithley/keithley2281S.rst](https://github.com/mcdo0486/pymeasure/commit/a216a77679507af084f7ea0b15aa92a8cfd8e8e6#diff-8a6be87ebce428c4c505546a13ffa1b099298e667a0dabf9e00d8a0a768030fd)

I made a small PR for your branch if you'd like to merge it, which removes the inheritance.

The inherited docstrings in `int` do not follow some linting rules. In fact, I think you can just remove the `Keithley2281SOperationEventRegister` & `Keithley2281SMeasurementEventRegister` classes from generated documentation, as the user wouldn't be using those classes directly.

BenediktBurger:
Ah, that makes sense. Thanks @mcdo0486 .

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,537 @@
...
+    _RESERVED_1 = 2
+    _RESERVED_2 = 4
+    _RESERVED_3 = 8
+    _RESERVED_4 = 16
+    _RESERVED_5 = 32
+    RAV = 64  # Reading available
+    BHF = 128  # Buffer half-full
+    BH = 256  # Buffer full
+    BTF = 512  # Buffer three-quarters full
+    BQF = 1024  # Buffer one-quarter-full
```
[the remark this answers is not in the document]
PfannenHans: I just took the register names from datasheet, but yes just writing it out is better 👍

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,537 @@
...
+        """Get line frequency.""",
+        cast=int,
+    )
+
+    @property
+    def cm_measurement_ongoing(self) -> bool:
+        """Get measurement status."""
+        if Keithley2281SOperationEventRegister.MEASUREMENT in self.cm_summary_event:
+            return True
+        return False
```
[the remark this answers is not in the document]
PfannenHans: Thanks, classic comparison to boolean 😅

pymeasure/instruments/keithley/keithley2281S.py
PfannenHans: Good idea, i will look into channels. They must be mutually exclusive thou.

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,537 @@
...
+
+    @property
+    def cm_reading_available(self) -> bool:
+        """Get availability of a reading."""
+        if Keithley2281SMeasurementEventRegister.RAV in self.cm_measurement_event:
+            return True
+        return False
```
PfannenHans, line 167:
```suggestion
        return Keithley2281SMeasurementEventRegister.RAV in self.cm_measurement_event

```

pymeasure/instruments/keithley/keithley2281S.py
PfannenHans: Indeed, the Docstring is not enough here. There are two ways to enable the output in test-mode.
One is for (dis-)charging to a certain voltage (`bt_output_enabled`), the other is for actually recording the parameters for a battery-model.  
This is the property that triggers the output for characterization

pymeasure/instruments/keithley/keithley2281S.py
BenediktBurger: You could add/remove the channels dynamically.
However, I think it is best to keep them active, as the driver never know, which mode the device is in.
Right now, you do not disable these "forbidden" methods either.

BenediktBurger: By the way: I thought about three different channels: One simulation, one measurement and one power supply.

BenediktBurger:
Your changes till now look good, only the question of channels remains open.

Please refrain from force-pushing commits as that prevents github from showing previous comments on the right spot (as the respective commit is not present anymore).

pymeasure/instruments/keithley/keithley2281S.py
PfannenHans: Finally came around to implement the channels. Unfortunately I had to put some constants in the global scope of the file, since I could not call `self.parent.var` from the channel's properties for some reason.   
The `_PLC_RANGE` 'constant' might be changed once on startup, if a 60Hz line-frequency is detected. Maybe the global scope might be a problem in the edge case where one uses two of these devices in the same script and one would be on 50Hz and the other on 60Hz. But I guess this is a very exotic use-case.

BenediktBurger: Don't worry about global constants. That's totally fine as the user will never see them.

PfannenHans: Allright. I will do some last tests on my device at work next week and see if it works after the addition of channels and then this can be merged I think.

PfannenHans: Sorry for the long delay. I could finally test all the functions with a device and fix or remove what did not work.
The code is now in a state where it can be shared with others. The latest master branch is merged and all the checks are run as well.

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,573 @@
...
+
+    Make sure that the parent's instrument `function_mode` property is set to `POWER` before using
+    this channel.
+    """
+
+    _query_buffer_data = Instrument.measurement(
+        ':DATA:DATA? "READ,SOUR,REL"',
+        """Get the buffer in power supply mode""",
+        separator=",",
+    )
```
BenediktBurger, line 103: I'd change the name of the property. Right now it sounds like a method. Maybe something like `_data_buffer` is better.

Small note: Add a period at the end of the docstring.

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,573 @@
...
+    output_enabled = Instrument.control(
+        ":OUTP:STAT?",
+        ":OUTP:STAT %s",
+        """Control the output state.""",
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    conc_nplc = Instrument.control(
```
BenediktBurger, line 146: What about `power_line_cycles` as a name? Wouldn't that be better?

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,573 @@
...
+        map_values=True,
+    )
+
+    simulation_mode = Instrument.control(
+        ":BATT:SIM:METH?",
+        ":BATT:SIM:METH %s",
+        """
+        Control simulation mode to use.
+        I.e. does the SoC change when charging or discharging.
+        """,
```
BenediktBurger, line 323: It is good to add the possible values to the docstring.
Do I understand it correctly, that there are only two possible modes?
What do you think about a boolean property `dynamic_simulation_enabled`? That makes it easier to use and to interpret as it is just a boolean and not strings being sent to the device/interpreted.

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,573 @@
...
+        ESR might not be displayed correctly until display is updated (e.g. via enabled output)
+        """,
+        validator=truncated_range,
+        values=[-100, 100],
+    )
+
+    soc_setpoint = Instrument.control(
+        ":BATT:SIM:SOC?",
+        ":BATT:SIM:SOC %g",
+        """Control the SoC of the simulated battery.""",
```
BenediktBurger, line 368: Consider spelling "SoC" out to make it more accessible.

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,573 @@
...
+        validator=truncated_range,
+        values=[0.0, 100],
+    )
+
+    voc_setpoint = Instrument.control(
+        ":BATT:SIM:VOC?",
+        ":BATT:SIM:VOC %g",
+        """Control the Voc of the simulated battery.""",
```
BenediktBurger, line 376: What is "Voc"? Would a user of the device understand it?

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,573 @@
...
+            return self.bs.buffer_data
+
+    display_text_data = Instrument.setting(
+        ":DISP:USER:TEXT '%s'", """Set control text to be displayed(24 characters)."""
+    )
+
+    function_mode = Instrument.control(
+        ":ENTR:FUNC?",
+        ":ENTR:FUNC %s",
+        """Control function mode to use.""",
```
BenediktBurger, line 459: Consider adding the corresponding values to the docstring.

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,573 @@
...
+        if function_mode == "TEST":
+            return self.bt.buffer_data
+        if function_mode == "SIMULATOR":
+            return self.bs.buffer_data
+
+    display_text_data = Instrument.setting(
+        ":DISP:USER:TEXT '%s'", """Set control text to be displayed(24 characters)."""
+    )
+
+    function_mode = Instrument.control(
```
BenediktBurger, line 456: What about a method for each channel, which enables that channel. E.g. `enable_channel`, such that you start using a channel with the channel itself?

pymeasure/instruments/keithley/keithley2281S.py
```diff
@@ -0,0 +1,573 @@
...
+        self.function_mode = "TEST"
+
+        # Set discharging conditions and start discharging
+        self.bt.charge_voltage_setpoint = lower_voltage
+        self.bt.charge_current_limit = charge_current / 10
+        self.bt.termination_current = charge_current / 100
+        self.bt.output_enabled = True
+        self.display_text_data = "Measurement Ongoing! Discharging..."
+        while self.measurement_ongoing:
+            time.sleep(1)
```
BenediktBurger, line 553: I think it is good to mention in the docstring, that this method blocks until the whole measurement has finished.

BenediktBurger (changes requested):
A few more requests regarding naming and documentation.

Thank you for your work and patience.

pymeasure/instruments/keithley/keithley2281S.py
PfannenHans: In this case i think it is clearer to have the `function_mode` in the parent device, since the channels are not actual channels, but groups of functionalities that are mutually exclusive. I added an explanation to the docstring of each channel.  
For the other comments i implemented the review suggestions.

BenediktBurger: Ok.

PR #1069, opened 2024-03-19
fix Stanford Research SR830 output conversion

dkriegner:
This fixes the `output_conversion` function of the Stanford Research SR830 lock-in amplifier. This seems to be a very useful function but in the current implementation did not work for me as intended (I do not understand how it actually ever worked like this). 

Rationale for the change:
The [manual](https://www.thinksrs.com/downloads/pdfs/manuals/SR830m.pdf) says (on page 1-7 / 9) that the `OEXP` command used by `get_scaling` returns Offset in _x_ percent ( -105.00 ≤ _x_ ≤ 105.00) and Expand as integer 0,1,2. The integer values of expand are converted to the values of 1, 10, and 100 in `get_scaling`. The offest is returned unchanged. Only the new equation seems to correspond to this logic. 

Edit: Testing with the hardware revealed that the value return as the `SR830.x` property does not change when the expand value is changed. Therefore the `expand` should not appear in the output conversion equation.

### DIFF
--- a/pymeasure/instruments/srs/sr830.py
+++ b/pymeasure/instruments/srs/sr830.py
@@ -414,9 +414,9 @@ class SR830(Instrument):
         """ Returns a function that can be used to determine
         the signal from the channel output (X, Y, or R)
         """
-        offset, expand = self.get_scaling(channel)
+        offset, _ = self.get_scaling(channel)
         sensitivity = self.sensitivity
-        return lambda x: x / expand + offset / 100. * sensitivity
+        return lambda x: x  + offset / 100 * sensitivity
 
     @property
     def sample_frequency(self):
--- a/pymeasure/instruments/srs/sr830.py
+++ b/pymeasure/instruments/srs/sr830.py
@@ -416,7 +416,7 @@ class SR830(Instrument):
         """
         offset, _ = self.get_scaling(channel)
         sensitivity = self.sensitivity
-        return lambda x: x  + offset / 100 * sensitivity
+        return lambda x: x + offset / 100 * sensitivity
 
     @property
     def sample_frequency(self):
--- a/pymeasure/instruments/srs/sr830.py
+++ b/pymeasure/instruments/srs/sr830.py
@@ -416,7 +416,7 @@ class SR830(Instrument):
         """
         offset, expand = self.get_scaling(channel)
         sensitivity = self.sensitivity
-        return lambda x: (x / (10. * expand) + offset) * sensitivity
+        return lambda x: x / expand + offset / 100. * sensitivity
 
     @property
     def sample_frequency(self):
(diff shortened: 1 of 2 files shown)


pymeasure/instruments/srs/sr830.py
```diff
@@ -416,7 +416,7 @@ def output_conversion(self, channel):
         """
         offset, expand = self.get_scaling(channel)
         sensitivity = self.sensitivity
-        return lambda x: (x / (10. * expand) + offset) * sensitivity
+        return lambda x: x / expand + offset / 100. * sensitivity
```
BenediktBurger, line 419: You do not need to divide by the float 100. The results (for python 3) are the same with `/ 100`.

BenediktBurger (approved):
Thanks for your fix.
Probably there were some misunderstandings of the manual.

Maybe you could write a (protocol) test, to ensure, that it works as expected, which is quite useful for such calculations.

pymeasure/instruments/srs/sr830.py
BenediktBurger: This is just a note for yourself.

BenediktBurger:
Hey @dkriegner , would you mind writing a test or do prefer merging without a test?

dkriegner:
I will write some test, sorry for the delay. I try to get to it in
reasonable time.

BenediktBurger:
I thought it would be good to include it in the next release we are going to prepare. Therefore, I am asking.

pymeasure/instruments/srs/sr830.py
dkriegner: I now removed the ".". While I was aware of the python 3 feature i wanted to be explicit here.

pymeasure/instruments/srs/sr830.py
```diff
@@ -414,9 +414,9 @@ def output_conversion(self, channel):
         """ Returns a function that can be used to determine
         the signal from the channel output (X, Y, or R)
         """
-        offset, expand = self.get_scaling(channel)
+        offset, _ = self.get_scaling(channel)
         sensitivity = self.sensitivity
-        return lambda x: (x / (10. * expand) + offset) * sensitivity
+        return lambda x: x  + offset / 100 * sensitivity
```
dkriegner, line 419: thanks for pushing for writing the tests. it forced me to test this with hardware and I found that even my changed version was not correct. It turns out that the expand value should not be used at all in the conversion function. The latest version now behaves as expected with my hardware.

dkriegner:
This is now ready to merge from my side. Sorry for the initially wrong change. The tests cover of course only a fraction of the features of the SR830, but they should cover the current change.

pymeasure/instruments/srs/sr830.py
BenediktBurger: Thanks for writing the tests.
I'm glad to hear that the tests improved code quality yet another time.

I don't know how many bugs I discovered by having tests.

BenediktBurger:
Thanks for your work!

PR #1019, opened 2024-01-04
Scpi instrument codebase

BenediktBurger:
Continuation of #905 .

In this PR all instruments have the SCPIMixin or SCPIUnknownMixin, unless they do not comply explicitly to SCPI.

- [x] Update the .rst files to include the methods from the Mixin, see https://github.com/pymeasure/pymeasure/pull/905#issuecomment-1546668335 -> moved to its own issue #1060 
- [x] Rework all recently added instruments in order to comply. (maybe with an all_instruments test against `includeSCPI=True`?)
- [x] Adding documentation #1047

PR #1077, opened 2024-03-27
Make query_delay=None the default.

BenediktBurger:
Reworking the codebase (#1019 ), I noticed the warning, that ATSBase still uses `query_delay`, which is deprecated (and does not work anymore, infact).

This PR fixes it.

While writing the fix, I noticed an inconvenience for instruments with a default delay (Anc300, for example):
`wait_for` cannot distinguish between an explicit `0` by the user and the _default_ `0` by `ask`. In both cases, they would wait for `self.query_delay` seconds

Therefore, I made the default `None`, such that you can distinguish between not set (in this case it will be `self.query_delay`) and between an explicit 0.

I adjusted the ANC300 as well and some other places where it was needed.

### DIFF
--- a/pymeasure/instruments/attocube/anc300.py
+++ b/pymeasure/instruments/attocube/anc300.py
@@ -138,7 +138,7 @@ class Axis(Channel):
     stepu = Instrument.setting(
         "stepu %d",
         """Set the steps upwards for N steps. Mode must be 'stp' and N must be
-        positive. 0 causes a continous movement until stop is called.
+        positive. 0 causes a continuous movement until stop is called.
 
         .. deprecated:: 0.13.0 Use meth:`move_raw` instead.
         """,
@@ -150,7 +150,7 @@ class Axis(Channel):
     stepd = Instrument.setting(
         "stepd %d",
         """Set the steps downwards for N steps. Mode must be 'stp' and N must be
-        positive. 0 causes a continous movement until stop is called.
+        positive. 0 causes a continuous movement until stop is called.
 
         .. deprecated:: 0.13.0 Use meth:`move_raw` instead.
         """,
@@ -217,7 +217,7 @@ class Axis(Channel):
         # perform the movement
         self.move_raw(steps)
         # wait for the move to finish
-        self.parent.wait_for(abs(steps) / self.frequency)
+        self.wait_for(abs(steps) / self.frequency)
         # ask if movement finished
         self.ask('stepw')
         if gnd:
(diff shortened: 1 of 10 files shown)


pymeasure/instruments/temptronic/temptronic_base.py
```diff
@@ -125,6 +125,12 @@ class ATSBase(SCPIUnknownMixin, Instrument):
     """The base class for Temptronic ATSXXX instruments.
     """
 
+    def __init__(self, adapter, name="ATSBase", **kwargs):
+        super().__init__(adapter, name=name, **kwargs)
+
+    def wait_for(self, query_delay=None):
+        super().wait_for(0.05 if query_delay is None else query_delay)
```
CasperSchippers, line 132: In this case, would it not make more sense to put the 0.05 as function-argument default? So:
``` suggestion 
    def wait_for(self, query_delay=0.05):
        super().wait_for(query_delay)
```

pymeasure/instruments/attocube/anc300.py
```diff
@@ -408,9 +408,9 @@ def read(self):
                              f"command with message {msg}")
         return self._extract_value(msg)
 
-    def wait_for(self, query_delay=0):
+    def wait_for(self, query_delay=None):
         """Wait for some time. Used by 'ask' to wait before reading.
 
-        :param query_delay: Delay between writing and reading in seconds.
+        :param query_delay: Delay between writing and reading in seconds. None is default delay.
```
CasperSchippers, line 414: Maybe indicate what happens if no delay is provided, i.e., that the `self.query_delay` value will be used?

CasperSchippers (approved):
Sounds good to me. Just two minor comments.

pymeasure/instruments/temptronic/temptronic_base.py
BenediktBurger: The `values` method calls wait_for with None.
Therefore, the default argument 0.5 would not be used.

pymeasure/instruments/attocube/anc300.py
BenediktBurger: I expanded th comments.

BenediktBurger:
@CasperSchippers  you may merge, if you deem the changes suitable.

pymeasure/instruments/temptronic/temptronic_base.py
CasperSchippers: Ahh check, good point.

PR #1086, opened 2024-04-10
Add Kepco BOP 36-12 Bipolar Power Supply

JAW90:
I have written a minimal driver for Kepco BOP power supplies with BIT 4886 digital interface cards. I have tested with a BOP 36-12 over GPIB and all is working as expected. It is possible to carry out more sophisticated control sequences etc but I thought it was best to keep things simple for the time being. In principle it is also possible to communicate over RS-232 but I haven't checked this for myself yet (I don't have an intention to use it this way).

This is my first time contributing to an open source project, so a bit new to the whole process! I followed the guidance in the documentation and think I have done most of the requirements, though I haven't included any tests just yet. I tried to follow the Keithley 2260B as an example in terms of naming conventions etc.

### DIFF
--- a/pymeasure/instruments/kepco/kepcobop.py
+++ b/pymeasure/instruments/kepco/kepcobop.py
@@ -83,7 +83,7 @@ class KepcoBOP3612(SCPIMixin, Instrument):
         map_values=True,
     )
 
-    # TODO 
+    # TODO
     # Return list of errors based on return number.
 
     def beep(self):
@@ -161,7 +161,7 @@ class KepcoBOP3612(SCPIMixin, Instrument):
         (see: `current`).
         If power supply in voltage mode, this sets the compliance current
         for the corresponding voltage set point.
-        Query returns programmed value, meaning of which is dependent on 
+        Query returns programmed value, meaning of which is dependent on
         power supply operating context (see: `operating_mode`).
 
         Output must be enabled separately (see: `output_enabled`)
@@ -181,7 +181,7 @@ class KepcoBOP3612(SCPIMixin, Instrument):
         (see: `voltage`).
         If power supply in current mode, this sets the compliance voltage
         for the corresponding current set point.
-        Query returns programmed value, meaning of which is dependent on 
+        Query returns programmed value, meaning of which is dependent on
         power supply operating context (see: `operating_mode`).
 
         Output must be enabled separately (see: `output_enabled`)
--- a/pymeasure/instruments/kepco/kepcobop.py
+++ b/pymeasure/instruments/kepco/kepcobop.py
@@ -79,9 +79,8 @@ class KepcoBOP3612(SCPIMixin, Instrument):
     def __init__(self, adapter, name="Kepco BOP 36-12 Bipolar Power Supply",
                  **kwargs):
         super().__init__(
-            adapter,
-            name,
-            baud_rate=9600,
+            adapter=adapter,
+            name=name,
             read_termination="\n",
             write_termination="\n",
             **kwargs
--- a/pymeasure/instruments/kepco/kepcobop.py
+++ b/pymeasure/instruments/kepco/kepcobop.py
@@ -62,10 +62,12 @@ class KepcoBOP3612(SCPIMixin, Instrument):
     _Imax = 12
 
     def __init__(self, adapter, name="Kepco BOP 36-12 Bipolar Power Supply",
-                 read_termination="\n", write_termination="\n", **kwargs):
+                 baud_rate=9600, read_termination="\n", write_termination="\n",
+                 **kwargs):
         super().__init__(
             adapter,
             name,
+            baud_rate=baud_rate,
             read_termination=read_termination,
             write_termination=write_termination,
             **kwargs
@@ -114,7 +116,7 @@ class KepcoBOP3612(SCPIMixin, Instrument):
         commands or queries. """
         self.write("*WAI")
 
-    voltage_measure = Instrument.measurement(
+    voltage = Instrument.measurement(
         "MEASure:VOLTage?",
         """
         Measures actual voltage that is across the output terminals.
@@ -122,7 +124,7 @@ class KepcoBOP3612(SCPIMixin, Instrument):
         cast=float
     )
 
-    current_measure = Instrument.measurement(
+    current = Instrument.measurement(
         "MEASure:CURRent?",
         """
         Measures the actual current through the output terminals.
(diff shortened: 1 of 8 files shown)


pymeasure/instruments/kepco/kepcobop.py
```diff
@@ -0,0 +1,180 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+# List of Kepco BOP's (Vmax,Imax) [single channel output]
```
BenediktBurger, line 25: Do I understand it correctly, that it has channels? If so, you might consider using a `Channel` class.

pymeasure/instruments/kepco/kepcobop.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    Represents the Kepco BOP 36-12 (M or D) 400 W bipolar power supply
+    fitted with BIT 4886 digital interface card (minimal implementation)
+    and provides a high-level interface for interacting with the instrument.
+    """
+    _Vmax = 36
+    _Imax = 12
+
+    def __init__(self, adapter, name="Kepco BOP 36-12 Bipolar Power Supply",
+                 baud_rate=9600, read_termination="\n", write_termination="\n",
```
BenediktBurger, line 65: Can you change the baud_rate and the termination of the instrument?
If not, you should hard code them in the super().__init__() call below.

pymeasure/instruments/kepco/kepcobop.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        Control whether the source is enabled, takes values True or False (bool)
+        """,
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    def beep(self):
+        """Causes the unit to emit a brief audible tone."""
```
BenediktBurger, line 88: A general comment: The first sentence of a docstring should be in imperative mode:
```suggestion
        """Cause the unit to emit a brief, audible tone."""
```

pymeasure/instruments/kepco/kepcobop.py
```diff
@@ -0,0 +1,180 @@
...
+        self.write("SYSTem:BEEP")
+
+    confidence_test = Instrument.measurement(
+        "*TST?",
+        """
+        Power supply interface self-test procedure.
+        Returns 0 if all tests passed,
+        otherwise corresponding error code.
```
BenediktBurger, line 96: Pymeasure convention: The docstring of a `measurement` should start with `"Measure"` or `"Get"`, for example: Get the amount of failed self tests.

For `control`, it should start with `"Control"` and for `setting` with `"Set"`.

pymeasure/instruments/kepco/kepcobop.py
```diff
@@ -0,0 +1,180 @@
...
+        """,
+        cast=int
```
BenediktBurger, line 98: You could use "get_process" to return a True/False or even raise an error, if the result is negative.

pymeasure/instruments/kepco/kepcobop.py
```diff
@@ -0,0 +1,180 @@
...
+    def wait_to_continue(self):
+        """ Causes the power supply to wait until all previously issued
+        commands and queries are complete before executing subsequent
+        commands or queries. """
+        self.write("*WAI")
+
+    voltage = Instrument.measurement(
+        "MEASure:VOLTage?",
+        """
+        Measures actual voltage that is across the output terminals.
```
BenediktBurger, line 122: As explained above, it should start with `"Measure"` or `"Get"`:
```suggestion
        Measure current voltage across the output terminals in Volts.
```

The units are good to mention.

BenediktBurger (changes requested):
Thank you for your contribution @JAW90

I left a few comments regarding style. Your code looks already good.

Consider adding tests, as they help us to verify, that the device works as expected, even if the code changes.

pymeasure/instruments/kepco/kepcobop.py
JAW90: Hi Benedikt, thanks for all the comments, they are very helpful.

Yes, there do exist some models which have two output channels. However, it turns out they are not very sophisticated in design and are essentially two independent power supplies in a single unit. They each have their own digital interface card and GPIB address, though they share a single, physical connector. The implementation for each sub-power supply looks to be the same as for the single channel output models (I have had a brief scan of the manual), since they use the same interface card (BIT 4886) as the single channel models.

What do you think is the best route to go down? I could imagine creating a class for the two channel model, e.g. `Kepco2X504` (two channels of 50 V, 4 A) that is simply two `Kepco504` as each channel, though I'm not sure what the advantage would be over just declaring two single channel outputs since they behave that anyway. There doesn't seem to be any functionality of the unit as a whole that affects each channel simultaneously.

pymeasure/instruments/kepco/kepcobop.py
JAW90: No, it looks like it can't be changed. Thanks.

pymeasure/instruments/kepco/kepcobop.py
BenediktBurger: If they have different GPIB addresses (very strange setup), they need two different instruments.
Therefore, you do not need to create a `Channel`.

pymeasure/instruments/kepco/kepcobop.py
JAW90: I found the section in the documentation about error codes and it was very simple to use `IntFlag` to determine the errors based on the 9-bit number.

JAW90:
I think I have covered all the suggested changes. Is there anything that I have missed? Or other changes needed?

BenediktBurger:
Thanks for your contribution.

PR #1087, opened 2024-04-11
Fix 1085

### LINKED ISSUE #1085: expected_protocol uses empty dictionary as default value
`expected_protocol` uses an empty dictionary as a default value, which is bad practice.
The underlying `ProtocolAdapter` already accepts `None`, so a change to `None` is a no brainer.

BenediktBurger:
Fix #1085


CasperSchippers (approved):
Looks fine. I think we do not need to wait for the codecov issue to be fixed, do we?

BenediktBurger:
> Looks fine. I think we do not need to wait for the codecov issue to be fixed, do we?

No, it is just, that no coverage reports are uploaded.

Issue #1006, opened 2023-12-20
hp8116a should have staticmethods


BenediktBurger:
The methods `_generate_1_2_5_sequence` and `_boolean_control` should be static methods.

For that, they have to be used as `HP8116A._boolean_control`.

PR #1088, opened 2024-04-11
Fix 1006 hp8116a

BenediktBurger:
Fixes the hp8116a not using staticmethods (#1006) and improves docstrings in that pr.

### DIFF
--- a/pymeasure/instruments/hp/hp8116a.py
+++ b/pymeasure/instruments/hp/hp8116a.py
@@ -197,6 +197,7 @@ class HP8116A(Instrument):
 
         return value
 
+    @staticmethod
     def _generate_1_2_5_sequence(min, max):
         """ Generate a list of a 1-2-5 sequence between min and max. """
         exp_min = int(np.log10(min))
@@ -209,6 +210,7 @@ class HP8116A(Instrument):
 
         return list(sequence)
 
+    @staticmethod
     def _boolean_control(identifier, state_index, docs, inverted=False, **kwargs):
         return Instrument.control(
             'CST', identifier + '%d', docs,
--- a/pymeasure/instruments/hp/hp8116a.py
+++ b/pymeasure/instruments/hp/hp8116a.py
@@ -533,7 +533,7 @@ class HP8116A(Instrument):
     def shutdown(self):
         """ Gracefully close the connection to the 8116A. """
         self.adapter.connection.clear()
-        self.adapter.connection.close()
+        self.adapter.close()
         super().shutdown()
 
     def check_errors(self):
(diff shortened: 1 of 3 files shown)


BenediktBurger:
Only from [python 3.10](https://docs.python.org/3/whatsnew/3.10.html) onward, static methods can be called directly, therefore these methods have to go outside of the instrument body.

pymeasure/instruments/hp/hp8116a.py
```diff
@@ -35,6 +35,30 @@
...
+
+def _boolean_control(identifier, state_index, docs, inverted=False, **kwargs):
+    return Instrument.control(
+        'CST', identifier + '%d', docs,
+        validator=strict_discrete_set,
+        values=[True, False],
+        get_process=lambda x: inverted ^ bool(int(x[state_index][1])),
+        set_process=lambda x: int(inverted ^ x),
+        **kwargs
+    )
```
CasperSchippers, line 59: Annoying behaviour in python<3.10 with these static-methods. Is this something we'd like to revert when drop support for pyton<=3.9 (so in about 1.5 year I think)? If so, we might want to mark this as something to change back by then.

pymeasure/instruments/hp/hp8116a.py
CasperSchippers, line 1: Nice that you conformize the docstrings. Just noticed a slight inconsistency: for some properties you changed the 'allowed range is X to Y' into '(strict float from X to Y)', while for others you left the range description as it is. Just wondered if this was intentionall.

I'll leave it up to you whether you want to keep it like this or make it consistent.

CasperSchippers (approved):
Looks good to me. Just some minor remarks.

tests/instruments/hp/test_hp8116a.py
```diff
@@ -31,6 +31,9 @@
 HP8116A.status = property(fget=lambda self: Status(5))
 
 
+init_comm = [(b"CST", b"x" * 87 + b' ,\r\n')]  # communication during init
```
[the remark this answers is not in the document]
BenediktBurger: Done.

pymeasure/instruments/hp/hp8116a.py
BenediktBurger: I added a comment.

pymeasure/instruments/hp/hp8116a.py
BenediktBurger: I made them consistent

BenediktBurger:
@CasperSchippers thanks for your reviews.
I addressed your issues. Feel free to merge, if they suit you.

Issue #1090, opened 2024-04-11
Codecov uploads fail


BenediktBurger:
See for example: https://github.com/pymeasure/pymeasure/actions/runs/8646275960/job/23705256912?pr=1086

That is the upstream issue:
- https://github.com/codecov/codecov-action/issues/1292

Log entry:
~~~
[2024-04-11T12:00:18.997Z] ['info'] Pinging Codecov: https://codecov.io/upload/v4?package=github-action-3.1.6-uploader-0.7.2&token=*******&branch=dev%2Fkepco-bop-bit-4886&build=8646275960&build_url=https%3A%2F%2Fgithub.com%2Fpymeasure%2Fpymeasure%2Factions%2Fruns%2F8646275960&commit=e15c0c18bccc682b73428ba47e74ad3724f9402a&job=PyMeasure+CI&pr=1086&service=github-actions&slug=pymeasure%2Fpymeasure&name=codecov-umbrella&tag=&flags=unittests&parent=
[2024-04-11T12:00:18.997Z] ['verbose'] Passed token was 0 characters long
[2024-04-11T12:00:18.997Z] ['verbose'] https://codecov.io/upload/v4?package=github-action-3.1.6-uploader-0.7.2&branch=dev%2Fkepco-bop-bit-4886&build=8646275960&build_url=https%3A%2F%2Fgithub.com%2Fpymeasure%2Fpymeasure%2Factions%2Fruns%2F8646275960&commit=e15c0c18bccc682b73428ba47e74ad3724f9402a&job=PyMeasure+CI&pr=1086&service=github-actions&slug=pymeasure%2Fpymeasure&name=codecov-umbrella&tag=&flags=unittests&parent=
        Content-Type: 'text/plain'
        Content-Encoding: 'gzip'
        X-Reduced-Redundancy: 'false'
[2024-04-11T12:00:19.180Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
[...]

PR #1091, opened 2024-04-11
Bump codecov action version.

BenediktBurger:
Maybe version 4 is a fix for #1090

### DIFF
--- a/.github/workflows/pymeasure_CI.yml
+++ b/.github/workflows/pymeasure_CI.yml
@@ -76,7 +76,7 @@ jobs:
           echo "::add-matcher::.github/pytest.json"
           xvfb-run -a pytest --cov=pymeasure --cov-report=xml
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           directory: ./coverage/reports/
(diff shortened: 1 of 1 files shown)


driftregion:
Here is a purportedly successful bump in codecov version from 3 to 4:
https://github.com/Open-Attestation/document-store/pull/170/files

(linked from https://github.com/codecov/codecov-action/issues/1292)

BenediktBurger:
> Here is a purportedly successful bump in codecov version from 3 to 4: https://github.com/Open-Attestation/document-store/pull/170/files
> 
> (linked from [codecov/codecov-action#1292](https://github.com/codecov/codecov-action/issues/1292))

Thanks for making me aware of that. Unfortunately, they have another issue with a workflow calling another one, but both are in the main repo.

Our problem is, that work flows in forks cannot access the secret token.

driftregion:
It seems the established solution is to use `pull_request_target`. Example: https://github.com/hyperledger/firefly-tezosconnect/pull/20

See also:
https://github.com/codecov/codecov-action/issues/29 

This solution seems to be unviable if there is potential for abuse of the codecov token or other secrets. I haven't yet found anything on whether or not it's possible to isolate secrets to a particular workflow.

BenediktBurger:
As this PR is from a fork, the upload will fail, such that it cannot be tested.

PR #1095, opened 2024-04-29
Use permanent Zenodo DOI in readme.rst

### LINKED ISSUE #1070: Update Zenodo DOI
The Zenodo-DOI is still of version 0.8 for "ralph-group/pymeasure". This should be the reason, why it does not work anymore.

@cjermain

BenediktBurger:
I updated the readme.rst to use the _permanent_ Zenodo DOI, which points always to the most recent version.

Closes #1070 as the most recent version is now 0.13.1.post1.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,8 +19,8 @@ PyMeasure runs on Python 3.8-3.11, and is tested with continuous-integration on 
     :target: http://pymeasure.readthedocs.io/en/latest/?badge=latest
     :alt: Documentation Status
 
-.. image:: https://zenodo.org/badge/DOI/10.5281/zenodo.3732545.svg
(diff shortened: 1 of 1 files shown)


cjermain (approved):
Looks good!

bilderbuchi:
@BenediktBurger CI failed with this commit, and now in master, did you notice that?

BenediktBurger:
Yes, I noticed, that CI failed due to an error in the dependency resolution of micromamba.

As I did not change any code and additionally the codedov upload (with running the test suite) succeeded, I did ignore these errors for this PR.

By the way, on windows and linux the CI succeed. They fail only on macos. If it happens again, we have to investigate it.

PR #1105, opened 2024-05-09
Add Keysight E3642A DC Power supply 

mzen228:
Add Keysight E3642A DC Power supply

### DIFF
--- a/docs/api/instruments/keysight/keysightE3642A.rst
+++ b/docs/api/instruments/keysight/keysightE3642A.rst
@@ -1,0 +1,9 @@
+##########################################
+Keysight E3642A DC Power Supply
+##########################################
+
+.. autoclass:: pymeasure.instruments.keysight.KeysightE3642A
+    :members:
+    :show-inheritance:
+    :inherited-members:
+
--- a/pymeasure/instruments/keysight/keysightE3642A.py
+++ b/pymeasure/instruments/keysight/keysightE3642A.py
@@ -42,6 +42,13 @@ class KeysightE3642A(SCPIMixin, Instrument):
         map_values=True
     )
 
+    voltage = Instrument.control(
+        "VOLT?",
+        "VOLT %f",
+        """Control the output voltage""",
+    )
+
+
 if __name__ == "__main__":
     adapter = PrologixAdapter('ASRL3::INSTR')
     adapter.connection.read_termination = '\r\n'
@@ -49,4 +56,9 @@ if __name__ == "__main__":
     adapter.auto = True
 
     ps = KeysightE3642A(adapter)
-    ps.output_state_enabled = False
+    ps.output_state_enabled = True
+    print(ps.output_state_enabled)
+    ps.voltage = 5.31
+
+
+
--- a/pymeasure/instruments/keysight/keysightE3642A.py
+++ b/pymeasure/instruments/keysight/keysightE3642A.py
@@ -23,7 +23,7 @@
 #
 from pymeasure.instruments import SCPIMixin, Instrument
 from pymeasure.adapters.prologix import PrologixAdapter
-from pymeasure.instruments.validators import strict_discrete_set
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
 
 
 class KeysightE3642A(SCPIMixin, Instrument):
@@ -46,6 +46,9 @@ class KeysightE3642A(SCPIMixin, Instrument):
         "VOLT?",
         "VOLT %f",
         """Control the output voltage""",
+        validator=strict_range,
+        values=[0, 20],
+        dynamic=True,
     )
 
     current = Instrument.control(
--- a/pymeasure/instruments/keysight/keysightE3642A.py
+++ b/pymeasure/instruments/keysight/keysightE3642A.py
@@ -35,10 +35,10 @@ class KeysightE3642A(SCPIMixin, Instrument):
 
     output_state_enabled = Instrument.control(
         "OUTPUT:STATE?",
-        "OUTPUT:STATE %s",
+        "OUTPUT:STATE %d",
         """Control if output state is enabled""",
         validator=strict_discrete_set,
-        values={True: 'on', False: 'off'},
+        values={True: 1, False: 0},
         map_values=True
     )
 
--- a/pymeasure/instruments/keysight/keysightE3642A.py
+++ b/pymeasure/instruments/keysight/keysightE3642A.py
@@ -27,6 +27,8 @@ from pymeasure.instruments.validators import strict_discrete_set, strict_range
 
 class KeysightE3642A(SCPIMixin, Instrument):
     def __init__(self, adapter, name="Keysight E3642A", **kwargs):
+        """ Represents the Keysight E3642A DC Power Supply interface for interacting with
+        the instrument. """
         super().__init__(
             adapter, name,
             **kwargs
--- a/pymeasure/instruments/keysight/keysightE3642A.py
+++ b/pymeasure/instruments/keysight/keysightE3642A.py
@@ -43,25 +43,25 @@ class KeysightE3642A(SCPIMixin, Instrument):
         map_values=True
     )
 
-    voltage = Instrument.control(
+    voltage_setpoint = Instrument.control(
         "VOLT?",
         "VOLT %f",
-        """Control the output voltage""",
+        """Control the output voltage setpoint""",
         validator=strict_range,
         values=[0, 20],
         dynamic=True,
     )
 
-    current = Instrument.control(
+    current_limit = Instrument.control(
         "CURR?",
         "CURR %f",
-        """Control the output current""",
+        """Control the output current limit""",
         validator=strict_range,
         values=[0, 5],
         dynamic=True,
     )
 
-    def apply(self, voltage, current):
-        self.voltage = voltage
-        self.current = current
+    def apply(self, voltage_setpoint, current_limit):
+        self.voltage_setpoint = voltage_setpoint
+        self.current_limit = current_limit
 
--- a/pymeasure/instruments/keysight/keysightE3642A.py
+++ b/pymeasure/instruments/keysight/keysightE3642A.py
@@ -1,0 +1,51 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2024 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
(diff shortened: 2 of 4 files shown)


mzen228:
Can you please offer some feedback as to why my tests are generated with a bunch of "read eoi" commands?  I don't think I am generating the tests correctly.

mzen228:
The power supply's APPLY comand takes both voltage and current.    An example use of apply would be "APPLY 3.0, 1.0".  Is it possible to use the Instrument.control method to implement APPLY?  Can Instrument.control takes multiple arguments?

docs/api/instruments/keysight/keysightE3642A.rst
```diff
@@ -0,0 +1,9 @@
+##########################################
+Keysight E3642A DC Power Supply
+##########################################
```
BenediktBurger, line 3: The hashes should be as long as the text:
```suggestion
###############################
Keysight E3642A DC Power Supply
###############################
```

pymeasure/instruments/keysight/keysightE3642A.py
```diff
@@ -0,0 +1,67 @@
...
+        the instrument. """
+        super().__init__(
+            adapter, name,
+            **kwargs
+        )
+
+    output_state_enabled = Instrument.control(
+        "OUTPUT:STATE?",
+        "OUTPUT:STATE %d",
+        """Control if output state is enabled""",
```
BenediktBurger, line 40: Docstrings should end with a period (or exclamation mark...):
```suggestion
        """Control whether output state is enabled.""",
```

pymeasure/instruments/keysight/keysightE3642A.py
```diff
@@ -0,0 +1,67 @@
...
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True
+    )
+
+    voltage_setpoint = Instrument.control(
+        "VOLT?",
+        "VOLT %f",
+        """Control the output voltage setpoint""",
```
BenediktBurger, line 49: For such properties, it is quite useful to define the units, for example (I assume, that the value is in Volts):
```suggestion
        """Control the output voltage setpoint in Volts.""",
```

pymeasure/instruments/keysight/keysightE3642A.py
```diff
@@ -0,0 +1,67 @@
...
+# THE SOFTWARE.
+#
+from pymeasure.instruments import SCPIMixin, Instrument
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+
+
+class KeysightE3642A(SCPIMixin, Instrument):
+    def __init__(self, adapter, name="Keysight E3642A", **kwargs):
+        """ Represents the Keysight E3642A DC Power Supply interface for interacting with
+        the instrument. """
```
BenediktBurger, line 31: Please move this docstring to the class (directly after `class`).
The __init__ does not need its own docstring, it will get that one of the class, therefore you need to describe it just once.

pymeasure/instruments/keysight/keysightE3642A.py
```diff
@@ -0,0 +1,67 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+from pymeasure.instruments import SCPIMixin, Instrument
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+
+
+class KeysightE3642A(SCPIMixin, Instrument):
```
BenediktBurger, line 28: We already have several Keysight power supplies, is your power supply similar to one of the other ones?
If that's the case, we could simplify the codebase by having just one code for several instruments.

tests/instruments/keysight/test_keysightE3642A.py
```diff
@@ -0,0 +1,99 @@
...
+    with expected_protocol(
+            KeysightE3642A,
+            [(b'CURR 2.670000', None)],
+    ) as inst:
+        inst.current_limit = 2.67
+
+
+@pytest.mark.parametrize("comm_pairs, value", (
+    ([(b'CURR?', None),
+      (b'++read eoi', b'+2.67000000E+00\n')],
```
BenediktBurger, line 25: You're using a Prologix Adapter, and you used the test generator, aren't you?

The prologix adapter writes the `read eoi` message, which is not part of the instrument code itself.
You have to replace it by `None`.
As the previous message pair has already None as read message, you can combine them:
```suggestion
    ([(b'CURR?', b'+2.67000000E+00\n')],
```

Unfortunately, you have to do that manually afterwards, due to the Prologix Adapter.

BenediktBurger (changes requested):
Thanks for your contribution @mzen228 and welcome to the contributors.

You may add yourself to the `authors.txt` as well.

> The power supply's APPLY comand takes both voltage and current. An example use of apply would be "APPLY 3.0, 1.0". Is it possible to use the Instrument.control method to implement APPLY? Can Instrument.control takes multiple arguments?

Yes, you can give two values to a `control` (or  `setting` which is more fitting for you here), as you hand a tuple of two values to control.
However, as it is an action, a method is more appropriate.

Regarding the failing tests, I wrote something in the test file itself.

pymeasure/instruments/keysight/keysightE3642A.py
mzen228: What would having just one code for several instruments entail?  Would E3642A not be added to PyMeasure at all? 
 Or would it get added and just inherit functionality from a similar power supply?
 
There are 3 other Keysight power supplies: E3631A, E36312A, and N5767A.  The first two are triple output supplies, whereas my E3642A addition is not.  The N5767A has a voltage range from 0-60 V, whereas the E3642A has a range from 0-8 or 0-20V.  My additions also have different commands from the other supplies.

mzen228:
Do you know why I am failing the linting test?

tests/instruments/keysight/test_keysightE3642A.py
mzen228: Thanks for letting me know!

pymeasure/instruments/keysight/keysightE3642A.py
BenediktBurger: The first goal is, to make that instrument work in pymeasure. 

How a similar instrument is implemented depends on the naming scheme and the similarity.
It could be a series name like "345x series", if the code is exactly the same and the names are from the same series (not here, I think).
In another case it is just inheriting.

If only slight differences exist (like different voltage ranges), you can inherit the instrument and adjust the range with dynamic properties, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#instruments-with-similar-features . (Note the `dynamic=True` at the property definition) 

Now for your case:
Thanks for checking the differences.
The most important question is, whether the features are similar and the commands are the same.
One option could be to make your instrument a single channel version.

BenediktBurger:
> Do you know why I am failing the linting test?

There are several issues:
- one file has two empty lines at the end instead of one (if you look here on github at the files, it is marked as well)
- you did not add your instrument to the _init__.py in the keysight folder, therefore the documentation did not find the instrument
- the doctests should succeed, could you merge the current master branch into your branch, please? I hope this resolves that last issue.

pymeasure/instruments/keysight/keysightE3642A.py
BenediktBurger: I had a look into the different drivers.
Only `KeysightN5767A` seems similar, however, it is in a not that good state (truncated validators, naming..).

A question out of curiosity: would that driver work for your device as well (with the `voltage` / `current` commands, the status...)?

Can your device read the current voltage/current (like the N5767A)?

pymeasure/instruments/keysight/keysightE3642A.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self, adapter, name="Keysight E3642A", **kwargs):
+        super().__init__(
+            adapter, name,
+            **kwargs
+        )
+
+    output_state_enabled = Instrument.control(
+        "OUTPUT:STATE?",
+        "OUTPUT:STATE %d",
+        """Control whether output state is enabled.""",
```
BenediktBurger, line 40: We defined it to be called `output_enabled` for a device with an output: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#frequent-properties
```suggestion
    output_enabled = Instrument.control(
        "OUTPUT:STATE?",
        "OUTPUT:STATE %d",
        """Control whether output is enabled.""",
```

pymeasure/instruments/keysight/keysightE3642A.py
mzen228: I was unable to test whether the N5767A driver works for my device because the N5767A driver wouldn't accept my PrologixAdapter.  Dunder init in N5767A is looking for a `config` method but the adapter has no such method.

`if isinstance(self.adapter, VISAAdapter):
            self.adapter.config(
                is_binary=False,
                datatype='float32',
                converter='f',
                separator=','
            )`

mzen228: After commenting out the `self.adapter.config` line from `__init__`, I found that the N5767A driver does indeed work for my device.

mzen228:
@BenediktBurger What would you like to see happen with this PR?

pymeasure/instruments/keysight/keysightE3642A.py
BenediktBurger: Thanks for checking.

BenediktBurger:
Thanks for your patience, I was very busy and did not have time for pymeasure.

Thanks for checking compatibility with the other power supply.

Would you mind making a `KeysightSingleChannelPowerSupply` (or similar) class out of your driver?
Then your E3642A can inherit this SingleChannelPowerSupply and set the appropriate limits for voltage and current.
I open an issue to make the other power supply inherit from this base class. If you want to, you can then implement this change, but in another PR.

mzen228:
> Thanks for your patience, I was very busy and did not have time for pymeasure.
> 
> Thanks for checking compatibility with the other power supply.
> 
> Would you mind making a `KeysightSingleChannelPowerSupply` (or similar) class out of your driver? Then your E3642A can inherit this SingleChannelPowerSupply and set the appropriate limits for voltage and current. I open an issue to make the other power supply inherit from this base class. If you want to, you can then implement this change, but in another PR.


Thanks for opening a new issue.  I can make a KeysightSingleChannelPowerSupply class in a month or two when I'm more free...

Does #1105 need any more changes?  I see a message that says "1 change requested" but I'm not sure what request the message is referring to.  Can you help?

BenediktBurger:
> Does https://github.com/pymeasure/pymeasure/pull/1105 need any more changes? I see a message that says "1 change requested" but I'm not sure what request the message is referring to. Can you help?

It states, that one person finished a review with the state "change requested", it does not track, whether these changes have been implemented or not, so this is not an obstacle.

>  I can make a KeysightSingleChannelPowerSupply class in a month or two when I'm more free...

I thought, that you move your code into a BaseClass and then inherit from this BaseClass, which would not require a lot of work.
However, if you prefer, we can merge this PR right now and do the baseclass in the issue opened.

Issue #1106, opened 2024-05-10
CI fails due to timing issue occasionally


BenediktBurger:
Quite often one test fails due to timing issues making the whole CI pipeline fail, probably the time limit should be reduced slightly:
~~~
    def test_check_query_delay():
        with expected_protocol(ATSBase, [("TTIM?", "7")]) as inst:
            start = perf_counter()
            assert inst.maximum_test_time == 7
            delay = perf_counter() - start
>           assert delay > 0.05
E           assert 0.0491846000000038 > 0.05

tests\instruments\temptronic\test_temptronic_base.py:36: AssertionError
~~~

CasperSchippers:
It _sounds_ a bit like a floating-point errors: the `start` and `delay` might be too large (numericall) such that that difference cannot be calculated accurately.

BenediktBurger:
Another solution could be `pytest.approx`, however, we should check first, what is actually tested, maybe the test is ok, but the code under test not.

BenediktBurger:
Having merged #1109, CI runs succeed. After the previous two merges, they always failed.

Seems to work, we should keep it open though for some more time.

BenediktBurger:
It did not work in [CI run](https://github.com/pymeasure/pymeasure/actions/runs/9091306319/job/24985650504?pr=1039):
~~~
    def test_check_query_delay():
        with expected_protocol(ATSBase, [("TTIM?", "7")]) as inst:
            start = perf_counter_ns()
            assert inst.maximum_test_time == 7
            delay = perf_counter_ns() - start
>           assert delay > 0.05 * 1e9
E           assert 47828800 > (0.05 * 1000000000.0)
~~~

PR #1109, opened 2024-05-14
Fix temptronic test to use ns perf counter

BenediktBurger:
Should be a solution to #1106


CasperSchippers (approved):
Looks good to me, let's hope this works indeed.

PR #1110, opened 2024-05-15
Fix temptronic test timing issue

### LINKED ISSUE #1106: CI fails due to timing issue occasionally

BenediktBurger:
Closes #1106 

This should solve the problem: reducing the time limit.
It's brute force, but I do not see any other option.

### DIFF
--- a/tests/instruments/temptronic/test_temptronic_base.py
+++ b/tests/instruments/temptronic/test_temptronic_base.py
@@ -33,4 +33,7 @@ def test_check_query_delay():
         start = perf_counter_ns()
         assert inst.maximum_test_time == 7
         delay = perf_counter_ns() - start
-        assert delay > 0.05 * 1e9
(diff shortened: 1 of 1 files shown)


bilderbuchi:
Don't have this instrument, but this change looks fine to me.. a 5% jitter in timing on various systems would not concern me, and it's only in test code anyway.
:+1:

PR #1123, opened 2024-06-11
Fix the encoding of datafiles

### LINKED ISSUE #1118: Change the default display of units (managed by pint) in live plot using ManagedDockWindow
Hello,

I am experimenting with the GUI, more specifically the managed dock window for operating a new measurement setup.
One quantity I obtain is resistance. Now when I define the DATA_COLUMNS etc. I can directly provide the unit, e.g. 
```python
DATA_COLUMNS = ['resistance (ohm)']
```
When it comes to the actual live plot, the axis labels automatically obtain their labels from this definition, which results in an axis label that just reads **"Resistance (ohm)"** or when the value is small it automatically changes the label to **"Resistance (mohm)"**.
Personally, I would prefer if the units are displayed with their corresponding symbols by default, that means Ohm = $\Omega$ or temperature in °C instead of degC or so.
I tried to change the default format of the units in the units.py file, but that didn't change anything.
```python
from pymeasure.units import ureg
ureg.default_format = '~'
``` 
Any ideas where I need to change the code so I can adjust the default settings for the displayed units?

Further, I noticed that when I use °C as unit, i.e.
```python
DATA_COLUMNS = ['temperature (°C)']
```
the procedure is executed without problems and the results data also exported and stored in a file. However, the live plot doesn't show anything or update itself. The progress bar for example does it, there is just no visible live plot.
Any suggestions on how to fix this?

CasperSchippers:
Explicitly set the encoding of the writen datafiles when the files is opened (either for reading and writing). This ensures the files to be written and read correclty, even with special characters (e.g., ° or Ω); this also ensures the data is correctly plotted (since it is read from file for plotting).

Resolves #1118.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Upcoming version
 
 Automation
 ----------
-- Explicitly set encoding when writing and reading data to file, allowing the use of special characters (@CasperSchippers, #1123).
+- Explicitly set encoding to utf8 when writing and reading data to file, allowing the use of special characters (@CasperSchippers, #1123).
 
 Version 0.14.0 (2024-05-22)
 ===========================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+Upcoming version
+================
+
+
+Automation
+----------
+- Explicitly set encoding when writing and reading data to file, allowing the use of special characters (@CasperSchippers, #1123).
 
 Version 0.14.0 (2024-05-22)
 ===========================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Upcoming version
 
 Automation
 ----------
-- Explicitly set encoding to utf8 when writing and reading data to file, allowing the use of special characters (@CasperSchippers, #1123).
+- Explicitly set encoding to utf8 when writing and reading data to file, allowing the use of special characters.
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -1,3 +1,10 @@
+Upcoming version
+================
+
+
+Automation
+----------
+- Explicitly set encoding when writing and reading data to file, allowing the use of special characters (@CasperSchippers, #1123).
```
BenediktBurger, line 7: Maybe:
```suggestion
- Explicitly set encoding to utf8 when writing and reading data to file, allowing the use of special characters (@CasperSchippers, #1123).
```

BenediktBurger (approved):
Thanks for the quick fix.

One question: Does this break backward compatibility of previously written files?

CasperSchippers:
I _think_ it does not break backwards compatibility, since the reading did implicitly assume utf-8.

What this could break is other codes (without PyMeasure) that assumed a certain encoding (although this encoding was platform dependent).

CasperSchippers:
@BenediktBurger do you think we should do something with this potential breaking of other codes?

BenediktBurger:
It only affects new files, so it should be sufficient to add a note to the changelog.
If it should cause any issues, the person can adjust the encoding.

Sorry for overlooking that issue.

CasperSchippers:
No worries, I've added a note in the changelog (and in doing so found another place where the encoding should be explicitly set).

Issue #985, opened 2023-11-02
SR830 pause buffer


FlyingCurryMonster:
I'm trying to diagnose some digitization issues on my SR830 and noticed there's a bug related to pauseBuffer().  In the SR830 py file there are several cases where self.pause_buffer() is used, when the actual function call should be  "self.pauseBuffer()".

Edit: It also seems like the buffer_data() method isn't defined in the SR830 class.

bmoneke:
Yes, indeed. 8 years ago `pauseBuffer` was renamed to `pause_buffer`, but not all occurrences.

Would you mind fixing it? If you need help with it, we're here.

FlyingCurryMonster:
For sure.  It seems like the support for measurements with the buffer isn't too well supported at the moment.  That's not too surprising, I didn't need it until I ran into issues where my device noise is lower than the resolution of the SR830.
I'm guessing the workaround for digitization is to use the SR830 buffer and fast sample rate to record ~1000 measurements in the buffer and take the mean of the buffer, rather than record the front reading.  I think by central limit theorem (I think?) that should cure the quantization problem.  

I'll try and fix the pause and the buffer measure methods.

Any input or thoughts are appreciated.

bmoneke:
> Any input or thoughts are appreciated

For the naming, `pause_buffer` is the better name, as that is the normal python naming scheme.

Regarding the `buffer_data`, I don't know anything

seb5g:
Hi guys, 
I'm actually working on making this driver working again. Indeed the buffer_data was not defined... I should be done with a first update soon

BenediktBurger:
Thanks for your comment @seb5g.

FlyingCurryMonster:
I've made a pull request here https://github.com/pymeasure/pymeasure/pull/1125

seb5g:
I'm a bit lost here, has my PR on the subject been merged?

seb5g:
> I'm a bit lost here, has my PR on the subject been merged?

actually yes it is the #999 
@FlyingCurryMonster did you use an earlier version of pymeasure without this patch?

BenediktBurger:
Oh, merging your PR I missed that it closes this issue.

@FlyingCurryMonster  could you please check, whether the current master branch or even the current release (#999 is already part of the current released version) solves your issue?

FlyingCurryMonster:
Current branch solves my problems!

seb5g:
you're welcome ;-)

[...]

PR #1125, opened 2024-06-13
Sr830 buffer functionality

FlyingCurryMonster:
I've added three methods for measuring buffer.  

1.  buffer read method that transfers bytes over GPIB.  Manual claims it's the fastest transfer method among the 3 methods (ascii, binary, and bytes).
2. method to convert the array of bytes to float.  
3. A `buffer_measure_from_bytes()` method that returns the buffer from both channels.  

I mixed up some changes I was trying to make to the tektronix AFG3152C, but there should be no effective modification to that driver file despite the commit log suggesting otherwise.  

~~The buffer measure functionality didn't seem to work of the shelf, and there was syntax error with `pauseBuffer()` should be `pause_buffer`.~~

Edit: Seems like `pauseBuffer()` and `buffer_data` were fixed by a while ago.  

Discussion link:  https://github.com/pymeasure/pymeasure/issues/985


seb5g:
could you see a gain in acquisition time because when I tried this, it was taking so much time to retrieve data I finally ended using an external DAQ card plugged to the analog output of the lockin...

FlyingCurryMonster:
Back when I had an older version of pymeasure I noticed the reading binary method always took 4.2 seconds no matter the buffer size.  For small buffers this seemed slow.  I experimented with reading bytes and thought it was a little faster.  

Since updating my fork my code doesn't seem to work with FAST2 mode. I also had to add 'STRT' at the end of the FAST0 mode...Seems like whenever I try to get the buffer count in FAST2 mode I end up reading bytes in the buffer instead of the buffer size.  Might be a better idea to modify my commits to utilize current functionality.

PR #1128, opened 2024-06-20
Add Keithley DAQ6510

Aphelion82:
Add support for the [Keithley DAQ6510 Data Acquisition and Logging Multimeter System](https://www.tek.com/en/products/keithley/digital-multimeter/keithley-daq6510).

### DIFF
--- a/pymeasure/instruments/keithley/keithleyDAQ6510.py
+++ b/pymeasure/instruments/keithley/keithleyDAQ6510.py
@@ -24,7 +24,13 @@
 
 from pymeasure.instruments import Instrument
 from pymeasure.instruments.generic_types import SCPIMixin
+from pymeasure.instruments.validators import strict_discrete_set, truncated_range
 from .buffer import KeithleyBuffer
+import logging
+
+# Set up logging
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
 
 
 class KeithleyDAQ6510(KeithleyBuffer, SCPIMixin, Instrument):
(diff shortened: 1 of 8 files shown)


pymeasure/instruments/keithley/keithleyDAQ6510.py
```diff
@@ -0,0 +1,103 @@
...
+
+    open_channel = Instrument.setting(
+        ":ROUT:OPEN (@%d)",
+        """Set a single channel to open."""
+    )
+
+    close_channel = Instrument.setting(
+        ":ROUT:CLOS (@%d)",
+        """Set a single channel to closed."""
+    )
```
BenediktBurger, line 56: You use it as an "action word", therefore these properties should be methods instead.

BenediktBurger:
Thanks for your contribution.

I do not get, what you can do with this device. Until now, you can only open a channel, but what does that do?

pymeasure/instruments/keithley/keithleyDAQ6510.py
Aphelion82: Changed to be methods

Aphelion82:
> Thanks for your contribution.
> 
> I do not get, what you can do with this device. Until now, you can only open a channel, but what does that do?

The DAQ6510 is able to measure voltage and current, among other things. I haven't yet implemented these

BenediktBurger:
That explains it.

Aphelion82:
Thank you—as far as the opening of channels, the DAQ6510 uses a MUX card with channels that can be opened and closed.

The codecov check fails due to a codecov token being missing. Do I need to add that on my end?

BenediktBurger:
> The codecov check fails due to a codecov token being missing. Do I need to add that on my end?

No, you cannot add the token. And for token less upload, codecov runs sometimes into api call limits. So, don't worry.

Aphelion82:
The checks fail on the docstrings, but I had been following the wording used by preexisting instrument files (see [keithley2400](https://github.com/pymeasure/pymeasure/blob/606fae7e8063594394a089ee88d2aa5df8f19570/pymeasure/instruments/keithley/keithley2400.py#L151-L163) and [keithley2450](https://github.com/pymeasure/pymeasure/blob/606fae7e8063594394a089ee88d2aa5df8f19570/pymeasure/instruments/keithley/keithley2450.py#L91-L104)) for similar/identical purposes. Should I still change the docstrings, or leave it as it is?

BenediktBurger:
> The checks fail on the docstrings, but I had been following the wording used by preexisting instrument files (see [keithley2400](https://github.com/pymeasure/pymeasure/blob/606fae7e8063594394a089ee88d2aa5df8f19570/pymeasure/instruments/keithley/keithley2400.py#L151-L163) and [keithley2450](https://github.com/pymeasure/pymeasure/blob/606fae7e8063594394a089ee88d2aa5df8f19570/pymeasure/instruments/keithley/keithley2450.py#L91-L104)) for similar/identical purposes. Should I still change the docstrings, or leave it as it is?

Please change the docstrings. We adjusted at some point in history the [coding standards](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings) (in order to be in line with python recommendations), however we did not change all the docstrings of all instruments yet, as it is quite some work to change the wrong docstrings of all instruments.

pymeasure/instruments/keithley/keithleyDAQ6510.py
```diff
@@ -0,0 +1,286 @@
...
+        The largest value is the longest time, and results in the lowest reading rate,
+        but increases the number of usable digits. """
+    )
+
+    ####################
+    # Methods        #
+    ####################
+
+    def measure_resistance(self, nplc=1, resistance=100e6, auto_range=True):
+        """ Configures the measurement of resistance.
```
BenediktBurger, line 183: Docstrings should be (according to [PEP 257](https://peps.python.org/pep-0257/)) in imperative mode:
```suggestion
        """ Configure the measurement of resistance.
```

Equally below.

pymeasure/instruments/keithley/keithleyDAQ6510.py
```diff
@@ -0,0 +1,286 @@
...
+            self.current_range = current
+        self.check_errors()
+
+    def no_errors(self):
+        """
+        Check to see if the instrument has any errors returned or not.
+
+        :return: ``True`` if there are no errors, ``False`` if there are errors.
+        """
+        return len(self.check_errors() == 0)
```
BenediktBurger, line 242: Did you know, that `bool([]) is False`? That means, you can use `if inst.check_errors(): handle_errors()`?

That would make that method superfluous.

A general suggestion: try to word a boolean property, that it works like a question: `no_errors() is True` (or `something_enabled = True`, as that makes it easier to interpret True/False.
In such a case (like your name), you can write in the docstring: `:return: whether there are errors.`. (or maybe `:return bool: whether there are errors.`

tests/instruments/keithley/test_keithleyDAQ6510_with_device.py
BenediktBurger, line 1: Thanks for adding these device tests.

However, we maintainers like to have [protocol tests](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#protocol-tests), which do not require the instrument (which we do not have).
These protocol tests allow us to verify, that the instrument continues to work, even if we change code (in the core).

If you need help, don't hesitate to ask.

BenediktBurger (changes requested):
Thanks for your contribution.

Your code is already in a good shape, only a few small issues.

pymeasure/instruments/keithley/keithleyDAQ6510.py
Aphelion82: Changed all that I spotted

pymeasure/instruments/keithley/keithleyDAQ6510.py
Aphelion82: Thanks—I just removed the method.

tests/instruments/keithley/test_keithleyDAQ6510_with_device.py
Aphelion82: I created a protocol test for the DAQ6510 going off of some [other tests](https://github.com/pymeasure/pymeasure/blob/b688a09a5d9b832d4bf65fea0227c780b1394c05/pymeasure/instruments/keithley/keithleyDMM6500.py) and the documentation. Let me know if there's anything wrong. Thanks for the help!

BenediktBurger: Thanks for adding them.

tests/instruments/keithley/test_keithleyDAQ6510.py
```diff
@@ -0,0 +1,88 @@
...
+#
+
+from pymeasure.test import expected_protocol
+from pymeasure.instruments.keithley import KeithleyDAQ6510
+
+
+def test_init():
+    with expected_protocol(
+            KeithleyDAQ6510,
+            [(b'*LANG SCPI', None)],
```
BenediktBurger, line 32: Could it be, that you removed the "lang scpi" command from the init?

At least all tests have it and they all fail, as this command is never issued (see the failing tests).

Aphelion82: Yeah, I had wondered if that was the case, looking at the tests

Aphelion82: Removed `*LANG SCPI`

tests/instruments/keithley/test_keithleyDAQ6510.py
```diff
@@ -26,63 +26,49 @@
 from pymeasure.instruments.keithley import KeithleyDAQ6510
 
 
-def test_init():
-    with expected_protocol(
```
BenediktBurger: You can keep the init test (albeit with an empty comm_pair_list), if you wish.

BenediktBurger (approved):
One last question, afterwards I can merge your pr, if you agree.

Thanks for your contribution.

tests/instruments/keithley/test_keithleyDAQ6510.py
Aphelion82: Do you think it's a good idea? I'll add it back anyways.

Aphelion82:
Thanks so much for your help.

BenediktBurger:

PR #1129, opened 2024-06-21
SRS SR850 lock-in amplifier

barentine:
I have an SR850 lock-in amplifier and I'd like to change settings on over GPIB, for which I found the SR830 implementation here useful (thanks!). I'm not too familiar with the package, but had a look at the contributing guidelines - please let me know if I missed anything.

The SR850 is fairly similar to the SR830, so I've subclassed it. I left the SR830 completely alone, which I figured is best without having one to test myself.

This implementation is not full-fledged. In particular, it doesn't have the buffer reads that the SR830 implementation does, but as I haven't yet needed them myself I left them unimplemented for now. Maybe it's useful for others already in its current state (configure some settings and read in current values).

For now I raise `NotImplementedError`s on sr830 calls that aren't implemented for the SR850. Not sure what the cleanest way to handle that is.

### DIFF
--- a/pymeasure/instruments/srs/sr850.py
+++ b/pymeasure/instruments/srs/sr850.py
@@ -1,0 +1,191 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2024 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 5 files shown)


BenediktBurger:
> Maybe it's useful for others already in its current state (configure some settings and read in current values).

If it is in a useful state, it is sufficient and you do not need to implement everything. 

> For now I raise NotImplementedErrors on sr830 calls that aren't implemented for the SR850. Not sure what the cleanest way to handle that is.

If the methods are present in the device, but not implemented, it is fine. (please write it in the docstring as well, as only docstrings are shown on the web page).

pymeasure/instruments/srs/sr850.py
```diff
@@ -0,0 +1,180 @@
...
+    def __init__(self, adapter, name="Stanford Research Systems SR850 Lock-in amplifier",
+                 **kwargs):
+        # write term lf or cr on RS232, lf or eoi on GPIB
+        kwargs.setdefault('write_termination', '\n')  
+        super().__init__(
+            adapter,
+            name,
+            **kwargs
+        )
+        if 'GPIB' in self.name.capitalize():
```
BenediktBurger, line 52: Why does the name contain the connection?

You could check, whether "gpib" is in the adapter string (if it is string), instead.

pymeasure/instruments/srs/sr850.py
```diff
@@ -0,0 +1,180 @@
...
+            self.output_interface = 'GPIB'
+    output_interface = Instrument.control(
+        "OUTX?", "OUTX %d",
+        """ An string property that controls the output interface. Inputs are
```
BenediktBurger, line 56: Docstrings are in imperative mode (at least the first sentence).

For a control, it should start with "Control", for a setting with "Set" and for a measurement with "Get" or "Measure".
For example:
```suggestion
        """Control the output interface. Inputs are

```

PR #1132, opened 2024-06-27
SR830 FAST2 buffer functionality

FlyingCurryMonster:
Discussion post: https://github.com/pymeasure/pymeasure/issues/985

The main addition in this PR is the ability to use the SR830's FAST2 method for data transfer.  In this mode, as soon as the buffer starts to populate with (X, Y) data, the instrument sends data to the computer without the need for an ask command.  This allows for the total measurement time (sampling of data and computer I/O) to be close to the `buffer_count` divided by the the sampling frequency.

Because of the nature of FAST2 transfer method, ordinary ask commands following a `start_buffer(fast=True)` call yield an error as the read buffer of the adapter has already been populated by data in the form of bytes.  And for high sampling rates, the adapter's read buffer may populate in between a successful `read_bytes` command and `pause_buffer` command.  To deal with  unintentional buffer population `read_bytes(-1)` is called to clear the buffer with a fast timeout setting.  

I have not made any changes to the old `buffer_measure` method as it can still work for `FAST0`, but I have not had success with it on my machine.  It also does not produce the buffer size that is passed into the arguments (see discussion link).

There is also a minor typo correction in the `set_scaling` method, and option to just control just the offset or just the expand.


BenediktBurger (changes requested):
Thanks for your contribution.

A few comments.

@seb5g a change to the SR830, would you like to look at it?

pymeasure/instruments/srs/sr830.py
```diff
@@ -582,6 +634,31 @@ def get_buffer(self, channel=1, start=0, end=None):
...
+        if end is None:
+            end = self.buffer_count
+        self.write("TRCL?%d,%d,%d" % (channel, start, end - start))
+        return self.buffer_bytes_convert(self.read_bytes(-1))
+
+    def buffer_bytes_convert(self, buffer):
+        '''
+        Converts the SR830's buffer in bytes to decimal numbers.
+        This formula was derived from the programming manual.
+        '''
```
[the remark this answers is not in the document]
FlyingCurryMonster: Got it, I think this should be corrected now

pymeasure/instruments/srs/sr830.py
```diff
@@ -421,16 +421,21 @@ def get_scaling(self, channel):
...
         """ Sets the offset of a channel (X=1, Y=2, R=3) to a
         certain percent (-105% to 105%) of the signal, with
         an optional expansion term (0, 10=1, 100=2)
         """
         if channel not in self.CHANNELS:
             raise ValueError('SR830 channel is invalid')
+        if percent is None:
+            percent = self.get_scaling(channel)[0]
+        if expand is None:
+            expand = self.get_scaling(channel)[1]
```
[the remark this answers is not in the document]
FlyingCurryMonster: Hmm, I find that I use `auto_offset` to set the percent scaling often, so I really just want to control the expand without having to pass anything into percent on the `set_scaling` method.  I think I implemented your suggestion for calling `get_scaling` just once.  What do you think?

pymeasure/instruments/srs/sr830.py
```diff
@@ -553,14 +558,61 @@ def buffer_measure(self, count, stopRequest=None, delay=1e-3):
...
+        Buffer measurement method that returns
+        both channel 1 and channel 2 buffers as np.arrays.
+        Args:
+        buffer_size:  Desired minimum buffer length.
+        If fast == True,  FAST2 is the data transfer method.
+        FAST2 sends data through the adapter ~0.5seconds
+        after `start_buffer` is called.
+        If  fast == False, FAST0 is the transfer mode and
+        an appropriate timeout needs to be spcified.
+        See programming section of the SR830 manual for more detail.
```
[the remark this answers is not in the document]
FlyingCurryMonster: Thank you for the edits!  These should be corrected now.

pymeasure/instruments/srs/sr830.py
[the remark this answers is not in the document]
BenediktBurger: Yes, you implemented, what I suggested.

pymeasure/instruments/srs/sr830.py
```diff
@@ -582,6 +638,30 @@ def get_buffer(self, channel=1, start=0, end=None):
...
 
+    def get_buffer_bytes(self, channel=1, start=0, end=None):
+        """ Acquires the 32 bit floating point data through bytes transfer
+        This is the fastest transfer method according to the manual.
+        The byte to float conversion is provided in the manual
+        and implemented in :meth:`buffer_bytes_convert`
+        """
+        if end is None:
+            end = self.buffer_count
+        self.write("TRCL?%d,%d,%d" % (channel, start, end - start))
```
BenediktBurger, line 649: In most cases it is better to use f-strings instead of this old percent style as it makes it easier to attribute the values to the position:
```suggestion
        self.write(f"TRCL?{channel},{start},{end - start}")
```

BenediktBurger:
@FlyingCurryMonster , would you mind adding (protocol) tests?
Especially the `buffer_bytes_convert` method could profit from tests, but maybe even `buffer_measure_fast`.

For the protocol tests, you can use the [test generator](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#test-generator), if you like.

FlyingCurryMonster:
It seems like there's a bug in the case that `fast=False`, it returns an empty array.   

I can also write a device test for these methods, SR830 is fairly common so maybe someone could verify.

FlyingCurryMonster:
@BenediktBurger The `fast=False` case is fixed for the method.  As for the protocol tests, it seems like it depends on signal fed into the instrument which will have some inherent noise as well.  

Maybe feeding in the lockin's internal oscillator and using a high time constant might work for a protocol test?

BenediktBurger:
The protocol tests are only messages sent to/from the driver, so a static test will always succeed, unless the written driver does something dependent on the current time.

FlyingCurryMonster:
I added a protocol test for `buffer_measure_fast` which tests for `fast=False` and `fast=True`.  Should I also add tests for "helper" methods like `buffer_bytes_convert` or `buffer_measure_bytes`?  Those two methods are used when `fast=False`.

pymeasure/instruments/srs/buffer_measure_fast test.py
BenediktBurger, line 1: Please move this file to the tests folder (`pymeasure/tests/instruments...`) and rename it to `test_sr830.py`, like all the other instrument test files.

pymeasure/instruments/srs/buffer_measure_fast test.py
```diff
@@ -0,0 +1,103 @@
+import pytest
+import numpy as np
+import numpy.testing as npt
```
BenediktBurger, line 3: I didn't know about numpy testing, good to know, thanks

BenediktBurger:
> I added a protocol test for `buffer_measure_fast` which tests for `fast=False` and `fast=True`. Should I also add tests for "helper" methods like `buffer_bytes_convert` or `buffer_measure_bytes`? Those two methods are used when `fast=False`.

If you deem it useful to test these methods individually (in order to catch bugs), add tests.

pymeasure/instruments/srs/buffer_measure_fast test.py
FlyingCurryMonster: seems like there's already a file there, should I add my methods to the current version?

FlyingCurryMonster: @BenediktBurger I've appended my protocol tests to `test_sr830.py`.

FlyingCurryMonster:
@seb5g could you check out this PR on your SR830's maybe?  I'm curious to see if `buffer_measure_fast(fast=True)` works for other users.

pymeasure/instruments/srs/sr830.py
```diff
@@ -421,16 +421,24 @@ def get_scaling(self, channel):
         offset, expand = self.ask("OEXP? %d" % channel).split(',')
         return float(offset), self.EXPANSION_VALUES[int(expand)]
 
-    def set_scaling(self, channel, precent, expand=0):
+    def set_scaling(self, channel, percent=None, expand=None):
```
BenediktBurger, line 424: Good catch.

However, please add a parameter `precent=None` for backward compatibility. This parameter should be used, if `percent=None` due to old code. In this case, you should raise a deprecation warning, see #672 for examples.
Please add also a line to deprecation section of the changelog, that `precent` is dperecated in favor of `percent`

pymeasure/instruments/srs/sr830.py
```diff
@@ -553,14 +561,63 @@ def buffer_measure(self, count, stopRequest=None, delay=1e-3):
         ch2[index:count] = self.get_buffer(2, index, count)
         return (ch1.mean(), ch1.std(), ch2.mean(), ch2.std())
 
+    def buffer_measure_fast(self, buffer_size, fast=True, timeout=60,):
+        """Measure the buffer in either the fast or slow
+        data transfer mode. Returns X, Y values in the fast
+        mode as np.arrays.
```
BenediktBurger, line 567: You can write about the return values after the parameters, with `:return: X, Y values as tuple of numpy arrays`.

pymeasure/instruments/srs/sr830.py
```diff
@@ -582,6 +639,30 @@ def get_buffer(self, channel=1, start=0, end=None):
         return self.binary_values("TRCB?%d,%d,%d" % (
             channel, start, end - start))
 
+    def get_buffer_bytes(self, channel=1, start=0, end=None):
+        """ Acquires the 32 bit floating point data through bytes transfer
+        This is the fastest transfer method according to the manual.
+        The byte to float conversion is provided in the manual
+        and implemented in :meth:`buffer_bytes_convert`
+        """
```
BenediktBurger, line 647: According to PEP257, you should have a single summary line, then an empty line and then a more detailed explanation.
All should be in imperative mode (Acquire instead of Acquires...)

BenediktBurger (approved):
Looks good, with a few small comments regarding docstrings.

@seb5g , will you have a look?

seb5g:
sorry guys but I have no time for this at the moment... I could not even finish my PR on elliptec motors...

pymeasure/instruments/tektronix/afg3152c.py
```diff
@@ -69,7 +73,7 @@ class AFG3152CChannel(Channel):
         validator=strict_discrete_set,
         values=UNIT_LIMIT,
     )
-
+
```
BenediktBurger, line 76: I guess you did not want to add spaces.

pymeasure/instruments/tektronix/afg3152c.py
```diff
@@ -100,6 +104,24 @@ class AFG3152CChannel(Channel):
         """ Control the amplitude offset. It is always in Volt. (float)""",
     )
 
+    # phase degree should not be a query because it always returns rad
+    phase_deg = Instrument.setting(
+        "phase:adjust %e DEG",
+        """ A floating point set comomand that controls
+          the phase in degree units.""",
```
BenediktBurger, line 111: Please use the correct [docstring style](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings) (different from earlier used style):
```suggestion
        """ Set the phase in degrees (float).""",
```

Also below (but with Control instead of Set)

pymeasure/instruments/tektronix/afg3152c.py
```diff
@@ -100,6 +104,24 @@ class AFG3152CChannel(Channel):
         """ Control the amplitude offset. It is always in Volt. (float)""",
     )
 
+    # phase degree should not be a query because it always returns rad
```
BenediktBurger, line 107: Why don't you modify the response (with a [get_process](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#processing-of-return-values))?

BenediktBurger (commented):
Small comments/suggestions.

FlyingCurryMonster:
Sorry those changes were not meant to be made on this PR, as it's for a different instrument.  Should I start a new pull request for these edits?

BenediktBurger:
Please tell, when you want a new review.

Issue #1130, opened 2024-06-25
id( ) command fails for Agilent B1500 SDA


AdMico:
Command gives:

TypeError: 'str' object is not callable

Might want to name it something other than id?

msmttchr:
`id` is a property which returns the output of scpi command `*IDN?`, so you should not add round brackets as you would when calling a method.

BenediktBurger:
You just use `device_ID = inst.id` to assign the ID to some variable. You use all these properties like variables of the device, not like methods.

AdMico:
OK, then I'd say it's a documentation error, as the brackets I had just came directly from the documentation.

msmttchr:
> OK, then I'd say it's a documentation error, as the brackets I had just came directly from the documentation.

Could you please send some reference where the documentation error is ?

AdMico:
https://pymeasure.readthedocs.io/en/latest/api/instruments/agilent/agilentB1500.html

About halfway down the Command Translation list in General Information.

msmttchr:
Indeed, you are right. It looks like a documentation bug. If you feel comfortable, you can open a PR for it, otherwise I will try to do it on my side. Thanks for spotting this problem.

AdMico:
Might be better that I let you do that one (unless you can point me at a good resource for how to do it) as I haven't done that before and don't know where to start.

PR #1133, opened 2024-06-29
Fix documentation errors

msmttchr:
Some properties were declared as methods by mistake.
(See #1130)


BenediktBurger:
Looks good to me, thanks for fixing it.

I mark this PR to resolve #1130

PR #1142, opened 2024-07-19
Added new meassurement device and corrected documentation comment

Konradrundfunk:


BenediktBurger:
Hey @Konradrundfunk , thanks for your contribution.

Is any other device similar to your new one, such that no new driver is needed?

If it remains a new device, please add also the device documentation (in the docs/api) folder.

Konradrundfunk:
Okay thanks, I think only that one device will support the specified ranges. Will add the documentation asap.

docs/api/instruments/keysight/keysightE36311A.rst
BenediktBurger, line 1: You need to include a link to this file in the parent's `index.rst` file. That will fix the warning shown below.

pymeasure/instruments/keysight/keysightE36311A.py
```diff
@@ -0,0 +1,104 @@
...
+        "OUTPut? (@{ch})",
+        "OUTPut %d, (@{ch})",
+        """Control whether the channel output is enabled (boolean).""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, False: 0},
+    )
+
+
+class KeysightE36311A(SCPIUnknownMixin, Instrument):
```
BenediktBurger, line 74: Does the device support SCPI? Then use `SCPIMixin` instead of `SCPIUnknownMixin`. Otherwise do not use a Mixin, but set `includeSCPI=False` in the super init call.

pymeasure/instruments/keysight/keysightE36311A.py
```diff
@@ -0,0 +1,104 @@
...
+import logging
+
+from pymeasure.instruments import Instrument, Channel, SCPIUnknownMixin
+from pymeasure.instruments.validators import strict_range, strict_discrete_set
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class VoltageChannel(Channel):
```
BenediktBurger, line 35: Does this channel differ from the E36312A version?

If not, what are the differences to E36312A, only the limits?

In this case, you could make a single file with the Voltage channel and below the two classes.

BenediktBurger (changes requested):
A few comments. Thanks for your patience.

PR #1145, opened 2024-07-30
Fix race condition in test suite due to HP8116A tests

### LINKED ISSUE #1144: Race condition in test suite involving HP8116A tests
The pass/fail outcome of the test suite is currently dependent on execution order. On my machine, I'm seeing the following failure:

```
FAILED tests/instruments/test_all_instruments.py::test_property_docstrings[HP8116A.status] - AttributeError: 'NoneType' object has no attribute 'split'
```

At first glance, this seems to indicate that the `status` property of the `HP8116A` instrument class does not have a docstring but, looking at the source code in `pymeasure/instruments/hp/hp8116a.py`, it clearly does:

```python
    @property
    def status(self):
        """Get the status byte of the 8116A as a :class:`Status` IntFlag-type enum."""
        return Status(self.adapter.connection.read_stb())
```

Upon closer inspection, it turns out that the tests for the HP8116A instrument (in `tests/instruments/hp/test_hp8116a.py`) modify the `status` property of the instrument object to mock the value of `status`:

```python
HP8116A.status = property(fget=lambda self: Status(5))
```

Depending on whether these tests run before or after the docstring test in `tests/instruments/test_all_instruments.py`, the test suite either passes or fails.

I have a plausible-looking fix for this issue that I'll propose as a PR.

kpet:
Do not modify the HP8116A instrument class in tests and derive a variant that overrides the status property for testing instead.

Fixes #1144

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -85,3 +85,4 @@ Nick James Kirkby
 Konrad Gralher
 David Ziliak
 David Sun
+Kévin Petit
--- a/tests/instruments/hp/test_hp8116a.py
+++ b/tests/instruments/hp/test_hp8116a.py
@@ -28,7 +28,11 @@ from pymeasure.test import expected_protocol
 from pymeasure.instruments.hp import HP8116A
 from pymeasure.instruments.hp.hp8116a import Status
 
-HP8116A.status = property(fget=lambda self: Status(5))
+
+class HP8116AWithMockStatus(HP8116A):
+    @property
+    def status(self):
+        return Status(5)
 
 
 init_comm = [(b"CST", b"x" * 87 + b' ,\r\n')]  # communication during init
@@ -36,7 +40,7 @@ init_comm = [(b"CST", b"x" * 87 + b' ,\r\n')]  # communication during init
 
 def test_init():
     with expected_protocol(
-            HP8116A,
+            HP8116AWithMockStatus,
             init_comm,
     ):
         pass  # Verify the expected communication.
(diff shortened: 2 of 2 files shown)


BenediktBurger:
Thanks for the fix.
As a side note: It is possible to give a docstring to the property (with `doc` parameter).
However, the `test_all_instruments` shall test the instruments as they are and not as they are modified by other tests.
Therefore, your approach is the correct one, I think.

BenediktBurger:
I'm pondering, whether it would be good to add a comment, why we mock the `status`.

Please tell me, whether I should merge directly or wait for a comment.

kpet:
>I'm pondering, whether it would be good to add a comment, why we mock the status.

It's a question for the original author of these tests really but it seems to be because the default STB value returned by `MagicMock` is not valid for this instrument. If I remove the mock status value, the tests fail with:

```
FAILED tests/instruments/hp/test_hp8116a.py::test_init - ValueError: <MagicMock name='mock.read_stb()' id='139761073010592'> is not a valid Status
FAILED tests/instruments/hp/test_hp8116a.py::test_duty_cycle - ValueError: <MagicMock name='mock.read_stb()' id='139761126050176'> is not a valid Status
FAILED tests/instruments/hp/test_hp8116a.py::test_duty_cycle_setter - ValueError: <MagicMock name='mock.read_stb()' id='139761073116688'> is not a valid Status
FAILED tests/instruments/hp/test_hp8116a.py::test_sweep_time - ValueError: <MagicMock name='mock.read_stb()' id='139761072210480'> is not a valid Status
FAILED tests/instruments/hp/test_hp8116a.py::test_limit_enabled - ValueError: <MagicMock name='mock.read_stb()' id='139761072924016'> is not a valid Status
```

>Please tell me, whether I should merge directly or wait for a comment.

I can add a comment stating this but it'd just be my guess :). Given this PR does not introduce the mocking but just changes how it's performed, I'm tempted to suggest to merge the PR as is if you're fine with it.

BenediktBurger:
Thanks for clarifying. In this case, I'll merge.

PR #1146, opened 2024-07-31
Implement Agilent E5062A

AlecVercruysse:
Hi,

I implemented basic functionality for the Agilent E5062A Vector Network Analyzer.

I used a mix of the Agilent 8722ES and HP8753E implementations for naming conventions for the more common knobs on the VNA.

I tested all functionality while connected to the device (with a `*with_device.py` test), but running the test generator did not seem to work for me. When changing channel properties, the output `pytest.mark.parameterize` list had objects like `<pymeasure.instruments.agilent.agilentE5062A.AgilentE5062A.VNAChannel object at 0x13bc08c10>` in the list, and all tests wrapped a `test_channels_getter` function that just checked `inst.channels == value`. Let me know if the device-based test cases are sufficient or if I should look further into the generator. 

Thank you for the review!

### DIFF
--- a/pymeasure/instruments/agilent/agilentE5062A.py
+++ b/pymeasure/instruments/agilent/agilentE5062A.py
@@ -72,30 +72,28 @@ class VNAChannel(Channel):
         tr_id = len(self.traces) + 1 if override_id is None else override_id
         self.add_child(VNATrace, tr_id, 'traces', 'tr_')
 
-    def _update_trace_count(self, active_traces=None):
+    def _update_trace_count(self, active_traces):
         """Internal method that updates the trace list based on the number
         of active traces.
         """
-        if active_traces is None:
-            active_traces = self.active_traces
         while len(self.traces) > active_traces:
             self.remove_child(self.traces[max(self.traces.keys())])
         while len(self.traces) < active_traces:
             self._add_trace()
 
-    def _update_power_values(self):
+    def _update_power_values(self, attenuation):
         """Internal method that updates the valid range of stimulus power
         based on the configured attenuation"""
         self.power_values = (
-            -self.attenuation - 5,
-            -self.attenuation + 10)
+            -attenuation - 5,
+            -attenuation + 10)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # self.traces = []
-        self._add_trace(override_id=1)   # add just a single trace to init it all
-        self._update_trace_count()       # add remaining traces if necessary
-        self._update_power_values()      # update dynamic limits of power
+        self._add_trace(override_id=1)                # add just a single trace to init it all
+        self._update_trace_count(self.active_traces)  # add remaining traces if necessary
+        self._update_power_values(self.attenuation)   # update dynamic limits of power
 
     start_frequency = Channel.control(
         "SENSe{ch}:FREQuency:STARt?", "SENSe{ch}:FREQuency:STARt %g",
@@ -200,7 +198,7 @@ class VNAChannel(Channel):
         value = int(float(value))
         value = strict_range(value, [1, 4])
         self.write("CALC{ch}:PARameter:COUNt %d" % value)
-        self._update_trace_count()
+        self._update_trace_count(value)
 
     power = Channel.control(
         "SOURce{ch}:POWer?", "SOURce{ch}:POWer %g",
@@ -227,7 +225,7 @@ class VNAChannel(Channel):
         value = int(float(value))
         value = strict_discrete_set(value, [0, 10, 20, 30, 40])
         self.write("SOURce{ch}:POWer:ATTenuation %d" % value)
-        self._update_power_values()
+        self._update_power_values(value)
 
     display_layout = Channel.control(
         "DISPlay:WINDow{ch}:SPLit?", "DISPlay:WINDow{ch}:SPLit %s",
@@ -325,15 +323,8 @@ class VNAChannel(Channel):
         """
         header = self.read_bytes(8)  # 2 start bytes + 6 <nbytes> bytes
         if not header[0:2].decode('ascii') == '#6':
-            fmt = self.ask("FORM:DATA?").strip()
-            if not (fmt == "REAL"):
-                raise ValueError(
-                    'Unrecognized header. Data transfer format is incorrect!' +
-                    f'is {fmt}, should be REAL')
-            else:
-                raise ValueError(
-                    'Unrecognized header, even though the configured ' +
-                    'data transfer format is correct.')
+            raise ValueError('''Unrecognized header. Ensure the data transfer
+            format (FORM:DATA?) is set to "REAL"''')
         nbytes = int(header[2:].decode('ascii'))
         binary_data = self.read_bytes(nbytes)
         terminator = self.read_bytes(1)
--- a/pymeasure/instruments/agilent/agilentE5062A.py
+++ b/pymeasure/instruments/agilent/agilentE5062A.py
@@ -289,8 +289,9 @@ class AgilentE5062A(SCPIMixin, Instrument):
             """Control the data format of the *active trace* of the channel
             (str). Default is MLOGarithmic. From the programmer's manual:
 
+            +--------------+-----------------------------------------------+
             | Setting      | Description                                   |
-            |--------------+-----------------------------------------------|
+            +--------------+-----------------------------------------------+
             | MLOGarithmic | Specifies the logarithmic magnitude format.   |
             | PHASe        | Specifies the phase format.                   |
             | GDELay       | Specifies the group delay format.             |
@@ -308,6 +309,8 @@ class AgilentE5062A(SCPIMixin, Instrument):
             | IMAGinary    | Specifies the imaginary format.               |
             | UPHase       | Specifies the expanded phase format.          |
             | PPHase       | Specifies the positive phase format.          |
+            +--------------+-----------------------------------------------+
+
             """,
             validator=strict_discrete_set,
             values=TRACE_FORMAT
--- a/pymeasure/instruments/agilent/agilentE5062A.py
+++ b/pymeasure/instruments/agilent/agilentE5062A.py
@@ -262,7 +262,7 @@ class VNAChannel(Channel):
     )
 
     TRACE_FORMAT = [
-        'MLOGarithmic'
+        'MLOGarithmic',
         'PHASe',
         'GDELay',
         'SLINear',
--- a/pymeasure/instruments/agilent/agilentE5062A.py
+++ b/pymeasure/instruments/agilent/agilentE5062A.py
@@ -139,12 +139,21 @@ class VNAChannel(Channel):
 
     sweep_type = Channel.control(
         "SENSe{ch}:SWEep:TYPE?", "SENSe{ch}:SWEep:TYPE %s",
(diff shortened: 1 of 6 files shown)


BenediktBurger:
Please add protocol tests, either manually or with the test generator.
What do you mean with "changed channel properties"?
Where all tests broken or just a few?

pymeasure/instruments/agilent/agilentE5062A.py
```diff
@@ -0,0 +1,540 @@
...
+    more significant missing functionality includes (TODO):
+
+    - editing the scale of the graphs
+    - markers
+    - calibration
+    - power sweeps
+
+    """
+
+    class VNAChannel(Channel):
```
BenediktBurger, line 75: Interesting solution to make the Channel an attribute of the instrument class.
Does it show in the api documentation? (you can in the docs folder and execute `make html` to create the documentation locally, which will be in the folder `builds`)

AlecVercruysse: The classes were showing up in the documentation, but they were missing the nice blue header, so I moved them outside. It seems that the docstring tests also did not check the methods of those classes while they were defined as attributes

AlecVercruysse:
I ended up manually implementing the protocol tests, but didn't look further into the issue. I attached (now outdated since the class definitions were moved) generator I had tried to use, and its output, if you're interested.
[generator_out.py.txt](https://github.com/user-attachments/files/16459671/generator_out.py.txt)
[temp_generator.py.txt](https://github.com/user-attachments/files/16459673/temp_generator.py.txt)

By changing channel properties, I meant something like  `instr.channels[1].active_traces = 3`, calling the setter of a property of the channel class.

pymeasure/instruments/agilent/agilentE5062A.py
BenediktBurger: Ah yes, the tests only find top level classes.

BenediktBurger:
> By changing channel properties, I meant something like instr.channels[1].active_traces = 3, calling the setter of a property of the channel class.

Ah, I guess I know, what the problem is: You access the list of channels (`channels`), which confuses the generator.
If you use the channel attributes (`ch_1`, `ch_2`, etc.) it should work as expected.

Would you mind to test it with `ch_1`, for example?
Then we can open an issue looking for a solution.

AlecVercruysse:
Apologies for the delay. The generator works great when I make that change (i.e., iterating through `channels = [inst.ch_1, inst.ch_2, inst.ch_3, inst.ch_4]`).

Do you prefer the protocol test to be auto-generated? The readability is much worse. For example, the class requires some communication upon initialization to set some default communication parameters and ask about what traces are currently activated (to dynamically add traces to the class). This is duplicated for every test case and results in a test file with thousands of lines of code:
```
def test_init():
    with expected_protocol(
            AgilentE5062A,
            [(b'CALC1:PARameter:COUNt?', b'+4\n'),
             (b'SOURce1:POWer:ATTenuation?', b'+0\n'),
             (b'CALC2:PARameter:COUNt?', b'+1\n'),
             (b'SOURce2:POWer:ATTenuation?', b'+0\n'),
             (b'CALC3:PARameter:COUNt?', b'+1\n'),
             (b'SOURce3:POWer:ATTenuation?', b'+0\n'),
             (b'CALC4:PARameter:COUNt?', b'+1\n'),
             (b'SOURce4:POWer:ATTenuation?', b'+0\n'),
             (b'FORMat:DATA REAL', None),
             (b'FORMat:BORDer SWAPped', None)],
    ):
        pass  # Verify the expected communication.
```
This is handled without code duplication in the hand-made protocol tests.


I'm also converting this PR to draft since I will change how strings are used to set properties. Right now if you set e.g. `inst.ch_x.trace_format = "MLOGarithmic"`, the getter returns just `"MLOG"`, which isn't how a property should work. If it makes sense, I'll change the setter to only accept the shortened string.

BenediktBurger:
> Do you prefer the protocol test to be auto-generated?

It does not matter how the tests are written. The Test Generator is meant as a tool to ease writing tests (especially with more complicated communication like binary one).
You can replace the default communication by a variable containing that communication, for example.

> Right now if you set e.g. inst.ch_x.trace_format = "MLOGarithmic", the getter returns just "MLOG", which isn't how a property should work. If it makes sense, I'll change the setter to only accept the shortened string.

You may use only the shortened versions or you might even consider using enums (be aware, that StrEnum is a recent python addition, but you might create it yourself with `class StrEnum(str, Enum)`).

Sionwage:
Hello, I just created [PR #1154](https://github.com/pymeasure/pymeasure/pull/1154) with some code that is probably similar to yours for the Keysight E5071C.

My code isn't complete and I have a ways to go on the unit testing but hopefully I will finish it soon.

We might be able to merge the best of both since the command structure is likely very close and utilize the `SERVice` class of SCPI commands with the `*OPT?` values to assign the specific model limits when initializing the instrument.

AlecVercruysse:
@BenediktBurger I finally ran tests on my (minor) changes and I think this PR is ready for review. I stuck with hand-written tests for readability and using the shortened strings rather than enums for simplicity.

Hi @Sionwage, It's a shame we seem to have duplicated some effort. This PR implements only E5062A with the S-param test option and extended power range, and the validation reflects that. For example, this instrument does not have the option for an 8-trace or channel display, like it seems you implement. Do you have a list of all the variations? I can't find any documentation about 8 channels in the Fifth Edition of the Agilent E5061A/E5062A ENA Series Programmer's Guide. 

Maybe it's useful if you take a look at this PR so we can agree on a public API such that future contributions of variations of this instrument don't require breaking changes to the existing API?

pymeasure/instruments/agilent/agilentE5062A.py
```diff
@@ -45,6 +45,362 @@
...
+            self._add_trace()
+
+    def _update_power_values(self, attenuation):
+        """Internal method that updates the valid range of stimulus power
+        based on the configured attenuation"""
+        self.power_values = (
+            -attenuation - 5,
+            -attenuation + 10)
+
+    def __init__(self, *args, **kwargs):
```
BenediktBurger, line 91: It's good to have the init at the top of the class, to see the configuration of the class easily.

pymeasure/instruments/agilent/agilentE5062A.py
```diff
@@ -45,6 +45,362 @@
...
+        super().__init__(*args, **kwargs)
+        # self.traces = []
+        self._add_trace(override_id=1)                # add just a single trace to init it all
+        self._update_trace_count(self.active_traces)  # add remaining traces if necessary
+        self._update_power_values(self.attenuation)   # update dynamic limits of power
+
+    start_frequency = Channel.control(
+        "SENSe{ch}:FREQuency:STARt?", "SENSe{ch}:FREQuency:STARt %g",
+        """Control the start frequency in Hz (float).""",
+        cast=lambda x: int(float(x)),
```
BenediktBurger, line 101: I guess, that the response contains something like `1E5`? Or why do you use `int(float(x))` instead of  `cast=int`?

pymeasure/instruments/agilent/agilentE5062A.py
```diff
@@ -45,6 +45,362 @@
...
+        validator=strict_range,
+        values=[3e5, 3e9]
+    )
+
+    stop_frequency = Channel.control(
+        "SENSe{ch}:FREQuency:STOP?", "SENSe{ch}:FREQuency:STOP %g",
+        """Control the stop frequency in Hz (float).""",
+        cast=lambda x: int(float(x)),
```
BenediktBurger, line 109: You state, that it is an float, but you return an integer?
I guess these are integers, but you might use a python float `1e5`, for example?

pymeasure/instruments/agilent/agilentE5062A.py
```diff
@@ -45,6 +45,362 @@
...
+        validator=strict_range,
+        values=[3e5, 3e9]
+    )
+
+    scan_points = Channel.control(
+        "SENSe{ch}:SWEep:POINts?", "SENSe{ch}:SWEep:POINts %g",
+        """Control the number of points used in a sweep (int). Valid range 2 - 1601.""",
+        cast=lambda x: int(float(x)),
+        validator=strict_range,
+        values=[2, 1601]
```
BenediktBurger, line 119: If you'd like to enforce, that they are integers, you could use `strict_set` and `range(2, 1602)`.

pymeasure/instruments/agilent/agilentE5062A.py
```diff
@@ -45,6 +45,362 @@
...
+        varies on config and the set value is truncated. Note that
+        ``sweep_time_auto`` needs to be ``False`` for changes to this
+        property to have an effect."""
+    )
+
+    sweep_time_auto = Channel.control(
+        "SENSe{ch}:SWEep:TIME:AUTO?", "SENSe{ch}:SWEep:TIME:AUTO %d",
+        """Control whether to automatically set the sweep time (bool). You
+        probably want this on to always keep the sweep time to a minimum (given
+        the range, IF BW, and sweep delay time).""",
```
BenediktBurger, line 134: For boolean properties names like `sweep_time_auto_enabled` are best, as they indicate directly, what True/False mean.

BenediktBurger (approved):
Thanks for the contribution.

A few comments.

If you, @Sionwage , could have a look regarding API, I'd really appreciate it.

pymeasure/instruments/agilent/agilentE5062A.py
AlecVercruysse: `strict_set` doesn't appear for me as an option [doc](https://pymeasure.readthedocs.io/en/latest/api/instruments/validators.html), so I tried `validator=functools.partial(strict_discrete_range, step=1)` instead

pymeasure/instruments/agilent/agilentE5062A.py
AlecVercruysse: Realizing that it's fine to leave this completely as a float.

pymeasure/instruments/agilent/agilentE5062A.py
AlecVercruysse: int appears to work fine

AlecVercruysse:
Thank you for the review! I've implemented your suggestions. I'm not sure if It's right that I resolve the conversations or if it's something you are supposed to do.

BenediktBurger:
The code looks good.

@Sionwage  does not respond.

What do you want to do: merge now and maybe create a unified version with Sionwage later on, or wait?

Sionwage:
Don't wait on my code as I can only work on it during lulls between projects. I still have issues I need to figure out for my use on how to activate additional windows layouts, channels, traces, and markers automatically with the various classes to avoid having to manually configure these instrument channels and ensure the commands go through.

I would say you should merge this and a unification will have to occur later. I think it would be best to maybe expand the documentation on expected vocabulary for controlling network analyzers, signal analyzers, power meters, etc to unify towards instruments that can be substituted without changing the commands.

In the past, I would have multiple types of each instrument but a factory environment wouldn't allow for using the same instrument each time so having interoperability would be nice for anyone dealing with the same.

BenediktBurger:
@AlecVercruysse , please tell me when you are ready, such that I merge this PR.

AlecVercruysse:
@BenediktBurger I think this is good to go, thanks!

PR #1154, opened 2024-08-19
Add Keysight E5071C VNA to the PyMeasure Builtin Instruments

Sionwage:
Working towards incorporating the Keysight E5071C VNA into PyMeasure's builtin instruments. These commands should be compatible with other Keysight VNA's such as the E5060, E5061, E5070, E5071, and the various revisions of each.

This PR is not complete yet and needs more work to finish in a few areas. Creating the PR to get ready for later integration.


BenediktBurger:
Hey @Sionwage , thanks for the pull request.

It is a large change (almost 2000 lines) and you think about including base classes for other VNAs etc.

If you find any logical separation for several pull requests, that would be great (e.g. the base class), as it reduces review effort.

PR #1168, opened 2024-10-04
Added LD400P electronic load

### LINKED ISSUE #1167: Add instrument for AimTTi LD400P
We have a LD400P that we'd like to automate. (See https://www.aimtti.com/product-category/electronic-loads/aim-ld400series.)

I'll be making a PR next, I figured I'd create this issue as a placeholder.

RobertoRoos:
Original code is made by Tim.

Resolves #1167

### DIFF
--- a/pymeasure/instruments/aimtti/ld400p.py
+++ b/pymeasure/instruments/aimtti/ld400p.py
@@ -23,10 +23,11 @@
 #
 
 from pymeasure.instruments import Instrument
+from pymeasure.instruments.generic_types import SCPIMixin
 from pymeasure.instruments.validators import strict_discrete_set
 
 
-class LD400P(Instrument):
+class LD400P(SCPIMixin, Instrument):
     """Interface for the LD400P electronic DC load from Aim-TTi.
 
     Note: the LD400 edition doesn't have digital interfacing, only the LD400P does.
@@ -50,15 +51,13 @@ class LD400P(Instrument):
             msg = msg.replace(suffix, "")
         return msg
 
-    def __init__(self, adapter, **kwargs):
-        super().__init__(adapter, "LD400", **kwargs)
-
-    id = Instrument.measurement("*IDN?", """Reads the instrument identification (string).""")
+    def __init__(self, adapter, name="LD400", **kwargs):
+        super().__init__(adapter, name, **kwargs)
 
     mode = Instrument.control(
         "MODE?",
         "MODE %s",
-        """The mode of the digital load. Can be one of:
+        """Control the mode of the digital load. Can be one of:
         ``("C", "P", "R", "G", "V")``, corresponding to (respectively) constant current,
         power, resistance, conductance or voltage (string).
         """,
@@ -70,7 +69,7 @@ class LD400P(Instrument):
     level_select = Instrument.control(
         "LVLSEL?",
         "LVLSEL %s",
-        """The selected level of the digital load. Can be one of:
+        """Control the selected level of the digital load. Can be one of:
         ``("A", "B", "T", "V", "E")``, corresponding to (respectively) Level A, Level B,
         Transient, Ext Voltage and Ext TTL (string).
         """,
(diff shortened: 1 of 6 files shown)


BenediktBurger:
You might want to add your name (or names) to the authors.txt file.

pymeasure/instruments/aimtti/ld400p.py
RobertoRoos, line 1: I still want to run the device test one more time, then it could be merged.

RobertoRoos: Test added, still looks good.

pymeasure/instruments/aimtti/ld400p.py
```diff
@@ -0,0 +1,126 @@
...
+        """Get the measured load current (float).""",
+        preprocess_reply=lambda v: v.replace("A", ""),
+    )
+
+    # Remove SCP properties that don't work on this device:
+    options = None
+    next_error = None
+
+    def check_errors(self):
+        raise NotImplementedError("This method is not available on this device")
```
RobertoRoos, line 126: @BenediktBurger I choose to un-define properties that don't work from `SCPIMixIn`, is that okay?
The alternative is to define the properties that do work, but to me that's less DRY.

pymeasure/instruments/aimtti/ld400p.py
```diff
@@ -0,0 +1,126 @@
...
+
+        The units of some replies depend on the mode of the device, hence all known
+        units are removed.
+        """
+        for suffix in cls.unit_suffixes:
+            msg = msg.replace(suffix, "")
+        return msg
+
+    def __init__(self, adapter, name="LD400", **kwargs):
+        super().__init__(adapter, name, **kwargs)
```
BenediktBurger, line 55: It is good to keep the init at the top of the class, to find (potential) configuration easily.

pymeasure/instruments/aimtti/ld400p.py
```diff
@@ -0,0 +1,126 @@
...
+
+    mode = Instrument.control(
+        "MODE?",
+        "MODE %s",
+        """Control the mode of the digital load. Can be one of:
+        ``("C", "P", "R", "G", "V")``, corresponding to (respectively) constant current,
+        power, resistance, conductance or voltage (string).
```
BenediktBurger, line 62:
```suggestion
        power, resistance, conductance, or voltage (string).
```
In English, you have a comma even before the "or" or "and" in an enumeration.

BenediktBurger (approved):
Small requests regarding documentation.

Thanks for your patience.

Again, you may add your name to the `authors.txt` file, if you wish so.

pymeasure/instruments/aimtti/ld400p.py
RobertoRoos: For some reason code like
```python
    def __init__(self, adapter, name="LD400", **kwargs):

        del self.options
```
keeps throwing the error `AttributeError: 'LD400P' object has no attribute 'options'`. Was this the code you had in mind?

BenediktBurger: Yes, that what, was I had in mind.

RobertoRoos: I just released my comment was ambiguous: the above doesn't work because an `AttributeError` is raised during _deletion_, not afterwards. So this snippet doesn't work. Do you have an example where it's used before?

BenediktBurger: I just did tests, confirming your findings.
So, yes, your raising "NotImplementedError" is fine.

RobertoRoos: I went with another option, to bring up a real `AttributeError`:
```python
    @property
    def options(self):
        raise AttributeError("No attribute 'options' exists")
```

BenediktBurger: You could add a suitable docstring (it has to start with set, get, measure, or control).
Here for example: "Get the options - does not exist for this device".

RobertoRoos: No, the errors are more complicated, I didn't see a quick way to fix it better: https://github.com/pymeasure/pymeasure/actions/runs/11418311223/job/31771599574#step:7:239

BenediktBurger: It tries to split the docstring. If you do not have a docstring (llike in your case), it is None, which is unsplittable.
Adding a docstring should fix it.

BenediktBurger:
Once you are ready, please tell me.

RobertoRoos:
Right, that's now. I think I'm all done!

RobertoRoos:
Thanks to you @BenediktBurger , for your tireless work for pymeasure.

PR #1169, opened 2024-10-07
Change 2182 read_termination to "\n"

panasee:
The read_termination of 2182a turns out to be "\n". 
Error messages can be avoided by changing the termination.

### DIFF
--- a/pymeasure/instruments/keithley/keithley2182.py
+++ b/pymeasure/instruments/keithley/keithley2182.py
@@ -179,7 +179,7 @@ class Keithley2182(SCPIMixin, KeithleyBuffer, Instrument):
     """
 
     def __init__(self, adapter, name="Keithley 2182 Nanovoltmeter",
-                 read_termination='\r', **kwargs):
+                 read_termination='\n', **kwargs):
         super().__init__(adapter, name, read_termination=read_termination, **kwargs)
 
     ch_1 = Instrument.ChannelCreator(Keithley2182Channel, 1)


BenediktBurger:
Actually, you can set the term char on the device, according to the [manual](https://download.tek.com/manual/2182A-900-01_May_2017.pdf).
Therefore, the driver offers the possibility to set the term_char at runtime: `Keithley2182(some_address, read_termination="\n")` will create an instrument driver for an instrument listening on "\n".

We can do this change, but I want to be sure, that it is what other users expect.

panasee:
Anyway to change it remotely instead of from the front panel? If it's only available on the front panel, then I think this parameter should just stick to the meter's default ("\n" in my case).
I referred to the LabView driver of 2182a (as the picture), line feed "\n" is used there. Maybe we should adapt to this configuration. 
image

Of course, this is the case with my meter. I cannot guarantee if this is the default setting for all 2182a meters. If you have more information about general settings, you could make the decision upon that.

BenediktBurger:
We cannot change the default termchar due to backwards compatibility.

Let's say some user uses the current code with a device set to `\r`.
After an update of pymeasure, the driver won't work anymore for that user, as the driver will use `\n`, but the device still uses `\r`.

I'm really sorry that I cannot incorporate your change, but I don't see any way to do such a change gracefully.

Therefore, you have either to change the device configuration (front panel) or call the python driver in this way: `Keithley2182(some_address, read_termination="\n") `.

Nevertheless, I thank you for seeing the issue and implementing a change.
I hope to see you again.

Issue #713, opened 2022-10-18
deprecated np.fromstring in Adapter.read_binary_values


BenediktBurger:
`Adapter.read_binary_values` raises
```DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead.```

Quote from [numpy documentation](https://numpy.org/doc/stable/reference/generated/numpy.fromstring.html) regarding `sep` parameter:

Deprecated since version 1.14: Passing sep='', the default, is deprecated since it will trigger the deprecated binary mode of this function. This mode interprets [string](https://docs.python.org/3/library/string.html#module-string) as binary bytes, rather than ASCII text with decimal numbers, an operation which is better spelt frombuffer(string, dtype, count). If [string](https://docs.python.org/3/library/string.html#module-string) contains unicode text, the binary mode of [fromstring](https://numpy.org/doc/stable/reference/generated/numpy.fromstring.html#numpy.fromstring) will first encode it into bytes using either utf-8 (python 3) or the default encoding (python 2), neither of which produce sane results.

Maybe Pyvisas implementation of "query_binary_values" might help.

bilderbuchi:
AFAICT, this breaks after numpy 2.1.2. 
Let's fix this before our next release.

PR #1173, opened 2024-10-16
Add Python 3.13 support.

BenediktBurger:


BenediktBurger:
Environment bumps are necessary.

BenediktBurger:
Note, especially Numpy 1.26.4, the last numpy 1 version, is **not** available for Python 3.13.

BenediktBurger:
Pyqt on conda-forge is version 5.15.9, which was updated last one year ago and is not built against python 3.13.
As long as that is not fixed, we cannot test python 3.13.
Most recent version on Pypi is 5.15.11.

For qt, it is the same.

bilderbuchi:
> Pyqt on conda-forge is version 5.15.9, which was updated last one year ago and is not built against python 3.13. As long as that is not fixed, we cannot test python 3.13. Most recent version on Pypi is 5.15.11.
> 
> For qt, it is the same.

If that is the only blocker, and for testing only, you can [specify ](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-file-manually)pip-installed dependencies in a conda yml file (presumably also with version bounds, only that example does not show them): 
```
  - pip
  - pip:
    - Flask-Testing
```

.github/pymeasure_python312.yml
```diff
@@ -3,24 +3,27 @@ channels:
...
   - setuptools_scm # don't pin, to get newest features
-  - sphinx=5.3.0
-  - sphinx_rtd_theme=1.2.2
+  - sphinx=8.1.3
+  - sphinx_rtd_theme=3.0.0
 #  pip is currently not needed, but the recommended tool when packages that are unavailable on conda are required
-#  - pip  # don't pin, to gain newest conda compatibility fixes
+  - pip  # don't pin, to gain newest conda compatibility fixes
+  - pip:
+    - pyqt5
```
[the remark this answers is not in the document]
BenediktBurger: I tried to make it work first and add versions if I succeed, but I'm encountering more issues (now for Python 3.12 and 3.13).
To be continued...

.github/pymeasure_python312.yml
```diff
@@ -3,24 +3,27 @@ channels:
...
+    - pint
```
bilderbuchi, line 29: what's wrong with pint? I can't check myself atm because of network issues.

bilderbuchi: weird, it seems like it should be available: https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fnoarch%2Fpint-0.24.3-pyhd8ed1ab_0.conda

.github/pymeasure_python312.yml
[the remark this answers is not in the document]
BenediktBurger: Now I pinned it again.

.github/pymeasure_python312.yml
BenediktBurger: Now it works.

bilderbuchi:
Now that we dropped py38, do we still need to to the workaround with `pymeasure_python312.yml`, or can we harmonize onto one environment file only? (cf #1051)

bilderbuchi:
> Ok, I can do that.
> 
> However, currently we use Numpy 1 for the general tests and a special test for Numpy 2.
> 
> Consolidating, should we use numpy 2 for all tests and a special numpy 1 test?
> 
> Or no numpy 1 test at all?

I'd say let's drop numpy 1., its security support period will end [really soon ](https://endoflife.date/numpy)anyway.

bilderbuchi:
> Either we keep two Environment with different Numpy versions (some hasse, but manageable) or we unpin Numpy Version.
> 
> I tend mildly to the latter. Due to the Numpy deprecation policy, we will test different Numpy versions, 

Neat solution! However, I'm nervous about allowing numpy version fluctuations in a test suite (as it's a pretty central package). Probably because at work I maintain a package where the (numerically compared) reference results change slightly even between Numpy minor versions... 😬 

OTOH, I don't think we have such sensitive tests in our codebase, right?

> which will Show errors due to changes in some Tests (Newer Python) only.

What do you mean, I don't understand?

bilderbuchi:
> And downgraded sphinx (to a version compatible with Python 3.9) does not work with the docs...

Isn't this just because of that warning
> WARNING: Calling get_html_theme_path is deprecated. If you are calling it to define html_theme_path, you are safe to remove that code.

which maybe stems from a very old template that was used to generate our spihinx config? Last touched 10 years ago https://github.com/pymeasure/pymeasure/blame/master/docs/conf.py#L29

BenediktBurger:
Now flake8 complains about code style, which are, however, strings...

BenediktBurger:
Now it is getting to work.
I limited the numpy version to the previously used one, such that the tests continue to pass, until #713 is fixed.

bilderbuchi:
`value_render = f"{value_render:.{self.float_digits:d}g}"` wow it did not occur to me to nest {} in f-strings like that. :-P, the inner one probably could be factored out into a variable. But a `noqa` is also fine with me to get this out the door.

PR #1174, opened 2024-10-17
Agilent E5062A: docs + rename `active_traces` to `visible_traces`

AlecVercruysse:
Sorry for not including this in the recently merged PR, but I was hoping to improve the docs and make a minor change to the API before a release containing this instrument goes out.

There might be some confusion between the concept of the "active trace" (the currently selected trace) and the `active_traces` parameter, which controls how many traces are visible. There is only ever one "active trace", so I renamed `active_traces` to `visible_traces`. I also cleaned up the docs a bit and added some example code, with a `wait_for_complete` helper method.


BenediktBurger:
The code looks good. Thanks for the clarification.

Do I understand it correctly, that the device is not yet included in any release?

AlecVercruysse:
Yes, the instrument was merged into main 2 weeks ago in PR #1176 so it is not in v0.14.

BenediktBurger:
Thanks!

PR #1180, opened 2024-10-25
Added an install group for building documentation

### LINKED ISSUE #1179: Pin development tools for the package
From what I see there is no quick way to get the additional Python packages installed to start developing `pymeasure`, like `pytest`, `black`, `flake8`, `sphinx`, etc. The errors are typically clear when required tools are missing, so installing them manually is no large concern. But then there is still the matter of versions.

A solution could be option requirement groups, such that when starting development the following installs everything to check, test and build docs:
```
pip install -e .[dev,test,doc]
``` 

Poetry also has optional groups for that, `poetry install --with dev`, but I don't know if we'd want to switch backends.

RobertoRoos:
Resolves #1179

### DIFF
--- a/docs/dev/contribute.rst
+++ b/docs/dev/contribute.rst
@@ -37,15 +37,20 @@ If you had already installed PyMeasure using :code:`pip`, make sure to uninstall
 
     pip uninstall pymeasure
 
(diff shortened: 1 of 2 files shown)


BenediktBurger:
Thanks.

PR #1182, opened 2024-10-28
Removed setup.cfg, replaced by pyproject.toml

### LINKED ISSUE #1178: Move from `setup.cfg` fully to `pyproject.toml`
A thought I was having: are we interested in replacing `setup.cfg` entirely by `pyproject.toml`? We already have the latter, but it's used for some tertiary config elements.

In my understanding `pyproject.toml` is the latest and preferred method of package definition.

I should add however that it looks like `setup.cfg` is not being deprecated soon though: https://stackoverflow.com/questions/44878600/is-setup-cfg-deprecated

RobertoRoos:
Resolves #1178


pyproject.toml
```diff
@@ -2,11 +2,70 @@
...
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "PyMeasure"
+# version is added dynamically
+authors = [
+    {name = "PyMeasure Developers"}
+]
+description = "Scientific measurement library for instruments, experiments, and live-plotting"
+# readme is merged dynamically from README and changelog
```
BenediktBurger, line 12: Move this line to the "dynamic" entry, please.

pyproject.toml
```diff
@@ -2,11 +2,70 @@
...
+    "pyqtgraph>=0.12",
+]
+dynamic = ["version", "readme"]
+
+[project.optional-dependencies]
+tcp = ["pyzmq>=16.0.2", "cloudpickle>=0.3.1"]
+python-vxi11 = ["python-vxi11>=0.9"]
+tests = ["pytest>=3.3.0", "pytest-cov>=4.1.0", "pytest-qt>=2.4.0", "pyvisa-sim>=0.4.0"]
+# install pyqt or pyside manually as desired
+docs = ["sphinx>=5.0.0", "sphinx_rtd_theme>=1.1.0"]
```
BenediktBurger, line 47: These dependencies should be one per line, in order to make version changes easier to spot in git.

BenediktBurger (approved):
A few small comments.

Thanks for implementing this change.

pyproject.toml
RobertoRoos: I disagree. The line `dynamic = ...` is very clear and doesn't need elaboration.

This is line is meant for a developer looking for a `readme = ...` entry. Instead of them incorrectly thinking we had forgotten it, with this comment I want to point them to the dynamic entry instead.

pyproject.toml
```diff
@@ -2,11 +2,70 @@
 requires = ["setuptools>=45", "wheel", "setuptools_scm>=8.1.0"]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "PyMeasure"
+# version is added dynamically
```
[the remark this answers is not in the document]
RobertoRoos: Yeah, that works for me. Also solves the other comment you raised.

BenediktBurger:
Thanks for your work.

PR #1185, opened 2024-10-29
Add has_next method to Procedure class

Did-Mas:
Pass information on whether the currently executing procedure is the last in the queue. The has_next method of the ExperimentQueue class is first monkeypatched to the Worker and then to the Procedure.

### DIFF
--- a/pymeasure/experiment/procedure.py
+++ b/pymeasure/experiment/procedure.py
@@ -284,8 +284,15 @@ class Procedure:
     def should_stop(self):
         raise NotImplementedError('should be monkey patched by a worker')
 
-    def has_next(self):
-        raise NotImplementedError('should be monkey patched by a worker')
+    def is_last(self) -> bool:
+        """Check if the procedure is the last one in the queue.
+
+        This method must be monkey patched by a worker to provide functionality.
+
+        Returns:
+            bool: True if the procedure is the last one in the queue, False otherwise.
+        """
+        raise NotImplementedError("should be monkey patched by a worker")
 
     def get_estimates(self):
         """ Function that returns estimates that are to be displayed by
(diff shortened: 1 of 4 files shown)


BenediktBurger:
Follow up to #1184

pymeasure/experiment/procedure.py
```diff
@@ -284,6 +284,9 @@ def emit(self, topic, record):
     def should_stop(self):
         raise NotImplementedError('should be monkey patched by a worker')
 
+    def has_next(self):
```
CasperSchippers, line 287: I realise this is also not done for the other monkeypatched methods, but it would be good to add a docstring to specify what the method does (or rather, what it returns) and thereby ensure (I presume at least) that the functionality/method is included in the documentation.

CasperSchippers, line 287: I do understand your doubt regarding the name. Maybe another/clearer name would help; something like `next_measurement_queued`?

If you think a negation would be the best option (which I can understand, as often some additional work is required only if the measurement is the last one) that is also fine. You could use either with a "private" intermediate method (e.g. `_has_next`) that is called by the actual `is_last`, or maybe you could monkeypatch with a lambda or local function in the worker (e.g. `self.procedure.is_last = lambda: not self.has_next()`) if that works.

I'm fine with either choice.

CasperSchippers (commented):
Nice, thanks for the PR!

Is this your first contribution to Pymeasure? In that case you can add your name to AUTHORS.txt.

pymeasure/experiment/procedure.py
Did-Mas: I added a docstring to the new function.

Did-Mas: "I decided on the name ```is_last``` as I believe it's the most intuitive, although it was a difficult choice. I implemented a lambda function at the ```BaseManager``` level, so both ```Worker``` and ```Procedure``` now have is_last methods.

CasperSchippers: Sounds good to me, nice work!

CasperSchippers (approved):
Sounds good to me, thanks for the nice work!

PR #1188, opened 2024-11-06
add "Reaveal in File Explorer" to browser item context menu

Did-Mas:


CasperSchippers (approved):
Looks good to me, thanks for the PR.

I've checked that it works on both linux (mint) and windows, so great job!

PR #1195, opened 2024-12-07
Fixed bug where setting waveform acquisition to get all points would …

fthouin:
…return none at all

### DIFF
--- a/pymeasure/instruments/siglenttechnologies/siglent_sds1072cml.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_sds1072cml.py
@@ -65,13 +65,24 @@ class VoltageChannel(Channel):
         self.write("C{ch}:WF? ALL")
         response = self.read_bytes(count=-1, break_on_termchar=True)
         descriptor_dictionary = self.get_descriptor()
-        rawWaveform = list(
-            struct.unpack_from(
-                "%db" % descriptor_dictionary["numDataPoints"],
-                response,
-                offset=descriptor_dictionary["descriptorOffset"],
-            ),
-        )
+        if descriptor_dictionary["numDataPoints"]==0:
+            rawWaveform = list(
+                struct.unpack_from(
+                    "%db" % descriptor_dictionary["pointsScreen"],
+                    response,
+                    offset=descriptor_dictionary["descriptorOffset"],
+                ),
+            )
+        else:
+            rawWaveform = list(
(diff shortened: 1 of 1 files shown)


pymeasure/instruments/siglenttechnologies/siglent_sds1072cml.py
```diff
@@ -65,13 +65,23 @@ def get_waveform(self):
...
-                response,
-                offset=descriptor_dictionary["descriptorOffset"],
-            ),
-        )
+        if descriptor_dictionary["numDataPoints"] == 0:
+            rawWaveform = list(
+                struct.unpack_from(
+                    "%db" % descriptor_dictionary["pointsScreen"],
+                    response,
+                    offset=descriptor_dictionary["descriptorOffset"],
```
BenediktBurger, line 73: If all data points are returned, should not the offset be 0?

pymeasure/instruments/siglenttechnologies/siglent_sds1072cml.py
```diff
@@ -65,13 +65,23 @@ def get_waveform(self):
...
+                ),
+            )
+        else:
+            rawWaveform = list(
+                struct.unpack_from(
+                    "%db" % descriptor_dictionary["numDataPoints"],
```
BenediktBurger, line 79: another idea to not repeat most of the code:
~~~python
data_points = descriptor_dictionary["numDataPoints"] if descriptor_dictionary["numDataPoints"] els escriptor_dictionary["pointsScreen"]
raw_waveform = list(... f"{data_points:d}b", response,...)
~~~

BenediktBurger (commented):
Thanks for your bug fix.

A suggestion for smaller code and a question regarding working code.

pymeasure/instruments/siglenttechnologies/siglent_sds1072cml.py
fthouin: The first bytes are a descriptor string and do not contain data. This is why they are skipped.

pymeasure/instruments/siglenttechnologies/siglent_sds1072cml.py
fthouin: Thanks for the suggestion. I implemented it in the code.

BenediktBurger:
Thanks for your work.

PR #1196, opened 2024-12-13
HP 66312A Dynamic measurement DC source

TEM01STAR:
Adding HP 66312A Dynamic measurement DC source. Intend to generalize for the 66332A once I have it working, although I only have the 66312A to test on.

### DIFF
--- a/pymeasure/instruments/hp/__init__.py
+++ b/pymeasure/instruments/hp/__init__.py
@@ -24,6 +24,7 @@
 
 from .hp33120A import HP33120A
 from .hp34401A import HP34401A
+from .hp66312A import HP66312A
 from .hp3478A import HP3478A
 from .hp3437A import HP3437A
 from .hp8116a import HP8116A
--- a/pymeasure/instruments/hp/hp66312A.py
+++ b/pymeasure/instruments/hp/hp66312A.py
@@ -1,0 +1,288 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2024 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+#work in progress by TEM01STAR
+from pymeasure.instruments import Instrument, SCPIUnknownMixin
+from pymeasure.instruments.validators import strict_discrete_set
+
+class HP66312A(SCPIUnknownMixin, Instrument):
+    """ Represents the HP / Agilent / Keysight 66312AA Multimeter and
+    provides a high-level interface for interacting with the instrument.
+
+    .. code-block:: python
+#garbage below, reminder to make example code
+        dmm = HP66312A("GPIB::1")
+        dmm.function_ = "DCV"
+        print(dmm.reading)  # -> Single float reading
+
+        dmm.nplc = 0.02
+        dmm.autozero_enabled = False
+        dmm.trigger_count = 100
+        dmm.trigger_delay = "MIN"
+        print(dmm.reading)  # -> Array of 100 very fast readings
+
+    """
+    FUNCTIONS = {"DCV": "VOLT", "DCV_RATIO": "VOLT:RAT",
+                 "ACV": "VOLT:AC", "DCI": "CURR", "ACI": "CURR:AC",
+                 "R2W": "RES", "R4W": "FRES", "FREQ": "FREQ",
+                 "PERIOD": "PER", "CONTINUITY": "CONT", "DIODE": "DIOD"}
+
+    FUNCTIONS_WITH_RANGE = {
+        "DCV": "VOLT", "ACV": "VOLT:AC", "DCI": "CURR",
+        "ACI": "CURR:AC", "R2W": "RES", "R4W": "FRES",
+        "FREQ": "FREQ", "PERIOD": "PER"}
+
+    BOOL_MAPPINGS = {True: 1, False: 0}
+
+    # Below: stop_bits: 20 comes from
+    # https://pyvisa.readthedocs.io/en/latest/api/constants.html#pyvisa.constants.StopBits
+    def __init__(self, adapter, name="HP 34401A", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            asrl={'baud_rate': 9600, 'data_bits': 8, 'parity': 0, 'stop_bits': 20},
+            **kwargs
+        )
+
+    #Source commands
+    output_enabled = Instrument.control(
+        "OUTP?", "OUTP %s",
+        """Enable[Disable] source output""",
+        validator=strict_discrete_set,
+        values=BOOL_MAPPINGS,
+        map_values=True,
+    )
+    
+    voltage_setpoint = Instrument.control(
+        "VOLT?","VOLT %s",
+        """Output voltage setpoint""",
+    )
+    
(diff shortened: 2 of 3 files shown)


pymeasure/instruments/hp/hp66312A.py
```diff
@@ -0,0 +1,288 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+#work in progress by TEM01STAR
+from pymeasure.instruments import Instrument, SCPIUnknownMixin
+from pymeasure.instruments.validators import strict_discrete_set
+
+class HP66312A(SCPIUnknownMixin, Instrument):
```
BenediktBurger, line 28: Does your device support the commands defined by the SCPIMixin (parent of the unknown mixin) ?
if so, please use SCPIMixin instead.
if not, please do not inherit the unknown mixin.

The unknown mixin is a crutch for us maintainers transitioning from default assuming Scpi compliance to explicitly defining it.
for older instruments we do not know, whether the device supports Scpi (having been assumed implicitly) or not (forgotten to define that it is not compliant).
no new instrument should inherit it.

TEM01STAR: Thank you for the clarification!

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -26,8 +26,8 @@
 from pymeasure.instruments.validators import strict_discrete_set
 
 
-class HP66312A(SCPIMixin, Instrument):
-    """ Represents the HP / Agilent / Keysight 66312AA Dynamic Measurement DC Source and
+class HP663X2A(SCPIMixin, Instrument):
+    """ Represents the HP / Agilent / Keysight 66312A/66332A Dynamic Measurement DC Source and
```
BenediktBurger, line 30: Better to rename to the docstring to "663X2A series", unless there are only these two devices.

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -221,3 +221,9 @@ def enable_cont_output_trigger(self):
     def write(self, command):
         """Write a command to the instrument."""
         super().write(command)
+
+class HP66312A(HP663X2A)
+    pass
+
+class HP66332A(HP663X2A)
```
BenediktBurger: The documentation (in docs folder) will only refer to the older device and won't show this one.

Please add it to the docs (and probably the parent device).

It these two devices are the only ones of the series, and the driver won't have any differences, the HP633X2A driver will be sufficient (with an appropriate docstring).

pymeasure/instruments/hp/__init__.py
BenediktBurger, line 1: For new devices, you have to add also the documentation, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation

pymeasure/instruments/hp/hp663X2A.py
BenediktBurger, line 1: Please add protocol tests, see the documentation: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#

There is also a test generator (see the same documentation) to make it easier to write tests based on communication with your device.

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+            adapter,
+            name,
+            asrl={'baud_rate': 9600, 'data_bits': 8, 'parity': 0, 'stop_bits': 20},
+            **kwargs
+        )
+
+    # Source commands
+    output_enabled = Instrument.control(
+        "OUTP?", "OUTP %s",
+        """Control Enable[Disable] the source output""",
```
BenediktBurger, line 59: Suggestion:
```suggestion
        """Control whether source output is enabled.""",
```

Also: Docstrings should end with a period (python convention).

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+    )
+
+    voltage_setpoint = Instrument.control(
+        "VOLT?", "VOLT %s",
+        """Control output voltage setpoint""",
+    )
+
+    voltage_trigger_level = Instrument.control(
+        "VOLT:TRIG?", "VOLT:TRIG %s",
+        """Control the pending triggered voltage level""",
```
BenediktBurger, line 72: What is the "pending triggered level"?

Maybe a user of that device will understand it.

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+        "MEAS:VOLT:MAX?",
+        """Measure the high voltage level from a sample.""",
+    )
+
+    # Output Trigger related commands
+    def init_single_output_trigger(self):
+        """Set the state of the triggering system to "wait-for-trigger".
+
+        The source will change the output levels to voltage_trigger_level and current_trigger_level.
+        It will return to Idle after a single trigger"""
```
BenediktBurger, line 146:
```suggestion
        It will return to 'Idle' after a single trigger."""
```

Suggestion to mark quotes with quote marks.

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+        self.write("INIT:NAME TRAN"),
+
+    def enable_cont_output_trigger(self):
+        """Set the state of the triggering system to "wait-for-trigger".
+
+        The source will change the output levels to voltage_trigger_level and current_trigger_level.
```
BenediktBurger, line 152: If you want to link to attributes, you can use this:
```suggestion
        The source will change the output levels to :attr:`voltage_trigger_level` and :attr:`current_trigger_level`.
```

~~~:meth:`name_of_method`~~~ refers to a method...

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+        It will return to Idle after a single trigger"""
+        self.write("INIT:CONT:NAME TRAN, ON"),
+
+    trigger_output = Instrument.setting(
+        "*TRG",
+        """Set the trigger once""",
+    )
```
BenediktBurger, line 159: Its an action, isn't it? So please make it a method instead of a setting, where you do not set anything.

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+        values=BOOL_MAPPINGS,
+        map_values=True,
+    )
+
+    display_mode = Instrument.control(
+        "DISP:MODE?", "DISP:MODE \"%s\"",
+        """Control the display, switch between normal and text display modes""",
+        validator=strict_discrete_set,
+        values=["NORM", "TEXT"],
+    )
```
BenediktBurger, line 175: Strings are always difficult.
As you have just two options, you could use the following: name the property `display_text_mode_enabled`, which switches between text and normal.
With `values={True: "TEXT", False: "NORM"}` and `map_values=True`, it will do the translation in the background.

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+        """Control whether remote control is enabled.""",
+        validator=strict_discrete_set,
+        values={True: "REM", False: "LOC"},
+        map_values=True,
+    )
+
+    error = Instrument.measurement(
+        "SYST:ERR?",
+        """Measure the next error in the buffer (FIFO)""",
+    )
```
BenediktBurger, line 198: As far as I know, the SCPIMixin defines this command already. (please check)

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+        "SYST: ", "SYST:%s",
+        """Control whether the beeper is enabled.""",
+        validator=strict_discrete_set,
+        values={True: "RWL", False: "LOC"},
+        map_values=True,
+    )
+
+    scpi_version = Instrument.measurement(
+        "SYST:VERS?",
+        """Measure the SCPI version of the source.""",
```
BenediktBurger, line 210: We allow "Get" as a synonym to "Measure", as it is sometimes more appropriate (like here).

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+    )
+
+    self_test_result = Instrument.measurement(
+        "*TST?",
+        """Measure a self-test of the source and return the result.
+
+        Be sure to set an appropriate connection timeout,
+        otherwise the command will fail.""",
+    )
```
BenediktBurger, line 219: Wouldn't a method be better: `self_test`?

pymeasure/instruments/hp/hp663X2A.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def write(self, command):
+        """Write a command to the instrument."""
+        super().write(command)
```
BenediktBurger, line 223: if you do not define a `write` method, it will automatically use the inherited one (which you call with `super()`), therefore, these lines are superfluous.

BenediktBurger (changes requested):
Thanks for your contribution.

A first round of comments.

Please verify, that your device differs from other similar devices.
Maybe there is already a very similar device, such that you can inherit from the other one.
See the [documentation ](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#instruments-with-similar-features)regarding dynamic properties for different limits on similar properties.

PR #1199, opened 2024-12-17
Fix documentation relating to Kepco BOP power supplies.

JAW90:
Fixed documentation for Kepco BOP Power supplies so that documentation is generated correctly.


BenediktBurger:
Thanks!

PR #1202, opened 2024-12-27
Phase control in tektronix AFG3152c

FlyingCurryMonster:
I added phase control for a Tektronix arbitrary waveform generator.  The get command always returns radians, but the user has the option to set the phase in radians or degrees.


pymeasure/instruments/tektronix/afg3152c.py
```diff
@@ -108,6 +112,21 @@ class AFG3152CChannel(Channel):
         values=FREQ_LIMIT,
     )
 
+    phase_rad = Instrument.control(
+        "phase:adjust?", "phase:adjust %e RAD",
+        """Control the phase in radians (float).""",
```
BenediktBurger, line 117: Maybe add the limits like "(float strictly between +-pi) "

BenediktBurger:
Sorry for the delay.

Would you mind adding tests for your new method?

PR #1209, opened 2025-01-05
add output_enabled property to individual channels (Keysight E3631A)

inonRAAAM:
The output_enabled property only existed for the Instrument class before, where it only showed (and controlled) the output status for the channel that was last used. This PR adds the ability to read and control the output status on every channel individually, regardless if it is the last used channel or not.

### DIFF
--- a/pymeasure/instruments/keysight/keysightE3631A.py
+++ b/pymeasure/instruments/keysight/keysightE3631A.py
@@ -113,7 +113,7 @@ class KeysightE3631A(SCPIMixin, Instrument):
     output_enabled = Instrument.control(
         "OUTPut?",
         "OUTPut %d",
-        """Control whether the channel output is enabled (boolean).""",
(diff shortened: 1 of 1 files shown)


BenediktBurger:
Thanks for your contribution. As a contributor you may add yourself to the authors.txt.

Your addition looks good.

Would it be good to modify the docstring from the generic property to make clear, that it is only the channel used last?

inonRAAAM:
Thanks for pointing that out, I forgot about the docstring. Changed it in a new commit.

BenediktBurger:
Thanks. Looks good.

You can add yourself to the authors.txt in this pull request, if you wish.
if you don't want to, please tell me and I'll merge the pull request right away.

inonRAAAM:
It's OK, you can merge :)

BenediktBurger:
Thanks for your work.

PR #1153, opened 2024-08-19
Add a `description` to `Parameter`s

SengerM:
I am new to pymeasure, but reading the documentation and source code I could not find a way to provide a description for each `Parameter` I add to my `Procedure`. I think it would be really nice to have an official `description` field in the constructor of `Parameter`s, so that the user can provide a human-friendly description with information that is not captured by the other fields or not obvious.

In this pull request I am simply suggesting that this would be a good idea and providing a trivial implementation. What exactly is done with the `description` afterwards I still don't know, as I am not experienced with pymeasure. However, I would guess it could probably be integrated with the graphical interface machinery and other things, instead as being mere documentation strings in the code.

Example:
```python
class ProcedureWithManyParameters(Procedure):
	repetition_rate = FloatParameter('repetition rate', units='s^-1', minimum=1e3, maximum=200e6, description='The inverse of the period between two write pulses.')
	read_fraction_start = FloatParameter('read fraction start', default=0.8, minimum=0, maximum=1, description='Fraction within each pulse at which to start the reading.')
	read_fraction_end = FloatParameter('read fraction end', default=1, minimum=0, maximum=1, description='Fraction within each pulse at which to stop the reading.')
```
as opposed to
```python
class ProcedureWithManyParameters(Procedure):
	"""A procedure with many parameters.
	
	Parameters
	----------
	repetition_rate: float
		The inverse of the period between two write pulses. units='s^-1', minimum=1e3, maximum=200e6.
	read_fraction_start: float
		Fraction within each pulse at which to start the reading. default=0.8, minimum=0, maximum=1.
	read_fraction_end: float
		Fraction within each pulse at which to stop the reading. default=1, minimum=0, maximum=1.
	"""
	
	repetition_rate = FloatParameter('repetition rate', units='s^-1', minimum=1e3, maximum=200e6)
	read_fraction_start = FloatParameter('read fraction start', default=0.8, minimum=0, maximum=1)
	read_fraction_end = FloatParameter('read fraction end', default=1, minimum=0, maximum=1)
```
which involves duplication between the code and the docstring.

Please let me know if this is already implemented somehow in Pymeasure, in that case I apologize for this.

PR #1211, opened 2025-01-09
Continue #1153: Add `description` to `Parameters`

Benbriel:
This PR continues the work from #1153 (as suggested by SengerM), introducing a new `description` argument to the `Parameter` class. The string is stored in the `_description` attribute. I’m unsure whether to keep it hidden or rename it as `description`, giving it more importance. If the latter is preferred, it should also be added to the `_help_fields` list attribute and the `__repr__` method in subclasses.

Additionally, tooltip support has been implemented. While keeping the PR simple, adding a default description that shows input limits could be considered.

image

Let me know if this should be rewritten, expanded upon, or entirely thrown away :)

### DIFF
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -61,10 +61,7 @@ class Input:
         if hasattr(parameter, 'units') and parameter.units:
             self.setSuffix(" %s" % parameter.units)
 
-        if hasattr(parameter, 'description'):
-            if not (description := parameter.description):
-                description = parameter._cli_help_fields()
-            self.setToolTip(description)
+        self.setToolTip(parameter._cli_help_fields())
 
     def update_parameter(self):
         """
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -61,6 +61,9 @@ class Input:
         if hasattr(parameter, 'units') and parameter.units:
             self.setSuffix(" %s" % parameter.units)
 
+        if hasattr(parameter, '_description') and parameter._description:
+            self.setToolTip(parameter._description)
+
     def update_parameter(self):
         """
         Update the parameter value with the Input GUI element's current value.
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -61,8 +61,10 @@ class Input:
         if hasattr(parameter, 'units') and parameter.units:
             self.setSuffix(" %s" % parameter.units)
 
-        if hasattr(parameter, 'description') and parameter.description:
-            self.setToolTip(parameter.description)
+        if hasattr(parameter, 'description'):
+            if not (description := parameter.description):
+                description = parameter._cli_help_fields()
+            self.setToolTip(description)
 
     def update_parameter(self):
         """
--- a/pymeasure/display/inputs.py
+++ b/pymeasure/display/inputs.py
@@ -61,8 +61,8 @@ class Input:
         if hasattr(parameter, 'units') and parameter.units:
             self.setSuffix(" %s" % parameter.units)
 
-        if hasattr(parameter, '_description') and parameter._description:
-            self.setToolTip(parameter._description)
+        if hasattr(parameter, 'description') and parameter.description:
+            self.setToolTip(parameter.description)
 
     def update_parameter(self):
         """
--- a/pymeasure/experiment/parameters.py
+++ b/pymeasure/experiment/parameters.py
@@ -42,8 +42,8 @@ class Parameter:
         list of strings, this argument can be either a single condition or
         a list of conditions. If the group_by argument is provided as a dict
         this argument is ignored.
-    :description: A string providing a human-friendly description for the 
-		parameter.
+    :description: A string providing a human-friendly description for the
+        parameter.
     """
 
     def __init__(self, name, default=None, ui_class=None, group_by=None, group_condition=True, description=None):
@@ -73,7 +73,7 @@ class Parameter:
         elif group_by is not None:
             raise TypeError("The provided group_by argument is not valid, should be either a "
                             "string, a list of strings, or a dict with {string: condition} pairs.")
-		
+
         if description is not None and not isinstance(description, str):
             raise TypeError("The provided description argument is not a string.")
         self._description = description
--- a/pymeasure/experiment/parameters.py
+++ b/pymeasure/experiment/parameters.py
@@ -133,7 +133,6 @@ class Parameter:
                 value = getattr(self, field[1])
                 message += ", {} {}".format(prefix, value)
 
-        message = message.replace("%", "%%")
         return message
 
     def __str__(self):
--- a/pymeasure/experiment/parameters.py
+++ b/pymeasure/experiment/parameters.py
@@ -59,7 +59,7 @@ class Parameter:
             self.value = default
         self.default = default
         self.ui_class = ui_class
-        self._help_fields = [('units are', 'units'), 'default']
+        self._help_fields = ['description', ('units are', 'units'), 'default']
 
         self.group_by = {}
         if isinstance(group_by, dict):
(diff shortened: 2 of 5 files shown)


pymeasure/experiment/parameters.py
```diff
@@ -72,6 +75,10 @@ def __init__(self, name, default=None, ui_class=None, group_by=None, group_condi
             raise TypeError("The provided group_by argument is not valid, should be either a "
                             "string, a list of strings, or a dict with {string: condition} pairs.")
 
+        if description is not None and not isinstance(description, str):
+            raise TypeError("The provided description argument is not a string.")
+        self._description = description
```
CasperSchippers, line 80: I think I would prefer `description` (i.e., without underscore), not so much as it it would give it more importance, but since the underscore signals it as a "private"-ish attribute (at least, by convention, technically there is of course no difference with a single underscore), and I think that description should be considered a "public" attribute.

Whether or not the description should also be added to ` _help_fields` and `__repr__` are (to me) a separate (from the name) questions: 

- Adding to `_help_fields` could be useful I think, as it extends the CLI docstring.
- I would refrain from adding it to `__repr__`, as I think it makes the repr to long and doesn´t seem to add much for debuging.

CasperSchippers (changes requested):
Thanks for this PR! See my thought on the attribute-name in the single comment.

I do like your suggestion of a default description to include the limits, but that could become dependend on the parameter type. One idea might be to use the `_help_fields`  to generate the tooltip: if you add the description to the `_help_fields`, maybe you can populate the tooltip with everything in there?

pymeasure/experiment/parameters.py
Benbriel: Would you agree that the best list order is `self._help_fields = ['description', ('units are', 'units'), 'default']`?

I believe `_help_fields` was intended for use with `ConsoleArgumentParser._cli_help_fields()`. It could be imported into  `inputs.py` to keep descriptions consistent between gui and cli:
```
if hasattr(parameter, 'description'):
    if not (description := parameter.description):
        description = ConsoleArgumentParser._cli_help_fields(
            parameter.name, parameter, parameter._help_fields
        )
    self.setToolTip(description)
```
I don't like the import, though. Maybe `_cli_help_fields` could be refactored into a `Parameter` method.

CasperSchippers: That sounds indeed as the correct order to me.

I agree that such an import is not very nice; I think your suggestion of making it a Parameter method is a very good one.

Benbriel:
I like that `_cli_help_fields` can now be reimplemented in a `Parameter` subclass. Also added extra kwargs in `ManagedConsole` to pass onto `ConsoleArgumentParser` for more flexibility.

pymeasure/experiment/parameters.py
```diff
@@ -115,6 +122,20 @@ def convert(self, value):
...
+        for field in self._help_fields:
+            if isinstance(field, str):
+                field = ["{} is".format(field), field]
+
+            if hasattr(self, field[1]) and getattr(self, field[1]) is not None:
+                prefix = field[0]
+                value = getattr(self, field[1])
+                message += ", {} {}".format(prefix, value)
+
+        message = message.replace("%", "%%")
```
CasperSchippers, line 136: Why this replacement?

pymeasure/experiment/parameters.py
```diff
@@ -115,6 +122,20 @@ def convert(self, value):
 
         return value
 
+    def _cli_help_fields(self):
+        message = self.name
+        for field in self._help_fields:
+            if isinstance(field, str):
+                field = ["{} is".format(field), field]
```
CasperSchippers, line 129: I´d suggest using f-strings
```suggestion
                field = [f"{field} is", field]
```

pymeasure/experiment/parameters.py
```diff
@@ -115,6 +122,20 @@ def convert(self, value):
...
+
+            if hasattr(self, field[1]) and getattr(self, field[1]) is not None:
+                prefix = field[0]
+                value = getattr(self, field[1])
+                message += ", {} {}".format(prefix, value)
```
CasperSchippers, line 134: Here I'd also suggest using f-strings:

```suggestion
                message += f", {prefix} {value}"
```

pymeasure/experiment/parameters.py
CasperSchippers, line 129: The only thing I like less is that the descript is also prefixed with "description is" in the CLI (e.g. `--delay DELAY         Delay Time, description is Delay between samples, units are s,....`). It might be nice if the description is added as something like `{name}: {description}` maybe (e.g. `Delay Time: Delay between samples`)

pymeasure/display/inputs.py
```diff
@@ -61,6 +61,11 @@ def set_parameter(self, parameter):
         if hasattr(parameter, 'units') and parameter.units:
             self.setSuffix(" %s" % parameter.units)
 
+        if hasattr(parameter, 'description'):
+            if not (description := parameter.description):
+                description = parameter._cli_help_fields()
+            self.setToolTip(description)
```
CasperSchippers, line 67: Maybe, I'd always using the cli_help_fields method, as that autmatically includes the minimum and maximum and units etc. which is kind of nice I think. In that case, maybe adding the different parts with a newline `\n` would be nice for the tooltip (such that it ends up in a multi-line tooltip, such as: 
```
{name}
{description}
Units are {},
Minimum is {},
Maximum is {},
 ```
This newline is automatically removed for the CLI anyways.

What do you think?

**Review by CasperSchippers — COMMENTED** [review]
Looks mostly good to me, just some minor remarks.

PS Sorry for the slow review.

File: pymeasure/experiment/parameters.py (hunk 2 above)
Comment (Benbriel) [reply pymeasure/experiment/parameters.py]: If not done, the ArgumentParser will get mad about a % sign not being escaped (for example, setting units as "%"). However, the Tooltip message would show "%%", so I moved that line to  `ConsoleArgumentParser.setup_parser` in [93fb581](https://github.com/pymeasure/pymeasure/pull/1211/commits/93fb581076699747d32f7cccccc052591411c120)

File: pymeasure/experiment/parameters.py (hunk 3 above)
Comment (Benbriel) [reply pymeasure/experiment/parameters.py]: I agree! Since `.format()` was predominantly used in the codebase, I looked for consistency.

Comment (Benbriel) [reply pymeasure/experiment/parameters.py]: Yeah, I didn't like that either. I decided to remove `"description"` from `_help_fields` and include it explicitly in `_cli_help_fields()`. Let me know what you think.

File: pymeasure/display/inputs.py (hunk 5 above)
Comment (Benbriel) [reply pymeasure/display/inputs.py]: Thank you for the review! I like the idea, plus it's better for long descriptions or parameters with long attributes. I changed the formatting a bit, let me know if you like it:

Tooltip:
```
{name}:
{description (+ "." if needed)}

Default is {}.
Units are {}.
[...]
```

CLI:
```
{name}: {description + "." if needed} Default is {}. Units are {}. [...]
```

Spaces between entries are correctly applied in both cases, and capitalization is done on the prefix.

This formatting could also be applied to `ConsoleArgumentParser._cli_help_fields()` for consistency.

pymeasure/experiment/parameters.py
CasperSchippers: Ah yeah fair enough; most of those are just there out of legacy I guess.

pymeasure/experiment/parameters.py
CasperSchippers: Ahh, check, didn't realize that.

pymeasure/experiment/parameters.py
CasperSchippers: Yeah, sounds good to me

pymeasure/display/inputs.py
CasperSchippers: Yes, I like that very much (including the appending of the "." if needed!!!).

CasperSchippers (approved):
This looks very nice! I think it is a very nice addition!

I think this can be merged. Let me know if you agree it can be merged.

CasperSchippers:
Oh, yeah, sure. That'd be nice for consistency.

Benbriel:
Also added wrapping in quotes if the value is a string. This could be useful for hinting the parameter type, plus it handles some edge cases (like a single dot being formatted as "Default is ..").

I think this PR is complete. Thanks for the reviews!

CasperSchippers (approved):

CasperSchippers:
Thanks for the work!

PR #1215, opened 2025-01-15
RigolDp932U driver and Unit test file

NIK7179:
Added Support to RigolDP932U Power Supply and Completed its Unit testing


pymeasure/instruments/rigol/Rigol_DP932U.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        Initialize the Rigol DP932U DC Power Supply Unit.
+
+        :param adapter: The communication adapter (e.g., USB or GPIB) to connect to the instrument.
+        :param kwargs: Additional arguments for instrument initialization.
+        """
+
+        if not adapter:
+            raise ValueError("Adapter cannot be None. Provide a valid communication adapter.")
+        kwargs.pop("name", None)
```
BenediktBurger, line 44: What is this line for? There cannot be an entry "name" I kwargs, as there is an explicit name parameter.

pymeasure/instruments/rigol/Rigol_DP932U.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def __init__(self, adapter, name="Rigol DP932U Power Supply", **kwargs):
+        """
+        Initialize the Rigol DP932U DC Power Supply Unit.
+
+        :param adapter: The communication adapter (e.g., USB or GPIB) to connect to the instrument.
+        :param kwargs: Additional arguments for instrument initialization.
+        """
+
+        if not adapter:
```
BenediktBurger, line 42: Why this check? Why should the adapter be None?

The super knit will complain if adapter is not defined properly.

pymeasure/instruments/rigol/Rigol_DP932U.py
```diff
@@ -0,0 +1,152 @@
...
+        """Control the current of the selected channel in Amps (0 to 3).
+
+        :param float value: Current level to set, within the range [0, 3].
+        :raises ValueError: If the current is outside the valid range.
+        """,
+        validator=strict_range,
+        values=[0, 3],
+    )
+
+    control_output_state = Instrument.control(
```
BenediktBurger, line 83: You can make it a boolean property, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html

In this case name it "output_enabled" such that usage is "output_enabled = True"

pymeasure/instruments/rigol/Rigol_DP932U.py
```diff
@@ -0,0 +1,152 @@
...
+        :param adapter: The communication adapter (e.g., USB or GPIB) to connect to the instrument.
+        :param kwargs: Additional arguments for instrument initialization.
+        """
+
+        if not adapter:
+            raise ValueError("Adapter cannot be None. Provide a valid communication adapter.")
+        kwargs.pop("name", None)
+        super().__init__(adapter, name, **kwargs)
+
+    control_channel = Instrument.control(
```
BenediktBurger, line 47: You do not have to add "Control" in front of a control, just name it "active_channel".

pymeasure/instruments/rigol/Rigol_DP932U.py
```diff
@@ -0,0 +1,152 @@
...
+        """Control the connection mode (OFF, PAR, or SER).
+
+        :param str value: Connection mode, either "OFF", "PAR" (parallel), or "SER" (series).
+        """,
+        validator=strict_discrete_set,
+        values={"OFF": "OFF", "PAR": "PAR", "SER": "SER", "PARALLEL": "PAR", "SERIES": "SER"},
+        map_values=True,
+    )
+
+    @check_error_decorator
```
BenediktBurger, line 107: Why don't you just use a pymeasure property "measurement" (like control, but without setting something)?
That will allow to call check_errors as well (see above linked documentation regarding properties).

BenediktBurger (changes requested):
Thanks for your contribution and welcome to pymeasure.

You may add yourself to the authors.txt, if you wish so.

Here a first round of comments.

pymeasure/instruments/rigol/Rigol_DP932U.py
NIK7179: You're correct that the if not adapter: check may be redundant since super().__init__(adapter, name, **kwargs) in the pymeasure library will handle validation and raise an error if the adapter is not properly defined.

However, this check was added as a defensive measure to provide a more specific and user-friendly error message ("Adapter cannot be None. Provide a valid communication adapter.") before delegating control to the superclass.

pymeasure/instruments/rigol/rigol_dp932u.py
```diff
@@ -0,0 +1,159 @@
...
+        """Control the connection mode (OFF, PAR, or SER).
+
+        :param str value: Connection mode, either "OFF", "PAR" (parallel), or "SER" (series).
+        """,
+        validator=strict_discrete_set,
+        values={"OFF": "OFF", "PAR": "PAR", "SER": "SER", "PARALLEL": "PAR", "SERIES": "SER"},
+        map_values=True,
+    )
+
+    measure_voltage = Instrument.measurement(
```
BenediktBurger, line 116: Here it should be `voltage` (and below `current`)

pymeasure/instruments/rigol/rigol_dp932u.py
```diff
@@ -0,0 +1,159 @@
...
+        ":MEASure:CURRent:DC?",
+        """Measure the current of the currently selected channel in Amps.
+
+        :return: Measured current in Amps (float).
+        """
+    )
+
+    def reset(self):
+        """
+        Resets the device to its factory defaults.
```
BenediktBurger, line 134: Docstrings are in imperative mode:
```suggestion
        Reset the device to its factory defaults.
```

pymeasure/instruments/rigol/rigol_dp932u.py
```diff
@@ -0,0 +1,159 @@
...
+        logging.info("Reset complete.")
+
+    def get_device_id(self):
+        """
+        Query the device identification string.
+
+        :return: Identification string (str), including manufacturer, model,
+            serial number, and firmware version.
+        """
+        return self.ask("*IDN?").strip()
```
BenediktBurger, line 147: For getting information, you should use a property.

By the way, this command is already implemented in the SCPIMixin: inst.id will result this value!. So you can remove this method.

pymeasure/instruments/rigol/rigol_dp932u.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Check for system errors.
+
+        :return: Error message (str) or "No error" if the system is operating correctly.
+        """
+        error = self.ask(":SYSTem:ERRor?").strip()
+        if error and "No error" not in error:
+            logging.error(f"System Error: {error}")
+            raise RuntimeError(f"System Error: {error}")
+        return error
```
BenediktBurger, line 159: This is also already implemented in SCPIMixin

tests/instruments/rigol/test_rigoldp932U.py
BenediktBurger, line 1: Thanks for adding tests.

BenediktBurger (changes requested):
A few more comments.

Sorry for the delay.

supervillain1010:
@BenediktBurger This module came from an intern at our company that is no longer with us and I'm not sure if he will address the requested changes since he does not have access to the hardware any more.   I am willing to pick this up (as well as the "Kepco" PR), but it's obviously up to you on how you want to handle it.  @NIK7179 Please chime in if I'm incorrect and you will still be working on this.

BenediktBurger:
I'm open for you taking over. There are still his commits, so authorship will be attributed accordingly.

NIK7179:
> I'm open for you taking over. There are still his commits, so authorship will be attributed accordingly.

NIK7179:
i will work on the above mentioned changes and commit them.

PR #1217, opened 2025-01-15
Keithley2600

NIK7179:
Added output_off_mode functionality


pymeasure/instruments/keithley/keithley_2600.py
```diff
@@ -82,7 +82,7 @@ def ask(self, cmd):
     def write(self, cmd):
         self.instrument.write(f'smu{self.channel}.{cmd}')
 
-    def values(self, cmd, **kwargs):
+    def values(self, cmd):
```
BenediktBurger, line 85: Why this change?

If we keep it, please change the docstring accordingly (currently it states, that it hands over the keyword arguments).

BenediktBurger:
I'm confused: The title states Keithley 2600,the text tells that a property is added (also in the files), but also a new instrument is added.

Please separate the two instruments, if possible, into two pull requests.
smaller pull requests (of logical units) are better to review.

PR #1229, opened 2025-01-22
Kepco

NIK7179:
Added Support to KepcoKlp75 Power Supply


pymeasure/instruments/kepco/kepcoklp75.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def __init__(self, adapter, name="Kepco KLP75", **kwargs):
+        """
+        Initialize the Kepco KLP75 instrument with the specified adapter and settings.
+
+        Parameters:
+            adapter: Communication adapter (e.g., GPIB address).
+            name (str): Name of the instrument.
+            kwargs: Additional keyword arguments for Instrument initialization.
+        """
```
BenediktBurger, line 66: Please move this information to the class docstring, as an __init__ docstring won't be visible typically. Besides, we are using sphinx formatting:
:param adapter: safasdfas
:param str name: Name of the instrument...

Actually, for this default information, you do not need to add anything.

pymeasure/instruments/kepco/kepcoklp75.py
```diff
@@ -0,0 +1,163 @@
...
+
+    current = Instrument.measurement(
+        "MEASure:CURRent?",
+        "Measure the current through the output terminals (A).",
+        cast=float,
+    )
+
+    voltage_setpoint = Instrument.control(
+        "SOURce:VOLTage?", "SOURce:VOLTage %.1f",
+        "Set or get the voltage setpoint (0-75 V).",
```
BenediktBurger, line 97: We use "Control" for "Set or get":
```suggestion
        "Control the voltage setpoint (0-75 V).",
```

pymeasure/instruments/kepco/kepcoklp75.py
```diff
@@ -0,0 +1,163 @@
...
+            ValueError: If the specified thresholds are out of valid range.
+        """
+        if ovp is not None:
+            if ovp < 0 or ovp > self._Vmax:
+                raise ValueError(f"OVP value {ovp} is out of range (0-{self._Vmax} V).")
+            self.write(f"SOURce:VOLTage:PROTection {ovp}")
+        if ocp is not None:
+            if ocp < self._Imin or ocp > self._Imax:
+                raise ValueError(f"OCP value {ocp} is out of range ({self._Imin}-{self._Imax} A).")
+            self.write(f"SOURce:CURRent:PROTection {ocp}")
```
BenediktBurger, line 127: Why don't you make two individual properties out of it?

If you'd like, you could make a method setting both properties to True.

pymeasure/instruments/kepco/kepcoklp75.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def status(self):
+        """
+        Query the instrument's status.
+
+        Returns:
+            str: The status message returned by the instrument.
+        """
+        return self.ask("STATus:QUEStionable?")
```
BenediktBurger, line 136: This could be a property as well, as it is not an action (doing), but getting a value.

pymeasure/instruments/kepco/kepcoklp75.py
```diff
@@ -0,0 +1,163 @@
...
+
+    confidence_test = Instrument.measurement(
```
BenediktBurger, line 138: As you return the result, you should name it `confidence_test_result`.
However, I guess this command will execute this test. Please make it a method: `execute_confidence_test` or something.

pymeasure/instruments/kepco/kepcoklp75.py
```diff
@@ -0,0 +1,163 @@
...
+        self.write("*CLS")
+
+    def get_id(self):
+        """
+        Get the instrument's identification string.
+
+        Returns:
+            str: The identification string of the instrument.
+        """
+        return self.ask("*IDN?")
```
BenediktBurger, line 163: These commands should be already defined in SCPIMixin (check it), so they are superfluous here.

BenediktBurger (changes requested):
Thanks for your contribution.

Here is a round of first comments.

PR #1239, opened 2025-02-13
Keithley2510

fg-lumentum:
Added new instrument, the Keithley 2510 TEC Sourcemeter. This device uses a thermistor and a peltier module to control temperature, which is the focus of the module.

All pretty standard of Keithley Sourcemeters, apart from:
- temperature_protection_range: Convenience property for getting and setting the low and high temperature protections as a tuple
- is_temperature_stable: Method for checking whether the temperature is stable
- wait_for_temperature_stable: Method for waiting until the temperature is stable

A unit test for the module has also been implemented

### DIFF
--- a/pymeasure/instruments/keithley/keithley2510.py
+++ b/pymeasure/instruments/keithley/keithley2510.py
@@ -22,9 +22,7 @@
 # THE SOFTWARE.
 #
 
-from time import sleep, time
-
-import numpy as np
+from time import time
 
 from pymeasure.instruments import Instrument, SCPIMixin
 from pymeasure.instruments.validators import strict_range, strict_discrete_set
--- a/pymeasure/instruments/keithley/keithley2510.py
+++ b/pymeasure/instruments/keithley/keithley2510.py
@@ -45,7 +45,7 @@ class Keithley2510(SCPIMixin, Instrument):
     source_enabled = Instrument.control(
         "OUTPut?",
         "OUTPut %d",
-        """Control whether the source is enabled. Takes values True or False.
+        """Control whether the source is enabled (bool).
         The convenience methods :meth:`~.Keithley2510.enable_source`
         and :meth:`~.Keithley2510.disable_source` can also be used.""",
         validator=strict_discrete_set,
@@ -197,11 +197,9 @@ class Keithley2510(SCPIMixin, Instrument):
             if temp_array[-1]:
                 mean_temp = np.mean(temp_array)
                 if np.all(abs(temp_array - mean_temp) < tolerance):
-                    print("Temperature stable!")
                     return True
 
             if should_stop():
-                print("Stopped waiting for temperature stability.")
                 return False
 
             sleep(delay)
--- a/pymeasure/instruments/keithley/keithley2510.py
+++ b/pymeasure/instruments/keithley/keithley2510.py
@@ -167,7 +167,6 @@ class Keithley2510(SCPIMixin, Instrument):
         self,
         tolerance=0.1,
         period=10,
-        points=64,
         should_stop=lambda: False,
         timeout=60,
     ):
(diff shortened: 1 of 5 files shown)


pymeasure/instruments/keithley/keithley2510.py
```diff
@@ -0,0 +1,209 @@
...
+        """Disables the source."""
+        self.write("OUTPUT OFF")
+
+    def enable_temperature_protection(self):
+        """Enable temperature protection."""
+        self.write(":SOURce:TEMPerature:PROTection:STATe ON")
+
+    def disable_temperature_protection(self):
+        """Disable temperature protection."""
+        self.write(":SOURce:TEMPerature:PROTection:STATe OFF")
```
BenediktBurger, line 146: Also here, why the duplication?

pymeasure/instruments/keithley/keithley2510.py
```diff
@@ -0,0 +1,209 @@
...
+    # Methods #
+    ###########
+
+    def enable_source(self):
+        """Enables the source."""
+        self.write("OUTPUT ON")
+
+    def disable_source(self):
+        """Disables the source."""
+        self.write("OUTPUT OFF")
```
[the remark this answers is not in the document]
fg-lumentum: They're included to align the 2510 to the other Keithley instruments, which use enable_source and disable_source as convenience methods

pymeasure/instruments/keithley/keithley2510.py
fg-lumentum: Again, to align it with other Keithleys

pymeasure/instruments/keithley/keithley2510.py
```diff
@@ -0,0 +1,209 @@
...
+        temp_array = points * [None]
+
+        t_start = time()
+
+        while time() - t_start < timeout:
+
+            temp_array.insert(0, self.temperature)
+            del temp_array[-1]
+
+            if temp_array[-1]:
```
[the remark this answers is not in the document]
fg-lumentum: I've gone for initializing an inf list, and a count modifying temp_array[count % points] on each loop.

BenediktBurger:
> All pretty standard of Keithley Sourcemeters

If it is a standard soucemeter. Did you check, whether you could inherit from another device?

fg-lumentum:
Just checked, and though it's similar to the Keithely 2400 there are many properties and methods in the 2400 that aren't in the 2510. It would be possible to create a mixin for Keithely SMUs, but that doesn't exist yet.

pymeasure/instruments/keithley/keithley2510.py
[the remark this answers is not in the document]
BenediktBurger: ok

pymeasure/instruments/keithley/keithley2510.py
```diff
@@ -0,0 +1,207 @@
...
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    temperature_setpoint = Instrument.control(
+        ":SOURce:TEMPerature?",
+        ":SOURce:TEMPerature %g",
+        """Control the temperature setpoint, in degrees centigrade.
+        Takes values from -50 to 225.""",
```
BenediktBurger, line 60: you can add `"(float truncated from -50 to 225)"` in line with the new docstring style. That way, you already indicate the used validator and the type (float or int).

pymeasure/instruments/keithley/keithley2510.py
```diff
@@ -0,0 +1,207 @@
...
+        validator=truncated_range,
```
BenediktBurger, line 61: We typically do not recommend truncated validators: The user might set the value to -60, believing, that it will be set to -60, as there is no error.

Typically, it is better to use the strict version, such that a user will notice, that the command did not execute as expected.

pymeasure/instruments/keithley/keithley2510.py
```diff
@@ -0,0 +1,207 @@
...
+        :return: True if stable, False otherwise.
+        """
+
+        delay = period / points
+
+        temp_array = []
+
+        for i in range(points):
+            temp_array.append(self.temperature)
+            sleep(delay)
```
BenediktBurger, line 165: another idea:
you compare each temperature measurement with the setpoint. If any value exceeds the temperature range, you return False.

BenediktBurger:
Thanks for your work!

PR #1089, opened 2024-04-11
Rework Agilent channels

BenediktBurger:
Agilent 4156 channels are subclasses of `Instrument`, (see #746 ), this PR makes them Channel subclasses and also improves some code.

PR #1240, opened 2025-02-15
Add HP4145x

neuschs:
Hi, I am about to start to write a driver for the HP4145A/B.

I have a specific issue, and I would like to have your opinion (maybe I am just too narrow minded right now).

The device has multiple channels, but they have different capabilites based on the instruments "mode".
I would like to assert / error out in case the the user uses capabilities that are not supported in this mode.
For this I would need access to the property of the  Instrument. 
This would be possible via `Channel.parent`, but not if I would like to use the convenient DynamicProperties.

### DIFF
--- a/pymeasure/instruments/hp/hp4145x.py
+++ b/pymeasure/instruments/hp/hp4145x.py
@@ -1,0 +1,233 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2025 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+import logging
+from enum import IntEnum
+
+from pymeasure.instruments import Instrument, Channel
+from pymeasure.instruments.validators import strict_discrete_set, truncated_discrete_set, strict_range
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+def mode_assert_lambda(v, parent, expected_mode):
+    assert parent.mode == expected_mode
+    return v
+
+
+def check_errors_user_mode_value(cmd_str):
+    data_status = cmd_str[0]
+    channel = cmd_str[1]
+    measurement_mode = cmd_str[2]
+    value = float(cmd_str[4:])
+
+    errors = {
+        'N': "Normal",
+        'L': "Interval is too short",
+        'V': "A-D converter saturation",
+        'X': "Oscillation",
+        'C': "This channel compliance error",
+        'T': "Other channel compliance error"
+    }
+
+    channels = {
+        'A': "SMU1",
+        'B': "SMU2",
+        'C': "SMU3",
+        'D': "SMU4",
+        'E': "VM1",
+        'F': "VM2"
+    }
+
+    if data_status != 'N':
+        log.error(f"HP4145x: Measurement Error - {data_status}, {errors[data_status]} "
+                  f"on channel {channels[channel]}")
+    return value
+
+
+class VS(Channel):
+    voltage = Channel.setting(
+        "DS {ch}, %f",
+        """
+        Controls the output voltage of on of the VS channels.
+
+        .. warning::
+            Only works with mode == 'USER_MODE' of instrument. In 'SYSTEM_MODE' no
+            direct access is allowed.
+        """,
+        validator=strict_range,
+        values=[0, 20],
+        # set_process=lambda v: mode_assert_lambda(v, super(), "USER_MODE")
+    )
+
+
+class VM(Channel):
+
+    def insert_id(self, command):
+        normal_id = self.id
+        self.id += 4
+        super().insert_id(command)
+        self.id = normal_id
+
+    voltage = Channel.measurement(
+        "TV {ch}",
+        """
+        Measures the voltage of one of the VM channels.
+
+        .. warning::
+            Only works with mode == 'USER_MODE' of instrument. In 'SYSTEM_MODE' no
+            direct access is allowed.
+        """,
+        # get_process=lambda v: mode_assert_lambda(v, super(), "USER_MODE")
+        get_process=check_errors_user_mode_value
+    )
+
+
+class SMU(Channel):
+    _voltage_range = 0
+    _voltage_ranges = {0: 0, 20: 1, 40: 2, 100: 3}
+    _current_range = 0
+    _current_ranges = {0: 0, 1e-9: 1, 10e-9: 2, 100e-9: 3, 1e-6: 4, 10e-6: 5, 100e-6: 6, 1e-3: 7, 10e-3: 8, 100e-3: 9}
+
+    _current_max = 100e-3
+    _current_compliance = 0.0
+
+    _voltage_max = 100
+    _voltage_compliance = 0.0
+
+    """
+    A SMU channel of the HP4145x Instrument
+    """
+    voltage = Channel.control(
+        "TV {ch}",
+        "DV {ch}, %d, %s, %f" % (_voltage_range, "%f", _current_compliance),
+        """
+        Controls the output voltage of on of the SMU channels.
+        Automatically puts the SMU into force voltage mode once set.
+        
+        .. warning::
+            Only works with mode == 'USER_MODE' of instrument. In 'SYSTEM_MODE' no
+            direct access is allowed.
+        """,
+        validator=strict_range,
+        values=[0, 100 if _voltage_range == 0 else _voltage_range],
+        dynamic=True,
+        # set_process=lambda v: modeAssertLambda(v, super(HP4145x), "USER_MODE"),
+        # get_process=lambda v: modeAssertLambda(v, super(HP4145x), "USER_MODE")
+        get_process=check_errors_user_mode_value
+    )
+
+    current = Channel.control(
+        "TI {ch}", "DI {ch}, %d, %s, %f" % (_current_range, "%f", _voltage_compliance),
+        """
+        Controls the output current of on of the SMU channels.
+        Automatically puts the SMU into force current mode once set.
+
+        .. warning::
+            Only works with mode == 'USER_MODE' of instrument. In 'SYSTEM_MODE' no
+            direct access is allowed.
+        """,
+        validator=strict_range,
+        values=[0, 100 if _voltage_range == 0 else _voltage_range],
+        dynamic=True,
+        # set_process=lambda v: modeAssertLambda(v, super(HP4145x), "USER_MODE"),
+        # get_process=lambda v: modeAssertLambda(v, super(HP4145x), "USER_MODE")
+        get_process=check_errors_user_mode_value
+    )
+
+    @property
+    def voltage_range(self):
+        for k, v in self._voltage_ranges.items():
+            if v == self._voltage_range:
+                return k
+
+    @voltage_range.setter
+    def voltage_range(self, value):
+        self._voltage_range = self._voltage_ranges[truncated_discrete_set(value, self._voltage_ranges)]
+        self.voltage_set_command = "DV {ch}, %d, %s, %f" % (self._voltage_range, "%f", self._current_compliance)
+
+    @property
+    def current_range(self):
+        for k, v in self._current_ranges.items():
+            if v == self._current_range:
+                return k
+
+    @current_range.setter
+    def current_range(self, value):
+        self._current_range = self._current_ranges[truncated_discrete_set(value, self._current_ranges)]
+        self.current_set_command = "DI {ch}, %d, %s, %f" % (self._current_range, "%f", self._voltage_compliance)
+
+    @property
+    def current_compliance(self):
+        return self._current_compliance
+
+    @current_compliance.setter
+    def current_compliance(self, value):
+        self._current_compliance = strict_range(value, [0, self._current_max])
+        self.voltage_set_command = "DV {ch}, %d, %s, %f" % (self._voltage_range, "%f", self._current_compliance)
+
+    @property
+    def voltage_compliance(self):
+        return self._voltage_compliance
+
+    @voltage_compliance.setter
+    def voltage_compliance(self, value):
+        self._voltage_compliance = strict_range(value, [0, self._voltage_max])
+        self.current_set_command = "DI {ch}, %d, %s, %f" % (self._current_range, "%f", self._voltage_compliance)
+
+
+class HP4145x(Instrument):
+    SMU1: SMU = Instrument.ChannelCreator(SMU, 1)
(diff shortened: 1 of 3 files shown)


BenediktBurger:
Typically, it is difficult to use these state dependent parts.
One idea is, to send the command to the device and let the device complain.

Maybe you could create a channel property which calls channel.parent?

neuschs:
Any objections to merge ?

BenediktBurger:
> The device has multiple channels, but they have different capabilites based on the instruments "mode".

Another idea: Just let the device itself complain, if a command is invalid and raise an Exception if you get an error message. For this we have check_get/set_errors parameter.

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+    manual_flush = False
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self._voltage_name = f"V{self.id}"
+        self._current_name = f"I{self.id}"
+
+    @property
+    def voltage(self):
```
BenediktBurger, line 313: Is it the actual voltage or the voltage setpoint? In the latter case, `voltage_setpoint` might be a better name.
Most of the time we use `setpoint` to distinguish the setpoint from the actual measurement.

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+            if self._channel_function == self._channel_functions['CONST']:
+                self._voltage = value
+                if not self.manual_flush:
+                    self.flush_source_setup()
+            else:
+                raise RuntimeError("In SYSTEM_MODE only allowed with channel_function = 'CONST'")
+        self.check_set_errors()
+
+    @property
+    def current(self):
```
BenediktBurger, line 351: see voltage

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+    def current_range(self, value):
+        self._current_range = self._current_ranges[truncated_discrete_set(
+            value, self._current_ranges)]
+
+    @property
+    def current_compliance(self):
+        """
+        Set the current compliance of the channel.
+
+        Returns not the actual state of the instrument but the state of the object.
```
BenediktBurger, line 438: it raises an Error and does not return anything!

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+
+        self._disabled = True
+        self.write("DE CH {ch}")
+
+    @property
+    def voltage_range(self):
+        """
+        Controls the voltage range used by :attr:`voltage`. Only applicable in `USER_MODE`.
+
+        Returns not the actual state of the instrument but the state of the object.
```
BenediktBurger, line 410: It returns nothing, but raises an error!

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+            raise ValueError(
+                "Channel can only be disabled, it gets enabled implicitly with the other functions")
+
+        self._disabled = True
+        self.write("DE CH {ch}")
+
+    @property
+    def voltage_range(self):
+        """
+        Controls the voltage range used by :attr:`voltage`. Only applicable in `USER_MODE`.
```
BenediktBurger, line 408:
```suggestion
        Set the voltage range used by :attr:`voltage`. Only applicable in `USER_MODE`.
```

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+        raise LookupError("Property can not be read.")
+
+    @disabled.setter
+    def disabled(self, value):
+        if not value:
+            raise ValueError(
+                "Channel can only be disabled, it gets enabled implicitly with the other functions")
+        self._disabled = True
+        self.write("DE VS {ch}")
+        self.check_set_errors()
```
BenediktBurger, line 155: In this case, I suggest to use a method instead, as it is an action:

~~~python
def disable(self)...
~~~

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+            if self._channel_function == self._channel_functions['CONST']:
+                if not self.manual_flush:
+                    self.flush_source_setup()
+            else:
+                raise RuntimeError("In SYSTEM_MODE only allowed with channel_function = 'CONST'")
+
+    @property
+    def voltage_compliance(self):
+        """
+        Set the voltage compliance of the channel.
```
BenediktBurger, line 457: what does "compliance" mean?

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+    manual_flush = False
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._voltage_name = f"VS{self.id}"
+
+    @property
+    def voltage(self):
+        """
+        Set the output voltage of one of the VS channels.
```
BenediktBurger, line 95: In which units? Volts?

It is always good to mention the units used.

tests/instruments/hp/test_hp4145x.py
BenediktBurger, line 1: Thanks for adding tests.

BenediktBurger (changes requested):
Thanks for your contribution.

I just answered your question, as it was a draft then.

Contrary to my response above, I do not see how to do it differently than with the plain properties you used.

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+
+    @property
+    def mode(self):
+        """
+        Control the mode of the instrument. Either 'USER_MODE' -
+        user controls manually all SMUs and voltage sources
+        or 'SYSTEM_MODE' where the analyzer gets configured to sweep
+        automatically the resources on it's own.
+        """
+        return self._mode
```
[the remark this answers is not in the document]
neuschs: I know this is a drawback of the design of this class, but there is no readout capability at all for the parameters on the HP4145. 
The only other alternativ is to get rid of all properties of these channels and just have two function calls per channel, that have all the properties values as parameters.
That makes using the class afterwards nasty.
I was in conflict with myself about this. But the limited interact ability with the device makes it hard.

pymeasure/instruments/hp/hp4145x.py
neuschs: It is a common term, an alternative is clamp.
If a voltage is forced, there is a certain current compliance and vice versa, as the SMU can reach an voltage of a certain load in Szenarios where the load is quite small and a high current would be needed.

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+        result = self.read().replace("\r", "").replace("\n", "")
+        removed_status = filter(None, result.replace("N", "").replace("C", "")
+                                .replace("T", "").replace("X", "").replace("L", "").split(","))
+
+        return [float(x) for x in removed_status]
+
+    hold_time = Instrument.setting(
+        "HT %f",
+        """
+        Set the wait time before sweep begins.
```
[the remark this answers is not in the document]
neuschs: See answer above. This device basically only allows to read the direct user mode parameters or the resulting trace data as well as srq byte. All other parameters are sadly write only

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+
+    def select_graphics_mode(self, mode, **kwargs):
+        """
+        Selects and configures the respective graphics mode.
+
+        For mode == 'MATRIX' provide parameter 'channel_name'.
+        """
+
+        if mode != "MATRIX":
+            raise ValueError("No graphics mode apart of matrix is supported")
```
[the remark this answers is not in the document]
neuschs: Theoretically the device supports multiple modes, I wanted to future proof the function for the other modes. Practically, they are not needed for most use, but they would need a ton of additional parameters and properties.
So I am just lazy

neuschs:
Although It doesn't perfectly fit, I would migrate to the Agent 4156 class design, if you merge your PR in before where you rework that.

The Agilent 4156 is fundamentally the successor to the HP 4145. They have like 90% functional overlap.
The 4156 has even a compatibility mode for the 4145. 

See
https://github.com/pymeasure/pymeasure/pull/1089

BenediktBurger:
I just merged it

BenediktBurger:
If they share a lot, could you not use parts (e.g. the channels) or reuse the code in another way?

If only the values and maybe a few methods differ, you could use dynamic properties and inheritance.

If the HP does only support a subset, maybe we could make something like:
~~~python
class HP:
    some_property = Agilent.some_property
    .
    .
    .
~~~

That copies only a subset of methods, but allows to reuse these methods without reinventing them.

neuschs:
> If they share a lot, could you not use parts (e.g. the channels) or reuse the code in another way?
> 
> If only the values and maybe a few methods differ, you could use dynamic properties and inheritance.
> 
> If the HP does only support a subset, maybe we could make something like:
> ~~~python
> class HP:
>     some_property = Agilent.some_property
>     .
>     .
>     .
> ~~~
> 
> That copies only a subset of methods, but allows to reuse these methods without reinventing them.


The issue is the actual command differs a lot only the use model and attributes are similar

BenediktBurger:
Ok, thanks for checking.
In this case, it is fine to create a new instrument.

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1116 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+import logging
+from enum import IntEnum
+
+from pymeasure.instruments import Instrument, Channel
+from pymeasure.instruments.validators import (strict_discrete_set, truncated_discrete_set,
```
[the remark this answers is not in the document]
BenediktBurger: What is your reasoning of using truncated validators?

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1252 @@
...
+    BUSY = 0b00010000
+    SELF_TEST_FAIL = 0b00100000
+    RQS = 0b01000000
+    EMERGENCY = 0b10000000
+    FIXTURE_LID_OPEN = 0b10000010
+    SMU_SHUT_DOWN = 0b10001000
+    POWER_FAILURE = 0b10100000
+
+
+class HP4145x(SCPIUnknownMixin, Instrument):
```
BenediktBurger, line 91: Does your device support SCPI commands? If so, please use the SCPIMixin. Otherwise, do not use any mixin.

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1252 @@
...
+        self.smu2 = SMU(self, 2)
+        self.smu3 = SMU(self, 3)
+        self.smu4 = SMU(self, 4)
+        self.vmu1 = VMU(self, 1)
+        self.vmu2 = VMU(self, 2)
+        self.vsu1 = VSU(self, 1)
+        self.vsu2 = VSU(self, 2)
+        self.var1 = VAR1(self)
+        self.var2 = VAR2(self)
+        self.vard = VARD(self)
```
BenediktBurger, line 199: See here how to add channels to an instrument: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html

That way, the channels are there already in the class (not only instance), and the channels are in a dictionary, which will make group actions (like the `manual_flush` property) easier.

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1252 @@
...
+    integration_time = Instrument.setting(
+        "SS %s",
+        """Set the integration time.
+
+        Values: :code:`SHORT` for no averages, :code:`MEDIUM` for 16 averages and :code:`LONG`
+        for 256 averages
+
+        .. code-block:: python
+
+            instr.integration_time = "MEDIUM"
```
BenediktBurger, line 261: Did you render the documentation locally (with `make html` in the docs folder) and look into it (in built folder)?

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1252 @@
...
+            new_settings_dict[obj_dict[key]] = value
+
+        for obj, setup in new_settings_dict.items():
+            for setting, value in setup.items():
+                setattr(obj, setting, value)
+                time.sleep(0.1)
+
+    def select_graphics_mode(self, mode, **kwargs):
+        """
+        Selects and configures the respective graphics mode.
```
BenediktBurger, line 396: Docstrings in imperative mode:
```suggestion
        Select and configure the respective graphics mode.
```

pymeasure/instruments/hp/hp4145x.py
BenediktBurger: You say, that you measure the voltage (i.e. read only), but you also have a setter, so do you measure it or do you control it?
If you control it, it is a setpoint, isn't it?

pymeasure/instruments/hp/hp4145x.py
BenediktBurger: That is a setpoint, as that is your desired current, right? It is not the current measured, is it?

BenediktBurger (changes requested):
A few more changes, please, but we are getting close.

Please do not resolve the conversations, such that I close them, once they are resolved.

pymeasure/instruments/hp/hp4145x.py
BenediktBurger: With use model you mean, that you have `voltage_setpoint = 5` and then you read the actual output voltage with `voltage == 5`?

We chose the setpoint version as for some instruments, `voltage` controls the setpoint (get and set), and sometimes it measures the actual voltage. For consistency and clarity in this library, we should use it here as well.

neuschs:
Will you check again ?

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1243 @@
...
+        name = new_name
+    return name.upper()
+
+
+##########
+# CHANNELS
+##########
+
+class HpMeasurementChannel(Channel):
+    """Offers some COMon properties."""
```
BenediktBurger, line 104: "COMon"?

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1243 @@
...
+        .. code-block:: python
+
+            instr.smu1.reset_settings()
+        """
+        self.write(f"DE {self._disable_command} {'{ch}'}")
+        self._disabled = True
+        self.check_set_errors()
+
+    def disable(self):
+        """Set the settings of this channel to default value.
```
BenediktBurger, line 166: would not reset be better to reset to default value instead of disable?

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1243 @@
...
+    def voltage_setpoint(self, value):
+        self._voltage_write_command = f"DV {'{ch}'}, 0, %f, {self.compliance:f}"
+        super().voltage_setpoint = value
+
+    @property
+    def current(self):
+        """
+        Measure the output current in 'USER_MODE'.
+        """
+        return super().current
```
BenediktBurger, line 453: properties get inherited, so I think this inherited method is superfluous.

pymeasure/instruments/hp/hp4145x.py
```diff
@@ -0,0 +1,1243 @@
...
+    def __init__(self, parent, id):
+        super().__init__(
+            parent,
+            id
+        )
+        self._steps = 1
+
+    def flush_source_setup(self):
+        """
+        Flushes the channel definition manually in case the :attr:`manual_flush` option is set.
```
BenediktBurger, line 754: Docstrings should be in imperative mode, that is: `Flush the channel`...

BenediktBurger (commented):
A few small comments.

Thanks for your patience.

pymeasure/instruments/hp/hp4145x.py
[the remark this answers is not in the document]
BenediktBurger: What about this point?

pymeasure/instruments/hp/hp4145x.py
neuschs: This was only to keep it similar to the Agilent successor.

BenediktBurger: Ok. I don't think we need to carry over this bad naming to this new device.
If someone uses the other device, they have to adjust their code.

Does the agilent have a reset method? If not, I would add it there and maybe even deprecate the `disable`. (In another PR)
Afterwards, both instruments behave in the same (better) way.

PR #1249, opened 2025-03-11
Fix typos in docs

ilev-sergey:


docs/dev/adding_instruments/properties.rst
```diff
@@ -92,7 +92,7 @@ In the examples below we assume you have imported the validators.
 
     from pymeasure.instruments.validators import strict_discrete_set, strict_range, truncated_range, truncated_discrete_set
 
-In many situations you will also need to process the return string in order to extract the wanted quantity or process a value before sending it to the device. The :func:`Instrument.control <pymeasure.instruments.common_base.CommonBase.control>`, :func:`Instrument.measurement <pymeasure.instruments.common_base.CommonBase.measurement>` and :func:`Instrument.setting <pymeasure.instruments.common_base.CommonBase.setting>` function also provide means to achieve this.
+In many situations you will also need to process the return string in order to extract the wanted quantity or process a value before sending it to the device. The :func:`Instrument.control <pymeasure.instruments.common_base.CommonBase.control>`, :func:`Instrument.measurement <pymeasure.instruments.common_base.CommonBase.measurement>` and :func:`Instrument.setting <pymeasure.instruments.common_base.CommonBase.setting>` function also provides means to achieve this.
```
BenediktBurger, line 95: In this case it is correct: These three functions provide means. However, there is another type:
```suggestion
In many situations you will also need to process the return string in order to extract the wanted quantity or process a value before sending it to the device.
The :func:`Instrument.control <pymeasure.instruments.common_base.CommonBase.control>`, :func:`Instrument.measurement <pymeasure.instruments.common_base.CommonBase.measurement>` and :func:`Instrument.setting <pymeasure.instruments.common_base.CommonBase.setting>` functions also provide means to achieve this.
```
Btw. In these files, you can write a single sentence per line, which is better for diffs.
Do this only for new files/touched lines.

docs/dev/adding_instruments/tests.rst
```diff
@@ -109,7 +109,7 @@ These adapter arguments are not written to tests.
 If you have arguments for the instrument itself, e.g. a RS485 address, you may give it as a keyword argument.
 These additional keyword arguments are included in the tests.
 
-Now we can use :code:`inst` as if it were created the normal way, i.e. :code:`inst = TC038(adapter)`, where ``adapter`` is some resource string.
```
BenediktBurger, line 112: That's correct English: as if it were (conjunctive instead of indicative).

BenediktBurger (changes requested):
Thanks for your suggested typos.

Here are a few comments.

docs/dev/adding_instruments/tests.rst
ilev-sergey: Agreed.

BenediktBurger:
Thanks!

PR #1250, opened 2025-03-11
Add Keysight 81160A Pulse Function Arbitrary Noise Generator

ilev-sergey:


docs/api/instruments/keysight/keysight81160A.rst
BenediktBurger, line 1: This file is still empty.

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,208 @@
...
+
+    limit_state = Instrument.control(
+        ":VOLT{ch}:LIM:STAT?",
+        ":VOLT{ch}:LIM:STAT %s",
+        """ Control the limit state (string).""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, "on": 1, "ON": 1, False: 0, "off": 0, "OFF": 0},
+        dynamic=True,
+    )
```
BenediktBurger, line 130: For such a boolean property, a name like `limit_state_enabled` is better, then you may just use boolean values (and not a combination of strings and bools).

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,208 @@
...
+LIMIT_HIGH_RANGE = [-9.99, 10]
+LIMIT_LOW_RANGE = [-10, 9.99]
+STATES = ["ON", "OFF"]
+IMPEDANCE_RANGE = [3e-1, 1e6]
+
+
+class Keysight81160AChannel(Agilent33500Channel):
+    shape_values = WF_SHAPES
+    shape_get_command = ":FUNC{ch}?"
+    shape_set_command = ":FUNC{ch} %s"
```
BenediktBurger, line 52: I'm unsure, whether it is better to write the properties anew or use their dynamic parts (as you do here).
In this case, you basically just copy the docstring and validator from the inherited class.

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,208 @@
...
+    pulse_period_values = PULSE_PERIOD_RANGE
+    pulse_period_get_command = ":PULS:PER?"
+    pulse_period_set_command = ":PULS:PER %e"
+    pulse_width_values = PULSE_WIDTH_RANGE
+    pulse_dutycycle_values = PULSE_DUTYCYCLE_RANGE
+    pulse_transition_values = PULSE_TRANSITION_RANGE
+    pulse_transition_set_command = ":FUNC:PULS:TRAN %e"
+    burst_period_values = BURST_PERIOD
+    burst_ncycles_values = BURST_NCYCLES
+    arb_file_get_command = ":FUNC:USER?"
```
BenediktBurger, line 208: As you just change the values (and a few commands), here it seems reasonable to inherit from the Agilent.

BenediktBurger (commented):
A few comments.

Thanks for opening the PR early on, and thanks for trying to inherit from other instruments, if possible.

ilev-sergey:
Seems like one of the existing tests doesn't always pass because of the randomness in the performance of CI
https://github.com/pymeasure/pymeasure/blob/bdae50f99ff523525189cf698d5045d6054d08d9/tests/instruments/temptronic/test_temptronic_base.py#L31-L39

I guess we should correct acceptable_factor to a lower value or remove/rework the test at all.

pymeasure/instruments/keysight/__init__.py
```diff
@@ -22,8 +22,9 @@
 # THE SOFTWARE.
 #
 
+from .keysight81160A import Keysight81160A
 from .keysightDSOX1102G import KeysightDSOX1102G
+from .keysightE3631A import KeysightE3631A
+from .keysightE36312A import KeysightE36312A
 from .keysightN5767A import KeysightN5767A
 from .keysightN7776C import KeysightN7776C
-from .keysightE36312A import KeysightE36312A
```
BenediktBurger, line 28: Thanks for fixing the order.

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,353 @@
...
+    )
+
+    coupling_enabled = Instrument.control(
+        ":TRAC:CHAN{ch}?",
+        ":TRAC:CHAN{ch} %s",
+        """Control the channel coupling (string). ``:TRAC:CHAN1 ON`` to copy values from
+        channel 1 to channel 2.""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, False: 0},
```
BenediktBurger, line 171: You use map values, so it is not string anymore, but boolean (for the user).

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,353 @@
...
+    burst_ncycles_set_command = ":BURS{ch}:NCYC %d"
+    burst_ncycles_get_command = ":BURS{ch}:NCYC?"
+
+    limit_state_enabled = Instrument.control(
+        ":VOLT{ch}:LIM:STAT?",
+        ":VOLT{ch}:LIM:STAT %s",
+        """Control the limit state (string).""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, False: 0},
```
BenediktBurger, line 136: Better would be a comment like `Control whether the state limit is enabled (bool).` As it is a boolean for the user.

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,353 @@
...
+        dynamic=True,
+    )
+
+    limit_high = Instrument.control(
+        ":VOLT{ch}:LIM:HIGH?",
+        ":VOLT{ch}:LIM:HIGH %f",
+        """Control the high-level voltage limit (float).""",
```
BenediktBurger, line 143: I guess in Volts?

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,353 @@
...
+    limit_low = Instrument.control(
+        ":VOLT{ch}:LIM:LOW?",
+        ":VOLT{ch}:LIM:LOW %f",
+        """Control the low-level voltage limit (float).""",
+        validator=strict_range,
+        values=LIMIT_LOW_RANGE,
+        dynamic=True,
+    )
+
+    memory_free = Instrument.measurement(
```
BenediktBurger, line 158: What about `free_memory_slots`?

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,353 @@
...
+    burst_ncycles_get_command = ":BURS{ch}:NCYC?"
+
+    limit_state_enabled = Instrument.control(
+        ":VOLT{ch}:LIM:STAT?",
+        ":VOLT{ch}:LIM:STAT %s",
+        """Control the limit state (string).""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, False: 0},
+        dynamic=True,
```
BenediktBurger, line 137: Why do you make everything dynamic?

If you do not need it, don't make it dynamic.

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,353 @@
...
+        return data_int
+
+    @property
+    def waveform_volatile(self):
+        """
+        Get volatile waveform data.
+
+        :return: waveform data (array-like).
+        """
+        return self._waveform_volatile
```
BenediktBurger, line 222: Warning: This property will not return anything meaningful until you have set it previously.
Always be careful, if you duplicate a state in the driver code as it might get out of sync with the actual device configuration.
For example, if you power cycle the device, the waveform is lost, but this variable still contains it. Or on the other side, you restart the python driver and do not return anything, while the device still has a waveform.
Or someone else changes the waveform (via another connection or on the front panel...).

At least you should warn, that it might not reflect the current device state.
And think about a return value without having set it. You could use `getattr` with a default value of None or you could set `self._waveform_volatile = None` at startup.
Or you state, that it raises a NameNotFound exception if it has not been set, such that a user can catch that exception appropriately.

pymeasure/instruments/keysight/keysight81160A.py
```diff
@@ -0,0 +1,353 @@
...
+            datatype="h",
+            is_big_endian=True,
+        )
+
+    def save_waveform(self, waveform, name):
+        """
+        Save a waveform to the generator's nonvolatile memory.
+
+        :param waveform: The waveform data.
+        :param name: The name of the waveform.
```
BenediktBurger, line 245: That is the name to identify it in the memory, right?

tests/instruments/keysight/test_keysight81160A_with_device.py
```diff
@@ -0,0 +1,508 @@
...
+    SQUARE_DUTYCYCLE_RANGE,
+    TRIGGER_COUNT,
+    TRIGGER_MODES,
+    VOLTAGE_HIGH_RANGE,
+    VOLTAGE_LOW_RANGE,
+    WF_SHAPES,
+    Keysight81160A,
+)
+
+pytest.skip("Only works with connected hardware", allow_module_level=True)
```
BenediktBurger, line 51: Not needed, as the `connected_device_address` fixture already skips the test unless you call it explicitly with and with a device address.
With that line, you will never run this test file!

See https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#device-tests

BenediktBurger (changes requested):
A few comments with some changes. Thanks for your work!

pymeasure/instruments/keysight/keysight81160A.py
BenediktBurger: Thanks for the comment.
I don't think, that you need to add the log, though.

BenediktBurger (approved):
A small comment is left, besides, it looks good. Thank you!

ilev-sergey:
Removed the log. Thanks for your review!

BenediktBurger:
Thanks for your work!

PR #1252, opened 2025-03-12
Add support for emitting sequences of values

jvdoorn:
Whilst writing a procedure to handle chunks of data (examples include snapshots from an oscilloscope or CCD but also data acquisition devices that send data in chunks), I ran into performance issues. Currently you have to emit each data point individually. For example:
```python
def execute(self):
    for _ in range(self.chunks):
        x = np.ones(self.chunk_size)  # Fictitious source of data, can be an oscilloscope snapshot for example.
        y = np.ones(self.chunk_size)
        
        for i in range(self.chunk_size):
            self.emit("results", {
                "x": x[i],
                "y": y[i]
            })
```

For large chunks of data this takes a very long time to process. Which can present issues for the data acquisition (full buffers). A major speedup can be achieved by instead emitting the whole array at once (reducing multiple TCP calls to a single call). I made a small edit to the `emit(self, record)` method of the `Worker` class such that it can handle arrays of data. This then allows one to do the following:

```python
def execute(self):
    for _ in range(self.chunks):
        x = np.ones(self.chunk_size)  # Fictitious source of data, can be an oscilloscope snapshot for example.
        y = np.ones(self.chunk_size)
        
        self.emit("results", {
            "x": x,
            "y": y
        })
```

I wrote a small script to demonstrate this and to show that it works based on the random procedure example (see below). Emitting points one by one also still works. You can check the procedure log to compare processing times. My approach is a (temporary) alternative to the changes proposed in https://github.com/pymeasure/pymeasure/pull/277 (which appears to be stale). 

```python
import logging
from itertools import count
from time import time

import numpy as np

import sys
import random
from pymeasure.experiment import BooleanParameter, Procedure
from pymeasure.experiment import IntegerParameter, Parameter

    runs = IntegerParameter('Number of Runs', default=1)
    iterations = IntegerParameter('Loop Iterations', default=100)
    use_batches = BooleanParameter('Use Batches', default=False)

    DATA_COLUMNS = ['Index', 'Random Number']

    def startup(self):
        random.seed(self.seed)

    def execute(self):
        for _ in range(self.chunks):
            x = np.ones(self.chunk_size)
            y = np.ones(self.chunk_size)

            for i in range(self.chunk_size):
                self.emit("results", {
                    "x": x[i],
                    "y": y[i]
                })

    def execute(self):
        start_time = time()

        index_counter = count()
        for j in range(self.runs):
            indices = np.fromiter([next(index_counter) for _ in range(self.iterations)], dtype=int)
            values = np.fromiter([random.random() for _ in range(self.iterations)], dtype=float)

            if not self.use_batches:
                    data = {
                        "Index": indices[i],
                        "Random Number": values[i]
                    }
                    self.emit('results', data)

            if self.should_stop():
                break

            if self.use_batches:
                data = {
                    "Index": indices,
                    "Random Number": values
                }
                self.emit('results', data)
            self.emit('progress', 100 * (j + 1) / self.runs)

        end_time = time()
        log.info(f"Average time per data point: {(end_time - start_time) / (self.iterations * self.runs) * 1E6:.2f} µs")


    def __init__(self):
        super().__init__(
            inputs=['runs', 'iterations', 'use_batches', 'seed'],
            displays=['runs', 'iterations', 'use_batches', 'seed'],
            x_axis='Index',
            y_axis='Random Number'
        )


if __name__ == "__main__":
    window = MainWindow()
    window.show()
    sys.exit(app.exec())
```

### DIFF
--- a/pymeasure/experiment/workers.py
+++ b/pymeasure/experiment/workers.py
@@ -26,6 +26,9 @@ import logging
 import time
 import traceback
 from queue import Queue
+from typing import Iterable, Sequence
+
+import numpy as np
 
 from .listeners import Recorder
 from .procedure import Procedure
@@ -115,9 +118,32 @@ class Worker(StoppableThread):
         except (NameError, AttributeError):
             pass  # No dumps defined
         if topic == 'results':
-            self.recorder.handle(record)
+            if self._should_record_be_iterated(record):
+                log.debug("Iterating results before handling")
+
+                lengths = [len(value) for value in record.values()]
+                if not all(length == lengths[0] for length in lengths):
+                    log.warning(
+                        f"Potential data loss: not all data columns have the same length (check your emitted results)"
+                    )
+
+                for index in range(lengths[0]):
+                    data = {key: value[index] for key, value in record.items()}
+                    self.recorder.handle(data)
+            else:
+                self.recorder.handle(record)
         elif topic == 'status' or topic == 'progress':
             self.monitor_queue.put((topic, record))
+
+    def _should_record_be_iterated(self, record) -> bool:
+        """
+        A record (dictionary) should be iterated if all values are sequences (lists, tuples, numpy arrays, etc.).
+        """
+        sequence_types = (Sequence, np.ndarray)
+        type_exceptions = (str, bytes)
+        return all(
+            isinstance(value, sequence_types) and not isinstance(value, type_exceptions) for value in record.values()
+        )
 
     def handle_abort(self):
         log.exception("User stopped Worker execution prematurely")


CasperSchippers:
Hey,

Thanks for the PR. This is indeed a recurring theme which absolutely should be fixed. #227 is indeed stale alas (and by now significantly outdated).

I'll have a look at this PR on Monday/Tuesday.

CasperSchippers:
When I wanted to start looking at the patch, I just now noted that your branch / PR is based on the develop branch. This branch is (significantly) outdated (maybe we ought to archive this somewhere :thinking:). I can apparently change the PR target to master, but although I guess that merging the PR will work as intended (i.e., only your patch) I think it is better to branch it off master instead of develop. This might need a new PR (or maybe you could rebase and force-push it, but I'll leave that up to you)

CasperSchippers (changes requested):
I have to admit that I find the simplicity and effectiveness quite striking.

I think making it a bit more explicit could be nice, but then I think I'm happy to include it; I do have two options/suggestions and wonder what your thoughts are?

1. Would it be a good idea to make a different "topic" (such as `"batch results"` such that you can `self.emit("batch results",  data)`), to make it more explicit? In that case I think we should somehow handle unequal lenght series (i.e., what you warn for in line 127).
2. Another option could be to choose handling based on wheter you emit a dict (for a single datapoint) or a pandas dataframe (for a series); this also ensures that all entries are equally long for the series option.

What do you think?

Finally, there are some linting issues found by the CI that´d need to be resolved.

BenediktBurger:
Hey @jvdoorn , thanks for your initiative. By sending a pull request you become a contributor. If you want to, you may add yourself to the authors.txt file (also in this PR).

jvdoorn:
Superseded this PR by #1256, will close this one in favor of it. Also goes into your feedback.

PR #1253, opened 2025-03-13
Santec tsl570

fg-lumentum:
Added the Santec TSL-570 Tunable Laser. Provides control over optical power, wavelength, optical-frequency, and sweep settings.

Valid values for wavelength and optical frequency are dependent on device model and settings, so instead of using pymeasure validators properties relating to the control of wavelength and optical-frequency query the device for errors after setting using check_errors from the SCPI mixin. If an error is found, a custom InstrumentError is raised, e.g. "InstrumentError: -222: Data out of range". For testing purposes this feature is patched out as it relies on having access to the instrument.

Properties to split sweep_mode into sweep_pattern and sweep_routing have been created to control these two separate aspects of sweep_mode independently.

### DIFF
--- a/pymeasure/instruments/santec/tsl570.py
+++ b/pymeasure/instruments/santec/tsl570.py
@@ -24,7 +24,48 @@
 
 from pymeasure.instruments import Instrument, SCPIMixin
 
+from pymeasure.instruments.validators import strict_discrete_set
+
 
 class TSL570(SCPIMixin, Instrument):
     """Represents the Santec TSL-570 Tunable Laser and provides a high-level interface for
     interacting with the instrument."""
+
+    def __init__(self):
+        """Set the device to use SCPI commands."""
+        Instrument.write(self, ":SYSTem:COMMunicate:CODe 1")
+
+    shutter_closed = Instrument.control(
+        ":POWer:SHUTter?",
+        ":POWer:SHUTter %d",
+        """A boolean property that controls whether shutter is closed.""",
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    wavelength = Instrument.control(
+        ":WAVelength?", ":WAVelength %e", """Set the output wavelength, in m."""
+    )
+
+    frequency = Instrument.control(
+        ":FREQuency?", "FREQuency %e", """Set the output wavelength in optical frequency, in Hz."""
+    )
+
+    # TODO
+    # power_setpoint
+    # power_reading
+    # power_unit
+    # wavelength_start
+    # wavelength_stop
+    # wavelength_step
+    # frequency_start
+    # frequency_stop
+    # frequency_step
+    # sweep_mode
+    # sweep_speed
+    # sweep_dwell
+    # sweep_delay
+    # single_sweep
+    # repeat_sweep
+    # sweep_status
--- a/pymeasure/instruments/santec/tsl570.py
+++ b/pymeasure/instruments/santec/tsl570.py
@@ -37,15 +37,24 @@ class TSL570(SCPIMixin, Instrument):
 
     def __init__(self, adapter, name="Santec TSL-570", **kwargs):
         super().__init__(adapter, name, **kwargs)
-        self.write(":SYSTem:COMMunicate:CODe 1")  # Set the device to use SCPI commands
         self.check_errors()  # Clear the error queue
 
     def check_instr_errors(self):
-        """Checks the instrument for errors, and raises an exeption if any are present."""
+        """Check the instrument for errors, and raise an exception if any are present."""
         errors = self.check_errors()
         if errors:
             raise Error(errors)
 
+    command_set = Instrument.control(
+        ":SYSTem:COMMunicate:CODe?",
+        ":SYSTem:COMMunicate:CODe %d",
+        """Control the command set (str Legacy of SCPI).
+        Legacy commands use units of nm and THz for wavelength and optical frequency respectively.
+        SCPI commands use units of m and Hz for wavelength and optical frequency respectively.""",
+        validator=strict_discrete_set,
+        values={"Legacy": 0, "SCPI": 1},
+        map_values=True,
+    )
     # --- Optical power control ---
 
     output_enabled = Instrument.control(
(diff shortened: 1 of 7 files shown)


fg-lumentum:
Python 3.10, windows-latest seems to be throwing "Error: Could not find micromamba activate pymeasure in C:\Users\runneradmin\.bash_profile", is this an issue on my end or yours?

BenediktBurger:
> Python 3.10, windows-latest seems to be throwing "Error: Could not find micromamba activate pymeasure in C:\Users\runneradmin.bash_profile", is this an issue on my end or yours?

I do not know, why the CI gets cancelled, maybe it is your `InstrumentException`.

Would you mind also to merge the most recent master into your branch? Just to make sure, that you are up to date.

fg-lumentum:
Hmmm, still confused

BenediktBurger:
Do the tests work locally on your machine (just execute `pytest`)?

fg-lumentum:
Found the problem, test_all_instruments was timing out because there was communication in the init of the new device

pymeasure/instruments/santec/tsl570.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class TSL570(SCPIMixin, Instrument):
+    """Represents the Santec TSL-570 Tunable Laser and provides a high-level interface for
+    interacting with the instrument."""
+
+    def __init__(self, adapter, name="Santec TSL-570", **kwargs):
+        super().__init__(adapter, name, **kwargs)
+        self.write(":SYSTem:COMMunicate:CODe 1")  # Set the device to use SCPI commands
+        self.check_errors()  # Clear the error queue
```
BenediktBurger, line 41: Consider adding a method "activate_SCPI_commands" with that command.
And then you could move that out of the init, but state in the docstring, that you might need to set it at least once (or on the front panel).

pymeasure/instruments/santec/tsl570.py
```diff
@@ -0,0 +1,263 @@
...
+        validator=strict_discrete_set,
+        values={"dBm": 0, "mW": 1},
+        map_values=True,
+    )
+
+    power_setpoint = Instrument.control(
+        ":POWer?",
+        ":Power %e",
+        """Control the output optical power, units defined by power_unit.""",
+        values_kwargs={"check_instr_errors": True},
```
BenediktBurger, line 73: It buzzles me, that `ask` does not complain about that unknown argument.

If you set `check_set_errors=True`  (or the get variant), then it will call check_set_errors (defaulting to check_errors). If an error is found, it will be logged and an exception raised.
So, it will do what you want to do.

pymeasure/instruments/santec/tsl570.py
```diff
@@ -0,0 +1,263 @@
...
+class TSL570(SCPIMixin, Instrument):
+    """Represents the Santec TSL-570 Tunable Laser and provides a high-level interface for
+    interacting with the instrument."""
+
+    def __init__(self, adapter, name="Santec TSL-570", **kwargs):
+        super().__init__(adapter, name, **kwargs)
+        self.write(":SYSTem:COMMunicate:CODe 1")  # Set the device to use SCPI commands
+        self.check_errors()  # Clear the error queue
+
+    def check_instr_errors(self):
```
BenediktBurger, line 43: I think this method is superfluous, as the parameter `check_get_errors` does, what you want to do here.

BenediktBurger (changes requested):
A few comments. Thanks

pymeasure/instruments/santec/tsl570.py
```diff
@@ -0,0 +1,266 @@
...
+
+from pymeasure.instruments import Instrument, SCPIMixin
+from pymeasure.instruments.validators import (
+    strict_range,
+    strict_discrete_range,
+    strict_discrete_set,
+)
+
+
+class InstrumentError(Exception):
```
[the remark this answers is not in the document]
fg-lumentum: No longer necessary after switching to using check_set_errors

pymeasure/instruments/santec/tsl570.py
```diff
@@ -0,0 +1,266 @@
...
+    """Represents the Santec TSL-570 Tunable Laser and provides a high-level interface for
+    interacting with the instrument."""
+
+    def __init__(self, adapter, name="Santec TSL-570", **kwargs):
+        super().__init__(adapter, name, **kwargs)
+        self.write(":SYSTem:COMMunicate:CODe 1")  # Set the device to use SCPI commands
+        self.check_errors()  # Clear the error queue
+
+    def check_instr_errors(self):
+        """Checks the instrument for errors, and raises an exeption if any are present."""
```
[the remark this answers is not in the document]
fg-lumentum: Again, no longer necessary

pymeasure/instruments/santec/tsl570.py
fg-lumentum: Switched to using check_set_errors

pymeasure/instruments/santec/tsl570.py
```diff
@@ -0,0 +1,263 @@
...
+        """Start repeated wavelength sweeps."""
+        self.write(":WAVelength:SWEep:REPeat")
+
+    def stop_sweep(self):
+        """Stop the wavelength sweep."""
+        self.write(":WAVelength:SWEep 0")
+
+    sweep_staus = Instrument.measurement(
+        "WAVelength:SWEep?",
+        """Get the current sweep status, "Stopped", "Running", "Standing by trigger",
```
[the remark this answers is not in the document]
fg-lumentum: I've switched sweep_status, sweep_mode, sweep_routing, and sweep_pattern to enums

tests/instruments/test_all_instruments.py
BenediktBurger, line 1: I think this is a relict of some formatter.

Typcially, it is good to format only those lines you touched in the content, or which are really bad (in this case you should do it in a separate commit).

Here, I do not care, whether you revert or not.

BenediktBurger (approved):
Thanks. Looks good. Please tell me, if you are ready to merge.

fg-lumentum:
Thanks, I'm ready to merge

PR #1256, opened 2025-03-20
Allow procedures to emit results in batches

jvdoorn:
This PR supersedes #1252, in addition it incorporates the feedback given there. In summary in addition to emitting results one by one, this PR adds the ability to emit a dictionary of sequences allowing the user to emit a lot of data at once. To demonstrate this ability see the example below.

```python
import logging
from itertools import count
from time import time

import numpy as np

import sys
import random
from pymeasure.experiment import BooleanParameter, Procedure
from pymeasure.experiment import IntegerParameter, Parameter

    runs = IntegerParameter('Number of Runs', default=1)
    iterations = IntegerParameter('Loop Iterations', default=100)
    use_batches = BooleanParameter('Use Batches', default=False)

    DATA_COLUMNS = ['Index', 'Random Number']

    def startup(self):
        random.seed(self.seed)

    def execute(self):
        start_time = time()

        index_counter = count()
        for j in range(self.runs):
            indices = np.fromiter([next(index_counter) for _ in range(self.iterations)], dtype=int)
            values = np.fromiter([random.random() for _ in range(self.iterations)], dtype=float)

            if not self.use_batches:
                    data = {
                        "Index": indices[i],
                        "Random Number": values[i]
                    }
                    self.emit('results', data)

            if self.should_stop():
                break

            if self.use_batches:
                data = {
                    "Index": indices,
                    "Random Number": values
                }
                self.emit('batch results', data)
            self.emit('progress', 100 * (j + 1) / self.runs)

        end_time = time()
        log.info(f"Average time per data point: {(end_time - start_time) / (self.iterations * self.runs) * 1E6:.2f} µs")


    def __init__(self):
        super().__init__(
            inputs=['runs', 'iterations', 'use_batches', 'seed'],
            displays=['runs', 'iterations', 'use_batches', 'seed'],
            x_axis='Index',
            y_axis='Random Number'
        )


if __name__ == "__main__":
    window = MainWindow()
    window.show()
    sys.exit(app.exec())
```

## Comments on previous feedback (see #1252)
@CasperSchippers 

> I have to admit that I find the simplicity and effectiveness quite striking.
> 
> I think making it a bit more explicit could be nice, but then I think I'm happy to include it; I do have two options/suggestions and wonder what your thoughts are?
> 
> 1. Would it be a good idea to make a different "topic" (such as `"batch results"` such that you can `self.emit("batch results",  data)`), to make it more explicit? In that case I think we should somehow handle unequal length series (i.e., what you warn for in line 127).

Agreed, gives the user much more control and might avoid issues with certain edge cases.

> 2. Another option could be to choose handling based on wheter you emit a dict (for a single datapoint) or a pandas dataframe (for a series); this also ensures that all entries are equally long for the series option.

This sounds like a good idea, however since I never really use Pandas I currently only added a check if it is a dataframe and if so tells the user it is not supported.

### DIFF
--- a/pymeasure/experiment/workers.py
+++ b/pymeasure/experiment/workers.py
@@ -122,12 +122,16 @@ class Worker(StoppableThread):
             self.recorder.handle(record)
         elif topic == 'batch results':
             if isinstance(record, pd.DataFrame):
-                log.error('Support for DataFrames when emitting batch results is not available yet.')
+                log.error(
+                    'Support for DataFrames when emitting batch results is not available yet.')
                 self.stop()
             elif self._is_dictionary_of_sequences(record):
                 lengths = [len(value) for value in record.values()]
                 if not all(length == lengths[0] for length in lengths):
-                    log.warning(f'Potential data loss: not all data columns have the same length (check your emitted results)')
+                    log.warning(
+                        'Potential data loss: not all data columns have the same length '
+                        '(check your emitted results)'
+                    )
 
                 for index in range(lengths[0]):
                     # Handle the records one by one.
@@ -141,15 +145,17 @@ class Worker(StoppableThread):
 
     def _is_dictionary_of_sequences(self, record: Any) -> bool:
         """
-        Checks if the record is a dictionary of sequences, there are a couple data types that we do not want to treat as
-        sequences, such as strings and bytes. This function will return False if any of the values in the dictionary are
-        strings or bytes or not a sequence.
+        Checks if the record is a dictionary of sequences, there are a couple data types that we do
+        not want to treat as sequences, such as strings and bytes. This function will return False
+        if any of the values in the dictionary are strings or bytes or not a sequence.
         """
         sequence_types = (Sequence, np.ndarray)
         type_exceptions = (str, bytes)
         if not isinstance(record, dict):
             return False
-        return all(isinstance(value, sequence_types) and not isinstance(value, type_exceptions) for value in record.values())
+        return all(
+            isinstance(value, sequence_types) and not isinstance(value, type_exceptions) for value
+            in record.values())
 
     def handle_abort(self):
         log.exception("User stopped Worker execution prematurely")
--- a/pymeasure/experiment/workers.py
+++ b/pymeasure/experiment/workers.py
@@ -119,25 +119,31 @@ class Worker(StoppableThread):
         except (NameError, AttributeError):
             pass  # No dumps defined
         if topic == 'results':
-            self.recorder.handle(record)
+            self.handle_record(record)
         elif topic == 'batch results':
-            if self._is_dictionary_of_sequences(record):
-                lengths = [len(value) for value in record.values()]
-                if not all(length == lengths[0] for length in lengths):
-                    log.warning(
-                        'Potential data loss: not all data columns have the same length '
-                        '(check your emitted results)'
-                    )
-
-                for index in range(lengths[0]):
-                    # Handle the records one by one.
-                    data = {key: value[index] for key, value in record.items()}
-                    self.recorder.handle(data)
-            else:
-                log.error(f'Unsupported type ({type(record)}) for batch results.')
-                self.stop()
+            self.handle_batch_record(record)
         elif topic == 'status' or topic == 'progress':
             self.monitor_queue.put((topic, record))
+
+    def handle_record(self, record: dict[str, Any]) -> None:
+        self.recorder.handle(record)
+
+    def handle_batch_record(self, record: Any) -> None:
+        if self._is_dictionary_of_sequences(record):
+            lengths = list(len(value) for value in record.values())
+            if not all(length == lengths[0] for length in lengths):
+                log.warning(
+                    'Potential data loss: not all data columns have the same length '
+                    '(check your emitted results)'
+                )
+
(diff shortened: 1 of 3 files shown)


pymeasure/experiment/workers.py
```diff
@@ -116,9 +120,43 @@ def emit(self, topic, record):
...
+                log.error(
+                    'Support for DataFrames when emitting batch results is not available yet.')
+                self.stop()
+            elif self._is_dictionary_of_sequences(record):
+                lengths = [len(value) for value in record.values()]
+                if not all(length == lengths[0] for length in lengths):
+                    log.warning(
+                        'Potential data loss: not all data columns have the same length '
+                        '(check your emitted results)'
+                    )
```
CasperSchippers, line 134: I do think it is indeed good to check on the length of the different array. I might be inclined to even be more harsh and give an error when they don´t match, since there should never be a reason that data is incomplete (and if it is, is there data missing at the beginning or the end of a column?)

(as a side note, the `range(lengths[0])` can be "dangerous" if they're not all the same length and if the first entry happens to be shorter than one of the other entries)

CasperSchippers (changes requested):
Thanks for this update; I've left some detail-remarks.

pymeasure/experiment/workers.py
```diff
@@ -116,9 +120,43 @@ def emit(self, topic, record):
...
+
+                for index in range(lengths[0]):
+                    # Handle the records one by one.
+                    data = {key: value[index] for key, value in record.items()}
+                    self.recorder.handle(data)
+            else:
+                log.error(f'Unsupported type ({type(record)}) for batch results.')
+                self.stop()
```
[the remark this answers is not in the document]
jvdoorn: Agreed, implemented in 9ae00d74.

pymeasure/experiment/workers.py
```diff
@@ -116,9 +120,43 @@ def emit(self, topic, record):
             pass  # No dumps defined
         if topic == 'results':
             self.recorder.handle(record)
+        elif topic == 'batch results':
+            if isinstance(record, pd.DataFrame):
+                log.error(
+                    'Support for DataFrames when emitting batch results is not available yet.')
+                self.stop()
```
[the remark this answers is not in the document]
jvdoorn: Agreed, implemented in a0f51793.

pymeasure/experiment/workers.py
jvdoorn: Good catch, I now log it as an error and return prematurely which also avoids the issue with `range(lengths[0])`, see 47905468.

pymeasure/experiment/workers.py
```diff
@@ -115,10 +118,47 @@ def emit(self, topic, record):
...
+            self.handle_record(record)
+        elif topic == 'batch results':
+            self.handle_batch_record(record)
         elif topic == 'status' or topic == 'progress':
             self.monitor_queue.put((topic, record))
 
+    def handle_record(self, record: dict[str, Any]) -> None:
+        self.recorder.handle(record)
+
+    def handle_batch_record(self, record: Any) -> None:
```
CasperSchippers, line 130: Currently, I think it is policy not to include type hints (also holds for _is_dictionary_of_sequences)

It also seems to be the cause of the failing CI (as `dict[...]`-style typing is not supported in python 3.8)

CasperSchippers (changes requested):
Is looking good; only have one minor remark at the moment.

pymeasure/experiment/workers.py
jvdoorn: Do you want me to remove (any) type hints? Personally I think it is always good to provide type hints as most editors can then already detect wrong usage of certain functions. It also makes it clearer to a user what they can expect from a function.

I believe I can fix the CI for py3.8 by using `Dict[str, Any]` instead of `dict`.

jvdoorn: Fixed the typing for python 3.8, but let me know if you want me to remove it.

BenediktBurger: Actually, some more recent Tools do not like these old typings.
Better is to import annotations from __future__, afterwards typing works even for older Python.

Originally (way back), it was decided against type hints. Now, little speaks against it, except, that the code base does not use them yet.

jvdoorn: I always forget about the `__future__` imports, changed the typing to use e.g. `dict[str, Any]` instead of `Dict[str, Any]`. Also took the liberty to specify the types for the other functions in `workers.py` where that made as a step towards improving the code base.

jvdoorn:
Also added documentation on the use of `emit('batch results', ...)` including a usage example and a small explanation why it might be beneficial.

pymeasure/experiment/workers.py
CasperSchippers: Sounds good to me; I'm generally in favour of type hints, so this sits fine with me.

CasperSchippers (approved):
Looks good to me; thanks for also adding a note in the documentation.

CasperSchippers:
BTW: I think it is good that we have this functionality now, so thanks a lot for that, and the simple implementation! I think the way you did it now is very nice, since, _if_ we want to change it under the hood, the user-side can still remain the same.

PR #1259, opened 2025-03-27
Add _write_bytes() and _read_bytes() for Prologix adapter

dl8dtl:
As in plain `write()`, `_write_bytes()` possibly needs to emit the `++addr` command first. Also, it must always append one CR in order for the Prologix adapter to actually transfer the data to the target.

In order to read something from a Prologix adapter, it is necessary to first emit a `++read` command, thus the default `_read_bytes` method alone is not sufficient.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -99,3 +99,4 @@ Fred Gillard
 Dimitrios Simatos
 Dongkai Pan
 Julian van Doorn
+Jörg Wunsch
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -283,6 +283,14 @@ class PrologixAdapter(VISAAdapter):
         return super()._read()
 
     def _read_bytes(self, count, break_on_termchar=False, **kwargs):
+        """Read bytes from the instrument.
+
+        :param int count: Number of bytes to read. A value of -1 indicates to
+            read from the whole read buffer.
+        :param bool break_on_termchar: Stop reading at a termination character.
+        :param \\**kwargs: Keyword arguments for the connection itself.
+        :returns bytes: Bytes response of the instrument (including termination).
+        """
         self.write("++read eoi")
         return super()._read_bytes(count, break_on_termchar, **kwargs)
 
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -25,6 +25,7 @@ import time
 from warnings import warn
 
 from pymeasure.adapters import VISAAdapter
+from pyvisa.constants import *
 
 
 class PrologixAdapter(VISAAdapter):
@@ -291,7 +292,10 @@ class PrologixAdapter(VISAAdapter):
         :param \\**kwargs: Keyword arguments for the connection itself.
         :returns bytes: Bytes response of the instrument (including termination).
         """
-        self.write("++read eoi")
+        avail = self.connection.get_visa_attribute(VI_ATTR_ASRL_AVAIL_NUM)
+        if avail == 0:
+            # nothing buffered, need to request data from Prologix
+            self.write("++read eoi")
         return super()._read_bytes(count, break_on_termchar, **kwargs)
 
     def gpib(self, address, **kwargs):
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -229,6 +229,12 @@ class PrologixAdapter(VISAAdapter):
             super().write("++addr %d" % self.address, **kwargs)
         super().write(command, **kwargs)
 
+    def _write_bytes(self, content, **kwargs):
+        if self.address is not None:
+            super().write("++addr %d" % self.address, **kwargs)
+        content += b'\x0d'
+        return super()._write_bytes(content, **kwargs)
+
     def _format_binary_values(self, values, datatype='f', is_big_endian=False, header_fmt="ieee"):
         """Format values in binary format, used internally in :meth:`.write_binary_values`.
 
@@ -281,6 +287,10 @@ class PrologixAdapter(VISAAdapter):
         if not prologix:
             self.write("++read eoi")
         return super()._read()
+
+    def _read_bytes(self, count, break_on_termchar=False, **kwargs):
+        self.write("++read eoi")
+        return super()._read_bytes(count, break_on_termchar, **kwargs)
 
     def gpib(self, address, **kwargs):
         """ Return a PrologixAdapter object that references the GPIB
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -25,7 +25,7 @@ import time
 from warnings import warn
 
 from pymeasure.adapters import VISAAdapter
-from pyvisa.constants import *
+from pyvisa.constants import VI_ATTR_ASRL_AVAIL_NUM
 
 
 class PrologixAdapter(VISAAdapter):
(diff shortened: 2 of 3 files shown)


dl8dtl:
As you suspect, `_format_binary_values` is already properly taking care of escaping all characters that need to be escaped. The appended (un-escaped) CR is then the indication to the Prologix firmware that the data submission is complete, and ought to be submitted to GPIB (where the trailing CR is normally stripped then).
Upon reading, there is no escaping of characters; it is only needed when writing to the Prologix so it can distinguish between Prologix-internal handling (like `++` commands), and actual GPIB data.

Right now, I'm still struggling with the pytest failures. I'll submit an update once I could resolve them.

BenediktBurger:
I'm thinking how to design the interface right: Should we do the escaping in the write bytes or in the format block.
Also, the `write_binary_values` method already sends the address to the adapter.

dl8dtl:
> Should we do the escaping in the write bytes or in the format block.

I initially added it to the write bytes method, but then noticed it's already there in the formatting subroutine.

Regarding the Prologix, both will certainly work.

> Also, the `write_binary_values` method already sends the address to the adapter.

Umm, I didn't realize that. But when thinking about it, it already has to issue whatever `cmd` has been given first, so makes sense.

dl8dtl:
I pushed an updated. pytest passes for me now.
Only thing with the HP3478A is a warning:
```
$ python
Python 3.11.11 (main, Jan  7 2025, 17:38:55) [Clang 18.1.6 (https://github.com/llvm/llvm-project.git llvmorg-18.1.6-0-g1118c2 on freebsd13
Type "help", "copyright", "credits" or "license" for more information.
>>> from pymeasure.adapters import PrologixAdapter
>>> from pymeasure.instruments.hp import HP3478A
>>> 
>>> adapter = PrologixAdapter("ASRL/dev/cu.0::INSTR", 13, visa_library="@py", timeout=1000)
>>> mm = HP3478A(adapter)
>>> x = mm.calibration_data
/home/joerg/src/pymeasure/pymeasure/adapters/visa.py:156: UserWarning: write message already ends with termination characters
  self.connection.write(command, **kwargs)
>>> print(x)
[0, 0, 0, 0, 2, 9, 5, 0, 4, 3, 0, 2, 14, 6, 0, 0, 0, 0, 3, 1, 0, 4, 5, 12, 4, 14, 2, 0, 0, 0, 0, 0, 3, 0, 4, 2, 15, 1, 14, 6, 9, 9, 9, 9, 9, 5, 1, 12, 13, 3, 1, 10, 15, 9, 9, 9, 9, 9, 9, 0, 5, 5, 12, 14, 10, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 15, 0, 0, 0, 6, 1, 3, 0, 3, 0, 4, 4, 14, 10, 0, 0, 0, 0, 5, 4, 0, 5, 1, 0, 13, 14, 3, 0, 0, 0, 0, 1, 1, 0, 5, 2, 1, 5, 15, 0, 0, 0, 0, 0, 0, 1, 0, 5, 15, 4, 13, 13, 9, 0, 0, 0, 0, 0, 1, 0, 5, 14, 12, 3, 13, 12, 0, 0, 0, 0, 0, 1, 0, 5, 12, 5, 12, 13, 12, 0, 0, 0, 0, 0, 0, 0, 5, 13, 4, 5, 14, 4, 0, 0, 0, 0, 0, 1, 1, 12, 13, 12, 3, 13, 5, 9, 9, 9, 8, 6, 9, 1, 4, 2, 14, 0, 11, 8, 9, 9, 9, 9, 8, 7, 1, 4, 3, 3, 2, 11, 15, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 15, 0, 0, 0, 6, 1, 3, 1, 3, 0, 12, 14, 13, 7, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 15, 0, 0, 0, 0, 0, 0, 0, 0]
```
The resulting data are correct though. Not sure how to get rid of the warning.

BenediktBurger:
I'm not sure, whether we want to modify the `write_bytes`, as that contradicts the expectation.
Currently, the idea behind write_bytes is, that you can send bytes without any modification, such that sending a message in two parts does not make any difference to sending it in one go.

Regarding the termination of the message with a carriage return: I would add this to the `write_binary_values` method. That one has in fact a "termination" option (inherited from the parent), which I would set per default to carriage return.

BenediktBurger:
> Only thing with the HP3478A is a warning:

Is it from an existing test or from your device?
The `write` method already appends a term char, so if you execute `write("something\n")`, it will send the bytes `b"something\n\n"`.

dl8dtl:
> I'm not sure, whether we want to modify the `write_bytes`, as that contradicts the expectation.
> Currently, the idea behind write_bytes is, that you can send bytes without any modification, such that sending a message in two parts does not make any difference to sending it in one go.

I already remove overloading `_write_bytes()`. As you mentioned, addressing is already handled correctly in `write_binary_values()`. So the PR is basically just to add the `++read eoi` command in `_read_bytes()` now.

dl8dtl:
> Is it from an existing test or from your device?

This is from my actual device. The code quoted above was everything, i.e. all the internals are from `pymeasure` itself.

BenediktBurger:
I'm unsure about the read change.
Here again, `read_bytes` should be plain access to the sent bytes.
On the other hand, the prologix adapter should appear like any adapter reading from the device.

A question: What happens, if the device sends 10 bytes and you call your `read_bytes(5)` two times (writing twice ++eoi)?
If that works, I'm happy to merge.

dl8dtl:
> A question: What happens, if the device sends 10 bytes and you call your `read_bytes(5)` two times (writing twice ++eoi)?

That's going to be difficult then.
Right now, it would issue `++read` twice, possibly triggering one read too much.
For sure, without any `++read`, the Prologix will never send anything at all. However, if only part of the previous response has been fetched from the upper layers, no additional `++read` is needed. That raises the question: is there any method to query the lower layer for wether the buffer already / still has data available? If so, emitting the `++read` can be made dependent on that information.

BenediktBurger:
Does a second ++read hurt if there is still data in the buffer?

dl8dtl:
It potentially triggers another read cycle:
```
>>> mm.read_bytes(3)
b'+1.'
>>> mm.read_bytes(-1)
b'75340E+0\r\n+1.75632E+0\r\n'
```
(The HP3478A always issues the current display value on the bus when asked, even without a command.)

I checked the underlying serial implementation (in pyvisa_py), but I don't see any method to figure out whether there's data pending.
Without that, the only option I see is to try querying the lower layers with a very low (temporary) timeout, catching the possible timeout exception. If the exception was thrown, issue the `++read`, and wait for the normal timeout.

Seems quite cumbersome …

dl8dtl:
I think I found a more elegant solution.
Prologix adapters all implement a (USB-based) serial connection, so we can access the respective VISA attributes.
Digging deep into that stuff, I noticed there's a VISA attribute for the number of bytes that can be read from the connection.
Thus, it is only needed to emit a `++read eoi` command if no bytes are buffered already.

```
>>> mm.read_bytes(5)
b'+1.72'
>>> mm.read_bytes(-1)
b'955E+0\r\n'
```
I think this is the expected behaviour now.

dl8dtl:
Nevermind, all is fine. The HP3478A change is on a different branch. :-)

BenediktBurger (approved):
Thanks for your investigation and contribution!

PR #1260, opened 2025-03-27
HP3478a calibration data are binary

dl8dtl:
When reading calibration data, treat the command as binary

Just using `write_bytes()` fails to call `_format_binary_values()`
on the data to send. Using this is crucial on Prologix adapters
as they need special escaping of certain binary values.

Thus, rather use `write_binary_values()` instead.

### DIFF
--- a/pymeasure/adapters/prologix.py
+++ b/pymeasure/adapters/prologix.py
@@ -229,6 +229,12 @@ class PrologixAdapter(VISAAdapter):
             super().write("++addr %d" % self.address, **kwargs)
         super().write(command, **kwargs)
 
+    def _write_bytes(self, content, **kwargs):
+        if self.address is not None:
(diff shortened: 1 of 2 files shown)


BenediktBurger:
Let's do the Prologix change in the #1259 PR.

dl8dtl:
> Let's do the Prologix change in the #1259 PR.

OK

BenediktBurger:
Sorry, we can change the HP device in here, but the Prologix Adapter part not.

dl8dtl:
Ah, now understood :)
The Prologix change made it here inadvertently (wrong base branch for the PR).
I'll finish up the Prologix stuff first, and then re-submit the HP3478A part as a new PR instead. So just let it closed right now.

dl8dtl:
(Reopened as #1263)

PR #1263, opened 2025-03-28
HP3478A: use write_binary_values() to request calibration data

dl8dtl:
The argument to the "W" command in HP3478A instruments is binary (0x00 through 0xFF). For certain adapters (e.g. Prologix), this requires special handling. Consequently, instead of a plain write_bytes(), use write_binary_values() instead.

### DIFF
--- a/pymeasure/instruments/hp/hp3478A.py
+++ b/pymeasure/instruments/hp/hp3478A.py
@@ -491,8 +491,7 @@ class HP3478A(HPLegacyInstrument):
         cal_data = []
         for addr in range(0, 256):
             # To fetch one nibble: 'W<address>', where address is a raw 8-bit number.
-            cmd = bytes([ord('W'), addr])
-            self.write_bytes(cmd)
+            self.write_binary_values('W', [addr], datatype='B', header_fmt="empty")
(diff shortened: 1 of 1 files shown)


BenediktBurger:
Best place is the code itself (meaningful names), second best place are the docstrings.
In this case, add it to the read method or the Prologix adapter itself.

Just a quick check: Does a test cover this change?

dl8dtl:
> Just a quick check: Does a test cover this change?

Not sure. The method itself has been there before. Whether to use `write()`, or `write_binary_values()` probably only makes a difference on the Prologix, where binary values need to be escaped. Thus, in order to meaningfully test it, it would require a kind of Prologix emulator.

dl8dtl:
I had a look into the tests. Yes, the calibration data read/write of the HP3478A is part of the existing tests, and the test still successfully runs.

Besides, my real HP3478A also successfully passed the test. ;-)

dl8dtl:
> In this case, add it to the read method or the Prologix adapter itself.

Did it that way: #1272

BenediktBurger:
Thanks for your implementation

PR #1266, opened 2025-03-29
Update the custom_ui example

CasperSchippers:
Update the custom_ui example to use the standard implementation of the queue method.

Incited by #1262.

Since nowadays there is a default implementation of the `queue` method, I think it is better to re-implement the `make_procedure` method than the `queue` method: this is typically a smaller method that _only_ handles creating the procedure instance and assigning the parameters from the ui. Therefore, this seems like the logical place to also handle getting the values from the custom UI. Finally, since this method is also used by e.g. the SequencerWidget, this approach also enables using the SequencerWidget in combination with a custom ui.


BenediktBurger:
Looks good to me. You may merge it.

PR #1267, opened 2025-03-31
Correct flake8 error in agilent33220A.py

Falk3579:
Add missing whitespace in line 26 as shown by flake8


BenediktBurger:
Thanks for spotting and directly fixing it.

You may add yourself to the authors.txt if you wish so.
Or you may do it in the next PR you are doing. We probably won't remind you though, as you will not be marked a "First-time contributor" anymore.

BenediktBurger:
Thanks again for your work and sorry for the confusion with the conflict resolution.

PR #1268, opened 2025-04-01
Remove deprecated parts

### LINKED ISSUE #1264: Remove deprecated preprocess_reply and maybe more
preprocess_reply (and similar) have been deprecated since version 0.11 (released end of 2022), we should be able to remove it safely by now.

@bilderbuchi , what do you think?

BenediktBurger:
Removes code which is deprecated since 0.11, mainly due to the adapter/instrument rework.

Closes #1264


CasperSchippers (approved):
Good thing to remove the deprecated parts.

I didn´t notice anything strange: looks good to me.

PR #1269, opened 2025-04-01
Fix flake8

Falk3579:
Fix two flake8 errors in instruments.


BenediktBurger:
Thanks for spotting them.

PR #1270, opened 2025-04-02
Fix raise errors to show full stack trace

BenediktBurger:
Raising an error anew (raise exc) will point the stack trace there, better use raise alone or do nothing at all.


CasperSchippers (approved):
Sorry for the late response; looks good to me.

PR #1277, opened 2025-04-08
Simplify test passing conditions

### LINKED ISSUE #1276: Test instability
Seems like one of the existing tests doesn't always pass because of the randomness in the performance of CI.
https://github.com/pymeasure/pymeasure/blob/bdae50f99ff523525189cf698d5045d6054d08d9/tests/instruments/temptronic/test_temptronic_base.py#L31-L39

See this job as an example. https://github.com/pymeasure/pymeasure/actions/runs/14314010550/job/40115738391#step:8:348 

I guess we should correct acceptable_factor to a lower value or remove/rework the test at all.

_Originally posted by @ilev-sergey in https://github.com/pymeasure/pymeasure/issues/1250#issuecomment-2783953859_

ilev-sergey:
Update the time required to pass the test, so it passes in CI

Closes #1276

### DIFF
--- a/tests/instruments/temptronic/test_temptronic_base.py
+++ b/tests/instruments/temptronic/test_temptronic_base.py
@@ -24,8 +24,8 @@
 
 from time import perf_counter_ns
 
-from pymeasure.test import expected_protocol
 from pymeasure.instruments.temptronic.temptronic_base import ATSBase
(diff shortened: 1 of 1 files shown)


BenediktBurger:
Thanks

PR #1278, opened 2025-04-09
Add SRS LDC500Series

fg-lumentum:
Added a new instrument, the SRS LDC500Series of laser diode controller.
Test of the new instrument has been included.
Kept as a draft for now until the behaviour of properties containing ``check_set_errors`` have been validated with the instrument (marked as TODO).

note: In this pull request .spyproject has been added to git ignore

### DIFF
--- a/pymeasure/instruments/srs/ldc500series.py
+++ b/pymeasure/instruments/srs/ldc500series.py
@@ -65,11 +65,12 @@ class LDC500Series(SCPIMixin, Instrument):
 
     @property
     def options(self):
-        """Not implemented, raises ``NotImplementedError``"""
+        """Get options not implemented, raises ``NotImplementedError``"""
         raise NotImplementedError("options not implemented in LDC500series")
 
+    @property
     def next_error(self):
-        """Not implemented, raises: ``NotImplementedError``"""
+        """Get next error not implemented, raises: ``NotImplementedError``"""
         raise NotImplementedError("next_error not implemented in LDC500series")
 
     last_execution_error = Instrument.measurement(
--- a/pymeasure/instruments/srs/ldc500series.py
+++ b/pymeasure/instruments/srs/ldc500series.py
@@ -31,7 +31,19 @@ from pymeasure.instruments.validators import strict_discrete_set, strict_range
 
 
 class ThermometerType(IntEnum):
-    """Enumerator of thermometer types supported by the SRS LDC500Series."""
+    """Enumerator of thermometer types supported by the SRS LDC500Series.
+    (NTC - negative temperature coefficient; RTD - Resistance temperature detector)
+
+    Members:
+
+        - **NTC10UA**  : 10 µA excitation for NTC thermistor.
+        - **NTC100UA** : 100 µA excitation for NTC thermistor.
+        - **NTC1MA**   : 1 mA excitation for NTC thermistor.
+        - **NTCAUTO**  : Automatically detects optimal excitation for NTC.
+        - **RTD**      : Typically Pt-100 or other PTC sensors
+        - **LM335**    : LM335 and compatible temperature-to-voltage transducers.
+        - **AD590**    : AD590 and compatible temperature-to-current transducers.
+    """
 
     NTC10UA = 0
     NTC100UA = 1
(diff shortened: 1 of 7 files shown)


BenediktBurger:
I just had a quick look at the code: Instead of using ld_current and tec_current, you could use Channels to distinguish the different parts.
Just as a suggestion.

fg-lumentum:
I don't think putting the currents under different channels would be useful as the laser diode and TEC are quite different components. However, it did get me thinking about splitting the class up into subsystems as a user may want to use just one system on the ldc500 (in my case the TEC). This would enable a user to do the following::

    ldc = LDC500Series("GPIB:0")
    tec = ldc.tec
    tec.temperature_setpoint = 45
    etc..

It makes sense from an instrument point of view, as the tec is separate to ld and pd. I don't know if this is done at all in any other instruments.

BenediktBurger:
You basically did implement, what I meant, but did not use channels.
If your Subsystems inherit from Channel instead of Instrument and the main instrument uses the ChannelCreator, then you use our correct implementation, but usage remains (almost) the same.

For your information: Channels can be used for several times the same functionality of a device (like oscilloscope channels) or for different functionalities of a combined device, like in your case.

pymeasure/instruments/srs/ldc500series.py
```diff
@@ -0,0 +1,516 @@
...
+    """
+
+    def __init__(self, adapter, name="LDC500Series", **kwargs):
+        super().__init__(adapter, name, **kwargs)
+        self.ld = LDC500SeriesLDSubsystem(self)
+        self.pd = LDC500SeriesPDSubsystem(self)
+        self.tec = LDC500SeriesTECSubsystem(self)
+
+
+class LDC500SeriesLDSubsystem(LDC500SeriesBase):
```
BenediktBurger, line 133: Should be based on `Channel` (or some class inheriting from `Channel`), as every `Instrument` should have its own adapter. All composite classes should be Channels.

pymeasure/instruments/srs/ldc500series.py
```diff
@@ -0,0 +1,516 @@
...
+    enabled = Instrument.control(
+        "LDON?",
+        "LDON %s",
+        """Control whether the laser diode current source is enabled (bool).""",
+        validator=strict_discrete_set,
+        values={True: "ON", False: "OFF"},
+        map_values=True,
+    )
+
+    mode = Instrument.control(
```
BenediktBurger, line 156: as an idea: if you call it `CC_mode_enabled` (or similar), you can use a boolean to control it

pymeasure/instruments/srs/ldc500series.py
```diff
@@ -0,0 +1,516 @@
...
+        """Control the laser diode current limit, in mA (float).
+
+        The ouput current is clamped to never exceed ``current_limit`` under any conditions.
+        If ``current_limit`` is reduced below ``current_setpoint``,
+        ``current_setpoint`` is "dragged" down with ``current_limit``.
+        """,
+        check_set_errors=True,
+    )
+
+    current_range = Instrument.control(
```
BenediktBurger, line 203: here also, `current_range_high_enabled` could give you a boolean, which is easier than a string property.

BenediktBurger (commented):
Thanks for your rework, it looks good, but just requires a migration to `Channel`. If you have questions, feel free to ask.

fg-lumentum:
In terms of the string vs boolean properties, I think I'd rather keep them as strings as it aligns better with the instrument documentation.

Though ``ld.cc_mode_enabled = False`` would imply being in CP mode, I feel it would be clearer to the user if ``ld.mode = "CP"`` was just stated explicitly. This is compounded by ``tec.cc_mode_enabled = False`` implying being in CT mode, not CP as this doesn't exist for the TEC.

BenediktBurger (approved):
Channel inits are superfluous and can be removed.

Thanks for your work!

fg-lumentum:
Channel inits removed and an internal property put in to control the check_set_errors delay, thanks :)

BenediktBurger:
Thanks for your work

PR #1282, opened 2025-04-18
Update keysightN7776C.py

langrock:
Changed sweep speeds to reflect available values present in current generation of N7776C.


BenediktBurger:
Thanks for your contribution.

You may add yourself also to the authors.txt.

More descriptive titles are helpful, especially if you open three PRs with the same title.

PR #1283, opened 2025-04-18
Update keysightN7776C.py

langrock:
Added lambda_zero and cavity_align function representing 'wavelength zero' and 'cavity realignment' from the N7776C programming guide.


BenediktBurger (changes requested):
Thanks for your contribution.

PR #1284, opened 2025-04-18
Update keysightN7776C.py

langrock:
Added instrument control for sweep cycles and dwell time when using stepped sweep.


langrock (commented):
Missing quotation mark.

langrock (commented):
Indentation and line length fixed.

PR #1285, opened 2025-04-21
Implement qtpy

Benbriel:
These changes should be enough to implement qtpy on the whole project. Technically pyqtgraph uses its own Qt bindings, so issues are possible, but unlikely if one keeps only one Qt installation.

I don't know how dependencies are added for conda-forge, though, so that's missing.

### DIFF
--- a/.github/pymeasure.yml
+++ b/.github/pymeasure.yml
@@ -12,7 +12,7 @@ dependencies:
   - pyvisa=1.13.0
   - pyzmq=24.0.1
   - qt=5.15.6
-  - qtpy=2.4.3
+  - qtpy=2.4.2
 # Development dependencies below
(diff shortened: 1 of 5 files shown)


CasperSchippers:
Nice work; feel free to ping me when you want it reviewed.

@bilderbuchi or @BenediktBurger: I also do not know how to add a dependency; is that updating the recipe.yaml in the conda feedstock repo?

BenediktBurger:
Yes, for conda-forge, we have to modify the recipe.yaml file.
However, it might be, that the automatic updater will pick up the new dependency in the pyproject.toml.

@CasperSchippers , it is fine to be merged from my side.

CasperSchippers (approved):
Looks good; thanks for the work!

PR #1286, opened 2025-04-22
Yokogawa aq6370 sweep status

fg-lumentum:
- Added ``sweep_complete`` measurement property to get the completion status of the sweep (``True`` if complete)
- Added ``wait_for_sweep_complete`` method to block the program until the sweep completes (Or ``should_stop`` returns ``True``, or times out)


fg-lumentum:
Thanks, both comments addressed :)

PR #1289, opened 2025-04-24
Dev/srs860 improvement

panasee:
This pull request refactors and enhances the `SR860` class in the `pymeasure` library by removing unknown mixins, fixing typos, adding new properties, and introducing a new method for retrieving multiple parameters simultaneously. Below is a summary of the most important changes:

### Refactoring and Cleanup:
* Removed the `SCPIUnknownMixin` from the `SR860` class, as it is no longer needed. Updated the `__init__` method to exclude SCPI-specific functionality by setting `includeSCPI=False`. Because some of the methods in SCPIMixin class is proved to be unsupported by SR860.

### Enhancements to Class Properties:
* Added new aliases for existing properties to provide compatibility with the SR830 driver on some basic functionalities:
  - `reference_source_trigger` as an alias for `reference_triggermode`.
  - `input_config` as an alias for `input_voltage_mode`.
  - `input_grounding` as an alias for `input_shields`.

### Bug Fixes:
* Corrected a typo in the `sensitivity` property name (previously misspelled as `sensitvity`). NOTE: **The original wrongly spelled parameters are removed, meaning this is a breaking change**.
* Fixed the range of valid values for the `filter_synchronous` property to include 24 dB/octave (adjusted from `range(0, 3)` to `range(0, 4)`).
* Reverse the order of the mapping list of sensitivity according to the manual of SR860.

### New Method:
* Added the `snap_all` method to retrieve (`X`, `Y`, `R`, `THETA`) in a single command, improving efficiency for users requiring these values simultaneously. While the `snap` method can only fetch up to 3 of them.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+Upcoming Release
+================
+
+Deprecated
+----------
+- Replaced :code:`sensitvity` attribute of :code:`pymeasure/instruments/srs/SR860.py` by :code:`sensitivity`
+- Replaced :code:`filer_synchronous` attribute of :code:`pymeasure/instruments/srs/SR860.py` by :code:`filter_synchronous`
+
 Version 0.15.0 (2025-01-15)
 ===========================
 Main items of this new release:
--- a/pymeasure/instruments/srs/sr860.py
+++ b/pymeasure/instruments/srs/sr860.py
@@ -30,10 +30,10 @@ from pymeasure.instruments import Instrument, SCPIUnknownMixin
 class SR860(SCPIUnknownMixin, Instrument):
 
     SENSITIVITIES = [
-        1e-9, 2e-9, 5e-9, 10e-9, 20e-9, 50e-9, 100e-9, 200e-9,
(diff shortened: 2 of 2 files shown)


BenediktBurger (changes requested):
Thanks for the work and detailed description.

Please create aliases for the typoed properties too be backward compatible.

And please add deprecation warnings there, see for a general note: https://github.com/pymeasure/pymeasure/issues/672

A concrete example with a typo is https://github.com/pymeasure/pymeasure/blob/c1ef467e884e01d7209a219612525de7bc74a192/pymeasure/instruments/lakeshore/lakeshore_base.py#L56-L64

panasee:
OK, I've added compatible aliases. The deprecated version number is preliminarily set as 0.15.0, you can change it according to the actual case.

BenediktBurger:
Please change the deprecated version number to the next release (0.16.0)

panasee:
Done

BenediktBurger:
Thank you very much for your work!

Issue #1288, opened 2025-04-23
FIX Preprocessing issue in Commonbase.control.fget


hvegh:
I have this instrument function that basically multiplexes the averaging with the measurement speed:

![Image](https://github.com/user-attachments/assets/f5326ddb-9ea8-47a2-9488-dd4e27f6af8d)

So the 'query' command basically returns 2 values: [ str, integer ] I would like to map the 'str' value SLOW MED FAST to the more generic measurements  per second:

```python
measurement_speed = Instrument.control(
        "APER?", "APER %s",
        """Control the measurement speed in meas. per second.""",
        validator=strict_discrete_set,
        values={2.5:'SLOW', 10:'MED', 40:'FAST'},
        map_values=True,
        get_process=lambda v: [ v[0] ]
    )
```
But in the current code if there are more than 2 values, there is never a mapping performed even if a single value is returned:

```python
            if len(vals) == 1:
                value = get_process(vals[0])
                if not map_values:
                    return value
                elif isinstance(values, (list, tuple, range)):
                    return values[int(value)]
                elif isinstance(values, dict):
                    for k, v in values.items():
                        if v == value:
                            return k
                    raise KeyError(f"Value {value} not found in mapped values")
                else:
                    raise ValueError(
                        'Values of type `{}` are not allowed '
                        'for Instrument.control'.format(type(values))
                    )
            else:
                vals = get_process(vals)
                return vals
```
I propose this change:

```patch
--- apymeasure/instruments/common_base.py      2025-04-22 21:37:23.652371008 +0200
+++ b/pymeasure/instruments/common_base.py   2025-04-23 19:58:39.424985981 +0200
@@ -570,7 +570,11 @@
                     log.error("Error received after trying to get a property with the command "
                               f"""'{command_process(get_command)}': '{"', '".join(errors)}'.""")
             if len(vals) == 1:
-                value = get_process(vals[0])
+                vals = [ get_process(vals[0]) ]
+            else:
+                vals = get_process(vals)
+            if len(vals) == 1:
+                value = vals[0]
                 if not map_values:
                     return value
                 elif isinstance(values, (list, tuple, range)):
@@ -586,7 +590,6 @@
                         'for Instrument.control'.format(type(values))
                     )
             else:
-                vals = get_process(vals)
                 return vals
 
         def fset(self,
```

new code:

```python
            if len(vals) == 1:
                vals = [ get_process(vals[0]) ]
            else:
                vals = get_process(vals)
[...]

PR #1291, opened 2025-04-25
Add get process list

### LINKED ISSUE #781: Unify get_process between single and multiple values
Right now, get process works differently, if `values` returns a single value or a list:
~~~python
            if len(vals) == 1:
                value = get_process(vals[0])  # works on a single entry
                if not map_values:
                    return value  # otherwise do mapping here
            else:
                vals = get_process(vals)  # works on the list as a whole
                return vals
~~~

However, #637 and continued in #740 , `control` should work the same in both cases (list or value).

### Possibly affected instrument properties

Group 1: Properties which use `get_process` on a single entry (members yet to be determined):

Group 2: Some instruments implement `get_process(list)` in order to filter the desired entries or similar:
* [ ] AgilentE4980.impedance
* [ ] AnritsuMS9710C.trace_marker
* [ ] BKPrecision9130B.channel
* [ ] Fluke7341.id
* [ ] HP8116A.operating_mode, control_mode, trigger_slope, shape
* [ ] Keithley2306.last_readings, readings, measured_voltages, measured_currents, dvm_voltages, pulse_currents, long_integration_currents
* [ ] Keithley2750.closed_channels
* [ ] Keithley6517B.current, voltage, resistance
* [ ] DSOX1102G.label

### Implementation ideas

Now the question is, on what works `get_process`?
- it works on the whole list (even if that list has a single entry): same behavior independent of the implementation. Group 1 properties have to be adjusted. (e.g. `lambda vals: 2 * vals` to `lambda vals: [2 * v for v in vals]`). This is easily possible without much thinking, but not that intuitive at first (if you expect a single value).
- it works on every entry of the list: Group 2 properties have to be adjusted, the solution is not as straight forward as no operation on the full list is possible. It is more intuitive for a single return value (`lambda v: 2 * v`).
- `get_process` is called **after** the mapping is done and immediately before returning (i.e. after the extraction of the single entry of a list). Group 1 and Group 2 properties have to be checked that they work together with mapping. get_process can stay the same in several implementations. Downside: you have to know which value to expect (list or single value) in order to choose `get_process` right.
- Introducing a `list_operation` method working on the whole list and `get_process` on every single entry of the list. All Group 2 properties could `get_process` to `list_operation` and should continue to work as expected. That way `get_process` and the values map work both on the individual entries.

BenediktBurger:
Closes #781 as it fixes all instruments mentioned there as well.

Alternative to #740, which never got adjusted.
Related to #1288

ToDo:
- [x] check that all occurances of get_process do not use a list anymore!


CHANGES.rst
```diff
@@ -1,3 +1,11 @@
+Upcoming release
+================
+
+Changed features
+----------------
+- :code:`Instrument.control` does not apply :code:`get_process` to a returned list anymore. Use :code:`get_process_list` parameter instead.
```
CasperSchippers, line 6: Maybe specify that get_process_list is applied only to returned list and that get_process is still applied when a single item is returned.

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -353,7 +353,7 @@ def long_integration_time_auto(self):
     last_readings = Channel.measurement(
         ":FETCH{ch}:ARR?",
         """A floating point array property that returns the last readings. """,
-        get_process=lambda v: v if isinstance(v, list) else [v]
+        get_process_list=lambda v: v if isinstance(v, list) else [v]
```
CasperSchippers, line 356: Shouldn´t this be:

```suggestion
        get_process=lambda v: [v],
```

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -379,7 +379,7 @@ def long_integration_time_auto(self):
         ":MEAS{ch}:ARR:VOLT?",
         """A floating point array property that triggers and returns
         voltage readings. """,
-        get_process=lambda v: v if isinstance(v, list) else [v]
+        get_process_list=lambda v: v if isinstance(v, list) else [v]
```
CasperSchippers, line 382: Shouldn´t this be:

```suggestion
        get_process=lambda v: [v],
```

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -392,7 +392,7 @@ def long_integration_time_auto(self):
         ":MEAS{ch}:ARR:CURR?",
         """A floating point array property that triggers and returns
         current readings. """,
-        get_process=lambda v: v if isinstance(v, list) else [v]
+        get_process_list=lambda v: v if isinstance(v, list) else [v]
```
CasperSchippers, line 395: Shouldn´t this be:

```suggestion
        get_process=lambda v: [v],
```

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -405,7 +405,7 @@ def long_integration_time_auto(self):
         ":MEAS{ch}:ARR:DVM?",
         """A floating point array property that triggers and returns
         DVM voltage readings. """,
-        get_process=lambda v: v if isinstance(v, list) else [v]
+        get_process_list=lambda v: v if isinstance(v, list) else [v]
```
CasperSchippers, line 408: Shouldn´t this be:

```suggestion
        get_process=lambda v: [v],
```

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -417,7 +417,7 @@ def long_integration_time_auto(self):
         ":MEAS{ch}:ARR:PCUR?",
         """A floating point array property that triggers and returns
         pulse current readings. """,
-        get_process=lambda v: v if isinstance(v, list) else [v]
+        get_process_list=lambda v: v if isinstance(v, list) else [v]
```
CasperSchippers, line 420: Shouldn´t this be:

```suggestion
        get_process=lambda v: [v],
```

pymeasure/instruments/keithley/keithley2306.py
```diff
@@ -430,7 +430,7 @@ def long_integration_time_auto(self):
         ":MEAS{ch}:ARR:LINT?",
         """A floating point array property that triggers and returns
         long integration current readings. """,
-        get_process=lambda v: v if isinstance(v, list) else [v]
+        get_process_list=lambda v: v if isinstance(v, list) else [v]
```
CasperSchippers, line 433: Shouldn´t this be:

```suggestion
        get_process=lambda v: [v],
```

CasperSchippers (approved):
Appart from a few minor remarks, looks good to me!

Nice work.

pymeasure/instruments/keithley/keithley2306.py
BenediktBurger: You're right.

Thanks!

CHANGES.rst
BenediktBurger: I changed it.

PR #1296, opened 2025-05-07
Add Agilent B2980 electrometer series

Falk3579:
Support for the following Agilent/Keysight instruments:
- B2981
- B2983
- B2985
- B2987


pymeasure/instruments/agilent/agilentB298x.py
```diff
@@ -0,0 +1,1124 @@
...
+from pymeasure.instruments.validators import (strict_discrete_set,
+                                              strict_range,
+                                              joined_validators
+                                              )
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class Battery(SCPIMixin, Instrument):
```
BenediktBurger, line 38: You use that `Battery` as a mixin, never as a standalone instrument, right?

Therefore, you should rename it to `BatteryMixin` and not inherit anything:
```suggestion
class BatteryMixin:
```

Consequently, you do not need the init method either.

Alternatively, you could implement the Battery as a Channel which you add to the respective instruments. See: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html

pymeasure/instruments/agilent/agilentB298x.py
```diff
@@ -0,0 +1,1124 @@
...
+        ":MEAS?",
+        """Measure the defined parameter(s) with a spot measurement.
+
+        :return: float or list of float
+        """
+        )
+
+    current = Instrument.measurement(
+        ":MEAS:CURR?",
+        """Measure current with a spot measurement.
```
BenediktBurger, line 110: In which units, Amps?

pymeasure/instruments/agilent/agilentB298x.py
```diff
@@ -0,0 +1,1124 @@
...
+# Trigger system #
+##################
+
+    def abort(self, action='ALL'):
+        """Abort the specified action.
+
+        :param str, optional action: strictly in ``ALL``, ``ACQ``
+        """
+        strict_discrete_set(action, ['ALL', 'ACQ'])
+        self.write(f":ABOR:{action}")
```
BenediktBurger, line 153: Maybe it is easier to add two methods: `abort_acquisition` and `abort` which aborts anything.

pymeasure/instruments/agilent/agilentB298x.py
```diff
@@ -0,0 +1,1124 @@
...
+        :param str, optional action: strictly in ``ALL``, ``ACQ``
+        """
+        strict_discrete_set(action, ['ALL', 'ACQ'])
+        self.write(f":INIT:{action}")
+
+###########################################
+# Trigger properties for action 'ACQuire' #
+###########################################
+
+    arm_acquire_bypass_once_enabled = Channel.control(
```
BenediktBurger, line 178: The parent class is an `Instrument`, so you should use `Instrument.control`. (currently there is no difference, but that might change).

tests/instruments/agilent/test_agilentB298x_with_device.py
```diff
@@ -0,0 +1,387 @@
...
+# Tested using SCPI over USB. call signature:
+# $ pytest test_agilentB298x_with_device.py --device-address USB0::0x2A8D::0x9B01::MY61390198::INSTR
+#
+# Test was performed with B2987B
+#
+
+import pytest
+from pymeasure.instruments.agilent.agilentB298x import AgilentB2987  # B2987 supports all features
+from time import sleep
+# from pyvisa.errors import VisaIOError
```
BenediktBurger, line 36:
```suggestion
```

If unneeded, please remove it.

BenediktBurger (changes requested):
Thanks for your work.

A first round of comments.

pymeasure/instruments/agilent/agilentB298x.py
Falk3579: Added unit to docstring.

pymeasure/instruments/agilent/agilentB298x.py
Falk3579: Split into explicit methods for the different trigger actions as recommended.

pymeasure/instruments/agilent/agilentB298x.py
Falk3579: Changed from `Channel` to `Instrument`

PR #1300, opened 2025-05-14
Add tests to AgilentE4980A

Falk3579:
- add files for the protocol tests and device tests of the Agilent E4980A impedance meter
- only test of frequency property as a starting point so far

### DIFF
--- a/tests/instruments/agilent/test_agilentE4980A.py
+++ b/tests/instruments/agilent/test_agilentE4980A.py
@@ -1,0 +1,41 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2025 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


BenediktBurger:
Do you want to merge it with that single test (better than no test at all), or do you want to add more tests?

You can use the [test generator](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#test-generator) to easily generate protocol tests, if you have a device to use.

Falk3579:
I see that single test as a placeholder for later work.
The majority of E4980A remote commands are compatible with the Agilent 4284A commands. So most of the tests should be done for the 4284A. I'll try to modify the E4980A driver so that it inherits the 4284A properties.

BenediktBurger:
Ok. Do you want that single test to be merged yet?

Falk3579:
Yes, please merge it with the single test.

BenediktBurger:
Thanks for your work.

PR #1310, opened 2025-05-30
Feeltech fy3200s instrument driver

dvalde0:
A driver for a very cheap chinese AWG: https://chinese-electronics-products-tested.blogspot.com/p/fy3200s-function-generator-tested.html

### DIFF
--- a/pymeasure/instruments/feeltech/fy3200s.py
+++ b/pymeasure/instruments/feeltech/fy3200s.py
@@ -116,7 +116,7 @@ class MainChannel(Channel):
     duty_cycle = Instrument.control(
         "{ch}\ncd",
         "{ch}d%i",
-        "Set the duty cycle (percentage) of the channel",
+        "Control the duty cycle (percentage) of the channel",
         validator=step_validator(1),
         values=(1, 99),
         get_process=lambda x: float(x.strip("cd")),
@@ -243,7 +243,7 @@ class SubsidiaryChannel(Channel):
     duty_cycle = Instrument.control(
         "{ch}\ncd",
         "{ch}d%i",
-        "Set the duty cycle (percentage) of the channel",
+        "Control the duty cycle (percentage) of the channel",
         validator=step_validator(1),
         values=(1, 99),
         get_process=lambda x: float(x.strip("cd")),


pymeasure/instruments/feeltech/__init__.py
```diff
@@ -0,0 +1,25 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2024 PyMeasure Developers
```
BenediktBurger:
```suggestion
# Copyright (c) 2013-2025 PyMeasure Developers
```

pymeasure/instruments/feeltech/fy3200s.py
BenediktBurger, line 1: Please add copyright notice at the top (with the correct year).

pymeasure/instruments/feeltech/fy3200s.py
```diff
@@ -0,0 +1,232 @@
...
+        """Set the phase (deg) of the waveform""",
+        validator=step_validator(1),
+        values=(0, 360),
+    )
+
+
+class FY3200S(SCPIUnknownMixin, Instrument):
+    """Class to control the Feeltech FY3200S arbitrary waveform generator"""
+
+    def __init__(self, adapter: SerialAdapter, name: str = "FY3200s AWG", **kwargs):
```
BenediktBurger: The adapter can be a string or a subclass of `Adapter`!

pymeasure/instruments/feeltech/fy3200s.py
```diff
@@ -0,0 +1,232 @@
...
+
+    phase = Instrument.setting(
+        "{ch}p%i",
+        """Set the phase (deg) of the waveform""",
+        validator=step_validator(1),
+        values=(0, 360),
+    )
+
+
+class FY3200S(SCPIUnknownMixin, Instrument):
```
BenediktBurger: SCPIUnknownMixin is for devices, where we did not know, whether it supports SCPI or not.

If it supports SCPI, do use SCPIMixin. If it does not support SCPI (as it seems), do not use the Mixin, but use scpi_included=False (or similar) in the constructor call.

pymeasure/instruments/feeltech/fy3200s.py
```diff
@@ -0,0 +1,232 @@
...
+    """Class to control the Feeltech FY3200S arbitrary waveform generator"""
+
+    def __init__(self, adapter: SerialAdapter, name: str = "FY3200s AWG", **kwargs):
+        """Initialize the instrument with a SerialAdapter
+        :param adapter: SerialAdapter for the connection. The write termination has to be \\n
```
BenediktBurger: You can specify the write termination in the call to super init.
Afterwards you just write the address as "adapter" and the device does the correct configuration for you.

tests/instruments/feeltech/test_fy3200s.py
BenediktBurger, line 1: Thanks for adding tests.

pymeasure/instruments/feeltech/fy3200s.py
```diff
@@ -0,0 +1,232 @@
...
+    def load(self, pos: int):
+        if pos not in range(10):
+            raise ValueError("Load position must be an integer between 0 and 9")
+        else:
+            self.write("bl%i" % pos)
+
+    @property
+    def id(self):
+        """Get the identification of the instrument."""
+        return "Feeltech FY3200S arbitrary waveform generator"
```
BenediktBurger: If you do not read anything from the device, what is the reason for that property?
Typically the id gives the serial number etc.

docs/api/instruments/feeltech/fy3200s.rst
BenediktBurger, line 1: As a new contributor, you may add your name to `authors.txt` as well.

pymeasure/instruments/feeltech/fy3200s.py
```diff
@@ -0,0 +1,232 @@
...
+        "ARB1": 17,
+        "ARB2": 18,
+        "ARB3": 19,
+    }
+
+    SWEEPMODES = {"LINEAR": 0, "LOG": 1}
+
+    waveform = Instrument.setting(
+        "{ch}w%i",
+        "Set the waveform of the channel.",
```
BenediktBurger: Maybe hint at the Waveforms dictionary, such that a user knows, which values are available.

pymeasure/instruments/feeltech/fy3200s.py
```diff
@@ -0,0 +1,232 @@
...
+    )
+
+    def start_sweep(self):
+        self.write("{ch}r1")
+
+    def pause_sweep(self):
+        self.write("{ch}r0")
+
+
+class SubsidiaryChannel(Channel):
```
BenediktBurger, line 165: Are both classes the same one?
What are the differences?
Maybe you could base one Channel on the other (e.g. the Main one on the subsidiary on, if only the main one has a few more options).

If they are the same, a single Class would be sufficient, you distinguish the instances in the Instrument.

pymeasure/instruments/feeltech/fy3200s.py
dvalde0: Added in [e127e3b](https://github.com/pymeasure/pymeasure/pull/1310/commits/e127e3b498726aac5b11dbf9353caec200d9fcfa)

docs/api/instruments/feeltech/fy3200s.rst
dvalde0: Added in [e127e3b](https://github.com/pymeasure/pymeasure/pull/1310/commits/e127e3b498726aac5b11dbf9353caec200d9fcfa)

pymeasure/instruments/feeltech/fy3200s.py
dvalde0: Changed it to read from device in [e127e3b](https://github.com/pymeasure/pymeasure/pull/1310/commits/e127e3b498726aac5b11dbf9353caec200d9fcfa)

pymeasure/instruments/feeltech/fy3200s.py
dvalde0: Added in [e127e3b](https://github.com/pymeasure/pymeasure/pull/1310/commits/e127e3b498726aac5b11dbf9353caec200d9fcfa)

pymeasure/instruments/feeltech/__init__.py
dvalde0: Changed in [e127e3b](https://github.com/pymeasure/pymeasure/pull/1310/commits/e127e3b498726aac5b11dbf9353caec200d9fcfa)

pymeasure/instruments/feeltech/fy3200s.py
dvalde0: Changed in [e127e3b](https://github.com/pymeasure/pymeasure/pull/1310/commits/e127e3b498726aac5b11dbf9353caec200d9fcfa)

pymeasure/instruments/feeltech/fy3200s.py
dvalde0: Changed in [e127e3b](https://github.com/pymeasure/pymeasure/pull/1310/commits/e127e3b498726aac5b11dbf9353caec200d9fcfa)

pymeasure/instruments/feeltech/fy3200s.py
dvalde0: Changed in [e127e3b](https://github.com/pymeasure/pymeasure/pull/1310/commits/e127e3b498726aac5b11dbf9353caec200d9fcfa)

pymeasure/instruments/feeltech/fy3200s.py
dvalde0: The classes are not the same, the SubsidiaryChannel has the phase property that is not on Main. Also they have different waveforms. I am unsure whether bundling the common properties into a separate class and subclassing is useful here.

BenediktBurger: For me the Channels seem quite similar, so they could share code. DRY: Don't repeat yourself is generally a good practice as you have to change just one place, if you want to change something.

However, device definitions are more static than other code, such that it could stay that way.

pymeasure/instruments/feeltech/fy3200s.py
BenediktBurger: You did not adjust the docstring below.
Also, you can make the description shorter: `param str name` instead of two lines.

pymeasure/instruments/feeltech/fy3200s.py
```diff
@@ -0,0 +1,314 @@
...
+        "{ch}o%.1f",
+        "Set the offset (V) of the channel",
+        validator=step_validator(0.1),
+        values=(-10, 10),
+    )
+
+    duty_cycle = Instrument.control(
+        "{ch}\ncd",
+        "{ch}d%i",
+        "Set the duty cycle (percentage) of the channel",
```
BenediktBurger: You have a `control`, so please use also "Control", which indicates getting and setting:
```suggestion
        "Control the duty cycle (percentage) of the channel",
```

pymeasure/instruments/feeltech/fy3200s.py
```diff
@@ -0,0 +1,314 @@
...
```
BenediktBurger: Also here: Control.

With a single parent class, it would have been one changed line instead of two.

PR #1316, opened 2025-06-08
Fix flake8 errors in tests

Falk3579:
fix E721 and E231 errors


BenediktBurger:
Thanks for improving the code base.

Issue #1314, opened 2025-06-06
Setting measurement modes seperate from taking measurements on multimeter type devices, why?


Teslafly:
I am using the keithleyDAQ6510, which is basically a multimeter with a relay mux attached.
[keithleyDAQ6510](https://pymeasure.readthedocs.io/en/latest/api/instruments/keithley/keithleyDAQ6510.html)

I was a bit confused at first, because the examples for taking measurements in the documentation does not actually work:
```
keithley = KeithleyDAQ6510("GPIB::1")
keithley = KeithleyDAQ6510("TCPIP::192.168.1.1::INSTR")

print(keithley.current)                 # Prints the current in Amps
keithley.current_range = 10E-6          # Select the 10 uA range

print(keithley.voltage)                 # Prints the voltage in Volts
keithley.voltage_range = 100e-3         # Select the 100 mV range

print(keithley.resistance)              # Prints the resistance in Ohms
keithley.offset_compensated = "ON"      # Turns offset-compensated ohms on

keithley.open_channels([134, 135])      # Open channels 134 and 135 on the MUX card
keithley.close_channel(133)             # Close channel 133 on the MUX card
```

To take a voltage measurement, you need to actually do:
```
keithly.measure_voltage()      # configures the instrument to read a voltage
print(keithley.voltage)        # triggers and gets reading
```

It appears at some point, that pymeasure separated configuring an instrument for taking a specific measurement type, from actually taking the measurement. You can see that in instruments like [HP34401A.current_ac](https://pymeasure.readthedocs.io/en/latest/api/instruments/hp/hp34401A.html#pymeasure.instruments.hp.HP34401A.current_ac )

This is problematic in 2 facets. 

1. Taking a measurement of some value is an atomic function that deserves full abstraction
2. Because most instruments implement triggering and returning the measurement as just a `:READ?` command. 
If you do:
```
keithly.measure_voltage()      # configures the instrument to read a voltage
print(keithley.current)        # triggers and gets reading
```

`keithley.current` will actually return the voltage reading.

This is the implementation of `.current`
```
 current = Instrument.measurement(
        ":READ?",
        """ Measure the current in Amps, if configured for this reading.
        """
    )
```

I would like to understand why this was implemented this way, and then either: 

[...]

PR #1318, opened 2025-06-11
Refactor keithlydaq6510 and add diode measurement

Teslafly:
This is a somewhat significant refactor of the Keithly DAQ6510 driver.

This breaks backwards compatibility by removing some measurement functions that did not actually function as described.
Closes https://github.com/pymeasure/pymeasure/issues/1314 

This PR:

- Adds diode measurement functionality
- Removes `KeithleyDAQ6510.voltage`, `KeithleyDAQ6510.current`, `KeithleyDAQ6510.resistance` properties. This breaks backwards compatibility.
- Implements all measurement functions with full "setup + measure" functionality. 
- adds `configure_X_measurement` functions
- Adds awareness to expansion cards and mux channels as nested objects.
  - This means that:
    - `daq.mux_channels` will list all available channels provided through cards. Also a "front_panel" channel.
    - `daq.mux_channels[101].closed = True/False` will open/close channel or get state.
    - `daq.mux_channels[101].measure_voltage()` will measure the voltage or other parameter of that specific channel. And close the required measurement path automatically. 
  - `daq.cards["slot1"].id` gets model of expansion card. The card object also pulls number of mux channels, serial number, and other info.
- More to come.

### DIFF
--- a/pymeasure/instruments/keithley/keithleyDAQ6510.py
+++ b/pymeasure/instruments/keithley/keithleyDAQ6510.py
@@ -22,7 +22,7 @@
 # THE SOFTWARE.
 #
 
-from pymeasure.instruments import Instrument
+from pymeasure.instruments import Instrument, Channel
 from pymeasure.instruments.generic_types import SCPIMixin
 from pymeasure.instruments.validators import strict_discrete_set, truncated_range
 from .buffer import KeithleyBuffer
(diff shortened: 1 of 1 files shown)


pymeasure/instruments/siglenttechnologies/siglent_sdg_1032x.py
BenediktBurger, line 1: Do you really intend to add this new instrument in this PR?

pymeasure/instruments/keithley/keithley2200.py
BenediktBurger, line 1: What does this change have to do in the PR (regarding DAQ6510)?

pymeasure/instruments/keithley/keithley2200.py
```diff
@@ -25,20 +25,30 @@
 import logging
 from pymeasure.instruments import Instrument, Channel, SCPIUnknownMixin
 from pymeasure.instruments.validators import strict_discrete_set, strict_range
+from pymeasure.adapters import VISAAdapter                                          #To call adapter from driver
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 
+class VisaSplitAdapter(VISAAdapter):                                        #Buffer adapter call to split SCPI commands into 2 lines
```
BenediktBurger, line 34: Why do you need this special adapter?

You could also modify the `write/ask` methods of the instrument itself, which will work then for all adapters.

pymeasure/instruments/keithley/keithleyDAQ6510.py
```diff
@@ -33,15 +33,113 @@
...
+    """
+    OPEN = False
+    CLOSED = True
+
+
+class MuxChannel(Channel):
+    """mux relay channels"""
+
+    def __init__(self, chan_number):
+        self.chan_num = chan_number
```
BenediktBurger: Channels already have an id as input parameter.

Please use `super.__init__` call in the init, in order to use the configuration of the Channel parent class.

BenediktBurger (commented):
A few comments.

Teslafly:
This was extremely rough before and yes I have a lot of cleanup to do. including moving the other instrument to a different PR. 
Beginning to work on this again with some of your feedback.

PR #1325, opened 2025-07-02
Add Siglent oscilloscope SDS1000XHD

enlightralucas:
Hello,

This pull request introduces support for the Siglent SDS1000XHD series of oscilloscopes. I have implemented the core functionalities that were necessary for my specific use cases, focusing on the most common measurement and setup commands.

I am adding the official programming guide to facilitate future development : 
[SDS1000XHD_Series_ProgrammingGuide_EN11F.pdf](https://github.com/user-attachments/files/21015443/SDS1000XHD_Series_ProgrammingGuide_EN11F.pdf)

While the SDS1000XHD series offers a comprehensive range of features, this initial implementation is not exhaustive. There are many other functionalities that can be added to expand the instrument's capabilities within pymeasure. However, I will not be implementing these additional features myself.

### DIFF
--- a/pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
@@ -738,35 +738,34 @@ class MeasureChannel(Channel):
         get_process=lambda v: v.strip(),
     )
 
+    simple_item = Channel.setting(
+        ":MEASure:SIMPle:ITEM %s,%s",
+        """Set the simple measurement item.
+        
+        This command sets the type of simple measurement and its state.
+        Takes a tuple of (item, state) where:
+        - item (str): The measurement item to set. Valid values are:
+            PKPK, MAX, MIN, AMPL, TOP, BASE, LEVELX, CMEAN, MEAN,
+            STDEV, VSTD, RMS, CRMS, MEDIAN, CMEDIAN, OVSN, FPRE,
+            OVSP, RPRE, PER, FREQ, TMAX, TMIN, PWID, NWID, DUTY,
+            NDUTY, WID, NBWID, DELAY, TIMEL, RISE, FALL, RISE20T80,
+            FALL80T20, CCJ, PAREA, NAREA, AREA, ABSAREA, CYCLES,
+            REDGES, FEDGES, EDGES, PPULSES, NPULSES, PACArea,
+            NACArea, ACArea, ABSACArea
+        - state (str): The state of the measurement item (ON or OFF).
+        
+        Example:
+            # Add maximum measurement to the simple measurements window
+            measure.simple_item = ("MAX", "ON")
+            # Remove frequency measurement from the simple measurements window  
+            measure.simple_item = ("FREQ", "OFF")
+        """
+    )
+
     # Simple measurement methods
     def clear_simple(self):
         """Clear simple measurements."""
         self.write(":MEASure:SIMPle:CLEar")
-    
-    def set_simple_item(self, item, state="ON"):
-        """Set the simple measurement item.
-        
-        This command sets the type of simple measurement and its state.
-        
-        Args:
-            item (str): The measurement item to set. Valid values are:
-                PKPK, MAX, MIN, AMPL, TOP, BASE, LEVELX, CMEAN, MEAN,
-                STDEV, VSTD, RMS, CRMS, MEDIAN, CMEDIAN, OVSN, FPRE,
-                OVSP, RPRE, PER, FREQ, TMAX, TMIN, PWID, NWID, DUTY,
-                NDUTY, WID, NBWID, DELAY, TIMEL, RISE, FALL, RISE20T80,
-                FALL80T20, CCJ, PAREA, NAREA, AREA, ABSAREA, CYCLES,
-                REDGES, FEDGES, EDGES, PPULSES, NPULSES, PACArea,
-                NACArea, ACArea, ABSACArea
-            state (str): The state of the measurement item (ON or OFF).
-                        Defaults to "ON".
-        
-        Example:
-            # Add maximum measurement to the simple measurements window
-            measure.set_item("MAX", "ON")
-            # Remove frequency measurement from the simple measurements window  
-            measure.set_item("FREQ", "OFF")
-        """
-        self.write(f":MEASure:SIMPle:ITEM {item},{state}")
     
     # Advanced measurement methods
     def clear_advanced(self):
--- a/pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
+++ b/pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
@@ -23,6 +23,9 @@
 #
 
 
+import struct
+import math
+import gc
 from pymeasure.instruments import Channel, Instrument
 from pymeasure.instruments.generic_types import SCPIMixin
 from pymeasure.instruments.validators import (
@@ -30,7 +33,7 @@ from pymeasure.instruments.validators import (
 )
 
 
-class VoltageChannel(Channel):
+class AnalogChannel(Channel):
     """
     ===========================================================
     Implementation of a SIGLENT SDS1000xHD Oscilloscope channel
@@ -82,6 +85,16 @@ class VoltageChannel(Channel):
         get_process=lambda v: v.strip(),
     )
 
+    switch = Channel.control(
+        ":CHANnel{ch}:SWITch?",
+        ":CHANnel{ch}:SWITch %s",
+        "Control the channel display switch (ON/OFF).",
+        validator=strict_discrete_set,
+        values={True: "ON", False: "OFF"},
+        map_values=True,
+        get_process=lambda v: v.strip(),
+    )
+
     bandwidth_limit = Channel.control(
         ":CHANnel{ch}:BWLimit?",
         ":CHANnel{ch}:BWLimit %s",
@@ -98,7 +111,8 @@ class VoltageChannel(Channel):
         "Control signal inversion (ON/OFF).",
         validator=strict_discrete_set,
         values={True: "ON", False: "OFF"},
-        map_values=True,        get_process=lambda v: v.strip(),
+        map_values=True,
+        get_process=lambda v: v.strip(),
     )
 
     label = Channel.control(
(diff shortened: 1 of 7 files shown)


tests/instruments/siglenttechnologies/test_siglent_sds1000xhd.py
BenediktBurger, line 1: Thanks for adding tests.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+    ===========================================================
+    Implementation of a SIGLENT SDS1000xHD Oscilloscope channel
+    ===========================================================
+    """
+
+    scale = Channel.control(
+        ":CHANnel{ch}:SCALe?",
+        ":CHANnel{ch}:SCALe %.3e",
+        "Control the vertical scale of a channel in V/divisions.",
+        validator=truncated_range,
```
BenediktBurger: A warning regarding truncation: It is not transparent to the user. If he calls `inst.achannel.scale = 20`, it will be set to 10 V/division, but the user won't know this.

Therefore, a strict validator is better as it will tell, whenever the device does not do, what you told it to do.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        values=[1e-3, 10],
+        get_process=lambda v: float(v),
```
BenediktBurger: the result of `values` is typically a float, so your `get_process` is superfluous.

By the way, if you want to use a different casting, you could use the `cast` parameter, e.g. `cast=int`.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+    )
+
+    offset = Channel.control(
+        ":CHANnel{ch}:OFFSet?",
+        ":CHANnel{ch}:OFFSet %.6f",
+        "Control the vertical offset of the channel in volts.",
+        get_process=lambda v: float(v),
+    )
+
+    visible = Channel.control(
```
BenediktBurger: For boolean properties, we typically use a name like `visible_enabled`, as it makes it clear, how to use it.

Here, however, you could keep it, as `visible=True` is self explanatory

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        ":CHANnel{ch}:VISible?",
+        ":CHANnel{ch}:VISible %s",
+        "Control whether the channel is visible (ON/OFF).",
```
BenediktBurger: You should indicate the user input, i.e. True/False:
```suggestion
        "Control whether the channel is visible (bool).",
```

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        validator=strict_discrete_set,
+        values={True: "ON", False: "OFF"},
+        map_values=True,
+        get_process=lambda v: v.strip(),
+    )
+
+    switch = Channel.control(
```
BenediktBurger: What is this `switch`? Maybe `display_enabled`?

What is the difference to `visible`?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        ":CHANnel{ch}:SWITch?",
+        ":CHANnel{ch}:SWITch %s",
+        "Control the channel display switch (ON/OFF).",
+        validator=strict_discrete_set,
+        values={True: "ON", False: "OFF"},
+        map_values=True,
+        get_process=lambda v: v.strip(),
+    )
+
+    bandwidth_limit = Channel.control(
```
BenediktBurger: Here better: `bandwidth_limit_enabled`. And later (bool) in the docstring.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        get_process=lambda v: v.strip().strip('"\''),
+    )
+
+    unit = Channel.control(
+        ":CHANnel{ch}:UNIT?",
+        ":CHANnel{ch}:UNIT %s",
+        "Control the channel unit (V, A).",
+        validator=strict_discrete_set,
+        values=["V", "A"],
+        get_process=lambda v: v.strip(),
```
BenediktBurger: Is this strip necessary for all the commands?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        validator=strict_discrete_set,
+        values=["C1", "C2", "C3", "C4", "F1", "F2", "F3", "F4",
+                "D0", "D1", "D2", "D3", "D4", "D5", "D6", "D7",
+                "D8", "D9", "D10", "D11", "D12", "D13", "D14", "D15"],
+        get_process=lambda v: v.strip(),
+    )
+
+    @property
+    def source_with_auto_set(self):
+        """Get the source, automatically setting it if needed."""
```
BenediktBurger: For a property, only the getter docstring will be used, so you should write here `Control`
```suggestion
        """Control the source, automatically setting it if needed."""
```

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        self._ensure_source_set()
+        return self.source
+
+    @source_with_auto_set.setter
+    def source_with_auto_set(self, value):
+        """Set the source."""
```
BenediktBurger: This docstring does nothing.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+    start_point = Channel.control(
+        ":WAVeform:STARt?",
+        ":WAVeform:STARt %d",
+        """Control the starting data point for waveform transfer.
+        This command specifies the starting data point for waveform transfer
+        using the query :WAVeform:DATA?. The value range is related to the
+        current waveform point and the value set by the command :WAVeform:POINt.
+        Value in NR1 format (integer with no decimal point).""",
+        validator=truncated_range,
+        values=[0, 1000000000],  # Range depends on current waveform points and :WAVeform:POINt
```
BenediktBurger: You do not need a validator (and values) at all.

So, you could
- either remove values/validator
- or make this a [dynamic property](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#dynamic-properties) and modify the range on the fly.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        get_process=lambda v: int(v),
```
BenediktBurger:
```suggestion
        cast=int,
```

Also below...

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+    @source_with_auto_set.setter
+    def source_with_auto_set(self, value):
+        """Set the source."""
+        self.source = value
+
+    start_point = Channel.control(
+        ":WAVeform:STARt?",
+        ":WAVeform:STARt %d",
+        """Control the starting data point for waveform transfer.
+        This command specifies the starting data point for waveform transfer
```
BenediktBurger: According to Python standards for docstrings, you have an empty line between the one line summary and the additional text.
```suggestion
        """Control the starting data point for waveform transfer.

        This command specifies the starting data point for waveform transfer
```

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        values=["BYTE", "WORD"],
+        get_process=lambda v: v.strip(),
+    )
+
+    def _parse_preamble_descriptor(self, raw_data):
+        """Parse the binary preamble descriptor data.
+        Args:
+            raw_data (bytes): Raw binary data from :WAVeform:PREamble? command
+        Returns:
+            dict: Parsed descriptor data with the same format as get_descriptor
```
BenediktBurger: for sphinx we use a different syntax:
```suggestion
        :param bytes raw_data: Raw binary data from :WAVeform:PREamble? command
        :return: Parsed descriptor data with the same format as get_descriptor (dict)
```

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+        """
+        try:
+            # Find the start of the binary data after the header
+            if isinstance(raw_data, str):
+                raw_data = raw_data.encode('latin-1')
```
BenediktBurger: Typically, we use `"` for strings.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1287 @@
...
+           self._auto_source_id in ["C1", "C2", "C3", "C4", "F1", "F2", "F3", "F4",
+                                    "D0", "D1", "D2", "D3", "D4", "D5", "D6", "D7",
+                                    "D8", "D9", "D10", "D11", "D12", "D13", "D14", "D15"]:
+            try:
+                # Always set the source to the channel ID without checking current value
+                # This avoids potential issues with reading the current source state
+                self.source = self._auto_source_id
+            except Exception:
+                # If setting source fails, continue without error
+                pass
```
BenediktBurger: Isn't that dangerous?
Maybe add logging that you skip it?

At least you should limit this "error ignoring" to specific Exceptions and not all possible Exceptions.

tests/instruments/siglenttechnologies/test_siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1752 @@
+import struct
```
BenediktBurger: Please add the copyright notice at the top of the file.

tests/instruments/siglenttechnologies/test_siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1752 @@
...
+import math
+import pytest
+from pymeasure.test import expected_protocol
```
BenediktBurger, line 30: It's good to separate imports: standard library, external packages, local packages:

```suggestion
import math

import pytest
from pymeasure.test import expected_protocol
```

tests/instruments/siglenttechnologies/test_siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1752 @@
...
+            [(b':MEASure:ADVanced:P1:STATistics? ALL', b'CUR,1.234E+00,MEAN,1.500E+00')],
+    ) as inst:
+        result = inst.measure.advanced_p1.statistics_all
+        # The result may be parsed as a list or returned as a string
+        # depending on the framework's processing
+        if isinstance(result, list):
+            # Check that it contains the expected elements
+            assert 'CUR' in result and 1.234 in result and 'MEAN' in result and 1.5 in result
+        else:
+            assert result == 'CUR,1.234E+00,MEAN,1.500E+00'
```
BenediktBurger: why does it vary?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+    def _parse_preamble_descriptor(self, raw_data):
+        """Parse the binary preamble descriptor data.
+        Args:
+            raw_data (bytes): Raw binary data from :WAVeform:PREamble? command
+        Returns:
+            dict: Parsed descriptor data with the same format as get_descriptor
+        """
+        # Find the start of the binary data after the header
+        if isinstance(raw_data, str):
+            raw_data = raw_data.encode('latin-1')
```
BenediktBurger: why latin-1?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+            'tdiv': tdiv_val,
+            'vcode_per': vcode_per,
+            'adc_bit': adc_bit_val,
+            'probe': probe_val
+        }
+
+    @property
+    def preamble(self):
+        """Get the waveform preamble descriptor data.
+        The preamble contains binary descriptor information about the waveform data format,
```
BenediktBurger, line 259: Python recommends an empty line between summary and description
```suggestion
        """Get the waveform preamble descriptor data.

        The preamble contains binary descriptor information about the waveform data format,
```

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        including voltage scale, offset, time interval, trigger delay, time division,
+        voltage codes per division, ADC bit depth, and probe attenuation factor.
+        This uses the same binary parsing approach as the get_descriptor method.
+        Returns:
+        dict: A dictionary with the following keys:
+        - vdiv: Voltage per division (float)
```
BenediktBurger: I guess that this does not render as you expect it:
- You need an empty line before returns and it should be `:return:` for sphinx syntax.
- you duplicate the return type dict

If you want to look at the result, you may execute `make html` in the docs folder and look at the generated html documentation.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+            block_start = recv_rtn.find(b'#')
+            data_digit = int(recv_rtn[block_start + 1:block_start + 2])
+            data_start = block_start + 2 + data_digit
+            recv_byte += recv_rtn[data_start:]
+
+        # Unpack signed byte data
+        if adc_bit > 8:
+            # Calculate actual length of data received and adjust points if needed
+            actual_points = len(recv_byte) // 2  # 2 bytes per point for WORD format
+            # print(f"Expected points: {int(points)}, Actual points from data: {actual_points}")
```
BenediktBurger: Please remove these debugging parts `#print...`

If you want to add logging, you may use the python logging module.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+                actual_points = len(recv_byte) // 2
+
+            convert_data = struct.unpack("%dh" % actual_points, recv_byte)
+        else:
+            # Calculate actual length of data received and adjust points if needed
+            actual_points = len(recv_byte)  # 1 byte per point for BYTE format
+            # print(f"Expected points: {int(points)}, Actual points from data: {actual_points}")
+            convert_data = struct.unpack("%db" % actual_points, recv_byte)
+        del recv_byte
+        gc.collect()
```
BenediktBurger: Is it necessary? After the method has returned, recv_byte is out of scope and will be garbage collected shortly after.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        # Constants - same as reference script
+        HORI_NUM = 10
+
+        # Set up waveform source and get preamble information
+        self.start_point = 0  # Reset start point to 0
+
+        # Get waveform descriptor information using the preamble property
+        preamble_data = self.preamble
+        vdiv = preamble_data['vdiv']
+        ofst = preamble_data['voffset']
```
BenediktBurger: Why don't you call the variable `offset`? That isn't too long either.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        validator=strict_discrete_set,
+        values=["C1", "C2", "C3", "C4", "F1", "F2", "F3", "F4",
+                "D0", "D1", "D2", "D3", "D4", "D5", "D6", "D7",
+                "D8", "D9", "D10", "D11", "D12", "D13", "D14", "D15"],
+        get_process=lambda v: v.strip(),
+    )
+
+    statistics_all = Channel.measurement(
+        ":MEASure:ADVanced:P{ch}:STATistics? ALL",
+        "Returns all the statistics",
```
BenediktBurger: It's a measurement, so it should be:

```suggestion
        "Get all the statistics",
```
(or "Measure", but that does not fit here that well)

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        get_process=lambda v: v.strip(),
+    )
+
+    # Individual statistics measurements
+    statistics_all = Channel.measurement(
+        ":MEASure:ADVanced:P{ch}:STATistics? ALL",
```
BenediktBurger, line 404: That's the same command as above, isn't it?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        """Get all statistics for the advanced measurement item.
+        Returns all statistical data when statistics are enabled, or 'OFF' when disabled.
+        """,
+        get_process=lambda v: v.strip(),
+    )
+
+    statistics_current = Channel.measurement(
+        ":MEASure:ADVanced:P{ch}:STATistics? CURRent",
+        """Get the current value of the statistics for the advanced measurement item.
+        Returns the current statistical value in NR3 format (e.g., 1.23E+2).
```
BenediktBurger: You do not duplicate the description:

```suggestion
        """Get the current value of the statistics for the advanced measurement item in NR3 format (e.g., 1.23E-2).
```

What is NR3 format?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        get_process=lambda v: v.strip(),
+    )
+
+    # Individual statistics measurements
+    statistics_all = Channel.measurement(
+        ":MEASure:ADVanced:P{ch}:STATistics? ALL",
+        """Get all statistics for the advanced measurement item.
+        Returns all statistical data when statistics are enabled, or 'OFF' when disabled.
+        """,
+        get_process=lambda v: v.strip(),
```
BenediktBurger: I think it should be part of `values` which is called during readout.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+    )
+
+    mode = Channel.control(
+        ":MEASure:MODE?",
+        ":MEASure:MODE %s",
+        "Control the measurement mode (SIMPle or ADVanced).",
+        validator=strict_discrete_set,
+        values=["SIMPle", "ADVanced"],
+        get_process=lambda v: v.strip(),
+    )
```
BenediktBurger: Why don't you make it a boolean property: `advanced_mode_enabled`, that way you do not have to worry about the strings as a user, but can use a simple boolean.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        validator=truncated_range,
+        values=[1, 12],
+        get_process=lambda v: int(v),
+    )
+
+    advanced_measurements = Channel.MultiChannelCreator(
+        AdvancedMeasurementItem,
+        list(range(1, 13)),  # P1 through P12
+        prefix="advanced_p"
+    )
```
BenediktBurger: Please put these child definitions (together with the __init__) at the top of the class.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        The default precision is 3 digits, maximum valid precision is 7 digits.
+        Use ":FORMat:DATA" command to set the data precision.
+        """,
+        get_process=lambda v: float(v.strip()) if isinstance(v, str) else float(v),
+    )
+
+    mode = Channel.control(
+        ":TRIGger:MODE?",
+        ":TRIGger:MODE %s",
+        """Control the trigger mode.""",
```
BenediktBurger: It's good to specify the options

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        validator=strict_discrete_set,
+        values=["FAST", "SLOW"],
+        get_process=lambda v: v.strip(),
+    )
+
+    acq_interpolation = Instrument.control(
+        ":ACQuire:INTerpolation?",
+        ":ACQuire:INTerpolation %s",
+        "Control acquisition interpolation method (ON selects sinx/x (sinc) "
+        "interpolation, OFF selects linear interpolation).",
```
BenediktBurger: Maybe also as a boolean? `acq_interpolation_enabled`?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+         "D8", "D9", "D10", "D11", "D12", "D13", "D14", "D15"],
+        prefix="wf_"
+    )
+
+    # Measurement subsystem
+    measure = Instrument.ChannelCreator(MeasureChannel, "")
+    # Trigger subsystem
+    trigger = Instrument.ChannelCreator(TriggerChannel, "")
+
+    acq_acquisition_mode = Instrument.control(
```
BenediktBurger: `acq` means acquisition? You prefix it to group the methods?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1110 @@
...
+        ":ACQuire:INTerpolation %s",
+        "Control acquisition interpolation method (ON selects sinx/x (sinc) "
+        "interpolation, OFF selects linear interpolation).",
+        validator=strict_discrete_set,
+        values={True: "ON", False: "OFF"},
+        map_values=True,
+        get_process=lambda v: v.strip(),
+    )
+
+    acq_memory_mgmt = Instrument.control(
```
BenediktBurger: Consider spelling the abbreviation out for better clarity: management

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: Yes exactly. Do you recommend a better way ?

BenediktBurger: Looking again at the methods, I do not think, that you need that prefix at all.

In general, it can be sometimes useful to use such a prefix, but they can also make it more complicated.

Below with the timebase, the prefixes are fine as they are part of the "name". E.g. plot_mode makes sense even without that acq prefix, while timebase_window without timebase would not.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: done

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: I added the difference in the docs

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: thanks. made proper cleaning

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: thank you

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: removed auto_source_id completely

tests/instruments/siglenttechnologies/test_siglent_sds1000xhd.py
lucasbraud: made proper modifications, thank you.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: not necessary indeed.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: You're absolutely right about the prefixes. I've refactored the code to use an AcquisitionChannel class for consistency with PyMeasure's architecture. This eliminates the need for acq_ prefixes while keeping related functionality properly grouped. Now users access acquisition settings via instrument.acquisition.mode, instrument.acquisition.type, etc., which is much cleaner and follows the same pattern as TimebaseChannel, TriggerChannel, and MeasureChannel.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: thank you. I fixed all methods to follow sphinx format.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: thank you, it was redundant indeed. my apologies.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: Fixed the duplicate description. NR3 is IEEE 488.2 floating-point format (e.g., 1.23E+2), but I've updated docstrings to focus on practical return values (frequency in Hz, voltage in V) rather than protocol details for better user experience.

tests/instruments/siglenttechnologies/test_siglent_sds1000xhd_with_device.py
```diff
@@ -0,0 +1,306 @@
...
+    # Test channel 1 settings
+    sds1000xhd.channel_1.coupling = "DC"
+    assert sds1000xhd.channel_1.coupling == "DC"
+
+    sds1000xhd.channel_1.coupling = "AC"
+    assert sds1000xhd.channel_1.coupling == "AC"
+
+    # Test scale (vertical sensitivity)
+    sds1000xhd.channel_1.scale = 0.5  # 500mV/div
+    assert abs(sds1000xhd.channel_1.scale - 0.5) < 0.01
```
BenediktBurger: Pytest offers a great tool to do such a comparison: https://docs.pytest.org/en/stable/reference/reference.html#pytest.approx

pyrightconfig.json
BenediktBurger, line 1: Please remove this file from the PR.

.gitignore
BenediktBurger, line 1: I see you used claude code. What's your experience?

I'm unsure, whether we want to add all those different files to the gitignore (or add them into the repo itself 🤔 ).

What do you think about adding an [Agents.md](https://agents.md) file and symlinks from claude.md etc. to that file?

tests/instruments/siglenttechnologies/test_siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1737 @@
...
+        assert callable(inst.clear_sweeps_acq)
+
+        # Verify waveform channels exist
+        assert hasattr(inst, 'wf_C1')
+        assert hasattr(inst, 'wf_C2')
+        assert hasattr(inst, 'wf_F1')
+        assert hasattr(inst, 'wf_D0')
+
+        # Verify measurement and trigger exist
+        assert hasattr(inst, 'measure')
```
BenediktBurger: At least some of these tests are already done indirectly. For example, you already set properties of the `trigger` , therefore, the trigger has to exist.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        :param str measurement_type: The measurement type to query. Valid values are:
+            PKPK, MAX, MIN, AMPL, TOP, BASE, LEVELX, CMEAN, MEAN,
+            STDEV, VSTD, RMS, CRMS, MEDIAN, CMEDIAN, OVSN, FPRE,
+            OVSP, RPRE, PER, FREQ, TMAX, TMIN, PWID, NWID, DUTY,
+            NDUTY, WID, NBWID, DELAY, TIMEL, RISE, FALL, RISE20T80,
+            FALL80T20, CCJ, PAREA, NAREA, AREA, ABSAREA, CYCLES,
+            REDGES, FEDGES, EDGES, PPULSES, NPULSES, PACArea,
+            NACArea, ACArea, ABSACArea, ALL
+
+            Note: ALL returns all measurement values of all measurement types
```
BenediktBurger: A suggestion to make it clear, that it is a value:
```suggestion
            Note: "ALL" returns all measurement values of all measurement types
```

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+            except for delay measurements.
+        :return: The measurement value in NR3 format (e.g., 1.23E+2) for
+            individual measurements, or a string containing all values
+            when measurement_type is "ALL"
+        :rtype: float or str
+        """
+        response = self.ask(f":MEASure:SIMPle:VALue? {measurement_type}")
```
BenediktBurger: Did you consider using `values` to get the values? It can split the data and convert it to float (etc.)?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        if measurement_type == "ALL":
+            return response.strip()
+        else:
+            return float(response.strip()) if isinstance(response, str) else float(response)
```
BenediktBurger: What type do you expect, if not string? ask always returns a string!

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+    This class provides comprehensive acquisition control for the SDS1000xHD oscilloscope
+    including acquisition modes, memory management, sampling, and data format settings.
+    """
+
+    mode = Channel.control(
+        ":ACQuire:AMODe?",
+        ":ACQuire:AMODe %s",
+        """Control sets the rate of waveform capture.
+        This command can provide a high-speed waveform capture rate to help capture signal
+        anomalies""",
```
BenediktBurger: That docstring is hard to understand.

Maybe something like `"Control whether to use the fast or slow capture rate"`.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+
+
+class AcquisitionChannel(Channel):
+    """
+    Acquisition channel for SDS1000xHD oscilloscope.
+    This class provides comprehensive acquisition control for the SDS1000xHD oscilloscope
+    including acquisition modes, memory management, sampling, and data format settings.
+    """
+
+    mode = Channel.control(
```
BenediktBurger: I would make it boolean: `fast_mode_enabled` or something similar, then you can use mapping and boolean values instead of strings.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        values=["YT", "XY", "ROLL"],
+    )
+
+    memory_depth = Channel.control(
+        ":ACQuire:MDEPth?",
+        ":ACQuire:MDEPth %s",
+        """Control the memory depth for acquisition (str).
+        Sets the maximum number of sample points that can be stored.""",
+        validator=strict_discrete_set,
+        values=["AUTO", "14K", "140K", "1.4M", "14M", "AUTO"],
```
BenediktBurger: You have "AUTO" twice.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        validator=strict_discrete_set,
+        values=["LAST_TRIG", "ACQ_START"],
+    )
+
+    edge_impedance = Channel.control(
+        ":TRIGger:EDGE:IMPedance?",
+        ":TRIGger:EDGE:IMPedance %s",
+        """Control the impedance of the edge trigger source.""",
+        validator=strict_discrete_set,
+        values=["ONEMeg", "FIFTy"],
```
BenediktBurger: See above regarding mixed case values.

Actually, here, you could change the property to `edge_high_impedance = True/False`, which is boolean.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        values=[8e-9, 30],
+        cast=float,
+    )
+
+    edge_hld_off = Channel.control(
+        ":TRIGger:EDGE:HOLDoff?",
+        ":TRIGger:EDGE:HOLDoff %s",
+        """Control the holdoff type for edge trigger.""",
+        validator=strict_discrete_set,
+        values=["OFF", "EVENts", "TIME"],
```
BenediktBurger: These mixed case strings are hard to memorize.
I know, the manual states such a string as you can use the short or long form.
However, they are more difficult to read/write.

What does the device return?
What does the device accept? -> I guess case insensitive.

What do you think about a `strict_discrete_case_insensitive_set` validator, which does not validate the case?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+from pymeasure.instruments import Channel, Instrument
+from pymeasure.instruments.generic_types import SCPIMixin
+from pymeasure.instruments.validators import (
+    truncated_discrete_set, strict_range, strict_discrete_set
+)
+
+
+class AnalogChannel(Channel):
+    """
+    ===========================================================
```
BenediktBurger: In the documentation, it is already marked as a class, so no need to use that additional marking as a header.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        validator=strict_discrete_set,
+        values=["V", "A"],
+    )
+
+
+class WaveformChannel(Channel):
+    """
+    Waveform channel for SDS1000XHD oscilloscope.
+    ===========================================================
+    This class provides methods to retrieve waveform data from the oscilloscope.
```
BenediktBurger: Also here, no need to add the header part.
In case you left it, make sure, that it is as long as the text you underline.

Finally, docstrings have an empty line between the first summary line and the body of the docstring (according to PEP 257).

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+    """
+    Acquisition channel for SDS1000xHD oscilloscope.
+    This class provides comprehensive acquisition control for the SDS1000xHD oscilloscope
+    including acquisition modes, memory management, sampling, and data format settings.
+    """
+
+    mode = Channel.control(
+        ":ACQuire:AMODe?",
+        ":ACQuire:AMODe %s",
+        """Control sets the rate of waveform capture.
```
BenediktBurger: "Control sets..."?

```suggestion
        """Control the rate of waveform capture.
```

Control is a verb.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        Sets the maximum number of sample points that can be stored.""",
+        validator=strict_discrete_set,
+        values=["AUTO", "14K", "140K", "1.4M", "14M", "AUTO"],
+    )
+
+    count = Channel.control(
+        ":ACQuire:NUMacq?",
+        ":ACQuire:NUMacq %d",
+        """Control the number of acquisitions for averaging mode (int).
+        Valid range: 1 to 1000000""",
```
BenediktBurger: In the docstrings example we used a more compact version, even including the validator type:

```suggestion
        """Control the number of acquisitions for averaging mode (int strictly between 1 and 1000000).""",
```

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        validator=strict_range,
+        values=[1, 1000000],
+        cast=int,
+    )
+
+    points = Channel.measurement(
+        ":ACQuire:POINts?",
+        "Get the number of sampled points of the current waveform on the screen (int).",
+        cast=lambda x: int(float(x)),
```
BenediktBurger: Why do you cast it to float first?
Would `cast=int` be sufficient?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+    )
+
+    resolution = Channel.control(
+        ":ACQuire:RESolution?",
+        ":ACQuire:RESolution %s",
+        "Control the vertical resolution for digitizing (str).",
+        validator=strict_discrete_set,
+        values=["8Bits", "10Bits"],
```
BenediktBurger: Maybe as an integer (`resolution_bits = 8`) or even boolean (`resolution_high_enabled`) property?
In the latter case, the docstring has to state what high/low means (10/8).

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        values={True: "ON", False: "OFF"},
+        map_values=True,
+    )
+
+    sequence_count = Channel.control(
+        ":ACQuire:SEQuence:COUNt?",
+        ":ACQuire:SEQuence:COUNt %d",
+        """Control the number of segments for sequence acquisition (int).
+        This command sets the number of acquisition segments when sequence mode is enabled.
+        Valid range: 1 to 1000.""",
```
BenediktBurger:
```suggestion
        """Control the number of acquisition segments for sequence mode (int strictly between 1 and 1000).""",
```

Can be all in a simple sentence.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        validator=strict_range,
+        values=[1, 1000],
+        cast=int,
+    )
+
+    sample_rate = Channel.measurement(
+        ":ACQuire:SRATe?",
+        "Get the current sample rate in samples per second (float).",
+        get_process=lambda v: float(v.strip()) if isinstance(v, str) else float(v),
```
BenediktBurger: `float`  is already able to process whitespaces: `float("   123.456   ") == 123.456`.
Therefore, `values` will already return a float and get_process casts that float again to float.
So, no need for `get_process` here.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+    scale = Channel.control(
+        ":TIMebase:SCALe?",
+        ":TIMebase:SCALe %.6e",
+        """Control the horizontal scale per division for the main window (float).
+        Note: Due to the limitation of the expansion strategy, when the time base is set
+        from large to small, it will automatically adjust to the minimum time base that
+        can be set currently.""",
+        validator=strict_range,
+        values=[200e-12, 1000],
+        cast=float,
```
BenediktBurger: float is the default cast

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+    """
+
+    frequency = Channel.measurement(
+        ":TRIGger:FREQuency?",
+        docs="""Get the current trigger frequency from the hardware frequency counter.
+        Returns the value of hardware frequency counter in hertz if available.
+        The default precision is 3 digits, maximum valid precision is 7 digits.
+        Use ":FORMat:DATA" command to set the data precision.
+        """,
+        get_process=lambda v: float(v.strip()) if isinstance(v, str) else float(v),
```
BenediktBurger: See above

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        - FTRIG: Force trigger to acquire a frame regardless of trigger conditions.
+        """,
+        validator=strict_discrete_set,
+        values=["AUTO", "NORMal", "SINGle", "FTRIG"],
+    )
+
+    status = Channel.measurement(
+        ":TRIGger:STATus?",
+        docs="""Get the current trigger status.
+        Returns the current trigger state such as STOP, READY, ARM, TD, WAIT, etc.""",
```
BenediktBurger: Both sentences are almost the same, so you can combine them:

```suggestion
        docs="""Get the current trigger status, such as STOP, READY, ARM, TD, WAIT, etc.""",
```

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        """Control the coupling mode of the edge trigger.""",
+        validator=strict_discrete_set,
+        values=["DC", "AC", "LFREJect", "HFREJect"],
+    )
+
+    edge_hld_event = Channel.control(
+        ":TRIGger:EDGE:HLDEVent?",
+        ":TRIGger:EDGE:HLDEVent %d",
+        """Control the number of holdoff events for edge trigger.
+        This command sets the number of holdoff events of the edge trigger.
```
BenediktBurger: Sets it? I thought it controls (sets/gets) it?

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+        ":TRIGger:EDGE:LEVel?",
+        ":TRIGger:EDGE:LEVel %.6e",
+        """Control the trigger level of the edge trigger.
+        This command sets the trigger level of the edge trigger in volts.
+        The query returns the current trigger level value of the edge trigger.
+        For SDS1000X HD models, the range is:
+        [-4.1*vertical_scale-vertical_offset, 4.1*vertical_scale-vertical_offset]
+        See programming manual for other models.
+        """,
+        get_process=lambda v: float(v.strip()) if isinstance(v, str) else float(v),
```
BenediktBurger: see above

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+    ==============================================
+    Represents the SIGLENT SDS1000xHD Oscilloscope
+    ==============================================
+    The SDS1000X-HD series are high-definition oscilloscopes with enhanced
+    measurement capabilities and improved user interface. This implementation
+    supports common oscilloscope operations including waveform acquisition,
+    measurements, and trigger control.
+    """
+
+    # Child definitions moved to top per maintainer feedback
```
BenediktBurger: Please remove that comment as it does not provide any information for future readers.
We wrote it in the documentation, that the init should be on top.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
```diff
@@ -0,0 +1,1064 @@
...
+    def __init__(self, adapter, name="Siglent SDS1000xHD Oscilloscope", timeout=2000,
+                 chunk_size=20*1024*1024, **kwargs):
+        super().__init__(adapter, name, timeout=timeout, chunk_size=chunk_size, **kwargs)
+
+    def auto_setup(self):
+        """Perform automatic setup of the oscilloscope."""
+        self.write(":AUToset")
+
+    def clear_sweeps_acq(self):
+        """Clear accumulated sweeps using ACQuire command."""
```
BenediktBurger: The user won't need the hint at the ACQ command, but more, what it means. So you can keep it short at: `Clear accumulated sweeps.` or a little more.

Basically, the user should read the docstrings and understand how to use the device without need for the manual.

.gitignore
lucasbraud: Claude Code has been very effective for this project because:
  1. Structured Guidance: The CLAUDE.md file provides excellent context about PyMeasure's architecture, testing
   patterns, and development commands, which helps maintain consistency with project conventions
  2. Code Quality: It automatically follows the established patterns (using strict_discrete_set vs
  truncated_discrete_set, proper docstring formats, etc.) and runs the appropriate linting/testing commands
  3. Maintainer Feedback Integration: It systematically addresses your code review feedback and tracks progress
  through todo lists

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: A proper case-insensitive validator would be cleaner and more comprehensive.
  The validator could:
  1. Accept any case from user input
  2. Normalize to device format (e.g., uppercase) for sending
  3. Handle any case in device responses
  4. Provide clear error messages if the value doesn't match (ignoring case)

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
lucasbraud: No, `cast=int` is not sufficient because the oscilloscope returns the value in scientific notation format as a string (e.g., `'1.00E+07'`). When I tested with the device, acquisition.points returned '1.00E+07' as a string. Python's `int() f`unction cannot directly convert scientific notation strings to integers:
`int('1.00E+07')  # Raises ValueError: invalid literal for int()`
However, `float()` can parse scientific notation, and then we convert to int:
`int(float('1.00E+07'))  # Works! Returns 10000000`
This is why I need `get_process=lambda v: int(float(v))` instead of just `cast=int`. The two-step conversion handles the scientific notation format that the instrument returns.

enlightralucas:
Hi @BenediktBurger,
Thank you for the detailed feedback. I have addressed all the issues you raised.

Agents.md
BenediktBurger, line 1: Please remove this file for now.
In #1362 we will discuss, whether to add such a file (I'm not able to decide that alone).
You may add the content of this file there as a suggestion.

Writing this file is too important to be part of another PR.

Sorry for the confusion, my comment was just meant to start a discussion, not for you adding already the file here.

.gitignore
BenediktBurger, line 1: We have to think, whether we want to check in the symlinks as well. Let us leave these changes for #1362.

pymeasure/instruments/siglenttechnologies/siglent_sds1000xhd.py
BenediktBurger: Ok, valid point.

PR #1329, opened 2025-07-04
Add support for Spellman XRV high voltage power supplies

Falk3579:
These power supplies are used to drive X-ray tubes.

### DIFF
--- a/pymeasure/instruments/spellmanhv/spellmanXRV.py
+++ b/pymeasure/instruments/spellmanhv/spellmanXRV.py
@@ -408,9 +408,7 @@ class SpellmanXRV(Instrument):
 
     baudrate = Instrument.setting(
         "07,%d",
-        """Set the baud rate.
-
-        :type: int, strictly in ``9600``, ``19200``, ``38400``, ``57600``, ``115200``
+        """Set the baud rate (int, strictly in ``9600``, ``19200``, ``38400``, ``57600``, ``115200``).
         """,
         validator=strict_discrete_set,
         map_values=True,
--- a/pymeasure/instruments/spellmanhv/spellmanXRV.py
+++ b/pymeasure/instruments/spellmanhv/spellmanXRV.py
@@ -233,6 +233,9 @@ class SpellmanXRV(Instrument):
 
     checksum_enabled = True  # RS232 and USB
 
+    filament = Instrument.ChannelCreator(Filament)
+    unscaled = Instrument.ChannelCreator(UnscaledData)
+
     def __init__(self, adapter,
                  name="Spellman XRV HV Power Supply",
                  query_delay=0.15,
@@ -402,10 +405,6 @@ class SpellmanXRV(Instrument):
             "preheat": int(v[6]),
             "anode_current": adc_amps_per_bit*int(v[7]),
             }
-
-    filament = Instrument.ChannelCreator(Filament)
-
-    unscaled = Instrument.ChannelCreator(UnscaledData)
 
     baudrate = Instrument.setting(
         "07,%d",
--- a/pymeasure/instruments/spellmanhv/spellmanXRV.py
+++ b/pymeasure/instruments/spellmanhv/spellmanXRV.py
@@ -255,7 +255,8 @@ class SpellmanXRV(Instrument):
 
         self.set_scaling()
 
-    def checksum(self, string_to_check):
+    @staticmethod
+    def checksum(string_to_check):
         """Calculate the checksum.
 
         :param string_to_check: string to calculate the checksum from
--- a/pymeasure/instruments/spellmanhv/spellmanXRV.py
+++ b/pymeasure/instruments/spellmanhv/spellmanXRV.py
@@ -409,7 +409,8 @@ class SpellmanXRV(Instrument):
 
     baudrate = Instrument.setting(
         "07,%d",
-        """Set the baud rate (int, strictly in ``9600``, ``19200``, ``38400``, ``57600``, ``115200``).
+        """
+        Set the baud rate (int, strictly in ``9600``, ``19200``, ``38400``, ``57600``, ``115200``).
         """,
         validator=strict_discrete_set,
         map_values=True,
--- a/pymeasure/instruments/spellmanhv/spellmanXRV.py
+++ b/pymeasure/instruments/spellmanhv/spellmanXRV.py
@@ -330,7 +330,6 @@ class SpellmanXRV(Instrument):
 
             if got_checksum is not calculated_checksum:
                 string = f"Checksum error: expected '{calculated_checksum}', got '{got_checksum}'."
-                print(string)
                 raise ConnectionError(string)
 
         return response[0].partition(",")[2]  # remove command from response
--- a/pymeasure/instruments/spellmanhv/spellmanXRV.py
+++ b/pymeasure/instruments/spellmanhv/spellmanXRV.py
@@ -1,0 +1,622 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2025 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Instrument, Channel
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+from enum import IntFlag
+from pyvisa.constants import InterfaceType
+
+# https://www.spellmanhv.com/en/high-voltage-power-supplies/XRV
+
+
+class StatusCode(IntFlag):
+    HV_ENABLED = 2**0
+    INTERLOCK_1_CLOSED = 2**1
+    INTERLOCK_2_CLOSED = 2**2
+    ECR_MODE_ACTIVE = 2**3  # Emission Current Regulation
+    POWER_SUPPLY_FAULT = 2**4
+    LOCAL_MODE = 2**5
+    FILAMENT_ENABLED = 2**6
+    LARGE_FILAMENT = 2**7
+    XRAYS_EMINENT = 2**8
+    LARGE_FILAMENT_CONFIRMATION = 2**9
+    SMALL_FILAMENT_CONFIRMATION = 2**10
+    RESERVED1 = 2**11
+    RESERVED2 = 2**12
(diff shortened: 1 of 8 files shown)


pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,622 @@
...
+        response = got.strip(self.STX).rpartition(",")
+
+        if self.checksum_enabled:
+            string_to_check = response[0] + response[1]
+            calculated_checksum = self.checksum(string_to_check)
+            got_checksum = response[2]
+
+            if got_checksum is not calculated_checksum:
+                string = f"Checksum error: expected '{calculated_checksum}', got '{got_checksum}'."
+                print(string)
```
BenediktBurger: Do not use print for logging, but the logging package (python standard library).

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,622 @@
...
+            "voltage_setpoint": int(v[3]),
+            "current_setpoint": int(v[4]),
+            "limit": int(v[5]),
+            "preheat": int(v[6]),
+            "anode_current": adc_amps_per_bit*int(v[7]),
+            }
+
+    filament = Instrument.ChannelCreator(Filament)
+
+    unscaled = Instrument.ChannelCreator(UnscaledData)
```
BenediktBurger: Please move these channels up to the init, such that these "structural information" is up on top.

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,622 @@
...
+
+    baudrate = Instrument.setting(
+        "07,%d",
+        """Set the baud rate.
+
+        :type: int, strictly in ``9600``, ``19200``, ``38400``, ``57600``, ``115200``
```
BenediktBurger: We use the format:

```suggestion
        """Set the baud rate (int, strictly in ``9600``, ``19200``, ``38400``, ``57600``, ``115200``).
```

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,622 @@
...
+        )
+
+    hv_on_timer = Instrument.measurement(
+        "21",
+        """Get the HV On time in hours (float).""",
+        )
+
+    status = Instrument.measurement(
+        "22",
+        """Get the power supply status (enum).""",
```
BenediktBurger: You could use the sphinx code for linking to this enum:
```suggestion
        """Get the power supply status (:code:`StatusCode` enum).""",
```

tests/instruments/spellmanhv/test_spellmanXRV.py
```diff
@@ -0,0 +1,666 @@
...
+from pymeasure.instruments.spellmanhv.spellmanXRV import SpellmanXRV, StatusCode, ErrorCode
+
+STX = chr(2)
+ETX = chr(3)
+
+# communication during class initialization
+INITIALIZATION = (f"{STX}28,j{ETX}", f"{STX}28,160,30,0,|")
+
+
+def checksum(string_to_check):
```
BenediktBurger: why don't you use it from SpellmanXRV?
E.g. `checksum = SpellmanXRV.checksum`? (works only for a staticmethod or a method outside the class)

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,622 @@
...
+        self.query_delay = query_delay
+
+        # disable checksum for LAN interface
+        interface_type = self.adapter.connection.interface_type
+        if interface_type is InterfaceType.tcpip:
+            self.checksum_enabled = False
+
+        self.set_scaling()
+
+    def checksum(self, string_to_check):
```
BenediktBurger: As this method does not need self at all, you can either extract it from the class or make it a staticmethod:

```suggestion
    @staticmethod
    def checksum(string_to_check):
```

tests/instruments/spellmanhv/test_spellmanXRV.py
```diff
@@ -0,0 +1,666 @@
...
+
+    @pytest.mark.parametrize("baudrate, mapping",
+                             [(9600, 1),
+                              (19200, 2),
+                              (38400, 3),
+                              (57600, 4),
+                              (115200, 5)])
+    def test_baudrate(self, baudrate, mapping):
+        set_cmd = f"07,{mapping},"
+        set_cmd_csum = checksum(set_cmd)
```
BenediktBurger: Warning, you always check, that the checksum is calculated twice in the same way.

It would be good to add a dedicated test for the checksum, or you could test it implicitly in a "normal" test, which has hardcoded responses (so it tests the checksum generation).

BenediktBurger (changes requested):
Thanks for your contribution.

Here a few comments.

pymeasure/instruments/spellmanhv/spellmanXRV.py
Falk3579: done

pymeasure/instruments/spellmanhv/spellmanXRV.py
Falk3579: done

tests/instruments/spellmanhv/test_spellmanXRV.py
Falk3579: Thanks for the hint. It's now implemented as you suggested.

tests/instruments/spellmanhv/test_spellmanXRV.py
Falk3579: I replaced the checksum function by hardcoded values.

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,620 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Instrument, Channel
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+from enum import IntFlag
+from pyvisa.constants import InterfaceType
```
BenediktBurger: typically, there are three groups of imports (each group separated by an empty line):
- standard library modules (e.g. enum)
- third party modules (the other imports)
- local imports (or from this package)

In each group, they are ordered alphabetically

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,620 @@
...
+    INVERTER_NOT_READY = 2**26
+
+
+class Filament(Channel):
+    """A class representing the functions for the filament of the x-ray tube."""
+
+    limit = Channel.control(
+        "16",
+        "12,%d",
+        """Control the filament limit setpoint (int, strictly from 0 to 4095).""",
```
BenediktBurger, line 95: Is there any translation of this integer to temperature etc.?

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,620 @@
...
+        "17",
+        "13,%d",
+        """Control the filament preheat setpoint (int, strictly from 0 to 4095).""",
+        check_set_errors=True,
+        validator=strict_range,
+        values=[0, 4095],
+        get_process_list=lambda v: int(v[0]),
+        )
+
+    size = Channel.setting(
```
BenediktBurger: As it is basically a boolean, you could rename it to `large_filament_size_enabled`, such that you can use True/False (via mapping) instead of strings.

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,620 @@
...
+             },
+        cast=int,
+        )
+
+
+class SpellmanXRV(Instrument):
+    """A class representing the Spellman XRV series high voltage power supplies."""
+
+    STX = chr(2)
+    ETX = chr(3)
```
BenediktBurger: You could (just an idea) move these outside the class, such that you can refer them just as `STX` instead of `self.STX`.

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,620 @@
...
+        Adds <STX> (0x02) in front and checksum + <ETX> (0x03) at end of every command before
+        sending it. The checksum is omitted for TCPIP connections.
+        """
+
+        command_with_comma = command + ","
+        if self.checksum_enabled:
+            checksum = self.checksum(command_with_comma)
+            super().write(f"{self.STX}{command_with_comma}{checksum}{self.ETX}")
+        else:
+            super().write(f"{self.STX}{command_with_comma}{self.ETX}")
```
BenediktBurger: You should not need to add ETX at the end, if you state, that ETX is the write_termination.

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,620 @@
...
+
+        :raise: ConnectionError if response doesn't start with <STX> or checksum is incorrect.
+                The checksum check is omitted for TCPIP connections.
+        """
+        got = super().read()
+
+        if not got.startswith(self.STX):
+            raise ConnectionError("Expected <STX> at begin of received message.")
+
+        response = got.strip(self.STX).rpartition(",")
```
BenediktBurger: Just as a suggestion: You check, whether it start with STX, so you can just return a view.
```suggestion
        response = got[1:].rpartition(",")
```
Your code, however, is more declarative as it explicitly states, what should be removed.

pymeasure/instruments/spellmanhv/spellmanXRV.py
```diff
@@ -0,0 +1,620 @@
...
+        "23",
+        """Get the DSP part number and version.
+
+        :return: dict
+
+        :dict keys: ``part_number``, ``version``
+        """,
+        get_process_list=lambda v: {"part_number": v[0],
+                                    "version": int(v[1])},
+        )
```
BenediktBurger: if you have two values, won't a tuple of part number and version be better suited than a dict?

BenediktBurger (commented):
A few more comments, but it is getting close.

pymeasure/instruments/spellmanhv/spellmanXRV.py
Falk3579: Unfortunately not.

pymeasure/instruments/spellmanhv/spellmanXRV.py
Falk3579: Imports are sorted now as suggested.

pymeasure/instruments/spellmanhv/spellmanXRV.py
Falk3579: Good idea. It's now declared outside the class.

pymeasure/instruments/spellmanhv/spellmanXRV.py
Falk3579: write_termination is now set to ETX

pymeasure/instruments/spellmanhv/spellmanXRV.py
Falk3579: Part number is something like 'SWM9999-999' and version '56234'. Do you recommend a tuple in such a case?

pymeasure/instruments/spellmanhv/spellmanXRV.py
Falk3579: Modified as suggested.

pymeasure/instruments/spellmanhv/spellmanXRV.py
Falk3579: renamed to 'large_size_enabled' and mapped as bool.

pymeasure/instruments/spellmanhv/spellmanXRV.py
BenediktBurger: Here, it is a matter of taste.

Dictionaries with strings are always error prone. On the other hand, they contain meta data about the value.

In this case I personally would prefer a tuple, but am not strongly opposed to a tuple.

Falk3579: `fpga` and `dsp` are returning lists now

PR #1332, opened 2025-07-14
Add support for Keithley 4200A-SCS

Falk3579:
Add initial support for the Keithley 4200A-SCS Parameter Analyzer

### DIFF
--- a/pymeasure/instruments/keithley/keithley4200.py
+++ b/pymeasure/instruments/keithley/keithley4200.py
@@ -56,7 +56,7 @@ class SMU(Channel):
         self.write("US;DV{ch}")
         self.check_set_errors()
 
-    source_voltage = Channel.setting(
+    voltage_setpoint = Channel.setting(
         "US;DV{ch},%d,%g,%g",  # range, value, compliance
         """Set range, output voltage and current compliance (int, float, float).
 
@@ -86,7 +86,7 @@ class SMU(Channel):
         get_process=lambda v: float(v[3:]),
         )
 
-    source_current = Channel.setting(
+    current_setpoint = Channel.setting(
         "US;DI{ch},%d,%g,%g",  # range, value, compliance
         """Set range, output current and voltage compliance (int, float, float).
 
--- a/tests/instruments/keithley/test_keithley4200.py
+++ b/tests/instruments/keithley/test_keithley4200.py
@@ -46,14 +46,14 @@ class TestKeithley4200SMU:
          (1, 1.345E2, 1e-6),
          (5, -25.45E1, 3e-3),
          ])
-    def test_source_voltage(self, ch, range, voltage, compliance):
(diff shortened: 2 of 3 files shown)


pymeasure/instruments/keithley/keithley4200.py
```diff
@@ -0,0 +1,206 @@
...
+    B7 = 2**7
+
+
+class SMU(Channel):
+    """A class representing the SMU (source/measure unit) channel."""
+
+    def disable(self):
+        """Disable the SMU."""
+        ch = self.id
+        self.write(f"US;DV{ch}")
```
BenediktBurger: If I remember correctly, the channel name is already present as self.id and will replace {ch}, so this will work as well (note that it is not an f-string anymore):

```suggestion
        self.write("US;DV{ch}")
```

pymeasure/instruments/keithley/keithley4200.py
```diff
@@ -0,0 +1,206 @@
...
+            - 5: 2 V range, only with a preamplifier
+
+        :return: dict
+        :dict keys: ``value``, ``status``
+
+        """,
+        check_set_errors=True,
+        get_process=lambda v: dict(value=float(v[3:]),
+                                   status=str(v[:3]),
+                                   ),
```
BenediktBurger: What do you get back? The current output setpoint, or actual output value (measured)?

Falk3579: It returns the measured value.

BenediktBurger: In this case I suggest to use one property for the setpoint (suffixed with `_setpoint`) and one for the actually measured value.
The latter one should just return the value and not the status or what does the status say?

Falk3579: I added 'source_current' and 'source_voltage' as new properties. These are the same names as in the other Keithley SMU drivers.
The status contains information if a SMU is oscillating or if it is in compliance. I removed it from the return value.

pymeasure/instruments/keithley/keithley4200.py
```diff
@@ -53,14 +53,12 @@ class SMU(Channel):
...
     def disable(self):
         """Disable the SMU."""
-        ch = self.id
-        self.write(f"US;DV{ch}")
+        self.write("US;DV{ch}")
         self.check_set_errors()
 
-    voltage = Channel.control(
-        "US;TV{ch}",
+    source_voltage = Channel.setting(
```
BenediktBurger: For consistency, we typically use `voltage_setpoint` to indicate what the setpoint (i.e. desired value) is.

Falk3579: voltage_setpoint and current_setpoint are used now

PR #1334, opened 2025-07-21
Dev/lake shore340

DF9AS:
Dear maintainers of pymeasure,

LakeShore Temperature Controllers can have more options (inputs and outputs) than implemented in the LakeShore331 device class.

As the naming for the implemented channels depends on the installed options (Input Option Cards) and the number of output channels varies between different series, one could implement multiple quite similar device classes or one to "rule them all".

I'm not sure about your conventions so I've added the LakeShore340 with respect to the up to four possible outputs in the 350 series. A better solution might be to replace the device class LakeShore331 with a new one (like this LakeShore340) and name it LakeShore3xx but I didn't want to break the already implemented measurement routines based on LakeShore331 which might be obsolete having the 340.

Thank you and kind regards.

### DIFF
--- a/pymeasure/instruments/lakeshore/lakeshore340.py
+++ b/pymeasure/instruments/lakeshore/lakeshore340.py
@@ -1,0 +1,95 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2025 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import logging
+
+from pymeasure.instruments import Instrument, SCPIUnknownMixin
+from pymeasure.instruments.lakeshore.lakeshore_base import LakeShoreTemperatureChannel, \
+    LakeShoreHeaterChannel
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class LakeShore340(SCPIUnknownMixin, Instrument):
+    """ Represents the Lake Shore 340/350 Temperature Controller and provides
+    a high-level interface for interacting with the instrument. 
+    In addition to the :ref:`lakeshore331`, multiple input-and output options are implemented.
+    Not all channels exist in all device versions!
+
+    Note that the 340 provides two input channels (A and B) and one output channel.
+    Additional input option cards can be used to add input channels.
+    - 3465 Single Capacitance Input Option Card adds one channel (C)
+    - 3462 Dual Standard Input Option Card or 3464 Dual Thermocouple Input Option Card 
+        adds two channels (C and D).
+    - 3468 Eight Channel Input Option Card adds eight channels (C1-C4, D1-D4)
+
+    The LS 350 can power four output channels (1-4).
+    - 3062 4-Channel Scanner option is installed in the Model 350, 
+        4 additional channels, D2, D3, D4, and D5, become available for use.
+
+    This driver makes use of the :ref:`LakeShoreChannels`.
+
+    .. code-block:: python
+
+        controller = LakeShore340("GPIB::1")
+
+        print(controller.output_1.setpoint)         # Print the current setpoint for loop 1
+        controller.output_1.setpoint = 50           # Change the loop 1 setpoint to 50 K
+        controller.output_1.heater_range = 'low'    # Change the heater range to low.
+        controller.input_A.wait_for_temperature()   # Wait for the temperature to stabilize.
+        print(controller.input_A.temperature)       # Print the temperature at sensor A.
+    """
+    input_A = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'A')
+
+    input_B = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'B')
+
+    input_C = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'C')
+
+    input_D = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D')
+
+    input_C1 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'C1')
+    input_C2 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'C2')
+    input_C3 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'C3')
+    input_C4 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'C4')
+
+    input_D1 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D1')
+    input_D2 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D2')
+    input_D3 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D3')
+    input_D4 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D4')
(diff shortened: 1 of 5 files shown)


pymeasure/instruments/lakeshore/lakeshore340.py
```diff
@@ -0,0 +1,95 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import logging
+
+from pymeasure.instruments import Instrument, SCPIUnknownMixin
```
BenediktBurger: If the device talks default SCPI commands (e.g. *IDN?, *RST, *STB?), use `SCPIMixin`. If not, do not use any mixin, but set `include_SCPI=False` in `__init__`.

The UnknownMixin indicates, that we do not know, whether it supports SCPI or not (initially SCPI support was default in pymeasure).

As you happen to know these devices, would you mind to adjust the other instruments (at least 331) as well? Thanks!

pymeasure/instruments/lakeshore/lakeshore340.py
```diff
@@ -0,0 +1,95 @@
...
+    input_D5 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D5')
+
+    output_1 = Instrument.ChannelCreator(LakeShoreHeaterChannel, 1)
+    output_2 = Instrument.ChannelCreator(LakeShoreHeaterChannel, 2)
+    output_3 = Instrument.ChannelCreator(LakeShoreHeaterChannel, 3)
+    output_4 = Instrument.ChannelCreator(LakeShoreHeaterChannel, 4)
+
+
+    def __init__(self, adapter, name="Lakeshore Model 336 Temperature Controller", **kwargs):
+        kwargs.setdefault('read_termination', "\r\n")
```
BenediktBurger: can you change the read_termination on the device?
If yes, consider adding it to the signature of `__init__`.
If not, write it directly into `super().__init__` call.

pymeasure/instruments/lakeshore/lakeshore340.py
```diff
@@ -0,0 +1,95 @@
...
+    input_D1 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D1')
+    input_D2 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D2')
+    input_D3 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D3')
+    input_D4 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D4')
+    input_D5 = Instrument.ChannelCreator(LakeShoreTemperatureChannel, 'D5')
+
+    output_1 = Instrument.ChannelCreator(LakeShoreHeaterChannel, 1)
+    output_2 = Instrument.ChannelCreator(LakeShoreHeaterChannel, 2)
+    output_3 = Instrument.ChannelCreator(LakeShoreHeaterChannel, 3)
+    output_4 = Instrument.ChannelCreator(LakeShoreHeaterChannel, 4)
```
BenediktBurger: Please make the spacings consistent (e.g. no empty lines between A, B, C, D).

BenediktBurger: Do I understand it correctly, that not all instruments have all these channels, but it depends on the actual device and expansion card?

Regarding pymeasure policy:
- We want to keep backward compatibility, so, if you wanted to replace the Lakeshore331, you would have to keep a Lakeshore331 class at the accustomed place, but it could inherit from this generic class.
- We try to keep it simple and to not duplicate too much code
  - As these instruments consist mainly of channel imports, I'm fine with one class per instrument.
  - It's also fine to have many channels and state, that not all channels are actually available, depending on the device.

pymeasure/instruments/lakeshore/lakeshore340.py
```diff
@@ -0,0 +1,95 @@
...
+#
+
+import logging
+
+from pymeasure.instruments import Instrument, SCPIUnknownMixin
+from pymeasure.instruments.lakeshore.lakeshore_base import LakeShoreTemperatureChannel, \
+    LakeShoreHeaterChannel
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
```
BenediktBurger: If you do not use the `log` to log anything, you can remove these lines and the logging import.

BenediktBurger (changes requested):
Thanks for your contribution, it is very appreciated.

Your suggestion looks good.

Here are a few comments, also addressing your questions.
See also the comments in github regarding style (only a single line between class methods, no trailing whitespaces...).

Typically we recommend to add tests, but in this case, there is no logic, as the logic is in the channels. Therefore, no tests are needed.

pymeasure/instruments/lakeshore/lakeshore340.py
DF9AS: Yes, that's right. The Lakeshore temperature controllers all have a certain number of sensor inputs and heater outputs. These numbers vary between the model series and expansion cards, but essentially they are identical in terms of the command structure.

From my point of view, the following would make sense:
- My pull request for LakeShore340 is rejected
- I extend the LakeShore331 class by the number of all possible input / output combinations according to the previous scheme, taking into account the different device operating manuals for Series 331, 340 and newer mentioned at LakeShore (https://www.lakeshore.com/products/categories/temperature-products/cryogenic-temperature-controllers ).
- I add the requirements already mentioned here (e.g. read_termination, SCPIMixin) to the LakeShore331 class.
- I supplement the documentation of the LakeShore331 class in such a way that it becomes clear that the class is universal for other LakeShore controllers of the 3xx series.

This gives us backward compatibility and at the same time a functional extension for the entire model series without further expanding the code base.

BenediktBurger: If it is useful for the 3xx series, it is better to make such a class in an appropriate file (you can just do it in this PR) and then make the 311 inherit from that other class.

The 311 can then (in its init) raise a deprecation warning, such that users can switch to the main class until the main class will be removed.

pymeasure/instruments/lakeshore/lakeshore340.py
DF9AS: Thanks for the explanation. I checked the manuals and decided to use SCPIMixin on both 331 and 340 / 3xx.

pymeasure/instruments/lakeshore/lakeshore340.py
DF9AS: Also thank you for this recommendation. The read_termination can bechanged, so I added it to the signature.

pymeasure/instruments/lakeshore/lakeshore331.py
```diff
@@ -56,8 +53,13 @@ class LakeShore331(SCPIUnknownMixin, Instrument):
...
     output_2 = Instrument.ChannelCreator(LakeShoreHeaterChannel, 2)
 
-    def __init__(self, adapter, name="Lakeshore Model 336 Temperature Controller", **kwargs):
-        kwargs.setdefault('read_termination', "\r\n")
+    def __init__(self,
+                 adapter,
+                 name="Lakeshore Model 336 Temperature Controller",
+                 read_termination="\r\n",
+                 **kwargs):
+        kwargs.setdefault('read_termination', read_termination)
```
BenediktBurger: In this case, you can just inject the read_termination to the super call.

```python
        warnings.warn("LakeShore331 is deprecated. Use LakeShore3xx instead.", DeprecationWarning)
super().__init__(
adapter,
name,
read_termination=read_termination,
**kwargs
```

pymeasure/instruments/lakeshore/lakeshore331.py
```diff
@@ -56,8 +53,13 @@ class LakeShore331(SCPIUnknownMixin, Instrument):
...
+        warnings.warn("LakeShore331 is deprecated. Use LakeShore3xx instead.", DeprecationWarning)
```
BenediktBurger, line 64: Please put also a note to the docstrings of the class. See https://github.com/pymeasure/pymeasure/issues/672#issuecomment-1371900222

pymeasure/instruments/lakeshore/lakeshore3xx.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self,
+                 adapter,
+                 name="Lakeshore Model 3xx Temperature Controller",
+                 read_termination="\r\n",
+                 **kwargs):
+        kwargs.setdefault('read_termination', read_termination)
+        super().__init__(
+            adapter,
+            name,
+            **kwargs
```
BenediktBurger: As the kwargs are set to the super call, the following is better:


```suggestion
        super().__init__(
            adapter,
            name,
            read_termination=read_termination,
            **kwargs
```

BenediktBurger (approved):
A few last comments.

Thanks for your work.

pymeasure/instruments/lakeshore/lakeshore331.py
```diff
@@ -46,7 +43,10 @@ class LakeShore331(SCPIUnknownMixin, Instrument):
         controller.output_1.setpoint = 50           # Change the loop 1 setpoint to 50 K
         controller.output_1.heater_range = 'low'    # Change the heater range to low.
         controller.input_A.wait_for_temperature()   # Wait for the temperature to stabilize.
-        print(controller.input_A.temperature)       # Print the temperature at sensor A.
+        print(controller.input_A.kelvin             # Print the temperature at sensor A.
+
+    .. deprecated:: v0.16.0
```
BenediktBurger: A last comment:

```suggestion
    .. deprecated:: 0.16.0
```

BenediktBurger:

PR #1337, opened 2025-07-28
Fix image_gui.py example

Falk3579:
Replaced `filename_input` and `directory_input` by `enable_file_input`. Now the example is working again.


CasperSchippers (approved):
Good catch! Looks good to me.

Thanks for the fix!!

PR #1338, opened 2025-07-28
Define 'Ohm' in units.py

Falk3579:
Define 'Ohm' as an alias for 'ohm'.
This makes the both examples in 'examples/Current-Voltage Measurements' working. Otherwise 'Ohm' has to be replaced by 'Ω' in the examples.


Falk3579:
The examples use `ohm` now.

Falk3579:
Yes, you can merge.

BenediktBurger:
Thanks for spotting the bugs in the examples.

Issue #1298, opened 2025-05-09
Missing color bar scale  in the image plotting functionality


PkRPintu:
Hello,

I am new to Pymeasure and would like to use it for live image scanning while grabbing data from various lock-ins. I have three questions if anyone can kindly help me. I really appreciate any help you can offer.

1. How can I add a color bar to the image scan display that changes dynamically depending on the Z axis values? (If the colour bar can be interactive, that would also be a great feature)
2. How to change the colormap from viridis to something else, probably a diverging colour map like ‘bwr’?
3. Also, there is a managed docked window for displaying multiple plots. Can we have managed docked image windows so that we can display multiple image plots at once?

I think these are very handy features for the image scan experiments and will be helpful for everyone.

Cheers

CasperSchippers:
Hey,

These are good, but difficult questions, to which I don't have a good awnser to all of them.

> 1. How can I add a color bar to the image scan display that changes dynamically depending on the Z axis values? (If the colour bar can be interactive, that would also be a great feature)

While I think it should be possible, I didn't manage to add one just yet... I do agree that this would be very, very useful!
By the way, what do you mean with interactive?

> 2. How to change the colormap from viridis to something else, probably a diverging colour map like ‘bwr’?

To change the colormap, you currently need to alter a single line in `curves.py`, where it specifies the colormap: https://github.com/pymeasure/pymeasure/blob/13a173fc465b0b0b5daabff2e4fe55ae4d2ff94a/pymeasure/display/curves.py#L87
Pyqtgraph does include colormaps of colorcet (https://colorcet.holoviz.org/, which I think are generally quite good, so for most purposes I expect this would suffice), so it includes a set of diverging colormaps, such as `CET-D1` for bwr, or `CET-D4` for bkr, or `CET-D6` for bky (these ones have the advantage that zero also stands out from the background of the plotting window), or (one which I particullarly like as it is isoluminant to prevent misguided interpretations) `CET-D12`/`CET-D11`.

For a diverging colormap, you may also want to adjust the edges of the colormap, such that zero is the center of the map (i.e. white, or gray, or black, depending on the exact map that is used). For this you can e.g. update the zmin and zmax values after these lines: https://github.com/pymeasure/pymeasure/blob/13a173fc465b0b0b5daabff2e4fe55ae4d2ff94a/pymeasure/display/curves.py#L104-L105
For example by adding some lines such as:
``` python
        zmax = max(abs(zmin), abs(zmax))
[...]

PR #1341, opened 2025-08-02
Rewrote ImageWidget algorithm and added colorbar and colormap selection

dvalde0:
Added colormaps to the image widget and rewrote the algorithm for image data to use numpy array operations. This way it should be faster. Also modified the example image_gui to actually run and be useful for testing. Please ignore the linting errors for the moment, I just want to see if it is actually working as before.

The colorbar is added to the plot of the image frame and it is updated on PlotFrame update. The colormap is chisen via a dropdown menu which shows all the colormaps available in pyqtgraph. Upon selection the colormap propagates from the ImageWidget to the ImageFrame to the ResultsImage items, updating all items.

Some sidenotes while I was developing this:
1. Why is the data polled on a timer in PlotFrame? Wouldn't it make more sense for the results class to emit a signal on file change, maybe via watchdog? This way it wouldn't update the graphics even when the data is not updating.
2. Should I impement a way to view the z-value of the image at the cursor position?

Edit: Ok, ignore 1. I see that update_curves checks if the Procedure is running. :)
Edit 2: Solves #1298


CasperSchippers:
Sure will do, I'll try to do it this weekend (had very little time this week, sorry)

examples/Basic/image_gui.py
```diff
@@ -77,43 +81,56 @@ def execute(self):
...
-        self.setWindowTitle('PyMeasure Image Test')
+        self.setWindowTitle("PyMeasure Image Test")
 
     def queue(self):
-        direc = '.'
-        filename = unique_filename(direc, 'test')
+        filename = unique_filename(tempfile.gettempdir(), "test")
         procedure = self.make_procedure()
         results = Results(procedure, filename)
         experiment = self.new_experiment(results)
```
CasperSchippers: Seeing that the example is revamped anyways, maybe it'd be god to remove this queue method, and the `enable_file_input=false,` option (line128) to use the default queueing method, and better focus this example on using the image gui; also removes the need for a temp file here.

pymeasure/display/widgets/image_frame.py
```diff
@@ -27,20 +27,25 @@
 from ..curves import ResultsImage
 from ..Qt import QtCore
 from .plot_frame import PlotFrame
+import pyqtgraph as pg
+from ...experiment import Procedure
```
CasperSchippers: You seem to have added two imports that you don't use, or am I missing something?

pymeasure/display/curves.py
```diff
@@ -58,92 +58,71 @@ def update_data(self):
...
-
-    def colormap(self, x):
-        """ Return mapped color as 0.0-1.0 floats RGBA """
-        return self.cm.map(x, mode='float')
-
-    # TODO: colormap selection
+        z_data = data[self.z].to_numpy(dtype=np.float64)
+
+        self.img_data.ravel()[: z_data.size] = z_data
+        print(self.img_data.shape, z_data.shape, self.xsize * self.ysize)
```
CasperSchippers: There is a print statement left of (I guess) debugging. Could you remove this please.

pymeasure/display/curves.py
```diff
@@ -58,92 +58,71 @@ def update_data(self):
...
-        self.img_data = np.zeros((self.ysize, self.xsize, 4))
+        self.xstart = getattr(self.results.procedure, self.x + "_start")
+        self.xend = getattr(self.results.procedure, self.x + "_end")
+        self.xstep = getattr(self.results.procedure, self.x + "_step")
+        self.xsize = int(np.ceil((self.xend - self.xstart) / self.xstep))
+        self.ystart = getattr(self.results.procedure, self.y + "_start")
+        self.yend = getattr(self.results.procedure, self.y + "_end")
+        self.ystep = getattr(self.results.procedure, self.y + "_step")
+        self.ysize = int(np.ceil((self.yend - self.ystart) / self.ystep))
+        self.img_data = np.zeros((self.ysize, self.xsize))
```
CasperSchippers: Is it an Idea to fill the initial matrix with nan's? These are treated as transparent pixels.

pymeasure/display/curves.py
```diff
@@ -58,92 +58,71 @@ def update_data(self):
...
 
-        # Scale and translate image so that the pixels are in the correct
-        # position in "data coordinates"
-        tr = QtGui.QTransform()
-        tr.scale(self.xstep, self.ystep)
-        tr.translate(int(self.xstart / self.xstep) - 0.5,
-                     int(self.ystart / self.ystep) - 0.5)  # 0.5 so pixels centered
-        self.setTransform(tr)
-
     def update_data(self):
```
CasperSchippers, line 99: Your version is a lot more compact, which is nice, but (unless I miss something) it misses the option to add data in an arbitrary order. Your version assumes that data is added in a column-major order (i.e., it assumes that first all data for the first x-value is given, then for the next x-value, and so on, try swapping the for-loops on line 82 and 83 of the image_gui example), but this might not always be that case that you receive data in that order (data can be recorded row-major, some different scanning approach, or maybe even randomly added, all of which are highly likely). That is why the old version colors its pixels based on the x,y locations.

I think it is important that this (i.e. basing which pixels to color on the (x, y)-location of the data) is a quite important feature to support and should be added/returned.

pymeasure/display/curves.py
```diff
@@ -58,92 +58,71 @@ def update_data(self):
...
+        if len(z_data) != 0:
+            self.setImage(
+                image=self.img_data,
+                rect=[
+                    self.xstart - self.xstep / 2,
+                    self.ystart - self.ystep / 2,
+                    self.xend - self.xstart,
+                    self.yend - self.ystart,
+                ],
+                autoLevels=True,
```
CasperSchippers: I'm not always a fan of blind autoleveling (you should for the fun of it try adding `(-2)**progit` to the `pixel_data` in the image_gui example, that is maybe a bit extreme, but shows the issue a bit).

One option that would help incredibly is to have the option to force symmetry in the colorbar (i.e. that the lower bound = - the upper bound). Would it be an options (I'm not sure at this moment how difficult it is) to add/expose this option to the user (as a toggle-button next to the colormap-dropdown for example and/or an keyword argument just like colormap for the ManagedImageWindow)? A more extensive option would be to do the same thing for the lower and upper limits of the colorbar, but maybe symmetric would be a big help already?

pymeasure/display/widgets/image_frame.py
```diff
@@ -27,20 +27,25 @@
...
 
 class ImageFrame(PlotFrame):
-    """ Extends :class:`PlotFrame<pymeasure.display.widgets.plot_frame.PlotFrame>`
+    """Extends :class:`PlotFrame<pymeasure.display.widgets.plot_frame.PlotFrame>`
     to plot also axis Z using colors
     """
+
     ResultsClass = ResultsImage
     z_axis_changed = QtCore.Signal(str)
+    item_updated = QtCore.Signal()
```
CasperSchippers, line 42: Is this one used somewhere?

pymeasure/display/widgets/image_widget.py
```diff
@@ -35,44 +35,67 @@
...
         )
-        self.updated = self.image_frame.updated
         self.plot = self.image_frame.plot
-        self.columns_z.setCurrentIndex(2)
+        self.colorbar = pg.ColorBarItem(
+            colorMap=self.colormap,
+            interactive=False,
+            colorMapMenu=False,
+        )
+        self.colorbar.setLabel("left", text=self.columns_z.currentText())
```
CasperSchippers: Feels a bit weird to have the label on the left, not; is the common placement not on the right? (I.e., on the side with the numbers, similar to the other axes of the figure)

Also applies to the updating of the color-bar on line 139.

ps. I have the impression that this line is not even needed, only the one on line 139 seems to be doing anything.

pymeasure/display/widgets/image_widget.py
```diff
@@ -35,44 +35,67 @@
...
-            self.check_status
+            self.x_axis, self.y_axis, self.columns[0], self.refresh_time, self.check_status
         )
-        self.updated = self.image_frame.updated
         self.plot = self.image_frame.plot
-        self.columns_z.setCurrentIndex(2)
+        self.colorbar = pg.ColorBarItem(
+            colorMap=self.colormap,
+            interactive=False,
+            colorMapMenu=False,
```
CasperSchippers, line 103: Why dissable this? it is actually a nice menu to change the colormaps, right?

PS. oh wait, hmmm, I think I understand, it doesn't work nicely with the dropdown menu.
How about letting the colorbar handle the colormap? especially when multiple imageItems are added to the colorbar (see my remark later in this file), it seems to work flawelessly. It would basically mean removing the `colormap` attribute from the `ResultImage` class. 

You can still keep the dropdown, which I think is a nice and convenient (because visible) way of setting the colormap, but letting the colorbar handle the coloring seems quite good. You can even update it with what you set in the colorMapMenu is you change this line to: 
``` suggestion
            colorMapMenu=ColorMapMenu(showColorMapSubMenus=True),
```
and add this line:
```python
       self.colorbar.colorMapMenu.sigColorMapTriggered.connect(lambda cmap: self.colormaps.setCurrentText(cmap.name))
```

You can (and probably should) also remove line 149 (`self.image_frame.set_colormap(self.colormap)` from `set_colormap`) then, since that is handled by the colorbar.

pymeasure/display/widgets/image_widget.py
```diff
@@ -35,44 +35,67 @@
...
-            self.refresh_time,
-            self.check_status
+            self.x_axis, self.y_axis, self.columns[0], self.refresh_time, self.check_status
         )
-        self.updated = self.image_frame.updated
         self.plot = self.image_frame.plot
-        self.columns_z.setCurrentIndex(2)
+        self.colorbar = pg.ColorBarItem(
+            colorMap=self.colormap,
+            interactive=False,
```
CasperSchippers, line 102: I wanted to say "why would you disable this, seems quite a nice feature", until I tried it, but it is extremely finicky..... not really usefull it seems... 😞

pymeasure/display/widgets/image_widget.py
```diff
@@ -113,3 +150,9 @@ def load(self, curve):
 
     def remove(self, curve):
         self.plot.removeItem(curve)
+
+    def _update_colorbar(self):
+        """Updates the colorbar to match the current image"""
+        for item in self.plot.items:
+            if isinstance(item, ResultsImage):
+                self.colorbar.setImageItem(item, insert_in=self.plot)
```
CasperSchippers, line 232: I think you can set multiple images at once to have a better link:
```suggestion
        items = [item for item in self.plot.items if isinstance(item, ResultsImage)]
        self.colorbar.setImageItem(items, insert_in=self.plot)
```

This way, the colorbar always matches with all of the figures. However the scale is auto-updated with the latest one (so the one that is currently running for example... I guess that is a limitation of leaving the autoleveling to the setImage function?

CasperSchippers (changes requested):
Hey; thanks for the PR. There is quite some added value in this! I added a couple of suggestions / remarks.

My main point regards setting the z-data, which in this version relies on the data being passed column-major, while that is not something we can always be sure of. I think that allowing for a random distribution of data-points is crucial.

I really like the addition of the colorbar and the ad-hoc selection of the colormap!

Regarding your question about the z-values; that sounds like a useful feature, but feel free to decide whether you want to add it or not!

As an asside; you seem to have used a forematter or something on the files. I would like to urge you to be very careful with that. If multiple people do that and use formatters with a different opionion, you get a lot of rubish commits. Additionally, it really really complicates a good review (as you have to filter what is an actual difference and just a formatting change).

pymeasure/display/curves.py
dvalde0: @CasperSchippers Could you check the latest commit [7451583](https://github.com/pymeasure/pymeasure/pull/1341/commits/74515831c758e0c13e0433dff5be18850dd7112a), I updated the algorithm and the example to test it and it seems to work.

Edit: The algorithm update was in [bf61e87](https://github.com/pymeasure/pymeasure/pull/1341/commits/bf61e87ea2052b0c7191c830d2c5c7d205d17c03).

CasperSchippers: I'm not sure if it works completely as expected; you get a bit of weird result during the measurement: it seems to sort the (unique) values rather than placing them in the correct place of the image. As an example: you can see that clearly if you use binary pixel data (e.g., `(x > 0.9) and (x < 1.1)`).

It is also a slight risk to use unique here, as float-point arithmatics can cause slight deviations in the returned values, and unique will not account for that.


I think I did find a reasonable solution: calculate bins upfront and use np.digitize to get the correct bins for the data; this also allows for uncertainties in the x- and y-values: I've added suggestions to this review to show what I meant.

pymeasure/display/curves.py
```diff
@@ -98,14 +98,19 @@ def update_data(self):
...
-        print(self.img_data.shape, z_data.shape, self.xsize * self.ysize)
+        x_data = data[self.x].to_numpy(dtype=np.float64)
+        y_data = data[self.y].to_numpy(dtype=np.float64)
 
         # set image data
         if len(z_data) != 0:
+            x_unique = np.unique(x_data)
+            y_unique = np.unique(y_data)
+            x_ids = np.searchsorted(x_unique, x_data)
+            y_ids = np.searchsorted(y_unique, y_data)
```
CasperSchippers: To prevent an awkward sorting of uncomplete data, you can use digitize rather than unique+searchsorted: (the bins are calculated in a separate suggestion).

```suggestion
            x_ids = np.digitize(x_data, self.xbins, right=True)
            y_ids = np.digitize(y_data, self.ybins, right=True)
```

pymeasure/display/curves.py
```diff
@@ -85,7 +85,7 @@ def __init__(
         self.yend = getattr(self.results.procedure, self.y + "_end")
         self.ystep = getattr(self.results.procedure, self.y + "_step")
         self.ysize = int(np.ceil((self.yend - self.ystart) / self.ystep))
-        self.img_data = np.zeros((self.ysize, self.xsize))
+        self.img_data = np.full((self.ysize, self.xsize), np.nan, dtype=float)
```
CasperSchippers, line 91: Pre-calculate bins (related to the other suggestion about the digitizing)

```suggestion

        self.xbins = np.linspace(self.xstart, self.xend, self.xsize + 1)
        self.ybins = np.linspace(self.ystart, self.yend, self.ysize + 1)
        self.img_data = np.full((self.ysize, self.xsize), np.nan, dtype=float)

```

CasperSchippers (commented):
Thanks for the update, and the patience on my slow replies; I've had a look, but it seems to not work completely as (I at least) expected, with some weird behaviour during the measurement. I have a suggestion for how to find the right location in the img_data matrix. Have a look at what you think.

pymeasure/display/curves.py
dvalde0: @CasperSchippers pushed the suggestion in [#6f39762](https://github.com/pymeasure/pymeasure/pull/1341/commits/6f39762e17804b953a3869aeabccd4e4ce115a2f)

uv.lock
CasperSchippers, line 1: Please remove this file

pymeasure/display/curves.py
CasperSchippers: I'm wondering whether here we should set the image data already? I'm getting a warning "all-NaN slice encountered", removing the `image=self.img_data` argument prevents this warning and seems to work fine.

CasperSchippers (changes requested):
Nice; that seems to work as intended. Have added two additional small remarks (also please do have look at the items that are yet open)

pymeasure/display/curves.py
dvalde0: Yeah, I also think that it shouldn't affect anything, I pushed it.

pymeasure/display/widgets/image_widget.py
dvalde0: So for me the ColorMapMenu was not working, but I implemented the colormap icons in the dropdown.

pymeasure/display/curves.py
dvalde0: I added an option to disable the auto-levels and change the lower and upper bounds to the colormap, see the example, awaiting feedback on this.

pymeasure/display/widgets/image_widget.py
dvalde0: If the scale is auto-updated with the latest one, how is this different from the current implementation?

dvalde0:
@CasperSchippers Reviewed and implemented some of the open items, please have a look. Awaiting feedback whether the user functionality is ok.

pymeasure/display/widgets/image_widget.py
CasperSchippers: Ahh fair enough, this is very nice now!!!

pymeasure/display/curves.py
```diff
@@ -117,14 +120,13 @@ def update_data(self):
...
-
-    def set_colormap(self, colormap):
-        """Sets the colormap for the image."""
-        self.colormap = colormap
-
+            print(self.auto_level)
+            if not self.auto_level:
+                print(self.lower_cmap_bound)
+                self.setLevels([self.lower_cmap_bound, self.upper_cmap_bound], update=True)
+                print(self.levels)
```
CasperSchippers, line 129: For the final version, do not forget to remove these print statements

pymeasure/display/widgets/image_widget.py
CasperSchippers: Maybe it's clearest below; it is two times the same data, but once with auto level on and once with auto level off. This is the default image_gui example but split into two experiments, on from x=0 to 1 and one form x=1 to 2. The pixel data is x*y so should be continuous, which is clearly what is happening if you turn of auto level, but with auto-level on I would've expected roughly the same thing to happen (at least I would've expected that the data is continous).

The suggestion is one way to resolve this behaviour (ensure that updates to the color scaling are transmitted to all experiments)

Image

Image

CasperSchippers (changes requested):
It's a while since I've had time to look at this, sorry about that.

Functionallity wise I think this is great! Only minor details to comment. Also have a look at the linting errors/warnings.

Issue #1258, opened 2025-03-26
Agilent33220A Offset Range Wrong


Patschke:
I'm using pymeasure to talk to an Agilent33220A AWG. I've noticed that the parameter range for the offset is too strict (and perhaps also for some other parameters, I did not test that in detail). It's currently hard coded to -4.995 to +4.995:

https://github.com/pymeasure/pymeasure/blob/c9e806c7c30d4c9ea48424c3bc8d9827570ff4e1/pymeasure/instruments/agilent/agilent33220A.py#L124

However, the actual  range depends on the output termination. You can change that (default is 50 Ohm) in the Utility->Output Setup menu. 

![Image](https://github.com/user-attachments/assets/cd0872bc-6b89-49e0-8869-52e24c01d784)

The voltage range can go up to -9.95...+9.95 V for larger load resistances: 

![Image](https://github.com/user-attachments/assets/eeb7906c-ca8c-4304-a9d3-db47e3d88d82)

Via pymeasure it's still only possible to set values in the 50 Ohm range of 5 V.

BenediktBurger:
Thanks for submitting this bug report.
Would you mind adjusting the limits?

Could you please check after the change, what the device (in 50 Ohm resistance) does, if you exceed +- 5V? Does pymeasure raise an exception?

Thanks!

BenediktBurger:
> Given that the current policies for the frequency already is to fail silently (setting a frequency above 200 kHz while in Ramp mode shows the same error on the device as setting an offset Voltage above 5 V in 50 Ohm termination) I could see simply raising all the limits?

Instead of raising the limits, you can remove the limits altogether, as they depend on the configuration which we do not know at all. Also the docstring should state, that the limits depend on the current configuration.

Thanks for showing the technical details.

PR #1342, opened 2025-08-04
Fix parameter ranges for Agilent33220A

Patschke:
I noticed that I never actually put a fix up for this. 

closes https://github.com/pymeasure/pymeasure/issues/1258

### DIFF
--- a/pymeasure/instruments/agilent/agilent33220A.py
+++ b/pymeasure/instruments/agilent/agilent33220A.py
@@ -88,23 +88,18 @@ class Agilent33220A(SCPIUnknownMixin, Instrument):
     )
 
     frequency = Instrument.control(
-        "FREQ?", "FREQ %s",
-        """Control the frequency of the output waveform, depending on the method
-        valid range depends on the current shape and supplying out-of-bound values will silently
-        be clipped by the device (float, strict from 1e-6 to 20e+6).""",
-        validator=strict_range,
-        values=[1e-6, 20e+6],
+        "FREQ?", "FREQ %f",
+        """Control the frequency of the output waveform. Depending on the output shape, the
+        supported frequency range changes. Supplying out-of-bound values may silently be clipped
+        by the device (float, in Hz)"""
     )
 
     amplitude = Instrument.control(
         "VOLT?", "VOLT %f",
-        """Control the amplitude of the output waveform. Depending on amplitude unit this
-        is Volt (peak-to-peak), Volt (RMS) or dBm and the range changes accordingly. The limits
-        also depend on the configured output termination (50 Ohm to High Impedance changes
-        by a factor of 2). Supplying out-of-bound values will silently be clipped by the device
-        (float strict from -53.05 (minimal dbm) to 19.9 (maximal VPP)).""",
-        validator=strict_range,
-        values=[-53.05, 19.9],
+        """Control the amplitude of the output waveform. Depending on amplitude unit, the unit
+        is Volt (peak-to-peak), Volt (RMS) or dBm. The limits depend on the configured output
+        termination (50 Ohm to High Impedance changes by a factor of 2) and the offset.
+        Supplying out-of-bound values may silently be clipped by the device (float, in V or dBm)"""
     )
 
     amplitude_unit = Instrument.control(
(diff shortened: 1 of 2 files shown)


pymeasure/instruments/agilent/agilent33220A.py
```diff
@@ -90,16 +90,21 @@ def __init__(self, adapter, name="Agilent 33220A Arbitrary Waveform generator",
     frequency = Instrument.control(
         "FREQ?", "FREQ %s",
         """Control the frequency of the output waveform, depending on the method
-        (float, strict from 1e-6 to 5e+6).""",
+        valid range depends on the current shape and supplying out-of-bound values will silently
+        be clipped by the device (float, strict from 1e-6 to 20e+6).""",
```
BenediktBurger: If the device clips the values, why do we limit the values in the driver?
Would it not be better to remove these limits in the docstring, the validator and the values field?

pymeasure/instruments/agilent/agilent33220A.py
```diff
@@ -258,8 +266,8 @@ def wait_for_trigger(self, timeout=3600, should_stop=lambda: False):
 
             if timeout != 0 and time() - t0 > timeout:
                 raise TimeoutError(
-                    "Timeout expired while waiting for the Agilent 33220A" +
-                    " to finish the triggering."
+                    "Timeout expired while waiting for the Agilent 33220A"
+                    + " to finish the triggering."
```
BenediktBurger: Actually, the `+`  is not needed as several strings in a parenthesis, without a comma in between, are considered a single string.

BenediktBurger:
As a Contributor, you may add yourself to the authors.txt file as well.

Thanks for offering the fix.

pymeasure/instruments/agilent/agilent33220A.py
Patschke: I don't know. The same behavior was already present for the combination of amplitude and offset. The device has a physical limit of the maximum output voltage, so a high amplitude would impose limits on the offset and vice versa. If setting invalid combinations, it would just be clipped. So I expected that this is how it's supposed to be done - I don't have any strong opinion what to put here. 

In principle, I can see why people could prefer getting an error when setting invalid stuff over just silently ignoring (and clipping) that. Given that this value is interpreted wildly different (dBm vs volt), it might just not be reasonable to have a validator here - it could even create a wrong sense of trusting in the validator?

pymeasure/instruments/agilent/agilent33220A.py
Patschke: I guess the `+` was there before and my autopep just moved it to another line...

pymeasure/instruments/agilent/agilent33220A.py
BenediktBurger: >  The same behavior was already present for the combination of amplitude and offset. 

We try to be backward compatible, but removing a validator does not break that. We do not have to keep bad code, just for keeping it.

> Given that this value is interpreted wildly different (dBm vs volt), it might just not be reasonable to have a validator here...

That is exactly my point: If the validator is doesn't serve any purpose, we should remove it.

Patschke: Removing the validator may break backwards compatibility. Someone could have some function that tries to set values without sanity checks beforehand. Instead, having try-except based error-handling to catch out-of-bounds values was an understandable way to go (let the driver handle values rather than the user-code - this allows to exchange a device much more easily). But that wasn't my point here. 

In the given setting, the driver is unable to determine what the correct bounds for the given situation are (as they may be subject to change). It can now either 
 * don't do any validation (expect the user do that themself)
 * do some generous validation (The  validator still does something useful in case of clearly wrong values, e.g. trying to use the AWG as laser by specifying a 300 THz output frequency, or setting the Voltage to 10 kV. However, it can't be relied on in all cases)

Given that the second option was chosen for offset and amplitude, I expected that this was the way-to-go. I can see how having a validator implies that "all values it accepts are indeed valid values", so I'll remove them where that's not possible.

BenediktBurger:
Thanks for your contribution.

PR #1343, opened 2025-08-04
Added ramp commands to heater channel

DF9AS:
Dear maintainers of pymeasure,

LakeShore temperature controllers can ramp the temperature to a given setpoint. I've added the commands to enable / disable the ramp and to set the ramp rate for the output loops.

### DIFF
--- a/pymeasure/instruments/lakeshore/lakeshore_base.py
+++ b/pymeasure/instruments/lakeshore/lakeshore_base.py
@@ -30,9 +30,6 @@ from time import sleep, time
 from pymeasure.instruments import Instrument, Channel
 from pymeasure.instruments.validators import strict_discrete_set
 
-log = logging.getLogger(__name__)
-log.addHandler(logging.NullHandler())
-
 
 class LakeShoreTemperatureChannel(Channel):
     """ Temperature input channel on a lakeshore temperature monitor. Reads the temperature in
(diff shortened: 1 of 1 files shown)


pymeasure/instruments/lakeshore/lakeshore_base.py
```diff
@@ -124,3 +121,32 @@ class LakeShoreHeaterChannel(Channel):
...
+    @property
+    def ramp_rate(self):
+        return self.ramp[1]
+
+    @ramp_rate.setter
+    def ramp_rate(self, value: float):
+        ramp_enabled = int(self.ramp_enabled)
+        self.ramp = ramp_enabled, value
+
+    rampst = Instrument.measurement(
```
BenediktBurger, line 149: Better `ramp_in_progress` or something similar. And then use a mapping from 0/1 to a boolean.

Also, the docstring should start with "Get" or "Measure" for a `measurement`.

pymeasure/instruments/lakeshore/lakeshore_base.py
```diff
@@ -124,3 +121,32 @@ class LakeShoreHeaterChannel(Channel):
         """Control the setpoint temperature
         in the preferred units of the control loop sensor."""
     )
+
+    ramp = Instrument.control(
+        'RAMP? {ch}',
+        'RAMP {ch}, %d, %f',
+        """Control the ramp settings. Needs and replies a tuple (0/1, rate)""",
+    )
```
BenediktBurger, line 129: What about a single value: `False` or 0 or something switches it off, while any number switches it on with that value?

That would make the

pymeasure/instruments/lakeshore/lakeshore_base.py
```diff
@@ -124,3 +121,32 @@ class LakeShoreHeaterChannel(Channel):
         """Control the setpoint temperature
         in the preferred units of the control loop sensor."""
     )
+
+    ramp = Instrument.control(
+        'RAMP? {ch}',
+        'RAMP {ch}, %d, %f',
+        """Control the ramp settings. Needs and replies a tuple (0/1, rate)""",
```
BenediktBurger, line 128: what is the units of the rate?

BenediktBurger (changes requested):
A few comments. Thanks for your contribution.

PR #1344, opened 2025-08-06
Add SPGU module support for Agilent B1500

ilev-sergey:


BenediktBurger:
A quick note on formatting: It makes it quite difficult to see, where are actual changes and where not.

So, it is best to restrict reformatting to the lines you change anyways. Especially as every formatter has slightly different opinions, so everyone will format it slightly differently.

If you want to format a whole file, because there are many issues or inconsistencies, it is best to do it in a single commit (best at the beginning), such that a reviewer can look at the changes without that reformatting, which should not change anything techically.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -155,6 +157,37 @@ def initialize_all_smus(self):
...
+        :type channel: int
+        :param spgu_type: SPGU type, e.g. ``'HRSPGU'``
+        :type spgu_type: str
+        :param name: SPGU name for pymeasure (data output etc.)
+        :type name: str
+        :return: SPGU instance
+        :rtype: :class:`.SPGU`
+        """
+        if channel in (list(range(101, 1101, 100)) + list(range(102, 1102, 100))):
+            channel = int(str(channel))
```
BenediktBurger: why this acrobatics: int -> str -> int?

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -155,6 +157,37 @@ def initialize_all_smus(self):
...
+        """Initializes SPGU instance by calling :class:`.SPGU`.
+
+        :param channel: SPGU channel
+        :type channel: int
+        :param spgu_type: SPGU type, e.g. ``'HRSPGU'``
+        :type spgu_type: str
+        :param name: SPGU name for pymeasure (data output etc.)
+        :type name: str
+        :return: SPGU instance
+        :rtype: :class:`.SPGU`
```
BenediktBurger: You can also use the shorter form to include the type:
`:param int channel: SPGU channel`

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -43,31 +47,27 @@
 ######################################
 
 
-class AgilentB1500(SCPIUnknownMixin, Instrument):
-    """ Represents the Agilent B1500 Semiconductor Parameter Analyzer
+class AgilentB1500(SCPIMixin, Instrument):
```
BenediktBurger, line 51: Thanks for fixing it.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -155,6 +157,37 @@ def initialize_all_smus(self):
...
+
+    def initialize_all_spgus(self):
+        """Initialize all SPGUs by querying available modules and creating
+        a SPGU class instance for each.
+        SPGUs are accessible via attributes ``.spgu1`` etc.
+        """
+        modules = self.query_modules()
+        i = 1
+        for channel, module_type in modules.items():
+            if module_type == "SPGU":
```
BenediktBurger, line 163: Why don't you use the number of module (maybe with `enumerate`) instead of the count of SPGU modules?

How can a user know, which module has which number?

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -175,6 +208,30 @@ def force_gnd(self):
         """
         self.write("DZ")
 
+    def restore_settings(self):
+        """Restores the settings of all channels to the state before
+        using `force_gnd`. (``RZ``)
```
BenediktBurger:
```suggestion
        using :meth:`force_gnd`. (``RZ``)
```
This :meth: will mark it as a method and link it (in the sphinx documentation) to the appropriate method.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -175,6 +208,30 @@ def force_gnd(self):
...
+        """
+        self.write("RZ")
+
+    control_mode = Instrument.control(
+        "ERMOD?",
+        "ERMOD %s",
+        """Control mode for the digital I/O ports. (``ERMOD``)""",
```
BenediktBurger:
```suggestion
        """Control the control mode for the digital I/O ports (:code:`ControlMode`). (``ERMOD``)""",
```

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -175,6 +208,30 @@ def force_gnd(self):
         """
         self.write("DZ")
 
+    def restore_settings(self):
+        """Restores the settings of all channels to the state before
```
BenediktBurger: Docstrings should be (according to PEP) in imperative mode:

```suggestion
        """Restore the settings of all channels to the state before
```

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -175,6 +208,30 @@ def force_gnd(self):
...
+    control_mode = Instrument.control(
+        "ERMOD?",
+        "ERMOD %s",
+        """Control mode for the digital I/O ports. (``ERMOD``)""",
+        get_process=lambda v: ControlMode(v),
+        set_process=lambda v: ControlMode(v).value,
+    )
+
+    def set_port_connection(self, port, status):
+        """Sets the connection status for a specific port. (``ERSSP``)
```
BenediktBurger:
```suggestion
        """Set the connection status for a specific port. (``ERSSP``)
```

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -155,6 +157,37 @@ def initialize_all_smus(self):
                 )
                 i += 1
 
+    def initialize_spgu(self, channel, name):
+        """Initializes SPGU instance by calling :class:`.SPGU`.
```
BenediktBurger:
```suggestion
        """Initialize a :class:`.SPGU` instance.
```

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -155,6 +157,37 @@ def initialize_all_smus(self):
...
+        self._spgu_names[channel] = name
+        spgu_reference = SPGU(self, channel, name)
+        self._spgu_references[channel] = spgu_reference
+        return spgu_reference
+
+    def initialize_all_spgus(self):
+        """Initialize all SPGUs by querying available modules and creating
+        a SPGU class instance for each.
+        SPGUs are accessible via attributes ``.spgu1`` etc.
+        """
```
BenediktBurger, line 160: According to [PEP-257](https://peps.python.org/pep-0257/), there should be a single summary line in imperative mode.
After a newline, there can be more content.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -155,6 +157,37 @@ def initialize_all_smus(self):
...
+        modules = self.query_modules()
+        i = 1
+        for channel, module_type in modules.items():
+            if module_type == "SPGU":
+                setattr(self, "spgu" + str(i), self.initialize_spgu(channel, "SPGU" + str(i)))
+                i += 1
+
     def pause(self, pause_seconds):
         """Pauses Command Execution for given time in seconds (``PA``)
```
BenediktBurger: Typically, we do not tell in a docstring, which command is sent to the device.
But to keep consistency in this device, we can keep them in.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -999,6 +1058,12 @@ def force_gnd(self):
         ``RZ`` (not implemented). (``DZ``)"""
         self.write("DZ %d" % self.channel)
 
+    def restore_settings(self):
+        """Restores the settings of the channel to the state before
+        using `force_gnd`. (``RZ``)
+        """
+        self.write("RZ %d" % self.channel)
```
BenediktBurger: Why don't you use a more convenient f-string:

```suggestion
        self.write(f"RZ {self.channel}")
```

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1590,6 +1655,238 @@ def __init__(self, smu_type):
         self.meas = Ranging(supported_meas_ranges, ranges, fixed_ranges=True)
 
 
+######################################
+# SPGU Setup
+######################################
+
+
+class SPGU:
```
BenediktBurger: This is a channel, right, at least in an abstract way? So please inherit from the `Channel` class!

See: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html

There is no problem to have Channels of Channels. They were designed that way.

You might even add these channels to the api documentation.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1625,7 +1922,7 @@ class ADCType(CustomIntEnum):
 
     HSADC = (0,)  #: High-speed ADC
     HRADC = (1,)  #: High-resolution ADC
-    HSADC_PULSED = (2,)  #: High-resolution ADC for pulsed measurements
+    HSADC_PULSED = (2,)  #: High-speed ADC for pulsed measurements
```
BenediktBurger: That still continues to work in all used places, as int is not tuple of one member?

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1590,6 +1655,238 @@ def __init__(self, smu_type):
...
+        "SPPER?",
+        "SPPER %f",
+        """Control the pulse period for SPGU channels (``SPPER``) in seconds (float).
+        Applies to all installed SPGU modules""",
+        validator=strict_range,
+        values=[2e-8, 10],
+    )
+
+    def set_output_mode(self, mode, condition=None):
+        """Sets the operating mode for SPGU channel outputs. This setting applies to
```
BenediktBurger:
```suggestion
        """Set the operating mode for SPGU channel outputs. This setting applies to
```

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1590,6 +1655,238 @@ def __init__(self, smu_type):
...
+    @property
+    def load_impedance(self):
+        """Control the load impedance (``SER``) in Ohm (float)."""
+        return self._spgu.ask(f"SER? {self.channel}")
+
+    @load_impedance.setter
+    def load_impedance(self, value):
+        self._spgu.write(f"SER {self.channel}, {value}")
+
+    def set_output_voltage(self, source=1, base_voltage=0, peak_voltage=0):
```
BenediktBurger, line 1657: Consider using a `output_voltage_setpoint` property instead.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1590,6 +1655,238 @@ def __init__(self, smu_type):
...
+    @property
+    def output_mode(self):
+        """Control the output mode of the SPGU channel. (``SPM``)
+        The SPGU operating mode must be set to PG with the SIM 0 command before setting the
+        output mode.
+
+        :param mode: Output mode
+        :type mode: :class:`.SPGUChannelOutputMode` or int
+        """
+        return self._spgu.ask(f"SPM? {self.channel}")
```
BenediktBurger: can't this be a `Channel.control`?

BenediktBurger (changes requested):
Thanks for your changes and sorry for the late response.

Consider adding protocol tests for some of the new features (especially Channels of Channels): https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html .

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: I don’t think using a property is a good option here, as it becomes less readable when multiple arguments are passed.

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: I don't really know, I think I didn't use this functionality in my measurements. Anyway, I think the change will break some existing scripts if anybody uses it, so I just fixed the typo

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: I think it'd be a bit complicated in this case.

The thing is, there are different modules for this instrument, and they can be installed in any available slots. So, for example, if SPGU module is installed in the 1st slot that corresponds to channels 101 and 102, it would be accessible through .spgu1. But then things get more complicated, as 1st slot is already taken.

For example, In my case, SMUs are installed in slots 4, 6, 8, 9, so they'd be accessible by smu4, smu6, etc, even though no smu1, 2, 3 are actually present.

I think it could be confusing for new users, as they are not familiar with the hardware itself, and it's not obvious why they should specifically address smu4, 6, 8, 9 instead of a simpler smu1, 2, 3, 4

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: If you don't mind, I'd like to keep it without inheritance for now. I ran into some issues when using Channels last time due to the above-mentioned features with dynamic module number. I think I wasn't able to resolve them by adding channels dynamically either. I'll try to refactor both SMU and SPGU modules into channels later and make a new PR if it'll work.

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: It can't be as SPGUChannel is not inherited from Channel atm. Because of that channel property that should be passed into command is not available as a static member

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: You can add channels also dynamically (with `add_child`), so I do not see any reason not to use Channel.
What are the problems exactly?

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: By the way: `Channel.control` and `Instrument.control` are exactly the same property creators (at the moment).
You can use both methods, even though the class (like here) does not inherit from Channel or Instrument, as the `Channel.control` is static and will create a property (independently of the parent class).

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: Ok, I see.
So, the user has to know his hardware, but that's fine.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -975,74 +996,80 @@ def status(self):
...
-        """ Force 0V immediately. Current Settings can be restored with
-        ``RZ`` (not implemented). (``DZ``)"""
-        self.write("DZ %d" % self.channel)
+        """Force 0V immediately. Current Settings can be restored with
+        ``RZ``. (``DZ``)"""
+        self.write(f"DZ {self.channel}")
+
+    def restore_settings(self):
+        """Restore the settings of the channel to the state before
+        using `force_gnd`. (``RZ``)
```
BenediktBurger:
```suggestion
        using :meth:`force_gnd`. (``RZ``)
```

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: I see it now that you have a "control mode", maybe you find a better wording?

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: Now I see it. It was already a tuple (with the comma), so no change in functionality, only in representation.

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: Well, didn't have any issues this time. Guess I did something wrong previously

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1567,39 +1550,29 @@ def __init__(self, smu_type):
...
-        """Start SPGU output (``SRP``)"""
-        self.write("SRP")
-
-    def stop_output(self):
-        """Stop SPGU output (``SPP``). The channel outputs the base voltage."""
-        self.write("SPP")
+    def __init__(self, parent, channel, **kwargs):
+        super().__init__(parent, channel, **kwargs)
+        self.ch1 = self.add_child(SPGUChannel, int(str(self.id) + "01"), prefix="ch")
+        self.ch2 = self.add_child(SPGUChannel, int(str(self.id) + "02"), prefix="ch")
```
BenediktBurger: Just a note for string manipulation: `f"{self.id}01"` will work instead of `str(self.id) + "01"`.
Those f-strings are quite powerful.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1524,56 +1498,246 @@ class SMUCurrentRanging():
...
-                            fixed_ranges=True)
+        self.meas = Ranging(supported_meas_ranges, ranges, fixed_ranges=True)
+
+
+######################################
+# SPGU Setup
+######################################
+
+
+class SPGU(Channel):
```
BenediktBurger, line 1553: Please add both channel classes to the documentation page (agilentB1500.rst), such that users can see also the API of the channels.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -480,86 +512,80 @@ def format_single(self, element):
...
+        self._data_format = self._data_formatting(f"FMT{output_format}", self._smu_names)
 
     ######################################
     # Measurement Settings
     ######################################
 
     @property
     def parallel_meas(self):
-        """ Enable/Disable parallel measurements.
+        """Enable/Disable parallel measurements.
```
BenediktBurger: As we are changing it, please make it compatible with the correct wording for properties:

```suggestion
        """Control whether parallel measurements are enabled (bool).
```

If all properties are rightly written, we could remove the B1500 from the list of "grandfathered  docstring instruments":
https://github.com/pymeasure/pymeasure/blob/37afa1baa8e9032cc66efa4e243f90743b724f0e/tests/instruments/test_all_instruments.py#L121

If more than a handful of properties have to be changed, it should be its own PR and not part of this one.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -910,16 +906,16 @@ def query_meas_ranges(self):
...
 
+
 ######################################
 # SMU Setup
 ######################################
 
 
-class SMU():
-    """ Provides specific methods for the SMUs of the Agilent B1500 mainframe
+class SMU:
```
BenediktBurger, line 915: Oh, that should be a Channel as well.

However, this should be another task, not in this PR. I added this device to https://github.com/pymeasure/pymeasure/issues/746

tests/instruments/agilent/test_agilentB1500.py
BenediktBurger, line 1: Please add the copyright note at the top of the test

BenediktBurger (approved):
Thanks for your rework.

A few small issues before we can merge.

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: There are quite a few, I'll change them in a separate PR then, along with other docs refinements

BenediktBurger: Would you mind to rename the method to `parallel_measurement` in order to be more clear?
Also, please change the wording of this docstring.

Changing the rest, can be done in another PR.

ilev-sergey: This property is not implemented as part of the PR, so the change will break backward compatibility, is it OK for you?

BenediktBurger: Oh sorry.
In this case, we leave it as is.

Thanks for pointing it out.

PR #1346, opened 2025-08-18
Srs ldc500series tec

fg-lumentum:
- Fixed typo in docstring (line 287)
- Added methods to wait and check for temperature stability, identical to the implementation for the Keithley 2510


BenediktBurger (approved):
Looks good, thanks for your work.

We cannot make the check code reused by both instruments, right?

fg-lumentum:
Technically you could, the syntax is the same with `temperature` and `temperature_setpoint`. I don't know whether it would be good practice to inherit the methods from a different instrument manufacturer though.

PR #1347, opened 2025-08-19
Keithley2400 Rework

fg-lumentum:
This is a major rework of the Keithley 2400. Where possible, I have kept undesirable methods and properties and labeled them as deprecated. As this is a large change with some subtle change in behaviour, the old version has been kept under `keithley2400_deprecated.py` in-case users encounter issues with the new implementation.

**Measurement**
- `current`, `voltage`, and `resistance` all used the command ":READ?".
- This meant that if one were to configure the device for measurement of current using ":SENS:FUNC 'CURR':FORM:ELEM CURR" (part of `measure_current`), and then get the property `voltage`, the quantity returned would actually be a current measurement. This would all pass silently.
- This has been changed by specifying the data format in the init (":FORM:ELEM VOLT, CURR, RES, TIME, STAT", giving the result of ":READ?" as the list `[voltage, current, resistance, time, status]`), and using the ":MEAS" commands.
- The ":MEAS" commands implicitly configure the device for the desired measurement. The processing to select the desired quantity from the returned list is done using `get_process_list=lambda v: v[index]`
- This also removes any need for explicitly configuring the device for any given measurement, so `measure_voltage`, `measure_current`, and `measure_resistance` can be deprecated.
- Finally, this makes for easy implementation of concurrent measurements through `measure_all`

**Measurement and sourcing configuration**
- The `apply_...` and `measure_...` methods set unnecessarily quantities 
- The `apply_...`  methods implicitly set the source range, and the compliance.
- The `measure_...` methods set the nplc, and the sense range (and the resistance mode for `measure_resistance`)
- These functions have been labeled as deprecated.
- Control of the measurement mode is now done implicitly by `voltage`, `current`, and `resistance`.
- Control of the source mode through `source_mode` is encouraged.
- Properties have been created for all of the quantities set implicitly by the `apply_...` and `measure_...` methods.

**Strict ranges**
- Truncated ranges were used heavily.
  - This was particularly problematic as most end users use mA or uA as their working units, making it very easy to forget to change units when setting the current in A and accidentally running a very high current through the device (a mistake I have made many times).
- All ranges have been switched to **strict ranges**

**Mapping**
- There was inconsistency in the format of strings, e.g. `source_mode` took "current" or "voltage", whereas `output_off_state` took "HIMP", "NORM", "ZERO", or "GUAR".
- This has been made consistent by making them all lower-case and descriptive.

**Testing**
- Comprehensive protocol and *with device* tests have been written.

**Other changes**
- Module has been re-ordered more logically.
- Almost every docstring has been re-written for consistency.

As a final comment, I understand that the implementation of many other Keithley instruments has been taken from the Keithley2400. As far as I am aware, none of it is inherited, it's all *copy and pasted* (evidenced by the fact that no other tests needed to be rewritten). It would be desirable to restructure this in future to inherit from the Keithley2400 class.


BenediktBurger:
Thanks for your work. The description sounds good. Let's see, when I will have time to look at it.

BenediktBurger:
Not yet. As it was a draft, I pushed it back on the PRs. Now I'll put it on top of the list to give you some feedback.

pymeasure/instruments/keithley/keithley2400_legacy.py
BenediktBurger, line 1: I think it is better to call it "legacy" than deprecated.

BenediktBurger, line 1: This file is just the old keithley2400 file, right?

pymeasure/instruments/keithley/keithley2400_deprecated.py
```diff
@@ -0,0 +1,743 @@
...
+from pymeasure.instruments.validators import truncated_range, strict_discrete_set
+
+from .buffer import KeithleyBuffer
+
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
```
BenediktBurger: In this case, rename it to ` Keithley2400Legacy` and import it in the init (and add it to the docs) etc.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -30,7 +30,7 @@
 
 from pymeasure.instruments import Instrument, SCPIMixin
 from pymeasure.errors import RangeException
-from pymeasure.instruments.validators import truncated_range, strict_discrete_set
+from pymeasure.instruments.validators import strict_range, strict_discrete_set
```
BenediktBurger: Strict ranges are good!

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
-        validator=strict_discrete_set,
-        values={"current": "CURR", "voltage": "VOLT"},
-        map_values=True,
-    )
+        # Ensures the data format is as expected. Setting `:FORM:ELEM` to anything else after
+        # initialization will break some part of the functionality.
+        # If comms in the init is undesirable, could put the formatter before any measurements
+        # e.g. for current, ":FORM:ELEM CURR;:MEAS:CURR?". I opted against this to reduce the
+        # number of commands sent to the instrument.
+        self.write(":FORM:ELEM VOLT, CURR, RES, TIME, STAT")
```
BenediktBurger: Init comms are not that nice (for tests etc.) and they have the problem, that a restarted device (without a restarted driver), won't have that setting.

However, it is better to send it once (like here) instead of sending it before every other command.

Please:
- make this a command, which can be called later on manually again.
- Clean up the comment (as I answered it). Probably the name and docstring of the method/property will describe it sufficiently.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
+    # Source properties #
 
     source_enabled = Instrument.control(
-        "OUTPut?",
-        "OUTPut %d",
-        """Control whether the source is enabled, takes
-        values True or False. The convenience methods :meth:`~.Keithley2400.enable_source` and
-        :meth:`~.Keithley2400.disable_source` can also be used.""",
+        "OUTP?",
+        "OUTP %d",
```
BenediktBurger: Why did you change it? It does not have any effect, right?
I just want to hear your opinion.

I personally would keep the slightly longer form, as it makes the commands more understandable, while the shorter command does not bring that much performance, but I don't have a strong opinion.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
         ":SOUR:DEL?",
         ":SOUR:DEL %g",
-        """ Control (floating) a manual delay for the source
-        after the output is turned on before a measurement is taken. When this
-        property is set, the auto delay is turned off. Valid values are
-        between 0 [seconds] and 999.9999 [seconds].""",
-        validator=truncated_range,
+        """Control a manual delay for the source after the output is turned on
+        before a measurement is taken (float strictly from 0 to 999.9999).
+        When this property is set, the auto delay is turned off.""",
```
BenediktBurger:
```suggestion
        """Control the manual delay in seconds for the source after the output is turned on
        before a measurement is taken (float strictly from 0 to 999.9999).
        When this property is set, the auto delay is turned off.""",
```

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
     )
 
     auto_zero = Instrument.control(
         ":SYST:AZER:STAT?",
         ":SYST:AZER:STAT %s",
-        """ Control whether the auto zero option is enabled. Valid values are
-        True (enabled) and False (disabled) and 'ONCE' (force immediate). """,
-        values={True: 1, False: 0, "ONCE": "ONCE"},
+        """Control whether the auto zero option is enabled
+        (bool or str, True (enabled), False (disabled), or 'ONCE' (force immediate)).""",
```
BenediktBurger: "ONCE" is the string sent to the device, while the property accepts the value "once"!

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
+            "guard": "GUAR",
+        },
+        map_values=True,
     )
 
-    line_frequency_auto = Instrument.control(
-        ":SYST:LFR:AUTO?",
-        ":SYST:LFR:AUTO %d",
-        """ Control the auto line frequency (boolean). """,
+    auto_output_off = Instrument.control(
```
BenediktBurger: Typically, we use such a name like `auto_output_enabled` as that makes boolean properties quite understandable.
However, due to backward compatibility, we have to think twice about renaming properties.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
+        ":SENS:AVER:TCON?",
+        ":SENS:AVER:TCON %s",
+        """Control the filter's type (str, strictly 'repeat' or 'moving'.""",
+        validator=strict_discrete_set,
+        values={
+            "repeat": "REP",
+            "moving": "MOV",
+        },
+        map_values=True,
+    )
```
BenediktBurger, line 329: For those two-value properties, boolean properties are quite handy, as true/false are easier to handle than string values.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
+        guard : I-Source is selected and set to 0A""",
         validator=strict_discrete_set,
-        values=[50, 60],
-        cast=int,
+        values={
+            "disconnect": "HIMP",
+            "normal": "NORM",
+            "zero": "ZERO",
+            "guard": "GUAR",
+        },
```
BenediktBurger, line 209: I like your new values, however, they are not backward compatible.
I would add the old values again, in order to be backward compatible, e.g. `"HIMP": "HIMP"`.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
-        validator=truncated_range,
+        ":SENS:CURR:RANG %g",
+        """Control the measurement current range in Amps (float, strictly from -1.05 to 1.05).
+        When set, the range selected will be the most sensitive range that will accommodate the
+        set value, and Auto-range is disabled.""",
+        validator=strict_range,
         values=[-1.05, 1.05],
     )
 
+    current_range_auto = Instrument.control(
```
BenediktBurger: Better `current_range_auto_enabled`, as that makes it clear, what true/false mean.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
+    )
+
     current_nplc = Instrument.control(
         ":SENS:CURR:NPLC?",
         ":SENS:CURR:NPLC %g",
-        """ Control (floating) the number of power line cycles
-        (NPLC) for the DC current measurements, which sets the integration period
-        and measurement speed. Takes values from 0.01 to 10, where 0.1, 1, and 10 are
-        Fast, Medium, and Slow respectively. """,
+        """Control the number of power line cycles (NPLC) (float, from strictly from 0.01 to 10).
```
BenediktBurger:
```suggestion
        """Control the number of power line cycles (NPLC) (float, strictly from 0.01 to 10).
```

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
     source_current_range = Instrument.control(
         ":SOUR:CURR:RANG?",
-        ":SOUR:CURR:RANG:AUTO 0;:SOUR:CURR:RANG %g",
-        """ Control the source current range in Amps (float, truncated from -1.05 to 1.05).
-        Auto-range is disabled when this property is set. """,
-        validator=truncated_range,
+        ":SOUR:CURR:RANG %g",
+        """Control the source current range in Amps (float, strictly from -1.05 to 1.05).
+        When set, the range selected will be the most sensitive range that will accommodate the
+        set value, and Auto-range is disabled.""",
```
BenediktBurger, line 468: No comma before and:
```suggestion
        set value and Auto-range is disabled.""",
```

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
+        """Control the souce voltage auto-range (bool).""",
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    # Voltage methods #
+
+    def ramp_to_voltage(self, target_voltage, steps=30, pause=20e-3):
+        """Ramp to a target voltage from the set voltage value over
```
BenediktBurger:
```suggestion
        """Ramp to a target voltage from the current voltage value over
```

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
-
-    wires = Instrument.control(
-        ":SYSTEM:RSENSE?",
-        ":SYSTEM:RSENSE %d",
-        """ Control the number of wires in use for resistance measurements
-        (integer, strictly 2 or 4). """,
+    resistance_range_auto = Instrument.control(
+        ":SENS:RES:RANG:AUTO?",
+        ":SENS:RES:RANG:AUTO %d",
+        """Control the measurement resistance auto-range (bool).""",
```
BenediktBurger:
```suggestion
        """Control whether resistance measurement auto-range is enabled (bool).""",
```

If this is a new property, `resistance_range_auto_enabled` would be better to make it clearer and consistent with other instruments.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
-        map_values=False,
-    )
+    def set_timed_arm(self, interval):
+        """Set up the measurement to be taken with the internal
+        trigger at a variable sampling rate defined by the interval
+        in seconds between sampling points
+        """
+        if interval > 99999.99 or interval < 0.001:
+            raise RangeException("Keithley 2400 can only be time triggered between 1 mS and 1 Ms")
+        self.write(":ARM:SOUR TIM;:ARM:TIM %.3f" % interval)
```
BenediktBurger, line 1088: Wouldn't be a property better? Or does it also start a measurement?

There is a good rule of thumb: If it is an action (do this or that) use a method, if it is a value, use a property. "Timed_arm" seems to be a property to me.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
+        line number, with the option of supplying the part of the
+        measurement after which the trigger should be generated
+        (default to delay, which is right before the measurement)
 
-    def measure_resistance(self, nplc=1, resistance=2.1e5, auto_range=True):
-        """Configure the measurement of resistance.
+        :param line: A trigger line from 1 to 4
+        :param after: An event string that determines when to trigger
+        """
+        self.write(":TRIG:OUTP %s;:TRIG:OLIN %d;" % (after, line))
```
BenediktBurger, line 1125: As above, this seems to be more appropriate for a property, than an action (method).

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
-        if auto_range:
-            self.write(":SENS:RES:RANG:AUTO 1;")
-        else:
-            self.resistance_range = resistance
-        self.check_errors()
+        self.write(f":SYST:BEEP {frequency:g}, {duration:g}")
 
-    def measure_voltage(self, nplc=1, voltage=21.0, auto_range=True):
-        """Configure the measurement of voltage.
+    def triad(self, base_frequency, duration):
```
BenediktBurger, line 1147: As stated above: A method should be an action, so "sound_triad" or similar is a better method name, as it contains a verb (an action).

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
+
+    # Misc methods #
+
+    def use_rear_terminals(self):
+        """Enable the rear terminals for measurement, and disable the front terminals."""
+        self.terminals = "rear"
+
+    def use_front_terminals(self):
+        """Enable the front terminals for measurement, and disable the rear terminals."""
+        self.terminals = "rear"
```
BenediktBurger: A boolean property makes it easier:
`front_terminals_enabled` could do the whole thing at once. The docstring should state though, that it switches between front/back terminals.

Ah, I see, we had these methods already present, so at least they should remain in the code, maybe deprecated.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -46,401 +46,740 @@ class Keithley2400(KeithleyBuffer, SCPIMixin, Instrument):
...
+    ##############
+
+    def auto_range_source(self):
+        """Configure the source to use an automatic range."""
+        warn(
+            "Deprecated, recommended to explicitly set the auto range for the desired source "
+            "using `Keithley2400.source_current_range_auto` "
+            "and/or `Keithley2400.source_voltage_range_auto`.",
+            FutureWarning,
+        )
```
BenediktBurger: For proper deprecation see #672.

1. The docstring should contain a deprecation note (including the version where deprecation starts, i.e. the next release). There you do not have to name, what is deprecated, as it is tied to the method.
2. The warning should contain **what** is deprecated, as you will see it in the logs, but you do not know, which method raised the warning.
3. For these deprecations, please add already a note to the changelog (under "upcoming release")


```suggestion
        """Configure the source to use an automatic range."""
        warn(
            "Deprecated to use `auto_range_source`, instead, set explicitly the auto range for the desired source "
            "using `Keithley2400.source_current_range_auto` "
            "and/or `Keithley2400.source_voltage_range_auto`.",
            FutureWarning,
        )
```

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -539,172 +874,18 @@ def error(self):
...
-        """Get the minimum resistance from the buffer."""
-        return self.minimums[2]
-
-    @property
-    def std_resistance(self):
-        """Get the resistance standard deviation from the buffer."""
-        return self.standard_devs[2]
-
     def status(self):
+        warn("Deprecated to use `status` method, use `status` property instead.", FutureWarning)
```
BenediktBurger: That won't work, as this status method overwrites the status property! You have to delete it (breaking change).
Does the status property work for this instrument (the command differs)?

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -539,172 +874,18 @@ def error(self):
...
         return self.ask("status:queue?;")
 
     def RvsI(self, startI, stopI, stepI, compliance, delay=10.0e-3, backward=False):
+        # Fred Gillard, fg-lumentum:
+        #   I would seriously consider removing this method, it doesn't work and is undocumented.
+        #   Internal sweeps offer a fractional time advantage over a sweep managed through python,
+        #   but due to the complex configuration required more care needs to be taken in setting
+        #   one up than shown here.
```
BenediktBurger: Why doesn't it work?
What is not documented? How it works?

You mean, the advantage of internal sweeps is only marginal (very small) compared to a python loop, but setup is more difficult, therefore, you suggest to write a python loop, right?

tests/instruments/keithley/test_keithley2400.py
BenediktBurger, line 1: Thanks for expanding tests.

pymeasure/instruments/keithley/keithley2400_legacy.py
BenediktBurger, line 1: Is there any breaking change, why your rework should not work?

tests/instruments/keithley/test_keithley_buffer.py
BenediktBurger, line 1: Shouldn't these tests be part of the test_keithley2400 file as you test that class.

BenediktBurger (changes requested):
Sorry for the delay.

Thanks for your work and patience.

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: I think I did it to make it consistent with the other commands all being shortform. I get your point though, I prefer the longform commands so I've switched the other commands to using them.

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: Auto output off is how it's called in the Keithley2400 documentation

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: I'd like to avoid current in any meaning other than the electrical context here. We could use present?

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: I've just gone ahead and deprecated this

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: Again, deprecated

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: This is what it was called in the previous version, we could deprecated `triad`, but for what is a rarely used function I don't think it's worth it

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: The status property works, I have no idea why that function was included, I have removed it!

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: Yeah, the advantage of internal sweeps is marginal. The procedure for setting up an internal sweep is laid out in the documentation, but this method doesn't match. I have removed it from this file.

pymeasure/instruments/keithley/keithley2400_legacy.py
fg-lumentum: No, I have tested it with the instrument and it all looks to be functional!

tests/instruments/keithley/test_keithley_buffer.py
fg-lumentum: I put them in a separate module as buffer.py is a separate module. I have removed it for now as I am honestly not particularly familiar about how Keithley buffers work.

pymeasure/instruments/keithley/__init__.py
BenediktBurger, line 1: Please add a note to the changelog (under the section "upcoming changes") regarding the move of the instrument class.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -39,592 +43,1065 @@
...
-        values={"current": "CURR", "voltage": "VOLT"},
-        map_values=True,
-    )
+    def reset_data_format(self):
+        """Resets the data format to the format expected by :class:`Keithley2400`.
+
+        The expected data format is [<current>, <voltage>, <resistance>, <time>, <status>].
+
+        .. caution::
+           Changing the data format with the ":FORMAT:ELEMENTS" command after initialization
```
BenediktBurger, line 77: You mean, if someone uses `inst.write(":FORMAT:ELEMENTS...")`, right?

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -39,592 +43,1065 @@
...
-        """ Control (string) the source mode, which can
-        take the values 'current' or 'voltage'. The convenience methods
-        :meth:`~.Keithley2400.apply_current` and :meth:`~.Keithley2400.apply_voltage`
-        can also be used. """,
-        validator=strict_discrete_set,
-        values={"current": "CURR", "voltage": "VOLT"},
-        map_values=True,
-    )
+    def reset_data_format(self):
+        """Resets the data format to the format expected by :class:`Keithley2400`.
```
BenediktBurger: Docstrings (at least the initial sentence) should be in imperative mode telling what the method should do ([PEP 257](https://peps.python.org/pep-0257/)), so:

```suggestion
        """Reset the data format to the format expected by :class:`Keithley2400`.
```

pymeasure/instruments/keithley/keithley2400.py
BenediktBurger: Yes, "present" would be fine.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -39,592 +43,1065 @@
...
-        self.disable_output_trigger()
-        self.trigger_immediately()
-
     def set_timed_arm(self, interval):
         """Set up the measurement to be taken with the internal
         trigger at a variable sampling rate defined by the interval
         in seconds between sampling points
+
+        .. deprecated:: 0.16
+           Set :attr:`~.arm_source` to "time" and control interval using :attr:`~.arm_timer`.
```
BenediktBurger:
```suggestion
           Instead set :attr:`~.arm_source` to "time" and control the interval using :attr:`~.arm_timer`.
```

Also in the deprecation note below.

Please add this deprecation to the changelog file.

pymeasure/instruments/keithley/keithley2400.py
BenediktBurger: Ok, then we leave it as is.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -634,128 +1111,121 @@ def output_trigger_on_external(self, line=1, after="DEL"):
...
+        validator=strict_discrete_set,
+        values={True: "FRON", False: "REAR"},
+        map_values=True,
+    )
 
-    def use_rear_terminals(self):
-        """Enable the rear terminals for measurement, and
-        disable the front terminals."""
-        self.write(":ROUT:TERM REAR")
+    def use_rear_terminals(self):  # Included for backwards compatibility
```
BenediktBurger: Instead of the comment, please add a deprecation note and warning, such that we can remove these methods one day.

BenediktBurger (commented):
Thank you very much for your rework.

Here a few more comments. We're getting close.

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: Yes, I do, is that clear enough?

pymeasure/instruments/keithley/keithley2400.py
fg-lumentum: Would you like me to add all the deprecated properties and methods to the changelog?

BenediktBurger: Depending on the number of deprecations. At least a note, that several methods/properties are deprecated.

pymeasure/instruments/keithley/keithley2400.py
```diff
@@ -39,592 +43,1067 @@
...
+        validator=strict_discrete_set,
+        values=["ON", "OFF"],
+        map_values=False,
+        get_process=_deprecate_process(
+            "Deprecated to use `Keithley2400.filter_state`. ",
+            "Instead use `Keithley2400.filter_enabled`.",
+        ),
+        set_process=_deprecate_process(
+            "Deprecated to use `Keithley2400.filter_state`. ",
+            "Instead use `Keithley2400.filter_enabled`.",
```
BenediktBurger: This makes the doctests and other tests fail:
You separate the two strings with a comma, therefore, they are two parameters, while `_deprecate_process` only takes one.

I understand that you want to combine both lines. You have to remove the commata in order to do this:

```suggestion
            "Deprecated to use `Keithley2400.filter_state`. "
            "Instead use `Keithley2400.filter_enabled`."
        ),
        set_process=_deprecate_process(
            "Deprecated to use `Keithley2400.filter_state`. "
            "Instead use `Keithley2400.filter_enabled`."
```

This happens in several places.

BenediktBurger (approved):

PR #1354, opened 2025-09-09
Wait for sweep completer timeout

fg-lumentum:
Increased timeout to 1 hour for `wait_for_sweep_complete` as these sweeps can take a very long time
Corrected the corresponding docstring (previously referenced temperature for no reason)


fg-lumentum:
Sweeps can be configured to run quickly, sub tenth of a second, as well. It could be worth polling rapidly for up to, say, 20s and then polling every second thereafter?

BenediktBurger:
Thanks for the clarification.

If you do not see a problem, we can leave it as is. Maybe the device communication takes longer resulting implicitly in a sleep (waiting for the response).

PR #1355, opened 2025-09-11
Fix racal read

### LINKED ISSUE #1350: Bug: Instrument Racal1992. Differing expectations in the length of instruments message
### How to reproduce

```python

from pymeasure.instruments.racal import Racal1992

racal_fc = Racal1992("GPIB::10")
racal_fc.operating_mode = "frequency_a"

racal_fc.measured_value
```

When the `measured_value` method is called we see the following:

```
"Length of instrument response must always be 19 characters"
```

This can be traced to the `Racal1992.decode`  static method in `pymeasure/instruments/racal/racal1992.py`
as it expects an argument length of 19 bytes bytes in length.

The `measured_value` method calls `read_and_decode`:

```python

    def read_and_decode(self, allowed_types=None):
       v = self.read_bytes(21).decode('utf-8')
       return Racal1992.decode(v, allowed_types)
```

`read_bytes` gets 21 bytes, which is correct (19 bytes + 2 bytes termination characters CR+LF).
The problem is the termination characters are not removed before sent to `decode`, which
expects a message to be 19 bytes.

### Possible solutions

Remove any whitespace before being sent to `Racal1992.decode`. This will not impact the 19 bytes message which
does not contain any whitespace. Note the `read()` method would still return 21 bytes.

```python

    def read(self):
        return self.read_bytes(21).decode('utf-8')
```
Another approach would be pass the 21 byte message to `Racal1992.decode` and let it take responsibility for
dealing with/processing the message.

I'm willing to take responsibility for this issue

BenediktBurger:
Closes #1350

### DIFF
--- a/pymeasure/instruments/racal/racal1992.py
+++ b/pymeasure/instruments/racal/racal1992.py
@@ -104,7 +104,9 @@ class Racal1992(Instrument):
         This here, for example, is math constant Z: MZ+001.00000000E+00
         """
         if len(v) != 19:
-            raise ReturnValueError("Length of instrument response must always be 19 characters")
+            raise ReturnValueError(
+                f"Length of instrument response must always be 19 characters, '{v}'"
+            )
 
         val_type = v[0:2]
         val = float(v[2:19])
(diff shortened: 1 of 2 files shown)


BenediktBurger:
@coderabbitai review

PR #1356, opened 2025-09-12
Add Keysight PNAs

Falk3579:
Add support for the Keysight PNA vector network analyzer family.

### DIFF
--- a/pymeasure/instruments/keysight/keysightPNA.py
+++ b/pymeasure/instruments/keysight/keysightPNA.py
@@ -240,9 +240,9 @@ class KeysightPNA(SCPIMixin, Instrument):
     """A class representing a Keysight PNA vector network analyzer.
 
     :param str data_format: set the :attr:`.data_format` attribute to strictly ``ascii``,
-    ``real32`` or ``real64``.
-    :param bool byte_order_swapped: Set the :attr:`.byte_order_swapped` attribute for 
-    the ``real32`` and ``real64`` data formats.
+        ``real32`` or ``real64``.
+    :param bool byte_order_swapped: Set the :attr:`.byte_order_swapped` attribute for
+        the ``real32`` or ``real64`` data format.
 
     .. code-block:: python
 


pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,354 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # add the 15 markers to traces
+        for marker in range(1, 16):
+            self.add_child(Marker,
+                           id=marker,
+                           collection="markers",
+                           prefix="mkr_",
+                           )
```
BenediktBurger: `add_child`  is there for dynamic addition of (sub)channels.
In this case, the "MultiChannelCreator" is better suited to add that many channels of the same type, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html#adding-multiple-channels-with-multichannelcreator

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,354 @@
...
+        "SENS{ch}:SWE:POIN?",
+        """Get the number of points of the channel (int).""",
+        cast=int,
+        )
+
+    measurements = Channel.measurement(
+        "SYST:MEAS:CAT? {ch}",
+        """Get the measurement trace numbers of the channel (list of int).""",
+        preprocess_reply=lambda v: v.strip('"'),
+        get_process=lambda v: [v],  # always return a list, even if only one trace exists
```
BenediktBurger, line 234: This will always wrap the result in a list. So, if the result is already a list, you will get a list of lists.

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,354 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # add the measurement traces
+        for trace in self.measurements:
+            self.add_child(Trace,
+                           id=trace,
+                           collection="traces",
+                           prefix="tr_",
+                           )
```
BenediktBurger: Could the associated traces change later?
In this case, I would create a method for the child creation.

Could it be somehow fixed with a ChannelCreator?

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,354 @@
...
+        )
+
+        self.data_format = data_format
+        self.byte_order = byte_order
+
+        # add the active channels
+        for channel in self.measurement_channels:
+            self.add_child(MeasurementChannel,
+                           id=channel,
+                           )
```
BenediktBurger: How do you handle, whenever another channel is added via the device's front panel?
Wouldn't it be better to add all channels at startup (via ChannelCreator or MultiChannelCreator) and these channels will complain whenever they are not active but are called?

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def abort(self):
+        """Stop all sweeps.
+
+        After the stop the sweep(s) resume per current trigger settings.
```
BenediktBurger: Maybe something like: "Note that the configured trigger will restart/resume the sweeps."?
(Depending whether they are resumed or restarted).

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,354 @@
...
+        """
+        self.write("ABOR")
+
+    def load_state(self, file_name):
+        """Load an instrument state from file.
+
+        :param file_name: str, e.g. ``D:/my_pna_state.csa``
```
BenediktBurger:
```suggestion
        :param str file_name: e.g. ``D:/my_pna_state.csa``
```
You can state the (python) type between the `param` keyword and name.
Actually, we started to allow Python typing, so you could add it to the signature as `def load_state(self, file_name: str):`, if you want to.

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,354 @@
...
+        "FORM:BORD %s",
+        """
+        Control the byte order used for data transfer (strictly ``normal`` or ``swapped``).
+        """,
+        map_values=True,
+        validator=strict_discrete_set,
+        values={"normal": "NORM",
+                "swapped": "SWAP",
+                },
+        )
```
BenediktBurger: Why not `byte_order_swapped` as a boolean property (True/False)?

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,354 @@
...
+        """Get the measurement trace numbers of the channel (list of int).""",
+        preprocess_reply=lambda v: v.strip('"'),
+        get_process=lambda v: [v],  # always return a list, even if only one trace exists
+        cast=int,
+        )
+
+
+class KeysightPNA(SCPIMixin, Instrument):
+    """A class representing a Keysight PNA vector network analyzer.
+
```
BenediktBurger, line 241: Please add a section with the init parameters, which are not self explanatory, especially, what the byte_order and the data_format do and what the options are.
You can link to the property (on the read the docs documentation page) with
```
:attr:`.byte_order`
```

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,354 @@
...
+        "*OPT?",
+        """Get the device options installed.""",
+        preprocess_reply=lambda v: v.strip('"'),
+        cast=str,
+        )
+
+    output_enabled = Instrument.control(
+        "OUTP?",
+        "OUTP %d",
+        """Control the RF power output status of the sources (bool).""",
```
BenediktBurger: The word "whether" helps often with booleans:

```suggestion
        """Control whether the RF power output of the sources is enabled (bool).""",
```
Does it deactivate all channels at once?

tests/instruments/keysight/test_keysightPNA.py
BenediktBurger, line 1: Please add the copyright notice at the top of the file.

tests/instruments/keysight/test_keysightPNA.py
```diff
@@ -0,0 +1,360 @@
+import pytest
+import numpy as np
+
+from pymeasure.test import expected_protocol
+from pymeasure.instruments.keysight import KeysightPNA
+
+# communication during class initialization
```
BenediktBurger: maybe a comment is good to state, that you initialize with a certain configuration of channels and traces, as  future contributors won't know directly, why they have to use certain channels.

tests/instruments/keysight/test_keysightPNA.py
```diff
@@ -0,0 +1,360 @@
...
+                 ("CALC1:MEAS3:X:AXIS:UNIT?", "FREQ"),
+                 ],
+            ) as inst:
+                assert "FREQ" == inst.ch_1.tr_3.x_unit
+
+
+@pytest.mark.parametrize("channel, trace",
+                         [(1, 4),
+                          (2, 3),
+                          (5, 7),
```
BenediktBurger: Consider reducing the number of tests, if they do not provide any benefit, as each test adds up in the pipeline.
Probably a single channel /trace would be sufficient.

Or do you test for something specific with these three cases?

Also for the tests below.

tests/instruments/keysight/test_keysightPNA_with_device.py
```diff
@@ -0,0 +1,224 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+# Call signature:
```
BenediktBurger, line 25: please state, that you have to exchange it for your own address and that these are just examples.

pymeasure/instruments/keysight/keysightPNA.py
Falk3579: Yes, it deactivates the RF power in all channels at once.

pymeasure/instruments/keysight/keysightPNA.py
Falk3579: `get_process` is only applied if the return value is a single value. If the return value is a list then `get_process_list` would be applied instead. `get_process_list` is not used here.
I tried it with a single channel and multiple channels. It always gave one list.

tests/instruments/keysight/test_keysightPNA_with_device.py
Falk3579: Done.

pymeasure/instruments/keysight/keysightPNA.py
Falk3579: Changed as recommended

pymeasure/instruments/keysight/keysightPNA.py
Falk3579: "MultiChannelCreator" is used now.

pymeasure/instruments/keysight/keysightPNA.py
Falk3579: Done as recommended.

tests/instruments/keysight/test_keysightPNA.py
Falk3579: I reduced the tests to a single channel/trace/marker as suggested.

tests/instruments/keysight/test_keysightPNA.py
Falk3579: Done

pymeasure/instruments/keysight/keysightPNA.py
Falk3579: I improved the docstring for the init parameters.

pymeasure/instruments/keysight/keysightPNA.py
Falk3579: Yes, the traces could change later. I implemented an update function for the traces.

The ChannelCreator can't be used since it is not known which trace numbers exist in each channel.

pymeasure/instruments/keysight/keysightPNA.py
Falk3579: That's a good point. Thanks for the hint.
I added an update function for the channels. It is now called when an instrument state is loaded.

I prefer to get an AttributeError if a channel is not existing rather then waiting for timeouts.

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,401 @@
...
+    y = Channel.measurement(
+        "CALC{{{{ch}}}}:MEAS{{tr}}:MARK{mkr}:Y?",
+        """Get the marker's Y value ([float, float]).""",
+        )
+
+
+class Trace(Channel):
+    """A class representing a Keysight PNA measurement trace."""
+
+    # add the 15 markers to the trace
```
BenediktBurger: It explains what you do, but that you can read from the code. Comments are useful to describe, **why** you do something, not what.

So, please remove it.

pymeasure/instruments/keysight/keysightPNA.py
```diff
@@ -0,0 +1,401 @@
...
+        )
+
+
+class KeysightPNA(SCPIMixin, Instrument):
+    """A class representing a Keysight PNA vector network analyzer.
+
+    :param str data_format: strictly ``ascii``, ``real32`` or ``real64``
+    :param bool byte_order_swapped:
+
+    See :attr:`.data_format` and :attr:`.byte_order_swapped` for more information.
```
BenediktBurger: I suggest to include the links in the description (maybe even a word about what it is):

```suggestion
    :param str data_format: set the :attr:`.data_format` attribute to strictly ``ascii``, ``real32`` or ``real64``.
    :param bool byte_order_swapped: Set the :attr:`.byte_order_swapped` attribute.
```

pymeasure/instruments/keysight/keysightPNA.py
Falk3579: Comment removed.

BenediktBurger:
Thank you!

PR #1359, opened 2025-09-24
AgilentE5270B: add 'display.enabled' property

Falk3579:
- add property "display_enabled' to AgilentE5270B to allow data display during remote operation

### DIFF
--- a/pymeasure/instruments/agilent/agilentE5270B.py
+++ b/pymeasure/instruments/agilent/agilentE5270B.py
@@ -35,7 +35,7 @@ class Display(Channel):
 
     enabled = Instrument.setting(
         "RED %d",
-        """Set the display during remote operation (bool).""",
+        """Set whether the display is enabled during remote operation (bool).""",
         map_values=True,
         values={True: 1, False: 0},
         )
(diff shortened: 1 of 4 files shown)


pymeasure/instruments/agilent/agilentE5270B.py
```diff
@@ -30,6 +30,17 @@
 log.addHandler(logging.NullHandler())
 
 
+class Display(Channel):
+    """A class representing the Agilent E5270B display."""
+
+    enabled = Instrument.setting(
```
BenediktBurger, line 36: Why did you choose to add this property to a dedicated class?

Are there more display properties or is this the only one?
In the latter case, I'd add a `display_enabled` property to the instrument class instead of an additional Channel.

Falk3579: There are more display related properties/methods that I want to implement in the next time. That's why I have added a dedicated class.

pymeasure/instruments/agilent/agilentE5270B.py
```diff
@@ -30,6 +30,17 @@
...
+        "RED %d",
+        """Set the display during remote operation (bool).""",
```
BenediktBurger:
```suggestion
        """Set whether the display is enabled during remote operation (bool).""",
```

BenediktBurger (approved):
LGTM

PR #1360, opened 2025-09-25
Fix Keithley2000 voltage_nplc

d-el:
Fix function 'voltage_nplc'


BenediktBurger:
Thank you for finding and fixing it!

PR #1361, opened 2025-09-26
Thorlabs mbx series

fg-lumentum:
Added control of the Thorlabs mbx series of modulator bias controllers
Includes protocol tests

### DIFF
--- a/pymeasure/instruments/thorlabs/thorlabsmbxseries.py
+++ b/pymeasure/instruments/thorlabs/thorlabsmbxseries.py
@@ -249,7 +249,20 @@ class ThorlabsMBXSeries(SCPIMixin, Instrument):
         t0 = time()
         while time() < t0 + duration:
             r, g, b = colorsys.hsv_to_rgb(random.random(), 1, 1)
-            self.rgb_red, self.rgb_green, self.rgb_blue = [int(x * 100) for x in (r, g, b)]
+            self.rgb_red, self.rgb_green, self.rgb_blue = [round(x * 100) for x in (r, g, b)]
+            sleep(pause)
+
+        self.rgb_power = LedPowerMode.OFF
+
+    def rainbow(self, duration=10, pause=0):
+        self.rgb_power = LedPowerMode.RGB
+
+        t0 = time()
+        h = 0
+        while time() < t0 + duration:
+            r, g, b = colorsys.hsv_to_rgb(h % 600 / 600, 1, 1)
+            self.rgb_red, self.rgb_green, self.rgb_blue = [round(x * 100) for x in (r, g, b)]
+            h += 1
             sleep(pause)
 
         self.rgb_power = LedPowerMode.OFF
--- a/pymeasure/instruments/thorlabs/thorlabsmbxseries.py
+++ b/pymeasure/instruments/thorlabs/thorlabsmbxseries.py
@@ -54,7 +54,7 @@ class VoaMode(IntEnum):
     POWER = 1
 
 
-class LedPowerMode(IntEnum):
+class RgbPowerMode(IntEnum):
     OFF = 0
     RGB = 1
     WHITE = 2
@@ -197,8 +197,8 @@ class ThorlabsMBXSeries(SCPIMixin, Instrument):
     rgb_power = Instrument.control(
         "RGB:POWER:?",
         "RGB:POWER: %d",
-        """Control the under-chassis LED accent lighting mode (LedPowerMode enum).""",
-        get_process=lambda v: LedPowerMode(v),
+        """Control the under-chassis LED accent lighting mode (RgbPowerMode enum).""",
+        get_process=lambda v: RgbPowerMode(v),
         set_process=lambda v: v.value,
     )  # pragma: no cover
 
@@ -207,9 +207,8 @@ class ThorlabsMBXSeries(SCPIMixin, Instrument):
         "RGB:RED: %d",
         """Control the brightness of the red under-chassis accent lighting LEDs
         (int, strictly in range 0 to 100).""",
-        validator=strict_discrete_range,
+        validator=lambda v, vs: strict_discrete_range(v, vs, 1),
         values=(0, 100),
-        values_kwargs={"step": 1},
     )  # pragma: no cover
 
     rgb_green = Instrument.control(
@@ -217,9 +216,8 @@ class ThorlabsMBXSeries(SCPIMixin, Instrument):
         "RGB:GREEN: %d",
         """Control the brightness of the green under-chassis accent lighting LEDs
         (int, strictly in range 0 to 100).""",
-        validator=strict_discrete_range,
+        validator=lambda v, vs: strict_discrete_range(v, vs, 1),
         values=(0, 100),
-        values_kwargs={"step": 1},
     )  # pragma: no cover
 
     rgb_blue = Instrument.control(
@@ -227,9 +225,8 @@ class ThorlabsMBXSeries(SCPIMixin, Instrument):
         "RGB:BLUE: %d",
         """Control the brightness of the blue under-chassis accent lighting LEDs
         (int, strictly in range 0 to 100).""",
-        validator=strict_discrete_range,
+        validator=lambda v, vs: strict_discrete_range(v, vs, 1),
         values=(0, 100),
-        values_kwargs={"step": 1},
     )  # pragma: no cover
 
     rgb_white = Instrument.control(
@@ -237,13 +234,12 @@ class ThorlabsMBXSeries(SCPIMixin, Instrument):
         "RGB:BLUE: %d",
         """Control the brightness of the white under-chassis accent lighting
         (int, strictly in range 0 to 100).""",
-        validator=strict_discrete_range,
+        validator=lambda v, vs: strict_discrete_range(v, vs, 1),
         values=(0, 100),
-        values_kwargs={"step": 1},
     )  # pragma: no cover
 
     def disco(self, duration=10, pause=0.5):  # pragma: no cover
-        self.rgb_power = LedPowerMode.RGB
+        self.rgb_power = RgbPowerMode.RGB
 
         t0 = time()
         while time() < t0 + duration:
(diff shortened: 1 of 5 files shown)


pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,267 @@
...
+        values=(0.01, 100),
+    )
+
+    voa_power = Instrument.measurement(
+        "VOA:TAP:MW?",
+        """Measure the VOA output optical power, in mW (float).""",
+    )
+
+    # === RGB CONTROL ===
+    # This section is excluded from coverage because it is purely aesthetic
```
BenediktBurger: Even if the effect is aesthetic, please do not exclude it from coverage.

fg-lumentum:
Re-opened as the instrument does actually communicate over serial, you just have to use the RS-232 port with a null-modem adapter. I'll make sure to put this in the class docstring.

On this note, you have to use a SerialAdapter with kwargs ``baudrate = 115200, write_termination="\r", read_termination="\r"``. What's the recommended way to manage this? I presume it's a bad idea to hardcode this requirement as follows?

```
class ThorlabsMBXSeries(SCPIMixin, Instrument):

    def __init__(self, port, name="ThorlabsMBXSeries modulator bias controller", **kwargs):
        
        adapter = SerialAdapter(
            port=port,
            baudrate=115200,
            timeout=1,
            write_termination="\r",
            read_termination="\r",
        )

```

BenediktBurger:
> Re-opened as the instrument does actually communicate over serial, you just have to use the RS-232 port with a null-modem adapter. I'll make sure to put this in the class docstring.
> 
> On this note, you have to use a SerialAdapter with kwargs `baudrate = 115200, write_termination="\r", read_termination="\r"`. What's the recommended way to manage this? I presume it's a bad idea to hardcode this requirement as follows?
> 
> ```
> class ThorlabsMBXSeries(SCPIMixin, Instrument):
> 
>     def __init__(self, port, name="ThorlabsMBXSeries modulator bias controller", **kwargs):
>         
>         adapter = SerialAdapter(
>             port=port,
>             baudrate=115200,
>             timeout=1,
>             write_termination="\r",
>             read_termination="\r",
>         )
> 
>         super().__init__(adapter, name, **kwargs)
> ```

See here: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#single-interface

You can define the defaults in the super init call, such that an (automatically generated) VISA adapter will be configured correctly.

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+        MZM - Mach Zender Modulator
+        VOA - Variable Optical Attenuator
+    """
+
+    def __init__(self, adapter, name="ThorlabsMBXSeries modulator bias controller", **kwargs):
+
+        kwargs.setdefault("baud_rate", 115200)
+        kwargs.setdefault("timeout", 1000)
+        kwargs.setdefault("write_termination", "\r")
+        kwargs.setdefault("read_termination", "\r")
```
BenediktBurger: Can you change these values?
If not, just add them to the super init call as hardcoded values.
If yes, consider putting them explicitly in the __init__ signature, depending whether you think they might be used more often or not (in this case kwargs.setdefault is fine).

If USB and RS-232 use different settings, you could also define it, see here: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#defining-default-connection-settings

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+        super().__init__(adapter, name, **kwargs)
+
+    # === MACH-ZENDER INTERFEROMETER (MZM) ===
```
BenediktBurger: Do you think channels (https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html) would be good to separate the MRM and VOA?

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def calibrate_mzm(self):
+        """Run a MZM bias calibration."""
+        self.ask("MZM:RESET")
+
+    mzm_calibrating = Instrument.measurement(
```
BenediktBurger: maybe "is_calibrating" to be more explicit.

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+    mzm_stable = Instrument.measurement(
+        "MZM:SETPOINT?",
+        """Get whether the MZM is stable and at the setpoint (bool).""",
+        cast=bool,
+    )
+
+    mzm_mode = Instrument.control(
+        "MZM:MODE?",
+        "MZM:MODE: %d",
+        """Control the MZM bias mode (MzmMode enum).""",
```
BenediktBurger: You can create links in the sphinx documentation (on read the docs) with this syntax:
```suggestion
        """Control the MZM bias mode (:class:`MzmMode` enum).""",
```

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+        """Run a MZM bias calibration."""
+        self.ask("MZM:RESET")
+
+    mzm_calibrating = Instrument.measurement(
+        "MZM:CALIBRATING?",
+        """Get whether the MZM is currently being calibrated (bool).""",
+        cast=bool,
+    )
+
+    mzm_stable = Instrument.measurement(
```
BenediktBurger: see above

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+        """Get whether the MZM is stable and at the setpoint (bool).""",
+        cast=bool,
+    )
+
+    mzm_mode = Instrument.control(
+        "MZM:MODE?",
+        "MZM:MODE: %d",
+        """Control the MZM bias mode (MzmMode enum).""",
+        get_process=lambda v: MzmMode(v),
+        set_process=lambda v: v.value,
```
BenediktBurger: Do you need it, as it is an IntEnum?

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+    )
+
+    mzm_ratio_setpoint = Instrument.control(
+        "MZM:HOLD:RATIO?",
+        "MZM:HOLD:RATIO: %d",
+        """Control the input to output power ratio setpoint
+        when `mzm_mode` is set to `MzmMode.AUTOPOWERPOS` or `MzmMode.AUTOPOWERNEG`
+        (float strictly in range 2.5 to 100).""",
```
BenediktBurger, line 92: I guess it is in percent or is it a factor (2 to 100 times)?

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+        "VOA:POWER: %d",
+        """Control whether the VOA is enabled (bool).""",
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    voa_stable = Instrument.measurement(
+        "VOA:SETPOINT?",
+        """Get whether the VOA attenuation is withing 0.1dB of the setpoint (bool).""",
```
BenediktBurger:
```suggestion
        """Get whether the VOA attenuation is within 0.1dB of the setpoint (bool).""",
```

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+        cast=bool,
+    )
+
+    # Handled differently as "VOA:MODE: {0 or 1}" returns "1" on receipt of command
```
BenediktBurger: Doesn't a more elaborate get_process work?

Also please move this comment closer (maybe into the property) to the code.

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
```diff
@@ -0,0 +1,272 @@
...
+
+        t0 = time()
+        h = 0
+        while time() < t0 + duration:
+            r, g, b = colorsys.hsv_to_rgb(h % 600 / 600, 1, 1)
+            self.rgb_red, self.rgb_green, self.rgb_blue = [round(x * 100) for x in (r, g, b)]
+            h += 1
+            sleep(pause)
+
+        self.rgb_power = RgbPowerMode.OFF
```
BenediktBurger: I understand, that it is fun, but does everyone using this instrument have to load this library?

It would be better to place it in the examples folder.

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
fg-lumentum: It's a factor, 2.5 to 100 times

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
fg-lumentum: You can't change them, so I've hardcoded them, with the exception of timeout but I doubt that will be changed often so I've kept it as kwargs.setdefault

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
fg-lumentum: I've removed them :(

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
fg-lumentum: I don't know, it's a value returned on setting which is strange, so I don't think `get_process` will help?

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
BenediktBurger: set_process should be not necessary (you removed it).
However, you might need get_process.
Does it return now an integer (I think so) or the enum?
The tests compare with the IntEnum, so the comparison is true, regardless whether it is an int or enum.

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
BenediktBurger: Ah ok, I see.

pymeasure/instruments/thorlabs/thorlabsmbxseries.py
fg-lumentum: Ah, I see, thank you

PR #1363, opened 2025-10-09
Add Python 3.14 support and drop Python 3.9 support

### LINKED ISSUE #1444: Drop Python 3.9 support
Without Python 3.9 (which is EOL), we can create a coherent testing environment again.

@pymeasure/pymeasure-maintainers I suggest to drop Python 3.9 while adding Python 3.14.

_Originally posted by @BenediktBurger in https://github.com/pymeasure/pymeasure/issues/1363#issuecomment-4413095670_
            

Here we can collect changes for dropped python 3.9 support.

In the codebase, only one instrument contains annotations that global methods could become static now.

BenediktBurger:


BenediktBurger:
@coderabbitai review

BenediktBurger:
Without Python 3.9 (which is EOL), we can create a coherent testing environment again.

@pymeasure/pymeasure-maintainers I suggest to drop Python 3.9 while adding Python 3.14.

CasperSchippers:
That sounds good to me indeed; 3.9 has reached EOL already.

bilderbuchi:
> That sounds good to me indeed; 3.9 has reached EOL already.

same for me!

pyproject.toml
```diff
@@ -84,7 +84,7 @@ profile = "black"
 line-length = 100
 
 [tool.ruff.lint]
-extend-select = ["E", "E9", "F63", "F7", "F82", "W",]
+extend-select = ["E", "E9", "F63", "F7", "F82", "W", "UP", "FURB"]
```
BenediktBurger, line 89: I tested it locally: These upgrade rules always respect the current minimum Python version. If you increase requires-python, it will surface new issues.

Therefore, we should keep it in and at every python upgrade we will get suggestions to improve the codebase.

BenediktBurger:
@coderabbitai review

PR #1370, opened 2025-10-24
Scpi keyword

fg-lumentum:
Created a new datatype, `SCPIKeyword`, to handle the shortform and longform versions of SCPI keywords, and created a new enumerator class, `SCPIKeywordEnum`, to handle enumeration of SCPI keywords.

`SCPIKeyword` inherits from string. It is created using the standard upper-lowercase formatting in SCPI conformant documentation:

    >>> k = SCPIKeyword("PYMeasure")

from which it derives the shortform:

    >>> k.shortform
    "PYM"

The rules for an argument to `SCPIKeyword` are:
- Must contain only alphanumeric characters.
- Uppercase letters define the shortform abbreviation.
- Must start with at least one uppercase letter to define the shortform.
- Must follow the pattern `<uppercase><lowercase><digits>`.

`SCPIKeyword` implements case insensitive comparisons to both the short and longform versions, equivalent to how SCPI handles them:

    >>> k == "PyMeasure"
    True
    >>> k == "PyM"
    True
    >>> k == "PyMeas"
    False

`SCPIKeyword` is inherited by `SCPIKeywordEnum`. `SCPIKeywordEnum` also incorporates a `_missing_` function to allow for lookup using the short or case-insensitive SCPI keyword string, e.g.:

    >>> class PyMeasureEnum(SCPIKeywordEnum):
    ... PYMEASURE = "PYMeasure"
    >>> PyMeasureEnum("Pym")
    ... <PyMeasureEnum.PYMEASURE: SCPIKeyword('PYMeasure')>

The motivation for this is to implement enumerators while not forcing the user into using them, for example:

    class SourceMode(SCPIKeywordEnum):
        CURRENT = "CURRent"
        VOLTAGE = "VOLTage"
    
    class SMUInstrument(Instrument):
    
        source_mode = Instrument.control(
            "SOURce:MODE?",
            "SOURce:MODE %s",
            "",
            values=SourceMode,
            get_process=lambda s: SourceMode(s),
        )

This then allows the user to set the source mode with a string matching the SCPI command, a `SCPIKeyword`, or the enumerator, e.g. all the following would be valid:

    >>> SMUInstrument.source_mode = SourceMode.CURRENT
    >>> SMUInstrument.source_mode = "Current"
    >>> SMUInstrument.source_mode = "CURR"
    >>> SMUInstrument.source_mode == SourceMode.CURRENT
    True
    >>> SMUInstrument.source_mode == "Current"
    True
    >>> SMUInstrument.source_mode == "CURR"
    True

This is particularly useful for implementing enumerators while maintaining backwards compatibility for situations where `values` is a set or dict. In this case, this would maintain backwards compatibility for the cases `values = {"CURR", "VOLT"}`, `values = {"current": "CURR", "voltage": "VOLT"}`, etc.

Tests of the new datatypes have also been implemented.

### DIFF
--- a/pymeasure/instruments/scpi_datatypes.py
+++ b/pymeasure/instruments/scpi_datatypes.py
@@ -53,7 +53,9 @@ class SCPIKeyword(str):
     def __new__(cls, value):
         """Validate and construct a new SCPIKeyword instance."""
         if isinstance(value, SCPIKeyword):
-            value = str(value)
+            instance = super().__new__(cls, value)
+            instance.shortform = value.shortform
+            return instance
 
         if not isinstance(value, str):
             raise TypeError(f"Expected a string or SCPIKeyword, got {type(value).__name__}")
@@ -67,7 +69,8 @@ class SCPIKeyword(str):
 
         if len(upper) == 0:
             raise ValueError(
-                f"Invalid SCPIKeyword '{value}': Must start with at least one uppercase character."
+                f"Invalid SCPIKeyword '{value}': "
+                "Must start with at least one uppercase character."
             )
 
         if value != upper + lower + decimal:
@@ -76,9 +79,9 @@ class SCPIKeyword(str):
                 "must be of the form '<uppercase><lowercase><decimal>'."
             )
 
-        self = str.__new__(cls, value)
-        self.shortform = upper + decimal
-        return self
+        instance = super().__new__(cls, value)
+        instance.shortform = upper + decimal
+        return instance
 
     def __eq__(self, other):
         """Compare against a string or SCPIKeyword, matching longform or shortform."""
@@ -87,7 +90,7 @@ class SCPIKeyword(str):
         return NotImplemented
 
     def __repr__(self):
-        return f"SCPIKeyword('{str(self)}')"
+        return f"SCPIKeyword('{self!s}')"
 
 
 class SCPIKeywordEnum(SCPIKeyword, Enum):
--- a/tests/instruments/test_scpi_datatypes.py
+++ b/tests/instruments/test_scpi_datatypes.py
@@ -22,34 +22,38 @@
 # THE SOFTWARE.
 #
 
-from contextlib import nullcontext
 import pytest
 
 from pymeasure.instruments import SCPIKeyword, SCPIKeywordEnum
+from pymeasure.instruments.validators import strict_discrete_set
 
 
 # === SCPIKeyword ===
 
 
 @pytest.mark.parametrize(
-    "input_value, shortform, expected_exception",
+    "input_value, shortform",
     (
-        ("ALPHa", "ALPH", None),
-        ("ALPH", "ALPH", None),
-        ("ALPHa1", "ALPH1", None),
-        (SCPIKeyword("ALPHa"), "ALPH", None),
-        (1234, None, TypeError),
-        ("ALPHa-1", None, ValueError),
-        ("alpha", None, ValueError),
-        ("alPHa", None, ValueError),
-        ("A1PHa", None, ValueError),
+        ("ALPHa", "ALPH"),
+        ("ALPH", "ALPH"),
+        ("ALPHa1", "ALPH1"),
+        (SCPIKeyword("ALPHa"), "ALPH"),
     ),
 )
-def test_scpi_keyword_init(input_value, shortform, expected_exception):
-    ctx = pytest.raises(expected_exception) if expected_exception else nullcontext()
-    with ctx:
-        keyword = SCPIKeyword(input_value)
-        assert keyword.shortform == shortform
+def test_scpi_keyword_init_valid(input_value, shortform):
+    assert SCPIKeyword(input_value).shortform == shortform
+
+
+@pytest.mark.parametrize("input_value", [1234])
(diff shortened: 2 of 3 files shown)


tests/instruments/test_scpi_datatypes.py
```diff
@@ -0,0 +1,111 @@
...
+        ("alpha", None, ValueError),
+        ("alPHa", None, ValueError),
+        ("A1PHa", None, ValueError),
+    ),
+)
+def test_scpi_keyword_init(input_value, shortform, expected_exception):
+    ctx = pytest.raises(expected_exception) if expected_exception else nullcontext()
+    with ctx:
+        keyword = SCPIKeyword(input_value)
+        assert keyword.shortform == shortform
```
BenediktBurger: You could split it into two tests: One with successful values and one with errors

pymeasure/instruments/scpi_datatypes.py
```diff
@@ -0,0 +1,125 @@
...
+        * Must contain only alphanumeric characters.
+        * Uppercase letters define the shortform abbreviation.
+        * Must start with at least one uppercase letter to define the shortform.
+        * Must follow the pattern <uppercase><lowercase><digits>.
+    """
+
+    def __new__(cls, value):
+        """Validate and construct a new SCPIKeyword instance."""
+        if isinstance(value, SCPIKeyword):
+            value = str(value)
```
BenediktBurger: Can't you just return value here?

pymeasure/instruments/scpi_datatypes.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if value != upper + lower + decimal:
+            raise ValueError(
+                f"Invalid SCPIKeyword '{value}': "
+                "must be of the form '<uppercase><lowercase><decimal>'."
+            )
+
+        self = str.__new__(cls, value)
+        self.shortform = upper + decimal
+        return self
```
BenediktBurger: I don't think `self` is appropriate here.
Better `instance` or keyword or something as `self` seems wrong to me for a classmethod.

pymeasure/instruments/scpi_datatypes.py
```diff
@@ -0,0 +1,125 @@
...
+                f"Invalid SCPIKeyword '{value}': Must start with at least one uppercase character."
+            )
+
+        if value != upper + lower + decimal:
+            raise ValueError(
+                f"Invalid SCPIKeyword '{value}': "
+                "must be of the form '<uppercase><lowercase><decimal>'."
+            )
+
+        self = str.__new__(cls, value)
```
BenediktBurger: I don't have a lot of experience with `__new__` methods.
Shouldn't it be better `super().__new__...` like for init methods?

pymeasure/instruments/scpi_datatypes.py
```diff
@@ -0,0 +1,125 @@
...
+        self.shortform = upper + decimal
+        return self
+
+    def __eq__(self, other):
+        """Compare against a string or SCPIKeyword, matching longform or shortform."""
+        if isinstance(other, str):
```
BenediktBurger, line 88: This also works with other keywords, because they are a subclass of str, right?

pymeasure/instruments/scpi_datatypes.py
```diff
@@ -0,0 +1,125 @@
...
+            return other.upper() in [self.upper(), self.shortform]
+        return NotImplemented
+
+    def __repr__(self):
+        return f"SCPIKeyword('{str(self)}')"
```
BenediktBurger: You can use the "[conversion](https://docs.python.org/3/library/string.html#format-string-syntax)" element of a format string (with `!`):

```suggestion
        return f"SCPIKeyword('{self!s}')"
```

tests/instruments/test_scpi_datatypes.py
```diff
@@ -0,0 +1,111 @@
...
+def test_scpi_keyword_eq(input_value, valid):
+    assert (ALPHA_KEYWORD == input_value) is valid
+
+
+# === SCPIKeywordEnum ===
+
+
+class AlphaKeywordEnum(SCPIKeywordEnum):
+    ALPHA = "ALPHa"
+
```
BenediktBurger, line 88: Could you also use a SCPIKeyword as a member?
E.g. `BETA = SCPIKeyword("BEta")`

pymeasure/instruments/scpi_datatypes.py
fg-lumentum: Yeah, that seems more sensible

pymeasure/instruments/scpi_datatypes.py
fg-lumentum: Not quite, it still needs to go through super().__new__(cls, value). I have changed it to::

      if isinstance(value, SCPIKeyword):
          instance = super().__new__(cls, value)
          instance.shortform = value.shortform
          return instance

to ensure it doesn't go through the validation logic multiple times

fg-lumentum:
Converted to a draft as, on further reading, I need to ensure equalities are commutative (a==b if and only if b==a), and I will likely need to be careful with __hash__. This isn't possible with the current implementation.

PR #1371, opened 2025-10-27
Add Voltcraft LCR500 instrument

cckssr:
The Voltcraft LCR500 supports SCPI and has documentation [e.g., from Conrad Electronics](https://www.conrad.de/de/p/voltcraft-lcr-500-lcr-messbruecke-digital-datenlogger-cat-i-anzeige-counts-40000-3088970.html?refresh=true). 

I implemented all SCPI commands and tested them with a real instrument.


docs/api/instruments/voltcraft/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.voltcraft
+
+####################
```
BenediktBurger: Please make the hashes as long as the text.

Also applies to the next file.

pymeasure/instruments/voltcraft/__init__.py
```diff
@@ -22,4 +22,4 @@
 # THE SOFTWARE.
 #
 
-from .spellmanXRV import SpellmanXRV
+from .voltcraft_lcr500 import LCR500
```
BenediktBurger: Files should end with an empty line.

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,162 @@
...
+#
+
+import logging
+from pymeasure.instruments import Instrument, SCPIMixin
+from pymeasure.instruments.validators import (
+    strict_discrete_set,
+)
+
+log = logging.getLogger(__name__)  # https://docs.python.org/3/howto/logging.html#library-config
+log.addHandler(logging.NullHandler())
```
BenediktBurger: If you do not use the logger, you do not need to create `log` at all.

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,162 @@
...
+    """Validator that accepts both strings and numbers in the same list."""
+    if isinstance(value, str):
+        if value.upper() not in [v for v in values if isinstance(v, str)]:
+            raise ValueError
+    elif isinstance(value, (int, float)):
+        if value not in [v for v in values if isinstance(v, (int, float))]:
+            raise ValueError
+    else:
+        raise ValueError
+    return value
```
BenediktBurger: Doesn't do `strict_discrete_set` validator ensure just that: Whether the value is in the list of values, regardless, whether the is a string or a number?

Or is it just for the case insensitivity?
In the latter case, I'd say it is sufficient to allow upper case only.

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class LCR500(SCPIMixin, Instrument):
+    def __init__(self, adapter, name="Voltcraft LCR-500", **kwargs):
+        super().__init__(adapter, name, **kwargs)
+        self.adapter.connection.write_termination = "\n"
+        self.adapter.connection.read_termination = "\n"
```
BenediktBurger: It is better to give the configuration to the super call. That will exactly do, what you want it to do, unless I give a (already preconfigured) adapter:

```suggestion
        super().__init__(adapter, name, write_termination="\n", read_termination="\n", **kwargs)
```

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,162 @@
...
+    level = Instrument.control(
+        "FUNC:LEV?",
+        "FUNC:LEV %d",
+        """Control the measurement signal level in mVrms (int).
+        Can be one of: 300, 600.
+        """,
+        validator=strict_discrete_set,
+        values=_LEVELS_MVRMS,
+    )
+    equivalent_circuit = Instrument.control(
```
BenediktBurger:
```suggestion
    )

    equivalent_circuit = Instrument.control(
```

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,162 @@
...
+        "FUNC:EQUI?",
+        "FUNC:EQUI %s",
+        """Control the equivalent circuit type (str).
+        Can be one of: 'SER', 'PAL'.
+        """,
+        validator=strict_discrete_set,
+        values=_EQUIV,
+    )
```
BenediktBurger: You could make it a boolean property, e.g. `equivalent_circuit_parallel_enabled` ant then use map_values to map True to PAL and False to SER.

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # -- Fetch Subsystem --
+    fetch = Instrument.measurement(
+        "FETC?",
+        """Perform a measurement and return the result as tuple of
+        (primary_value, secondary_value, range_used).""",
+        parser=_parse_fetch,
```
BenediktBurger: Does this work?
I did not know that we have a `parser` argument!

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,162 @@
...
+        Can be one of: 'SER', 'PAL'.
+        """,
+        validator=strict_discrete_set,
+        values=_EQUIV,
+    )
+
+    # -- Fetch Subsystem --
+    fetch = Instrument.measurement(
+        "FETC?",
+        """Perform a measurement and return the result as tuple of
```
BenediktBurger: Docstrings should start with Get/Measure for a `measurement` and Set for `setting` and Control for `Control`.

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,162 @@
...
+        self.adapter.connection.read_termination = "\n"
+
+    def _parse_fetch(self, value):
+        """Parse the fetch response into a tuple of floats.
+        The response is expected to be in the format: 'primary,secondary,range'
+        """
+        parts = value.split(",")
+        if len(parts) != 3:
+            raise ValueError("Invalid fetch response: %s" % value)
+        return float(parts[0]), float(parts[1]), int(parts[2])
```
BenediktBurger: That is superfluous, the `values` method (called by the property creators) offers already the functionality to split the response and cast each part to something you want to (default float).

BenediktBurger (commented):

Would you mind adding at least a few protocol tests, please? Here you can look at them: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#protocol-tests

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
cckssr: Done

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
cckssr: Done

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
cckssr: Done

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
cckssr: Not my favourite method, but done.

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
cckssr: Done

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
cckssr: The super call doesn't work all the time. If I am initialising with a VISAAdapter instance instead of the resource_name, the correct termination is not written. Thus, keeping it in.

BenediktBurger: Yes, it is known that nothing of the configuration enters a manually given adapter.
If you hand over any adapter instead of a resource name, you have to configure it manually.

docs/api/instruments/voltcraft/voltcraft_lcr500.rst
```diff
@@ -0,0 +1,11 @@
+##########################
+Voltcraft LCR500 LCR-Meter
+##########################
+
+.. warning::
+
+    This instrument uses different write and read terminators. The write and read terminators are each ``\n``. A mismatch can lead to communication issues when sending commands, as responses may not be properly terminated or recognized. If a custom adapter is used, ensure that these are set correctly.
```
BenediktBurger, line 7: Put this information into the docstring. That way it is visible both in the code as well as here (the API docs show the docstring as well).

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,201 @@
...
+    frequency = Instrument.control(
+        "FREQ?",
+        "FREQ %d",
+        """Control the measurement frequency in Hz.
+
+        :type: int
+
+        Allowed values: 100, 120, 400, 1000, 4000, 10000, 40000, 50000,
+        75000, 100000.
+        """,
```
BenediktBurger, line 107: Typically, we write docstrings like this, see https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings:

```suggestion
        """Control the measurement frequency in Hz (int strictly in 100, 120, 400, 1000, 4000, 10000, 40000, 50000, 75000, 100000).
        """,
```

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
```diff
@@ -0,0 +1,201 @@
...
+        :type: str
+
+        Allowed values:
+
+        * ``'r'``: Resistance
+        * ``'l'``: Inductance
+        * ``'c'``: Capacitance
+        * ``'z'``: Impedance
+        * ``'auto'``: Automatic
+        """,
```
BenediktBurger, line 127: For such an enumeration of values, you can keep a longer docstring.

pymeasure/instruments/voltcraft/voltcraft_lcr500.pyi
BenediktBurger, line 1: We don't use such an interface file. Instead, you can add type hints to the main class.

tests/instruments/voltcraft/test_voltcraft_lcr500.py
BenediktBurger, line 1: Please add copyright notes to the test files as well.

BenediktBurger (approved):
Looks, good. Only minor comments.

Sorry for the long waiting time.

pymeasure/instruments/voltcraft/voltcraft_lcr500.py
cckssr: The suggested
```
"""Control the measurement frequency in Hz (int strictly in 100, 120, 400, 1000, 4000, 10000, 40000, 50000, 75000, 100000).
"""
```
 docstring contradicts the coding standards. 

- One short summary line in imperative voice, with a period at the end.
Values, types and ranges are not summary.

Also, the line is over 80 characters and thus neither short nor compliant with the linter.

pymeasure/instruments/voltcraft/voltcraft_lcr500.pyi
cckssr: Typing the class variables isn’t straightforward while avoiding Pylance exceptions. I needed to use casting to get the correct types. Since I think inline typing helps users immensely, in addition to the docstring, this is the only approach I could think of.

docs/api/instruments/voltcraft/voltcraft_lcr500.rst
cckssr: Done

PR #1375, opened 2025-11-04
Extend parameter test

Falk3579:
Add test to BooleanParameter to check whether it can handle data type numpy.bool.

### DIFF
--- a/tests/experiment/test_parameters.py
+++ b/tests/experiment/test_parameters.py
@@ -79,7 +79,7 @@ def test_integer_bounds():
         p.value = -100  # below minimum
 
 
-def test_boolean_value():
+def test_boolean_value_error():
     p = BooleanParameter('Test')
     with pytest.raises(ValueError):
         _ = p.value  # not set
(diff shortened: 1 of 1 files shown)


tests/experiment/test_parameters.py
```diff
@@ -100,6 +101,10 @@ def test_boolean_value():
     assert p.value is False
     p.value = True
     assert p.value is True
+    p.value = np.bool(True)  # True as numpy boolean
+    assert bool(p.value) is True
```
BenediktBurger: Why do you need `bool(p.value)` while all the other tests do not?

tests/experiment/test_parameters.py
```diff
@@ -100,6 +101,10 @@ def test_boolean_value():
     assert p.value is False
     p.value = True
     assert p.value is True
+    p.value = np.bool(True)  # True as numpy boolean
```
BenediktBurger: I think the comment is superfluous.

tests/experiment/test_parameters.py
```diff
@@ -100,6 +101,10 @@ def test_boolean_value():
...
+    assert bool(p.value) is True
+    p.value = np.bool(False)  # False as numpy boolean
+    assert bool(p.value) is False
```
BenediktBurger: It is quite repetitive to write `p.value = xyz; assert p.value is True`.
Therefore, I think, it is good to parametrize the test: Have a generic one for the first and last test, another one for True and one for False.

Do you know how it works?

BenediktBurger (commented):
Thanks for writing tests. It would be good to have the tests with the PR itself.

tests/experiment/test_parameters.py
Falk3579: Removed it

tests/experiment/test_parameters.py
Falk3579: The 'is' operator compares whether objects are from the same class. So it had to be converted first.
In newer version I use '=='.

tests/experiment/test_parameters.py
Falk3579: The test is now parametrized. I merged the 'True' and 'False' test into one test.

BenediktBurger (approved):
Thanks for adjusting the test.
The parametrization makes it better, as it clearly indicates which value breaks the implementation.

PR #1376, opened 2025-11-04
Add vector parameter input

Falk3579:
Add class 'VectorInput' to handle 'VectorParameter' in the inputs of the GUI.

Previously, VectorParameter was handled as StringInput. This raised an error since VectorParameter returns a list.


CasperSchippers (approved):
Also here, nothing to remark; thanks for this addition!!!

PR #1377, opened 2025-11-04
Vector parameter allow numpy array

Falk3579:
- Allow numpy.ndarray as data type for VectorParameter. This enables the usage of vector parameter in the sequencer.
- Add corresponding test.
- Parameterize the test of VectorParameter.

### DIFF
--- a/tests/experiment/test_parameters.py
+++ b/tests/experiment/test_parameters.py
@@ -192,14 +192,8 @@ def test_list_order():
                              ('choices are', 'choices')]
 
 
-def test_vector():
+def test_vector_error():
     p = VectorParameter('test', length=3, units='tests')
-    p.value = [1, 2, 3]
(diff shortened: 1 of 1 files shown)


tests/experiment/test_parameters.py
```diff
@@ -212,4 +206,16 @@ def test_vector():
     assert p.cli_args[0] is None
     assert p.cli_args[1] == [('units are', 'units'), 'default', '_length']
 
+
+@pytest.mark.parametrize("value, mapping", (
```
CasperSchippers, line 210: Nice, that is a lot neater; thanks

CasperSchippers (approved):
Thanks for this addition; good idea indeed, and nothing to remark!

(except for the thanks for the parametrized test, but cannot have an open "issue" before merging)

PR #1380, opened 2025-11-11
Update documentation of Agilent B1500

ilev-sergey:
Updates documentation of Agilent B1500

- fixes docstring style to correspond to Pymeasure style
- removes the B1500 from the list of "grandfathered docstring instruments"
- adds missing return type for functions
- adds sphinx links for some of the methods

### DIFF
--- a/pymeasure/instruments/agilent/agilentB1500.py
+++ b/pymeasure/instruments/agilent/agilentB1500.py
@@ -71,7 +71,6 @@ class AgilentB1500(SCPIMixin, Instrument):
 
     def query_learn(self, query_type):
         """Query settings from the instrument. (``*LRN?``)
-        Return dict of settings.
 
         :param int or str query_type: Query type (number according to manual)
         :rtype: dict


pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -61,30 +61,38 @@ def __init__(self, adapter, name="Agilent B1500 Semiconductor Parameter Analyzer
 
     @property
     def smu_references(self):
-        """Return all SMU instances."""
+        """Get all SMU instances.
+
+        :rtype: dict_values
+        """
         return self._smu_references.values()
```
BenediktBurger, line 65: Maybe combine it: `"Get all SMU instances as dict."`

Actually, we now allow type hints, so you could also use these:

```suggestion
def smu_references(self) -> dict:
        """Get all SMU instances."""
return self._smu_references.values()
```

If you know the types in the dict, you can also return them, e.g. `dict[str, str]`.
If you have more questions regarding type hints, don't hesitate to ask.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -61,30 +61,38 @@ def __init__(self, adapter, name="Agilent B1500 Semiconductor Parameter Analyzer
...
+        """Get all SMU names.
+
+        :rtype: dict
+        """
         return self._smu_names
 
     def query_learn(self, query_type):
-        """Query settings from the instrument (``*LRN?``).
+        """Query settings from the instrument. (``*LRN?``)
         Return dict of settings.
```
BenediktBurger: While you touch it:

Docstrings should have a single summary line, then (if applicable) an empty line ant then more context, see https://peps.python.org/pep-0257/.
So this "Return dict of settings." should be after an empty line.

However, here, it could be removed as it is redundant.
Please take it into consideration for the other methods.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -164,21 +172,21 @@ def initialize_all_spgus(self):
...
         """Abort the present operation but channels may still output
-        current/voltage (``AB``)
+        current/voltage. (``AB``)
         """
         self.write("AB")
 
     def force_gnd(self):
-        """Force 0V on all channels immediately. Current Settings can
-        be restored with RZ. (``DZ``)
+        """Force 0V on all channels immediately. Current settings can
```
BenediktBurger: Actually, according to SI, there should be some space between magnitude and units: 0 V.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -975,20 +974,21 @@ def _query_status_raw(self):
 
     @property
     def status(self):
-        """Query status of the SMU."""
+        """Control status of the SMU."""
```
BenediktBurger: Control indicates, that you can get **and** set the value, here the status. Here, we only get the status -> Get status.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -975,20 +974,21 @@ def _query_status_raw(self):
...
 
     def disable(self):
-        """Disable Source/Measurement Channel (``CL``)"""
+        """Disable Source/Measurement Channel. (``CL``)"""
         self.write(f"CL {self.channel}")
 
     def force_gnd(self):
-        """Force 0V immediately. Current Settings can be restored with
-        ``RZ``. (``DZ``)"""
+        """Force 0V immediately. Current settings can be restored with
```
BenediktBurger: Force output to 0 V immediately.

"Force 0 V" does not seem right to me.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1123,14 +1117,13 @@ def ramp_source(
...
+        :param float target_output: Target output voltage or current
         :param int irange: Output range index
         :param float, optional comp: Compliance, defaults to previous setting
-        :param CompliancePolarity comp_polarity: Compliance polairty, defaults to auto
+        :param CompliancePolarity, optional comp_polarity: Compliance polairty, defaults to auto
         :param int or str, optional comp_range: Compliance ranging type, defaults to auto
-        :param stepsize: Maximum size of steps
-        :param pause: Duration in seconds to wait between steps
+        :param float, optional stepsize: Maximum size of steps
+        :param float, optional pause: Duration in seconds to wait between steps
```
BenediktBurger: I don't think that you need to state that it is optional, as you can see it in the signature (by having a default value).
The signature should be the single point of truth.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1409,8 +1402,8 @@ def __call__(self, input_value):
         """Give named tuple (name/index) of given Range.
         Throws error if range is not supported by this SMU.
 
-        :param str or int input: Range name or index
-        :return: named tuple (name/index) of range
+        :param str or int input_value: Range name or index
```
BenediktBurger, line 1399: Good catch!

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: It's actually `dict_values`, not `dict` itself. Anyway, I updated the description with type.

As for type hints, thanks for the suggestion, but I'll probably implement them for the whole instrument later, once I'm done with other changes. I don't want to add inconsistencies in the code now.

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: > The signature should be the single point of truth.

By the way, there is a [sphinx extension](https://github.com/tox-dev/sphinx-autodoc-typehints) that allows to use type hints and default values from the signature to render parameters in the documentation (see the README for an example). Since type hints are now allowed, maybe it's worth to look into this extension as a docs dependency. This would let to use the docstring for actual description while keeping types and defaults in the signature.

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: You're right, its just the values.

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: I created a new ticket with your suggestion (thank you) under #1392 .

PR #1381, opened 2025-11-11
ENH add gwinstek GPD-230S power supply

Helikopterconni:
Support for the two channel GPD-230S power supply by GWInstek. Rather old power supply with only USB support.
[manual](https://www.gwinstek.com/en-global/products/downloadSeriesDownNew/7837/1474)

### DIFF
--- a/pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
+++ b/pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
@@ -27,14 +27,18 @@ from pymeasure.instruments import Instrument, Channel
 from pymeasure.instruments.validators import strict_discrete_set, truncated_range
 
 
-import logging
-
-log = logging.getLogger(__name__)
-log.addHandler(logging.NullHandler())
-
 _VOLTAGE_RANGE = [0.0, 30]
 _CURRENT_RANGE = [0.0, 3]
-_TRACK_MODES = [0, 1, 2]  # 0=Independent, 1=Series, 2=Parallel
+
+class TRACK_MODE(enum.Enum):
+    INDEPENDENT = 0
+    SERIES = 1
+    PARALLEL = 2
+
+class BAUD_RATE(enum.Enum):
+    BAUD_115200 = 0
+    BAUD_57600 = 1
+    BAUD_9600 = 2
 
 
 class GWInstekGPD2303SStatusRegister(enum.Enum):
@@ -53,30 +57,32 @@ class GWInstekGPD230SChannel(Channel):
     current_limit = Channel.control(
         "ISET{ch}?",
         "ISET{ch}:%g",
-        """Control maximum current of the power supply.""",
+        """Control maximum current [A] of the power supply.""",
         get_process=lambda v: float(v.replace("A", "")),
         set_process=lambda v: round(v, 3),
         validator=truncated_range,
         values=_CURRENT_RANGE,
     )
+
     voltage_limit = Channel.control(
         "VSET{ch}?",
         "VSET{ch}:%g",
-        """Control maximum voltage of the power supply.""",
+        """Control maximum voltage [V] of the power supply.""",
         get_process=lambda v: float(v.replace("V", "")),
         set_process=lambda v: round(v, 3),
         validator=truncated_range,
         values=_VOLTAGE_RANGE,
     )
+
     current = Channel.measurement(
         "IOUT{ch}?",
-        """Get the actual output current.""",
+        """Get the actual output current [A].""",
         get_process=lambda v: float(v.replace("A", "")),
     )
 
     voltage = Channel.measurement(
         "VOUT{ch}?",
-        """Get the actual output voltage.""",
+        """Get the actual output voltage [V].""",
         get_process=lambda v: float(v.replace("V", "")),
     )
 
@@ -85,47 +91,14 @@ class GWInstekGPD230S(Instrument):
     """
     Represents the GW Instek two channel Power Supply (minimal implementation)
     and provides a high-level interface for interacting with the instrument.
-
-    The device only has a serial USB interface that is ususally set at
-    a baud rate of 9600. \n
-
-    the GPD-230s series only has two channels.
-
-    Supported commands (as per the manual):
-
-      - VSET<CH>:<volts>
-      - VSET<CH>?
-      - ISET<CH>:<amps>
-      - ISET<CH>?
-
-      - VOUT<CH>?
-      - IOUT<CH>?
-      - TRACK<NR1> (0=Independent, 1=Series, 2=Parallel)
-
-      - OUT<BOOLEAN> (0=Off, 1=On)
-      - LOCAL / REMOTE
-
-      - BEEP<BOOLEAN>
-      - SAV<NR1> / RCL<NR1> (memory save/recall)
-
-      - BAUD<NR1> (0:115200, 1:57600, 2:9600)
-      - *IDN? / STATUS? / ERR? / HELP?
-
-      - SAV/RCL do not affect output state; OUT may be turned off on save/recall as safety
-
-    .. code-block:: python
-
-    ps.output_enabled = 1
-    ps.channel_1.voltage_limit=5
-    ps.channel_1.current_limit=1
-    print(ps.error)
-    print(ps.status)
-    print(ps.channel_1.voltage)
-    print(ps.channel_1.current)\n
+    The device only has a serial USB interface that is usually set at
+    a baud rate of 9600. The GPD-230s series only has two channels.
 
     Known issue:
-    - When the instrument throws an error, it is best to flush the read buffer
-    `ps.adapter.flush_read_buffer()`
+    - When the instrument throws an error, it is best to flush the read buffer with
+    `ps.adapter.flush_read_buffer()`.
+    -  The`REMOTE` command is not implemented, as it does not have any effect. It
+    does not work, when the instrument is already in local mode.
     """
 
     channel_1 = Instrument.ChannelCreator(GWInstekGPD230SChannel, 1)
@@ -135,14 +108,13 @@ class GWInstekGPD230S(Instrument):
         self,
         adapter,
         name="GW Instek GPD-2303S DC Power Supply",
-        read_termination="\n",
         baud_rate=9600,
         **kwargs,
     ):
         super().__init__(
             adapter,
             name,
-            read_termination=read_termination,
+            read_termination="\n",
             baud_rate=baud_rate,
             includeSCPI=False,
             **kwargs,
@@ -162,8 +134,8 @@ class GWInstekGPD230S(Instrument):
             flag = register.value["flag"]
             states = register.value["states"]
 
-            trailing_zeors = (flag & -flag).bit_length() - 1
-            state = (flag & status_byte) >> trailing_zeors
+            trailing_zeros = (flag & -flag).bit_length() - 1
+            state = (flag & status_byte) >> trailing_zeros
 
             return_dict[name] = states.get(state, "Error")
 
@@ -175,29 +147,26 @@ class GWInstekGPD230S(Instrument):
         status_byte = int(self.ask("STATUS?").strip(), 2)
         return self.decode_status_byte(status_byte)
 
-    # remote = Instrument.setting(
-    #     "REMOTE",
-    #     """ Exits local mode and sets the instrument to remote mode""",
-    # )
-
     tracking_mode = Instrument.setting(
         "TRACK%d",
         """Set the operation mode: independent, tracking series, or tracking parallel.
         NR1 0: Independent, 1: Series, 2: Parallel""",
         validator=strict_discrete_set,
-        values=_TRACK_MODES,
+        values=TRACK_MODE,
+        map_values=True,
     )
 
     beep_enabled = Instrument.setting(
         "BEEP%d",
         """Set beep on or off.""",
         validator=strict_discrete_set,
-        values=_TRACK_MODES,
+        values={True: 1, False: 0},
+        map_values=True,
     )
 
     output_enabled = Instrument.setting(
         "OUT%d",
-        """Control whether the source is enabled, takes values True or False. (bool)""",
+        """Control whether the source is enabled. (bool)""",
         validator=strict_discrete_set,
         values={True: 1, False: 0},
         map_values=True,
@@ -214,13 +183,8 @@ class GWInstekGPD230S(Instrument):
         "BAUD%d",
         """Set the baud rate to 2:9600, 1:57600, 0:115200""",
         validator=strict_discrete_set,
-        values=[0, 1, 2],
-    )
-
-    local = Instrument.setting(
-        "LOCAL",
-        """Set to exit remote mode and sets the instrument ot local mode.
-        Beware, remote operation is not possible afterwards.""",
+        values=BAUD_RATE,
+        map_values=True,
     )
 
     error = Instrument.measurement(
@@ -228,3 +192,8 @@ class GWInstekGPD230S(Instrument):
     )
 
     help = Instrument.measurement("HELP?", """Get the command list.""")
+
+    def disable_remote(self):
+        """Set to exit remote mode and sets the instrument ot local mode.
+        Beware, remote operation is not possible afterwards."""
+        self.write("LOCAL")
--- a/tests/instruments/gwinstek/test_gwinstekGPD2303S.py
+++ b/tests/instruments/gwinstek/test_gwinstekGPD2303S.py
@@ -1,3 +1,26 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2025 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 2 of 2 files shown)


pymeasure/instruments/gwinstek/__init__.py
BenediktBurger, line 1: Please also add the API documentation, as written here: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+import enum
+
+from pymeasure.instruments import Instrument, Channel
+from pymeasure.instruments.validators import strict_discrete_set, truncated_range
+
+
+import logging
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
```
BenediktBurger: If you do not log anything, you do not need to import it and create a logger.

tests/instruments/gwinstek/test_gwinstekGPD2303S.py
BenediktBurger, line 1: Please add the copyright notice at the top of the file.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+    BEEP = {"flag": 0b00001000, "states": {0b0: "Off", 0b1: "On"}}
+    OUTPUT = {"flag": 0b00000100, "states": {0b0: "Off", 0b1: "On"}}
+    BAUD = {"flag": 0b00000011, "states": {0b00: 115200, 0b01: 57600, 0b10: 9600}}
+
+
+class GWInstekGPD230SChannel(Channel):
+    current_limit = Channel.control(
+        "ISET{ch}?",
+        "ISET{ch}:%g",
+        """Control maximum current of the power supply.""",
```
BenediktBurger: It is good to add the units (I guess Ampere) to the docstring to be complete.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+        get_process=lambda v: float(v.replace("A", "")),
+        set_process=lambda v: round(v, 3),
+        validator=truncated_range,
+        values=_CURRENT_RANGE,
+    )
+    voltage_limit = Channel.control(
```
BenediktBurger, line 67: Please have a new line between two properties.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+    voltage = Channel.measurement(
+        "VOUT{ch}?",
+        """Get the actual output voltage.""",
+        get_process=lambda v: float(v.replace("V", "")),
+    )
+
+
+class GWInstekGPD230S(Instrument):
+    """
+    Represents the GW Instek two channel Power Supply (minimal implementation)
```
BenediktBurger, line 92: do we need this "minimal implementation" in the docstring?

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+    and provides a high-level interface for interacting with the instrument.
+
+    The device only has a serial USB interface that is ususally set at
+    a baud rate of 9600. \n
```
BenediktBurger: Why this new line?

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+      - OUT<BOOLEAN> (0=Off, 1=On)
+      - LOCAL / REMOTE
+
+      - BEEP<BOOLEAN>
+      - SAV<NR1> / RCL<NR1> (memory save/recall)
+
+      - BAUD<NR1> (0:115200, 1:57600, 2:9600)
+      - *IDN? / STATUS? / ERR? / HELP?
+
+      - SAV/RCL do not affect output state; OUT may be turned off on save/recall as safety
```
BenediktBurger: This does not help anybody using this device driver, does it?

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+
+    .. code-block:: python
+
+    ps.output_enabled = 1
+    ps.channel_1.voltage_limit=5
+    ps.channel_1.current_limit=1
+    print(ps.error)
+    print(ps.status)
+    print(ps.channel_1.voltage)
+    print(ps.channel_1.current)\n
```
BenediktBurger: I guess you have to indent it, such that it works properly. Did you try it?

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+
+    channel_1 = Instrument.ChannelCreator(GWInstekGPD230SChannel, 1)
+    channel_2 = Instrument.ChannelCreator(GWInstekGPD230SChannel, 2)
+
+    def __init__(
+        self,
+        adapter,
+        name="GW Instek GPD-2303S DC Power Supply",
+        read_termination="\n",
+        baud_rate=9600,
```
BenediktBurger: Can you change these values (at least the baud rate seems to).
If not, you should hard code them into the super call.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+        **kwargs,
+    ):
+        super().__init__(
+            adapter,
+            name,
+            read_termination=read_termination,
+            baud_rate=baud_rate,
+            includeSCPI=False,
```
BenediktBurger, line 119: > I am not sure how to disable the standard scpi commands that are included per default, even though it should be disabled, when SCPIMixin is not inherited to GWInstekGPD230S(). At least, that is how i understand it. I would appreciate advice.

You did it correctly. With `includeSCPI=False` you disable SCPI support in this device.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+        return_dict["raw"] = bin(status_byte)
+        for register in GWInstekGPD2303SStatusRegister:
+            name = register.name
+            flag = register.value["flag"]
+            states = register.value["states"]
+
+            trailing_zeors = (flag & -flag).bit_length() - 1
+            state = (flag & status_byte) >> trailing_zeors
+
+            return_dict[name] = states.get(state, "Error")
```
BenediktBurger, line 140: Did you look into `enum.IntFlag`?
It might make it way easier, just give it the int and you get an Enum with all the values enabled, which are in the byte.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+    @property
+    def status(self) -> dict:
+        """Get the status byte and and decode it."""
+        status_byte = int(self.ask("STATUS?").strip(), 2)
+        return self.decode_status_byte(status_byte)
+
+    # remote = Instrument.setting(
+    #     "REMOTE",
+    #     """ Exits local mode and sets the instrument to remote mode""",
+    # )
```
BenediktBurger: What do you want to do with this comment? If not needed, please remove it.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+
+    tracking_mode = Instrument.setting(
+        "TRACK%d",
+        """Set the operation mode: independent, tracking series, or tracking parallel.
+        NR1 0: Independent, 1: Series, 2: Parallel""",
+        validator=strict_discrete_set,
+        values=_TRACK_MODES,
```
BenediktBurger: Consider using an enum or strings and then use a dict `{"independent": 0, ...}` and set `map_values=True`. This will map the strings to numbers and send the numbers to the device.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+    beep_enabled = Instrument.setting(
+        "BEEP%d",
+        """Set beep on or off.""",
+        validator=strict_discrete_set,
+        values=_TRACK_MODES,
+    )
+
+    output_enabled = Instrument.setting(
+        "OUT%d",
+        """Control whether the source is enabled, takes values True or False. (bool)""",
```
BenediktBurger: "takes values True or False. (bool)" is redundant, the hint at bool is sufficient (also output_enabled calls already for a boolean).

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+        """Set to recall a panel setting""",
+        validator=strict_discrete_set,
+        values=[1, 2, 3, 4],
+    )
+
+    baude_rate = Instrument.setting(
+        "BAUD%d",
+        """Set the baud rate to 2:9600, 1:57600, 0:115200""",
+        validator=strict_discrete_set,
+        values=[0, 1, 2],
```
BenediktBurger: Here again a mapping or an enum would be good.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+    )
+
+    local = Instrument.setting(
+        "LOCAL",
+        """Set to exit remote mode and sets the instrument ot local mode.
+        Beware, remote operation is not possible afterwards.""",
+    )
```
BenediktBurger: In this case, a method is better, e.g `enable_local` or better `disable_remote` as you do an action (enabling local and disabling remote) and do not set some value.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,230 @@
...
+
+    error = Instrument.measurement(
+        "ERR?", """Get the error status of the instrument and returns the last error message."""
```
BenediktBurger, line 191:
```suggestion
        "ERR?", """Get the error status of the instrument and return the last error message."""
```

Take the imperative mood to the whole docstring.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
Helikopterconni: Baud rate can be changed. The read termination seems to be fixed.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
Helikopterconni: Thank you :)

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
Helikopterconni: Yes, i did look into `enum.IntFlag`, but this status register is a bit special, as some of the bits need to be interpreted pairwise. This is the easiest solution i came up with.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
Helikopterconni: This is a bit confusing. The `LOCAL` command sends the instrument into the local state. It is not reachable afterwards. The `REMOTE` command, which should set the instrument back to the remote state does not work in this case, which makes the command useless even though it exists.
How should a useless SCPI command be handled?
At least i think it is useless.

Helikopterconni: i added it to the known issues.

docs/api/instruments/gwinstek/gpd2303s.rst
```diff
@@ -0,0 +1,14 @@
+#################################################################
```
BenediktBurger, line 1: Please make these hashes as long as the text below.

docs/api/instruments/gwinstek/gpd2303s.rst
BenediktBurger, line 1: If you add a new instrument, you have to create a manufacturer index.rst file (see other manufacturers). It is similar to the init.py you created.

pymeasure/instruments/gwinstek/__init__.py
```diff
@@ -0,0 +1,25 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2025 PyMeasure Developers
```
BenediktBurger, line 4: Now we are in 2026 (also the files below).

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
BenediktBurger: I see.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
BenediktBurger: You did not change returns to return yet.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,199 @@
...
+    )
+
+    help = Instrument.measurement("HELP?", """Get the command list.""")
+
+    def disable_remote(self):
+        """Set to exit remote mode and sets the instrument ot local mode.
```
BenediktBurger, line 197: Here also, the whole sentence should be imperative:

```suggestion
        """Exit remote mode and set the instrument to local mode.
```

It is a method so describing an "action" with an action verb (here exit).

Set/Get/Control apply only to properties.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,199 @@
...
+        Beware, remote operation is not possible afterwards."""
```
BenediktBurger, line 198: According to PEP 257, there should be an empty line between the first summary line of a docstring and the rest:

```suggestion

        Beware, remote operation is not possible afterwards."""
```

Maybe more clear: "Beware, it is not possible to return to remote operation afterwards."

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,199 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2025 PyMeasure Developers
```
BenediktBurger, line 4: We're now in 2026.

tests/instruments/gwinstek/test_gwinstekGPD2303S.py
```diff
@@ -0,0 +1,233 @@
...
+    ):
+        pass  # Verify the expected communication.
+
+
+@pytest.mark.parametrize(
+    "comm_pairs, value",
+    (
+        ([(b"IOUT1?", b"0.000A\r\n")], 0.0),
+        ([(b"IOUT1?", b"0.000A\r\n")], 0.0),
+        ([(b"IOUT1?", b"0.000A\r\n")], 0.0),
```
BenediktBurger, line 43: Why three times the same values?
Please change to a single value as three equal tests do not offer any benefit but slow down the pipeline.

In this case you do not need to parametrize the test at all.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,199 @@
...
+    BEEP = {"flag": 0b00001000, "states": {0b0: "Off", 0b1: "On"}}
+    OUTPUT = {"flag": 0b00000100, "states": {0b0: "Off", 0b1: "On"}}
+    BAUD = {"flag": 0b00000011, "states": {0b00: 115200, 0b01: 57600, 0b10: 9600}}
+
+
+class GWInstekGPD230SChannel(Channel):
+    current_limit = Channel.control(
+        "ISET{ch}?",
+        "ISET{ch}:%g",
+        """Control maximum current [A] of the power supply.""",
```
BenediktBurger, line 60: Afaik `[A]` is wrong according to SI, as it states: Give me the unit of A, which is nonsense.
Better write it as text:

```suggestion
        """Control maximum current in A of the power supply.""",
```
(or "in Ampere")

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @property
+    def status(self) -> dict:
+        """Get the status byte and and decode it."""
+        status_byte = int(self.ask("STATUS?").strip(), 2)
+        return self.decode_status_byte(status_byte)
+
+    tracking_mode = Instrument.setting(
+        "TRACK%d",
+        """Set the operation mode: independent, tracking series, or tracking parallel.
```
BenediktBurger, line 152:
```suggestion
        """Set the operation mode: independent, tracking series, or tracking parallel (strictly from :class:`TRACK_MODE`).
```

This will create a link to that enum class.

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,199 @@
...
+        NR1 0: Independent, 1: Series, 2: Parallel""",
+        validator=strict_discrete_set,
+        values=TRACK_MODE,
+        map_values=True,
```
BenediktBurger, line 156: If you use an enum, you do not need to use map_values.
`tracking_mode = TRACK_MODE.PARALLEL` will work out of the box. (Probably you will need to make it an IntEnum)

map_values is useful for a dictionary, where the key will be mapped to a value (as you did with the booleans) or an entry of a list will be mapped to the key (which you could use quite well for the baud rate).

pymeasure/instruments/gwinstek/gwinstekGPD2303S.py
```diff
@@ -0,0 +1,199 @@
...
+        validator=strict_discrete_set,
+        values=[1, 2, 3, 4],
+    )
+
+    baude_rate = Instrument.setting(
+        "BAUD%d",
+        """Set the baud rate to 2:9600, 1:57600, 0:115200""",
+        validator=strict_discrete_set,
+        values=BAUD_RATE,
+        map_values=True,
```
BenediktBurger, line 187: See above regarding enum and map_values.


```suggestion
        values=[115200, 57600, 9600],
        map_values=True,
```
This makes `baud_rate = 57600` send BAUD1 to the device.
No need for an enum.

docs/api/instruments/gwinstek/gpd2303s.rst
BenediktBurger, line 1: Please make the hashes as long as the text.

PR #1384, opened 2025-11-19
Input parameter widgets update

dvalde0:
Input parameter widgets update and code cleanup. Planned changes and features:
- [x] trailing input widget buttons (small buttons that appear on hover on the left and interact with the parameter)
- [x] reset trailing button - resets the input to the default if one is set
- [x] range parameter - a separate parameter that sets start, step/no_steps, stop and returns an iterable - or an update to vector parameter with a trailing button that opens a GUI that sets the value of the parameter to the range.
- [x] a trailing button for range parameter that lets the user pick the range from a plot via a pyqtgraph ROI.
- [x] collapsible animated input groups with trailing button support
- [x] typing for inputs and parameters
   - [x] display/inputs.py
   - [x] experiment/parameters.py
   - [ ] display/widgets/inputs_widget.py
- [ ] documentation for adding custom inputs

I am trying to not make breaking changes, but some I found inevitable (though I consider minor, awaiting your review). Current breaking changes:
- inputs are no longer declared with the `Input` mixin, instead they are declared by inheriting from the generic `Input[QWidget]`. The input also has now its own separate API, no longer being tied to the Qt namespace (setValue -> set_value, setPrefix -> set_prefix, etc). This breaks current custom input declaration passed through ui_class in the Parameter class.

### DIFF
--- a/examples/Basic/inputs_gallery.py
+++ b/examples/Basic/inputs_gallery.py
@@ -23,21 +23,24 @@
 #
 
 """
-This example demonstrates how to make a graphical interface, and uses
-a random number generator to simulate data so that it does not require
-an instrument to use. It also demonstrates the use of the sequencer module.
+This is a gallery of all the available inputs in the package.
 
 Run the program by changing to the directory containing this file and calling:
 
-python gui_sequencer.py
+python gui.py
 
 """
-
 import sys
-import random
-from time import sleep
 
-from pymeasure.experiment import Procedure, IntegerParameter, Parameter, FloatParameter
+from pymeasure.experiment import (
+    Procedure,
+    IntegerParameter,
+    Parameter,
+    FloatParameter,
+    BooleanParameter,
+    ListParameter,
+    VectorParameter,
+)
 from pymeasure.display.Qt import QtWidgets
 from pymeasure.display.windows import ManagedWindow
 
@@ -48,44 +51,33 @@ log.addHandler(logging.NullHandler())
 
 
 class TestProcedure(Procedure):
-    iterations = IntegerParameter("Loop Iterations", default=100)
-    delay = FloatParameter("Delay Time", units="s", default=0.2)
-    seed = Parameter("Random Seed", default="12345")
+    float_param = FloatParameter("Float Parameter", units="s", default=0.2, step=0.01)
+    int_param = IntegerParameter("Integer Parametr", units="A", default = 1, step=1)
+    param = Parameter("Parameter", default = "text")
+    bool_param = BooleanParameter("Boolean Parameter", default=True)
+    list_param = ListParameter("List Parameter", choices = ['Choice1', 'Choice 2', 'Choice 3'])
+    vector_param = VectorParameter("Vector Parameter", default = [1,2,3], units = "m")
 
     DATA_COLUMNS = ["Iteration", "Random Number"]
 
     def startup(self):
-        log.info("Setting up random number generator")
-        random.seed(int(self.seed))
-
+        pass
+    
     def execute(self):
-        log.info("Starting to generate numbers")
-        for i in range(self.iterations):
-            data = {"Iteration": i, "Random Number": random.random()}
-            log.debug("Produced numbers: %s" % data)
-            self.emit("results", data)
-            self.emit("progress", 100 * i / self.iterations)
-            sleep(self.delay)
-            if self.should_stop():
-                log.warning("Catch stop command in procedure")
-                break
-
+        pass
+    
     def shutdown(self):
-        log.info("Finished")
+        pass
 
 
 class MainWindow(ManagedWindow):
     def __init__(self):
         super().__init__(
             procedure_class=TestProcedure,
-            inputs=["iterations", "delay", "seed"],
-            displays=["iterations", "delay", "seed"],
+            inputs=["float_param", "int_param", "param", "bool_param", "list_param", "vector_param"],
+            displays=["float_param"],
             x_axis="Iteration",
             y_axis="Random Number",
-            sequencer=True,
-            sequencer_inputs=["iterations", "delay", "seed"],
-            sequence_file="gui_sequencer_example_sequence.txt",
-            inputs_in_scrollarea=True,
         )
         self.setWindowTitle("GUI Example")
 
--- a/examples/Basic/inputs_gallery.py
+++ b/examples/Basic/inputs_gallery.py
@@ -46,15 +46,14 @@ from pymeasure.display.windows import ManagedWindow
 
 import logging
 
-from pymeasure.experiment.parameters import PhysicalParameter
+from pymeasure.experiment.parameters import PhysicalParameter, ParameterGroup, RangeParameterGroup
 
 log = logging.getLogger("")
 log.addHandler(logging.NullHandler())
 
-
 class TestProcedure(Procedure):
     float_param = FloatParameter("Float Parameter", units="s", default=0.2, step=0.01)
-    int_param = IntegerParameter("Integer Parametr", units="A", default = 1, step=1)
+    int_param = IntegerParameter("Integer Parameter", units="A", default = 1, step=1)
     param = Parameter("Parameter", default = "text")
     bool_param = BooleanParameter("Boolean Parameter", default=True)
     list_param = ListParameter(
@@ -64,11 +63,17 @@ class TestProcedure(Procedure):
         "Vector Parameter", default = [1,2,3], units = "m", group_name="Test"
     )
     phys_param = PhysicalParameter("Physical Parameter", default = [20,3])
+    param_group = ParameterGroup("Test group",
+                                 test1 = FloatParameter("test1", default=0),
+                                 test2 = FloatParameter("test2", default=0)
+                                 )
+    wl_range = RangeParameterGroup("Wavelength Range")
+    
     
     DATA_COLUMNS = ["Iteration", "Random Number"]
 
     def startup(self):
-        pass
+        print(self.wl_range)
     
     def execute(self):
         pass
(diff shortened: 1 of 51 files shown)


dvalde0:
As the size of the pull request is pretty large, I will close it and open separate more digestible ones.

pymeasure/display/widgets/inputs_widget.py
```diff
@@ -25,45 +25,68 @@
...
+                element.selection_triggered.connect(partial(self.selection_triggered.emit, name, "1D"))
+
+            elif isinstance(parameter, parameters.Range2DParameterGroup):
+                element = Range2DInputGroup(parameter)
+                element.selection_triggered.connect(partial(self.selection_triggered.emit, name, "2D"))
+
+            elif isinstance(parameter,parameters.ParameterGroup):
+                element = InputGroup(parameter)
+                
+            elif parameter.ui_class is not None:
```
CasperSchippers, line 83: Feels like it makes most sense to have this one as first option, such that someone can "overrule" the default if he wants to  Or would you disagree?

pyrightconfig.json
CasperSchippers, line 1: Could you remove this file?

pyproject.toml
```diff
@@ -42,6 +42,7 @@ dependencies = [
     "pyserial>=2.7",
     "pyqtgraph>=0.12",
     "qtpy>=2.4",
+    "pyside6>=6.10.1",
```
CasperSchippers: We´d rather not have a pyqt or pyside as explicit dependency (people have to install one themselves)

pymeasure/experiment/parameters.py
```diff
@@ -460,34 +518,36 @@ def convert(self, value):
...
+    ) -> None:
         oldType = self._utype.value
         self._utype.value = uncertaintyType
         newType = self._utype.value
 
         if self.is_set():
             # Convert uncertainty value to the new type
             if (oldType, newType) == ('absolute', 'relative'):
-                self._value[1] = abs(self._value[1] / self._value[0])
+                self.value[1] = abs(self.value[1] / self.value[0])
```
CasperSchippers, line 539: What is the reason for changing this to `.value` instead of `._value`? `._value` skips the converter /.is_set check, which is fine in this case, right?

pymeasure/display/windows/managed_window.py
```diff
@@ -51,6 +54,20 @@
 log.addHandler(logging.NullHandler())
 
 
+class TabArea(QtWidgets.QTabWidget):
+    selection_started = QtCore.Signal()
+    selection_completed = QtCore.Signal(object)
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
```
CasperSchippers: This method does not seem needed, right?

pymeasure/display/widgets/tab_widget.py
```diff
@@ -74,3 +79,19 @@ def clear_widget(self):
         """
 
         return None
+
+    def selection_tags(self) -> list[str]:
+        return []
+    
+    def enter_selection_mode(self):
```
CasperSchippers, line 87: Maybe a check that selection_active==false before entering selection mode? I ran into the issue that I activated multiple selection_modes

CasperSchippers, line 87: What happens if the variable for which the selector option is clicked is not available in the current view? Maybe a check for this could be useful?

pymeasure/experiment/parameters.py
```diff
@@ -498,7 +558,125 @@ def __repr__(self):
         return "<{}(name={},value={},units={},uncertaintyType={})>".format(
             self.__class__.__name__, self.name, self._value, self.units, self._utype.value)
 
+class ParameterGroup:
+
```
CasperSchippers, line 566: Just a reminder: don´t forget the docstring

CasperSchippers (commented):
I went through some of the files, but as it is quite a bit, I'll need a second round (but that will be later this week, or maybe this weekend). Here's already a number of remarks. 

As of yet, it is unclear why CI fails :thinking:

pyrightconfig.json
dvalde0: Can i add it to .gitignore? While the vars for qtpy are in pyproject.toml, I would prefer to use PySide6 instead of the default set now.

CasperSchippers: That is fine I think indeed

examples/Basic/inputs_gallery.py
CasperSchippers, line 1: This example does not seem to work: it errors on the fact that the image widget cannot find the 'voltage1 (V)_start' parameter (and after that probably also the '_stop' and the same for voltage2)

examples/Basic/inputs_gallery.py
```diff
@@ -0,0 +1,128 @@
...
+                                         "current (A)"],
+                               "voltage1 (V)",
+                               "voltage2 (V)",
+                               selectors=["voltages"]),]
+        super().__init__(
+            procedure_class = TestProcedure,
+            inputs = inputs,
+            displays = ["float_param"],
+            widget_list = widgets
+        )
```
CasperSchippers, line 126: Maybe good to pass the option `inputs_in_scrollarea=True` here: currently the example does not fit on my (and hence probably also many other) monitors (widget is too high).

examples/Basic/inputs_gallery.py
CasperSchippers, line 1: I also noted in this example that the input fields are now fairly small (not full width as they used to be); could that be solved?

pymeasure/display/input_groups.py
```diff
@@ -0,0 +1,182 @@
...
+    def getAngle(self) -> float:
+        return self._angle
+
+    def setAngle(self, v: float):
+        self._angle = v
+        self.update()
+
+    angle = QtCore.Property(float, getAngle, setAngle)
+    
+    def animate(self, checked: bool):
```
CasperSchippers, line 53: This looks quite good!!

pymeasure/display/input_groups.py
```diff
@@ -0,0 +1,182 @@
...
+    def setup_ui(self):
+        self.elements = {}
+        for _,parameter in self.parameter.parameters.items():
+            if parameter.ui_class is not None:
+                element = parameter.ui_class(parameter)
+                
+            elif isinstance(parameter,parameters.ParameterGroup):
+                element = InputGroup(parameter)
+                
+            elif isinstance(parameter, parameters.FloatParameter):
```
CasperSchippers, line 109: What is not completely clear to me: it seems that the float-parameters in this group do not have a unit or the up-down arrows that normal float-parameters have; is that something that can be fixed?

pymeasure/display/inputs.py
```diff
@@ -25,52 +25,124 @@
...
 
     :param parameter: The parameter to connect to this input box.
     :attr parameter: Read-only property to access the associated parameter.
     """
-
-    def __init__(self, parameter, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(self, widget: Q, parameter: P, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.setMouseTracking(True)
```
CasperSchippers, line 76: Out of curiosity: what is this used for?

Edit: Is it for the trailing icon that shows when the cursor hovers over the input? That looks very neat!

pymeasure/display/inputs.py
```diff
@@ -81,47 +153,44 @@ def parameter(self):
...
+        return self.widget.setText(value)
 
-    def setValue(self, value):
-        # QtWidgets.QLineEdit has a setText() method instead of setValue()
-        return super().setText(value)
-
-    def setSuffix(self, value):
-        pass
+    def set_units(self, units: str) -> None:
+        return None
```
CasperSchippers: Small thing: this seems unnecessary as it overwrites a method with the same implementation, right?

CasperSchippers (commented):
Here are some additional comments.

I do want to say that I like the groups!

CasperSchippers:
The failing CI seems to be related to the inclusion of pyside in pyproject.toml, removing that line seems to fix it (tested this in a separate fork and branch)

pymeasure/display/widgets/tab_widget.py
dvalde0:
```python
class TabArea(QtWidgets.QTabWidget):
    selection_started = QtCore.Signal()
    selection_completed = QtCore.Signal(object)

    def start_selection(self, selector, tag, initial_range):
        self.selection_started.emit()
        wdg = self.currentWidget()
        if tag in wdg.selection_tags():
            if selector in wdg.selectors:
                wdg.selection_finished.connect(lambda result: self.selection_completed.emit([selector, result]))
                wdg.enter_selection_mode(initial_range)

```
This is the code for the TabArea. When an input widget emits to start a selection it sends a tag and this checks if the current widget supports it.

pymeasure/display/inputs.py
dvalde0: Yes, exactly.

examples/Basic/inputs_gallery.py
dvalde0: Hmm, I think i turned off the expanding size policy, as the trailing button would flicker the UI an every hover. I would appreciate another solution to this.

```python
self.widget.setSizePolicy(QtWidgets.QSizePolicy.Policy.Fixed, QtWidgets.QSizePolicy.Policy.Preferred)
```

dvalde0: I fixed the size of the input widget to always include the trailing buttons, but this results in the trailing buttons floating to the left instead of being left aligned to the widget. Please check the example in the latest commit.

pymeasure/display/input_groups.py
dvalde0: They do not have them cause the example does not define any units or step sizes, i added them and a way to define attributes for all params at the same time.

pymeasure/display/inputs.py
dvalde0: Yes, I just overwrote all the base class functions when reimplementing the inputs, will delete it.

pymeasure/display/widgets/inputs_widget.py
```diff
@@ -39,31 +41,54 @@ class InputsWidget(QtWidgets.QWidget):
...
+
+    def _setup_ui(self) -> None:
         parameter_objects = self._procedure.parameter_objects()
         for name in self._inputs:
             parameter = parameter_objects[name]
-            if parameter.ui_class is not None:
-                element = parameter.ui_class(parameter)
 
+            if isinstance(parameter,parameters.ParameterGroup):
+                element = InputGroup(parameter)
```
CasperSchippers, line 73: I think this case should be after the 1D and 2D input groups, as this will also be true for those groups (as they are subclasses)

PR #1385, opened 2025-11-20
Initial commit for FormFactor Velox

Falk3579:
Add support for FormFactor Waferprobers that use the Velox software.

### DIFF
--- a/pymeasure/instruments/formfactor/velox.py
+++ b/pymeasure/instruments/formfactor/velox.py
@@ -44,7 +44,6 @@ PROBER_REFERENCES = {"home": "H",
                      "center": "C",
                      "diehome": "D",
                      }
-EXPECTED_ERRORS = [703]  # list of errors that do not raise a ConnectionError
 
 
 class Chuck(Channel):
@@ -128,6 +127,9 @@ class Chuck(Channel):
 class WaferMap(Channel):
     """A class representing the functions of the Wafermap module."""
 
+    end_reached = False
+    """Get whether the end of the wafermap is reached or not (bool)."""
+
     def step_first_die(self):
         """Move the chuck to the first die and clear the the binning results.
 
@@ -178,7 +180,7 @@ class WaferMap(Channel):
 
     enabled = Channel.measurement(
         "IsAppRegistered WaferMap",
-        """Get whether the WaferMap module is loaded or not.""",
+        """Get whether the WaferMap module is loaded or not (bool).""",
         map_values=True,
         values={True: 1, False: 0}
         )
@@ -195,7 +197,7 @@ class Velox(SCPIMixin, Instrument):
 
         prober = Velox("GPIB0::28::INSTR")  # replace the resource to your needs
         coordinates = prober.wafermap.step_first_die()  # move to the first die in the wafermap
-        while not prober.error_code:
+        while not self.wafermap.end_reached:
             prober.chuck.move_contact()  # move to contact height
             # ...
             # code for measurement, processing and storing data
@@ -241,6 +243,7 @@ class Velox(SCPIMixin, Instrument):
 
         self.error_code = 0
         self.error_message = ""
+        self.wafermap.end_reached = False
 
         got = super().read()
 
@@ -255,10 +258,13 @@ class Velox(SCPIMixin, Instrument):
 
         if self.error_code:
             self.error_message = response
-            if self.error_code not in EXPECTED_ERRORS:
-                raise ConnectionError(f"{self.error_code}: {self.error_message}")
-            else:
+            if self.error_code == 703:
+                # 703 means end of wafermap reached, it's not a hard error
+                # more a status indicator, so it doesn't throw an exception
+                self.wafermap.end_reached = True
                 log.info(f"{self.error_code}: {self.error_message}")
+            else:
+                raise ConnectionError(f"{self.error_code}: {self.error_message}")
 
         return response
 


docs/api/instruments/formfactor/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.formfactor
+
+##########
+FormFactor
+##########
+
+This section contains specific documentation on the FormFactor wafer probers that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
```
BenediktBurger: Does FormFactor only manufactur wafer probers?
Typically we have a folder per manufacturer. So, it should be more like FormFactor instruments/devices or something like that.

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+#
+
+
+import logging
+
+from pymeasure.instruments import Channel, Instrument, SCPIMixin
+from pymeasure.instruments.validators import strict_range, strict_discrete_set
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
```
BenediktBurger, line 32: If you do not use the logger, you do not have to import it.

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+        if ":" not in got:
+            return got
+
+        got = got.partition(":")
+        self.error_code = int(got[0])
+        response = got[2].strip()
+
+        if self.error_code:
+            self.error_message = response
+            if self.error_code not in EXPECTED_ERRORS:
```
BenediktBurger: Why do you expect that error?

Actually, here you could use the logger to log these errors instead of dropping them silently.

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+                raise VeloxError(f"{self.error_code}: {self.error_message}")
```
BenediktBurger: I typically prefer to use builtin exceptions as you do not have to import the exception as well.

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+        """Control the chuck index (x, y) in micrometer (float, float).""",
+        separator=" ",
+        check_set_errors=True,
+        )
+
+
+class WaferMap(Channel):
+    """A class representing the functions of the Wafermap module."""
+
+    def step_firstdie(self):
```
BenediktBurger: Why do you go only half the way with snake case, why not `step_first_die`?

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+
+class VeloxError(BaseException):
+    """A class representing Velox errors."""
+    pass
+
+
+class Chuck(Channel):
+    """A class representing the Chuck functions of the wafer prober."""
+
+    def move_contact(self, velocity=100):
```
BenediktBurger, line 52: Maybe `move_until_contact` to be more explicit?

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+        :param  int y_steps: Index steps in Y direction
+        :param  str pos_ref: Position reference, strictly ``home``,
+            ``zero``, ``center`` or ``diehome``.
+        :param  float velocity: Speed in percent, strictly from ``0.1`` to ``100``.
+
+        """
+        pos_ref = pos_ref.lower()
+        strict_discrete_set(pos_ref, PROBER_REFERENCES.keys())
+        strict_range(velocity, VELOCITY_RANGE)
+        _ref = pos_ref[0].upper()
```
BenediktBurger: why don't you use the dictionary if you already defined it?

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+    def step_nextdie(self, *args):
+        """Move the chuck to the specified die coordinates. If no argument is passed,
+        the prober moves to the next logical die.
+
+        :param \\*args: Die coordinates. The coordinates can be passed as individual arguments or
+            as a single argument.
+            Examples of valid arguments to move to x=5, y=6:
+            ``5, 6``, ``"5 6"``, ``"5, 6"``, ``"5; 6"``, ``[5, 6]``, ``(5, 6)``
+
+        :return: List of int containing the new die coordinates.
```
BenediktBurger: That sounds complicated and error prone to allow that many possibilities. Why don't you limit it to a single option? E.g. two parameters or a tuple parameter?

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+
+    error_code = 0
+    error_message = ""
+
+    chuck = Instrument.ChannelCreator(Chuck)
+    wafermap = Instrument.ChannelCreator(WaferMap)
+
+    def __init__(self, adapter, name="FormFactor Velox",
+                 timeout=40000,
```
BenediktBurger, line 217: That is a quite long timeout per default.

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+        self.error_code = int(got[0])
+        response = got[2].strip()
+
+        if self.error_code:
+            self.error_message = response
+            if self.error_code not in EXPECTED_ERRORS:
+                raise VeloxError(f"{self.error_code}: {self.error_message}")
+
+        if response == "":
+            response = "OK"  # for commands that only return error code 0
```
BenediktBurger: As you raise an exception on error, you could just return the empty string if that is the result instead of a "OK" string.

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,270 @@
...
+            coordinates += " "
+
+        if len(coordinates):
+            coordinates = " ".join(coordinates.split())  # ensure single space between elements
+
+        command = f"StepNextDie {coordinates}".strip()
+        return self.values(command,
+                           separator=" ",
+                           cast=int,
+                           )
```
BenediktBurger: Why do you return something?

I do not expect a method which does an action (here step to next die) to return anything.

According to Command-Query-Separation, a method should either do something (like here) or return something (without doing anything).

The device returns an error if it happened and you would raise an exception.

If you wanted to check for errors, you could add a check here: `result = self.values(...); if result: raise Error`. However, I thought that `read` already handles it.

tests/instruments/formfactor/test_formfactorVelox_with_device.py
```diff
@@ -0,0 +1,120 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+# Call signature:
```
BenediktBurger, line 25: Maybe adding a note that you have to use your own device address.

tests/instruments/formfactor/test_formfactorVelox_with_device.py
```diff
@@ -0,0 +1,120 @@
...
+#############################################
+
+import pytest
+from time import sleep
+
+from pymeasure.instruments.formfactor.velox import Velox, VeloxError
+
+############
+# FIXTURES #
+############
```
BenediktBurger: I think we do not need this separation, especially as there is only a single fixture.

BenediktBurger (changes requested):

tests/instruments/formfactor/test_formfactorVelox_with_device.py
Falk3579: Comment removed

tests/instruments/formfactor/test_formfactorVelox_with_device.py
Falk3579: Note regarding own device address added.

docs/api/instruments/formfactor/index.rst
Falk3579: Replaced 'wafer probers' by 'devices'.

pymeasure/instruments/formfactor/velox.py
Falk3579: The device has an internal timeout of 30 seconds. If a movement command cannot be exected within 30 seconds than the device will respond with its own timeout error. The default timeout if 40 seconds for the driver allows to capture that error.

pymeasure/instruments/formfactor/velox.py
Falk3579: Replaced 'VeloxError' by 'ConnectionError'.

pymeasure/instruments/formfactor/velox.py
Falk3579: Renamed 'step_firstdie' to 'step_first_die' and 'step_nextdie' to 'step_next_die'.

pymeasure/instruments/formfactor/velox.py
Falk3579: Done as recommended.

pymeasure/instruments/formfactor/velox.py
Falk3579: Now used to log expected errors.

pymeasure/instruments/formfactor/velox.py
Falk3579: Expected errors are logged now.
Currently error 703 is an expected error. It is signaled when the end of the wafer is reached after multiple step_next_die executions. So it is more or less a wanted response and tells the user that the measurements are done. Other errors are 'real'.

pymeasure/instruments/formfactor/velox.py
Falk3579: Good point. I use the dictionary now.

pymeasure/instruments/formfactor/velox.py
Falk3579: Somehow 'move_until_contact' implies to me that the system is searching for the correct contact height. But this is not the case.
Other wafer prober vendors also call this command 'move_contact'.

pymeasure/instruments/formfactor/velox.py
Falk3579: Renamed the method to 'step_to_die'.
Reduced it to a single option with two/three parameters.

pymeasure/instruments/formfactor/velox.py
Falk3579: Now only 'step_next_die' and 'step_first_die' return the new coordinates.

pymeasure/instruments/formfactor/velox.py
```diff
@@ -0,0 +1,267 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2025 PyMeasure Developers
```
BenediktBurger: We're in 2026 now (also for the files below).

pymeasure/instruments/formfactor/velox.py
BenediktBurger: I'd like to have a comment somewhere (maybe in the EXPECTED_ERRORS list) why it is an expected error, or better, why it is more a warning/indicator and not an actual error.

pymeasure/instruments/formfactor/velox.py
Falk3579: Year changed to 2026.

pymeasure/instruments/formfactor/velox.py
Falk3579: EXPECTED_ERRORS are removed now. Instead I use the variable end_reached in the WaferMap class to indicate whether the end of the map is reached or not.

PR #1387, opened 2025-11-28
Rework the Agilent B1500 SMU class to inherit from Channel

ilev-sergey:
A bit of context about the instrument and the current implementation to help understand this PR.

The instrument has 10 slots that can host various units (SMUs, SPGUs, CMUs, etc.). By querying the instrument, it is possible to determine which unit is installed in each slot. Thus, dynamic initialization of channels is required.

For example, 4 basekit SMUs can be installed in slots 4, 6, 7, 8. Current implementation of `AgilentB1500` class implies that users access SMUs through references such as `.smu1`, `.smu2`, etc. These identifiers do not correspond to the physical slot numbers but rather to the sequential count of detected SMUs. Thus, in the example above, the SMUs would be accessible as `.smu1`, `.smu2`, `.smu3`, `.smu4` respectively. The same approach of counting SMUs starting from the lowest slot number is mentioned in manual and used in Agilent EasyEXPERT Software.

In the PR I reworked the SMU class into the subclass of `Channel`, but kept the same logic of numerating SMUs, by using SMU count number for creating SMU references and SMU slot number for initialization of the SMU itself. This way, SMUs are available by their count number via `.smus[1]` or `.smu1`, but the commands itself use `self.id` = 4 that is the actual slot number.

Existing `AgilentB1500` SMU properties were reworked as well to support backward compatibility.

### DIFF
--- a/pymeasure/instruments/agilent/agilentB1500.py
+++ b/pymeasure/instruments/agilent/agilentB1500.py
@@ -443,7 +443,7 @@ class AgilentB1500(SCPIMixin, Instrument):
                 return channel
 
     class _data_formatting_FMT1(_data_formatting_generic):
-        """Data formatting for FMT1 format"""
+        """Data formatting for FMT1 format."""
 
         def __init__(self, smu_names={}, output_format_string="FMT1"):
             super().__init__(smu_names, output_format_string)
@@ -468,13 +468,13 @@ class AgilentB1500(SCPIMixin, Instrument):
             return (status, channel, data_name, value)
 
     class _data_formatting_FMT11(_data_formatting_FMT1):
-        """Data formatting for FMT11 format (based on FMT1)"""
+        """Data formatting for FMT11 format (based on FMT1)."""
 
         def __init__(self, smu_names={}):
             super().__init__(smu_names, "FMT11")
 
     class _data_formatting_FMT21(_data_formatting_generic):
-        """Data formatting for FMT21 format"""
+        """Data formatting for FMT21 format."""
 
         def __init__(self, smu_names={}):
             super().__init__(smu_names, "FMT21")
@@ -884,7 +884,7 @@ class AgilentB1500(SCPIMixin, Instrument):
         return self.query_learn_header(46)
 
     def query_meas_ranges(self):
-        """Read measruement ranging status (32) for all SMUs."""
+        """Read measurement ranging status (32) for all SMUs."""
         return self.query_learn_header(32)
 
 
@@ -896,9 +896,11 @@ class AgilentB1500(SCPIMixin, Instrument):
 class SMU(Channel):
     """Provide specific methods for the SMUs of the Agilent B1500 mainframe.
 
-    :param int slot: Slot number of the SMU
-    :param str smu_type: Type of the SMU
+    :param AgilentB1500 parent: Instance of the B1500 mainframe class
+    :param int index: Index of the SMU
+    :param str type: Type of the SMU
     :param str name: Name of the SMU
+    :param int slot: Slot number of the SMU
     """
 
     def __init__(self, parent, index, type, name, slot, **kwargs):
@@ -1050,7 +1052,7 @@ class SMU(Channel):
         :param int or str source_range: Output range index or name
         :param float output: Source output value in A or V
         :param float comp: Compliance value, defaults to previous setting
-        :param CompliancePolarity comp_polarity: Compliance polairty, defaults to auto
+        :param CompliancePolarity comp_polarity: Compliance polarity, defaults to auto
         :param int or str comp_range: Compliance ranging type, defaults to auto
         """
         if source_type.upper() == "VOLTAGE":
@@ -1092,10 +1094,10 @@ class SMU(Channel):
         Each step is separated by a pause duration.
 
         :param str source_type: Source type (``'Voltage'`` or ``'Current'``)
+        :param int or str source_range: Output range index or name
         :param float target_output: Target output voltage or current
-        :param int irange: Output range index
         :param float comp: Compliance, defaults to previous setting
-        :param CompliancePolarity comp_polarity: Compliance polairty, defaults to auto
+        :param CompliancePolarity comp_polarity: Compliance polarity, defaults to auto
         :param int or str comp_range: Compliance ranging type, defaults to auto
         :param float stepsize: Maximum size of steps
         :param float pause: Duration in seconds to wait between steps
@@ -1554,7 +1556,7 @@ class SPGU(Channel):
         "SPPER?",
         "SPPER %f",
         """Control the pulse period for SPGU channels (``SPPER``) in seconds (float).
-        Applies to all installed SPGU modules""",
+        Applies to all installed SPGU modules.""",
         validator=strict_range,
         values=[2e-8, 10],
     )
@@ -1691,6 +1693,7 @@ class SPGUChannel(Channel):
 
     def get_pulse_timings(self, source=1):
         """Get the timing parameters for the SPGU channel. (``SPT?``)
+
         The SPGU operating mode must be set to PG with the ``SIM 0`` command before getting the
         pulse timings.
 
@@ -1726,7 +1729,7 @@ class CustomIntEnum(IntEnum):
     """
 
     def __str__(self):
-        """Give title case string of enum value"""
+        """Give title case string of enum value."""
         return str(self.name).replace("_", " ").title()
         # str() conversion just because of pylint bug
 
@@ -1744,7 +1747,7 @@ class CustomIntEnum(IntEnum):
 
 
 class ADCType(CustomIntEnum):
-    """ADC Type"""
+    """ADC Type."""
 
     HSADC = (0,)  #: High-speed ADC
     HRADC = (1,)  #: High-resolution ADC
@@ -1756,7 +1759,7 @@ class ADCType(CustomIntEnum):
 
 
 class ADCMode(CustomIntEnum):
-    """ADC Mode"""
+    """ADC Mode."""
 
     AUTO = 0  #:
     MANUAL = 1  #:
@@ -1765,14 +1768,14 @@ class ADCMode(CustomIntEnum):
 
 
 class AutoManual(CustomIntEnum):
-    """Auto/Manual selection"""
+    """Auto/Manual selection."""
 
     AUTO = 0  #:
     MANUAL = 1  #:
 
 
 class ControlMode(CustomIntEnum):
-    """Control mode for the digital I/O ports"""
+    """Control mode for the digital I/O ports."""
 
     GENERAL = 0  #: General purpose control mode (default)
     SMU_PGU_SELECTOR = 1  #: 16440A SMU/PGU selector (B1500A-A04) control mode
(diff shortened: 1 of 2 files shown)


tests/instruments/agilent/test_agilentB1500.py
```diff
@@ -73,6 +74,29 @@ class AgilentB1500Mock(AgilentB1500):
...
+class TestSMU:
+    """Tests for SMU module functionality."""
+
+    channel = 1
+
+    def test_enable(self):
+        """Test enable method."""
+        with expected_protocol(
+            AgilentB1500Mock,
+            [(f"CN {self.channel}", None)],
```
BenediktBurger, line 94: You can hardcode the 1 into the string inside these tests.

Thanks for using tests for ensuring consistent behaviour.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -941,42 +928,36 @@ def __init__(self, parent, channel, smu_type, name, **kwargs):
                 "UHVU",
             ],
         )
-        self.voltage_ranging = SMUVoltageRanging(smu_type)
-        self.current_ranging = SMUCurrentRanging(smu_type)
+        self.voltage_ranging = SMUVoltageRanging(type)
+        self.current_ranging = SMUCurrentRanging(type)
         self.name = name
+        self.type = type
```
BenediktBurger: Do you need this attribute later on?

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -941,42 +928,36 @@ def __init__(self, parent, channel, smu_type, name, **kwargs):
                 "UHVU",
             ],
         )
-        self.voltage_ranging = SMUVoltageRanging(smu_type)
-        self.current_ranging = SMUCurrentRanging(smu_type)
+        self.voltage_ranging = SMUVoltageRanging(type)
+        self.current_ranging = SMUCurrentRanging(type)
```
BenediktBurger: You rename it, as it is not only SMU but more general, right?

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -912,22 +901,20 @@ def query_meas_ranges(self):
...
-    :param AgilentB1500 parent: Instance of the B1500 mainframe class
     :param int channel: Channel number of the SMU
     :param str smu_type: Type of the SMU
     :param str name: Name of the SMU
     """
 
-    def __init__(self, parent, channel, smu_type, name, **kwargs):
-        # to allow garbage collection for cyclic references
-        self._b1500 = weakref.proxy(parent)
+    def __init__(self, parent, number, type, name, channel, **kwargs):
```
BenediktBurger: Maybe `index` instead of number and `slot` instead of `channel` are better names?

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -912,22 +901,20 @@ def query_meas_ranges(self):
...
 
 
-class SMU:
+class SMU(Channel):
     """Provide specific methods for the SMUs of the Agilent B1500 mainframe
 
-    :param AgilentB1500 parent: Instance of the B1500 mainframe class
     :param int channel: Channel number of the SMU
     :param str smu_type: Type of the SMU
     :param str name: Name of the SMU
```
BenediktBurger, line 910: If you change the init signature, you have to change the docstring as well.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -912,22 +901,20 @@ def query_meas_ranges(self):
...
     """
 
-    def __init__(self, parent, channel, smu_type, name, **kwargs):
-        # to allow garbage collection for cyclic references
-        self._b1500 = weakref.proxy(parent)
+    def __init__(self, parent, number, type, name, channel, **kwargs):
+        super().__init__(parent, channel, **kwargs)
         channel = strict_discrete_set(channel, range(1, 11))
         self.channel = channel
```
BenediktBurger: Do we still need this `self.channel`, as we have `self.id` with the same content.
I would prefer if we changed to the standard naming `self.id`.

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: Here, and in similar places, I would use the `{ch}` placeholder for the channel id, as used by `insert_id` method instead of `self.write(f"CN {self.channel}"):

```suggestion
        self.write("CN {ch}")
```
Note that I removed the `f`, in order to give a plain string. The replacement is done later.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -941,42 +928,36 @@ def __init__(self, parent, channel, smu_type, name, **kwargs):
...
     def check_errors(self):
         """Wrap :meth:`~.AgilentB1500.check_errors` method of B1500."""
-        return self._b1500.check_errors()
+        return self.parent.check_errors()
 
     ##########################################
 
     def _query_status_raw(self):
-        return self._b1500.query_learn(str(self.channel))
+        return self.parent.query_learn(str(self.channel))
```
BenediktBurger: We could make sure that the `self.id` is a string, then we do not need these string conversions in every call.

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: Property docstrings like these cause the pipeline issues:

Property docstrings have to start with Get, Set, or Control depending whether they can read, write, or read and write.
If you rework all property docstrings in this SMU class, the tests will pass.

BenediktBurger (changes requested):

Here is a list of comments.

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: Different types of SMU have different available limits. E.g. it's useful to know type of SMU that you use when specifying `comp` in `.force` method. Otherwise, you can get an error from the instrument.

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: I rename it as I think it's more natural to call `SMU.type` instead of `SMU.smu_type`. It was an internal variable for initialization before anyway.

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: Resolved by #1380

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: Great idea! Thanks!

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: I wouldn’t mind, but it would break backward compatibility.

tests/instruments/agilent/test_agilentB1500.py
ilev-sergey: I’d keep it for explicitness.

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: What do you think about methods that use f-strings to specify other arguments beside `self.channel`?  See for example `.filter`
https://github.com/pymeasure/pymeasure/blob/3ad4aa6e6751807506be2e4da568a226d6fcd4cf/pymeasure/instruments/agilent/agilentB1500.py#L1011-L1015
F-string is convenient here to specify `setting` parameter but prevents using `{ch}` placeholder.
Do you think I should change the `self.channel` to placeholder where it's possible and keep other methods unaltered?

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: It feels more natural for me to have `self.id` as an int instead of string, as it's a number from 1 to 9. But I don't have a strong opinion on this point, so I can change it if you insist.

ilev-sergey:
@coderabbitai review

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: You can do `f"FL {setting} {{ch}}"`. The f-string formatting will inject the setting and remove the outer pair of curly braces from ch resulting in `"FL value {ch}"` if `setting="value"`.

So, you can still use a f-string while keeping "ch".

It works with more curly braces for even deeper nesting.

pymeasure/instruments/agilent/agilentB1500.py
BenediktBurger: I wanted to make it easier and save "str" calls. Most important is consistency.

You can keep it, if you want to.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -952,77 +956,73 @@ def check_errors(self):
...
-        """Query status of the SMU."""
-        return self.parent.query_learn_header(str(self.channel))
+        """Get status of the SMU."""
+        return self.parent.query_learn_header(str(self.id))
 
     def enable(self):
-        """Enable Source/Measurement Channel (``CN``)"""
-        self.write(f"CN {self.channel}")
+        """Enable source/measurement channel. (``CN``)"""
+        self.write(f"CN {self.id}")
```
BenediktBurger: Please use the `"CN {ch}"` syntax as already said.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1198,21 +1197,20 @@ def meas_range_voltage(self):
...
-                                   defaults to 50
+        :param int rate: Parameter used to calculate the *current* value, defaults to 50
         """
         mode = strict_range(mode, range(1, 4))
         if mode == 1:
-            self.write(f"RM {self.channel}, {mode}")
+            self.write(f"RM {self.id}, {mode}")
         else:
-            self.write(f"RM {self.channel}, {mode}, {rate}")
+            self.write(f"RM {self.id}, {mode}, {rate}")
```
BenediktBurger: Here it could be `f"RM {{ch}}, {mode}, {rate}" to inject the channel name later.

BenediktBurger (approved):
Except for that small syntax thing.

Thanks for your work and patience.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -64,7 +63,7 @@ def __init__(self, adapter, name="Agilent B1500 Semiconductor Parameter Analyzer
     @property
     def smu_references(self):
         """Get all SMU instances as dict_values."""
-        return self._smu_references.values()
+        return self.smus.values()
```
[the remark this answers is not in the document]
ilev-sergey: I think that's fine behavior. I'd rather get an explicit error than a "safe" empty dict. E.g. when using it like this
```py
for smu in self.smu_references:
    # do smth
```
I'd rather get an exception straight away than silently skip SMUs configuration because they weren't initialized properly.

ilev-sergey:
Fixed

Thank you very much for your review!

BenediktBurger:
Thank you very much!

PR #1389, opened 2025-12-01
Adding different Keithley 26*** models with Ch# and limits

AnotherPseudonym:
I wanted to add a 2602B model because it has different current limits than the configuration for the 2600. After some research, I see there are no 2600 models directly. The main differences between models from pymeasure's perspective are the number of channels and the current and voltage limits, so I created entries for models (based on user manuals) and added entries for them.
It might make sense to consolidate the models, but I know that if I didn't find an exact match, I would have to look up the specifications of the listed model and my model to see if the features that matter are the same, so I think it makes sense to have all the models, though using as much inheritance as possible.

### DIFF
--- a/pymeasure/instruments/keithley/keithley2600.py
+++ b/pymeasure/instruments/keithley/keithley2600.py
@@ -41,7 +41,7 @@ log.addHandler(logging.NullHandler())
 #    LV   HV  HV+fine current
 # 1 2601 2611 2635
 # 2 2602 2612 2636
-# 2 2604 2614 2634 - No Ethernet
+# 2 2604 2614 2634 - No Ethernet for lower cost
 
 class Keithley260X(SCPIMixin, Instrument):
     """Represents the Keithley 2600* series SourceMeter with at least one channel"""
@@ -87,7 +87,7 @@ class Keithley2600(Keithley260X):
     """Backward compatible 2600 model - Not a real product"""
     def __init__(self, adapter, name="Keithley 2600 SourceMeter", **kwargs):
         warn("This is a generic 2 channel model. Using the actual model may " \
-                " match limits better.", FutureWarning)
+                " match limits and channels better.", FutureWarning)
         super().__init__(
             adapter,
             name,
@@ -136,7 +136,7 @@ class Keithley2602(Keithley260X):
         self.ChB = ChannelLV(self, 'b')
 
 class Keithley2602A(Keithley2602):
-    '''Represents the 2602A Single Channel SMU'''
+    '''Represents the 2602A Dual Channel SMU'''
     def __init__(self, adapter, name="Keithley 2602A SourceMeter", **kwargs):
         super().__init__(
             adapter,
@@ -145,7 +145,7 @@ class Keithley2602A(Keithley2602):
         )
 
 class Keithley2602B(Keithley2602):
-    '''Represents the 2602B Single Channel SMU'''
+    '''Represents the 2602B Dual Channel SMU'''
     def __init__(self, adapter, name="Keithley 2602B SourceMeter", **kwargs):
         super().__init__(
             adapter,
@@ -154,7 +154,7 @@ class Keithley2602B(Keithley2602):
         )
 
 class Keithley2604B(Keithley2602):
-    '''Represents the 2604B Single Channel SMU'''
+    '''Represents the 2604B Dual Channel SMU'''
     # It doesn't look like 2604 non-Bs were made.
     def __init__(self, adapter, name="Keithley 2604B SourceMeter", **kwargs):
         super().__init__(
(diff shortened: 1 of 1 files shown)


pymeasure/instruments/keithley/keithley2600.py
```diff
@@ -76,8 +83,194 @@ def error(self):
...
+    def __init__(self, adapter, name="Keithley 2600 SourceMeter", **kwargs):
+        warn("This is a generic 2 channel model. Using the actual model may " \
+                " match limits and channels better.", FutureWarning)
+        super().__init__(
+            adapter,
+            name,
+            **kwargs
+        )
+        self.ChA = ChannelHV(self, 'a')
+        self.ChB = ChannelHV(self, 'b')
```
BenediktBurger, line 97: Please use the ChannelCreator for adding channels:
https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html
For this, the Channels have to inherit from channel.Channel class.
See also #746

Consider writing protocol tests for a good transition.

pymeasure/instruments/keithley/keithley2600.py
```diff
@@ -76,8 +83,194 @@ def error(self):
         return self.next_error
 
 
-class Channel:
+class Keithley2600(Keithley260X):
+    """Backward compatible 2600 model - Not a real product"""
```
BenediktBurger, line 87: You mean, this class is for backward compatibility with older code? Please make it more explicit.

docs/api/instruments/keithley/index.rst
```diff
@@ -18,6 +18,22 @@ This section contains specific documentation on the Keithley instruments that ar
...
+   keithley2602A
+   keithley2611A
+   keithley2612A
+   keithley2600B
+   keithley2601B
+   keithley2602B
+   keithley2604B
+   keithley2611B
+   keithley2612B
+   keithley2634B
```
BenediktBurger, line 36: These names refer to files (with the suffix .rst) containing documentation.
As you do not add these files, these will lead to void.
Actually, as you do keep the keithley2600.py file, you should not add here any additional files, but add the new classes (instruments and channesl) to the keithley2600.rst file.

pymeasure/instruments/keithley/__init__.py
```diff
@@ -31,6 +31,10 @@
 from .keithley2450 import Keithley2450
 from .keithley2510 import Keithley2510
 from .keithley2600 import Keithley2600
+from .keithley2600 import Keithley2601, Keithley2602, Keithley2611, Keithley2612
+from .keithley2600 import Keithley2601A, Keithley2602A, Keithley2611A, Keithley2612A
+from .keithley2600 import Keithley2601B, Keithley2602B, Keithley2604B, Keithley2611B, Keithley2612B
+from .keithley2600 import Keithley2634B, Keithley2635B, Keithley2636B
```
BenediktBurger, line 37: It is better to use a parenthesis or `\ ` to continue importing from the same module instead of this combination of several lines importing from the same module.

pymeasure/instruments/keithley/keithley2600.py
```diff
@@ -27,25 +27,32 @@
 from warnings import warn
 
 import numpy as np
-from pymeasure.instruments import Instrument, SCPIUnknownMixin
+from pymeasure.instruments import Instrument, SCPIMixin
```
BenediktBurger, line 30: You know that it is SCPI compliant? Thanks for fixing it!

pymeasure/instruments/keithley/keithley2600.py
```diff
@@ -27,25 +27,32 @@
...
+# We care about the limits not resolution.
+# For a given number, the limits are the same, but enumerating each may save user time.
+#    LV   HV  HV+fine current
+# 1 2601 2611 2635
+# 2 2602 2612 2636
+# 2 2604 2614 2634 - No Ethernet for lower cost
 
-class Keithley2600(SCPIUnknownMixin, Instrument):
-    """Represents the Keithley 2600 series (channel A and B) SourceMeter"""
+class Keithley260X(SCPIMixin, Instrument):
```
BenediktBurger, line 46: Sometimes it is also called 2600Base or similar to indicate that it is a parent for many devices and should not be used on its own.

pymeasure/instruments/keithley/keithley2600.py
```diff
@@ -76,8 +83,194 @@ def error(self):
...
+        )
+
+class Keithley2601B(Keithley2601):
+    '''Represents the 2601B Single Channel SMU'''
+    def __init__(self, adapter, name="Keithley 2601B SourceMeter", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            **kwargs
+        )
```
BenediktBurger, line 125: I don't see the reason to have three classes which are exactly the same.
Better rename the Keithley2601 to Keithley2601Series to indicate that it is suitable for several instruments of the same series.

Also with the instruments below.

pymeasure/instruments/keithley/keithley2600.py
```diff
@@ -226,7 +347,6 @@ def measure_voltage(self, nplc=1, voltage=21.0, auto_range=True):
         :param auto_range: Enables auto_range if True, else uses the set voltage
         """
         log.info("%s is measuring voltage." % self.channel)
-        self.write('measure.v()')
```
BenediktBurger, line 229: Why do you remove this?

pymeasure/instruments/keithley/keithley2600.py
```diff
@@ -323,3 +443,155 @@ def shutdown(self):
...
+    current = Instrument.measurement(
+        'measure.i()',
+        """ Get the current in Amps """
+    )
+
+    source_current = Instrument.control(
+        'source.leveli', 'source.leveli=%f',
+        """ Control the applied source current """,
+        validator=truncated_range,
+        values=[-3, 3]
```
BenediktBurger, line 539: I guess [dynamic properties](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#dynamic-validity-range) are a better solution to change the limits instead of duplicating the channels.

BenediktBurger (commented):
A few comments for your draft.

Thanks for your contribution.

PR #1391, opened 2025-12-01
SCPI mixin test

fg-lumentum:
Added a pytest module to test whether a generic device can safely inherit from SCPIMixin. Can be run using `pytest -k scpi_mixin --device-address "GPIB0::0"`. Added a section in the documentation to explain this.

The motivation here is that there are devices that, while not conforming *exactly* to IEEE 488.2, have enough compatibility to warrant inheriting from SCPIMixin. This test provides an easy way for developers to check this.


BenediktBurger:

PR #1394, opened 2025-12-28
New instrument: Ocean Optics Flame

bicarlsen:
Add the Ocean Optics Flame spectrometer.

### DIFF
--- a/pymeasure/instruments/oceanoptics/flame.py
+++ b/pymeasure/instruments/oceanoptics/flame.py
@@ -52,7 +52,9 @@ class CommandFlameS(IntEnum):
 
 class FlameS(Instrument):
     """Control the Ocean Optics Flame S instrument."""
+    BYTE_ORDER = "little"
     NUM_PIXELS = 2048
+
 
     def __init__(self, adapter, name="Ocean Optics Flame S", **kwargs):
         super().__init__(adapter, name, **kwargs)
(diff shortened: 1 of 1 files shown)


docs/api/instruments/oceanoptics/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.newport
```
BenediktBurger, line 1: why do you refer to the newport folder?

pymeasure/instruments/oceanoptics/flame.py
```diff
@@ -0,0 +1,136 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from enum import IntEnum
+from pymeasure.instruments import Instrument, validators
+
+
+class CommandFlameS(IntEnum):
```
BenediktBurger, line 29: Interesting idea to use enums for the commands.

Typically, you do not need these numbers outside the methods, so you could use them only in the appropriate methods. However, it is a nice overview, so you can keep them, if you like.

pymeasure/instruments/oceanoptics/flame.py
```diff
@@ -0,0 +1,136 @@
...
+    PSOC_WRITE = 0x69
+    WRITE_REGISTER_INFO = 0x6A
+    READ_REGISTER_INFO = 0x6B
+    READ_PCB_TEMPERATURE = 0x6C
+    READ_IRRADIANCE_CALIBRATION_FACTORS = 0x6D
+    WRITE_IRRADIANCE_CALIBRATION_FACTORS = 0x6E
+    QUERY_INFO = 0xFE
+
+
+class FlameT(Instrument):
```
BenediktBurger, line 100: We typically use one file per instrument, unless there are several instruments which inherit from the same parent and only differ in small details.

PR #1399, opened 2026-01-23
Add new instruments Cryomagnetics 4G Magnet Power Supply family

JAW90:
This adds support for the Cryomagnetics, Inc. family of '4G Magnet Power Supply'; specifically, the 4G-100, 4G-150s, 4G-200s and 4G-dual. The family follow the same instruction set, except for an additional command for the 4G-Dual. Also included are tests performed with a model 4G-100 power supply. Communication has only been tested with GPIB, but USB/RS-232 is also supported (but not verified).

I have kept the instruction set to the minimum required to carry out typical measurements using the magnet power supply. There are additional instructions for adjusting shim magnets, if installed. However, I do not have a superconducting magnet with shim magnets with which to test this functionality, so it has been omitted for now.

### DIFF
--- a/docs/api/instruments/cryomagnetics/index.rst
+++ b/docs/api/instruments/cryomagnetics/index.rst
@@ -4,7 +4,7 @@
 Cryomagnetics, Inc.
 ###################
 
-This section contains specific documentation on the Cryomagnetics Ltd power supplies for superconducting magnets that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/indeֱx>`.
+This section contains specific documentation on the Cryomagnetics Ltd power supplies for superconducting magnets that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
 
 .. toctree::
     :maxdepth: 2


tests/instruments/cryomagnetics/test_cm4gmagnetPSU.py
BenediktBurger, line 1: Thanks for adding tests.

pymeasure/instruments/cryomagnetics/cm4GmagnetPSU.py
```diff
@@ -0,0 +1,543 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import strict_discrete_set, truncated_range
```
BenediktBurger, line 26: We discourage the use of truncated validators as they don't give any feedback that your command did not have the intended effect, for example, if you set `voltage= 100`, you think it sets the output to 100 V, but it might be truncated to 50 V.
A strict validator, however, always gives some feedback.

pymeasure/instruments/cryomagnetics/cm4GmagnetPSU.py
```diff
@@ -0,0 +1,543 @@
...
+
+    As an example:
+
+    .. code-block:: python
+
+        # Instantiate the power supply
+        magnet = Cryomagnetics4G100("GPIB::1")
+
+        # Give it a name
+        magnet.name = "my magnet"
```
BenediktBurger, line 131: Why not in the init itself (just an idea)?

Actually, the name is most of the time not needed.

pymeasure/instruments/cryomagnetics/cm4GmagnetPSU.py
```diff
@@ -0,0 +1,543 @@
...
+        "ERROR?",
+        "ERROR %d",
+        """
+        Control whether errors messages are reported over the USB / RS-232 interface,
+        takes values True of False (bool).
+        """,
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
```
BenediktBurger, line 207: For boolean properties, we recommend naming like `usb_error_report_enabled`, that way it is clear what True/False indicate.

pymeasure/instruments/cryomagnetics/cm4GmagnetPSU.py
```diff
@@ -0,0 +1,543 @@
...
+        """
+        Get the magnet current in Amps.
+
+        If the persistent switch heater is **ON**, the magnet current returned will be the same as
+        the power supply output current. If the persistent switch heater is **OFF**, the magnet
+        current will be the value of the power supply output current when the persistent switch
+        heater was last turned off. The magnet current will be set to zero if the power supply
+        detects a quench.
+
+        N.B. The ``units`` will always be switched to Amps when using this command.
```
BenediktBurger, line 232: You can link to the other property with this sphinx syntax:

```suggestion
        N.B. The :attr:`units` will always be switched to Amps when using this command.
```

:attr: links to attributes, :meth:  to a (class) method, :func: to a function (outside the class)...

pymeasure/instruments/cryomagnetics/cm4GmagnetPSU.py
```diff
@@ -0,0 +1,543 @@
...
+        """,
+        preprocess_reply=lambda response: response.replace("A", ""),
+    )
+
+    output_current = Instrument.measurement(
+        "UNITS A;IOUT?",
+        """
+        Get the power supply output current in Amps.
+
+        This may or may not be the same as ``magnet_current`` depending on the state of the
```
BenediktBurger, line 242:
```suggestion
        This may or may not be the same as :attr:`magnet_current` depending on the state of the
```

pymeasure/instruments/cryomagnetics/cm4GmagnetPSU.py
```diff
@@ -0,0 +1,543 @@
...
+    )
+
+    magnet_name = Instrument.control(
+        "NAME?",
+        "NAME %s",
+        """
+        Control the name of the currently selected module on the display.
+        Upper and lower case area accepted; however the string is converted to upper case.
+
+        Must be between 0 and 16 characters in length. Longer names will be truncated.
```
BenediktBurger, line 291: You could add your own validator to check that the length is 16 chars or less, if you wanted to.

pymeasure/instruments/cryomagnetics/cm4GmagnetPSU.py
```diff
@@ -0,0 +1,543 @@
...
+    persistent_switch_heater = Instrument.control(
+        "PSHTR?",
+        "PSHTR %s",
+        """
+        Control the status of the persistent switch heater, either ``'ON'`` or ``'OFF'``.
+        """,
+        validator=strict_discrete_set,
+        values=PERSISTENT_SWITCH_HEATER_STATES,
+        get_process=lambda x: PERSISTENT_SWITCH_HEATER_STATES[int(x)]
+    )
```
BenediktBurger, line 304: Why don't you map it to True/False and rename it to `persistent_switch_heater_enabled`?

pymeasure/instruments/cryomagnetics/cm4GmagnetPSU.py
```diff
@@ -0,0 +1,543 @@
...
+
+    range_1_rate = Instrument.control(
+        "RATE? 0",
+        "RATE 0 %g",
+        """
+        Control the rate, in Amps/second, of Range 1.
+        """,
+        validator=truncated_range,
+        values=[0, _MAXIMUM_CURRENT]
+    )
```
BenediktBurger, line 368: Would it make sense to create "range" channels which have a rate and a boundary?

pymeasure/instruments/cryomagnetics/cm4GmagnetPSU.py
```diff
@@ -0,0 +1,543 @@
...
+class Cryomagnetics4G150s(Cryomagnetics4G100):
+    """
+    Represents the Cryomagnetics model 4G-150s magnet power supply.
+    """
+    _MAXIMUM_CURRENT = 150
+    _MAXIMUM_VOLTAGE = 10
+
+    def __init__(self,
+                 adapter,
+                 name="Cryomagnetics 4G-150s Magnet Power Supply",
```
BenediktBurger, line 490: Do you expect people to use typically several different instruments in the same setup?
If not, you could use a generic name in the parent class and leave remove the inits from all these classes.

PR #1401, opened 2026-02-04
Added new instrument Inficon VGC501

will-fern:
This adds support for the Inficon VGC-501 Pressure Gauge. The implimentation is very minimal, with reading the gauge, instrument information and error status only. In future it should be simple to inherit the VGC501 to VGC50x instruments, which can measure multiple pressures.

### DIFF
--- a/tests/instruments/inficon/test_vgc501.py
+++ b/tests/instruments/inficon/test_vgc501.py
@@ -1,3 +1,26 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2026 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
 
 from pymeasure.test import expected_protocol
 from pymeasure.instruments.inficon import VGC501


pymeasure/instruments/inficon/__init__.py
```diff
@@ -23,3 +23,5 @@
 #
 
 from .sqm160 import SQM160
+
```
BenediktBurger: Typically, for those imports you do not need an empty line.

pymeasure/instruments/inficon/vgc501.py
```diff
@@ -0,0 +1,95 @@
...
+                )
+        else:
+            raise ValueError(
+                f"Expected positive acknowledgement {VGCDef.ACK.value}, received {response}."
+                )
+
+    pressure_1 = Instrument.measurement(
+        "PR1",
+        """Measure the pressure of gauge 1 in mbar.""",
+    )
```
BenediktBurger: As it has only one pressure channel, you might rename it just to `pressure`.

Are there more "channel-related" features (e.g. setting some value per channel)?
In this case, you could think about an actual Channel.
However, as you have only one channel in this instrument, you do not have to do it.

pymeasure/instruments/inficon/vgc501.py
```diff
@@ -0,0 +1,95 @@
...
+
+    information = Instrument.measurement(
+        "AYT",
+        """Measure information about the connected instrument.""",
```
BenediktBurger: You are right, that `"Measure"` is the proper beginning of the docstring.
However, sometimes "Measure" does not sound right, like here, thus, we also allowed "Get" as the initial word for a `measurement`:
"Get the information" or "get the error status" (method below).

tests/instruments/inficon/test_vgc501.py
```diff
@@ -0,0 +1,34 @@
+from pymeasure.test import expected_protocol
```
BenediktBurger, line 25: Please add the copyright notice at the top as well.

.gitignore
```diff
@@ -110,3 +110,6 @@ settings.json
 
 # Vim
 *.swp
+
+#Generatore
+test_vgc501_generator.py
```
BenediktBurger: That file is only short lived, and you can choose to not commit it. Please do not add it to the gitignore.

BenediktBurger (changes requested):

Here are a few comments.

pymeasure/instruments/inficon/vgc501.py
```diff
@@ -0,0 +1,95 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from enum import Enum
+from pymeasure.instruments import Instrument
+
+
+class VGCDef(Enum):
```
BenediktBurger, line 29: You can make it a "byte Enum":

```suggestion
class VGCDef(bytes, Enum):
```

that allows to compare it directly to bytes without using `.value` .

pymeasure/instruments/inficon/vgc501.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+
+    def read(self, **kwargs):
+        """
+        Read from the device and check the response.
+
+        The first read line contains an acknowledgement of the read command.
+        The second line contains the value to be parsed to the measurement.
+        """
+        response = super().read_bytes(3)
```
BenediktBurger, line 67: As you do not overwrite `read_bytes`, you may just call it with self:

```suggestion
        response = self.read_bytes(3)
```

PR #1406, opened 2026-02-18
Add sphinx-autodoc-typehints extension

### LINKED ISSUE #1392: Add sphinx autodoc typehints extension
By the way, there is a [sphinx extension](https://github.com/tox-dev/sphinx-autodoc-typehints) that allows to use type hints and default values from the signature to render parameters in the documentation (see the README for an example). Since type hints are now allowed, maybe it's worth to look into this extension as a docs dependency. This would let to use the docstring for actual description while keeping types and defaults in the signature.

_Originally posted by @ilev-sergey in https://github.com/pymeasure/pymeasure/pull/1380#discussion_r2586438554_

ilev-sergey:
Closes #1392

The PR adds sphinx-autodoc-typehints extension that allows to use type hints and default values from the signature to render parameters in the documentation. So, there is no need to duplicate types and defaults both in signature and docstring.

I also added some optional configuration parameters:
- `typehints_use_rtype = True` unites Return type and Returns sections into a single one which is typically better looking
- `typehints_defaults = "comma"` is used to insert default values into the description section, right after the specified type
- - There is also an option to use `typehints_defaults = "braces-after"` which renders default values in the end of the parameter description. That might be a better option as it is more similar to currently used approach to specifying default values in docstrings.

Here is a quick demo of the changes:

screenshot

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -272,7 +272,6 @@ autodoc_mock_imports = ['zmq', 'cloudpickle', 'pyvirtualbench']
 
 # -- Options for sphinx_autodoc_typehints ---------------------------------
 
-always_document_param_types = True
 typehints_use_rtype = False
 typehints_defaults = "braces-after"
 


ilev-sergey:
In general, I prefer to use the first option in my projects, since you can usually understand how to use a function by skimming through the argument names, types, and defaults. It’s more convenient when they are all in the same place.

However, I’m not sure this is relevant for PyMeasure. Especially for instruments with complex setup options, you need to read the descriptions anyway.

Moreover, the default values for arguments are currently specified at the end of the description. So, if the second option is used, the documentation for typed instruments would have the same style as non-typed instruments, with hardcoded values in the docstring.

I think we can keep the second option for now and change it later if needed. It's one line change anyway.

bilderbuchi:
👍 for less duplication, great if there's now tooling that supports that.

I don't have a strong preference either way of the two discussed options. However, considering that any transition of pymeasure to typehints (and away from putting type info in docstrings; remember that many pymeasure docstrings come from a time before type hinting in Python was common) will be gradual, it's probably better to have a consistent appearance between the two approaches (so from that perspective `braces-after` seems preferable).

docs/conf.py
```diff
@@ -270,6 +270,12 @@
 autodoc_mock_imports = ['zmq', 'cloudpickle', 'pyvirtualbench']
 
 
+# -- Options for sphinx_autodoc_typehints ---------------------------------
+
```
BenediktBurger, line 274: We should add to show the parameter even if it is not given in the docstring.
```suggestion

always_document_param_types = True
```

BenediktBurger (approved):
Looks good to me. Thank you very much!

docs/conf.py
ilev-sergey: It seems there was a bug in the extension with this combination of parameters. It was fixed just recently (release [3.6.3](https://github.com/tox-dev/sphinx-autodoc-typehints/releases/tag/3.6.3), [PR](https://github.com/tox-dev/sphinx-autodoc-typehints/pull/605)), but we won't be able to use it since the current version of the extension supports python>=3.12.

BenediktBurger: Thanks for looking it up, then we leave it out for now.
Let's see where typing will go.

Issue #1404, opened 2026-02-13
Korad KA3000D/KA6000D series PSU


diggit:
I am trying to add support for Korad PSUs. Protocol is mostly ASCII over Virtual COM. Looks like SCPI, but it is not compliant. Unfortunately it is not using delimiters, but rather waits for timeouts (around 18 ms). This also implies the need to implement delay between writes (done).
What is the right approach to consider timeout as a delimiter?

1. create subclass of `SerialAdapter` with this behavior, as `SerialAdapter` already has all necessary functions there
2. override `ask(...)` and use `adapter._read_bytes(...)` directly
3. ???

PR #1407, opened 2026-02-18
Add support for KORAD KA3005P PSU

diggit:
This is a followup to #1404

I've got inspired from other devices and tries to follow their style.
This is my first contribution to pymeasure, so please let me know if something is not right.

Fields where I am not sure:
- channel name: `ch1` or `ch_1`
- formatting
- proper validator usage (strict vs truncated)
- python version compatibility (typing)

## Summary by CodeRabbit

* **New Features**
  * Added support for the Korad KA3005P power supply with voltage and current control and measurement capabilities.
  * Enabled overvoltage and overcurrent protection settings.
  * Added preset management (store and recall functions).
  * Implemented device output and beep controls with version identification retrieval.


diggit:
Also, please let me know preferred way, how to fix those test fails (doc-string issues).

ilev-sergey:

You may add yourself to [AUTHORS.txt](https://github.com/pymeasure/pymeasure/blob/3ad4aa6e6751807506be2e4da568a226d6fcd4cf/AUTHORS.txt).

There are two main reasons why the pipeline fails:
- First of all, you can open the [Files changed](https://github.com/pymeasure/pymeasure/pull/1407/changes) tab and see issues connected to the linting. You can see the documentation for details about [code style](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#python-style-guides). In general, you can install a formatter (e.g. [ruff](https://docs.astral.sh/ruff/) or [black](https://black.readthedocs.io/en/stable/)) to automatically format files on save. Or run the formatter as a command line tool. That will solve most issues.
- Secondly, there are specific [pymeasure conventions](https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings) for property docstrings: start property docstrings with “Control” (read and write values), “Get” (read), or “Set” (write). If you go into the logs of the failing [Python 3.11 job](https://github.com/pymeasure/pymeasure/actions/runs/22150592807/job/64040539393?pr=1407), you can see which properties do not comply with the conventions (see example of an error below).
```
FAILED tests/instruments/test_all_instruments.py::test_property_docstrings[KoradKABase.recall_preset] - AssertionError: 'KoradKABase.recall_preset' docstring does start with 'Recall', not 'Control', 'Measure', 'Get', or 'Set'.
assert 'Recall' in ('Control', 'Measure', 'Set', 'Get')
```

---

> channel name: ch1 or ch_1

I think `ch_1` is preferred. For example, the channel creator uses `ch_` prefix by default.
https://github.com/pymeasure/pymeasure/blob/3ad4aa6e6751807506be2e4da568a226d6fcd4cf/pymeasure/instruments/common_base.py#L204

---

> proper validator usage (strict vs truncated)

I would say that a strict validator should be preferred. It's usually better to raise an error if the value is out of range rather than implicitly use a different value.

By the way, you can write `Closes #XXX` in the PR description to link the PR to corresponding issue. This way, the issue will be closed automatically after the merge of the PR.

tests/instruments/korad/test_ka3005p.py
```diff
@@ -0,0 +1,145 @@
+#
```
ilev-sergey, line 1: Test files must start with `test_` in order to be automatically discovered by pytest. Please rename this file to `test_ka3005p.py`.

pymeasure/instruments/korad/__init__.py
ilev-sergey, line 1: You should add documentation for your instrument. See [this section](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation) for details.

Since this manufacturer is new to pymeasure, please also do the following:
- in `docs/api/instruments/korad/`, add the file for your instrument and an `index.rst` that lists all instruments of Korad manufacturer (for now, it will contain only this instrument)
- Add a reference to `korad/index.rst` in [docs/api/instruments/index.rst](https://github.com/pymeasure/pymeasure/blob/2049dc20577216ad0aee34da081508f771ed44d3/docs/api/instruments/index.rst) so that Korad instruments are listed in the main [instruments index](https://pymeasure.readthedocs.io/en/latest/api/instruments/index.html)

pymeasure/instruments/korad/ka30xxp.py
```diff
@@ -0,0 +1,59 @@
...
+class KoradKA3005Channel(KoradKAChannel):
+    def __init__(self, parent: "KoradKA3005P", channel: int):
+        super().__init__(parent, channel)
+        assert channel in [1, 2], "Channel must be either 1 or 2."
+
+    voltage_setpoint_values = (0, 31.0)
+    current_setpoint_values = (0, 5.1)
+
+
+class KoradKA3005P(KoradKABase):
```
ilev-sergey, line 39: Consider adding a `.. code-block:: python` section with examples of basic usage of the instrument. They'll be rendered and shown on the instrument documentation page.

You can look at other instrument pages for inspiration.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+    # Properties
+    # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    @property
+    def id(self) -> str:
+        """Get the identity of the instrument"""
+        if self.cached_idn:
+            return self.cached_idn
+        self.cached_idn = self.ask("*IDN?")
+        return self.cached_idn
```
ilev-sergey: I'm not sure id caching is a good idea. `.id` is often used to check whether the connection to the instrument is established. In this case, if the connection breaks, you won't be able to diagnose it with `.id`.

Is there a particular reason to cache it?

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+class KoradKAChannel(Channel):
+    def __init__(self, parent: "KoradKABase", id: int):
+        super().__init__(parent, id)
+        assert id in [1, 2], "Channel must be either 1 or 2."
+
+    voltage_setpoint: property = Instrument.control(
+        "VSET{ch}?", "VSET{ch}:%g",
+        """Control the output voltage setpoint.""",
```
ilev-sergey: Please specify units in properties.

```suggestion
        """Control the output voltage setpoint in Volts (float).""",
```

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+        validator=lambda v, vs: strict_range(v, vs),
+        values=(0, 61.0),
+        dynamic=True
+    )
+
+    voltage: property = Instrument.measurement(
+        "VOUT{ch}?",
+        """Measure the actual output voltage."""
```
ilev-sergey:
```suggestion
        """Measure the actual output voltage iv Volts (float)."""
```

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+
+class KoradKAChannel(Channel):
+    def __init__(self, parent: "KoradKABase", id: int):
+        super().__init__(parent, id)
+        assert id in [1, 2], "Channel must be either 1 or 2."
+
+    voltage_setpoint: property = Instrument.control(
+        "VSET{ch}?", "VSET{ch}:%g",
+        """Control the output voltage setpoint.""",
+        validator=lambda v, vs: strict_range(v, vs),
```
ilev-sergey: You can use it like this:
```suggestion
        validator=strict_range,
```

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+
+    # on KA3005P V2.0
+    # - selects the preset slot
+    # - reloads values from the slot and overwrites anything unsaved
+    recall_preset: property = Instrument.setting(
+        "RCL%d",
+        """Set the current settings from preset slot.""",
+        validator=strict_discrete_set,
+        values=(1, 2, 3, 4, 5)
+    )
```
ilev-sergey: I'm not sure that using properties is a good option here.

`recall_preset = 1` isn't very intuitive in my opinion. Maybe it would be better to use methods instead.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+        Valid values are ``True`` to turn output on and ``False`` to turn output off, shutting down
+        any voltage or current.
+        """,
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True
+    )
+
+    # does nothing on KA3005P V2.0
+    beep: property = Instrument.control(
```
ilev-sergey: What about using `_enabled` postfix here as well? This way it's more obvious that it's boolean property
```suggestion
    beep_enabled: property = Instrument.control(
```

pymeasure/instruments/korad/ka30xxp.py
```diff
@@ -0,0 +1,59 @@
...
+#
+
+from pymeasure.adapters.adapter import Adapter
+from pymeasure.instruments.korad.ka_base import KoradKABase, KoradKAChannel
+
+
+class KoradKA3005Channel(KoradKAChannel):
+    def __init__(self, parent: "KoradKA3005P", channel: int):
+        super().__init__(parent, channel)
+        assert channel in [1, 2], "Channel must be either 1 or 2."
```
ilev-sergey: It's better to raise an exception rather than use assertions. Assertions are primarily intended for debug purposes and could be removed when Python is run in optimized mode.

```suggestion
        if channel not in [1, 2]:
            raise ValueError("Channel must be either 1 or 2.")
```

You can also use strict_discrete_set here:

```suggestion
        strict_discrete_set(channel, [1, 2])
```

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+        state = self.parent.state
+        if self.id == 1:
+            return state.ch_1
+        else:
+            return state.ch_2
+
+
+class KoradKABase(Instrument):
+    """Represents a generic Korad KAxxxxP power supply
+    and provides a high-level for interacting with the instrument
```
ilev-sergey:
```suggestion
    and provides a high-level interface for interacting with the instrument
```

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+
+    last_write_timestamp: float  # hold timestamp fo the last write for enforcing write_delay
```
ilev-sergey:
```suggestion
    last_write_timestamp: float  # hold timestamp of the last write for enforcing write_delay
```

pymeasure/instruments/korad/ka30xxp.py
ilev-sergey: Also, this check is already implemented in the parent `KoradKAChannel` class, so there’s no need to repeat it here.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+    def _wait_before_writing(self):
+        actual_write_delay = time.time() - self.last_write_timestamp
+        to_sleep = self.write_delay - actual_write_delay
+        if to_sleep > 0:
+            time.sleep(to_sleep)
+
+    def write(self, command, **kwargs):
+        """Overrides Instrument write method for including write_delay time after the parent call.
+
+    :param command: command string to be sent to the instrument
```
ilev-sergey: Please adjust indents in docstrings in this way.
```suggestion
        :param command: command string to be sent to the instrument
```

ilev-sergey (changes requested):
Thank you for your work.

Here are some comments:

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+
+class Mode(IntEnum):
+    CC = 0
+    CV = 1
+
+
+class Combination(IntEnum):
+    INDEPENDENT = 0
+    TRACKING_SERIES = 1
+    # ?
```
BenediktBurger, line 43: Do you want to indicate a missing "2"?

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+        lock = bool((byte >> 5) & 1)
+        output = bool((byte >> 6) & 1)
+        return cls(ch_1, ch_2, combination, beep, lock, output)
+        # 7 N/A
+
+
+class KoradKAChannel(Channel):
+    def __init__(self, parent: "KoradKABase", id: int):
+        super().__init__(parent, id)
+        assert id in [1, 2], "Channel must be either 1 or 2."
```
BenediktBurger: If you want to raise an error, it is better to do it before you call the super init.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        Control the over-voltage protection.
+
+        Valid values are ``True`` to turn OVP on and ``False`` to turn OVP off.
+        When OVP is on, the PSU will shut down the output
+        if the output voltage exceeds the OVP setpoint.""",
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True
+    )
```
BenediktBurger, line 96: For boolean properties we recommend a naming like `over_voltage_protection_enabled` to make it clear what True/False mean.

diggit:
Thanks for the extensive review and recommendations! I will incorporate all the feedback soon.

pymeasure/instruments/korad/ka_base.py
diggit: Each command take at least 50 ms and version/model/id commands use this. As it is constant, I did not see a reason to ask for it multiple times. I'll remove it.

docs/api/instruments/korad/ka3005p.rst
```diff
@@ -0,0 +1,7 @@
+##########################
+Korad KA3005P Power Supply
+##########################
+
+.. autoclass:: pymeasure.instruments.korad.KoradKA3005P
+    :members:
+    :show-inheritance:
```
ilev-sergey, line 7: It's better to include other classes that you implemented as well, since KoradKA3005P is basically a wrapper around them and doesn't contain the documentation of available methods.

```suggestion
    :show-inheritance:

.. autoclass:: pymeasure.instruments.korad.ka_base.KoradKABase
    :members:
    :show-inheritance:

.. autoclass:: pymeasure.instruments.korad.ka_base.KoradKAChannel
    :members:
    :show-inheritance:
```

pymeasure/instruments/korad/ka30xxp.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class KoradKA3005P(KoradKABase):
+    """Represents a Korad KA3005P power supply.
+
+    Voltage range: 0-31V
+    Current range: 0-5.1A
+
+    This model has various FW versions and some features may not work properly on all versions.
+    (e.g.: preset store/recall on V2.0)
```
ilev-sergey, line 46: Some small adjustments for better readability:


```suggestion
    - Voltage range: 0 - 31 V
    - Current range: 0 - 5.1 A

    .. warning::

        This model has various FW versions and some features may not work properly on all versions.
        (e.g.: preset store/recall on V2.0)
```

pymeasure/instruments/korad/ka30xxp.py
```diff
@@ -0,0 +1,76 @@
...
+
+        psu.ch_1.current_setpoint  = 0.5  # set current setpoint to 0.5 A
+        psu.ch_1.voltage_setpoint  = 5.0  # set voltage setpoint to 5 V
+        psu.output_enabled = True  # enable output
+
+        psu.ch_1.over_current = True # enable output auto-disable on overcurrent
+
+        print(psu.ch_1.voltage)  # measure output voltage
+        print(psu.ch_1.current)  # measure output current
+        print(psu.ch_1.mode)  # get current mode of the channel
```
ilev-sergey, line 62:
```suggestion
        psu.output_enabled = True         # enable output

        psu.ch_1.over_current = True  # enable output auto-disable on overcurrent

        print(psu.ch_1.voltage)  # measure output voltage
        print(psu.ch_1.current)  # measure output current
        print(psu.ch_1.mode)     # get current mode of the channel
```

pymeasure/instruments/korad/ka30xxp.py
```diff
@@ -0,0 +1,76 @@
...
+    ch_1: KoradKA3005Channel
+
+    def __init__(self, adapter: Adapter, name: str = "KA3005P", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            **kwargs
+        )
+        self.ch_1 = KoradKA3005Channel(self, 1)
+        # assert "KORADKA3005P" in self.id
```
ilev-sergey, line 76: Remove the comment if this is not needed.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+
+    # KA3005P V2.0 adds "K" suffix
+    current_setpoint: property = Instrument.control(
+        "ISET{ch}?", "ISET{ch}:%g",
+        """Control the output current setpoint in Amperes (float).""",
+        validator=lambda v, vs: strict_range(v, vs),
```
ilev-sergey, line 102:
```suggestion
        validator=strict_range,
```

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True
+    )
+
+    def store_preset(self, slot: int):
+        """Store the current settings into preset slot.
+
+        on KA3005P V2.0
+          - works only when the preset is currently selected (caused by recall_preset)
```
ilev-sergey, line 219:
```suggestion
          - works only when the preset is currently selected (caused by :meth:`recall_preset`)
```

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+        :param slot: preset slot number (1-5)
+        """
+        _ = strict_discrete_set(slot, (1, 2, 3, 4, 5))
+        _ = self.write(f"RCL{slot}")
+
+    @property
+    def state(self) -> State:
+        """Get the current state of the system
+
+        :return State.
```
ilev-sergey, line 243:
```suggestion
        :return: State.
```

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+        got = self.binary_values("STATUS?", dtype=np.uint8)
+        if len(got) != 1:
+            raise ValueError(f"Expected 1 byte from STATUS? query, got {len(got)} bytes.")
+        return State.from_byte(got[0])
+
+    # at least KA3005P V2.0 does not use traditionally delimiters for requests/responses,
+    # but relies on timeouts. It was observed that PSU starts sending reply
+    # ~18 ms after the end of the command (if the command is query).
+    # To make sure all commands are registered by the PSU,
+    # 50 ms pause between writes proved to be sufficient.
```
ilev-sergey, line 254: This note should be somewhere in the docs to be visible outside of the code as well. You can include it into write/ask or the class documentation.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+        """Store the current settings into preset slot.
+
+        on KA3005P V2.0
+          - works only when the preset is currently selected (caused by recall_preset)
+          - saves values as preset
+
+        :param slot: preset slot number (1-5)
+        """
+        _ = strict_discrete_set(slot, (1, 2, 3, 4, 5))
+        _ = self.write(f"SAV{slot}")
```
ilev-sergey, line 225: Why is the assignment needed here? Especially for write, since it doesn’t return anything.

tests/instruments/korad/test_ka3005p.py
```diff
@@ -0,0 +1,165 @@
...
+def test_recall_preset():
+    """Verify the recall preset functionality."""
+    with expected_protocol(
+        KoradKA3005P,
+        [("RCL4", None),
+         ],
+    ) as inst:
+        inst.recall_preset(4)
+
+# TODO: it would be nice to also validate write timing
```
ilev-sergey, line 165: You can add some [device tests](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#device-tests) to check that timings work fine.

pymeasure/instruments/korad/ka30xxp.py
```diff
@@ -0,0 +1,76 @@
...
+    write_delay: float  # minimum time between writes
+    ch_1: KoradKA3005Channel
+
+    def __init__(self, adapter: Adapter, name: str = "KA3005P", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            **kwargs
+        )
+        self.ch_1 = KoradKA3005Channel(self, 1)
```
ilev-sergey, line 75: You should probably still use ChannelCreator for adding the channel. I don't think recall having the problems you mentioned, although it was a while ago and I might not have had type checking configured at the time. Maybe this will be fixed after the merge of PR #1331

ilev-sergey (commented):
Thanks for your fixes.

Here are few more small suggestions.

pymeasure/instruments/korad/ka_base.py
diggit: I can move it, but IMO it is not strictly important to the pymeasure user, as it is handled internally...

pymeasure/instruments/korad/ka_base.py
diggit: basedpyright complained in both cases, but I can definitely remove it

ilev-sergey: Let's keep it then.

pymeasure/instruments/korad/ka_base.py
ilev-sergey: I see your point. My thinking was that this behavior might not be obvious when someone is reading the code without having a deep knowledge about the instrument. But I agree it’s not strictly required for user to know if everything is handled internally. I’m okay leaving it as is.

pymeasure/instruments/korad/ka30xxp.py
```diff
@@ -0,0 +1,76 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from pymeasure.adapters.adapter import Adapter
+from pymeasure.instruments.korad.ka_base import KoradKABase, KoradKAChannel
+
+
+class KoradKA3005Channel(KoradKAChannel):
+    def __init__(self, parent: "KoradKA3005P", channel: int):
```
BenediktBurger, line 30: Theoretically, the parent could be any `Instrument`, but you can keep it.

pymeasure/instruments/korad/ka30xxp.py
```diff
@@ -0,0 +1,76 @@
...
+        super().__init__(parent, channel)
+        if channel != 1:
+            raise ValueError("KoradKA3005P only supports channel 1.")
```
BenediktBurger, line 33: You should do the parameter validation before calling super init.

pymeasure/instruments/korad/ka30xxp.py
```diff
@@ -0,0 +1,76 @@
...
+        print(psu.ch_1.voltage)  # measure output voltage
+        print(psu.ch_1.current)  # measure output current
+        print(psu.ch_1.mode)  # get current mode of the channel
+    """
+
+    last_write_timestamp: float  # hold timestamp fo the last write for enforcing write_delay
+    write_delay: float  # minimum time between writes
+    ch_1: KoradKA3005Channel
+
+    def __init__(self, adapter: Adapter, name: str = "KA3005P", **kwargs):
```
BenediktBurger, line 69: It's not just `Adapter` which you can give, but also a string or int, see in #1331 for `Instrument`.

pymeasure/instruments/korad/ka30xxp.py
BenediktBurger: Exactly, you should use the ChannelCreator.
I know, typing with it does not work yet.
We chose this approach of the ChannelCreator that you can inspect a class (not an instance). Also it gives some uniformity to the codebase.

You (@diggit ) are free to figure out a way to make typing work with ChannelCreator and open a PR. Unfortunately, I'm quite busy doing reviews at the moment.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+        When OVP is on, the PSU will shut down the output
+        if the output voltage exceeds the OVP setpoint.""",
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True
+    )
+
+    # KA3005P V2.0 adds "K" suffix
+    current_setpoint: property = Instrument.control(
+        "ISET{ch}?", "ISET{ch}:%g",
```
BenediktBurger, line 100: With a dynamic property, you can even change the commands as well adding that suffix "K".

Sou you could do it dynamically.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+
+class KoradKABase(Instrument):
+    """Represents a generic Korad KAxxxxP power supply
+    and provides a high-level interface for interacting with the instrument
+    """
+
+    last_write_timestamp: float  # hold timestamp of the last write for enforcing write_delay
+    write_delay: float  # minimum time between writes
+
+    def __init__(self, adapter: Adapter, name: str = "Korad KA base", **kwargs):
```
BenediktBurger, line 149: See base instrument: You can give a string as adapter.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+        if not isinstance(self.adapter, SerialAdapter):
+            self.adapter.log.warning(
+                "adapter should be instance of SerialAdapter,assuming test environment"
+            )
+        else:
+            self.adapter.read_termination = ""
+            self.adapter.write_termination = ""
+            self.adapter.connection.timeout = 0.05
+            self.adapter.connection.baudrate = 9600
+        self.last_write_timestamp = 0.0
```
BenediktBurger, line 165: Why do you enforce a SerialAdapter?

Our default is the VisaAdapter which will be also created, if you just give a string to the instrument. This is the default way to connect (no manual adapter creation).
In order to configure the VisaAdapter, you give these parameters to the super init call: read_termination = ""....
Note though that the timeout of the VisaAdapter is milliseconds (so 50 in your case).

BenediktBurger, line 165: why don't you use that as a starting value of the class?

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+        self.write_delay = 0.05
+
+    # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    # Properties
+    # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    @property
+    def id(self) -> str:
+        """Get the identity of the instrument"""
+        return self.ask("*IDN?")
```
BenediktBurger, line 175: You could also use the property creators `Instrument.measurement` to do exactly that. Yes, they do a little more, but essentially they just ask.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+
+        Valid values are ``True`` to turn output on and ``False`` to turn output off, shutting down
+        any voltage or current.
+        """,
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True
+    )
+
+    # does nothing on KA3005P V2.0
```
BenediktBurger, line 203: This should be in the docstring to warn users.

pymeasure/instruments/korad/ka_base.py
BenediktBurger: This `:meth:` syntax will create a link to that method on the generated API on the website.

pymeasure/instruments/korad/ka_base.py
BenediktBurger: We never used that syntax in pymeasure, I think.
However, you are right, that way you make clear that you get something and we do not use it.

pymeasure/instruments/korad/ka_base.py
BenediktBurger: That's a duplication of the type hint. If you have a type hint, no need to add it in the docstring (actually bad as they might diverge in the future).

pymeasure/instruments/korad/ka_base.py
BenediktBurger: Generally, it should be at least in the ask or write method to indicate why it behaves as it does.
You could even argue for the class docstring as it describes to the user why there is a delay between commands.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+    # 50 ms pause between writes proved to be sufficient.
+
+    def _wait_before_writing(self):
+        actual_write_delay = time.monotonic() - self.last_write_timestamp
+        to_sleep = self.write_delay - actual_write_delay
+        if to_sleep > 0:
+            time.sleep(to_sleep)
+
+    def write(self, command, **kwargs):
+        """Overrides Instrument write method for including write_delay time after the parent call.
```
BenediktBurger, line 263: Docstrings should be imperative what the method should do.


```suggestion
        """Write to the device with proper time between consecutive write operations.
```

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+        self.adapter.connection.timeout = 0.1
+        try:
+            first = self.read_bytes(1, break_on_termchar=False)
+        finally:
+            self.adapter.connection.timeout = timeout_backup
+        if len(first) == 0:
+            raise TimeoutError("No response received from instrument.")
+        return (first + self.read_bytes(-1, break_on_termchar=False)).decode(errors="ignore")
+        # if inter_byte_timeout woudl work, it would be easy
+        # return self.read_bytes(-1, break_on_termchar=False).decode(errors="ignore")
```
BenediktBurger, line 301: Here you have to either implement it or remove the comment.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+                "ask method may not work correctly if adapter"
+                " is not instance of SerialAdapter, assuming test environment"
+            )
+            return super().ask(command, query_delay=query_delay)
+
+        self.write(command)
+        if query_delay is not None:
+            time.sleep(query_delay)
+        # override timeout until first character arrives
+        # after that, restore the timeout and read until that
```
BenediktBurger, line 288: Comments should state the why, not the what. You can read from code that you override the timeout temporarily.

pymeasure/instruments/korad/ka_base.py
```diff
@@ -0,0 +1,301 @@
...
+        super().write(command, **kwargs)
+        self.last_write_timestamp = time.monotonic()
+
+    def ask(self, command, query_delay=None):
+        """Overrides Instrument ask method for including write_delay time after the parent call.
+
+        :param command: command string to be sent to the instrument
+        :returns: response from the instrument
+        """
+        if not isinstance(self.adapter, SerialAdapter):
```
BenediktBurger, line 277: For tests, you can check for the ProtocolAdapter (used by tests).

tests/instruments/korad/test_ka3005p.py
BenediktBurger: Yes, TODO comments should be resolved or removed.

Actually, the write method uses the timing and you can use the ProtocolAdapter and do some tests with it.
Please make sure that you do not have too long time waiting in these tests (all together), as it will increase the overall test duration.

BenediktBurger (changes requested):
Thank you for your contribution. And welcome to the authors, @diggit .

A few requests also from my side.

Don't get discouraged from the comments. It won't take forever with these comments.

@ilev-sergey , thank you very much for the review over several rounds. You already addressed many issues which I did not have to point out myself.

diggit:
Hi,
Some other project landed on my shoulders and got highest priority. I am not sure when I will be able to get back to this topic.
Anyway, most of requested changes are about formatting/docs, so anyone can take over this PR.

BenediktBurger:
Thanks for letting know. Feel free to continue when you have time.

PR #1415, opened 2026-03-20
Replace load_module() by exec_module()

Falk3579:
Method "load_module()" is deprecated and will be removed in Python 3.15. It is replaced by "exec_module" method.
Affected files:
 - experiment/procedure.py
 - experiment/results.py

The code modification was supported by Mistral.ai

### DIFF
--- a/pymeasure/experiment/procedure.py
+++ b/pymeasure/experiment/procedure.py
@@ -26,7 +26,7 @@ import logging
 import sys
 import inspect
 from copy import deepcopy
-from importlib.machinery import SourceFileLoader
+import importlib.util
 import re
 from pint import UndefinedUnitError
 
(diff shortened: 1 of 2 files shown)


BenediktBurger:
I guess it works for all supported python versions. And older ones?

Falk3579:
This new version of loading and executing modules was introduced in Python version 3.4. So it will work with all the supported Python versions.

CasperSchippers (approved):
I guess it looks good to me.

Only thing I am unsure of is where this code is actually used; @BenediktBurger, would you happen to know? If not I am fine with merging regardless, since the code basically does the same as the old version, as far as I can tell.

BenediktBurger:
I don't know whether it is used.

So, I merge it.

PR #1416, opened 2026-03-27
Smaract ascii

WiktorJJ:
This PR adds a new driver for the Smaract SCU controller when it is configured with ascii communication protocol.

The code should encompass either step, linear or angular channels but only linear ones could be tested.

@seb5g is our tutor on this project.

Wiktor Jankowski and Joan Solé-Lohan as authors under the github WiktorJJ account


pymeasure/instruments/smaract/__init__.py
BenediktBurger, line 1: Please add also documentation files, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
+from typing import Union
```
BenediktBurger, line 1: Please add to this (and all other files) the copyright notices at the top.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+    unit: str = ''
+
+    frequency_max = Instrument.control(
+        ":GCLF{ch}",
+        ":SCLF{ch}F%d",
+        """Control the maximum frequency in an absolute move""",
+        set_process=lambda v: check_quantity_unit(v, 'Hz'),
+        get_process=lambda s: Q_(int(s[6:]), 'Hz'),
+    )
+    safe_direction = Instrument.control(
```
BenediktBurger, line 26: Please keep empty lines between properties:
```suggestion
    )

    safe_direction = Instrument.control(
```

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+
+
+# <channel> : zero-based channel index. Valid indices are 0,1 and 2
+# <angel> : the current angle of the position in millidegree
+
+
+class SCUChannel(Channel):
+    unit: str = ''
+
+    frequency_max = Instrument.control(
```
BenediktBurger, line 19: As it is in the Channel, use `Channel.control`:

```suggestion
    frequency_max = Channel.control(
```

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        ":GCLF{ch}",
+        ":SCLF{ch}F%d",
+        """Control the maximum frequency in an absolute move""",
```
BenediktBurger, line 22: In which unit?

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        set_process=lambda v: check_quantity_unit(v, 'Hz'),
+        get_process=lambda s: Q_(int(s[6:]), 'Hz'),
+    )
+    safe_direction = Instrument.control(
+        ":GSD{ch}",
+        ":SSD{ch}D%d",
+        """Control the safe direction for a given channel for move_to_ref
+           :param  must be either 0 (down) or 1 (up). In minor letters""",
```
BenediktBurger, line 30: I suggest to make it a boolean variable: `safe_direction_up_enabled`, so you can set it to True (up) or False (down), or use +1/-1 for the direction as an int (or any positive/negative number).

Regarding the docstring: You do not specify the "for a given channel", as this is a method of a certain channel.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        validator=strict_discrete_set,
+        map_values=True,
+        values={'up': 1, 'down': 0},
+        get_process=lambda x: int(x[-1]),
+    )
+
+    def calibrate_sensor(self):
+        """Calibrate the sensor. Has to be done before the use of move to reference.
```
BenediktBurger, line 38: You can link to another method with :meth:`move_to_ref` (at least in the API documentation):

```suggestion
        """Calibrate the sensor. Has to be done before the use of :meth:`move_to_ref`.
```

Similarly, :class: links to a class, :func: to a function (outside a class) etc.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+           The user must ensure himself that the positioner is not close to a mechanical
+           limit before using this method"""
+        self.write(f':CS{self.id}')
+        status = self.ask(f":M{self.id}")
+        return status
```
BenediktBurger, line 43: What does the return value mean?

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def set_zero_pos(self):
```
BenediktBurger, line 45: Typically, it is better to have a longer, but more descriptive name like `set_zero_position`.
Autocompletion makes it easy to write this longer name but reading it will be a lot better.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        """Set the current position as position zero."""
+        if self.check_sensor_present():
+            self.write(f":SZ{self.id}")
+        else:
+            raise NotImplementedError
```
BenediktBurger, line 50: Why not ImplementedErrror, do you want to do it later?

I guess a ValueError (with a meaningful message) would be better.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def set_positioner_type(self, t: int):
+        """Set the positioner/sensor type.
+
+           CAUTION! The user has to use the 'SmarAct ASCII Programming Interface' for a
+           thorough knowledge of the differnet modes. It ensures the user, that the desired type
+           exists, and has the possibility to be implemented to the program.
+
+           The list 'positioner_types' contains the available t values and corresponding properties
```
BenediktBurger, line 59: This list is not available outside the method, so it is better to add this as a comment to the dictionary inside the method.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+
+        """
+        positioner_types = {
```
BenediktBurger, line 62: You could use also an [Enum](https://docs.python.org/3/library/enum.html), if you want to.

With an Enum, you could accept that instead of an arbitrary integer.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+            36: "Iris – SI…S1L1S",
+            37: "Tip-Tilt – ST…S1I2E2",
+            38: "Rotary – SR…T5L3S",
+            39: "Iris – SI…S1L4E",
+            40: "Iris – SI…S1L1E",
+        }
+        if t in positioner_types:
+            self.write(f":SST{self.id}T{t}")
+        else:
+            raise ValueError
```
BenediktBurger, line 105: please add a meaningful message, e.g. `f"'{t}' is not a valid positioner type. It has to be an integer between 0 and 40."`

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def get_positioner_type(self):
+        """Get the positioner/sensor type, could be linear or angular"""
+        self.write(f":GST{self.id}")
+        return self.read()
```
BenediktBurger, line 110: do you return an integer?

Wouldn't it better to return an enum (see above)?

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def move_steps_up(self, steps: int,
+                      freq: Union[int, float, Q_] = None,
+                      ampl: Union[int, float, Q_] = None):
```
BenediktBurger, line 114: you say that freq and ampl have to be int, float, or Q_, but you assign None, which is invalid!

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        """Moves up
+        :param steps: Number of steps, an int in range [1;30000]
+        :param freq: Frequency, a quantity given in Hz in range [1;18500] (or int)
+        :param ampl: Amplitude, a quantity given in dV in range [150;1000] (or int)
+        """
```
BenediktBurger, line 119:
```suggestion
        """Move up.

        :param steps: Number of steps, an int in range [1;30000]
        :param freq: Frequency, a quantity given in Hz in range [1;18500] (or int)
        :param ampl: Amplitude, a quantity given in dV in range [150;1000] (or int)
```

A few hints:
- docstrings should have a first sentence (single line) in imperative mode giving a short summary.
- After an empty line, you can have a longer description, more context etc.
- Afterwards the parameters etc.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        a_val = int(valid_ampl.to('dV').magnitude)
+
+        self.write(f":D{self.id}F{f_val}A{a_val}S{steps}")
+        # self._current_steps -= steps
+
+    def check_sensor_present(self) -> bool:
+        """Check if the sensor is present
+
+        :returns: True/False
+
```
BenediktBurger, line 152: That you return True/False is nothing new, as you declared the return type as a boolean.
Better:


```suggestion
        """Check whether the sensor is present.
```

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        self.write(f':GSP{self.id}')
+        response = self.read()
+        return response == f':SP{self.id}P'
+
+    def move_to_ref(self):
+        """Moves up/down to reference"""
+        self.write(f":MTR{self.id}H0Z1")
+        self._current_steps = 0
+
+    def move_to_end_up(self):
```
BenediktBurger, line 163: Maybe `move_up_to_end`?

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        return status
+
+    def set_zero_pos(self):
+        """Set the current position as position zero."""
+        if self.check_sensor_present():
+            self.write(f":SZ{self.id}")
+        else:
+            raise NotImplementedError
+
+    def set_positioner_type(self, t: int):
```
BenediktBurger, line 52: instead of those getter/setter methods, it is better to use properties:
- either via Instrument/Channel.control/setting/measurement
- or via plain python `@property` annotation.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+    def move_to_end_down(self):
+        """Moves down until end of line"""
+        self.write(f":MES{self.id}DD")
+        self._current_steps = 20000
+
+    def stop(self):
+        """Stops any process."""
+        self.write(f":S{self.id}")
+
+    def get_position(self):
```
BenediktBurger, line 177: this should be a property (`position`) and not a method, as it represents a value.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def move_abs(self, position: Q_):
+        """Moves to the absolute angle given in m° from the reference position via
+           closed-loop control.
+
+           :param angle: A quantity with angular units (m°)
+        """
+        self.write(f":MAA{self.id}P{check_quantity_unit(position, self.unit)}")
+
+    def get_angle(self):
```
BenediktBurger, line 211: This should be a property (`angle`).

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+            self.move_steps_up(steps_val)
+            self._current_steps = steps_val + old_steps
+        elif steps_val < 0:
+            self.move_steps_down(abs(steps_val))
+            self._current_steps = steps_val + old_steps
+
+    def move_abs(self, position: Union[int, Q_]):
+        """Moves to the absolute position given in steps from the reference possition
+
+        :param steps : A quantity with the step as unit, given as integer
```
BenediktBurger, line 264: This parameter definition  is already contained in the summary line, so it is superfluous.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+    for driving piezo stages, with optional support for closed-loop
+    positioning using position or angle sensors.
+
+    SCU controllers communicate via USB or RS232. Program uses a simple ASCII command
+    protocol, which is implemented by this driver.
+
+    This base class provides the common communication layer and shared
+    functionality, while axis-specific behavior (e.g. linear or angular motion)
+    is implemented in subclasses.
+
```
BenediktBurger, line 291: Could you actually mix and match, e.g. one angular and one linear channel?

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+    :param str adapter: Name of the COM-port.
+    :param int timeout: Timeout in ms.
+    """
+
+    channel0 = Instrument.ChannelCreator(SCUChannel, "0")
+    channel1 = Instrument.ChannelCreator(SCUChannel, "1")
+    channel2 = Instrument.ChannelCreator(SCUChannel, "2")
+
+    def __init__(self, adapter, name='SCUController',
+                 includeSCPI=False, **kwargs):
```
BenediktBurger, line 301: You do not need to add includeSCPI in the signature of the __init__ method (but below where you call super, it is needed).

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        ":I", """Get the device model information. """,
+        get_process=lambda s: s[2:]
+    )
+
+    # CHECK AMPLITUDE
+    def check_amplitude(self, ampl: Union[int, str, Q_] = None) -> Q_:
+        """Check if amplitude is present and if it is inside the given boundary.
+
+               :param ampl : a quantity with amplitude units dV, if int then given in dV
+               """
```
BenediktBurger, line 330: This is indented too much

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+    )
+
+    model = Instrument.measurement(
+        ":I", """Get the device model information. """,
+        get_process=lambda s: s[2:]
+    )
+
+    # CHECK AMPLITUDE
+    def check_amplitude(self, ampl: Union[int, str, Q_] = None) -> Q_:
+        """Check if amplitude is present and if it is inside the given boundary.
```
BenediktBurger, line 327: Maybe voltage amplitude instead of just amplitude?

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+
+               :param ampl : a quantity with amplitude units dV, if int then given in dV
```
BenediktBurger, line 329: I'd say the unit of voltage, even amplitude, is just Volts (or abbreviated V), not dV.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        """ Control default frequency """
+        return self._amplitude
+
+    @amplitude.setter
+    def amplitude(self, value: Union[int, Q_]):
+        self._amplitude = self.check_amplitude(value)
+
+    # CHECK FREQUENCY
+
+    def check_freq(self, freq: Union[int, str, Q_] = None) -> Q_:
```
BenediktBurger, line 355: better `check_frequency` and `frequency` instead of `freq`.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        self._freq = self.check_freq(value)
+
+    # CHECK STEPS
+
+    def check_steps(self, steps: int):
+        """Check if steps are present and if it is inside the given boundary."""
+        if steps is None:
+            steps = self._steps
+        else:
+            self._steps = steps
```
BenediktBurger, line 391: You should mention that you store the steps value and not only check it. This is an unexpected side-effect (not only checking, but also changing things)!

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        return steps
+
+    # MOVEMENT
+
+    def move_abs(self, position: Union[Q_, int]):
+        """Moves to the absolute position given in µm from the reference possition
+
+        :param position : A quantity with length units (µm), if integer, then given in µm
+        """
+        raise NotImplementedError
```
BenediktBurger, line 403: In order to overwrite it in a subclass, right?

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        """ Control default frequency """
+        return self._freq
+
+    @frequency.setter
+    def frequency(self, value: Union[int, Q_]):
+        self._freq = self.check_freq(value)
+
+    # CHECK STEPS
+
+    def check_steps(self, steps: int):
```
BenediktBurger, line 386: I see that you use type hints for the input parameters, why not also for the return values?

```suggestion
    def check_steps(self, steps: int) -> int:
```

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
...
+        """
+        if baudrate in self.BAUDRATE:
+            self.write(f':CB{baudrate}')
+            self.write(':R')
+        else:
+            raise pyvisa.VisaIOError(-1)
+
+    def reset(self):
+        """Reset will be performed on the device. Has the same effect
+           as a power-down/power-up cycle
```
BenediktBurger, line 432: A docstring should be in imperative mode: Reset the device.

pymeasure/instruments/smaract/utils.py
```diff
@@ -0,0 +1,21 @@
...
+
+def check_quantity_unit(input: Union[str, int, Q_], unit: str) -> int:
+    if isinstance(input, str):
+        input = Q_(input)
+    if isinstance(input, Q_):
+        return int(input.to(unit).magnitude)
+    elif isinstance(input, int):
+        return input
+    else:
+        raise TypeError('Invalid input for the instrument')
```
BenediktBurger, line 21: You mean you convert some input to the magnitude in a given unit?
Consider a more meaningful name, like `convert_quantity_to_magnitude`.

Actually, this might be useful for other instruments as well, such that we should put it into the `units.py` module.
But with int/float and not only int (or just float)

pymeasure/generator.py
```diff
@@ -434,7 +434,11 @@ def parse_stream(self):
         self._index = self._stream.tell()
         return self._init_comm_pairs + comm
 
-    def instantiate(self, instrument_class, adapter, manufacturer, adapter_kwargs=None, **kwargs):
+    def instantiate(self, instrument_class: object,
+                    adapter: object,
+                    manufacturer: object,
+                    adapter_kwargs: object = None,
+                    **kwargs: object) -> TestInstrument:
```
BenediktBurger, line 441: Just adding `object` does not give us any advantages. Type hints should be meaningful.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,454 @@
+from typing import Union
+from pint import Quantity as Q_
```
BenediktBurger, line 2: Please use the unit registry `ureg` from `units.py` such that whole pymeasure uses the same unit registry.
You could define Q_ = ureg.Quantity if you wanted to.

pymeasure/instruments/smaract/utils.py
```diff
@@ -0,0 +1,21 @@
+from typing import Union
+
+from pint import Quantity as Q_
```
BenediktBurger, line 3: also here, please use the ureg from units.py.

tests/instruments/smaract/generator_protocol_smaract.py
BenediktBurger, line 1: Typically, we do not keep these files.
However, you could convert it to a test with device, see https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#device-tests, just add a few sensible assertions.

tests/instruments/smaract/test_scu_ascii_protocol.py
```diff
@@ -0,0 +1,208 @@
...
+    ):
+        pass  # verify the expected communication.
+
+
+def test_close():
+    with expected_protocol(
+            SmarActSCULinear,
+            [],  # no communication expected
+    ) as inst:
+        inst.close()
```
BenediktBurger, line 24: shouldn't you verify that it does what you expect it to do (closing the adapter)?

tests/instruments/smaract/test_scu_ascii_protocol.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@pytest.mark.parametrize("channel", CHANNELS)
```
BenediktBurger, line 27: why do you parametrize with only a single parameter?
Wouldn't it better to just use some hard-coded values (sometimes one channel, sometimes another one)?

tests/instruments/smaract/test_scu_ascii_protocol.py
BenediktBurger, line 1: The protocol tests do not contain `protocol` in the name.

tests/instruments/smaract/test_scu_ascii_with_device.py
```diff
@@ -0,0 +1,223 @@
...
+
+# Importing driver classes
+from pymeasure.instruments.smaract.scu_ascii import SmarActSCULinear, SmarActSCUStepper
+
+
+CHANNELS = ['0']
+# Here you may add multiple channels ['0', '1', '2', ...]
+SENSOR = [True]
+# TYPE = ['Linear'] hesitating on adding this function, since it may
+# not be as simple as imagined, and can be easily checked by SENSOR
```
BenediktBurger, line 13: Please finish it or remove the unfinished parts.

tests/instruments/test_all_instruments.py
```diff
@@ -40,8 +40,8 @@ def find_devices_in_module(module):
     base_import = module.__package__ + "."
     for inst_file in Path(base_dir).rglob("*.py"):
         relative_path = inst_file.relative_to(base_dir)
-        if inst_file == "__init__.py":
-            # import parent module when filename __init__.py
+        if inst_file == "test_smaract_protocol.py":
+            # import parent module when filename test_smaract_protocol.py
```
BenediktBurger, line 44: What do you try to achieve here?

generated_test_smaract.py
BenediktBurger, line 1: What do you try to test with this file?

BenediktBurger (changes requested):

You may add yourself (as part of this PR) to the authors.txt file if you wish so.

Here are a few requests for changes. If something is unclear, don't hesitate to ask.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Thank you for the suggestion. I had it implemented as well as for the docstring I added :meth: move_to_ref as advised in another comment.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Added :param definition of unit in Herz

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Done for both frequency_max and safe_direction

pymeasure/instruments/smaract/__init__.py
WiktorJJ: I based myself on the way the redpitaya library did it, and I believe it should be correct now.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: It returns a status, which in this case is the letter 'C' for 'calibrating'. We use this in the test_protocol in order to verify that a calibration is happening

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Acknowledged and added, thank you.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Added the full phrase for readability

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Good point. I added a ValueError with the indication that a sensor is missing.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: I apologise, but I do not understand what you mean with 'accepting of an arbitrary integer'. The numbers assigned to each mode come from the respective values given in the datasheet of SmarAct.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Not quite sure what you mean with this. I am to take the list out of the method and put it where exactly?

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Done

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: We figured that for the the user can use the list that is already defined inside 'set_positioner_type'

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Changed it to move_up_to_end and move_down_to_end. Changed it as well on the two test files

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: I implemented it via plain python now with @property

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Deleted it now.

pymeasure/instruments/smaract/__init__.py
BenediktBurger: There are also some reStructured Text (rst) files for the api documentation, see: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/instrument.html#adding-documentation

pymeasure/instruments/smaract/scu_ascii.py
BenediktBurger: Maybe add it to the docstring, that it returns "C" during calibration, or even, you could return `status == "C"`, so it returns True during calibration or something else.

pymeasure/instruments/smaract/scu_ascii.py
BenediktBurger: I just made you aware, that nobody could use the list programatically. You could put it (if you deem it useful) into the class or even outside the class just in this file, depending where you want it to be used.

pymeasure/instruments/smaract/scu_ascii.py
BenediktBurger: Yes, currently, the **method** accepts any integer and will fail later, if not valid.
With an enum, the accepted type (at least with type hints enabled) is just that enum.

pymeasure/instruments/smaract/scu_ascii.py
BenediktBurger: but the user cannot use that list in his own code, because he cannot access that list (it only lives in this method), so you have to make it a class list (defined outside this method).

WiktorJJ: Made the list global now. Thank you for making it clear.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Added status == 'M0C' which now returns true if the calibration is underway.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Got it now. I have it now in the class.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Made it a property

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Now its only in the super

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Corrected now.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: I added it to the docstring.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: It was given as dV by the supplier of the controler. We decided not to change this parameter.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: changed all check_frequency and frequency in #CHECK FREQUENCY

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Mention it in the docstring. As well as mentioning the fact that the old value stored will be used if None is given.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Correct.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Done.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Good point. Now its coherent, being in check_frequency, check_amplitude and check_steps

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Understood.

pymeasure/instruments/smaract/utils.py
WiktorJJ: Yes, it converts the inputs into a magnitude. Good name suggestion. I changed it now so that it hints to Union[int, float] as the return value. But I cast that if an int is given in, then an int will come out. Would you like for me to implement it into `units.py` module?

pymeasure/generator.py
WiktorJJ: I am not sure why this is a comment, since its ot a file we have contributed to the PyMeasure library. Maybe I have missunderstood the case though.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Changed it for scu_ascii, test with device and the protocol tests

pymeasure/instruments/smaract/utils.py
WiktorJJ: done

tests/instruments/smaract/generator_protocol_smaract.py
WiktorJJ: Deleted the file.

tests/instruments/smaract/test_scu_ascii.py
```diff
@@ -0,0 +1,208 @@
...
+    ):
+        pass  # verify the expected communication.
+
+
+def test_close():
+    with expected_protocol(
+            SmarActSCULinear,
+            [],  # no communication expected
+    ) as inst:
+        inst.close()
```
WiktorJJ: Verifying now with a command sent, and expected a None as return. Kind of like in test_with_device

tests/instruments/smaract/test_scu_ascii.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@pytest.mark.parametrize("channel", CHANNELS)
```
WiktorJJ: It was conceived this way, in order to stay away from hard-coding. In case the user wants to check if all three channels work simultaneously, he now has the possibility.

tests/instruments/smaract/test_scu_ascii.py
WiktorJJ: removed

tests/instruments/smaract/test_scu_ascii_with_device.py
WiktorJJ: removed

tests/instruments/test_all_instruments.py
WiktorJJ: Not sure what it is. Neither something that we have implemented.

generated_test_smaract.py
WiktorJJ: deleted the file. I had it deleted earlier but did not delete it from path

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Added to all five files. `scu_ascii, __init__, utils, test_scu_ascii, test_scu_ascii_with_device`

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: I have changed it into a Enum class now. Thank you for the tip.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: changed now to include None in the Union

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Ok, thank you

WiktorJJ:
Hello,
We have added the changes demanded. We have also heard that a new release is soon to come and were wondering if we could be included. 
Thank you a lot in advance.

pymeasure/instruments/smaract/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from .scu_ascii import SmarActSCU_ASCII, SCUChannel
```
BenediktBurger: Here, you do not need to import the Channel, as typically nobody uses the Channel alone. The instrument is enough.

pymeasure/instruments/smaract/scu_ascii.py
BenediktBurger: A small point: You still mention "list", but is an enum now. As it is a class, you can even link to it in the documentation, if you write

```suggestion
           The enum :class:'PositionerType' contains the available t values and corresponding properties
```

If you add now also (via autoclass) this enum to the `scu_ascii.rst` file, it will create links and make it clickable to discover.

docs/api/instruments/smaract/index.rst
```diff
@@ -0,0 +1,13 @@
+.. module:: pymeasure.instruments.smaract
+
+#########
+SmarAct
+#########
```
BenediktBurger: Please make the hashes as long as the text.

docs/api/instruments/smaract/scu_ascii.rst
```diff
@@ -0,0 +1,7 @@
+############################################################################################
+SmarAct board for analog signal acquisition and generation as well as digital input/output
+############################################################################################
```
BenediktBurger: see index.rst

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,493 @@
...
+    def calibrate_sensor(self) -> bool:
+        """Calibrate the sensor. Has to be done before the use of :meth: 'move_to_ref'
+           The user must ensure himself that the positioner is not close to a mechanical
+           limit before using this method
+           :return: True if the sensor is calibrating, False otherwise"""
+        self.write(f':CS{self.id}')
+        status = self.ask(f":M{self.id}")
+        return status == f":M{self.id}C"
+
+    def set_zero_position(self):
```
BenediktBurger: Just for your information: If you want to add a type hint that a method does not return anything, you return "None":

```suggestion
    def set_zero_position(self) -> None:
```

We are by far away from adding it everywhere. So it is up to you to add it everywhere (you seem to add types).

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,493 @@
...
+        self.write(f":D{self.id}F{f_val}A{a_val}S{steps}")
+
+    def check_sensor_present(self) -> bool:
+        """Check whether the sensor is present."""
+        self.write(f':GSP{self.id}')
+        response = self.read()
+        return response == f':SP{self.id}P'
+
+    def move_to_ref(self):
+        """Moves up/down to reference"""
```
BenediktBurger: A docstring should be in imperative mode and end with a period:

```suggestion
        """Move up/down to reference."""
```

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,493 @@
...
+        """Moves up a distance + current position"""
+        self.write(f":MPR{self.id}P{convert_quantity_to_magnitude(position, self.unit)}")
+
+
+class SCUChannelAngular(SCUChannel):
+    unit = 'm°'
+
+    @property
+    def angle(self):
+        """Get the current angle in m°"""
```
BenediktBurger: In milli-degrees? I have not seen it, yet.

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,493 @@
...
+    unit = ''
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._current_steps = 0
+
+    @property
+    def position_steps(self):
+        """ Get the current estimated position in steps. """
+        # We simply read our internal counter variable
```
BenediktBurger: This comment just describes what the code does, so it is superfluous.

Comments should state **why**, and only if needed.

pymeasure/instruments/smaract/scu_ascii.py
BenediktBurger: Why do you actually couple the checking with storing the current position?
Wouldn't a method like `set_steps` be better for that?

pymeasure/generator.py
BenediktBurger: Anything is a "object", so this type hint does not convey any information. The reader of the code (or the API docs) won't get any new information from that.

I guess adapter is of the type `Union[str, int, Adapter]`, manufacturer a `str` and adapter_kwargs a `dict`.

These type hints will be useful.

tests/instruments/smaract/test_scu_ascii_with_device.py
```diff
@@ -0,0 +1,248 @@
...
+
+Q_ = ureg.Quantity
+
+CHANNELS = ['0']
+# Here you may add multiple channels ['0', '1', '2', ...]
+SENSOR = [True]
+
+# You can parameterize the following test"
+# with @pytest.mark.parametrize for your number of channels:")
+
```
BenediktBurger, line 39: This comment looks like a left-over.

tests/instruments/test_all_instruments.py
BenediktBurger: But it is part of your PR, so it is a change in your branch.

If you did not intend it, please revert the change.

BenediktBurger (changes requested):
Thank you for your work.

Here a few comments (and answers).

Depending on how far I get with reviews and you with answering, this driver might end up in v0.16.0 or not.
However, I plan to have another release short after with official support for Python 3.14.

pymeasure/instruments/smaract/__init__.py
WiktorJJ: Changed it

docs/api/instruments/smaract/index.rst
WiktorJJ: Changed it now

WiktorJJ: Not sure if the added `PositionerType` autoclass now is implemented correctly

docs/api/instruments/smaract/scu_ascii.rst
WiktorJJ: Done

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Thank you for the tip. We have figured to not add it on movements

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Changed all concerned parts in `scu_ascii`

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Yes in milli-degrees. It was provided as the base variable by SmarAct

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Deleted it

tests/instruments/smaract/test_scu_ascii_with_device.py
WiktorJJ: Changed the `generator.py` and deleted the comment. For the question about set_step we figured that steps are modelled by us, and not really a physical value. Thus we only have a simple check_steps which in the same time does the storing.

WiktorJJ:
Thank you for the response. 
I have added the requested changes and answered the incoming questions.
Great to hear that it could be possible.

docs/api/instruments/smaract/scu_ascii.rst
```diff
@@ -1,7 +1,15 @@
...
 
 .. autoclass:: pymeasure.instruments.smaract.scu_ascii.SmarActSCU_ASCII
     :members:
-    :show-inheritance:
\ No newline at end of file
+    :show-inheritance:
+
+##############
+PositionerType
+##############
```
BenediktBurger: Do you intend to have this as a page title?

You can look at the page if you go into the docs folder and run `make html` afterwards you have a build folder where you can see the _read the docs_ documentation also this file.

Typically, we do not have separations, just the classes.

pymeasure/instruments/smaract/scu_ascii.py
BenediktBurger: I just notice, that you miss a period at the end of the sentence:
~~~suggestion
        """Get the current angle in m°."""
~~~

BenediktBurger (approved):
It looks good, thank you very much for your work and your patience.

Just address the open comments.

You do not need to take care of the SerialAdapter (as coderabbit wants you to).

docs/api/instruments/smaract/scu_ascii.rst
WiktorJJ: Corrected it into just being a class.

WiktorJJ:
Thank you as well for all the advice, I have answered the two last comments. We are very happy to be able to contribute to PyMeasure.

pymeasure/instruments/smaract/scu_ascii.py
BenediktBurger: This comment is still open.

BenediktBurger:
There were a few older comments (scroll through the "files changed" and look at the not resolved - that is open - conversations) for example:

- https://github.com/pymeasure/pymeasure/pull/1416/changes/BASE..1978ffe5502c0afbabf673eb21b644955fa0baf0#r3078850473
- https://github.com/pymeasure/pymeasure/pull/1416/changes/BASE..1978ffe5502c0afbabf673eb21b644955fa0baf0#r3079218634
- ...

Also, you have to resolve the conflict with the authors file (click on resolve conflict).

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Yes, I have this verified with the SmarAct company which produces the microcontrolers

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: the stepper mode for this driver is a model made by us, which is implemented for simplicity. We thought that this could be a small part of the driver, so we put them in as one method.

pymeasure/instruments/smaract/scu_ascii.py
WiktorJJ: Added periods to all end of sentences

pymeasure/instruments/smaract/__init__.py
WiktorJJ: Added

WiktorJJ:
I have adressed the open comments left. The one with `check_steps` and forgotten period. If I am correct, then the rabbitcode ai comments were supposed to be ignored and there is some comments that do not let me resolve them, like the `check_steps` which is a doubble comment and I am not allowed to resolve both at the same time. 

If there is anything else I could do, please feel free to tell me.

pymeasure/generator.py
```diff
@@ -434,7 +435,11 @@ def parse_stream(self):
         self._index = self._stream.tell()
         return self._init_comm_pairs + comm
 
-    def instantiate(self, instrument_class, adapter, manufacturer, adapter_kwargs=None, **kwargs):
+    def instantiate(self, instrument_class,
+                    adapter: Union[str, int],
+                    manufacturer: str,
+                    adapter_kwargs: Optional[dict] = None,
```
BenediktBurger: With the most recent Python version switch, we use `dict | None` instead of Optional/Union:


```suggestion
                    adapter: str | int,
                    manufacturer: str,
                    adapter_kwargs: dict | None = None,
```

That way you won't need the imports either.

tests/test_generator.py
```diff
@@ -254,7 +254,7 @@ def generator(self):
         TC038.string_test = TC038.control("test?", "test %s",  # type: ignore
                                           "Control some string.", cast=str,  # type: ignore
                                           get_process=lambda v: v[7:-1])
-        generator.instantiate(TC038, adapter, "hcp")
+        generator.instantiate(TC038, adapter, "hcp",{})
```
BenediktBurger: If you have None as default, you do not need to give an empty dictionary here.

Also below.

I guess you added it for the type hints?

BenediktBurger (commented):
Two small requests left. Then I will merge it.

Thanks for your work!

tests/test_generator.py
WiktorJJ: Deleted the two empty dictionaries

WiktorJJ:
Should be good now. Thank you for your patience!

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,492 @@
...
+    :param int timeout: Timeout in ms.
+    """
+
+    channel0 = Instrument.ChannelCreator(SCUChannel, "0")
+    channel1 = Instrument.ChannelCreator(SCUChannel, "1")
+    channel2 = Instrument.ChannelCreator(SCUChannel, "2")
+
+    def __init__(self, adapter, name='SCUController',
+                 **kwargs):
+        super().__init__(adapter, name, includeSCPI=False,
```
BenediktBurger: Due to recent changes, you have to remove that includeSCPI (no behaviour change):

```suggestion
        super().__init__(adapter, name,
```

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,492 @@
...
+    unit: str = ''
+
+    frequency_max = Channel.control(
+        ":GCLF{ch}",
+        ":SCLF{ch}F%d",
+        """Control the maximum frequency in an absolute move
+        :param frequency, an int given in Hz in range [1;18500]
+        """,
+        set_process=lambda v: convert_quantity_to_magnitude(v, 'Hz'),
+        get_process=lambda s: Q_(int(s[6:]), 'Hz'),
```
BenediktBurger, line 99: I guess you read a string from the device, right? You need to cast to str:

```suggestion
        get_process=lambda s: Q_(int(s[6:]), 'Hz'),
        cast=str,
```

tests/instruments/smaract/test_scu_ascii.py
```diff
@@ -0,0 +1,250 @@
...
+    with expected_protocol(
+            SmarActSCULinear,
+            [(f':GCLF{channel}'.encode(), b':CLF0F500')],
+    ) as inst:
+        assert inst.channels[channel].frequency_max.magnitude == 500
+
+
+@pytest.mark.parametrize("channel", CHANNELS)
+def test_safe_direction_setter(channel):
+    for comm_pairs, value in ((((f':SSD{channel}D1'.encode(), None),), True),
```
BenediktBurger: Actually, you do not need encode, the expected_protocol can also take strings!

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,492 @@
...
+
+    safe_direction_up_enabled = Channel.control(
+        ":GSD{ch}",
+        ":SSD{ch}D%d",
+        """Control the safe direction of the :meth: 'move_to_ref'
+           :param  must be either 0 (down) or 1 (up). In minor letters""",
+        validator=strict_discrete_set,
+        map_values=True,
+        values={True: 1, False: 0},
+        get_process=lambda x: int(x[-1]),
```
BenediktBurger, line 111: You read strings, right -> cast=str:

```suggestion
        get_process=lambda x: int(x[-1]),
        cast=str,
```

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,492 @@
...
+        self.adapter.close()
+
+    serial_nb = Instrument.measurement(
+        ":GID", """Get the serial number. """,
+        get_process=lambda s: s[3:]
+    )
+
+    model = Instrument.measurement(
+        ":I", """Get the device model information. """,
+        get_process=lambda s: s[2:]
```
BenediktBurger:
```suggestion
        get_process=lambda s: s[2:],
        cast=str,
```

pymeasure/instruments/smaract/scu_ascii.py
```diff
@@ -0,0 +1,492 @@
...
+        self._freq: Q_ = Q_(3000, 'Hz')
+        self._steps: int = 2000
+        self._current_steps: int = 0
+
+    def close(self):
+        self.adapter.close()
+
+    serial_nb = Instrument.measurement(
+        ":GID", """Get the serial number. """,
+        get_process=lambda s: s[3:]
```
BenediktBurger:
```suggestion
        get_process=lambda s: s[3:],
        cast=str,
```

BenediktBurger (approved):
There have been a few changes, which you have to address, most of them I prepared as commitable ones:
- includeSCPI was obsolete and should be removed now.
- if you want to read a string, you have to say `cast=str`, you are not anymore allowed to expect the cast to float to fail.

Only, I noticed, that you encode all the tests to bytes, this is not necessary.

tests/instruments/smaract/test_scu_ascii.py
WiktorJJ: Done. I have deleted all .encode() on each expected_protocol and comm_pairs.

WiktorJJ:
I am sorry for a late answer, but I hope this fix will prove helpful for PyMeasure. I have also removed the encoding of bytes in the test file.

BenediktBurger:
Thank you for your work and patience.

WiktorJJ:
Thank you as well for the patience and the help provided throughout this merge.

seb5g:
Well done Wiktor and Juan, I'm glad you went to the end of this work!

WiktorJJ:
Thank you Mr. Weber for the help and patience you had with us.

PR #1293, opened 2025-04-26
Keysight B29XXB Source-Measure Units.

langrock:
Some of the most basic functions to control the B29XXB version of Keysight's Source-Measure Units (SMU).

PR #1421, opened 2026-04-18
Implement KeysightB2912A PSMU

AlecVercruysse:
Hi, I have basic functionality for the Keysight B2900 series PSMUs. I see, in hindsight, that PR #1293 is already open for this series, but it looks stale (1 year old). The Programming Manual I referenced is for all "B2900 Series", but limits are set by sub-name (i.e., B2901C), and I only had the Keysight B2912A on hand, so maybe a future PR can refactor to set limits properly by sub-name?

If this looks ok, I will write tests that aren't just targeting `_with_device.py`

Thanks,
Alec


AlecVercruysse:
Closing as there's an issue with the git history.

PR #1424, opened 2026-04-28
Update Pandas version to 3 and 4

maduka-ariyasiri:
This change updates the pandas and NumPy dependencies to their latest versions to ensure compatibility with current environments. The update has been verified against existing functionality, and no breaking changes were observed.


BenediktBurger:
Thank you very much for your suggestion.

We added the upper bound in order to keep the code working. The last change (numpy 1 to 2) added some incompatibilities which we had to fix first. With this upper bound, the code will work until we made the adjustments.
Numpy even recommends to do that.

With the pandas major update, we have to check, whether any of these breaking changes apply to us: https://pandas.pydata.org/docs/whatsnew/v3.0.0.html#backwards-incompatible-api-changes
A major version means (possible) backward incompatibility, so requires more attention.

Otherwise it is not good to remove the limit as it might break the code.

maduka-ariyasiri:
Thank you for the detailed explanation; that makes complete sense, and I agree that removing upper bounds without validation would be risky.

To address those concerns, I’ve taken a compatibility-first approach in this PR:

## 1. NumPy 2.x compatibility

You mentioned the incompatibilities introduced in NumPy 2.0. While testing against NumPy ≥1.24 / 2.x, I identified one concrete issue:

- Usage of deprecated `np.bool`, which raises an `AttributeError` in newer versions

This has been fixed by switching to `np.bool_`, which is the correct and stable type going forward. A regression test has also been added to ensure this remains compatible.

## 2. pandas 3.0 compatibility

Following your point about major-version breakage, I reviewed the pandas 3.0 backward-incompatible changes:

- The relevant change affecting this codebase is the requirement to explicitly specify `sort` in `pd.concat`

All affected call sites have been updated with `sort=False`, which:

- preserves existing behaviour  
- is backward-compatible with pandas 2.x  

## 3. Verification via tests (not assumptions)

To avoid relying on assumptions, I added:

- A dedicated CI job that runs against latest NumPy (2.x) and pandas (3.x)  
- Targeted tests covering the main risk areas:
  - dtype stability across reads/writes  
  - protection against unintended object/string coercion  
  - column ordering behaviour in `concat`  
  - numpy scalar compatibility (`np.bool_`)  
  - empty dataset handling  

All existing tests continue to pass unchanged, and the new tests pass under both current and latest versions.

## 4. On removing upper bounds

Given:

- the identified incompatibilities have been addressed  
- and compatibility is now verified in CI  

the upper bounds no longer appear necessary for the currently tested scope.

pyproject.toml
```diff
@@ -35,8 +35,8 @@ classifiers = [
     "Topic :: Scientific/Engineering",
 ]
 dependencies = [
-    "numpy>=1.6.1,<3",
-    "pandas>=0.14,<3",
+    "numpy>=1.6.1",
+    "pandas>=0.14",
```
BenediktBurger: I'm fine with **increasing** the upper bound for pandas to 4, but for numpy, we should not stop the limit!

```suggestion
    "numpy>=1.6.1,<3",
    "pandas>=0.14,<4",
```

maduka-ariyasiri: Thanks, I’ll add the upper bounds.

maduka-ariyasiri:
Hi @BenediktBurger 

I’ve addressed the review comments, so please could you review and approve this PR?

I also wanted to ask about the release cycle, as I can see there are some recently added features in the project that still aren’t available on PyPI until the next release.

@coderabbitai resume

pyproject.toml
```diff
@@ -30,7 +30,7 @@ classifiers = [
 ]
 dependencies = [
     "numpy>=1.6.1,<3",
-    "pandas>=0.14,<3",
+    "pandas>=0.14,<4",
```
maduka-ariyasiri, line 33: The dependency upper bound was intentionally updated from <3 to <4 as part of enabling compatibility with pandas 3.x. Existing functionality was validated against the supported versions.

maduka-ariyasiri (commented):
I’ve addressed the review comments, so please could you review and approve this PR?

<sub>Reviewed in [CodeRabbit Review Stack](https://app.coderabbit.ai/change-stack/pymeasure/pymeasure/pull/1424)</sub>

pymeasure/experiment/parameters.py
```diff
@@ -231,8 +231,8 @@ def convert(self, value):
             value = bool(value)
         elif isinstance(value, bool):
             value = value
-        elif isinstance(value, np.bool):
-            value = value
+        elif isinstance(value, np.bool_):
```
BenediktBurger, line 234: Why do you change to `bool_`?

According to the docs (https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bool_) is bool_ an alias for bool.

tests/experiment/test_parameters.py
```diff
@@ -95,11 +95,11 @@ def test_boolean_value_error():
                          ["True", True],
                          ["true", True],
                          [1, True],
-                         [np.bool(True), True],
+                         [np.bool_(True), True],
```
BenediktBurger, line 98: See above.

.github/workflows/pymeasure_CI.yml
BenediktBurger, line 1: I'm unsure whether I like this addition of tests. On the other hand, it allows us to test pandas 3, so I guess we need it (for some time).

BenediktBurger (approved):
Looks good, only minor points.

I'm preparing a release, so this might end up on pypi soon.

BenediktBurger:
Hey @maduka-ariyasiri , once you address my comments, especially regarding `bool_`, I can merge your PR.

Also, you may add yourself to the `authors.txt` file in this PR, if you want to.

pymeasure/experiment/parameters.py
maduka-ariyasiri: One reason: np.bool was removed in NumPy 1.24 and only reintroduced in 2.0, so isinstance(value, np.bool) raise AttributeError on NumPy 1.24 to 1.26. np.bool_ exists in every supported version. 

I think you may have misread, The numpy docs says(NumPy 2.x docs) bool_ is an alias of numpy.bool; that's numpy's own scalar type, not python's built-in bool. np.bool_ is the genuine numpy boolean scalar; it isn't an instance of python's bool, without this change fall though and raise valuerror. the bool(value) cast normalise it to a plain python bool.

tests/experiment/test_parameters.py
maduka-ariyasiri: same answer,  np.bool_(True) exists in every supported NumPy version and produces a real numpy.bool_ scalar.

.github/workflows/pymeasure_CI.yml
maduka-ariyasiri: As you said, if we extend support to pandas 3, it would be better to extend the tests as well.

maduka-ariyasiri:
@coderabbitai review

I’ve changed the PR name, since this PR only handles the pandas version update.

.github/workflows/pymeasure_CI.yml
```diff
@@ -124,3 +127,49 @@ jobs:
...
+      - name: Run pytest with xvfb
+        if: runner.os == 'Linux'
+        run: |
+          echo "::add-matcher::.github/pytest.json"
+          xvfb-run -a pytest
+      - name: Run pytest
+        if: runner.os != 'Linux'
+        run: |
+          echo "::add-matcher::.github/pytest.json"
+          pytest
```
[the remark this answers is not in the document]
maduka-ariyasiri: I’ve changed the PR name, since this PR only handles the pandas version update.

maduka-ariyasiri:
Hi @BenediktBurger, I’ve added myself to authors.txt, addressed the review comments, pegged the micromamba version due to the failure, and also adjusted the new test section for pandas 3.

Hopefully this satisfies the requirements and the PR can now be merged.

PR #1426, opened 2026-04-29
Change the SR860 min ampl

chieh0517:
SR860 can set the amplitude to 0
Also, SR860 support the SCPI command

### DIFF
--- a/pymeasure/instruments/srs/sr860.py
+++ b/pymeasure/instruments/srs/sr860.py
@@ -317,28 +317,28 @@ class SR860(Instrument):
         """Floating point property representing the preset sine out amplitude, for the A1 preset button.
         This property can be set.""",  # noqa: E501
         validator=truncated_range,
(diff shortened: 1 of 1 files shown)


BenediktBurger:
Thanks for your PR.

> Also, SR860 support the SCPI command

What do you mean with that? That it supports the default SCPI commands which are defined by the SCPIMixin class?

chieh0517:
> What do you mean with that? That it supports the default SCPI commands which are defined by the SCPIMixin class?

I am using the latest version of PyMeasure (v0.15.0) and received the following warning:
"It is not known whether this device support SCPI commands or not. Please inform the pymeasure maintainers if you know the answer."

chieh0517:
I don't know whether this is okay.
If it is a problem, I can PR it again without that description.

BenediktBurger:
Thanks for your work

PR #1428, opened 2026-04-30
Remove chunked reading from logic in `Results.data` property and `Results.reload()`

mcdo0486:
Remove chunked reading from logic in `Results.data` property and `Results.reload()` Previously the file was read in chunks then concat in memory, now it reads the entire file to memory removing unnecessary overhead. 

A benchmark of loading a Results file with 500K rows and 5 columns showed a ~68% time improvement (634.7 ms to 200.3 ms).

### DIFF
--- a/pymeasure/experiment/results.py
+++ b/pymeasure/experiment/results.py
@@ -435,8 +435,7 @@ class Results:
     def data(self):
         # Need to update header count for correct referencing
         if self._header_count == -1:
-            self._header_count = len(
-                self.header()[-1].split(Results.LINE_BREAK))
+            self._header_count = len(self.header()[-1].split(Results.LINE_BREAK))
         if self._data is None or len(self._data) == 0:
             # Data has not been read
             try:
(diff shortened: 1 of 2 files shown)


BenediktBurger:
I don't know, why it was done in chunks and a quick check did not reveal anything.

Probably we could do this improvement.

@CasperSchippers, what do you think?

CasperSchippers:
I'm not entirely sure (I'm on holidays so can't quickly play around with it) but I would expect that the chunked reading is meant for (very) large datasets (e.g. multiple GB of data, maybe even more that the RAM you have). I wonder what happens to the benchmarking comparison you did if you up the chunk size (currently at 1000, which may be very small for modern PCs, could set is too 100000 for example).

CasperSchippers:
I've very briefly played around with it, but I couldn´t see much of a difference between chunked with chunksize=1000, with chunksize=100000000, or without chunking (tried with 2 columns and 11M rows, file of ~280MB). I didn´t have time yet to create a file>memory to see how that behaves (or something with multiple columns to replicate your results).

So while I'm all in favour if we can speed this part up, I do want to understand why the current chunking is in place and test it for a few scenarios (before we break intential behaviour for maybe very very large files).

PR #1429, opened 2026-04-30
Add file size checking to `Results.data` property to skip reading file if data hasn't changed.

mcdo0486:
Add file size checking to `Results.data` property. When there is potentially new data, first check to see if the size of the file has changed. If it hasn't, there is no new data, return current _data.

`Results.data` is frequently polled by display widgets. For example, the default `PlotWidget` creates `PlotFrame` with a `refresh_time` of 0.2 seconds. On each tick, `PlotFrame.update_curves()` calls `ResultsCurve.update_data()`, which accesses `Results.data`. This change prevents the file being read if there are no changes.

### DIFF
--- a/pymeasure/experiment/results.py
+++ b/pymeasure/experiment/results.py
@@ -217,6 +217,7 @@ class Results:
         self.parameters = procedure.parameter_objects()
         self._header_count = -1
         self._metadata_count = -1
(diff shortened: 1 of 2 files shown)


BenediktBurger (approved):
Looks good, just address coderabbit's comment.

PR #1432, opened 2026-05-04
Allow sharing a single pyvisa.ResourceManager across instruments

nusssreiner:
﻿## Summary

Today every `VISAAdapter` (and therefore every `Instrument`) creates its own
`pyvisa.ResourceManager`. When a measurement setup uses several VISA
instruments at the same time this leads to multiple resource managers being
opened against the same VISA backend, which is wasteful and on some backends
(e.g. NI-VISA on Windows) can cause locking / device-discovery issues.

This PR lets the user pass an already-existing `pyvisa.ResourceManager` to
PyMeasure so that one RM can be reused by several instruments. Existing
behaviour is preserved when no RM is passed.

## Changes

- `pymeasure/adapters/visa.py`
  - `VISAAdapter` accepts a `pyvisa.ResourceManager` instance via the
    existing `visa_library` parameter and reuses it instead of creating a
    new one. Anything else is dispatched to the previous code path.
- `pymeasure/instruments/instrument.py`
  - `Instrument(adapter=rm, name=resource_name)` is now valid: when
    `adapter` is a `pyvisa.ResourceManager`, a `VISAAdapter` is built that
    reuses that RM and uses `name` as the VISA resource string.
  - Adds identity attributes (`resource_name`, `vendor`, `serial_number`,
    `firmware_ref`) on the base class so SCPI subclasses can populate them
    without re-declaring (used by the follow-up PR).

## Backwards compatibility

Fully backwards compatible. All existing call sites continue to work
unchanged; the new behaviour is only triggered when a `ResourceManager`
instance is passed.

## Tests

- `tests/adapters/test_visa.py` - verifies that a passed-in RM is reused.
- `tests/instruments/test_instrument.py` - verifies the
  `Instrument(adapter=rm, name=...)` path.

## Notes for reviewers

This is the base of a small stack. Two follow-up PRs build on it:

1. SCPIMixin helpers for `open` / `close` / `shutdown` / `get_device_info`
   (uses the new identity attributes).
2. A new Rohde & Schwarz NGPx power-supply driver (uses both of the above).

Each will be opened separately and rebased once this one is merged.

### DIFF
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -25,6 +25,7 @@
 import logging
 
 import pyvisa
+from pyvisa import ResourceManager
 
 from .adapter import Adapter
 from .protocol import ProtocolAdapter
@@ -42,8 +43,12 @@ class VISAAdapter(Adapter):
     :param resource_name: A
         `VISA resource string <https://pyvisa.readthedocs.io/en/latest/introduction/names.html>`__
         or GPIB address integer that identifies the target of the connection
-    :param visa_library: PyVISA VisaLibrary Instance, path of the VISA library or VisaLibrary spec
-        string (``@py`` or ``@ivi``). If not given, the default for the platform will be used.
+    :param visa_library: PyVISA VisaLibrary Instance, an existing
+        :class:`pyvisa.ResourceManager`, a path of the VISA library, or a
+        VisaLibrary spec string (``@py`` or ``@ivi``). If not given, the
+        default for the platform will be used. Passing an already existing
+        :class:`pyvisa.ResourceManager` allows several instruments to share a
+        single resource manager instead of each creating its own.
     :param log: Parent logger of the 'Adapter' logger.
     :param \\**kwargs: Keyword arguments for configuring the PyVISA connection.
 
@@ -87,7 +92,10 @@ class VISAAdapter(Adapter):
             resource_name = "GPIB0::%d::INSTR" % resource_name
 
         self.resource_name = resource_name
-        self.manager = pyvisa.ResourceManager(visa_library)
+        if isinstance(visa_library, ResourceManager):
+            self.manager = visa_library
+        else:
+            self.manager = pyvisa.ResourceManager(visa_library)
 
         # Clean up kwargs considering the interface type matching resource_name
         if_type = self.manager.resource_info(self.resource_name).interface_type
--- a/pymeasure/adapters/visa.py
+++ b/pymeasure/adapters/visa.py
@@ -87,6 +87,7 @@ class VISAAdapter(Adapter):
             self.resource_name = getattr(resource_name, "resource_name", None)
             self.connection = resource_name.connection
             self.manager = resource_name.manager
+            self._owns_manager = False
             return
         elif isinstance(resource_name, int):
             resource_name = "GPIB0::%d::INSTR" % resource_name
@@ -94,8 +95,10 @@ class VISAAdapter(Adapter):
         self.resource_name = resource_name
         if isinstance(visa_library, ResourceManager):
             self.manager = visa_library
+            self._owns_manager = False
         else:
             self.manager = pyvisa.ResourceManager(visa_library)
+            self._owns_manager = True
 
         # Clean up kwargs considering the interface type matching resource_name
         if_type = self.manager.resource_info(self.resource_name).interface_type
@@ -122,10 +125,16 @@ class VISAAdapter(Adapter):
 
             This closes the connection to the resource for all adapters using
             it currently (e.g. different adapters using the same GPIB line).
+            The underlying :class:`pyvisa.ResourceManager` is only closed when
+            this adapter created it; a manager passed in via ``visa_library``
+            is left open so that other adapters sharing it remain usable.
         """
         super().close()
         try:
-            if self.manager.visalib.library_path == "unset":
+            if (
+                getattr(self, "_owns_manager", True)
+                and self.manager.visalib.library_path == "unset"
+            ):
                 # if using the pyvisa-sim library the manager has to be also closed.
                 # this works around https://github.com/pyvisa/pyvisa-sim/issues/82
                 self.manager.close()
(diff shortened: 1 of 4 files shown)


pymeasure/adapters/visa.py
```diff
@@ -42,8 +43,12 @@ class VISAAdapter(Adapter):
     :param resource_name: A
         `VISA resource string <https://pyvisa.readthedocs.io/en/latest/introduction/names.html>`__
         or GPIB address integer that identifies the target of the connection
-    :param visa_library: PyVISA VisaLibrary Instance, path of the VISA library or VisaLibrary spec
-        string (``@py`` or ``@ivi``). If not given, the default for the platform will be used.
+    :param visa_library: PyVISA VisaLibrary Instance, an existing
```
BenediktBurger, line 46: Warning: I added type hints to this file on master, so you have to adjust it from str to `Union[str, ResourceManager]`.

pymeasure/adapters/visa.py
```diff
@@ -82,12 +87,18 @@ def __init__(self, resource_name, visa_library="", log=None, **kwargs):
...
             resource_name = "GPIB0::%d::INSTR" % resource_name
 
         self.resource_name = resource_name
-        self.manager = pyvisa.ResourceManager(visa_library)
+        if isinstance(visa_library, ResourceManager):
+            self.manager = visa_library
+            self._owns_manager = False
+        else:
+            self.manager = pyvisa.ResourceManager(visa_library)
+            self._owns_manager = True
```
BenediktBurger, line 101: So, the first Adapter owns the resource manager and closing the first adapter will close the connection for all, right?

Wouldn't it be better to have a counter how many adapters use a resource manager and close it once nobody uses it?

Maybe a "force_close" method could close it even if someone else uses it?

pymeasure/adapters/visa.py
```diff
@@ -114,10 +125,16 @@ def close(self):
 
             This closes the connection to the resource for all adapters using
             it currently (e.g. different adapters using the same GPIB line).
```
BenediktBurger, line 127: It won't be only GPIB anymore, right?

pymeasure/adapters/visa.py
```diff
@@ -114,10 +125,16 @@ def close(self):
...
+            The underlying :class:`pyvisa.ResourceManager` is only closed when
+            this adapter created it; a manager passed in via ``visa_library``
+            is left open so that other adapters sharing it remain usable.
```
BenediktBurger, line 130: Even if the "first adapter" is closed, the other adapters remain usable?

pymeasure/instruments/instrument.py
```diff
@@ -74,12 +80,15 @@ class Instrument(CommonBase):
     def __init__(self, adapter, name, includeSCPI=None,
```
BenediktBurger, line 80: We added type hints to the init, so now, it should also accept a ResourceManager.

pymeasure/instruments/instrument.py
```diff
@@ -91,6 +100,12 @@ def __init__(self, adapter, name, includeSCPI=None,
         self.SCPI = includeSCPI
         self.isShutdown = False
         self.name = name
+        self.resource_name = getattr(self.adapter, "resource_name", name)
+        # Only seed the identity fields on subclasses that don't already
+        # define them as (read-only) properties of their own.
+        for _attr in ("vendor", "serial_number", "firmware_ref"):
+            if not hasattr(type(self), _attr):
+                setattr(self, _attr, "")
```
BenediktBurger, line 108: Why do you redefine them?

pymeasure/instruments/instrument.py
```diff
@@ -91,6 +100,12 @@ def __init__(self, adapter, name, includeSCPI=None,
         self.SCPI = includeSCPI
         self.isShutdown = False
         self.name = name
+        self.resource_name = getattr(self.adapter, "resource_name", name)
```
BenediktBurger, line 103: Why do we need this? Just for tests?

tests/instruments/test_instrument.py
```diff
@@ -129,6 +129,21 @@ def test_init_visa_fail():
...
+    import pyvisa
+
+    rm = pyvisa.ResourceManager("@sim")
+    instr_a = Instrument(rm, "ASRL1::INSTR", includeSCPI=False)
+    instr_b = Instrument(rm, "ASRL2::INSTR", includeSCPI=False)
+    assert instr_a.adapter.manager is rm
+    assert instr_b.adapter.manager is rm
+    assert instr_a.resource_name == "ASRL1::INSTR"
+    assert instr_b.resource_name == "ASRL2::INSTR"
+    assert instr_a.vendor == "" and instr_a.serial_number == "" and instr_a.firmware_ref == ""
```
BenediktBurger, line 144: If you assert several things, consider using a fixture which sets the test up and use individual tests for the different asserts.
That way, you know whether the vendor/serial_number thing fails or whether the resource name etc, or whether both fail.

Currently, you know only the first assertion which fails.

pymeasure/instruments/instrument.py
```diff
@@ -47,6 +49,10 @@ class Instrument(CommonBase):
     defining the target of your connection.
 
     When ``adapter`` is an integer, a GPIB resource name is created based on that.
+    When ``adapter`` is a :class:`pyvisa.ResourceManager`, a
+    :py:class:`~pymeasure.adapters.VISAAdapter` is constructed that reuses this
+    resource manager; in that case ``name`` is interpreted as the VISA resource
```
BenediktBurger, line 54: I don't like the "abuse" of the field `name`.

Actually, if you give `InstrumentClass("some_resource_string", visa_library=rm)`  for some resource manager `rm`, it will work as the instrument hands all other arguments to the VISAAdapter init call.

Then you won't need the changes below. It will work out of the box.

pymeasure/adapters/visa.py
```diff
@@ -114,10 +125,16 @@ def close(self):
...
         """
         super().close()
         try:
-            if self.manager.visalib.library_path == "unset":
+            if (
+                getattr(self, "_owns_manager", True)
```
BenediktBurger, line 135: You always set _owns_manager in the init, so you shouldn't need this security step, do you?

EDIT: I see, the ProtocolAdapter path does not set it.

pymeasure/adapters/visa.py
BenediktBurger, line 101: Why don't you set _owns_manager to False and only in this single case you set it to True?

BenediktBurger (changes requested):
Thank you very much for your contribution. A very interesting idea.

As a new contributor, please either add yourself to the `authors.txt`  or tell me that you do not want to add yourself.

PR #1433, opened 2026-05-04
Add open/close/shutdown and *IDN?-based device-info helpers to SCPIMixin

nusssreiner:
﻿## Summary

Many SCPI drivers re-implement the same handful of helpers: opening and
closing the underlying VISA session, marking the instrument as shut down,
parsing `*IDN?` into vendor / model / serial / firmware, and refusing to
talk to a device that is not in a list of supported models. This PR moves
those helpers into `SCPIMixin` so every SCPI instrument gets them for free.

## Changes

- `pymeasure/adapters/adapter.py`
  - Adds an `Adapter.open()` no-op base implementation, mirroring the
    existing `Adapter.close()`, so `SCPIMixin.open()` has something to
    delegate to in the generic case.
- `pymeasure/instruments/generic_types.py` (`SCPIMixin`)
  - `open()` / `close()` toggle the underlying VISA connection through the
    adapter.
  - `shutdown()` flags the instrument as shut down.
  - `get_device_info()` queries `*IDN?` and populates `vendor`, `name`,
    `serial_number` and `firmware_ref`. Note: `self.name` is overwritten
    with the model designation reported by the device.
  - `check_is_dev_supported(instr_list, errMsg)` raises `AssertionError`
    if the connected model is not in `instr_list`.

## Backwards compatibility

Additive only - no existing method is changed. Drivers that already define
their own `open` / `close` / `shutdown` continue to take precedence via
normal MRO.

## Tests

- `tests/instruments/test_generic_types.py` covers all five helpers,
  including the supported / unsupported model cases.

## Depends on

This branch is stacked on top of #1432
("Allow sharing a single pyvisa.ResourceManager across instruments"),
because `get_device_info()` writes into the identity attributes
(`vendor`, `serial_number`, `firmware_ref`) introduced there.

After PR #1432 is merged I would rebase this branch onto `master`
and force-push, so the diff will shrink to only the SCPIMixin changes.

### DIFF
--- a/pymeasure/instruments/generic_types.py
+++ b/pymeasure/instruments/generic_types.py
@@ -120,28 +120,35 @@ class SCPIMixin:
         ``firmware_ref`` and stores them on the instrument. Note that
         ``self.name`` is overwritten with the model designation reported by
         the device.
+
+        :raises ValueError: If the ``*IDN?`` response does not contain four
+            comma-separated fields.
         """
         resp_str = self.id
-        vendor, name, serial_number, firmware_ref = resp_str.split(",")
-        self.vendor = vendor
-        self.name = name
-        self.serial_number = serial_number
-        self.firmware_ref = firmware_ref
+        parts = [p.strip() for p in resp_str.split(",", 3)]
+        if len(parts) != 4:
+            raise ValueError(f"Unexpected *IDN? response format: {resp_str!r}")
+        self.vendor, self.name, self.serial_number, self.firmware_ref = parts
 
     def check_is_dev_supported(self, instr_list, errMsg):
         """Check whether the connected instrument is in ``instr_list``.
 
-        :param instr_list: Iterable of supported model name substrings, or ``None``
-            to skip the check.
+        Each token in ``instr_list`` is treated as a model-name substring and
+        is searched for in the device name reported by the instrument
+        (typically the model field of ``*IDN?``).
+
+        :param instr_list: Iterable of supported model-name substrings, or
+            ``None`` to skip the check.
         :param errMsg: Message appended to ``self.name`` when raising.
-        :raises AssertionError: If no connection has been opened (``self.name`` is
-            ``None``) or the instrument is not in ``instr_list``.
+        :raises AssertionError: If no connection has been opened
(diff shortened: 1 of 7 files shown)


pymeasure/instruments/generic_types.py
```diff
@@ -100,6 +100,58 @@ def check_errors(self):
                 break
         return errors
 
+    def close(self):
+        """Close the underlying VISA connection."""
+        self.adapter.close()
+
+    def open(self):
+        """Reopen the underlying VISA connection."""
+        self.adapter.open()
```
BenediktBurger, line 109: They don't have to do anything with SCPI, so they should belong to the instrument itself.

pymeasure/instruments/generic_types.py
```diff
@@ -100,6 +100,58 @@ def check_errors(self):
...
+        the device.
+
+        :raises ValueError: If the ``*IDN?`` response does not contain four
+            comma-separated fields.
+        """
+        resp_str = self.id
+        parts = [p.strip() for p in resp_str.split(",", 3)]
+        if len(parts) != 4:
+            raise ValueError(f"Unexpected *IDN? response format: {resp_str!r}")
+        self.vendor, self.name, self.serial_number, self.firmware_ref = parts
```
BenediktBurger, line 131: Why do you want to store that information? You rarely need to read it and if you read it, you can just read it from the device. No need to store it.

pymeasure/instruments/generic_types.py
```diff
@@ -100,6 +100,58 @@ def check_errors(self):
...
+        Each token in ``instr_list`` is treated as a model-name substring and
+        is searched for in the device name reported by the instrument
+        (typically the model field of ``*IDN?``).
+
+        :param instr_list: Iterable of supported model-name substrings, or
+            ``None`` to skip the check.
+        :param errMsg: Message appended to ``self.name`` when raising.
+        :raises AssertionError: If no connection has been opened
+            (``self.name`` is ``None``) or none of the supported tokens is a
+            substring of the connected device name.
```
BenediktBurger, line 145: What do you want to do with this method?

PR #1435, opened 2026-05-04
Add Keysight 35670A dynamic signal analyzer

PeterPirog:
Summary
Add Keysight35670A driver for Keysight/Agilent/HP 35670A Dynamic Signal Analyzer.
Support input channels, trace boxes, display windows, source/output, trigger, sense, status/system, memory/MMEM/program helpers, and block-transfer helpers.
Add protocol tests and safe hardware tests.
Tests
.\.venv\Scripts\python.exe -m pytest tests/instruments/keysight/test_keysight35670A.py -q

Result: 655 passed

.\.venv\Scripts\python.exe -m pytest tests/instruments/keysight/test_keysight35670A_with_device.py -q

Result: 10 skipped

.\.venv\Scripts\python.exe -m pytest tests/instruments/keysight/test_keysight35670A_with_device.py --device-address "GPIB0::14::INSTR" -q -s

Result: 10 passed

Hardware
Instrument ID: HEWLETT-PACKARD,35670A,MY42509215,A.01.11
Options: "1D2,1D3,1D4,1C2,UFF,UFC,AY6"
Final SYSTem:ERRor?: +0,"No error"
Safety
Destructive or long-running operations require explicit confirmation in the API and are not executed in hardware tests.
Source output is forced off in safe hardware tests.
Hardware tests are skipped unless a device address is explicitly provided.
Notes
The driver accepts vendor IDs HEWLETT-PACKARD, HP, AGILENT, and KEYSIGHT.
Four-channel paths are option-dependent and were tested on hardware with option AY6.

### DIFF
--- a/docs/api/instruments/keysight/keysight35670A.rst
+++ b/docs/api/instruments/keysight/keysight35670A.rst
@@ -1,0 +1,44 @@
+##########################################
+Keysight 35670A Dynamic Signal Analyzer
+##########################################
+
+.. code-block:: python
+
+    from pymeasure.instruments.keysight import Keysight35670A
+
+The Keysight/Agilent/HP 35670A driver provides a dynamic signal analyzer API for
+measurement and configuration workflows.
+It supports 2 or 4 input channels depending on option AY6, trace boxes 1..4,
+source-output control, and block transfer helpers for binary/definite-block
+data paths.
+Destructive memory, system, and program operations require explicit
+confirmation arguments.
+
+.. autoclass:: pymeasure.instruments.keysight.Keysight35670A
+    :members:
+    :show-inheritance:
+    :inherited-members:
+
+.. autoclass:: pymeasure.instruments.keysight.keysight35670A.Keysight35670AInputChannel
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysight35670A.Keysight35670ASenseWindow
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysight35670A.Keysight35670ADisplayWindow
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysight35670A.Keysight35670AWindow
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysight35670A.Keysight35670AOrderTrack
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysight35670A.Keysight35670ATrace
+    :members:
+    :show-inheritance:
--- a/pymeasure/instruments/keysight/keysight35670A.py
+++ b/pymeasure/instruments/keysight/keysight35670A.py
@@ -1,0 +1,4959 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2026 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import logging
+from csv import reader
+from io import StringIO
+from typing import Optional
+
+from pymeasure.instruments import Channel, Instrument, SCPIMixin
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+SOURCE_FUNCTIONS = {
+    "sine": "SIN",
+    "random": "RAND",
+    "burst_random": "BRAN",
+    "periodic_chirp": "PCH",
+    "burst_chirp": "BCH",
+    "pink": "PINK",
+    "user": "USER",
+    "capture": "CAPT",
+}
+
+SOURCE_USER_REGISTERS = {
+    1: "D1",
+    2: "D2",
+    3: "D3",
+    4: "D4",
+    5: "D5",
+    6: "D6",
+    7: "D7",
+    8: "D8",
+}
+
+SOURCE_REFERENCE_CHANNELS = {
+    1: "INP1",
+    2: "INP2",
+    3: "INP3",
+    4: "INP4",
+}
+
+INSTRUMENT_MODES = {
+    "fft": "FFT",
+    "correlation": "CORR",
+    "histogram": "HIST",
+    "octave": "OCT",
+    "order": "ORD",
+    "swept_sine": "SINE",
+}
+
+TRACE_FEEDS = {
+    "data_register_1": "D1",
+    "data_register_2": "D2",
+    "data_register_3": "D3",
+    "data_register_4": "D4",
+    "power_spectrum_ch1": "XFR:POW 1",
+    "power_spectrum_ch2": "XFR:POW 2",
+    "linear_spectrum_ch1": "XFR:POW:LIN 1",
+    "linear_spectrum_ch2": "XFR:POW:LIN 2",
+    "frequency_response_2_1": "XFR:POW:RAT 2,1",
+    "coherence_1_2": "XFR:POW:COH 1,2",
+    "cross_spectrum_1_2": "XFR:POW:CROS 1,2",
+    "time_ch1": "XTIM:VOLT 1",
+    "time_ch2": "XTIM:VOLT 2",
+    "windowed_time_ch1": "XTIM:VOLT:WIND 1",
+    "windowed_time_ch2": "XTIM:VOLT:WIND 2",
+}
+
+TRACE_ACTIVE_VALUES = {
+    "a": "A",
+    "b": "B",
+    "c": "C",
+    "d": "D",
+    "ab": "AB",
+    "cd": "CD",
+    "abcd": "ABCD",
+}
+
+TRACE_DISPLAY_FORMATS = {
+    "linear_magnitude": "MLIN",
+    "log_magnitude": "MLOG",
+    "phase": "PHAS",
+    "real": "REAL",
+    "imaginary": "IMAG",
+    "nyquist": "NYQ",
+    "unwrapped_phase": "UPH",
+    "group_delay": "GDEL",
+    "polar": "POL",
+}
+
+AVERAGE_TYPES = {
+    "maximum": "MAX",
+    "rms": "RMS",
+    "time": "TIME",
+    "vector": "VECT",
+    "equal_confidence": "ECON",
+}
+
+AVERAGE_TCONTROLS = {
+    "freeze": "FREE",
+    "repeat": "REP",
+    "exponential": "EXP",
+}
+
+AVERAGE_HOLD_VALUES = {
+    "off": "OFF",
+    "maximum": "MAX",
+    "minimum": "MIN",
+}
+
+AVERAGE_PREVIEW_VALUES = {
+    "off": "OFF",
+    "manual": "MAN",
+    "timed": "TIM",
+}
+
+REFERENCE_CHANNEL_VALUES = {
+    "single": "SING",
+    "pair": "PAIR",
+}
+
+WINDOW_TYPES = {
+    "hanning": "HANN",
+    "flattop": "FLAT",
+    "uniform": "UNIF",
+    "force": "FORC",
+    "exponential": "EXP",
+    "lag": "LAG",
+    "llag": "LLAG",
+}
+
+TRIGGER_SOURCES = {
+    "immediate": "IMM",
+    "external": "EXT",
+    "internal1": "INT1",
+    "internal2": "INT2",
+    "internal3": "INT3",
+    "internal4": "INT4",
+    "output": "OUTP",
+    "bus": "BUS",
+}
+
+TRIGGER_SLOPES = {
+    "positive": "POS",
+    "negative": "NEG",
+}
+
+TACHOMETER_SLOPES = {
+    "positive": "POS",
+    "negative": "NEG",
+}
+
+ARM_SOURCES = {
+    "immediate": "IMM",
+    "manual": "MAN",
+    "rpm": "RPM",
+    "timer": "TIM",
+}
+
+ARM_RPM_MODES = {
+    "off": "OFF",
+    "up": "UP",
+    "down": "DOWN",
+}
+
+EXTERNAL_TRIGGER_RANGES = {
+    "high": "HIGH",
+    "low": "LOW",
+}
+
+TACHOMETER_TRIGGER_RANGES = {
+    "high": "HIGH",
+    "low": "LOW",
+}
+
+DISPLAY_FORMATS = {
+    "single": "SING",
+    "upper_lower": "ULOW",
+    "feedback": "FBAC",
+    "subl": "SUBL",
+    "upper_lower_feedback": "ULFB",
+    "quad": "QUAD",
+}
+
+DISPLAY_VIEWS = {
+    "trace": "TRAC",
+    "measurement_state": "MSTAT",
+    "mass_memory": "MMEM",
+    "state_table": "STAB",
+    "calc_table": "CTAB",
+    "frequency_table": "FTAB",
+    "time_table": "TTAB",
+    "memory": "MEM",
+    "option": "OPT",
+    "capture": "CAPT",
+    "instrument_state": "IST",
+    "message": "MESS",
+}
+
+DISPLAY_PROGRAM_MODES = {
+    "off": "OFF",
+    "full": "FULL",
+    "upper": "UPP",
+    "lower": "LOW",
+}
+
+TRACE_X_AUTOSCALE = {
+    "off": "OFF",
+    "once": "ONCE",
+}
+
+TRACE_Y_AUTOSCALE = {
+    "off": "OFF",
+    "on": "ON",
+    "once": "ONCE",
+}
+
+DISPLAY_AXIS_SPACING_VALUES = {
+    "linear": "LIN",
+    "logarithmic": "LOG",
+}
+
+DISPLAY_TRACE_Y_REFERENCE_VALUES = {
+    "top": "TOP",
+    "center": "CENT",
+    "bottom": "BOTT",
+    "range": "RANG",
+}
+
+TRACE_AMPLITUDE_UNITS = ("PEAK", "PP", "RMS")
+
+TRACE_ANGLE_UNITS = {
+    "degree": "DEG",
+    "radian": "RAD",
+}
+
+TRACE_X_UNITS = {
+    "hertz": "HZ",
+    "cycles_per_minute": "CPM",
+    "orders": "ORD",
+    "user": "USER",
+}
+
+DB_REFERENCE_UNITS = {
+    "dbv": "DBV",
+    "dbm": "DBM",
+    "dbspl": "DBSPL",
+    "dbuser": "DBUSER",
+    "user": "USER",
+}
+
+TRACE_MECHANICAL_UNITS = (
+    "G",
+    "M/S2",
+    "M/S",
+    "M",
+    "INCH/S2",
+    "INCH/S",
+    "INCH",
+    "MILS",
+)
+
+TRACE_VOLTAGE_UNITS = (
+    "V",
+    "V2",
+    "V/RTHZ",
+    "V2/HZ",
+    "V2S/HZ",
+)
+
+MARKER_FUNCTION_VALUES = {
+    "off": "OFF",
+    "harmonic_power": "HPOW",
+    "thd": "THD",
+    "band_power": "BPOW",
+    "band_rms": "BRMS",
+    "spectrum_power": "SPOW",
+    "overshoot": "OVER",
+    "rise_time": "RTIM",
+    "settling_time": "STIM",
+    "delay_time": "DTIM",
+    "steady_state_level": "SSL",
+    "gain_margin": "GMAR",
+    "phase_margin": "PMAR",
+    "gain_crossover": "GCR",
+    "phase_crossover": "PCR",
+    "frequency": "FREQ",
+    "damping": "DAMP",
+    "s_info": "SINF",
+    "weight": "WEIG",
+    "total_power": "TPOW",
+    "weighted_power": "WPOW",
+}
+
+MARKER_MODE_VALUES = {
+    "absolute": "ABS",
+    "relative": "REL",
+}
+
+MATH_SELECTED_FUNCTION_VALUES = ("F1", "F2", "F3", "F4", "F5")
+SYNTHESIS_SPACING_VALUES = {
+    "linear": "LIN",
+    "logarithmic": "LOG",
+}
+SYNTHESIS_TABLE_TYPE_VALUES = {
+    "pole_zero": "PZER",
+    "pole_fraction": "PFR",
+    "polynomial": "POLY",
+}
+
+BOOL_VALUES = {True: 1, False: 0}
+STATUS_MASK_VALUES = [0, 32767]
+DATA_RANGE_VALUES = [1.0e-20, 1.0e20]
+LIMIT_VALUE_RANGE = [-9.9e37, 9.9e37]
+SENSE_FEED_VALUES = {
+    "input": "INP",
+    "time_capture": "TCAP",
+}
+FREQUENCY_RESOLUTION_OCTAVE_VALUES = {
+    "third": "THIR",
+    "full": "FULL",
+    "twelfth": "TWEL",
+}
+FREQUENCY_SPAN_LINK_VALUES = {
+    "start": "STAR",
+    "center": "CENT",
+}
+SWEEP_DIRECTION_VALUES = {
+    "up": "UP",
+    "down": "DOWN",
+}
+SWEEP_MODE_VALUES = {
+    "auto": "AUTO",
+    "manual": "MAN",
+}
+SWEEP_SPACING_VALUES = {
+    "linear": "LIN",
+    "logarithmic": "LOG",
+}
+AUTORANGE_DIRECTION_VALUES = {
+    "up": "UP",
+    "either": "EITH",
+}
+INPUT_RANGE_TRANSDUCER_LABELS = (
+    "PA",
+    "G",
+    "M/S2",
+    "M/S",
+    "M",
+    "KG",
+    "N",
+    "DYN",
+    "INCH/S2",
+    "INCH/S",
+    "INCH",
+    "MIL",
+    "LB",
+    "USER",
+)
+SERIAL_BAUD_RATES = (300, 1200, 2400, 4800, 9600)
+SERIAL_BITS = (5, 6, 7, 8)
+SERIAL_STOP_BITS = (1, 2)
+SERIAL_RECEIVE_PACE_VALUES = {
+    "none": "NONE",
+    "xon": "XON",
+}
+SERIAL_PARITY_VALUES = {
+    "none": "NONE",
+    "even": "EVEN",
+    "odd": "ODD",
+}
+SERIAL_TRANSMIT_PACE_VALUES = {
+    "none": "NONE",
+    "xon": "XON",
+    "dsr": "DSR",
+}
+DATA_FORMAT_VALUES = {
+    "ascii": "ASC",
+    "real": "REAL",
+}
+FAN_STATE_VALUES = {
+    "off": "OFF",
+    "full": "FULL",
+    "auto": "AUTO",
+}
+SYSTEM_DATE_VALUES = ([0, 9999], [1, 12], [1, 31])
+SYSTEM_TIME_VALUES = ([0, 23], [0, 59], [0, 60])
+HARDCOPY_DEVICE_LANGUAGE_VALUES = ("HPGL", "PCL", "PHPGL")
+HARDCOPY_PAGE_POINT_VALUES = ([-32767, 32767], [-32767, 32767])
+HARDCOPY_DESTINATION_VALUES = ("MMEM", "SYST:COMM:GPIB:RDEV", "SYST:COMM:CENT", "SYST:COMM:SER")
+HARDCOPY_SOURCE_VALUES = ("ALL", "TRAC", "MARK", "REF", "GRID")
+HARDCOPY_LINE_TYPE_STYLE_RANGE = [0, 12]
+HARDCOPY_LABEL_COLOR_RANGE = [0, 16]
+HARDCOPY_TIMESTAMP_FORMAT_VALUES = ("FORM1", "FORM2", "FORM3", "FORM4", "FORM5")
+MEMORY_CATALOG_ITEM_VALUES = {
+    "time_capture": "TCAPture",
+    "waterfall": "WATerfall",
+    "waterfall_register": "WREGister",
+    "program": "PROGram",
+    "ram_disk": "RDISk",
+    "limits": "LIMits",
+}
+MASS_MEMORY_STORE_PROGRAM_FORMAT_VALUES = {
+    "ascii": "ASC",
+    "binary": "BIN",
+}
+MASS_MEMORY_STORE_TRACE_FORMAT_VALUES = ("SDF", "ASCII")
+PROGRAM_STATE_VALUES = {
+    "stop": "STOP",
+    "pause": "PAUSe",
+    "run": "RUN",
+    "continue": "CONTinue",
+}
+
+
+def _parse_ascii_floats(reply: str) -> list[float]:
+    """Parse a comma-separated ASCII response into a list of floats."""
+    if reply.strip() == "":
+        return []
+    values = []
+    for token in reply.split(","):
+        stripped = token.strip()
+        if stripped == "":
+            continue
+        values.append(float(stripped))
+    return values
+
+
+def _parse_ascii_ints(reply: str) -> list[int]:
+    """Parse a comma-separated ASCII response into a list of ints."""
+    if reply.strip() == "":
+        return []
+    values = []
+    for token in reply.split(","):
+        stripped = token.strip()
+        if stripped == "":
+            continue
+        values.append(int(stripped))
+    return values
+
+
+def _parse_csv_strings(reply: str) -> list[str]:
+    """Parse a CSV response into a list of unquoted strings."""
+    if reply.strip() == "":
+        return []
+    row = next(reader(StringIO(reply), skipinitialspace=True))
+    return [_strip_quotes(item.strip()) for item in row]
+
+
+def _strip_quotes(reply: str) -> str:
+    """Strip matching single or double quotes from string boundaries."""
+    value = reply.strip()
+    if len(value) >= 2 and value[0] == value[-1] and value[0] in {"'", '"'}:
+        return value[1:-1]
+    return value
+
+
+def _quote_string(value: str) -> str:
+    """Quote a string for SCPI arguments unless already quoted."""
+    if not isinstance(value, str):
+        raise TypeError("SCPI string arguments must be of type str.")
+    stripped = value.strip()
+    if len(stripped) >= 2 and stripped[0] == stripped[-1] and stripped[0] in {"'", '"'}:
+        return stripped
+    escaped = value.replace('"', '""')
+    return f'"{escaped}"'
+
+
+def _require_confirmation(action: str, confirmed: bool):
+    """Require explicit confirmation for risky actions."""
+    if confirmed is not True:
+        raise ValueError(
+            f"Action '{action}' requires explicit confirmation. "
+            "Pass confirmed=True to execute."
+        )
+
+
+def _trace_selector(trace):
+    """Return a valid TRACe selector token."""
+    if isinstance(trace, int):
+        if trace not in (1, 2, 3, 4):
+            raise ValueError("Trace index must be in the range 1..4.")
+        return f"TRACe{trace}"
+    return str(trace)
+
+
+def _tcap_selector(capture_channel):
+    """Return a valid TCAP selector token."""
+    if isinstance(capture_channel, int):
+        if capture_channel not in (1, 2, 3, 4):
+            raise ValueError("Capture channel index must be in the range 1..4.")
+        return f"TCAP{capture_channel}"
+
+    token = str(capture_channel).strip().upper()
+    if token in {"TCAP1", "TCAP2", "TCAP3", "TCAP4"}:
+        return token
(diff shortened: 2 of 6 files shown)


tests/instruments/keysight/test_keysight35670A_with_device.py
```diff
@@ -0,0 +1,291 @@
...
+            analyzer.source_frequency = original_frequency
+        finally:
+            try:
+                analyzer.source_voltage_offset = original_offset
+            finally:
+                try:
+                    analyzer.source_function = original_function
+                finally:
+                    analyzer.source_output_enabled = False
+                    assert analyzer.source_output_enabled is False
```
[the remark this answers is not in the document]
PeterPirog: Addressed. The source-output assertion was moved out of the cleanup path, the system-error helper is now reused, and cleanup exceptions now emit warnings instead of being silently ignored.

tests/instruments/keysight/test_keysight35670A_with_device.py
```diff
@@ -0,0 +1,295 @@
...
+            try:
+                adapter.close()
+            except Exception as exc:
+                warnings.warn(
+                    f"Ignored error during analyzer teardown cleanup: {exc!r}",
+                    RuntimeWarning,
+                )
+            finally:
+                if hasattr(adapter, "connection"):
+                    adapter.connection = None
```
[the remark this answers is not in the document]
PeterPirog: Addressed. The hardware fixture now captures the adapter immediately, drains teardown errors without calling pytest.fail before cleanup, closes the adapter before reporting teardown failures, and keeps normal cleanup behavior for the safe hardware tests.

docs/api/instruments/keysight/keysight35670A.rst
```diff
@@ -0,0 +1,44 @@
+##########################################
+Keysight 35670A Dynamic Signal Analyzer
+##########################################
```
BenediktBurger, line 3: Please keep the hashes as long as the text

docs/api/instruments/keysight/keysight35670A.rst
```diff
@@ -0,0 +1,44 @@
...
+
+    from pymeasure.instruments.keysight import Keysight35670A
+
+The Keysight/Agilent/HP 35670A driver provides a dynamic signal analyzer API for
+measurement and configuration workflows.
+It supports 2 or 4 input channels depending on option AY6, trace boxes 1..4,
+source-output control, and block transfer helpers for binary/definite-block
+data paths.
+Destructive memory, system, and program operations require explicit
+confirmation arguments.
```
BenediktBurger, line 15: Please move this comment to the docstring of the instrument.
That way, everyone can see it; either in the instrument's docstring or here on the API (which shows the docstring as well).

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    "total_power": "TPOW",
+    "weighted_power": "WPOW",
+}
+
+MARKER_MODE_VALUES = {
+    "absolute": "ABS",
+    "relative": "REL",
+}
+
+MATH_SELECTED_FUNCTION_VALUES = ("F1", "F2", "F3", "F4", "F5")
```
BenediktBurger, line 324: If you use it only once, you do not need to create a variable for it, you could just write:
~~~python
values=("F1", "F2"...)
~~~
in the property creator.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    """Parse a comma-separated ASCII response into a list of floats."""
+    if reply.strip() == "":
+        return []
+    values = []
+    for token in reply.split(","):
+        stripped = token.strip()
+        if stripped == "":
+            continue
+        values.append(float(stripped))
+    return values
```
BenediktBurger, line 450: You know that the `Instrument.values` method does call `ask` and afterwards splits the response at comma (or another separator). Finally it casts every member to a specific type.
So `Instrument.values` can return a list of float or int, depending on the cast type (or any other type you give it).

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+
+def _quote_string(value: str) -> str:
+    """Quote a string for SCPI arguments unless already quoted."""
+    if not isinstance(value, str):
+        raise TypeError("SCPI string arguments must be of type str.")
+    stripped = value.strip()
+    if len(stripped) >= 2 and stripped[0] == stripped[-1] and stripped[0] in {"'", '"'}:
+        return stripped
+    escaped = value.replace('"', '""')
```
BenediktBurger, line 489: why do you replace a single quote with two ones? Does that work as expected?

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        return f"TRACe{trace}"
+    return str(trace)
+
+
+def _tcap_selector(capture_channel):
+    """Return a valid TCAP selector token."""
+    if isinstance(capture_channel, int):
+        if capture_channel not in (1, 2, 3, 4):
+            raise ValueError("Capture channel index must be in the range 1..4.")
+        return f"TCAP{capture_channel}"
```
BenediktBurger, line 516: You are repeating yourself (same code as trace_selector).

Actually, we have validators which you could use easily: They take the value and possible values.
Probably they could serve you well for verification, here: `strict_range(capture_channel, range(1,5))`

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    validated = strict_range(value, values)
+    if float(validated) == 0.0:
+        raise ValueError("Value must not be 0.")
+    return validated
+
+
+def _normalize_on_off_once(value: str) -> str:
+    """Normalize ON/OFF/ONCE tokens returned as symbolic or numeric values."""
+    token = _strip_quotes(str(value)).strip().upper()
+    return {"0": "OFF", "1": "ON"}.get(token, token)
```
BenediktBurger, line 552: Why don't you map on/off to True/False?

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    if "CENT" in token:
+        return "SYST:COMM:CENT"
+    if "SER" in token:
+        return "SYST:COMM:SER"
+    raise ValueError(f"Invalid hardcopy destination '{value}'.")
+
+
+def _validate_hardcopy_destination(value, values):
+    """Validate hardcopy destination token."""
+    del values
```
BenediktBurger, line 639: values exists only inside this method, no need to delete it. It will be deleted once the method returns.

In the end, you could replace _validate with _normalize...

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    if suffix not in {"1", "2", "3", "4", "5"}:
+        raise ValueError(f"Invalid hardcopy timestamp format '{value}'.")
+    return f"FORM{suffix}"
+
+
+def _validate_hardcopy_timestamp_format(value, values):
+    """Validate hardcopy timestamp format token."""
+    del values
+    return _normalize_hardcopy_timestamp_format(value)
```
BenediktBurger, line 661: See above.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    token = _strip_quotes(str(value)).strip().upper().replace(" ", "")
+    if token in {"MMEM", "MMEMORY"}:
+        return "MMEM"
+    if "GPIB" in token and "RDEV" in token:
+        return "SYST:COMM:GPIB:RDEV"
+    if "CENT" in token:
+        return "SYST:COMM:CENT"
+    if "SER" in token:
+        return "SYST:COMM:SER"
+    raise ValueError(f"Invalid hardcopy destination '{value}'.")
```
BenediktBurger, line 634: Why don't you just allow a single version for the user, that will be easier.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        values=BOOL_VALUES,
+        map_values=True,
+        cast=int,
+        validator=strict_discrete_set,
+    )
+
+    @property
+    def state(self):
+        """Control whether the input channel is enabled."""
+        return type(self)._state.fget(self)
```
BenediktBurger, line 971: Why so complicated:

```suggestion
        return self._state
```

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    @state.setter
+    def state(self, value):
+        enabled = bool(strict_discrete_set(value, BOOL_VALUES))
+        if self.id == 1 and not enabled:
+            raise ValueError("Input channel 1 cannot be disabled.")
+        type(self)._state.fset(self, enabled)
```
BenediktBurger, line 978: Here again, too complicated:

```suggestion
        self._state = enabled
```

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    enabled = state
```
BenediktBurger, line 980: Please choose one name (enabled is better than state).

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    @property
+    def state(self):
+        """Control whether the input channel is enabled."""
+        return type(self)._state.fget(self)
+
+    @state.setter
+    def state(self, value):
+        enabled = bool(strict_discrete_set(value, BOOL_VALUES))
+        if self.id == 1 and not enabled:
+            raise ValueError("Input channel 1 cannot be disabled.")
```
BenediktBurger, line 977: I see that you want to add this check, therefore the work with _state and state.
Better use a dynamic property and set the allowed values differently in this channel's init depending on the id.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        validator=strict_range,
+        cast=int,
+    )
+
+    hardcopy_trace_limit_line_type = Channel.control(
+        "HCOPy:ITEM:WINDow{ch}:TRACe:LIMit:LTYPe?",
+        "HCOPy:ITEM:WINDow{ch}:TRACe:LIMit:LTYPe %s",
+        """Control the hardcopy line type used for limit lines.""",
+        values=(),
+        validator=_validate_hardcopy_line_type,
```
BenediktBurger, line 1291: If you use a custom validator, which does not care about values, you could just not define values at all.

However, you can use default validators and specify possible values here, instead of your custom validators.

Also for methods below.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        "DISPlay:WINDow{ch}:WATerfall:TOP %g",
+        """Control the waterfall Z-axis top value.""",
+        values=[0.0, 9.9e37],
+        validator=strict_range,
+        cast=float,
+    )
+
+    def x_match(self, source_window) -> None:
+        """Match this window X-axis scaling to another display window."""
+        source = int(strict_range(int(source_window), [1, 4]))
```
BenediktBurger, line 1504: Why do you do the casting to int and why twice?
Do you expect someone to give something else than an int? If you give a type hint, users will know that only ints are allowed.

The outer cast to int is utterly useless: The validator returns the input valute (already forced to int) if it is valid, otherwise it raises an error. No chance that it returns something differeent than an int.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        self.write(f"DISPlay:WINDow{{ch}}:TRACe:X:MATCh{source}")
+
+    def y_match(self, source_window) -> None:
```
BenediktBurger, line 1507: See above

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        """Match this window Y-axis scaling to another display window."""
+        source = int(strict_range(int(source_window), [1, 4]))
+        self.write(f"DISPlay:WINDow{{ch}}:TRACe:Y:MATCh{source}")
+
+
+class Keysight35670AWindow(Keysight35670ADisplayWindow):
```
BenediktBurger, line 1513: Is it necessary to create that subclass without its own code?

Could you use a generic KeysightWindow, maybe with some series name like Keysight356SeriesWindow?
Then we would not need this extra class.

This is just a suggestion, maybe there are reasons for different WindowClasses.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    )
+
+
+class Keysight35670ATrace(Channel):
+    """Represent a trace box of the Keysight 35670A."""
+
+    active = Channel.control(
+        "CALCulate{ch}:ACTive?",
+        "CALCulate{ch}:ACTive %s",
+        """Control which traces are active in the selected display group.""",
```
BenediktBurger, line 1546: Could you add some hints what values are possible to the docstring? A user will just see the docstring and not the code with the values argument.

The same is with the methods below.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        cast=float,
+    )
+
+    x_user_frequency_factor = Channel.control(
+        "CALCulate{ch}:UNIT:X:USER:FREQuency:FACTor?",
+        "CALCulate{ch}:UNIT:X:USER:FREQuency:FACTor %g",
+        (
+            """Control the user frequency-domain X-axis conversion factor """
+            """(float from 1e-15 to 1e15)."""
+        ),
```
BenediktBurger, line 1684: Triple quotes already allow multi line docstrings, so no need to group two multi-line-docstrings (of a single line each) together with parentheses. Actually, as it is already inside parentheses, the strings would be combined even without the parenthesis.

```suggestion
        """Control the user frequency-domain X-axis conversion factor
        (float from 1e-15 to 1e15).""",
```

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        values=[1e-15, 1e15],
+        validator=strict_range,
+        cast=float,
+    )
+
+    x_user_frequency_label = Channel.control(
+        "CALCulate{ch}:UNIT:X:USER:FREQuency:LABel?",
+        "CALCulate{ch}:UNIT:X:USER:FREQuency:LABel %s",
+        """Control the user frequency-domain X-axis unit label.""",
+        validator=lambda v, vs: v,
```
BenediktBurger, line 1694: That is the default validator, not validating anything, just remove that line.
Also, for that validator you do not need to set any value to `values`, so you can remove the line below as well.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        """Control whether the limit-fail beeper is enabled.""",
+        values=BOOL_VALUES,
+        map_values=True,
+        cast=int,
+        validator=strict_discrete_set,
+    )
+
+    limit_failed = Channel.measurement(
+        "CALCulate{ch}:LIMit:FAIL?",
+        """Measure whether the last limit test failed.""",
```
BenediktBurger, line 1733: In such a case I'd prefer `Get` over `Measure`. Exactly for such a reason we allowed both Get and Measure as sometimes one is more fitting sometimes the other one.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        cast=int,
+        get_process=bool,
+    )
+
+    lower_limit_segment = Channel.control(
+        "CALCulate{ch}:LIMit:LOWer:SEGMent?",
+        "CALCulate{ch}:LIMit:LOWer:SEGMent %s",
+        """Control lower limit line segments as (x0, y0, x1, y1) tuples.""",
+        validator=lambda v, vs: v,
+        values=(),
```
BenediktBurger, line 1743: Again superfluous.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        set_process=_format_limit_segment_data,
+        get_process=_parse_limit_segment_data,
+        cast=str,
+        maxsplit=0,
+    )
+
+    limit_enabled = Channel.control(
+        "CALCulate{ch}:LIMit:STATe?",
+        "CALCulate{ch}:LIMit:STATe %d",
+        """Control whether limit testing is enabled.""",
```
BenediktBurger, line 1753: Are limits enabled or limit testing? The name suggests that the limits themselves are enabled.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    def set_math_constant(
+        self,
+        constant_register: int,
+        real_part: float,
+        imaginary_part: Optional[float] = None,
+    ) -> None:
+        """Set a math constant register value."""
+        index = _math_register_selector(constant_register)
+        real_value = float(strict_range(float(real_part), LIMIT_VALUE_RANGE))
```
BenediktBurger, line 2220: again you are calling `float(float(number: float)`, just remove the casting to float (two times):

```suggestion
        real_value = strict_range(real_part, LIMIT_VALUE_RANGE)
```

Also below.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        if imaginary_part is None:
+            self.write(f"CALCulate{{ch}}:MATH:CONStant{index} {real_value:g}")
+            return
+
+        imaginary_value = float(strict_range(float(imaginary_part), LIMIT_VALUE_RANGE))
+        self.write(
+            f"CALCulate{{ch}}:MATH:CONStant{index} {real_value:g},{imaginary_value:g}"
+        )
+
+    def math_constant(self, constant_register: int) -> tuple[float, float]:
```
BenediktBurger, line 2230: This is the name of some property. I suggest you make this and above method (which is the setter for this constant, right) a property: I see that it is too complicated for a property creator like `control`, however, you could use a normal python property.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        """Read a math constant register as (real, imaginary)."""
+        index = _math_register_selector(constant_register)
+        values = _parse_ascii_floats(self.ask(f"CALCulate{{ch}}:MATH:CONStant{index}?"))
+        if not values:
+            raise ValueError("Math constant query returned no values.")
+        if len(values) == 1:
+            return values[0], 0.0
+        return values[0], values[1]
+
+    def set_math_expression(self, expression: str, function_register: int = 1) -> None:
```
BenediktBurger, line 2240: This should also be a property. For attributes/numbers/values etc. a property is best. For an action (something to do), a method is best.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        self.write(
+            f"CALCulate{{ch}}:MATH:EXPRession{index} {_quote_string(str(expression))}"
+        )
+
+    def math_expression(self, function_register: int = 1) -> str:
+        """Read the math expression in the selected function register."""
+        index = _math_register_selector(function_register)
+        return _strip_quotes(self.ask(f"CALCulate{{ch}}:MATH:EXPRession{index}?"))
+
+    def math_data(self, raw=False):
```
BenediktBurger, line 2252: A method does some action and  thus the name should indicate it, so it should contain some verb. Here again a property would be best as it is a value, not an action.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    def load_cfit_data(self, data, raw=False) -> None:
+        """Load complete curve-fit table block data."""
+        payload = _coerce_bytes(data)
+        block = payload if raw else _encode_definite_block(payload)
+        prefix = self.insert_id("CALCulate{ch}:CFIT:DATA ").encode("ascii")
+        self.write_bytes(prefix + block)
+
+    def run_curve_fit(self) -> None:
+        """Start a curve-fit operation and use wait_for_completion() to synchronize."""
```
BenediktBurger, line 2295: In this case, I would rename the method to `start_curve_fit`.
You mean, that you could use the `wait_for_completion` method to wait until you can read the result?
You can link to other methods (in the API documentation):

```suggestion
        """Start a curve-fit operation and use :meth:`wait_for_completion` to synchronize."""
```

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        """Copy the selected waterfall slice to a data register."""
+        selector = _data_register_selector(register)
+        self.write(f"CALCulate{{ch}}:WATerfall:SLICe:COPY {selector}")
+
+    def copy_waterfall_trace_to_register(self, register) -> None:
+        """Copy the selected waterfall trace to a data register."""
+        selector = _data_register_selector(register)
+        self.write(f"CALCulate{{ch}}:WATerfall:TRACe:COPY {selector}")
+
+    def data_points(self) -> int:
```
BenediktBurger, line 2342: This should be a property again.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        _require_confirmation("clear upper limit segment", confirmed)
+        value = float(strict_range(float(segment), LIMIT_VALUE_RANGE))
+        self.write(f"CALCulate{{ch}}:LIMit:UPPer:SEGMent:CLEar {value:g}")
+
+    def make_upper_limit_from_trace(self, confirmed=False) -> None:
+        """Convert the current trace into an upper limit line after explicit confirmation."""
+        _require_confirmation("convert trace to upper limit line", confirmed)
+        self.write("CALCulate{ch}:LIMit:UPPer:TRACe:IMMediate")
+
+    def marker_to_left_maximum(self) -> None:
```
BenediktBurger, line 2447: For an action you need a verb, `move_marker_to_left_maximum` describes that action.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    window1 = Instrument.ChannelCreator(Keysight35670ADisplayWindow, 1)
+    window2 = Instrument.ChannelCreator(Keysight35670ADisplayWindow, 2)
+    window3 = Instrument.ChannelCreator(Keysight35670ADisplayWindow, 3)
+    window4 = Instrument.ChannelCreator(Keysight35670ADisplayWindow, 4)
+
+    def __init__(self, adapter, name="Keysight 35670A Dynamic Signal Analyzer", **kwargs):
+        kwargs.setdefault("timeout", 10000)
+        kwargs.setdefault("read_termination", "\n")
+        kwargs.setdefault("write_termination", "\n")
```
BenediktBurger, line 2497: Can someone change the read termination or does it depend only on the device? In the latter case, just put it into the super init call.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    selected_instrument_number = Instrument.control(
+        "INSTrument:NSELect?",
+        "INSTrument:NSELect %d",
+        """Control the selected instrument mode by number.""",
+        values=[0, 5],
+        validator=strict_range,
+        cast=int,
+    )
+
+    _selected_instrument_number_setting = Instrument.setting(
```
BenediktBurger, line 2518: What should this setting do, what the control above cannot do?

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        """Control the source waveform shape.""",
+        values=SOURCE_FUNCTIONS,
+        map_values=True,
+        validator=strict_discrete_set,
+    )
+
+    source_frequency_cw = Instrument.control(
+        "SOURce:FREQuency?",
+        "SOURce:FREQuency %g",
+        """Control the sine source frequency in Hz using SOURce:FREQuency[:CW].""",
```
BenediktBurger, line 2678: You do not need to put the string that is sent to the device, into the docstring. That is the job of the driver you are writing right now.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    source_frequency_fixed = Instrument.control(
+        "SOURce:FREQuency:FIXed?",
+        "SOURce:FREQuency:FIXed %g",
+        """Control the sine source frequency in Hz using SOURce:FREQuency:FIXed.""",
+        values=[0.0, 115000.0],
+        validator=strict_range,
+        cast=float,
+    )
+
+    source_frequency = source_frequency_fixed
```
BenediktBurger, line 2693: Why this duplication?

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        cast=str,
+    )
+
+    source_voltage_reference_tolerance = Instrument.control(
+        "SOURce:VOLTage:LEVel:REFerence:TOLerance?",
+        "SOURce:VOLTage:LEVel:REFerence:TOLerance %g",
+        """Control the source autolevel sensitivity in dB.""",
+        values=[0.1, 20.0],
+        validator=strict_range,
+        cast=float,
```
BenediktBurger, line 2786: cast=float is the default.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    external_trigger_range = Instrument.control(
+        "TRIGger:EXTernal:RANGe?",
+        "TRIGger:EXTernal:RANGe %s",
+        """Control external trigger input range.""",
+        values=EXTERNAL_TRIGGER_RANGES,
+        map_values=True,
+        validator=strict_discrete_set,
+    )
+
+    trigger_start1 = Instrument.control(
```
BenediktBurger, line 2871: Should these be part of the channels?

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    mass_memory_load_continue_status = Instrument.measurement(
+        "MMEMory:LOAD:CONTinue?",
+        """Measure whether multi-disk mass-memory load continuation is requested.""",
+        cast=int,
+    )
+
+    mass_memory_store_continue_status = Instrument.measurement(
+        "MMEMory:STORe:CONTinue?",
+        """Measure whether multi-disk mass-memory store continuation is requested.""",
```
BenediktBurger, line 3364: whether and the cast=int do not match in my opinion.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    time_capture_length = Instrument.control(
+        "SENSe:TCAPture:LENGth?",
+        "SENSe:TCAPture:LENGth %g",
+        """Control the time-capture buffer length.""",
+        values=[0.0, 9.9e37],
+        validator=strict_range,
+        cast=float,
+    )
+
+    time_capture_start1 = Instrument.control(
```
BenediktBurger, line 3688: These properties should be part of a channel, then you do not have to repeat the same property.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        "SENSe:TCAPture:TACHometer:STATe?",
+        "SENSe:TCAPture:TACHometer:STATe %d",
+        """Control whether tachometer data is included in time capture.""",
+        values=BOOL_VALUES,
+        map_values=True,
+        cast=int,
+        validator=strict_discrete_set,
+    )
+
+    _installed_options = Instrument.measurement(
```
BenediktBurger, line 3779: Why do you mark it as private (with an underscore)?

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        "*OPT?",
+        """Get the installed option configuration.""",
+        cast=str,
+        maxsplit=0,
+    )
+
+    _next_system_error = Instrument.measurement(
```
BenediktBurger, line 3786: I think `next_error` (which is part of SCPIMixin) does already the same.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        "*SRE %d",
+        """Control the Service Request enable register.""",
+        values=[0, 255],
+        validator=strict_range,
+        cast=int,
+    )
+
+    status_byte = Instrument.measurement(
+        "*STB?",
+        """Measure the status byte register.""",
```
BenediktBurger, line 3840: "Get" would be more appropriate than Measure.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        """Control the time-step arming interval in seconds.""",
+        values=[0.0, 500000.0],
+        validator=strict_range,
+        cast=float,
+    )
+
+    calibration_auto = Instrument.control(
+        "CALibration:AUTO?",
+        "CALibration:AUTO %s",
+        """Control the autocalibration mode.""",
```
BenediktBurger, line 3892: It would be good to inicate the possible values.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        map_values=True,
+        validator=strict_discrete_set,
+        cast=str,
+        get_process=_normalize_on_off_once,
+    )
+
+    operation_condition = Instrument.measurement(
+        "STATus:OPERation:CONDition?",
+        """Measure the operation status condition register.""",
+        cast=int,
```
BenediktBurger, line 3903: Consider using IntFlag (from the standard lib enum package) for these status registers as it gives you easily which bits are set.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    def wait_for_completion(self) -> int:
+        """Wait for pending operations to complete."""
+        return self.operation_complete()
+
+    def arm(self) -> None:
+        """Arm the trigger when manual arming is selected."""
+        self.write("ARM:IMMediate")
+
+    def pass_control_back_address(
```
BenediktBurger, line 4279: In this case, you should add set to the name.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    def bode(self) -> None:
+        """Display a Bode diagram layout."""
+        self.write("DISPlay:BODE")
+
+    def display_error(self, text: str) -> None:
+        """Display analyzer-style text in the message area."""
+        self.write(f"DISPlay:ERRor {_quote_string(str(text))}")
+
+    def hardcopy_color_default(self) -> None:
```
BenediktBurger, line 4311: You need a verb for methods: reset_hardcopy_color (reset implies to default).

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+    def define_explicit_program(self, text, program="program1", raw=False) -> None:
+        """Define a selected Instrument BASIC program buffer from controller-provided block data."""
+        program_selector = _normalize_program_name(program)
+        block = _encode_program_block(text, raw=raw)
+        prefix = f"PROGram:EXPLicit:DEFine {program_selector},".encode("ascii")
+        self.write_bytes(prefix + block)
+
+    def read_explicit_program_definition(self, raw=False):
+        """Read an Instrument BASIC program definition via PROGram:EXPLicit:DEFine?."""
```
BenediktBurger, line 4345: You should tell what raw does. Either in the docstring or by renaming it to something like `return_raw` or skip_processing...

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        self.write("PROGram:EXPLicit:DEFine?")
+        reply = self.read_bytes(-1)
+        if raw:
+            return reply
+        return _parse_ascii_or_definite_block_text(reply)
+
+    def set_explicit_program_label(self, label: str, program="program1") -> None:
```
BenediktBurger, line 4352: If you already add type hints, please be consistent and add it to all parameters:

```suggestion
    def set_explicit_program_label(self, label: str, program: str = "program1") -> None:
```

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+            confirmed=False) -> None:
+        """Initialize a mass-memory disk after explicit confirmation."""
+        _require_confirmation("initialize mass-memory disk", confirmed)
+        parts = ["MMEMory:INITialize"]
+        if disk is not None:
+            parts.append(_quote_string(disk))
+        if filesystem is not None:
+            parts.append(_strip_quotes(str(filesystem)).strip().upper())
+        if format_option is not None:
+            parts.append(str(int(format_option)))
```
BenediktBurger, line 4461: Here format_option is only an int, so you do not need to cast to int again, also similarly the filesystem above is already a string.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        _require_confirmation("load synthesis table from mass memory", confirmed)
+        self.write(f"MMEMory:LOAD:SYNThesis {_quote_string(filename)}")
+
+    def mass_memory_load_time_capture(self, filename: str, confirmed=False) -> None:
+        """Load a time-capture file after explicit confirmation."""
+        _require_confirmation("load time-capture file from mass memory", confirmed)
+        self.write(f"MMEMory:LOAD:TCAPture {_quote_string(filename)}")
+
+    def mass_memory_load_trace(
+            self, register, filename: str, no_scale: Optional[bool] = None,
```
BenediktBurger, line 4536: We now use the direct (piped) way to indicate different possible types:

```suggestion
            self, register, filename: str, no_scale: bool | None = None,
```

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        trace_selector = _normalize_mass_memory_trace_selector(trace)
+        self.write(f"MMEMory:STORe:LIMit:UPPer:{trace_selector} {_quote_string(filename)}")
+
+    def mass_memory_store_math(self, filename: str, confirmed=False) -> None:
+        """Store a math-definition file after explicit confirmation."""
+        _require_confirmation("store math definitions to mass memory", confirmed)
+        self.write(f"MMEMory:STORe:MATH {_quote_string(filename)}")
+
+    def mass_memory_store_program(self, filename: str, confirmed=False) -> None:
+        """Store an Instrument BASIC program file after explicit confirmation."""
```
BenediktBurger, line 4581: I guess to store it on the mass memory store, right?

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+            if not values:
+                raise ValueError("Program key bracket query returned no values.")
+            return bool(values[-1])
+
+        state = int(strict_discrete_set(bool(enabled), BOOL_VALUES))
+        self.write(f"DISPlay:PROGram:KEY:BRACket {first},{last},{state}")
+        return None
+
+    def status_preset(self) -> None:
+        """Preset the status register masks."""
```
BenediktBurger, line 4645: What do you mean with "Preset the register"? I guess to reset it to the preset values, right? or to set it to preset values?

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        self.write("STATus:PRESet")
+
+    def pulse_user_status(self, mask: int) -> None:
+        """Pulse selected bits in the user status condition register."""
+        validated_mask = strict_range(mask, STATUS_MASK_VALUES)
+        self.write(f"STATus:USER:PULSe {int(validated_mask)}")
```
BenediktBurger, line 4651: The int is superfluous again, as validated_mask is already an int.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        """Measure the number of points available in a time-capture buffer."""
+        selector = _tcap_selector(capture_channel)
+        return int(self.ask(f"SENSe:DATA:HEADer:POINts? {selector}"))
+
+    def sense_data_range(self, capture_channel=1) -> float:
+        """Measure the configured range metadata for a time-capture buffer."""
+        selector = _tcap_selector(capture_channel)
+        return float(self.ask(f"SENSe:DATA:RANGe? {selector}"))
+
+    def set_sense_data_range(self, capture_channel, value) -> None:
```
BenediktBurger, line 4776: Should be a property.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+        raise TypeError("Waterfall data must be bytes, str, sequence, int, or float.")
+
+    def trace_data(self, trace=1) -> list[float]:
+        """Read trace data from a data register (compatibility wrapper)."""
+        return self.read_trace_raw_data(register=trace, raw=False)
+
+    def trace_x_data(self, trace=1) -> list[float]:
+        """Read trace X-axis data (compatibility wrapper)."""
+        del trace
```
BenediktBurger, line 4890: why do you delete the variable? Then you would not need to put it to the parameters in the first place!

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+    def trace_z_unit(self, trace=1) -> str:
+        """Read trace Z-axis unit (compatibility wrapper)."""
+        del trace
+        return self.read_trace_z_unit()
+
+    def options(self) -> str:
+        """Get the installed option configuration."""
+        return self._installed_options
+
+    def system_error(self) -> str:
```
BenediktBurger, line 4912: why this indirection instead of doing whatever you need directly?

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+        return self._next_system_error
+
+    def _raise_if_errors(self, max_errors=10) -> None:
+        """Raise runtime errors if the system error queue contains failures."""
+        if max_errors < 1:
+            raise ValueError("max_errors must be >= 1.")
+
+        errors = []
+        for _ in range(max_errors):
+            reply = self.ask("SYSTem:ERRor?")
```
BenediktBurger, line 4923: Here you should either use system_error or next_error instead of adding the string yet another time.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+                code = int(code_token)
+            except ValueError:
+                try:
+                    code = int(float(code_token))
+                except ValueError:
+                    code = None
+
+            message = ",".join(tokens[1:]).strip() if len(tokens) > 1 else ""
+            if code == 0 and "NO ERROR" in message.upper():
+                break
```
BenediktBurger, line 4938: next_error from SCPIMixin does already this logic, so no need to repeat it.

pymeasure/instruments/keysight/keysight35670A.py
```diff
@@ -0,0 +1,4969 @@
...
+
+            errors.append(reply)
+
+        if errors:
+            raise RuntimeError(f"Instrument error queue is not empty: {errors}")
+
+    def drain_errors(self, max_errors=10) -> None:
```
BenediktBurger, line 4945: check_errors from SCPIMixin does it already!

BenediktBurger (changes requested):
I went through the instrument. Sorry for the delay, but several thousand lines is quite something.
I will go later through the tests etc.

PR #1442, opened 2026-05-07
Update PTW UNIDOS with interval_measurement, typing, and docstrings

Falk3579:
- Rename intervall() to interval_measurement, adapt tests to renaming
- Remove docstrings from tests
- Divide the tests in classes
- Add tests for methods
- Add tests for JSON configuration
- Add typing
- Change the docstrings to follow the guidelines

### DIFF
--- a/pymeasure/instruments/ptw/ptwUNIDOS.py
+++ b/pymeasure/instruments/ptw/ptwUNIDOS.py
@@ -135,7 +135,8 @@ wrong format of the parameter",
 
         for n in range(len(err_txt)):
             if err_code & (2**n):
-                err_msg.append(err_txt[n])
+                if err_txt[n] is not None:
+                    err_msg.append(err_txt[n])
 
         return err_msg
 
--- a/pymeasure/instruments/ptw/ptwUNIDOS.py
+++ b/pymeasure/instruments/ptw/ptwUNIDOS.py
@@ -25,6 +25,7 @@
 
 import logging
 import json
+import warnings
 from typing import Any, Union, Optional
 
 from pymeasure.adapters import Adapter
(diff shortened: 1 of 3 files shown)


pymeasure/instruments/ptw/ptwUNIDOS.py
```diff
@@ -139,49 +143,49 @@ def errorflags_to_text(self, flags):
...
+    def hold(self) -> None:
         """Set the measurment to HOLD state.
 
         .. note:: Write permission is required.
         """
         self.ask("HLD")
 
-    def intervall(self, intervall=None):
-        """Execute an intervall measurement.
+    def interval_measurement(self, interval=None) -> None:
```
BenediktBurger:
```suggestion
    def interval_measurement(self, interval: Optional[int] = None) -> None:
```

pymeasure/instruments/ptw/ptwUNIDOS.py
```diff
@@ -139,49 +143,49 @@ def errorflags_to_text(self, flags):
...
+        """Execute an interval measurement.
 
-        :param int intervall: optional, measurement intervall in seconds
+        :param int interval: optional, measurement interval in seconds
```
BenediktBurger:
```suggestion
        :param interval: optional, measurement interval in seconds
```
If specified as a type hint, we do not need the type here again.

pymeasure/instruments/ptw/ptwUNIDOS.py
BenediktBurger: Thanks for improving the name. However, due to backwards compatibility, please keep the old name (additionally to the new one). The old one should use the new one and please add deprecation notes, see https://github.com/pymeasure/pymeasure/issues/672.

BenediktBurger (approved):
Thanks for your improvements.

pymeasure/instruments/ptw/ptwUNIDOS.py
Falk3579: Note added.

PR #1445, opened 2026-05-10
In workers, check should_stop after startup() before execute()

breakU1:
Fix workers.py: Add a simple return after startup() to skip the execution phase and jump to the finally part.
Documentation Update about use should_stop() in startup()


docs/tutorial/procedure.rst
```diff
@@ -127,7 +127,7 @@ At the top of the SimpleProcedure class we define the required Parameters. In th
...
+The :python:`execute` methods defines the main body of the procedure. Our example method consists of a loop over the number of iterations, in which we emit the data to be recorded (the Iteration number). The data is broadcast to any number of listeners by using the :code:`emit` method, which takes a topic as the first argument. Data with the :python:`'results'` topic and the proper data columns will be recorded to a file. The sleep function in our example provides two very useful features. The first is to delay the execution of the next lines of code by the time argument in units of seconds. The seconds is that during this delay time, the CPU is free to perform other code. Successful measurements often require the intelligent use of sleep to deal with instrument delays and ensure that the CPU is not hogged by a single script. After our delay, we check to see if the Procedure should stop by calling :python:`self.should_stop()`. By checking this flag, the Procedure will react to a user canceling the procedure execution. If your :python:`startup` method is very time-consuming, it is also recommended to query :python:`should_stop()` within it to allow for early cancellation.
```
BenediktBurger: While you are at it: Would you mind putting each sentence into its own line?
That will make it easier in the future to spot differences instead of comparing two whole paragraphs.

BenediktBurger (approved):
Thanks for your work.

PR #1447, opened 2026-05-13
ENH: Add Rigol DHO804 Oscilloscope

niklasschulz99:
### Summary
Added support for the Rigol DHO804 4 channel Osciloscope. The Code creation has been assisted by Claude Sonnet 4.6, but every line was read and verified by humans. 

### Tests
`pytest tests/instruments/rigol/test_rigol_dho804.py `
96 passed, 19 skipped, 0 failed

`pytest tests/instruments/rigol/test_rigol_dho804.py --device-address "USB0::6833::1101::DHO8<SERIAL>::0::INSTR"`
115 passed, 0 skipped, 0 failed

`pytest tests/instruments/rigol/test_rigol_dho804.py --device-address "TCPIP::<IP>::INSTR"`
115 passed, 0 skipped, 0 failed

(IP Adress and Serialnumber are redacted)

### Additional Information
In our groups ([PTT](https://etit.ruhr-uni-bochum.de/ptt/) and [PULS](https://www.puls.ruhr-uni-bochum.de/) at Ruhr University Bochum) we would like to shift our instrument control to pymeasure. This is the first device to test the process of contribution.


pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -0,0 +1,555 @@
...
+        self._set_waveform_source(channel)
+        self.write(f":WAV:MODE {mode}")
+        self.write(f":WAV:FORM {fmt}")
+    
+        pre = self.get_waveform_preamble(channel)
+    
+        if mode in ("MAX", "RAW"):
+            n_total = int(self.acquisition_memory_depth)
+        else:
+            n_total = pre['points']
```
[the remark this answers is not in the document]
niklasschulz99: The preamble always returns 1000 points regardless of the waveform mode. 
This was verified on the physical device:
```
    scope.write(":WAV:MODE MAX")
    scope.write(":WAV:SOUR CHAN1")
    print(scope.ask(":WAV:PRE?"))
    # 0,1,1000,1,4.000000E-7,...

    print(scope.ask(":ACQ:MDEP?"))
    # 1.0000E+04
```
The DHO804 preamble point count reflects the screen display points (1000), not the memory depth, in MAX and RAW mode. Therefore `:ACQ:MDEP?` is required to get the actual number of points available for transfer.

docs/api/instruments/rigol/rigol_dho804.rst
```diff
@@ -0,0 +1,10 @@
+##############################
+Rigol DHO804 Oscilloscope
+##############################
```
BenediktBurger: Please keep the hashes as long as the text.

pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -0,0 +1,625 @@
...
+)
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class DHO804Channel(Channel):
+    """A single analog input channel of the Rigol DHO804."""
+
+    display = Channel.control(
```
BenediktBurger: For boolean properties we typically use names like `display_enabled`, because that makes it clear what True/False mean: `display_enabled = True` is self-evident.

pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -0,0 +1,625 @@
...
+        ":CHAN{ch}:OFFS?",
+        ":CHAN{ch}:OFFS %g",
+        """Control the vertical offset in Volts (float).""",
+        cast=float,
+    )
+
+    probe = Channel.control(
+        ":CHAN{ch}:PROB?",
+        ":CHAN{ch}:PROB %g",
+        """Control the probe attenuation ratio (float, e.g. 1, 10, 100).""",
```
BenediktBurger: I'm thinking how we could make the possible values known to the user.

pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -0,0 +1,625 @@
...
+    )
+
+    label = Channel.control(
+        ":CHAN{ch}:LAB:CONT?",
+        ":CHAN{ch}:LAB:CONT %s",
+        """Control the label content shown for the channel (string, max
+        10 chars).""",
+    )
+
+    label_show = Channel.control(
```
BenediktBurger: Maybe `label_enabled`?

pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -0,0 +1,625 @@
...
+    ch_2 = Instrument.ChannelCreator(DHO804Channel, 2)
+    ch_3 = Instrument.ChannelCreator(DHO804Channel, 3)
+    ch_4 = Instrument.ChannelCreator(DHO804Channel, 4)
+
+    def __init__(self, adapter, name="Rigol DHO804", **kwargs):
+        super().__init__(adapter, name, **kwargs)
+
+    def wait_for_opc(self, timeout=10):
+        """Block until the oscilloscope reports operation complete."""
+        deadline = time.time() + timeout
```
BenediktBurger: For a timeout consider `monotonic` or `perf_counter`, as both are always increasing, while `time` depends on the timezone which changes during daylight saving, additional seconds, changing time settings etc.

pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -0,0 +1,625 @@
...
+            time.sleep(0.1)
+
+    def clear_status(self):
+        """Clear the event status register (CLS)."""
+        self.write("*CLS")
+
+    @property
+    def status_byte(self):
+        """Get the status byte (STB, int)."""
+        return int(self.ask("*STB?"))
```
BenediktBurger: You could use the normal property creator `measurement` as well.

I guess the status byte has some meaning: Consider using an [IntFlag](https://docs.python.org/3/library/enum.html#enum.IntFlag), which will easily show which bits of the status byte are set and what they mean.

pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -0,0 +1,625 @@
...
+            100_000_000,
+            200_000_000,
+        ],
+    )
+
+    @property
+    def sample_rate(self):
+        """Get the current sample rate in Sa/s (read-only, float)."""
+        return float(self.ask(":ACQ:SRAT?"))
+
```
BenediktBurger: For read only properties we have `Instrument.measurement` (a special form of `control`), and for write only `setting`.

pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -0,0 +1,625 @@
...
+    trigger_holdoff = Instrument.control(
+        ":TRIG:HOLD?",
+        ":TRIG:HOLD %g",
+        """Control the trigger holdoff time in seconds (float).""",
+        cast=float,
+    )
+
+    @property
+    def trigger_status(self):
+        """Get the current trigger status string (read-only).
```
BenediktBurger: We have the convention: "Control" = get and set, "Get" or "Measure" = read-only, "Set" = write only.

So your addition (read-only) is redundant.

pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -0,0 +1,625 @@
...
+        self.write(":DISP:CLE")
+
+    display_type = Instrument.control(
+        ":DISP:TYPE?",
+        ":DISP:TYPE %s",
+        """Control the waveform display type: ``"VECT"`` (vector) or
+        ``"DOTS"``.""",
+        validator=strict_discrete_set,
+        values=["VECT", "DOTS"],
+    )
```
BenediktBurger: For properties with two options I typically recommend a boolean one: `display_type_vector_enabled` with True/False (you can use the `map_values` argument). That makes it easier than string values.

tests/instruments/rigol/test_rigol_dho804.py
```diff
@@ -0,0 +1,722 @@
...
+    parser.addoption(
+        "--device-address",
+        action="store",
+        default=None,
+        help="VISA resource string, e.g. 'TCPIP::10.0.0.100::INSTR'",
+    )
+
+
+@pytest.fixture(scope="module")
+def real_instrument(request):
```
BenediktBurger: Tests with an actual device (in contrast to protocol tests), should be in their own file with the `_with_device` suffix.

Also, we already have a fixture which gets the device address from the command line and skips the tests otherwise.

For more information regarding tests with devices see: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#device-tests

tests/instruments/rigol/test_rigol_dho_base.py
BenediktBurger, line 1: Thanks for adding protocol tests (very helpful for maintenance later) and device tests.

BenediktBurger (changes requested):
Thank you very much for your contribution of a new instrument.

You are very welcome to use pymeasure in your lab.
We're glad if people share their work which other can use as well.

Note that you can use pymeasure as a library for measurement instrument drivers while using other software to do the actual data acquisition (I, for example, had code written with PyLECO, others use PyMoDAQ etc.).

If you have questions feel free to ask.

pymeasure/instruments/rigol/rigol_dho804.py
niklasschulz99: Thanks for the tip! I wasn't aware of that, learned something new today. I've updated it to use `time.monotonic()`. 👍

pymeasure/instruments/rigol/rigol_dho804.py
niklasschulz99: I've refactored the status byte to use PyMeasure's native `measurement` property creator. I also implemented `IntFlag` enums for both the Status Byte (STB) and the Standard Event Register (ESR) to make the individual bits clearly readable.

tests/instruments/rigol/test_rigol_dho804.py
niklasschulz99: All device tests are now in their own file. I also changed to the recommended fixture.

tests/instruments/rigol/test_rigol_dho_base.py
niklasschulz99: You're welcome! All tests have been updated according to the changes and are passing successfully both with and without the physical device. Thank you very much for the great review and feedback!

Regarding the CI build: It seems some tests are failing, but they appear to be unrelated to my changes (am I right?). The failure occurs in the post-install setup for PyMeasure requirements on the Python 3.10 / Windows-latest runner.

pymeasure/instruments/rigol/rigol_dho804.py
niklasschulz99: I agree, making these values visible to the user would be a great improvement. I'd propose extracting the long list into a class constant (e.g., `PROBE_ATTENUATIONS`) and referencing it in the docstring. This keeps the property code clean and allows the user to easily inspect the values. 

Additionally, we could consider switching the validator from strict_discrete_set to truncated_discrete_set so that PyMeasure automatically snaps to the closest valid value if the user provides a slight mismatch.

Let me know what you think, and I can implement it! 👍

tests/instruments/rigol/test_rigol_dho_base.py
BenediktBurger: Yes, a problem with micromamba, we already fixed it on main.

pymeasure/instruments/rigol/rigol_dho804.py
niklasschulz99: Hi @BenediktBurger,
just checking in to see whether you’d like me to proceed with the proposed changes or if you have any additional feedback.
Happy to adjust things as needed.

BenediktBurger: Yes, use a list or enum is the best. With truncation you never know what you get. 0.015 will be 0.02, but you won't know it.

niklasschulz99: Thanks for the clarification! I added the list as a class constant and it is referenced in the docstring.

sjgallagher2:
FYI this also supports DHO1000 and DHO4000 series scopes. Maybe it could be named something more generic, or a parent class could be created to be subclassed by particular models

pymeasure/instruments/rigol/rigol_dho804.py
```diff
@@ -241,10 +259,11 @@ def status_byte(self):
...
     )
 
-    @property
-    def sample_rate(self):
-        """Get the current sample rate in Sa/s (read-only, float)."""
-        return float(self.ask(":ACQ:SRAT?"))
+    sample_rate = Instrument.measurement(
+        ":ACQ:SRAT?",
+        """Get the current sample rate in Sa/s (float).""",
+        cast=float
```
BenediktBurger: `cast=float` is the default, so you do not need to specify it.

tests/instruments/rigol/test_rigol_dho_base_with_device.py
```diff
@@ -0,0 +1,129 @@
...
+@pytest.fixture(scope="module")
+def scope(connected_device_address):
+    """Connect to a real DHO804; skip if --device-address is not given."""
+    instr = DHO804(connected_device_address)
+    return instr
+
+
+# ======================================================================= #
+#  Live hardware tests                                                     #
+# ======================================================================= #
```
BenediktBurger, line 38:
```suggestion
# ======================================================================= #
#  Live hardware tests                                                    #
# ======================================================================= #
```

BenediktBurger (approved):
> FYI this also supports DHO1000 and DHO4000 series scopes. Maybe it could be named something more generic, or a parent class could be created to be subclassed by particular models

In this case I would rename it to something more generic. Subclasses are only useful if they differ.

Thanks for your work.

pymeasure/instruments/rigol/rigol_dho804.py
niklasschulz99: Thanks, I removed it!

niklasschulz99:
> > FYI this also supports DHO1000 and DHO4000 series scopes. Maybe it could be named something more generic, or a parent class could be created to be subclassed by particular models
> 
> In this case I would rename it to something more generic. Subclasses are only useful if they differ.
> 
> Thanks for your work.

From what I can tell, the DHO900, DHO1000, and DHO4000 series are also compatible with this class, as they share the same underlying SCPI command set for all standard functions.

However, looking at the respective programming guides, there are differences (Bus decoders, bit-depth config, extended channel command for impedance/bias voltage/probe calibration/..., digital channels, ...).

I would therefore suggest renaming this class to `DHOSeries` or `DHOBase`, as it covers the large common subset of functionality shared across all DHO models. Anyone with a specific device who wants to add the model-specific features (additional decoders, digital channels, etc.) could then create a subclass that extends this base.

In short: this class handles the standard functionality well, but each model has its own quirks that would justify separate subclasses down the road.

Is there any preference on the naming, or a different idea on how to structure this?

BenediktBurger:
DHOBase is good as it shows that there are actual implementations. DHOSeries would imply that it works for several devices without explicit subclasses (e.g. same driver works for several instruments without changes).

Sorry for the broken tests: The `values` method (used internally) would try to cast to float and return a string if it fails. Now during test runs, it is an error to rely on that silent fallback mechanism. If you do not want to cast to float, give a `cast=str`, or `cast=cast_or_str(float)` to return a float or str.

pymeasure/instruments/rigol/rigol_dho_base.py
```diff
@@ -0,0 +1,637 @@
...
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    coupling = Channel.control(
+        ":CHAN{ch}:COUP?",
+        ":CHAN{ch}:COUP %s",
+        """Control the input coupling: ``"AC"``, ``"DC"``, or ``"GND"``.""",
+        validator=strict_discrete_set,
+        values=["AC", "DC", "GND"],
```
BenediktBurger, line 94: This method (and the bandwidth_limit and some others) cause the pipeline to break.
If you want to read strings, you have to specify `cast=str`.

```suggestion
        values=["AC", "DC", "GND"],
        cast=str,
```

niklasschulz99: Thank you for the feedback! I added `cast=str` to `coupling`, `label`, `trigger_sweep`, `trigger_slope`. I also renamed DHO804 to DHOBase in device class, docs and tests.

BenediktBurger:
Thank you for your work and your patience.

PR #1456, opened 2026-05-20
Update Agilent B298x with type hints

Falk3579:
- Remove unused logging import
- add type hints


pymeasure/instruments/agilent/agilentB298x.py
```diff
@@ -63,9 +60,10 @@ class AgilentB2981(SCPIMixin, Instrument):
 
     The B2981 is a Femto/Picoammeter."""
 
-    def __init__(self, adapter,
-                 name="Agilent/Keysight B2980A/B series",
-                 **kwargs):
+    def __init__(self,
+                 adapter: Union[Adapter, int, str],
```
BenediktBurger: With the most recent python change, you can use:

```suggestion
                 adapter: Adapter | int | str,
```

No need to use Union/Optional anymore.

BenediktBurger (approved):
Thanks!

PR #1460, opened 2026-05-23
Change CI linter to ruff

BenediktBurger:


BenediktBurger:
With output set to github, github can generate annotations on the code like flake8 does.

bilderbuchi (approved):

PR #1464, opened 2026-06-01
Add AI guidelines to contribute.rst

### LINKED ISSUE #1369: How to deal with AI-generated code contributions
> Another AI-related topic that we should probably formulate a stance on is AI-generated code contributions. It seems already the first PRs for this are coming in #1325.
> 
> It will be quite easy for people to generate huge PRs (+3000 lines in #1325) with AI help, but it's not easier for us to _review_ those, and we are manpower-overstretched already and behind on reviews. AI (slop or not) will only make that worse.
>  
> This compounds the problem of identifying code quality issues hiding in lots of easily-generated code. Just from a quick browse, I found patterns like 
> ```python
>     # Set timebase delay to 0
>     sds1000xhd.timebase.delay = 0
> ```
> ("comment why, not what" is not observed - the comment just repeats what the code does, adding no additional information)
> I can't even tell if this comes from AI or a human.
>  

 _Originally posted by @bilderbuchi in [#1362](https://github.com/pymeasure/pymeasure/issues/1362#issuecomment-3430634144)_

> I'm not against AI generated code, as it will happen and it can be a great support.
> However, we should make it clear, that we may refuse a proper review, if the code is not in line with at least certain standards. Maybe, we should add a note (in contribution) to also mention that you should think about the amount of a driver added, such that the people do not add drivers for the sake of adding them (like "here are ten manuals, translate them to drivers, even though I do not have such an instrument").

_Originally posted by @BenediktBurger in [#1362](https://github.com/pymeasure/pymeasure/issues/1362#issuecomment-3430829899)_

BenediktBurger:
Closes #1369

### DIFF
--- a/docs/dev/contribute.rst
+++ b/docs/dev/contribute.rst
@@ -105,9 +105,36 @@ There is some automation in place to run the unit tests and check some coding st
 Your pull-request will be reviewed by the PyMeasure maintainers. Frequently there is some iteration and discussion based on that feedback until a pull request can be merged. This will happen either in the conversation tab or in inline code comments.
 
 Be aware that due to maintainer manpower limitations it might take a long time until PRs get reviewed and/or merged.
-In general, review effort scales badly with PR size. Therefore, **smaller PRs are much preferred**. Try to limit your contribution to one "aspect", e.g. one instrument (or a few if closely related), one bug fix, or one feature contribution.
+In general, review effort scales badly with PR size. Therefore, **smaller PRs are much preferred** — see :ref:`contribution-standards` for details.
 
 If you placed your contribution in a separate branch as suggested above, you can easily use your contribution in the meantime -- just check out your feature branch instead of `master`.
+
+.. _contribution-standards:
+
+Contribution standards
+======================
+
+All contributions (human or AI-assisted) must:
+
+1. Adhere to the :doc:`coding standards </dev/coding_standards>`.
+2. Be reviewed *by the contributor* before submission.
+3. Focus on **single logical changes** — avoid "shotgun" PRs. Try to limit your
+   contribution to one "aspect", e.g. one instrument (or a few if closely
+   related), one bug fix, or one feature.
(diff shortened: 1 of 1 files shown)


docs/dev/contribute.rst
```diff
@@ -105,10 +105,37 @@ There is some automation in place to run the unit tests and check some coding st
...
+- Treat AI as a pair programmer — *you* are responsible for its output.
+- Code must pass the same review bar as human-written contributions.
+- PRs exceedingly large or showing signs of unreviewed generation
+  (e.g. hallucinated APIs, style inconsistencies) may be rejected with
+  reference to these guidelines.
+
+.. hint::
+
+   Remember: Your name goes on the commit. Whether you wrote code with a
+   keyboard, Copilot, or ChatGPT, *you* own its quality.
```
bilderbuchi, line 137: Yes, very good, wholeheartedly agree!

docs/dev/contribute.rst
```diff
@@ -105,10 +105,37 @@ There is some automation in place to run the unit tests and check some coding st
...
+All contributions (human or AI-assisted) must:
+
+1. Adhere to the :doc:`coding standards </dev/coding_standards>`.
+2. Be reviewed *by the contributor* before submission.
+3. Focus on **single logical changes** — avoid "shotgun" PRs. Try to limit your
+   contribution to one "aspect", e.g. one instrument (or a few if closely
+   related), one bug fix, or one feature.
+4. Respect reviewer time through clear documentation and atomic commits.
+
+**For AI-assisted contributions:**
```
bilderbuchi, line 126: Do we want to require people to explicitly acknowledge having used AI for a commit? I'm not sure if that ends up meaningful?

BenediktBurger: I expect a lot of people going to use AI, so the **whether** won't be useful, maybe the **how much**.

As long as the results are fine, I don't care how much AI support was used.

bilderbuchi (approved):

PR #1465, opened 2026-06-02
ENH: Add Koheron CTL200 digital laser driver

niklasschulz99:
### Summary
Added supoort for Koheron CTL200 digital laser controller.

### Tests
`pytest tests/instruments/koheron/test_koheron_ctl200.py`
33 passed

### Additional Information
The Koheron CTL200 laser driver echoes commands and appends ">>" prompts to emulate an interactive terminal. I have implemented a custom serial adapter for this Koheron device to handle this specific behavior.

### DIFF
--- a/pymeasure/instruments/koheron/koheron_ctl200.py
+++ b/pymeasure/instruments/koheron/koheron_ctl200.py
@@ -1,0 +1,397 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2026 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import logging
+import time
+from enum import IntFlag
+from pymeasure.adapters import SerialAdapter
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.validators import (
+    strict_discrete_set,
+    strict_range,
+)
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
+
+
+class KoheronError(IntFlag):
+    """IntFlag for CTL200 Error Codes."""
+
+    NONE = 0
+    UART_BUFFER_OVERFLOW = 1 << 0
+    UART_CMD_BEFORE_PROMPT = 1 << 1
+    LASER_UNDERTEMPERATURE = 1 << 2
+    LASER_OVERTEMPERATURE = 1 << 3
+    CMD_UNKNOWN = 1 << 4
+    CMD_INVALID_ARG = 1 << 5
+    LASER_ON_WHILE_INTERLOCK = 1 << 6
+    INTERLOCK_TRIGGERED = 1 << 7
+
+
+class CTL200Adapter(SerialAdapter):
+    """Serial adapter for Koheron CTL200 do deal with echo responses."""
+
+    def __init__(self, port, **kwargs):
+        kwargs.setdefault("baudrate", 115200)
+        kwargs.setdefault("timeout", 2)
+        super().__init__(
+            port,
+            write_termination="\r\n",
+            read_termination="\r\n",
+            **kwargs
+        )
+        self._last_command: str = ""
+        self._response_buffer: list[str] = []
+        time.sleep(0.05)
+        self.connection.reset_input_buffer()
+
+    def _read_until_prompt(self) -> list[str]:
+        raw = b""
+        total_timeout = max(self.connection.timeout or 0, 2.0)
+        deadline = time.monotonic() + total_timeout
+        while time.monotonic() < deadline:
+            waiting = self.connection.in_waiting
+            if waiting:
+                raw += self.connection.read(waiting)
+                if b">>" in raw:
+                    # read trailing chars (\r\n, ...)
+                    time.sleep(0.02)
+                    trailing = self.connection.in_waiting
+                    if trailing:
+                        raw += self.connection.read(trailing)
+                    break
+            else:
+                time.sleep(0.005)
+
+        log.debug("raw RX: %r", raw)
+        text = raw.decode(errors="replace")
+        lines = [line.strip() for line in text.replace("\r", "").split("\n")]
+        lines = [line for line in lines if line]  # remove blanks
+        if lines and self._last_command and lines[0] == self._last_command:
+            lines.pop(0)  # remove echo
+        lines = [line for line in lines if line != ">>"]  # remove prompt chars
+        log.debug("parsed lines: %r", lines)
+        return lines
+
+    def write(self, command: str) -> None:
+        self._last_command = command.strip()
+        super().write(command)
+        self._response_buffer = self._read_until_prompt()
+        log.debug("write %r -> buffer: %r", command,
+                  self._response_buffer)
+
+    def read(self) -> str:
+        if self._response_buffer:
+            value = self._response_buffer.pop(0)
+            log.debug("read -> %r", value)
+            return value
+        log.debug("read -> buffer empty, return ''")
+        return ""
+
+
+class CTL200(Instrument):
+    """PyMeasure driver for the Koheron CTL200-0 digital
+    laser diode controller."""
+
+    def __init__(self, adapter, name="Koheron CTL200", **kwargs):
+        if isinstance(adapter, str):
+            adapter = CTL200Adapter(adapter, **kwargs)
+        super().__init__(
+            adapter,
+            name,
+            includeSCPI=False,
+            **kwargs
+        )
+
+    # -- Laser -----------------------------------------------------------
+
+    laser_enabled = Instrument.control(
+        "lason",
+        "lason %d",
+        """Control whether the laser output is enabled (bool).""",
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    laser_current = Instrument.control(
+        "ilaser",
+        "ilaser %g",
+        """Control the laser current setpoint in A (float).""",
+        cast=float,
+        get_process=lambda v: v * 1e-3,
+        set_process=lambda v: v * 1e3,
+    )
+
+    laser_current_limit = Instrument.control(
+        "ilmax",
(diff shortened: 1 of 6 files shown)


BenediktBurger:
> The Koheron CTL200 laser driver echoes commands and appends ">>" prompts to emulate an interactive terminal. I have implemented a custom serial adapter for this Koheron device to handle this specific behavior.

Please do not put these communication parts into an adapter, but into the instrument itself. We already have such an instrument which always responds like a shell, but I do not remember which one.

niklasschulz99:
> > The Koheron CTL200 laser driver echoes commands and appends ">>" prompts to emulate an interactive terminal. I have implemented a custom serial adapter for this Koheron device to handle this specific behavior.
> 
> Please do not put these communication parts into an adapter, but into the instrument itself. We already have such an instrument which always responds like a shell, but I do not remember which one.

Thank you for your feedback! 

I have moved the communication functions into the device class and removed the custom KoheronCTL200Adapter. I also figured out how to use the standard way for adapter with visa instead of creating a serialadapter in the class init and defining the connection defaults via kwargs.setdefault().

Additionally, I needed to implement a check for the tests, bc the ProtocolAdapter expects a strict 1-to-1 write-read sequence. The custom terminal cleanup loop with multiple read() calls would crash the tests with an AssertionError, so I implemented a check to safely bypass the echoing logic during pytest runs.

If there is anything wrong with that please let me know and I will adjust the implementation to match the repository's coding standards. Thanks again for your guidance!

pymeasure/instruments/koheron/koheron_ctl200.py
```diff
@@ -0,0 +1,389 @@
...
+    """PyMeasure driver for the Koheron CTL200-0 digital
+    laser diode controller."""
+
+    _STATUS_KEYS = ["lason", "vlaser", "itec", "vtec", "rtact", "iphd", "ain1",
+                    "ain2"]
+
+    def __init__(self, adapter, name="Koheron CTL200", **kwargs):
+        kwargs.setdefault("baud_rate", 115200)
+        kwargs.setdefault("write_termination", "\r\n")
+        kwargs.setdefault("read_termination", "\r\n")
```
BenediktBurger: Are these options changeable on the device?
If not, just put it into the super init call.

pymeasure/instruments/koheron/koheron_ctl200.py
```diff
@@ -0,0 +1,389 @@
...
+            return self.read()
+
+        cmd_clean = sent_command.strip()
+        while True:
+            line = self.read().replace("\x00", "").strip()
+            if line in ("", ">>"):
+                continue
+            if line == cmd_clean:
+                continue
+            if line.startswith(">>") and line.removeprefix(">>").strip() == cmd_clean:
```
BenediktBurger: I guess you could combine both lines (the second one also tests for the first one).

pymeasure/instruments/koheron/koheron_ctl200.py
```diff
@@ -0,0 +1,389 @@
...
+
+        return self._read_cleaned_response(command)
+
+    def ask(self, command):
+        """Query device and read answer without echos and >> chars."""
+        super().write(command)
+        return self._read_cleaned_response(command)
+
+    def _is_test_run(self):
+        return "Protocol" in type(self.adapter).__name__ or (
```
BenediktBurger: Why not `isinstance(self.adapter, ProtocolAdapter)`?

pymeasure/instruments/koheron/koheron_ctl200.py
```diff
@@ -0,0 +1,389 @@
...
+    laser_enabled = Instrument.control(
+        "lason",
+        "lason %d",
+        """Control whether the laser output is enabled (bool).""",
+        validator=strict_discrete_set,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    laser_current = Instrument.control(
```
BenediktBurger: As it is the setpoint, better name it `laser_current_setpoint` in order to distinguish it from the actual laser_current.

pymeasure/instruments/koheron/koheron_ctl200.py
```diff
@@ -0,0 +1,389 @@
...
+        "ilaser",
+        "ilaser %g",
+        """Control the laser current setpoint in A (float).""",
+        cast=float,
```
BenediktBurger: `cast=float` is the default and not needed.

pymeasure/instruments/koheron/koheron_ctl200.py
```diff
@@ -0,0 +1,389 @@
...
+    )
+
+    thermistor_setpoint = Instrument.control(
+        "rtset",
+        "rtset %g",
+        """Control the thermistor resistance setpoint in Ω (float).""",
+        cast=float,
+    )
+
+    tec_current = Instrument.measurement(
```
BenediktBurger, line 168: Does this relate to the laser_current and laser_current_limit? Wouldn't it be better to rename the latter ones tec instead of laser?

pymeasure/instruments/koheron/koheron_ctl200.py
```diff
@@ -0,0 +1,389 @@
...
+        cast=float,
+    )
+
+    analog_input_2 = Instrument.measurement(
+        "ain2",
+        """Get the voltage on auxiliary input AI2 in V (float).""",
+        cast=float,
+    )
+
+    laser_mod_gain = Instrument.control(
```
BenediktBurger: I'd spell it out `laser_modulation_gain`.

pymeasure/instruments/koheron/koheron_ctl200.py
```diff
@@ -0,0 +1,389 @@
...
+
+    def check_errors(self):
+        """Read the error status flag and extract occurring errors."""
+        status = self.error_status
+        if status == KoheronError.NONE:
+            return []
+        errors = []
+        for error in KoheronError:
+            if error != KoheronError.NONE and error in status:
+                errors.append(f"Koheron CTL200 Error: {error.name}")
```
BenediktBurger: I guess a list comprehension would also work:
~~~python
errors = [f"Koheron CTL200 Error: {error.name}" for error in status]
~~~

I don't know whether you really want to add the name of the device, or just the name of the error, as you know from which device you read the errors.

tests/instruments/koheron/test_koheron_ctl200.py
```diff
@@ -0,0 +1,325 @@
...
+        assert inst.laser_enabled is True
+
+
+def test_laser_current():
+    """Verify laser current setting and its mA to Ampere conversion (scaling
+    by 1e3 / 1e-3)."""
+    with expected_protocol(
+            CTL200, [("ilaser 150", None), ("ilaser", "150.0")]) as inst:
+        inst.laser_current = 0.15
+        assert inst.laser_current == pytest.approx(0.15)
```
BenediktBurger: Very good to add a test for the unit conversion!

tests/instruments/koheron/test_koheron_ctl200.py
```diff
@@ -0,0 +1,325 @@
...
+    with expected_protocol(CTL200, [("save", None), ("errclr", None)]) as inst:
+        inst.save()
+        inst.clear_error()
+
+
+def test_error_status():
+    """Verify processing and integer parsing of error status flags."""
+    with expected_protocol(CTL200, [("err", "0")]) as inst:
+        # Casts output to int to evaluate matching underlying flags
+        assert int(inst.error_status) == 0
```
BenediktBurger: why don't you compare it to the IntFlag?

BenediktBurger (changes requested):
Thank you for your work. Here are a few comments.

pymeasure/instruments/koheron/koheron_ctl200.py
```diff
@@ -0,0 +1,389 @@
...
+        kwargs.setdefault("write_termination", "\r\n")
+        kwargs.setdefault("read_termination", "\r\n")
+        super().__init__(adapter, name, **kwargs)
+
+    def _read_cleaned_response(self, sent_command):
+        """Read lines from device and filter for echos and >> chars."""
+        if self._is_test_run():
+            return self.read()
+
+        cmd_clean = sent_command.strip()
```
BenediktBurger, line 75: Actually, you can put the exact response of the device into the tests (with all the echoing etc.), such that the communication is exactly as with the device.

That is preferred over such a hack.

Some other devices have more than one command_pair for a single command exactly for that reason (with some more reading/writing).
You can use `[(None, "first response"), (None, "second response")]` if you read two times consecutively.

pymeasure/instruments/koheron/koheron_ctl200.py
niklasschulz99: The terminations are fixed by the protocol, so I moved them into the `super().__init__()` call. The baud rate is changeable on the device though (`brate`, 9600–460800, persisted with `save`), so hard-coding it would lock out anyone who changed it on their unit. It's an explicit keyword argument with the `115200` default now. This also made me notice `brate` and `serial` were the only commands from the manual I hadn't exposed — added both.

pymeasure/instruments/koheron/koheron_ctl200.py
niklasschulz99: Yes, thanks!

pymeasure/instruments/koheron/koheron_ctl200.py
niklasschulz99: Dropped the check entirely instead - see your comment below; the tests now contain the real device traffic, so the driver no longer needs to know it's under test.

pymeasure/instruments/koheron/koheron_ctl200.py
niklasschulz99: Right, removed all of them.

pymeasure/instruments/koheron/koheron_ctl200.py
niklasschulz99: No - these are two separate circuits. `ilaser`/`ilmax` are the drive current and software limit for the laser diode; `itec` is the current through the TEC (Peltier) that stabilizes its temperature. The TEC has no current limit command; its limits are voltage-based (`vtmin`/`vtmax`, exposed as `tec_voltage_limit_min/_max`), which is why `ilmax` can only refer to the laser.

pymeasure/instruments/koheron/koheron_ctl200.py
niklasschulz99: Done. I renamed `tec_mod_gain` to `tec_modulation_gain` as well, to keep the pair consistent.

pymeasure/instruments/koheron/koheron_ctl200.py
niklasschulz99: Iterating the flag instance already yields only the set flags, so the NONE guard could go. I also dropped the device name from the entries and log them with `self.name` instead, as `SCPIMixin.check_errors` does.

tests/instruments/koheron/test_koheron_ctl200.py
niklasschulz99: Fixed, and parametrized it to cover non-zero flags — the old test only checked `0`, so it would have passed even with completely broken decoding. Also exported KoheronError from the package `__init__`; it's the return type of `error_status`, so users need to be able to import it.

pymeasure/instruments/koheron/koheron_ctl200.py
niklasschulz99: I agree - `_is_test_run` is gone now. 
This flushed out two real bugs the hack was hiding. The echo-filtering logic was never actually executed under test (the test branch bypassed it), and `error_status` was parsing the error code as decimal when the device returns it in hex - `errclr` replies `ffffffff`, and a code like `a0` would have raised on float("a0"). Both fixed, with tests. Separately I corrected the `vtmin`/`vtmax` ranges to ±3.0 V per the manual; they were ±3.3.

pymeasure/instruments/koheron/koheron_ctl200.py
niklasschulz99: Correction to my last comment: iterating the flag instance only works on Python 3.11+, and pymeasure supports 3.10 (which CI runs), so that failed with `TypeError: 'KoheronError' object is not iterable`. The `NONE` guard was there for a reason, I remember now. I kept it a comprehension but mask the members instead, which works on all supported versions:
```
status = self.error_status
errors = [error.name for error in KoheronError if error.value & status]
```
`NONE` has value `0`, so it drops out on its own. All CI checks pass now.

BenediktBurger:
Thank you for your work!

PR #1467, opened 2026-06-05
Add MAUI-specific waveform saving methods

sjgallagher2:
The default methods in `teledyne/teledyne_oscilloscope.py` need to be overridden for MAUI-based oscilloscopes. It will fail to read for several reasons, including missing `sampling_rate` and a couple off-by-one errors in sanity-checking the sizes of payloads. The default scaling methods are also incorrect. [Per the manual](https://cdn.teledynelecroy.com/files/manuals/maui-remote-control-and-automation-manual.pdf), the vertical and horizontal scaling should be performed using data from the waveform descriptor, which must be read each time a channel is captured. 

This commit:
* Copies several auxiliary classes and member functions from the parent class `TeledyneOscilloscope` to provide the small adjustments required. 
* Adds a new `MAUIWaveformDescriptor` dataclass to parse and access the descriptor data. 
* Updates the horizontal and vertical scaling in `_process_data()` to use the data in the waveform descriptor instead of the previous defaults, and this was verified against the scope itself to be accurate, including with `sparsing` (decimating) of the data
* Adds a `block` parameter to several methods, because the default was `DAT2` which is not a sane default. 
* Adds a couple checks mapping `sparsing=0 -> 1`. This avoids any divide-by-zero errors, since e.g. the WaveSurfer uses `sparsing=0` as a default.

The waveform descriptor parsing was placed in the `_digitize()` method. The off-by-one errors appeared in the default `_header_footer_sanity_checks()` method and the `_npoints_sanity_checks()`. 

Testing and validation was performed on a WaveSurfer 4104HD. Example:
```python
from pymeasure.instruments.teledyne import TeledyneMAUI
import matplotlib.pyplot as plt

visa_addr = 'TCPIP0::LCRY4902C18134::INSTR'
scope = TeledyneMAUI(visa_addr)

wave = scope.download_waveform('c1',sparsing=4)

ch1 = wave[0]
ts = wave[1]
plt.plot(ts,ch1,'k')
plt.grid()
```

Test results:
```
======================= 8861 passed, 2194 skipped, 1 xfailed, 929 warnings in 71.49s (0:01:11) ========================
```

### DIFF
--- a/pymeasure/instruments/teledyne/teledyneMAUI.py
+++ b/pymeasure/instruments/teledyne/teledyneMAUI.py
@@ -141,7 +141,7 @@ class MAUIWaveformDescriptor:
     wave_src            : int
 
     @classmethod
-    def parse_desc(cls,desc: bytes):
+    def parse_descriptor(cls,desc: bytes):
         """Parse waveform descriptor from bytes, assuming COMM_HEADER=="OFF".
 
         .. Note::
@@ -396,6 +396,16 @@ class TeledyneMAUI(TeledyneOscilloscope):
     #    Waveform    #
     ##################
 
+    def read_waveform_descriptor(self) -> MAUIWaveformDescriptor:
+        """Read and parse instrument waveform descriptor.
+
+        :return: MAUIWaveformDescriptor
+        """
+        self.write(f"{self.waveform_source}:WF? DESC")
+        dd = self.read_bytes(-1)
+        descb = dd[len('DESC,#9000000000'):]
+        return MAUIWaveformDescriptor.parse_descriptor(descb)
+
     def _digitize(self, src, num_bytes=None, block='DAT1'):
         """Acquire waveforms according to the settings of the acquire commands.
 
@@ -409,11 +419,8 @@ class TeledyneMAUI(TeledyneOscilloscope):
         :return: bytearray with raw data.
         """
         with _ChunkResizer(self.adapter, num_bytes):
-            # Get descriptor  TODO Refactor into its own method
-            self.write(f"{self.waveform_source}:WF? DESC")
-            dd = self.read_bytes(-1)
-            descb = dd[len('DESC,#9000000000'):]
-            self.waveform_descriptor = MAUIWaveformDescriptor.parse_desc(descb)
+            # Get descriptor
+            self.waveform_descriptor = self.read_waveform_descriptor()
             # Get data
             binary_values = self.binary_values(f"{src}:WF? {block}", dtype=np.uint8)
             # NOTE: Payload is stored as raw bytes for now. But it must be converted to bytes or
--- a/pymeasure/instruments/teledyne/teledyneMAUI.py
+++ b/pymeasure/instruments/teledyne/teledyneMAUI.py
@@ -29,8 +29,7 @@ from pymeasure.instruments.teledyne.teledyne_oscilloscope import TeledyneOscillo
 from struct import unpack
 from dataclasses import dataclass
 import numpy as np
-from decimal import Decimal
-import re 
+import re
 
 
 def sanitize_source(source):
@@ -204,16 +203,15 @@ class MAUIWaveformDescriptor:
         acq_vos             = unpack('<f',desc[340:344])[0]
         wave_src            = unpack('<h',desc[344:346])[0]
 
-        return cls(desc_name, desc_templ_name, desc_comm_type, desc_comm_order, wave_desc_len  , user_txt_len   ,
-                    trig_time_arr, ris_time_arr, wave_arr_1, wave_arr_2, instr_name,
-                    instr_num, trace_label, wave_arr_count, pts_per_screen , first_valid,
-                    last_valid, first_pt, sparsing_factor, segment_num, subarr_count,
-                    sweeps_per_acq , pts_per_pair, pair_offset, vgain, vos,
-                    max_value, min_value, nominal_bits, nom_subarr_count, horiz_ival,
-                    horiz_os, pixel_os, vunit, hunit, huncert,
-                    acq_duration, ca_record_type , processing_done, ris_sweeps,
-                    time_base, vcoupling, probe_atten, fixed_vgain, bwlimit,
-                    vertical_vernier, acq_vos, wave_src)
+        return cls(desc_name, desc_templ_name, desc_comm_type, desc_comm_order, wave_desc_len,
+                    user_txt_len, trig_time_arr, ris_time_arr, wave_arr_1, wave_arr_2, instr_name,
+                    instr_num, trace_label, wave_arr_count, pts_per_screen , first_valid, 
+                    last_valid, first_pt, sparsing_factor, segment_num, subarr_count, 
+                    sweeps_per_acq , pts_per_pair, pair_offset, vgain, vos, max_value, 
+                    min_value, nominal_bits, nom_subarr_count, horiz_ival, horiz_os, 
+                    pixel_os, vunit, hunit, huncert, acq_duration, ca_record_type, 
+                    processing_done, ris_sweeps, time_base, vcoupling, probe_atten, 
+                    fixed_vgain, bwlimit, vertical_vernier, acq_vos, wave_src)
 
 
 class TeledyneMAUIChannel(TeledyneOscilloscopeChannel):
@@ -487,12 +485,15 @@ class TeledyneMAUI(TeledyneOscilloscope):
                 bytes_per_point = 2
             else:
                 raise ValueError(f"Expected data format {format}. Options are BYTE or WORD.")
-            requested_bytes = requested_points*bytes_per_point + self._header_size + self._footer_size
+            requested_bytes = requested_points*bytes_per_point + self._header_size \
+                + self._footer_size
             # read the next chunk starting from this points
             first_point = read_points * sparsing
             self.waveform_first_point = first_point
             # read chunk of points
-            values = self._digitize(src=self.waveform_source, block=block, num_bytes=requested_bytes)
+            values = self._digitize(src=self.waveform_source, 
+                                    block=block, 
+                                    num_bytes=requested_bytes)
             # perform many sanity checks on the received data
             self._header_footer_sanity_checks(values,block=block)
             self._npoints_sanity_checks(values)
@@ -501,7 +502,7 @@ class TeledyneMAUI(TeledyneOscilloscope):
             i += 1
         data = np.concatenate(data)
         preamble = self.waveform_preamble
-        
+
         # Determine sampling rate (horizontal interval) from waveform descriptor
         preamble['sampling_rate'] = 1/self.waveform_descriptor.horiz_ival
 
@@ -512,19 +513,24 @@ class TeledyneMAUI(TeledyneOscilloscope):
         """Apply scale and offset to the data points acquired from the scope.
         - Y axis : the scale is given by descriptor.vgain and the offset -descriptor.vos. the
         offset is not applied for the MATH source.
-        - X axis : the scale is given by descriptor.horiz_ival*i + descriptor.horiz_os for i = 0,1,...,N. 
-        :return: tuple of (numpy array of Y points, numpy array of X points, waveform preamble) """
+        - X axis : the scale is given by descriptor.horiz_ival*i + descriptor.horiz_os for
+        i = 0,1,...,N.
+        :return: tuple of (numpy array of Y points, numpy array of X points, waveform preamble)"""
 
         if self.waveform_descriptor.desc_comm_type == 0:
             data_points = ydata.view(np.int8)
         elif self.waveform_descriptor.desc_comm_type == 1:
             data_points = ydata.view(np.int16)
         else:
-            raise ValueError(f"Unexpected communication type {self.waveform_descriptor.desc_comm_type}, expected 0 (byte) or 1 (word).")
+            raise ValueError(
+                f"Unexpected communication type {self.waveform_descriptor.desc_comm_type},"\
+                     +" expected 0 (byte) or 1 (word)."
+                )
         data_points = data_points*self.waveform_descriptor.vgain
         if preamble['source'] != 'MATH':
-            data_points = data_points - self.waveform_descriptor.vos 
-        time_points = np.arange(len(data_points))*self.waveform_descriptor.horiz_ival*preamble['sparsing'] + self.waveform_descriptor.horiz_os
+            data_points = data_points - self.waveform_descriptor.vos
+        time_points = np.arange(len(data_points))*self.waveform_descriptor.horiz_ival \
+            *preamble['sparsing'] + self.waveform_descriptor.horiz_os
         return data_points, time_points, preamble
 
 
@@ -580,8 +586,9 @@ class TeledyneMAUI(TeledyneOscilloscope):
         kwargs.setdefault("destination", "REMOTE")
         self.hardcopy_setup(**kwargs)
         return super().download_image()
-    
-    def download_waveform(self, source, requested_points=None, sparsing=None, block='DAT1', format='WORD'):
+
+    def download_waveform(self, source, requested_points=None, sparsing=None, 
+                          block='DAT1', format='WORD'):
         """Get data points from the specified source of the oscilloscope.
 
         The returned objects are two np.ndarray of data and time points and a dict with the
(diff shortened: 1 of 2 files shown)


pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +390,132 @@ def force_trigger(self):
...
+            # number of points requested in a single chunk
+            requested_points = chunk_points if remaining_points > chunk_points else remaining_points
+            self.waveform_points = requested_points
+            # number of bytes requested in a single chunk
+            requested_bytes = requested_points + self._header_size + self._footer_size
+            # read the next chunk starting from this points
+            first_point = read_points * sparsing
+            self.waveform_first_point = first_point
+            # read chunk of points
+            values = self._digitize(src=self.waveform_source, block=block, num_bytes=requested_bytes)
```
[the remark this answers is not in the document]
sjgallagher2: Not sure where "DAT2 needs two bytes per point" comes from? No mention in the [control and automation manual](https://cdn.teledynelecroy.com/files/manuals/maui-remote-control-and-automation-manual.pdf).

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +390,132 @@ def force_trigger(self):
...
+        with _ChunkResizer(self.adapter, num_bytes):
+            # Get descriptor
+            self.write(f"{self.waveform_source}:WF? DESC")
+            dd = self.read_bytes(-1)
+            descb = dd[len('DESC,#9000000000'):]
+            self.waveform_descriptor = MAUIWaveformDescriptor.parse_desc(descb)
+            # Get data
+            binary_values = self.binary_values(f"{src}:WF? {block}", dtype=np.uint8)
+        if num_bytes is not None and len(binary_values) != num_bytes-1:
+            raise BufferError(f"read bytes ({len(binary_values)}) != requested bytes ({num_bytes})")
```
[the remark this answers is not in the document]
sjgallagher2: Tracked this down to different footer length (1 byte for my MAUI scope, `\n`, not two bytes, `\n\n` as in the parent class

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -26,6 +26,193 @@
...
+
+        return cls(desc_name, desc_templ_name, desc_comm_type, desc_comm_order, wave_desc_len,
+                    user_txt_len, trig_time_arr, ris_time_arr, wave_arr_1, wave_arr_2, instr_name,
+                    instr_num, trace_label, wave_arr_count, pts_per_screen , first_valid, 
+                    last_valid, first_pt, sparsing_factor, segment_num, subarr_count, 
+                    sweeps_per_acq , pts_per_pair, pair_offset, vgain, vos, max_value, 
+                    min_value, nominal_bits, nom_subarr_count, horiz_ival, horiz_os, 
+                    pixel_os, vunit, hunit, huncert, acq_duration, ca_record_type, 
+                    processing_done, ris_sweeps, time_base, vcoupling, probe_atten, 
+                    fixed_vgain, bwlimit, vertical_vernier, acq_vos, wave_src)
```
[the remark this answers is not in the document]
sjgallagher2: These lines do not trip the linter, and do not exceed 100 characters or contain trailing whitespace, as of the latest commit.

pymeasure/instruments/teledyne/teledyne_oscilloscope.py
```diff
@@ -458,7 +458,7 @@ def current_configuration(self):
         return ch_setup
 
 
-class TeledyneOscilloscope(SCPIUnknownMixin, Instrument, metaclass=ABCMeta):
+class TeledyneOscilloscope(SCPIMixin, Instrument, metaclass=ABCMeta):
```
BenediktBurger, line 461: You know that they support SCPI?

Thanks for fixing it!

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -26,6 +26,193 @@
 from pymeasure.instruments.teledyne.teledyne_oscilloscope import TeledyneOscilloscope, \
     TeledyneOscilloscopeChannel, _results_list_to_dict
 
+from struct import unpack
+from dataclasses import dataclass
+import numpy as np
+import re
```
BenediktBurger: Typically, imports are ordered in three groups:

1. standard library (e.g. dataclass, struct, re)
2. Third party libraries (here numpy)
3. Local packages (here pymeasure)

Each group is ordered alphabetically and between each group is an empty space.

So your imports should go up.

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -26,6 +26,193 @@
...
+    vcoupling           : int
+    probe_atten         : float
+    fixed_vgain         : int
+    bwlimit             : int
+    vertical_vernier    : float
+    acq_vos             : float
+    wave_src            : int
+
+    @classmethod
+    def parse_desc(cls,desc: bytes):
```
BenediktBurger: You can keep the longer names, IDE offers autocompletion so no need to save chars:

```suggestion
    def parse_description(cls, description: bytes):
```

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -26,6 +26,193 @@
...
+        """
+        Get description with something like this:
+            self.write(f"{self.waveform_source}:WF? DESC")
+            desc = self.read_bytes(-1)
+        Assumes COMM_HEADER == OFF
+        """
+        # Check for header
+        if desc[0:7].decode('ascii') == 'DESC,#9':
+            desc = desc[len('DESC,#9000000000'):]  # don't need this
```
BenediktBurger: If you do not need it, why do you do it?

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +389,151 @@ def force_trigger(self):
...
+
+        :param src: source of data: "C1", "C2", "C3", "C4", "MATH".
+        :param block: waveform block to be read: "DESC", "TEXT", "TIME", "DAT1", "DAT2"
+        :param: num_bytes: number of bytes expected from the scope (including the header and
+        footer).
+        :return: bytearray with raw data.
+        """
+        with _ChunkResizer(self.adapter, num_bytes):
+            # Get descriptor
+            self.write(f"{self.waveform_source}:WF? DESC")
```
BenediktBurger: Why don't you extract it into its own method to get the descriptor?

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +389,151 @@ def force_trigger(self):
...
+            dd = self.read_bytes(-1)
+            descb = dd[len('DESC,#9000000000'):]
+            self.waveform_descriptor = MAUIWaveformDescriptor.parse_desc(descb)
+            # Get data
+            binary_values = self.binary_values(f"{src}:WF? {block}", dtype=np.uint8)
+            # NOTE: data stored as bytes for now, but if format is WORD it must be converted
```
BenediktBurger: so it is a TODO?

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +389,151 @@ def force_trigger(self):
...
+            # to signed 16-bit integer
+        if num_bytes is not None and len(binary_values) != num_bytes:
+            raise BufferError(f"read bytes ({len(binary_values)}) != requested bytes ({num_bytes})")
+        return binary_values
+
+    def _header_footer_sanity_checks(self, message, block='DAT1'):
+        """Check that the header follows the predefined format.
+        The format of the header is <block>,#9XXXXXXX where XXXXXXX is the number of acquired
```
BenediktBurger, line 436: According to PEP 257, there should be an empty line between the summary and the more expansive description.

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +389,151 @@ def force_trigger(self):
...
+        points, and it is zero padded.
+        Then check that the footer is present. The footer is a double line-carriage \n\n
```
BenediktBurger: You do not do that check in the method.

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +389,151 @@ def force_trigger(self):
...
+        :param message: raw bytes received from the scope """
+        message_header = bytes(message[0:self._header_size]).decode("ascii")
+        # Sanity check on header and footer
+        if message_header[0:7] != f"{block},#9":
+            raise ValueError(f"Waveform data in invalid : header is {message_header}")
```
BenediktBurger:
```suggestion
            raise ValueError(f"Waveform data is invalid : header is {message_header}")
```

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +389,151 @@ def force_trigger(self):
...
+        transmitted_points = int(message_header[-9:])
+        received_points = len(message) - self._header_size - self._footer_size
+        if transmitted_points != received_points:
+            raise ValueError(f"Number of transmitted points ({transmitted_points}) != "
+                             f"number of received points ({received_points})")
+
+    def _acquire_data(self, requested_points=0, sparsing=1, block='DAT1', format='WORD'):
+        """Acquire raw data points from the scope. The header, footer and number of points are
+        sanity-checked, but they are not processed otherwise. For a description of the input
+        arguments refer to the download_waveform method.
```
BenediktBurger: You can link to other methods with:

```suggestion
        arguments refer to the :meth:`download_waveform` method.
```

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +389,151 @@ def force_trigger(self):
...
+        If the number of expected points is big enough, the transmission is split in smaller
+        chunks of 20k points and read one chunk at a time. I do not know the reason why,
+        but if the chunk size is big enough the transmission does not complete successfully.
```
BenediktBurger:
```suggestion
        but if the chunk size is too large, the transmission does not complete successfully.
```

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -196,6 +389,151 @@ def force_trigger(self):
...
+        :return: raw data points as numpy array and waveform preamble
```
BenediktBurger:
```suggestion
        :return: raw data points as tuple of numpy array and waveform preamble.
```

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -248,3 +586,50 @@ def download_image(self, **kwargs):
...
+        preamble["first_point"] = 0
+        # Scale the Y-data and create the X-data
+        return self._process_data(ydata, preamble)
+
+    ###############
+    # Acquisition #
+    ###############
+    def acquisition_sample_size(self,source = None):
+        """Get acquisition sample size. For MAUI oscilloscopes this seems to be the memory size."""
+        return self.memory_size
```
BenediktBurger, line 652: This seems to be a property for me, not an action. So you should make it a `property` and not a method.

BenediktBurger (changes requested):

I have a few comments.

Also, you may add yourself to the authors.txt file if you want to.

pymeasure/instruments/teledyne/teledyneMAUI.py
sjgallagher2: Sorry, ambiguous comment, I meant that we don't need the text "DESC,#9xxxxxxxxx" that prefixes the description. This removes that prefix in a way that makes it clear why there's a hard-programmed length. I'll change the comment

pymeasure/instruments/teledyne/teledyneMAUI.py
sjgallagher2: Not a TODO, I mean that `binary_values` cannot be used as it is now, you would first need to convert it using the data format (BYTE or WORD). That's done in `_process_data()`, but this note is a reminder not to use it as actual scope data yet. For moving the binary data around, it's stored as bytes. When it comes time to interpret those bytes, you check the format and convert. I tried to keep the structure of the original code (from `teledyneoscilloscope`) with its data length checks etc, which assume bytes, and I did the conversion in `_process_data()` right before scaling. I'll reword the comment.

pymeasure/instruments/teledyne/teledyneMAUI.py
sjgallagher2: This overrides the `acquisition_sample_size()` method used in `teledyne_oscilloscope.py`. The property is `memory_size`, it just happens that, for this instrument, the `acquisition_sample_size()` is equal to `memory_size`.

pymeasure/instruments/teledyne/teledyneMAUI.py
sjgallagher2: Agree with formatting changes. These were copied from `teledyne_oscilloscope.py`, should I expand the changes to include that file too? (Just apply the same changes there as in `teledyneMAUI.py`.)

pymeasure/instruments/teledyne/teledyneMAUI.py
sjgallagher2: I will implement this

pymeasure/instruments/teledyne/teledyne_oscilloscope.py
sjgallagher2: I might be confused by GPIB and SCPI terminology, I did that to get rid of the warning and because typical commands like `*IDN?` etc work. At first I thought the MAUI instruments only supported VBS, not GPIB, and thought this was referring to that

pymeasure/instruments/teledyne/teledyneMAUI.py
BenediktBurger: If you want to, yes.

pymeasure/instruments/teledyne/teledyneMAUI.py
BenediktBurger: Ok, then leave it as is.

pymeasure/instruments/teledyne/teledyneMAUI.py
sjgallagher2: Reformatted the relevant parts

pymeasure/instruments/teledyne/teledyne_oscilloscope.py
sjgallagher2: The instrument supports the core GPIB/IEEE 488.2 commands, per SCPI sec. 4.1.1., 
* `*CAL?`
* `*CLS`
* `*ESE`
* `*ESE?`
* `*ESR?`
* `*IDN?`
* `*OPC`
* `*OPT?`
* `*PRE`
* `*RST`
* `*SAV`
* `*SRE`
* `*STB?`
* `*TST`
* `*WAI`

It does NOT support `SYST:ERR?` which is included in `SCPIMixin`, and it doesn't use much `:` nesting, but I'm not sure how strict you are about being considered SCPI-compliant.

BenediktBurger: In this case (especially with not SYST:ERR?) we can remove the SCPIMixin and reimplement the methods, or just add our own `check_errors` method overwriting the one from the Mixin. What do you prefer?

The former one might be more work, but the latter one implies a device which supports a lot of SCPI commands (not only that base selection)

pymeasure/instruments/teledyne/teledyneMAUI.py
```diff
@@ -22,11 +22,199 @@
...
+    vcoupling           : int
+    probe_atten         : float
+    fixed_vgain         : int
+    bwlimit             : int
+    vertical_vernier    : float
+    acq_vos             : float
+    wave_src            : int
+
+    @classmethod
+    def parse_descriptor(cls,desc: bytes):
```
BenediktBurger, line 144:
```suggestion
    def parse_descriptor(cls, desc: bytes):
```

BenediktBurger (approved):
Looks good. Thank you for your work!

Also this SCPI point is still open.

PR #1470, opened 2026-06-09
Update datetime format for `date` and `time` placeholders

### LINKED ISSUE #1468: `{time}` filename placeholder doesn't work on Windows
At the moment using `{time}` placeholder in [filename input](https://pymeasure.readthedocs.io/en/latest/tutorial/graphical.html#the-filename-and-directory-input) leads to the error on Windows.

The issue is that `replace_placeholders` uses `"%H:%M:%S"` time format, but the usage of colon `:` is not allowed for filenames on Windows.

https://github.com/pymeasure/pymeasure/blob/d644577e69bb2ebaaae0788f31f87479f84e1d7a/pymeasure/experiment/results.py#L45

 Thus, the error appears when trying to save file with name smth like `14:34:13_1.csv`:

```python
with open(filename, 'w', encoding=Results.ENCODING) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 22] Invalid argument:
```

I would suggest to change the default to safe format (e.g. `"%H-%M-%S"`) or make it configurable if colons are preferred for other platforms.

ilev-sergey:
Closes #1468 
Updates time format for `time` placeholder to work on Windows. Also updates date format as discussed in the issue.

### DIFF
--- a/pymeasure/experiment/results.py
+++ b/pymeasure/experiment/results.py
@@ -42,7 +42,7 @@ log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 
-def replace_placeholders(string, procedure, date_format="%Y-%m-%d", time_format="%H:%M:%S"):
+def replace_placeholders(string, procedure, date_format="%Y_%m_%d", time_format="%H_%M_%S"):
     """Replace placeholders in string with values from procedure parameters.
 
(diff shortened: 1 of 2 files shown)


BenediktBurger:
Thank you!

PR #1471, opened 2026-06-09
Add Chroma 63600 electronic load bank

sjgallagher2:
This commit is part of my effort to add Chroma (and related) power test equipment to PyMeasure. The [Chroma 63600](https://www.chromausa.com/product/modular-dc-electronic-load-63600/) series of electronic load mainframes support up to 5 electronic loads and 10 channels. The user manual is available from the [Chroma USA website](https://www.chromausa.com/document-library/manuals-63600-series/). 

The basic interface is implemented here, with configurable channels basic formatting of outputs and status bits. Some protocol tests are included, and the documentation page.

### DIFF
--- a/docs/api/instruments/chroma/chroma_63600.rst
+++ b/docs/api/instruments/chroma/chroma_63600.rst
@@ -6,10 +6,10 @@ Chroma 63600 Electronic Load Bank
     :members:
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.chroma.Chroma63610_80_20
+.. autoclass:: pymeasure.instruments.chroma.chroma_63600.Chroma63630_80_60
     :members:
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.chroma.Chroma63630_80_60
+.. autoclass:: pymeasure.instruments.chroma.chroma_63600.Chroma63610_80_20
     :members:
     :show-inheritance:
--- a/docs/api/instruments/chroma/chroma_63600.rst
+++ b/docs/api/instruments/chroma/chroma_63600.rst
@@ -1,7 +1,15 @@
-##################################
+#################################
 Chroma 63600 Electronic Load Bank
-##################################
+#################################
 
 .. autoclass:: pymeasure.instruments.chroma.Chroma63600
     :members:
     :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.chroma.Chroma63610_80_20
+    :members:
+    :show-inheritance
+
+.. autoclass:: pymeasure.instruments.chroma.Chroma63630_80_60
+    :members:
+    :show-inheritance
--- a/docs/api/instruments/chroma/chroma_63600.rst
+++ b/docs/api/instruments/chroma/chroma_63600.rst
@@ -8,8 +8,8 @@ Chroma 63600 Electronic Load Bank
 
 .. autoclass:: pymeasure.instruments.chroma.Chroma63610_80_20
     :members:
-    :show-inheritance
+    :show-inheritance:
 
 .. autoclass:: pymeasure.instruments.chroma.Chroma63630_80_60
     :members:
-    :show-inheritance
+    :show-inheritance:
--- a/docs/api/instruments/chroma/chroma_63600.rst
+++ b/docs/api/instruments/chroma/chroma_63600.rst
@@ -1,0 +1,7 @@
+##################################
+Chroma 63600 Electronic Load Bank
+##################################
+
+.. autoclass:: pymeasure.instruments.chroma.Chroma63600
+    :members:
+    :show-inheritance:
--- a/docs/api/instruments/chroma/index.rst
+++ b/docs/api/instruments/chroma/index.rst
@@ -1,8 +1,8 @@
 .. module:: pymeasure.instruments.chroma
 
-############
+######
 Chroma
-############
+######
 
 This section contains specific documentation on the Chroma instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
 
--- a/docs/api/instruments/chroma/index.rst
+++ b/docs/api/instruments/chroma/index.rst
@@ -4,9 +4,7 @@
 Chroma
 ############
 
-This section contains specific documentation on the Chroma instruments
-that are implemented. If you are interested in an instrument not included,
-please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
+This section contains specific documentation on the Chroma instruments that are implemented. If you are interested in an instrument not included, please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
 
 .. toctree::
    :maxdepth: 2
--- a/docs/api/instruments/chroma/index.rst
+++ b/docs/api/instruments/chroma/index.rst
@@ -1,0 +1,14 @@
+.. module:: pymeasure.instruments.chroma
+
+############
+Chroma
+############
+
+This section contains specific documentation on the Chroma instruments
+that are implemented. If you are interested in an instrument not included,
+please consider :doc:`adding the instrument </dev/adding_instruments/index>`.
+
+.. toctree::
+   :maxdepth: 2
+
+   chroma_63600
--- a/pymeasure/instruments/chroma/__init__.py
+++ b/pymeasure/instruments/chroma/__init__.py
@@ -23,5 +23,3 @@
 #
 
 from .chroma_63600 import Chroma63600
-
-
--- a/pymeasure/instruments/chroma/__init__.py
+++ b/pymeasure/instruments/chroma/__init__.py
@@ -22,6 +22,6 @@
 # THE SOFTWARE.
 #
 
-from .chroma_63600 import Chroma63600, Chroma63630_80_60, Chroma63610_80_20
+from .chroma_63600 import Chroma63600
 
 
--- a/pymeasure/instruments/chroma/__init__.py
+++ b/pymeasure/instruments/chroma/__init__.py
@@ -22,6 +22,6 @@
 # THE SOFTWARE.
 #
 
-from .chroma_63600 import Chroma63600,Chroma63630_80_60,Chroma63610_80_20
+from .chroma_63600 import Chroma63600, Chroma63630_80_60, Chroma63610_80_20
 
 
--- a/pymeasure/instruments/chroma/__init__.py
+++ b/pymeasure/instruments/chroma/__init__.py
@@ -1,0 +1,27 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2026 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from .chroma_63600 import Chroma63600,Chroma63630_80_60,Chroma63610_80_20
+
+
--- a/pymeasure/instruments/chroma/chroma_63600.py
+++ b/pymeasure/instruments/chroma/chroma_63600.py
@@ -89,6 +89,7 @@ class Chroma63600_Channel(Channel):
         values=_BOOLS,
         map_values=True,
     )
+
     enabled = Instrument.control(
         "LOAD?",
         "LOAD %s",
--- a/pymeasure/instruments/chroma/chroma_63600.py
+++ b/pymeasure/instruments/chroma/chroma_63600.py
@@ -406,43 +406,10 @@ class Chroma63600(SCPIMixin, Instrument):
             **kwargs
         )
 
-        # Populate default channel for testing purposes
-        self.add_child(Chroma63630_80_60,1)
+        # Populate a default channel for fallback
+        self.add_child(Chroma63630_80_60, 1)
         # Auto-discover channels
         self.discover_channels()
-
-    def create_channels(self, channel_class_list: list[type[Chroma63600_Channel]]):
-        """Populate the channels of the Chroma 63600-x mainframe.
-
-        Pass a list of Chroma63600_Channel subclasses (not objects). The classes are
-        used to add channels, dependent on the type of load (single or double). The
-        mainframe always leaves space for two channels per load (left and right).
-        The 63610-80-20 has a left and right channel. Other loads (63630-80-60,
-        6363-600-16, 63640-80-80, 63640-150-60) have a single channel per load.
-
-        Note that even if a single-channel load is added, the channel ID increases by
-        _two_. For example, adding five 63630-80-60 loads would result in having channels
-        1, 3, 5, 7, and 9. If the third load were a 63610-80-20 (two channel load), the
-        channels would be 1, 3, 5, 6, 7, and 9.
-        """
-        if self.channels:
-            channels_copy = self.channels.copy()
-            for i,ch in channels_copy.items():
-                self.remove_child(ch)
-
-        i = 1
-        for ch_cls in channel_class_list:
-            if ch_cls == Chroma63610_80_20:  # Double, add two channels
-                self.add_child(ch_cls,i)
-                i += 1
-                self.add_child(ch_cls,i)
-                i += 1
-            elif ch_cls == Chroma63630_80_60:  # Single, add one channel
-                self.add_child(ch_cls,i)
-                i += 2  # Still add 2, one channel skipped
-            else:
-                raise NotImplementedError("Only Chroma 63610-80-20 and 63630-80-60 channels are" \
-                                          +" currently supported.")
 
     def discover_channels(self,Nslots: int = 5):
         """Discover and populate channels programmatically.
--- a/pymeasure/instruments/chroma/chroma_63600.py
+++ b/pymeasure/instruments/chroma/chroma_63600.py
@@ -78,19 +78,19 @@ class Chroma63600_Channel(Channel):
              "UDWH"]
 
     def insert_id(self, command):
-        """Set current channel before performing command"""
+        """Set current channel before performing command."""
         return f"CHAN {self.id};{command}"
 
-    active = Instrument.control(
+    active = Channel.control(
         "CHANNEL:ACTIVE?",
         "CHANNEL:ACTIVE %s",
-        """Set enabled or disabled for the load module, can be ON or OFF""",
+        """Control whether to enable the load module (bool).""",
         validator=strict_discrete_set,
         values=_BOOLS,
         map_values=True,
     )
 
-    enabled = Instrument.control(
+    enabled = Channel.control(
         "LOAD?",
         "LOAD %s",
         """Control current channel active status, can be ``True`` (ON) or ``False`` (OFF).""",
@@ -98,7 +98,7 @@ class Chroma63600_Channel(Channel):
         values=_BOOLS,
         map_values=True,
     )
-    status = Instrument.measurement(
+    status = Channel.measurement(
         "LOAD:PROTECTION?",  # Synonym: FETCH:STATUS?
         """Get the status of the electronic load.
 
@@ -128,31 +128,31 @@ class Chroma63600_Channel(Channel):
         """,
         get_process=lambda v: InstrStatus(v)
     )
-    identify = Instrument.measurement(
+    identify = Channel.measurement(
         "CHAN:ID?",
         """Get module identification string."""
     )
-    current = Instrument.measurement(
+    current = Channel.measurement(
         "FETCH:CURRENT?",
         """Get current measured at electronic load input."""
     )
-    frequency = Instrument.measurement(
+    frequency = Channel.measurement(
         "FETCH:FREQUENCY?",
         """Get the frequency measured in frequency sweep mode or sine wave dynamic mode."""
     )
-    power = Instrument.measurement(
+    power = Channel.measurement(
         "FETCH:POWER?",
         """Get power measured at electronic load input."""
     )
-    voltage = Instrument.measurement(
+    voltage = Channel.measurement(
         "FETCH:VOLTAGE?",
         """Get the voltage measured at the electronic load input."""
     )
-    protection_clear = Instrument.setting(
+    protection_clear = Channel.setting(
         "LOAD:PROTECTION:CLEAR",
         """Set the status of the electronic load to a clear state."""
     )
-    activate_short = Instrument.control(
+    activate_short = Channel.control(
         "LOAD:SHORT?",
         "LOAD:SHORT %s",
         """Set active or unactive for the short-circuit simulation.""",
@@ -160,7 +160,7 @@ class Chroma63600_Channel(Channel):
         values=_BOOLS,
         map_values=True,
     )
-    mode = Instrument.control(
+    mode = Channel.control(
         ":MODE?",
         ":MODE %s",
         """Set the operational mode of the electronic load.
@@ -194,15 +194,15 @@ class Chroma63600_Channel(Channel):
     )
 
     # TIMING MODE
-    amp_hours = Instrument.measurement(
+    amp_hours = Channel.measurement(
         "FETCH:AH?",
         """Get ampere-hour measured in timing mode.""",
     )
-    time = Instrument.measurement(
+    time = Channel.measurement(
         "FETCH:TIME?",
         """Get time measured in timing mode."""
     )
-    watt_hours = Instrument.measurement(
+    watt_hours = Channel.measurement(
         "FETCH:WH?",
         """Get the watt-hour measured in timing mode."""
     )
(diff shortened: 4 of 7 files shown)


docs/api/instruments/chroma/chroma_63600.rst
```diff
@@ -0,0 +1,7 @@
+##################################
+Chroma 63600 Electronic Load Bank
+##################################
```
BenediktBurger: Please make sure that the hashes are as long as the text.

docs/api/instruments/chroma/index.rst
```diff
@@ -0,0 +1,12 @@
+.. module:: pymeasure.instruments.chroma
+
+############
+Chroma
+############
```
BenediktBurger: see chroma

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+    SYNC_TIMEOUT = 2**5
+    REVERSE_VOLTAGE_ON_INPUT = 2**4
+    OVER_POWER = 2**3
+    OVER_CURRENT = 2**2
+    OVER_VOLTAGE = 2**1
+    OVER_TEMPERATURE = 2**0
+    OK = 0
+
+
+class InstrMode(str,Enum):
```
BenediktBurger: There is an enum.StrEnum, however it works only for Python 3.11+.

See mks974b.py file for an example how to make it work.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+
+
+class InstrModeRange(str,Enum):
+    """Instrument mode ranges."""
+    LOW_RANGE="L"
+    MEDIUM_RANGE="M"
+    HIGH_RANGE="H"
+
+
+class Chroma63600_Channel(Channel):
```
BenediktBurger, line 76: Please add the channel also to the api documentation (chroma_63600.rst) such that a user sees also the channel's methods.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+    _BOOLS = {True: "ON", False: "OFF"}
+    MODES = ["CCL","CCM","CCH", "CRL","CRH", "CVL","CRM","CRH", "CPL","CPM","CPH", "CZL","CZM",
+             "CZH", "CCDL","CCDM","CCDH", "CCFSL","CCFSM","CCFSH", "TIML","TIMM","TIMH", "SWDL",
+             "SWDM","SWDH", "OCPL","OCPM","OCPH", "PROG", "MPPTL","MPPTM","MPPTH","UDWL","UDWM",
+             "UDWH"]
+
+    def insert_id(self, command):
+        """Set current channel before performing command"""
+        return f"CHAN {self.id};{command}"
+
```
BenediktBurger, line 122: You know, that the default insert_id would replace `{ch}` with the id, so you could remove that method and write, for example for `active`:  `"CHAN {ch};CHANNEL:ACTIVE?"`.

I understand your idea to not repeat that CHAN part. You can keep it as you have. This is just a note for you.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+    active = Instrument.control(
+        "CHANNEL:ACTIVE?",
+        "CHANNEL:ACTIVE %s",
+        """Set enabled or disabled for the load module, can be ON or OFF""",
```
BenediktBurger: If it is a `control`, the docstring should start with `"Control"`. `"Set"` is for a `setting` (write-only), while "Get"/"Measure" is the beginning for a `measurement` (read-only) variant.

Also, you can formulate it differently:

```suggestion
        """Control whether to enable the load module (bool).""",
```

(following the docstring convention: https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings)

Note that docstrings (according to [PEP 257](https://peps.python.org/pep-0257/)) should end with a period.

map_values makes True/False the values to enter, not ON/OFF.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+    MODES = ["CCL","CCM","CCH", "CRL","CRH", "CVL","CRM","CRH", "CPL","CPM","CPH", "CZL","CZM",
+             "CZH", "CCDL","CCDM","CCDH", "CCFSL","CCFSM","CCFSH", "TIML","TIMM","TIMH", "SWDL",
+             "SWDM","SWDH", "OCPL","OCPM","OCPH", "PROG", "MPPTL","MPPTM","MPPTH","UDWL","UDWM",
+             "UDWH"]
+
+    def insert_id(self, command):
+        """Set current channel before performing command"""
+        return f"CHAN {self.id};{command}"
+
+    active = Instrument.control(
```
BenediktBurger: In a Channel use `Channel.control`, please?

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+        "CHANNEL:ACTIVE?",
+        "CHANNEL:ACTIVE %s",
+        """Set enabled or disabled for the load module, can be ON or OFF""",
+        validator=strict_discrete_set,
+        values=_BOOLS,
+        map_values=True,
+    )
+    enabled = Instrument.control(
```
BenediktBurger:
```suggestion
    )

    enabled = Instrument.control(
```
Please keep empty lines between these (multi-line) definitions.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+        |   4    | Reverse voltage on input    |
+        +--------+-----------------------------+
+        |   3    | Over power                  |
+        +--------+-----------------------------+
+        |   2    | Over current                |
+        +--------+-----------------------------+
+        |   1    | Over voltage                |
+        +--------+-----------------------------+
+        |   0    | Over temperature            |
+        +--------+-----------------------------+
```
BenediktBurger: As you return the enum, the user does not need to know which bit means what.

Instead, you can point towards that enum:


```suggestion
    Returns :class:`InstrStatus`.
```

This will create a documentation link to that class, in this case an enum.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+        """,
+        get_process=lambda v: InstrStatus(v)
+    )
+    identify = Instrument.measurement(
+        "CHAN:ID?",
+        """Get module identification string."""
+    )
+    current = Instrument.measurement(
+        "FETCH:CURRENT?",
+        """Get current measured at electronic load input."""
```
BenediktBurger: In which physical unit?


```suggestion
        """Measure current at electronic load input in Amps (float)."""
```

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+        """Get power measured at electronic load input."""
+    )
+    voltage = Instrument.measurement(
+        "FETCH:VOLTAGE?",
+        """Get the voltage measured at the electronic load input."""
+    )
+    protection_clear = Instrument.setting(
+        "LOAD:PROTECTION:CLEAR",
+        """Set the status of the electronic load to a clear state."""
+    )
```
BenediktBurger: In this case, a method `clear_protection_state` would be better as you do not set any value, but do an "action".

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+    activate_short = Instrument.control(
+        "LOAD:SHORT?",
+        "LOAD:SHORT %s",
+        """Set active or unactive for the short-circuit simulation.""",
+        validator=strict_discrete_set,
+        values=_BOOLS,
+        map_values=True,
+    )
```
BenediktBurger: For boolean properties we recommend a name like `short_circuit_enabled` (with enabled suffix), which makes it clear what true/false mean: true=enabled, false=disabled.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+
+    def __init__(self, adapter, name="Chroma 63600", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            read_termination='\n',
+            **kwargs
+        )
+
+        # Initialize default channels for a 63600-5. Makes testing easier.
```
BenediktBurger: Testing ideas should not enter production code.

If it is a generally useful method, we can keep it, otherwise, I recommend you to have it in your personal code or in the test class (you can have it in the [device test](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/tests.html#device-tests)).

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+        self.init_default_channels()
+
+    def init_default_channels(self):
+        """Populate default channels."""
+        self.add_child(Chroma63630_80_60,1)
+        self.add_child(Chroma63630_80_60,3)
+        self.add_child(Chroma63630_80_60,5)
+        self.add_child(Chroma63630_80_60,7)
+        self.add_child(Chroma63610_80_20,9)
+        self.add_child(Chroma63610_80_20,10)
```
BenediktBurger: Normally you would use the ChannelCreator for adding channels, not add_child directly.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+    def __init__(self, adapter, name="Chroma 63600", **kwargs):
+        super().__init__(
+            adapter,
+            name,
+            read_termination='\n',
+            **kwargs
+        )
+
+        # Initialize default channels for a 63600-5. Makes testing easier.
+        self.init_default_channels()
```
BenediktBurger: This will break if you do have a different card setup, right? So it will break for other users.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+    def init_default_channels(self):
+        """Populate default channels."""
+        self.add_child(Chroma63630_80_60,1)
+        self.add_child(Chroma63630_80_60,3)
+        self.add_child(Chroma63630_80_60,5)
+        self.add_child(Chroma63630_80_60,7)
+        self.add_child(Chroma63610_80_20,9)
+        self.add_child(Chroma63610_80_20,10)
+
+    def create_channels(self, channel_class_list: list[type[Chroma63600_Channel]]):
```
BenediktBurger: Another idea for that: Couldn't you iterate over the possible number and try to read the channel type and add it then dynamically?

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,493 @@
...
+                self.add_child(ch_cls,i)
+                i += 2  # Still add 2, one channel skipped
+            else:
+                raise NotImplementedError("Only Chroma 63610-80-20 and 63630-80-60 channels are" \
+                                          +" currently supported.")
+
+    run = Instrument.setting(
+        ":RUN",
+        "Set all electronic loads to ON."
+    )
```
BenediktBurger: Should be a method, not a property.

tests/instruments/chroma/test_chroma_63600.py
BenediktBurger, line 1: Thanks for adding tests.

pymeasure/instruments/chroma/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from .chroma_63600 import Chroma63600, Chroma63630_80_60, Chroma63610_80_20
```
BenediktBurger: In the instrument's init you only need to mention the Instrument, not the Channels which you normally do not need individually.

However, you want to use the channels in the create_channels method, so you would need it, right?

Maybe also change the channel names to somehow distinguish them from the instrument (e.g. adding 80_60_load) or something?

BenediktBurger (changes requested):
Thanks for your contribution.

You may add yourself to the `authors.txt` file as well.

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: I originally used StrEnum but replaced it with multiple inheritance since CI was failing with it. Functionality should be the same

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: That's a good idea, go through channels and use the `identify` property, I'll implement this

pymeasure/instruments/chroma/__init__.py
sjgallagher2: For assigning or overriding channels, you would need them. Might not be necessary when using auto-discovery of channels now though. FWIW they are distinguished by part numbers, `63630` and `63610` (loads) instead of `63600` (mainframe). Adding `load` or similar seems clunky but if it's done like that elsewhere that's fine, it's a painless change

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: Correct, but I wanted to have some kind of default because otherwise testing gets tricky. With channel discovery this shouldn't be an issue for other users, it will get overwritten. I'll revisit once channel discovery is implemented

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: I'm stuck on testing again. How do I use `expected_protocol()` on a class that runs SCPI commands in its constructor? In this case, constructing the instrument object will automatically call `discover_channels()` which will query the instrument for its channels and will try to populate them. Running tests fails because these first dozen or so queries are always the channel querying commands, and they picked up first instead of the actual test code after initialization.

sjgallagher2: Here's a code example for the instrument discovery:

```python
class Chroma63600(SCPIMixin, Instrument):
    """Control the Chroma 63600-5 Electronic Load Bank.

    Programming guide is UM-63600 (this interface is based on v2.6, Jan 2021), available from the
    Chroma USA website.
    https://www.chromausa.com/document-library/manuals-63600-series/
    """

    def __init__(self, adapter, name="Chroma 63600", **kwargs):
        super().__init__(
            adapter,
            name,
            read_termination='\n',
            **kwargs
        )

        # Populate default channels for testing purposes
        self.add_child(Chroma63630_80_60,1)
        self.add_child(Chroma63630_80_60,3)
        self.add_child(Chroma63630_80_60,5)
        self.add_child(Chroma63630_80_60,7)
        self.add_child(Chroma63610_80_20,9)
        self.add_child(Chroma63610_80_20,10)
        # Auto-discover channels
        self.discover_channels()

    def discover_channels(self,Nslots : int = 5):
        """Discover and populate channels programmatically.

        The mainframe always leaves space for two channels per load (left and right).
        The 63610-80-20 has a left and right channel. Other loads (63630-80-60,
        6363-600-16, 63640-80-80, 63640-150-60) have a single channel per load.

        Note that even if a single-channel load is added, the channel ID increases by
        _two_. For example, adding five 63630-80-60 loads would result in having channels
        1, 3, 5, 7, and 9. If the third load were a 63610-80-20 (two channel load), the
        channels would be 1, 3, 5, 6, 7, and 9.

        :param Nslots: Number of mainframe slots in the unit, either 1, 2, or 5 for the -1, -2,
                       and -5 mainframes. For N slots, there are up to 2N channels.
        """
        # Reset channels
        if self.channels:
            channels_copy = self.channels.copy()
            for i,ch in channels_copy.items():
                self.remove_child(ch)

        # Auto-discover and add channels
        for i in range(1,Nslots*2+1):
            if self.ask(f'CHAN {i};CHAN?') == str(i):
                # This channel exists, get its id and add it
                chid = self.ask('CHAN:ID?').split(',')[1]
                if chid == '63630-80-60':
                    self.add_child(Chroma63630_80_60,i)
                elif chid in ['63610-80-20L','63610-80-20R']:
                    self.add_child(Chroma63610_80_20,i)
                else:
                    raise NotImplementedError("Only Chroma 63610-80-20 and 63630-80-60 channels " \
                                             +"are currently supported.")
```

Every test fails with `AssertionError: Written bytes 'b'CHAN 1;CHAN?'' do not match expected 'b'CHAN 1;CHANNEL:ACTIVE ON''` (or whatever the particular test is).

sjgallagher2: Ok I used `test_AWG401x.py` as a template and just added the init commands to the expected response.

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: Can `ChannelCreator` be used to add channels programmatically? I was following [this note in the docs](https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/channels.html#adding-removing-channels),


> In order to add or remove programmatically channels, use the parent’s [add_child()](https://pymeasure.readthedocs.io/en/latest/api/instruments/instruments.html#pymeasure.instruments.common_base.CommonBase.add_child), [remove_child()](https://pymeasure.readthedocs.io/en/latest/api/instruments/instruments.html#pymeasure.instruments.common_base.CommonBase.remove_child) methods.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,522 @@
...
+            adapter,
+            name,
+            read_termination='\n',
+            **kwargs
+        )
+
+        # Populate default channel for testing purposes
+        self.add_child(Chroma63630_80_60,1)
+        # Auto-discover channels
+        self.discover_channels()
```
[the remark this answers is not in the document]
sjgallagher2: Not going to be implemented, testing gets around this by including preamble

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: Removed method

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: Implemented

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: Implemented

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: Accepted

tests/instruments/chroma/test_chroma_63600.py
```diff
@@ -0,0 +1,211 @@
...
+             (b'CHAN 10;CHAN?', b'10'),
+             (b'CHAN:ID?', b'CHROMA,63610-80-20R,636108008723,1.92,1.92')]
+
+
+def test_init():
+    with expected_protocol(
+            Chroma63600,
+            init_cmds,
+    ):
+        pass  # Verify the expected communication.
```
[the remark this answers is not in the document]
sjgallagher2: Not going to be implemented

pymeasure/instruments/chroma/chroma_63600.py
BenediktBurger: The ChannelCreator is used for static definitions (like you did here).
However, as you pointed out correctly, for dynamic additions, you have to resort to add_child etc.

pymeasure/instruments/chroma/__init__.py
BenediktBurger: For me (as someone who does not have the instrument) it is difficult to see that these are channels and not instruments.
You use Manufactuerer and Part? You can keep that naming scheme.

If you do not need the channels for manual override, please remove them from that import (someone can always import them from the chroma_63600.py file anyways).

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: Can't figure out a way to use `ChannelCreator` while also overwriting/replacing the default channel. Class-level channels can't be removed, and using `ChannelCreator` in `__init__()` doesn't create `self.channels` (and generally seems to be against the intention of `ChannelCreator`. 

Not trying to be difficult, but is there a good reason for avoiding `add_child()` and `remove_child()`? Is it for stability or functionality, or is it mostly aesthetics? Should those methods be private if they're not supposed to be used or if that API is liable to change? If there's an easy way to use `ChannelCreator` to make a  default channel with a random class, such that it can be removed and changed into a different class later, then that's fine, I'm just not sure what problem there is using `add_child()` in `__init__()` instead in this case.

pymeasure/instruments/chroma/__init__.py
sjgallagher2: I've removed the method for adding channels manually and removed load modules from imports. In theory, `discover_channels()` should always work. If it ever becomes an issue we can revisit.

pymeasure/instruments/chroma/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+from .chroma_63600 import Chroma63600
+
```
BenediktBurger: Only one empty line is needed.

```suggestion
```

pymeasure/instruments/chroma/chroma_63600.py
BenediktBurger: I extracted that logic into a _strenum module, which contains that logic, please use StrEnum from there.

https://github.com/pymeasure/pymeasure/blob/master/pymeasure/instruments/_strenum.py

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+        """Get module identification string.""",
+        cast=str,
+    )
+    current = Channel.measurement(
+        "FETCH:CURRENT?",
+        """Measure current at electronic load input in Amps (float)."""
+    )
+    frequency = Channel.measurement(
+        "FETCH:FREQUENCY?",
+        """Get the frequency measured in frequency sweep mode or sine wave dynamic mode."""
```
BenediktBurger: Why not?

```suggestion
        """Measure the frequency in frequency sweep mode or sine wave dynamic mode."""
```

also methods below.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+    enabled = Channel.control(
+        "LOAD?",
+        "LOAD %s",
+        """Control current channel active status, can be ``True`` (ON) or ``False`` (OFF).""",
+        validator=strict_discrete_set,
+        values=_BOOLS,
+        map_values=True,
+        cast=str,
+    )
+    status = Channel.measurement(
```
BenediktBurger, line 143: Please keep an empty line between (these multi-line) property definitions.

```suggestion
    )

    status = Channel.measurement(
```

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+        """Get the voltage measured at the electronic load input."""
+    )
+    def clear_protection_state(self):
+        """Set the status of the electronic load to a clear state."""
+        self.write("LOAD:PROTECTION:CLEAR")
+
+    short_circuit_enabled = Channel.control(
+        "LOAD:SHORT?",
+        "LOAD:SHORT %s",
+        """Set active or unactive for the short-circuit simulation.""",
```
BenediktBurger: It is a control, so you can get and set:

```suggestion
        """Control whether the short-circuit simulation is enabled.""",
```

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+        validator=strict_discrete_set,
+        values=_BOOLS,
+        map_values=True,
+        cast=str,
+    )
+    mode = Channel.control(
+        ":MODE?",
+        ":MODE %s",
+        """Set the operational mode of the electronic load.
```
BenediktBurger: Again: Control, not set.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+
+class Chroma63630_80_60(Chroma63600_Channel):
+    """Represents a Chroma 63630-80-60 single load module."""
+
+    # CONSTANT CURRENT MODE (STATIC)
+    current_setpoint_1 = Channel.control(
+        "CURRENT:STATIC:L1?",
+        "CURRENT:STATIC:L1 %g",
+        """Control the L1 static current set point in constant current mode.""",
+        validator=truncated_range,
```
BenediktBurger: We prefer strict_range over truncated_range, as a user won't get feedback that his values are truncated.
E.g. he will think that `current_setpoint_1 = 100` will set it to 100, but it does not.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+        values=[0.015,3000],
+    )
+    # CONSTANT IMPEDANCE MODE
+    parallel_load_capacitance_setpoint = Channel.control(
+        "IMPEDANCE:STATIC:CL?",
+        "IMPEDANCE:STATIC:CL %g",
+        """Set the equivalent parallel load capacitance in constant impedance mode.""",
+        # validator=truncated_range,
+        # values=[0,0],
```
BenediktBurger: Please remove these empty comments.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+        "RESISTANCE:STATIC:L2 %g",
+        """Control the L2 resistance set point in constant resistance mode.""",
+        validator=truncated_range,
+        values=[0.015,3000],
+    )
+    # CONSTANT IMPEDANCE MODE
+    parallel_load_capacitance_setpoint = Channel.control(
+        "IMPEDANCE:STATIC:CL?",
+        "IMPEDANCE:STATIC:CL %g",
+        """Set the equivalent parallel load capacitance in constant impedance mode.""",
```
BenediktBurger: Again a `control` should start with "Set".

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+        "FETCH:WH?",
+        """Get the watt-hour measured in timing mode."""
+    )
+
+
+class Chroma63630_80_60(Chroma63600_Channel):
+    """Represents a Chroma 63630-80-60 single load module."""
+
+    # CONSTANT CURRENT MODE (STATIC)
+    current_setpoint_1 = Channel.control(
```
BenediktBurger, line 237: You could think about making these a Channel of a Channel (with those 1/2 thingy).

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+class Chroma63610_80_20(Chroma63600_Channel):
+    """Represents one channel of a Chroma 63610-80-20 double load module."""
+
+    # CONSTANT CURRENT MODE (STATIC)
+    current_setpoint_1 = Channel.control(
+        "CURRENT:STATIC:L1?",
+        "CURRENT:STATIC:L1 %g",
+        """Control the L1 static current set point in constant current mode.""",
+        validator=truncated_range,
+        values=[0,20],
```
BenediktBurger: Do I understand it correctly, that this channel just differs in the values from the other channel?

You could use [dynamic properties](https://github.com/pymeasure/pymeasure/blob/master/pymeasure/instruments/_strenum.py) for changing the values without rewriting the rest of the creator.
That makes it easier to have the same code but different ranges.

pymeasure/instruments/chroma/chroma_63600.py
```diff
@@ -0,0 +1,476 @@
...
+            read_termination='\n',
+            **kwargs
+        )
+
+        # Populate a default channel for fallback
+        self.add_child(Chroma63630_80_60, 1)
+        # Auto-discover channels
+        self.discover_channels()
+
+    def discover_channels(self,Nslots: int = 5):
```
BenediktBurger:
```suggestion
    def discover_channels(self, Nslots: int = 5):
```

BenediktBurger (changes requested):
A few more ideas: Some changes apply on more places than I marked:
- empty space between property creators
- `control` should start with "Control" string.

Also I noticed some similarities between channels: You could parametrize them instead of repeating yourself: https://pymeasure.readthedocs.io/en/latest/dev/adding_instruments/properties.html#instrument-family-with-different-parameter-values

Maybe you could create channels of channels to reduce the "voltage_1", "voltage_2" properties in the channels.

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: It seems so, though it's been a bit since I wrote those. I can look into refactoring at a later time

pymeasure/instruments/chroma/chroma_63600.py
sjgallagher2: This seems like it would fit well with a later refactor, merged with the dynamic properties

PR #1477, opened 2026-06-15
Update ESP_300 to use standard list instead of individual axes

Ashwolaa:
Axes in ESP300 are getting built dynamically from a property by comparing all the objects present in the controller class. and checking if they belong to the Axis class.

In my use, it leads to some problems as it will check with all the objects in Intruments and can go through some SCPI command such as `complete`. which leads to a crash/stall of the loop.

I suggest to remove the axes property and instead just store it during init as a bare list. This avoids the unrequired complexity and stays as general as possible without enforcing some specific names to the Axis.

### DIFF
--- a/docs/about/changes.rst
+++ b/docs/about/changes.rst
@@ -5,3 +5,5 @@ Changelog
 =========
 
 .. include:: ../../CHANGES.rst
+
+ESP300: x, y, phi are no longer attributes and have been replaced by axes[0], axes[1], axes[2]. 
--- a/docs/about/changes.rst
+++ b/docs/about/changes.rst
@@ -5,5 +5,3 @@ Changelog
 =========
 
 .. include:: ../../CHANGES.rst
-
-ESP300: x, y, phi are no longer attributes and have been replaced by axes[0], axes[1], axes[2]. 
--- a/pymeasure/instruments/newport/esp300.py
+++ b/pymeasure/instruments/newport/esp300.py
@@ -257,8 +257,10 @@ class ESP300(SCPIMixin, Instrument):
     @property
     def x(self):
         """ Get the first axis of the controller.
+
         .. deprecated:: 0.17.0
             Use :attr:`axes[0]` instead.
+
         """
         warn("The x attribute is deprecated. "
              "Axes are now directly accessed through the axes attribute.",
@@ -269,8 +271,10 @@ class ESP300(SCPIMixin, Instrument):
     @property
     def y(self):
         """ Get the second axis of the controller.
+
         .. deprecated:: 0.17.0
             Use :attr:`axes[1]` instead.
+
         """
         warn("The x attribute is deprecated. "
              "Axes are now directly accessed through the axes attribute.",
@@ -281,8 +285,10 @@ class ESP300(SCPIMixin, Instrument):
     @property
     def phi(self):
         """ Get the third axis of the controller.
+
         .. deprecated:: 0.17.0
             Use :attr:`axes[2]` instead.
+
         """
         warn("The x attribute is deprecated. "
              "Axes are now directly accessed through the axes attribute.",
--- a/pymeasure/instruments/newport/esp300.py
+++ b/pymeasure/instruments/newport/esp300.py
@@ -254,19 +254,23 @@ class ESP300(SCPIUnknownMixin, Instrument):
         )
         self.axes = [Axis(1, self), Axis(2, self), Axis(3, self)]
 
+    @property
     def x(self):
         warn("The x attribute is deprecated."
-             "Axes are now directly accessed through the axes attribute.", FutureWarning)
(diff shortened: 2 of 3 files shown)


pymeasure/instruments/newport/esp300.py
BenediktBurger: Does it support SCPI or not?

Could you clarify it and either remove the UnknownMixin or replace it with the SCPIMixin?

Thank you!

BenediktBurger (approved):
Looks good.

Do I see it correctly, that the resulting API to the end user does not change? i.e. x,y, phi, axes are accessible as before?

pymeasure/instruments/newport/esp300.py
Ashwolaa: It follows SCPI protocol (but does not support most of the SCPI default properties listed in Instruments).
Should I replace it by SCPIMixin then?

Ashwolaa:
> Looks good.
> 
> Do I see it correctly, that the resulting API to the end user does not change? i.e. x,y, phi, axes are accessible as before?

Yes they are all accessible. axes is no longer a property but an attribute now

pymeasure/instruments/newport/esp300.py
BenediktBurger: Yes, if it does not support the default ones, remove the SCPIUnknownMixin.

Just copy those methods which do work into this driver. (In this PR or another one, as you wish)

Thank you!

BenediktBurger:
You have to fix the failing tests. If you need help, feel free to ask.

pymeasure/instruments/newport/esp300.py
```diff
@@ -253,10 +252,37 @@ def __init__(self, adapter, name="Newport ESP 300 Motion Controller", **kwargs):
...
-        # Defines default axes, which can be overwritten
-        self.x = Axis(1, self)
-        self.y = Axis(2, self)
-        self.phi = Axis(3, self)
+        self.axes = [Axis(1, self), Axis(2, self), Axis(3, self)]
+
+    @property
+    def x(self):
+        """ Get the first axis of the controller. Deprecated, Use axes[0] instead.
+        """
```
BenediktBurger:
```suggestion
        """ Get the first axis of the controller.
        
        .. deprecated:: 0.17.0
            Use :attr:`axes[0]` instead.
        """
```
This makes it in the API overview a nice deprecation warning.

pymeasure/instruments/newport/esp300.py
```diff
@@ -253,10 +252,37 @@ def __init__(self, adapter, name="Newport ESP 300 Motion Controller", **kwargs):
...
+        warn("The x attribute is deprecated. "
+             "Axes are now directly accessed through the axes attribute.",
+             FutureWarning,
+             stacklevel=2)
+        return self.axes[0]
+
+    @property
+    def y(self):
+        """ Get the second axis of the controller. Deprecated, Use axes[1] instead.
+        """
```
BenediktBurger:
```suggestion
        """ Get the second axis of the controller.
        
        .. deprecated:: 0.17.0
            Use :attr:`axes[1]` instead.
        """
```

pymeasure/instruments/newport/esp300.py
```diff
@@ -253,10 +252,37 @@ def __init__(self, adapter, name="Newport ESP 300 Motion Controller", **kwargs):
...
+        warn("The x attribute is deprecated. "
+             "Axes are now directly accessed through the axes attribute.",
+             FutureWarning,
+             stacklevel=2)
+        return self.axes[1]
+
+    @property
+    def phi(self):
+        """ Get the third axis of the controller. Deprecated, Use axes[2] instead.
+        """
```
BenediktBurger:
```suggestion
        """ Get the third axis of the controller.

        .. deprecated:: 0.17.0
            Use :attr:`axes[0]` instead.
        """
```

BenediktBurger (approved):
A small (commitable) suggestion for better deprecation notes.
If you want to, you can put a small sentence to the changes.rst as well.

Sorry for the last minute request.

Ashwolaa:
Hey thanks for the feedback, I included your recommendation. 
I am not sure for the .rst, you can modify/delete it as you prefer.

docs/about/changes.rst
```diff
@@ -5,3 +5,5 @@ Changelog
 =========
 
 .. include:: ../../CHANGES.rst
+
+ESP300: x, y, phi are no longer attributes and have been replaced by axes[0], axes[1], axes[2].
```
BenediktBurger: In the CHANGES.rst in the root folder. This file here loads the root folder one.

BenediktBurger:
```suggestion
```

BenediktBurger:
Thank you for your work!

Ashwolaa:
> Thank you for your work!

Thanks for your help! I'll try do add some new instruments in the coming months

PR #1482, opened 2026-06-21
Remove Keithley2000 SCPIUnknownMixin

chieh0517:


BenediktBurger:
Thank you for clarifying these capabilities and changing it!

PR #1486, opened 2026-06-22
Switch to pyright and improve Property typing

BenediktBurger:


ilev-sergey:
Just curious: why did you decide to move from ty to pyright? What was your experience with ty?

PR #1492, opened 2026-06-23
Update Keithley2450 with additional properties and methods

ilev-sergey:


ilev-sergey:
@coderabbitai review

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -341,6 +342,89 @@ def __init__(self, adapter, name="Keithley 2450 SourceMeter", **kwargs):
...
+    ############################
+
+    current_autozero = Instrument.control(
+        ":SENS:CURR:AZER?",
+        ":SENS:CURR:AZER %s",
+        """ Control (string) the auto-zero state for current measurements.
+        Valid values are 'ON' and 'OFF'. """,
+        validator=strict_discrete_set,
+        values=["ON", "OFF"],
+    )
```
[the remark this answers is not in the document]
BenediktBurger: I guess on/off are the right strings to send to the device.
In this case, cast=Str might be the right option.
Or, if the device requires an int (0/1), you miss map_values=True.

ilev-sergey:
@coderabbitai review

pymeasure/instruments/keithley/buffer.py
```diff
@@ -35,9 +35,8 @@
 log.addHandler(logging.NullHandler())
 
 
-class KeithleyBuffer:
-    """ Implements the basic buffering capability found in
-    many Keithley instruments. """
+class KeithleyBufferBase:
+    """Implement the subset of buffer capability that is available in :class:`Keithley2450`."""
```
BenediktBurger, line 39: For a "Base class", it seems odd to mention a single instrument.

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -24,28 +24,52 @@
...
 
 # Setup logging
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 
-class Keithley2450(KeithleyBuffer, SCPIMixin, Instrument):
-    """ Represents the Keithley 2450 SourceMeter and provides a
-    high-level interface for interacting with the instrument.
+def _validated_sweep_delay(delay: float, allow_auto: bool) -> float:
```
BenediktBurger, line 47: A function/method should have an "action" name, so better `validate_sweep_delay`.

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -66,22 +90,23 @@ class Keithley2450(KeithleyBuffer, SCPIMixin, Instrument):
 
     """
 
-    def __init__(self, adapter, name="Keithley 2450 SourceMeter", **kwargs):
+    def __init__(self, adapter, name: str = "Keithley 2450 SourceMeter", **kwargs):
```
BenediktBurger, line 93: As you add typing here, you might use `AdapterType` from the instrument module:

```suggestion
    def __init__(self, adapter: AdapterType, name: str = "Keithley 2450 SourceMeter", **kwargs):
```

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -313,7 +338,8 @@ def __init__(self, adapter, name="Keithley 2450 SourceMeter", **kwargs):
         MOV : Moving filter""",
         validator=strict_discrete_set,
         values=['REP', 'MOV'],
-        map_values=False)
+        map_values=False,
+        cast=str)
```
BenediktBurger, line 342: While you already are at it:

```suggestion
        cast=str,
    )
```

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -337,7 +363,8 @@ def __init__(self, adapter, name="Keithley 2450 SourceMeter", **kwargs):
...
```
BenediktBurger, line 367:
```suggestion
        cast=str,
    )
```

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -347,6 +374,78 @@ def __init__(self, adapter, name="Keithley 2450 SourceMeter", **kwargs):
...
 
+    ############################
+    # Auto-zero and Sense Count #
+    ############################
+
+    current_autozero: InstrumentProperty[bool] = Instrument.control(
+        ":SENS:CURR:AZER?",
+        ":SENS:CURR:AZER %d",
+        """ Control (bool) whether the internal reference measurements (auto-zero) are
+        updated automatically for current measurements. Valid values are True and False. """,
```
BenediktBurger, line 385: Typically, we write bool at the end. This already defines which values are valid:

```suggestion
        """ Control whether the internal reference measurements (auto-zero) are
        updated automatically for current measurements (bool). """,
```

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -347,6 +374,78 @@ def __init__(self, adapter, name="Keithley 2450 SourceMeter", **kwargs):
...
+        automatically. When set to True the instrument picks the optimal range. """,
+        values={True: 1, False: 0},
+        map_values=True,
+    )
+
+    sense_count = Instrument.control(
+        ":SENS:COUNT?",
+        ":SENS:COUNT %d",
+        """ Control (integer) the number of measurements made per trigger,
+        from 1 to 300 000. """,
```
BenediktBurger, line 403: We state the type and limits (and how to enforce at the end), see https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings :

```suggestion
        """ Control the number of measurements made per trigger,
        (int truncated from 1 to 300 000). """,
```

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -347,6 +374,78 @@ def __init__(self, adapter, name="Keithley 2450 SourceMeter", **kwargs):
...
+        validator=truncated_range,
```
BenediktBurger, line 404: Truncated validators ( @pymeasure/instruments/validators.py ) have the disadvantage, that a user won't get feedback: e.g setting output_voltage = 20, with a limit of 10, will make the user think, the output voltage were 20, but it is only 10.

Therefore, we prefer strict validators.

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -555,82 +661,191 @@ def ramp_to_voltage(self, target_voltage, steps=30, pause=20e-3):
...
 
-    def use_rear_terminals(self):
+    def autozero_once(self) -> None:
+        """Perform a single auto-zero correction on the sense subsystem."""
+        self.write(":SENS:AZER:ONCE")
+
+    def sweep_voltage_linear(self, v_from: float, v_to: float, n_steps: int,
+                             delay: float = 1e-4, count: int = 1, range_type: str = "BEST",
+                             fail_abort: bool = True, dual: bool = False,
+                             buffer_name: str = "defbuffer1") -> None:
```
BenediktBurger, line 736: Here I would prefer different formatting (use ruff format or your favourite formatter):

```suggestion
    def sweep_voltage_linear(
        self, v_from: float,
        v_to: float,
        n_steps: int,
        delay: float = 1e-4,
        count: int = 1,
        range_type: str = "BEST",
        fail_abort: bool = True,
        dual: bool = False,
        buffer_name: str = "defbuffer1"
    ) -> None:
```

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -555,82 +661,191 @@ def ramp_to_voltage(self, target_voltage, steps=30, pause=20e-3):
...
+                           whole sweep, or 'FIXED' for the present range
+        :param fail_abort: Abort the sweep if the source limit is exceeded
+        :param dual: Sweep from start to stop and then back from stop to start
+        :param buffer_name: Name of the buffer to store the readings in
+        :raises ValueError: If any argument lies outside the accepted values
+        """
+        v_from = strict_range(v_from, [-210, 210])
+        v_to = strict_range(v_to, [-210, 210])
+        n_steps = strict_range(n_steps, [2, 1000000])
+        count = strict_range(count, [0, 268435455])
```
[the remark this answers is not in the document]
BenediktBurger: That's right.
You could use `in_set(range(start, stop))` to check for individual members.

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -555,82 +661,191 @@ def ramp_to_voltage(self, target_voltage, steps=30, pause=20e-3):
...
+                        for an infinite loop
+        :param delay: Delay in seconds between each voltage step and measurement, either
+                      0 for no delay or a value from 50e-6 to 10000
+        :raises ValueError: If any argument lies outside the accepted values
+        """
+        n_points = len(waveform)
+        if not 1 <= n_points <= 2500:
+            raise ValueError(
+                f"A voltage list sweep takes 1 to 2500 points, but {n_points} were given."
+            )
```
BenediktBurger, line 789: Why don't you use a validator here?

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -555,82 +661,191 @@ def ramp_to_voltage(self, target_voltage, steps=30, pause=20e-3):
...
+        for voltage in waveform:
+            strict_range(voltage, [-210, 210])
```
BenediktBurger, line 791: We already use numpy in this instrument, so you could use numpy array comparison.

pymeasure/instruments/keithley/keithley2450.py
```diff
@@ -555,82 +661,191 @@ def ramp_to_voltage(self, target_voltage, steps=30, pause=20e-3):
...
+        def fmt(values: Iterable[float]) -> str:
+            """Format a sequence of voltages as a comma-separated command argument.
+
+            :param values: A sequence of voltage values in Volts
+            :returns: The values as a comma-separated string
+            """
+            return ", ".join(f"{v:.3g}" for v in values)
+
+        self.write(":SOUR:FUNC VOLT")
+        if n_points > 100:
```
BenediktBurger, line 804: why exactly 100?

tests/instruments/keithley/test_keithley2450_with_device.py
BenediktBurger, line 1: Thanks for adding device tests.

Why no protocol tests?
Protocol tests are far more useful for us maintainers as they run in each deployment and check that the code still works.

BenediktBurger (changes requested):
Thank you for your work!

PR #1493, opened 2026-06-23
Add type hints to AgilentB1500

ilev-sergey:
- Adds missing type hints for AgilentB1500 and its units
- Fixes all existing type errors highlighted by Pyright
- Fixes bugs connected to incorrect type handling when querying instrument's settings


ilev-sergey:
@coderabbitai review

ilev-sergey:
@coderabbitai review

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -55,49 +56,51 @@ class AgilentB1500(SCPIMixin, Instrument):
...
 
     @property
-    def smu_references(self):
+    def smu_references(self) -> ValuesView[SMU]:
         """Get all SMU instances as dict_values."""
         return self.smus.values()
 
     @property
-    def smu_names(self):
+    def smu_names(self) -> dict:
```
BenediktBurger: If you know the values of the dict, you can do it thus:

```suggestion
    def smu_names(self) -> dict[int, str]:
```

the first value (here int) is the key type and the second one the value (I guess names are strings).

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -2249,16 +2307,15 @@ def query_learn_header(cls, ask, query_type, smu_references, single_command=Fals
         return ret
 
     @staticmethod
-    def to_dict(parameters, names, *args):
+    def to_dict(parameters: str | list, names: list, *args) -> OrderedDict:
```
BenediktBurger: Also for lists you can define the type of their content (I just guessed some types):

```suggestion
    def to_dict(parameters: str | list[str], names: list[str], *args) -> OrderedDict[str, Parameter]:
```

BenediktBurger (commented):
I also started some typing for several instruments, but prefer if someone who has an instrument does it.
Here my changes to the same instrument: #1458 
Note that I also did some changes to some property creators (especially cast=str). However, I only guessed these changes based on tests/descriptions.
Feel free to take from my PR whatever you deem good.

ilev-sergey:
Maybe it would be better if I wait until your changes are in the master, so you don't have to deal with potential conflicts on merge?
I'm also a bit confused about the scope of #1458, since #1490 seems to partially implement typing changes from #1458. Can you share your thoughts on this?

And big thanks for all your work on improving the project's code quality!

BenediktBurger:
I had a large branch where I developed (#1458) which grew over time. In #1490 I extracted the changes just related to the core in order to keep the PR smaller. I guess 1458 won't be merged in its current form, but piecewise.

I prefer if we merge the AgilentB1500 in this PR, or I can extract my AgiltenB1500 changes into its own PR and I would kindly ask you to review these changes.
Smaller PRs (if sensible) are better and also if someone knows the instrument (I do not).

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -338,7 +348,7 @@ class _data_formatting_generic:
         }
         data_names_int = {"Sampling index"}  # convert to int instead of float
 
-        def __init__(self, smu_names, output_format_str):
+        def __init__(self, smu_names: dict, output_format_str: str = ""):
```
BenediktBurger: I guess it is a int, str or a str, str dictionary.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -816,56 +847,57 @@ def sampling_timing(self, hold_bias, interval, number, hold_base=0):
...
+        heads = data_mapped.iloc[[0]].map(lambda x: " ".join(x[1:3]))
         # channel & data_type
-        heads = heads.to_numpy().tolist()  # 2D List
-        heads = heads[0]  # first row
-        data = data.applymap(lambda x: x[3])
-        data.columns = heads
-        return data
+        heads_list = heads.to_numpy().tolist()  # 2D List
+        first_row = heads_list[0]
+        data_twice_mapped = data_mapped.map(lambda x: x[3])
```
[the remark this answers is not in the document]
BenediktBurger: Alternatively, we could increase the pandas minimum to 2.1

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1045,23 +1079,31 @@ def adc_type(self):
...
     ######################################
     # Force Constant Output
     ######################################
-    def force(self, source_type, source_range, output, comp="", comp_polarity="", comp_range=""):
+    def force(
+        self,
+        source_type: str,
+        source_range: int | str,
+        output: float,
+        comp: float | str = "",
```
BenediktBurger: I think None is better to indicate "no new value" than an empty string.
We could keep the logic inside the code (maybe with a deprecation warning), but for the type hints just float | None = None

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1079,36 +1121,36 @@ def force(self, source_type, source_range, output, comp="", comp_polarity="", co
...
-        comp_polarity="",
-        comp_range="",
-        stepsize=0.001,
-        pause=20e-3,
-    ):
+        source_type: str,
+        source_range: int | str,
+        target_output: float,
+        comp: float | str = "",
+        comp_polarity: CompliancePolarity | str = "",
```
BenediktBurger, line 1170: I don't like this "Enum | str", but for backward compatibility, we have to allow it, unfortunately.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1582,49 +1646,50 @@ def __init__(self, parent, channel, **kwargs):
         values=[2e-8, 10],
     )
 
-    def set_output_mode(self, mode, condition=None):
+    def set_output_mode(
+        self, mode: SPGUOutputMode, condition: int | float | None = None
```
BenediktBurger, line 1694: Typically, you do not have to write `int | float`  for a parameter as an int is a subclass of a float and anywhere, where you can use a float, you can also use an int. E.g. here condition: float could also accept 5 (an int).

Ah, I understand it now: int means cycles and float duration, so it is fine.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1634,47 +1699,48 @@ def get_output_mode(self):
...
+    enabled: InstrumentProperty[bool] = Channel.setting(
         "%s {ch}",
         """Control SPGU channel enable/disable state. (``CN``, ``CL``)""",
         validator=strict_discrete_set,
         values={False: "CL", True: "CN"},
         map_values=True,
     )
 
-    load_impedance = Channel.control(
+    load_impedance: InstrumentProperty[float] = Channel.control(
```
BenediktBurger, line 1754: That should not be necessary as this is the default type.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1783,7 +1847,7 @@ def set_measurement_mode(self, measurement_mode: MFCMUMeasurementMode) -> None:
 
         :param measurement_mode: Measurement mode.
         """
-        if hasattr(self.parent, "_data_format") and self.parent._data_format.format in [
+        if self.parent._data_format is not None and self.parent._data_format.format in [
```
BenediktBurger, line 1894: Is it possible that the parent does not have that attribute?
In this case you could go safe:

```suggestion
        if getattr(self.parent, "_data_format", None) is not None and self.parent._data_format.format in [
```
This defaults to None if the attribute does not exist, otherwise it reads it and returns it.

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1931,11 +1995,11 @@ def get(cls, input_value):
 class ADCType(CustomIntEnum):
     """ADC Type."""
 
-    HSADC = (0,)  #: High-speed ADC
-    HRADC = (1,)  #: High-resolution ADC
-    HSADC_PULSED = (2,)  #: High-speed ADC for pulsed measurements
+    HSADC = 0  #: High-speed ADC
+    HRADC = 1  #: High-resolution ADC
```
BenediktBurger, line 2043: That was my guess, would you mind verifying that the corresponding functionality still works?

BenediktBurger (changes requested):
Thank you very much for your work!

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: I added explicit `_data_format = None` in initialization [here](https://github.com/pymeasure/pymeasure/pull/1493/changes/342ea10e59041d946a62de4d52a0f0111c008afe#diff-0a9f5185d06142806c6c4dfc46f4133c909f59f90ff3a8c8dc7292f018460e33R71)

In principle, this part of if can be even omitted now,
```python
if self.parent._data_format is not None
```
but I think it's worth to keep it for explicitness as well

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: Yes, I checked that - works like before
Tuples were unpacked as ints anyway - don't know why they were present initially

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: I'd keep it for consistency and explicitness anyway, if you don't mind

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -403,6 +418,7 @@ def log_failed():
                 name = f" {name}"
 
             status = re.search(r"(?P<number>[0-9]*)(?P<letter>[ A-Z]*)", status_string)
+            assert status is not None  # pattern always matches, assert for type checking
```
BenediktBurger: Either use cast or put some useful text to the assertion. Actually, it could fail if it does not find anything, then it returns None!

pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -879,39 +918,57 @@ def read_channels(self, nchannels):
...
+def _compliance_or_none(value: float | str | None) -> float | None:
+    """Normalize a compliance argument, mapping the deprecated ``""`` sentinel to ``None``.
+
+    .. deprecated:: 0.17.0
+        Passing an empty string to indicate "no value" is deprecated; ``None`` should
+        be used instead.
+    """
+    if value == "":
+        warnings.warn(
+            'Passing "" to indicate "no compliance value" is deprecated;pass None instead.',
```
BenediktBurger: The warning should include where (which method) it is raised, because the user only sees the warning without context.
Also, there should be a space after semi colon.

BenediktBurger (approved):
Thank you for your work!

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: This regex always matches, even for an empty string `""`, so None is not really achievable.

ilev-sergey:
Thank you for review!

BenediktBurger:
Thank you for your work!

PR #1495, opened 2026-06-25
Fix SR830/SR860 aux outputs silently truncating sub-microvolt values

### LINKED ISSUE #821: Critical Bug:  Use of %f instead of %e for values truncates values
Hey, I've been working with this library for a while using the DSOX1102G I noticed that once I use a value smaller than 1e-6, the scope does not receive it correctly and uses the smallest value possible.

This issue occurs as the python default %f max decimal points is 6:
image

The more concise solution would be to use %e which is always supported by SCPI devices:
image

Looking at the repo there are about 19 devices use %f and not %e

This is uncaught by testing as it was never tested for.

binggao1230:
## Summary

The `aux_out_1`–`aux_out_4` properties on both SR830 and SR860 used `%f` in their set commands. Python's `%f` format has a fixed precision of 6 decimal places, so any value with magnitude below 1 µV is silently rounded to `0.000000` before being sent to the instrument.

**Example:**
```python
inst.aux_out_1 = 1e-7   # intended: 100 nV
# → sends "AUXV1,0.000000;"  (WRONG – device receives 0 V instead)
```

Fix: replace `%f` with `%g`, which switches to scientific notation automatically for small magnitudes while keeping a compact decimal representation for normal values. `%g` is a valid SCPI numeric format and is already used in other pymeasure instrument drivers (e.g. ami430, teledyne_oscilloscope, hp34401A).

Fixes #821.

## Changes

- `pymeasure/instruments/srs/sr830.py`: `%f` → `%g` in all four `aux_out_N` set commands.
- `pymeasure/instruments/srs/sr860.py`: same.
- `tests/instruments/srs/test_sr830.py`: add parametrised tests covering sub-microvolt, millivolt, and boundary values for all four outputs.
- `tests/instruments/srs/test_sr860.py`: new file with equivalent tests for SR860 (no test file existed before).

## Test plan

- [ ] `pytest tests/instruments/srs/test_sr830.py` – all 36 tests (8 pre-existing + 28 new) pass
- [ ] `pytest tests/instruments/srs/test_sr860.py` – all 28 new tests pass

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/tests/instruments/srs/test_sr830.py
+++ b/tests/instruments/srs/test_sr830.py
@@ -93,3 +93,22 @@ def test_output_conversion():
     ) as inst:
         conv = inst.output_conversion("X")
         assert conv(inst.x) == pytest.approx(-2.66e-7)
+
+
+@pytest.mark.parametrize("channel, cmd", (
+    ("aux_out_1", "AUXV1,%g;"),
+    ("aux_out_2", "AUXV2,%g;"),
+    ("aux_out_3", "AUXV3,%g;"),
+    ("aux_out_4", "AUXV4,%g;"),
+))
+@pytest.mark.parametrize("value", (1e-9, 1e-7, -1e-7, 0.5, 1.5, -10.5, 10.5))
+def test_aux_out_small_values(channel, cmd, value):
+    """Verify that aux outputs transmit sub-microvolt values without truncation.
+
+    Using %f would silently round values smaller than 1e-6 to zero.
+    """
+    with expected_protocol(
+        SR830,
+        [(cmd % value, None)],
+    ) as inst:
+        setattr(inst, channel, value)
--- a/tests/instruments/srs/test_sr830.py
+++ b/tests/instruments/srs/test_sr830.py
@@ -95,20 +95,18 @@ def test_output_conversion():
         assert conv(inst.x) == pytest.approx(-2.66e-7)
 
 
-@pytest.mark.parametrize("channel, cmd", (
-    ("aux_out_1", "AUXV1,%g;"),
-    ("aux_out_2", "AUXV2,%g;"),
-    ("aux_out_3", "AUXV3,%g;"),
-    ("aux_out_4", "AUXV4,%g;"),
+@pytest.mark.parametrize(("value", "command"), (
+    (1e-9, "AUXV1,1e-09;"),
+    (1e-7, "AUXV1,1e-07;"),
+    (-1e-7, "AUXV1,-1e-07;"),
 ))
-@pytest.mark.parametrize("value", (1e-9, 1e-7, -1e-7, 0.5, 1.5, -10.5, 10.5))
-def test_aux_out_small_values(channel, cmd, value):
+def test_aux_out_small_values(value, command):
     """Verify that aux outputs transmit sub-microvolt values without truncation.
 
     Using %f would silently round values smaller than 1e-6 to zero.
     """
     with expected_protocol(
         SR830,
-        [(cmd % value, None)],
+        [(command, None)],
     ) as inst:
-        setattr(inst, channel, value)
+        inst.aux_out_1 = value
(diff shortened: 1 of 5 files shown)


tests/instruments/srs/test_sr830.py
```diff
@@ -93,3 +93,22 @@ def test_output_conversion():
...
+))
+@pytest.mark.parametrize("value", (1e-9, 1e-7, -1e-7, 0.5, 1.5, -10.5, 10.5))
+def test_aux_out_small_values(channel, cmd, value):
+    """Verify that aux outputs transmit sub-microvolt values without truncation.
+
+    Using %f would silently round values smaller than 1e-6 to zero.
+    """
+    with expected_protocol(
+        SR830,
+        [(cmd % value, None)],
```
BenediktBurger: I think it would be better to explicitly give the string sent to the device, now we only verify that we use %g, not that we send a number unequal zero.

tests/instruments/srs/test_sr830.py
```diff
@@ -93,3 +93,22 @@ def test_output_conversion():
     ) as inst:
         conv = inst.output_conversion("X")
         assert conv(inst.x) == pytest.approx(-2.66e-7)
+
+
+@pytest.mark.parametrize("channel, cmd", (
+    ("aux_out_1", "AUXV1,%g;"),
+    ("aux_out_2", "AUXV2,%g;"),
+    ("aux_out_3", "AUXV3,%g;"),
+    ("aux_out_4", "AUXV4,%g;"),
```
BenediktBurger: I think we do not need to test all four channels, one channel as an example is enough (they are all the same class).

BenediktBurger (changes requested):

You may add yourself to the authors.txt file as well.

binggao1230:
Thanks, updated.

I reduced the aux-output regression tests to `aux_out_1` as the representative channel for each instrument class, and the test table now spells out the exact command string sent to the device (`1e-09`, `1e-07`, `-1e-07`) instead of formatting the expected command with `%g` in the assertion setup.

I also added myself to `AUTHORS.txt`.

Verification run locally:

```text
python -m pytest tests/instruments/srs/test_sr830.py tests/instruments/srs/test_sr860.py -q
python -m ruff check pymeasure/instruments/srs/sr830.py pymeasure/instruments/srs/sr860.py tests/instruments/srs/test_sr830.py tests/instruments/srs/test_sr860.py
git diff --check
```

tests/instruments/srs/test_sr830.py
```diff
@@ -93,3 +93,20 @@ def test_output_conversion():
     ) as inst:
         conv = inst.output_conversion("X")
         assert conv(inst.x) == pytest.approx(-2.66e-7)
+
+
+@pytest.mark.parametrize(("value", "command"), (
+    (1e-9, "AUXV1,1e-09;"),
+    (1e-7, "AUXV1,1e-07;"),
+    (-1e-7, "AUXV1,-1e-07;"),
```
BenediktBurger, line 101: I liked your previous selection of values (including some around 1, 10, that is numbers which are not going to be exponential ones).

I will merge it either way, just tell me (or push the change).

BenediktBurger (approved):

binggao1230:
Rebased onto master to resolve the AUTHORS.txt conflict — no code changes, SRS tests still pass.

tests/instruments/srs/test_sr830.py
BenediktBurger: @gaoflow, please respond to this comment (either saying not to change it or after having added a case for something like 10).

binggao1230:
Rebased onto master — the only conflict was the AUTHORS.txt line, both names kept. No code changes; SRS tests still 14 passed.

tests/instruments/srs/test_sr830.py
binggao1230: Added it in a4cc1ac5 — the parameter table now includes (10.0, "AUXV1,10;") (and the SR860 equivalent), so the non-exponential path is pinned too: with %f the device would get "10.000000", with %g it gets "10". All 22 SRS tests pass.

BenediktBurger:
Thank you for your fix!

PR #1496, opened 2026-06-27
Added support for the Keysight E364xA series of power supplies

jgrage:
I added support for the Keysight E364xA power supply series. The series has six single channel versions (E3640A to E3645A) and four dual output channel versions (E3646A to E3649A), so I created two base classes ```KeysightE364XASingleOutput``` and ```KeysightE364XADualOutput``` together with ```KeysightE364XAChannel``` for the individual device classes. The individual devices inherit their behavior from the base classes and provide their own maximum voltage and maximum current capabilities via class attributes.

One special feature of the device series is that each channel can be set into either "LOW" or "HIGH" output voltage mode via the ```VOLT:RANG``` SCPI command. I therefore declared the ```voltage_setpoint``` and ```current_limit``` of the device/channel as dynamic properties to be able to change the allowed validator range on the fly.
To update the validator range each time the user changes the output range, I registered an ```update_validator_range``` method as a set_process for the channels ```range``` property.

I tested correct operation on hardware with a E3643A (single output) and a E3649A (dual output) device connected with a [USBGpib v2](https://github.com/xyphro/UsbGpib) dongle, autogenerated the protocol tests and combined them into a single file under ```tests/instruments/keysight/test_keysightE364xA.py```.

### DIFF
--- a/docs/api/instruments/keysight/keysightE364xA.rst
+++ b/docs/api/instruments/keysight/keysightE364xA.rst
@@ -1,0 +1,45 @@
+##########################################
+Keysight E364xA Series of Power Supplies
+##########################################
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3640A
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3641A
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3642A    
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3643A    
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3644A    
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3645A    
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3646A    
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3647A    
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3648A    
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3649A    
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE364XASingleOutput
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE364XADualOutput
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE364XAChannel
+    :members:
+    :show-inheritance:
--- a/docs/api/instruments/keysight/keysightE364xA.rst
+++ b/docs/api/instruments/keysight/keysightE364xA.rst
@@ -1,6 +1,6 @@
-##########################################
+########################################
 Keysight E364xA Series of Power Supplies
-##########################################
+########################################
 
 .. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3640A
     :show-inheritance:
--- a/docs/api/instruments/keysight/keysightE364xA.rst
+++ b/docs/api/instruments/keysight/keysightE364xA.rst
@@ -8,28 +8,28 @@ Keysight E364xA Series of Power Supplies
 .. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3641A
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3642A    
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3642A
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3643A    
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3643A
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3644A    
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3644A
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3645A    
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3645A
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3646A    
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3646A
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3647A    
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3647A
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3648A    
+.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3648A
     :show-inheritance:
 
-.. autoclass:: pymeasure.instruments.keysight.keysightE364xA.KeysightE3649A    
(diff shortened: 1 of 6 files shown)


docs/api/instruments/keysight/keysightE364xA.rst
```diff
@@ -0,0 +1,45 @@
+##########################################
+Keysight E364xA Series of Power Supplies
+##########################################
```
BenediktBurger: Please keep the hashes as long as the text.

pymeasure/instruments/keysight/keysightE364xA.py
```diff
@@ -0,0 +1,352 @@
...
+    :param voltage_range: select the initial output range (default="LOW") during initialization.
+    :type voltage_range: str ("LOW") or str ("HIGH")
+    """
+
+    _default_name = "KeysightE364XASingleOutput"
+    _max_voltage = {"LOW": 35.0, "HIGH": 60.0}
+    _max_current = {"LOW": 1.4, "HIGH": 0.8}
+    _range_map = {"LOW": "P35V", "HIGH": "P60V"}
+
+    def __init__(self, adapter, name=None, voltage_range="LOW", **kwargs):
```
BenediktBurger, line 115: Why don't you just add the default name string as default?

pymeasure/instruments/keysight/keysightE364xA.py
```diff
@@ -0,0 +1,352 @@
...
+        super().__init__(
+            adapter,
+            name if name is not None else self._default_name,
+            **kwargs,
+        )
+        # Set 'update_validator_range' as a set process for the channels range command
```
BenediktBurger: Don't add comments which describe the what (as the code already describes that), use comments only for the why (if needed).

pymeasure/instruments/keysight/keysightE364xA.py
```diff
@@ -0,0 +1,352 @@
...
+
+    The Channel supports `HIGH` and a `LOW` output range that can be changed during operation.
+    However switching to different range while the output is set to parameters that the newly
+    selected range does not support will automatically alter those to match the maximum
+    allowed values of the new range setting.
+    """
+    voltage_setpoint = Channel.control(
+        "INST:NSEL {ch};:VOLT?",
+        "INST:NSEL {ch};:VOLT %g",
+        """Control output voltage setpoint. Range depends on selected LOW or HIGH mode.""",
```
BenediktBurger: In which units (I guess in Volts - but that is quite important to know)?

pymeasure/instruments/keysight/keysightE364xA.py
```diff
@@ -0,0 +1,352 @@
...
+        )
+        # Set 'update_validator_range' as a set process for the channels range command
+        self.ch_1.range_set_process = self.ch_1.update_validator_range
+        self.ch_2.range_set_process = self.ch_2.update_validator_range
+        # Set mapping of "HIGH" and "LOW" to device specific range keywords
+        self.ch_1.range_values = self._range_map
+        self.ch_2.range_values = self._range_map
+        # Initialize device with desired output range modes
+        self.ch_1.range = voltage_range[0]
+        self.ch_2.range = voltage_range[1]
```
BenediktBurger: Even though it is not a large improvement, but a suggestion and maybe something good to know:
Per default, the channels are added to a `channels` dictionary, so you could iterate over it:
~~~python
for device in self.channels.values():
    device.range_set_process = device.update_validator_range
    ...
~~~

tests/instruments/keysight/test_keysightE364xA.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+@pytest.mark.parametrize("comm_pairs, value", (
+    ([(b'VOLT:RANG P35V', None),
+      (b'OUTP 1', None)],
+     True),
+    ([(b'VOLT:RANG P35V', None),
+      (b'OUTP 0', None)],
+     False),
+))
```
BenediktBurger, line 65: you could also parametrize just the string "OUTP 0/1" and the value True/False, instead of the whole communication -> or you used the generator?

tests/instruments/keysight/test_keysightE364xA.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+def test_ch_2_current_limit_getter():
+    with expected_protocol(
+            KeysightE3649A,
+            [(b'INST:NSEL 1;:VOLT:RANG P35V', None),
+             (b'INST:NSEL 2;:VOLT:RANG P35V', None),
+             (b'INST:NSEL 2;:CURR?', b'+6.00000000E-01\n')],
+    ) as inst:
+        assert inst.ch_2.current_limit == 0.6
```
BenediktBurger: A repetition of the tests (just a different channel) is not needed. You could also make one of the above tests for channel 1 and the other one for channel 2. The channels are identical, so one channel could be a prototype for all channels.

BenediktBurger (changes requested):
Thank you very much for your contribution (and your patience).

pymeasure/instruments/keysight/keysightE364xA.py
jgrage: Sorry for my late reply, I was quite busy at work.
Setting the default argument ```name=self._default_name``` in the constructor does not work since default arguments are evaluated at function definition time. A reference to ```self``` is not valid here. Thats why I used the common pattern of setting the default argument to none and let the assignment be handled inside the constructor at call time.

tests/instruments/keysight/test_keysightE364xA.py
jgrage: Yes, I indeed used the automated test generator and two of my instruments to generate the tests from the device communication.

tests/instruments/keysight/test_keysightE364xA.py
jgrage: You're right. I removed those redundant test cases.

pymeasure/instruments/keysight/keysightE364xA.py
BenediktBurger: I meant to write it explicitly as a string: `name="whatever you want to add"`.

pymeasure/instruments/keysight/keysightE364xA.py
```diff
@@ -0,0 +1,348 @@
...
+    def __init__(self, adapter, name=None, voltage_range=("LOW", "LOW"), **kwargs):
+        super().__init__(
+            adapter,
+            name if name is not None else self._default_name,
+            **kwargs,
+        )
+        for i, ch in enumerate(self.channels.values()):
+            ch.range_set_process = ch.update_validator_range
+            ch.range_values = self._range_map
+            ch.range = voltage_range[i]
```
BenediktBurger: Actually, you could use the key (I don't know by heart whether we store it as a string, so maybe you have to convert it to int):

```suggestion
        for ident, ch in self.channels.items():
            ch.range_set_process = ch.update_validator_range
            ch.range_values = self._range_map
            ch.range = voltage_range[ident]
```

tests/instruments/keysight/test_keysightE364xA.py
BenediktBurger: Ok. If you want to, you can refactor it, but I don't require it.

BenediktBurger (approved):
A few minor remarks.

pymeasure/instruments/keysight/keysightE364xA.py
jgrage: For that each subclass would need its own constructor with its respective model number supplied as the default ```name``` argument. Currently each child class just substitutes their own model number as the ```_default_name``` class attribute and the name is then set when the inherited constructor from the base class is called. So it saved me from writing custom ```__init__```s for every child class.
Let me know if adding a custom constructor for each child class is your preferred way of handling the device names and I will change it accordingly.

pymeasure/instruments/keysight/keysightE364xA.py
jgrage: Yes, that way looks cleaner. I looked it up and the key is indeed an integer value, but the channel indexing starts at 1 so i modified ```ch.range = voltage_range[ident - 1]```

jgrage:
That is strange. Now the CI ruff linting fails with "Import block is un-sorted or un-formatted" although imports did not change compared to the previous commits and successful CI runs. On my local machine, ruff is not complaining...

tests/instruments/keysight/test_keysightE364xA.py
jgrage: Lets keep it that way

pymeasure/instruments/keysight/keysightE364xA.py
BenediktBurger: Ok, that is fine with me.
I personally would use a generic name in this init call here and not bother with explicit names for the subclasses.

In general, you can use these name to distinguish devices. And if someone needs that feature, they can supply their own names.

BenediktBurger:
This is a new ruff rule from 0.16.0. It failed on master and I had to adjust quite some files.

If you update ruff, you may fix it easily with `ruff check --fix`.

BenediktBurger:
May I merge?

jgrage:
> May I merge?

Yes please :)

PR #1498, opened 2026-06-30
Add step_type argument for `FloatParameter`

### LINKED ISSUE #1491: Suggestion: extending step functionality for `FloatParameter`
At the moment, `FloatParameter` has a `step` argument
https://github.com/pymeasure/pymeasure/blob/b52b10b8718bd0cb6a44c2ab4e72af83151b9e01/pymeasure/experiment/parameters.py#L255

that is used to create arrow buttons in the input field in GUI:
https://github.com/pymeasure/pymeasure/blob/b52b10b8718bd0cb6a44c2ab4e72af83151b9e01/pymeasure/display/inputs.py#L212-L215

I suggest extending its functionality by adding a `step_type` argument to `FloatParameter` that can be either `"linear"` or `"log"` for now. For `"linear"`, which will be the default value, the arrows will behave as previously, adding linear step to initial value after each click. For `"log"`, the buttons would instead multiply or divide the current value by `step` on each click.

This would be useful for parameters that span several orders of magnitude (for example, frequency or pulse duration), where multiplicative stepping is often more convenient than additive stepping.

ilev-sergey:
Closes #1491


pymeasure/display/inputs.py
```diff
@@ -267,6 +267,14 @@ def textFromValue(self, value):
...
         return string
 
+    def stepBy(self, steps):
+        value = self.value()
+        if self._parameter.step_type == "log":
+            sign = 1 if value >= 0 else -1
+            self.setValue(value * self._parameter.step ** (sign * steps))
+        else:
+            super().stepBy(steps)
+
```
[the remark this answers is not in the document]
ilev-sergey: > * When the current `value` is `0`, multiplicative stepping leaves it at `0` permanently — the arrows become inert. This is reachable when a log `FloatParameter` is left at a `0` value/default.

I think that's expectable behavior 

> * FloatParameter permits step_type="log" with step=None. The UI guards this (buttons disabled via stepEnabled), but a direct stepBy call would raise TypeError on None ** ....

I'm not sure that additional guarding is really required here: this combination of values is not smth reasonable and the case when user calls `stepBy` from code for this combination is highly unlikable. UI handles this automatically, as buttons are not available for `step=None`

pymeasure/experiment/parameters.py
```diff
@@ -253,16 +253,37 @@ class FloatParameter(Parameter):
...
+        decimals=15,
+        step=None,
+        step_type="linear",
+        **kwargs,
+    ):
+        if step_type not in ("linear", "log"):
+            raise ValueError(
+                f"FloatParameter step_type must be 'linear' or 'log', not '{step_type}'"
+            )
+        if step_type == "log" and step is not None and step <= 0:
```
CasperSchippers: I´d even argue that the value should be (striclty) larger then 1, having it smaller then 1 is very confusing (value decreases when pressing up arrow).

pymeasure/display/inputs.py
```diff
@@ -267,6 +267,14 @@ def textFromValue(self, value):
         string = re.sub(r"e(-?)0*(\d+)", r"e\1\2", string)
         return string
 
+    def stepBy(self, steps):
+        value = self.value()
+        if self._parameter.step_type == "log":
+            sign = 1 if value >= 0 else -1
```
CasperSchippers, line 273: Is this what we want?? When playing around I was very confused that the behaviour changed between values >0 and <0; for me having the up arrow always multiply with 10 makes more sense (and is in line with the description).

pymeasure/experiment/parameters.py
```diff
@@ -253,16 +253,37 @@ class FloatParameter(Parameter):
...
 
-    def __init__(self, name, units=None, minimum=-1e9, maximum=1e9,
-                 decimals=15, step=None, **kwargs):
+    def __init__(
+        self,
+        name,
+        units=None,
+        minimum=-1e9,
+        maximum=1e9,
+        decimals=15,
```
CasperSchippers: Maybe with the log-step-type added, it might make sense to widen these default ranges a bit? The decimals also controls the smallest number it can hold, going down you very quickly reach a point where this decimals starts trimming accuracy of the number you entered (or completely zeroing it), so maybe setting the default valuee to 30 or 60 would make sense? Maybe even 300 (higher is unwise, because you get close to the smallest value a double can hold, making the value unpredictable)?

CasperSchippers (changes requested):
I like this addition a lot. Have a few minor suggestions, but apart from that it looks good.

pymeasure/display/inputs.py
ilev-sergey: I personally expect that up arrow always increases the value itself (not abs), no matter which sign it has. I don't mind changing it though, if you find this behavior to be less obvious

CasperSchippers: I understand your expectation as well...
Would it help if we replace the up/down arrows with ×÷ symbols? Allegedly (I haven't tried it yet) we can change the symbols using something like this:

```python
spinbox.setStyleSheet("QDoubleSpinBox::up-arrow { image: url(\"data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg'><text x='50%' y='65%' font-size='12' text-anchor='middle'>×</text></svg>\"); } QDoubleSpinBox::down-arrow { image: url(\"data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg'><text x='50%' y='75%' font-size='12' text-anchor='middle'>÷</text></svg>\"); }")
```

What do you think about this kind of solution?

If you feel that even with such modification your original configuration feels more natural I'm also fine with that. In the end, I don't have to work with it.

CasperSchippers (approved):
Looks good, only open item is the discussion on the arrow.

BenediktBurger:
Thank you both for that improvement!

pymeasure/display/inputs.py
ilev-sergey: I like your idea. I think with ×÷ symbols the behavior is more unambiguous. However, I wasn't able to implement it with your snippet - arrows were replaced with empty boxes.

I asked Claude Fable about this and after a few iterations it made following changes. I'm not an expert in Qt styles so I don't really understand what's going on here and can't review that.

________

New style class
```py
class _MultiplyDivideSpinStyle(QtWidgets.QProxyStyle):
    """Proxy style that draws × and ÷ signs on the spin box buttons instead
    of the default up and down arrows, to indicate multiplicative stepping.
    """

    def drawComplexControl(self, control, option, painter, widget=None):
        CC = QtWidgets.QStyle.ComplexControl
        if control != CC.CC_SpinBox:
            super().drawComplexControl(control, option, painter, widget)
            return
        SC = QtWidgets.QStyle.SubControl
        State = QtWidgets.QStyle.StateFlag
        # Draw the spin box without its buttons, then draw the buttons
        # ourselves with × and ÷ symbols instead of the arrows
        opt = QtWidgets.QStyleOptionSpinBox(option)
        opt.subControls &= ~(SC.SC_SpinBoxUp | SC.SC_SpinBoxDown)
        opt.buttonSymbols = QtWidgets.QAbstractSpinBox.ButtonSymbols.NoButtons
        super().drawComplexControl(control, opt, painter, widget)

        step_flags = QtWidgets.QAbstractSpinBox.StepEnabledFlag
        for sub_control, symbol, step_flag in (
                (SC.SC_SpinBoxUp, "×", step_flags.StepUpEnabled),
                (SC.SC_SpinBoxDown, "÷", step_flags.StepDownEnabled)):
            if not option.subControls & sub_control:
                continue
            rect = self.subControlRect(control, option, sub_control, widget)
            button = QtWidgets.QStyleOption(option)
            button.rect = rect
            button.state &= ~(State.State_Sunken | State.State_MouseOver
                              | State.State_HasFocus)
            if not option.stepEnabled & step_flag:
                button.state &= ~State.State_Enabled
            elif option.activeSubControls & sub_control:
                button.state |= option.state & (State.State_Sunken
                                                | State.State_MouseOver)
            self.drawPrimitive(QtWidgets.QStyle.PrimitiveElement.PE_PanelButtonBevel,
                               button, painter, widget)
            color_group = (QtGui.QPalette.ColorGroup.Normal
                           if button.state & State.State_Enabled
                           else QtGui.QPalette.ColorGroup.Disabled)
            painter.save()
            # scale the symbol to the button size, based on its actual extent
            font = painter.font()
            font.setPixelSize(24)
            extent = QtGui.QFontMetricsF(font).tightBoundingRect(symbol)
            scale = min(rect.width() * 0.6 / extent.width(),
                        rect.height() * 0.6 / extent.height())
            font.setPixelSize(max(6, round(24 * scale)))
            painter.setFont(font)
            # draw the symbol optically centered in the button
            extent = QtGui.QFontMetricsF(font).tightBoundingRect(symbol)
            origin = QtCore.QPointF(
                rect.x() + (rect.width() - extent.width()) / 2 - extent.x(),
                rect.y() + (rect.height() - extent.height()) / 2 - extent.y())
            painter.setPen(option.palette.color(color_group,
                                                QtGui.QPalette.ColorRole.ButtonText))
            painter.drawText(origin, symbol)
            painter.restore()
```

New ScientificInput internal method
```py

    def _set_log_step_symbols(self):
        """Show × and ÷ signs on the spin buttons instead of the default
        arrows, indicating that stepping is multiplicative.
        """
        # base the proxy on the current application style, since a
        # default-constructed QProxyStyle uses the platform style instead
        base_style = QtWidgets.QStyleFactory.create(
            QtWidgets.QApplication.style().objectName())
        # keep a reference, since setStyle does not take ownership
        self._log_step_style = _MultiplyDivideSpinStyle(base_style)
        self.setStyle(self._log_step_style)

```

Update in ScientificInput `__init__`
```py
  if parameter.step_type == "log":
      self._set_log_step_symbols()
```
________

I tested these changes on Windows 11 (both light and dark themes) and Windows 7 - it works fine, but buttons are a bit out of style.

image_2026-07-15_17-37-23

image_2026-07-15_17-36-23

Capture

If you're fine with adding this custom style class, I'd prefer these ×÷ buttons with default multiplication behavior. If not, I'd keep sign-aware behavior for arrows

CasperSchippers: Hmm that is more complicated that I thought/hoped.... I tried looking at I a bit but I didn't have enough time to delve into this deeper, so I'd propose to for now stick to the original arrows With the behaviour that you feel works best, if you agree.

ilev-sergey: Let's keep the current behavior then. Anyway, thank you for the idea!

PR #1504, opened 2026-07-08
Actually hide columns that are not selected to be shown.

### LINKED ISSUE #1501: Bug: TableWidget ignores `columns` argument
## Describe the bug
Supplying the `columns` argument to `TableWidget` has no effect on which columns are shown in the table after running a procedure. 

## To Reproduce
Here is a modified version of the TableWidget example with two random numbers, and only `Random Number 1` and `Iteration` are included in `columns`. 

```python
 
import logging

import sys
import tempfile
import random
from time import sleep
from pymeasure.display.windows import ManagedWindowBase
from pymeasure.display.widgets import TableWidget, LogWidget

    iterations = IntegerParameter('Loop Iterations', default=10)

    DATA_COLUMNS = ['Iteration', 'Random Number 1', 'Random Number 2']

    def startup(self):
        random.seed(self.seed)

    def execute(self):
            data = {
                'Iteration': i,
                'Random Number 1': random.random(),
                'Random Number 2': random.random(),
            }
            self.emit('results', data)
            self.emit('progress', 100 * i / self.iterations)
            sleep(self.delay)
            if self.should_stop():
                break


class MainWindow(ManagedWindowBase):

    def __init__(self):
        widget_list = (TableWidget("Experiment Table",
                                columns=['Iteration', 'Random Number 1'],
                                # columns=[],        # this also works just fine
                                by_column=True,
                                ),
                    LogWidget("Experiment Log"),
                    )
        super().__init__(
            inputs=['iterations', 'delay', 'seed'],
            displays=['iterations', 'delay', 'seed'],
            widget_list=widget_list,
        )
        logging.getLogger().addHandler(widget_list[1].handler)
        log.setLevel(self.log_level)
        log.info("ManagedWindow connected to logging")

if __name__ == "__main__":
    window = MainWindow()
    window.show()
    sys.exit(app.exec())
```

## Expected behavior
Only the specified columns should appear in the results table after queueing/running the procedure. 

## Version
Version: master @ 11e261cc8
Python 3.14.6

CasperSchippers:
closes #1501

### DIFF
--- a/pymeasure/display/widgets/table_widget.py
+++ b/pymeasure/display/widgets/table_widget.py
@@ -570,7 +570,8 @@ class Table(QtWidgets.QTableView):
         if model.columnCount() != current_col_count:
             # Hide no-show columns
             for i in range(model.columnCount()):
-                label = model.headerData(i, QtCore.Qt.Orientation.Horizontal, QtCore.Qt.ItemDataRole.DisplayRole)
+                label = model.headerData(i, QtCore.Qt.Orientation.Horizontal,
+                                         QtCore.Qt.ItemDataRole.DisplayRole)
                 self.setColumnHidden(i, label not in self.parent().columns)
 
     def set_color(self, table, color):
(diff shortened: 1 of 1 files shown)


BenediktBurger:
Thanks for addressing it!

PR #1507, opened 2026-07-12
Fix DSPBase.auto_gain and .gain returning list instead of scalar

### LINKED ISSUE #1478: DSB_Base auto_gain broken
https://github.com/pymeasure/pymeasure/blob/ca2b837acea629b810cfc0656b0c21604f82860b/pymeasure/instruments/signalrecovery/dsp_base.py#L301

This cannot work as values always returns a list.

Also check the rest of the driver.

hmms:
## Summary

`DSPBase.auto_gain` (`pymeasure/instruments/signalrecovery/dsp_base.py`) called `int(self.values("AUTOMATIC")) == 1`, but `Instrument.values()` always returns a `list`, so reading `auto_gain` raises `TypeError` instead of returning a bool.

While fixing that, I found the same anti-pattern in the neighboring `gain` getter: it returned the raw single-element list from `self.values("ACGAIN")` instead of a scalar.

Both are fixed by indexing into the result, matching the existing `sensitivity` property a few lines above (`self.values("SEN.")[0]`).

Fixes #1478

## Test plan

- Added `test_gain_getter`, `test_auto_gain_getter`, `test_auto_gain_setter_enable`, `test_auto_gain_setter_disable` to `tests/instruments/signalrecovery/test_dspbase.py`, following the existing `expected_protocol` pattern in that file.
- `pytest tests/instruments/signalrecovery/` — 31/31 passed.
- `ruff check` on both changed files — clean.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -112,3 +112,4 @@ Niklas Schulz
 Haotian Zhang
 Maduka Ariyasiri
 Max Tweedale
+Muralidhar M Shenoy
--- a/pymeasure/instruments/signalrecovery/dsp_base.py
+++ b/pymeasure/instruments/signalrecovery/dsp_base.py
@@ -258,7 +258,7 @@ class DSPBase(Instrument):
     @property
     def gain(self):
         """Control the AC gain of signal channel amplifier."""
-        return self.values("ACGAIN")
+        return self.values("ACGAIN")[0]
 
     @gain.setter
     def gain(self, value):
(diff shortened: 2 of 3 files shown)


BenediktBurger (approved):

If you want to, you can add yourself to the authors.txt file. If not, please tell me to merge the PR as is.

hmms:
Thanks, Benedikt! I've added myself to AUTHORS.txt in the latest commit. Please go ahead and merge whenever you're ready.

PR #1508, opened 2026-07-13
Add flake8 whitespace checks

### LINKED ISSUE #1506: Ruff does not find all whitespace errors
TL;DR: Ruff is bad in whitespace errors, so we might need to use flake8 for some rules.

## Analysis

Why Ruff didn’t flag it

Even with extend-select = ["E"], Ruff currently:

    Does not enforce all inline comment spacing rules consistently
    Especially misses E261 in many real-world cases
    This is a known gap due to Ruff’s parser-based implementation (not token-based like pycodestyle)

So again, this is not your config — it’s a limitation.
Big picture (what you’re running into)

You’ve essentially discovered this:

    Ruff ≠ strict pycodestyle clone

Specifically:

    Weak spots:
        E231 (comma spacing in signatures)
        E261 (inline comment spacing)
    Strong spots:
        correctness (F rules)
        imports, dead code, complexity
        speed

Whitespace is intentionally de-emphasized in the linter because Ruff expects: → formatting tools to handle it

BenediktBurger:
Ruff is currently not good in whitespace checks, so we need flake as fallback.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,8 @@
 [flake8]
-exclude = .git,__pycache__,docs/conf.py,build,dist
+exclude = .git,__pycache__,docs/conf.py,build,dist,.venv
 max-line-length = 100
 max-complexity = 15
+select = E2, E3
+extend-ignore = E121, E123, E126, E226, E227, E241, E704, W503, W504
 per-file-ignores =
     __init__.py:F401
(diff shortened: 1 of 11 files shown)


BenediktBurger:
In this run (https://github.com/pymeasure/pymeasure/pull/1508/changes), errors are shown as annotations (for a test commit with broken formatting.

PR #1510, opened 2026-07-16
Improve Parameter and Procedure

BenediktBurger:


pymeasure/experiment/procedure.py
```diff
@@ -59,17 +78,19 @@ class Procedure:
...
-        ABORTED: 'Aborted', QUEUED: 'Queued',
-        RUNNING: 'Running'
-    }
+    FINISHED = ProcedureStatus.FINISHED
+    FAILED = ProcedureStatus.FAILED
+    ABORTED = ProcedureStatus.ABORTED
+    QUEUED = ProcedureStatus.QUEUED
+    RUNNING = ProcedureStatus.RUNNING
+
+    STATUS_STRINGS = STATUS_STRINGS
```
BenediktBurger: These are unused, but I do not know whether someone outside this repo uses them, therefore kept them.

pymeasure/experiment/procedure.py
```diff
@@ -37,6 +39,23 @@
...
+    RUNNING = 4
+
+
+STATUS_STRINGS = {
+    ProcedureStatus.FINISHED: "Finished",
+    ProcedureStatus.FAILED: "Failed",
+    ProcedureStatus.ABORTED: "Aborted",
+    ProcedureStatus.QUEUED: "Queued",
+    ProcedureStatus.RUNNING: "Running",
+}
```
CasperSchippers: If you're reworking this; is there a reason to not use a StrEnum? Then you can cover this in a single use?

By the way: I like the enums a lot better!

pymeasure/experiment/procedure.py
```diff
@@ -125,34 +146,35 @@ def get_datapoint(self):
...
         """
         if not self._parameters:
             self._parameters = {}
+        self._param_values: dict[str, Any] = {}
         for item, parameter in inspect.getmembers(self.__class__):
             if isinstance(parameter, Parameter):
                 self._parameters[item] = deepcopy(parameter)
                 if parameter.is_set():
-                    setattr(self, item, parameter.value)
+                    self._param_values[item] = parameter.value
```
CasperSchippers: ~I'm not a huge fan of the assignment we did here (a lot of magic going on, and you notice that a lot of linters don't really understand what is going on), but does this work correctly? How can you now from a procedure get the parameter values?~

Whoops, forgot to remove this comment when I realized how the parameters now work.

Open question still though: is this now working properly, or is this now double work with the inner workings of the parameters?

(same question for the metadata)

pymeasure/experiment/parameters.py
```diff
@@ -48,22 +55,60 @@ class Parameter:
...
+    def __set_name__(self, owner: type, name: str) -> None:
+        self._attr_name = name
+
+    @overload
+    def __get__(self, obj: None, objtype: type): ...
+
+    @overload
+    def __get__(self, obj: object, objtype: type) -> T: ...
+
+    def __get__(self, obj, objtype=None):
```
CasperSchippers: When I first saw this, I was a bit sceptical (wondering whether this is not too complicated/confusing) but maybe this actually a decent approach, turning it into an elaborate property, forgoing the "magic" switcharoo that we did before. This at least is (I think)  a lot more consistent/predictable (if not still maybe difficult to debug if something isn´t working properly, but maybe not worse than with the current approach).

Anyways, I just wanted to double check if we think this is a good approach. After some thinking (and revising this comment at least 8 times), I'm fine with it.

pymeasure/experiment/parameters.py
```diff
@@ -48,22 +55,60 @@ class Parameter:
...
+        if obj is None:
+            return self
```
CasperSchippers: Is there a good reason to store the value in the containing class (procedure) in a newly created _param_values dict? why not use the _value field of the parameter class itself? that already exists, and keeps the information with the parameter instance.

(same argument for the metadata)

pymeasure/experiment/parameters.py
```diff
@@ -22,10 +22,17 @@
...
+from collections.abc import Callable
+from typing import Any, Generic, TypeVar, overload
+
 import numpy as np
 
+T = TypeVar('T')
+
 
-class Parameter:
+class Parameter(Generic[T]):
```
CasperSchippers: Seeing that the Parameter and Metadata share the same property-like behaviour (depending a bit though on the design choice of storing the value in the _value or in the _parameter_values list), it could be an option to make a base-class with that logic?

pymeasure/experiment/parameters.py
```diff
@@ -48,22 +55,60 @@ class Parameter:
...
+
+    def __get__(self, obj, objtype=None):
+        if obj is None:
+            return self
+        param_values = getattr(obj, '_param_values', None)
+        if param_values is not None and self._attr_name in param_values:
+            return param_values[self._attr_name]
+        parameters = getattr(obj, '_parameters', None)
+        if parameters is not None and self._attr_name in parameters:
+            param = parameters[self._attr_name]
```
CasperSchippers: (only relevant if the the self._values vs obj._param_values comment concludes in keeping the current obj._param_values)

Is this not a roundabout way of _just_ getting `self`?

(same argument for the metadata)

pymeasure/experiment/procedure.py
CasperSchippers: For backwards compatibility that makes sense, but maybe then it is good to add a remark regarding that in a comment.

I noticed you replaced the `Procedure.<status>` everywehere in the code with the `ProcedureStatus.<status>` which, I guess, makes sense. And I guess you kept the above block with `FINISHED = ProcedureStatus.Finished` there as well for backwards compatibility? In that case I would also add a comment there regarding specifically that.

Only question is then: is backwards compatibility needed here, as this is internal workings rather than advertised functionality? And if yes, do we plan to keep this indefinately?

CasperSchippers (commented):
Hey, thanks for this work; It does improve the code a lot!

In general, I think I like this new property-like behaviour of the Parameters, but I think we need to be conscious of thes design desicion, so I added some remarks about that.

pymeasure/experiment/procedure.py
BenediktBurger: I changed it to StrEnum and removed the obsolete definitions in Procedure itself.

pymeasure/experiment/parameters.py
BenediktBurger: Done.

pymeasure/experiment/parameters.py
BenediktBurger: The overloads are not that magic, they just distinguish whether it is a class attribute or an instance attribute (depending on the parent `obj`). It is only relevant for typing.

BenediktBurger:
@CasperSchippers would you mind taking a look at my changes? Thanks!

pymeasure/experiment/parameters.py
CasperSchippers: Sure, I didn't specifically mean this overloading, but in general the use of the __get__ method. It is not really magic, but to someone who doesn't expect it or isn't used to this kind of things it can be a bit of magic.

CasperSchippers (approved):
Looks good to me. I quite like this improvement I have to say.

PR #1513, opened 2026-07-25
Add `unit_names` property to Agilent B1500

ilev-sergey:
Adds `unit_names` property and uses it for parsing results of queries instead of existing `smu_names`

The logic behind that PR is that previously only SMUs were supported for Agilent B1500, so when parsing measurement data, it was enough to transform SMU's slot numbers into their names. Now, CMU and SPGU are also supported, so in order to use existing data formatting methods for them, it is required to have a mapping of all units to slot numbers in which they are installed.

Basically, the purpose of the PR is to fix `format_single` behavior of `data_formating` classes. Now it maps the slot number of SMUs into their name, e.g. slot `8` that returned by the instrument as part of the data response is transformed into `"SMU4"`. But for CMU it returns the slot number without transformation, since it can't be found in `smu_names`. That breaks existing pandas logic in `read_data` and `read_channels` methods.

The existing `smu_names` mapping that is `{slot_number (int): smu_name (str)}` dict is now derived from `unit_names`, so that there is a single source of truth. I also decided to use an internal Channel `_name` field for the mapping. If you think that's not a good idea, I can rollback to defining a new `.name` field in `__init__` of each unit class.


pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -77,9 +76,28 @@ def smu_references(self) -> ValuesView[SMU]:
         return self.smus.values()
 
     @property
-    def smu_names(self) -> dict[int, str]:
-        """Get all SMU names as dict."""
-        return self._smu_names
+    def unit_names(self) -> dict[IdType, str]:
+        """Get the channel number to unit name mapping for all initialized units.
```
BenediktBurger, line 81: The channel "number" is of IdType, could also be a str:

```suggestion
        """Get the channel id to unit name mapping for all initialized units.
```

Similarly below for smu_names.

tests/instruments/agilent/test_agilentB1500.py
```diff
@@ -86,8 +86,11 @@ class AgilentB1500Mock(AgilentB1500):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.spgu1 = SPGU(self, 1)
+        self.spgu1._name = "spgu1"
```
BenediktBurger: I think it would be better to use the `add_child` method. That one will also set the name and will be used by the device itself while adding channels.

That way you may test better actual behavior instead of manually setting some name.

BenediktBurger (approved):
I like it to use the channel name (`_name`) to have a single source of truth.

The liniting failed because of some rule updates in ruff. If you merge master in your branch, it should be green (maybe some new rules will require slight adjustments on your changes).

pymeasure/instruments/agilent/agilentB1500.py
ilev-sergey: It's actually strictly int for this instrument. Do you think that I should use `IdType` here and update docstring accordingly or it's better to use plain int in this case?

BenediktBurger: I'd keep it int, if that is what the instrument requires.

When I first started adding typing, I also made id IdType, as the Channel required it. But I made it possible that the different subclasses can have different types of the id.

BenediktBurger:
Tell me once you are ready for merge.

ilev-sergey:
It can be merged, thank you for review!

Issue #1446, opened 2026-05-12
Split SCPIMixin into the two separate standards the class currently represents


OnePieceWithoutWax:
The SCPIMixin class implements basic commands like "*OPC?" which actually belong to the mandatory commands implementation defined in IEEE488.2 and are not actually part of SCPI. While SCPI 1999 is the most common SCPI standard out there that standardized the error command "SYST:ERR?" along with the "SYST" subsystem and more. 
SCPI is a subset of the IEEE standard (I'm pretty sure): i.e. all instruments that implement SCPI also implement the IEE standard, but not all instruments that implement the IEEE standard implement SCPI

This is important because instruments like the Tektronix scopes MSO5x series implement the IEEE standard, it does not implement SCPI. 

Solution:
Refactor SCPIMixin into IEEE4882Mixin and a SCPIMixin that inherits from that. Split the proper commands into each class based on their respective standards. 

Instruments that only implement the basic standard can now implement the basic command set they support without the SCPI commands that they don't

chieh0517:
Hi @BenediktBurger ,

Recently, I found this issue. It lines up with PRs I closed after being blocked for the same reason (#1483, #1484, #1485, to make SR830 and SR860 inherit `SCPIMixin`).
I hope that I can get the opportunity to enhance this program.

The structure in my mind will be:
```
class IEEE4882Mixin(CommonBase):
    ...


class SCPIMixin(IEEE4882Mixin):
    ...
```

Does the naming `IEEE4882Mixin` and this split look right to you?

florisla:
This issue may be closed now that #1514 is merged.

BenediktBurger:
Done in #1514

PR #1514, opened 2026-07-25
Split SCPIMixin to SCPIMixin and IEEE4882Mixin

chieh0517:
Spliting the IEEE 488.2 common commands from the SCPI-specific commands, as discussed in #1446.

`SCPIMixin` previously mixed both together, so instruments that only support IEEE 488.2 (but not the full SCPI command tree) had no correct base class to inherit from.

With this split, such instruments can now inherit `IEEE4882Mixin` directly, instead of inheriting `SCPIMixin` and implicitly claiming SCPI support they don't actually have.


pymeasure/instruments/generic_types.py
```diff
@@ -108,6 +117,7 @@ class SCPIUnknownMixin(SCPIMixin):
     """
 
     def __init__(self, *args, **kwargs):
+        """Initialize the mixin and warn that SCPI support for this instrument is unverified."""
```
BenediktBurger: Inits should not have a docstring as you never see that. You only see the class docstring.

pymeasure/instruments/generic_types.py
```diff
@@ -31,13 +31,16 @@
...
 
-class SCPIMixin(CommonBase):
-    """Mixin class for SCPI instruments with the default implementation of base SCPI commands."""
+class IEEE4882Mixin(CommonBase):
+    """Mixin class for IEEE 488.2 protocol instruments with
+    the default implementation of base commands.
+    """
 
     def __init__(self, *args, **kwargs):
+        """Initialize the mixin and pass through arguments to the next class in the MRO."""
```
BenediktBurger: An init does not need a docstring, you never call `IEEE4882Mixin.__init__()`, where you would see it, but directly the class.

BenediktBurger (approved):
Looks good, thank you very much.

pymeasure/instruments/generic_types.py
chieh0517: Really sorry for the wrong docstring.

pymeasure/instruments/generic_types.py
chieh0517: Really sorry for the wrong docstring. I've push a new commit for solving them.

chieh0517:
Closes #1446

PR #1518, opened 2026-07-29
Add Support for Keysight 681xB Family

sjgallagher2:
This PR adds the HP/Agilent/Keysight 681xB family of programmable AC sources ([Keysight website](https://www.keysight.com/us/en/product/6811B/performance-ac-power-source-375-va-300-v-325-a.html)). The implementation uses a base class and breaks out three models, the 6811B, 6812B, and 6813B. These models differ chiefly in their limits, which are managed by dynamic properties. In theory, the same instrument should work for the 681xC series, the latest offered from Keysight, but I don't have any available to test. Class tested on a 6812B over GPIB. 

I added API documentation pages for each of the subclasses as well as the base class, thinking this might make it easier to find the classes you're looking for, but I'm not sure if this is the right way to do it.

### DIFF
--- a/docs/api/instruments/keysight/keysight6811B.rst
+++ b/docs/api/instruments/keysight/keysight6811B.rst
@@ -1,7 +1,0 @@
-########################
-Keysight 6811B AC Source
-########################
-
-.. autoclass:: pymeasure.instruments.keysight.Keysight6811B
-    :members:
-    :show-inheritance:
--- a/docs/api/instruments/keysight/keysight6811B.rst
+++ b/docs/api/instruments/keysight/keysight6811B.rst
@@ -1,0 +1,7 @@
+########################
+Keysight 6811B AC Source
+########################
+
+.. autoclass:: pymeasure.instruments.keysight.Keysight6811B
+    :members:
+    :show-inheritance:
--- a/docs/api/instruments/keysight/keysight681xB.rst
+++ b/docs/api/instruments/keysight/keysight681xB.rst
@@ -1,0 +1,7 @@
+#########################
+Keysight 681xB AC Sources
+#########################
+
+.. autoclass:: pymeasure.instruments.keysight.Keysight681xB
+    :members:
+    :show-inheritance:
--- a/docs/api/instruments/keysight/keysight681xB.rst
+++ b/docs/api/instruments/keysight/keysight681xB.rst
@@ -2,6 +2,18 @@
 Keysight 681xB AC Sources
 #########################
 
-.. autoclass:: pymeasure.instruments.keysight.Keysight681xB
+.. autoclass:: pymeasure.instruments.keysight.keysight681xB.Keysight681xB
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.Keysight6811B
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.Keysight6812B
+    :members:
+    :show-inheritance:
+
+.. autoclass:: pymeasure.instruments.keysight.Keysight6813B
     :members:
     :show-inheritance:
--- a/pymeasure/instruments/keysight/keysight681xB.py
+++ b/pymeasure/instruments/keysight/keysight681xB.py
@@ -153,7 +153,7 @@ class Keysight681xB(SCPIMixin, Instrument):
         "TRIG:SEQ1:SOUR %s",
         """Control the trigger source for first sequence. Can be BUS|EXT|IMM.
 
-        When set to BUS, the trigger will activate after receiving a *TRG command over GPIB.
+        When set to BUS, the trigger will activate after receiving a `*TRG` command over GPIB.
         When set to EXTernal, the AC source backplane BNC trigger input is used as the trigger.
         When set to IMMediate, the trigger is generated as soon as the trigger system is initiated.
         """,
@@ -275,7 +275,7 @@ class Keysight681xB(SCPIMixin, Instrument):
 
         This method uses the existing voltage setpoint to set a voltage step level on trigger.
         The voltage setpoint is set to 0V, the output is enabled, and the trigger system is used
-        to step the voltage to the prior setpoint when we send *TRG and the waveform phase
+        to step the voltage to the prior setpoint when we send `*TRG` and the waveform phase
         reaches `phase`. A GPIB trigger is sent.
 
         After this method, the voltage setpoint will be the same as before, the trigger SYNC
--- a/pymeasure/instruments/keysight/keysight681xB.py
+++ b/pymeasure/instruments/keysight/keysight681xB.py
@@ -429,7 +429,7 @@ class Keysight681xB(SCPIMixin, Instrument):
             raise ValueError(
                 f"Length error, received array of length {len(data)}; length must be 1024."
             )
-        data_f = data.astype(float)
+        data_f = np.asarray(data, dtype=float)
 
         # Preprocess name, delete existing trace if present
         name = name.upper()
--- a/pymeasure/instruments/keysight/keysight681xB.py
+++ b/pymeasure/instruments/keysight/keysight681xB.py
@@ -1,0 +1,492 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2026 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+from time import sleep
+
+import numpy as np
+
+from pymeasure.instruments import Instrument
+from pymeasure.instruments.generic_types import SCPIMixin
+from pymeasure.instruments.validators import strict_discrete_set, strict_range
+from pymeasure.instruments._strenum import StrEnum
+
+
+class Functions(StrEnum):
+    SINE = "SIN"
+    SQUARE = "SQU"
+    CLIPPED_SINE = "CSIN"  # Clipped sinewave output. Both positive and negative peak amplitudes
+    # are clipped at a value determined by the FUNC:CSIN command.
+
+
+class Modes(StrEnum):
+    FIXED = "FIX"  # unaffected by a triggered output transient.
+    STEP = "STEP"  # programmed to the value set by X:TRIG when a triggered transient occurs.
+    PULSE = "PULS"  # changed to the value set by X:TRIG for a duration determined by the pulse
+    # commands.
+    LIST = "LIST"  # controlled by the waveform shape list when a triggered transient occurs.
+
+
+class Keysight681xB(SCPIMixin, Instrument):
+    """Represents the Keysight 681xB series of AC Power Source/Analyzers.
+
+    Do not use this class directly; use one of its subclasses.
+    """
+
+    _BOOLS = {True: 1, False: 0}
+    FREQ_RANGE = [45, 1000]
+    ROUT_RANGE = [0, 1]
+    LOUT_RANGE = [20e-6, 1e-3]  # Henries
+
+    def __init__(self, adapter, name="Keysight 681xB AC Power Source/Analyzer", **kwargs):
+        super().__init__(adapter, name, **kwargs)
+
+    voltage_setpoint = Instrument.control(
+        "VOLT?",
+        "VOLT %f",
+        """Control the AC RMS voltage amplitude setpoint in volts.""",
+        validator=strict_range,
+        values=[0, 1],
+        dynamic=True,
+    )
+
+    clipped_sine_setpoint_pct = Instrument.control(
+        "FUNC:CSIN?",
+        "FUNC:CSIN %f",
+        """Control clipped sine clipping point as a percent (0-100) of peak amplitude.""",
+        validator=strict_range,
+        values=[0.0, 100.0],
+    )
+
+    current_setpoint = Instrument.control(
+        "CURRENT?",
+        "CURRENT %f",
+        """Control the AC RMS current limit setpoint in amperes.""",
+        validator=strict_range,
+        values=[0, 1],
+        dynamic=True,
+    )
+
+    frequency_setpoint = Instrument.control(
+        "FREQ?",
+        "FREQ %f",
+        """Control the frequency setpoint in hertz""",
+        validator=strict_range,
+        values=FREQ_RANGE,
+    )
+
+    voltage_dc = Instrument.measurement("MEAS:VOLT:DC?", """Measure DC voltage in volts.""")
+    voltage_ac = Instrument.measurement("MEAS:VOLT:AC?", """Measure AC RMS voltage in volts.""")
+    voltage_acdc = Instrument.measurement("MEAS:VOLT:ACDC?", """Measure ACDC voltage in volts.""")
+    current_dc = Instrument.measurement("MEAS:CURR:DC?", """Measure DC current in amperes.""")
+    current_ac = Instrument.measurement("MEAS:CURR:AC?", """Measure AC RMS current in amperes.""")
+    current_acdc = Instrument.measurement("MEAS:CURR:ACDC?", """Measure ACDC current in amperes.""")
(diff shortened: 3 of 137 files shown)


docs/api/instruments/keysight/keysight681xB.rst
```diff
@@ -0,0 +1,7 @@
+#########################
+Keysight 681xB AC Sources
+#########################
+
+.. autoclass:: pymeasure.instruments.keysight.Keysight681xB
+    :members:
+    :show-inheritance:
```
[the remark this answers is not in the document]
BenediktBurger: Alternatively:

```suggestion
.. autoclass:: pymeasure.instruments.keysight.keysight681xB.Keysight681xB
    :members:
    :show-inheritance:
```

docs/api/instruments/keysight/keysight6811B.rst
BenediktBurger, line 1: I would put these subclasses in the same documentation file.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+
+    waveform = Instrument.control(
+        "FUNC?",
+        "FUNC %s",
+        """Control the output function of the ac source. Can be SIN, SQU, CSIN, or a user
+        waveform.""",
+        cast=str,
+    )
+
+    output_state = Instrument.control(
```
BenediktBurger: Why not `output_enabled`? Then True/False is self evident.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+        "OUTPUT:STATE?",
+        "OUTPUT:STATE %s",
+        """Control the enable/disable state of the AC source (bool).
+
+        See also :py:method:`output_enable()`.
```
BenediktBurger:
```suggestion
        See also :meth:`output_enable`.
```

Sphinx will show the parentheses as it knows it is a method.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+        "TRIG:SOUR?",
+        "TRIG:SEQ1:SOUR %s",
+        """Control the trigger source for first sequence. Can be BUS|EXTernal|IMMediate.
+
+        When set to BUS, the trigger will activate after receiving a *TRG command over GPIB.
+        When set to EXTernal, the AC source backplane BNC trigger input is used as the trigger.
+        When set to IMMediate, the trigger is generated as soon as the trigger system is initiated.
+        """,
+        validator=strict_discrete_set,
+        values=["BUS", "EXT", "EXTERNAL", "IMM", "IMMEDIATE"],
```
BenediktBurger: Above, you allow EXTernal, here only uppercase. I'd suggest to allow only the short or the long form. Best the way as it is given by the device, such that getter and setter use the same value.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+        """,
+        validator=strict_discrete_set,
+        values=["IMM", "IMMEDIATE", "PHAS", "PHASE"],
+        cast=str,
+    )
+
+    trigger_sync_phase = Instrument.control(
+        "TRIG:SYNC:PHASE?",
+        "TRIG:SYNC:PHASE %f",
+        """Control the trigger synchronization phase value.
```
BenediktBurger: From the limit of 360 I infer it is in degrees. Both data points (the units and the limits) are important for the user.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+        validator=strict_range,
+        values=[1, 9.9e37],
+    )
+
+    pulse_period = Instrument.control(
+        "PULSE:PER?",
+        "PULSE:PER %f",
+        """Control pulse period in seconds when trigger mode is set to PULSE.""",
+        validator=strict_range,
+        values=[0, 4.30133e5],
```
BenediktBurger:
```suggestion
        values=[0, 430133],
```

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+    )
+
+    voltage_sense_source = Instrument.control(
+        "VOLTAGE:SENSE:SOURCE?", "VOLTAGE:SENSE:SOURCE %s",
+        """Control the source from which the output voltage is sensed. Can be INTernal or
+        EXTernal.""",
+        validator=strict_discrete_set,
+        values=['EXT', 'EXTERNAL', 'INT', 'INTERNAL'],
```
BenediktBurger: See above.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+        self.output_enable(True)
+        self.voltage_trigger_mode = "STEP"
+        self.voltage_trigger_level = vset
+        self.trigger_sync_source = "PHASE"
+        self.trigger_sync_phase = trig_phase
+        sleep(1)  # MUST dwell here for trigger to work.
+        self.arm_immediate_trigger()
+        self.send_GPIB_trigger()
+
+    def output_pulse(self, Vdefault, Vpulse, pulse_period, N_pulses=1, pulse_ON_time=-1):
```
BenediktBurger: regarding naming: see the coderabbit comment on output_pulse_at_phase below.
probably `default_voltage` and `pulse_voltage` are better names.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+            self.trigger_source = trigger_source
+
+    # WAVEFORMS #
+    user_wfm_catalog = Instrument.measurement(
+        "TRACE:CATALOG?",
+        """Get the user waveform catalog.""",
+        get_process_list=lambda names: [str(name).replace('"', "") for name in names],
+    )
+
+    def get_user_wfm_data(self, name: str):
```
BenediktBurger: Thanks for adding type hints. Previously, many instruments did not have them, now I add them piece by piece.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+        :param name: internal name of user waveform.
+        :return: numpy array of y-data points with dtype float.
+        """
+        data_str = self.ask(f"TRACE:DATA? {name.upper()}")
+        data = np.array(data_str.strip().split(","), dtype=float)
+        return data
+
+    def get_user_waveform_catalog(self):
+        """Query the list of user waveforms.."""
+        print("Retrieving user waveform catalog. This might take a minute...")
```
BenediktBurger, line 388: Please don't use printing for log statements. Use the python logger instead. If you need help, don't hesitate to ask.

https://docs.python.org/3/library/logging.html

Often we use `log = logging.getLogger(__name__)` and then `log.info("Message...")` (depending on the level). See for example `AgilentB1500`.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+        names = self.user_wfm_catalog
+        cat = {}
+        for name in names:
+            if name not in ["SINUSOID", "SQUARE", "CSINUSOID"]:
+                namedata = self.get_user_wfm_data(name)
+                cat[name] = namedata
+        return cat
+
+    def delete_user_waveform(self, name: str):
```
BenediktBurger: For a method not returning anything, you can add:

```suggestion
    def delete_user_waveform(self, name: str) -> None:
```

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+        """Delete a user waveform by name."""
+        self.write(f"TRACE:DEL {name}")
+
+    def define_user_waveform_name(self, name: str):
+        """Define a user waveform name without data."""
+        self.write(f"TRACE:DEF {name}")
+
+    def add_user_waveform(self, name, data1024, delete_existing=False):
```
BenediktBurger: the 1024 is for the length, right? I would keep that number in the docstring only.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+        data point defines the amplitude that will be output at 0 degrees phase reference."
+
+        "Data points can be in any arbitrary units. The ac source scales the data to an
+        internal format that removes the dc component and ensures that the correct ac rms
+        voltage is output when the waveform is selected. When queried, trace data is returned
+        as normalized values in the range of 1. Waveform data is stored in nonvolatile memory
+        and is retained when input power is removed. Up to 12 user defined waveforms may be
+        created and stored."
+
+        :param name: str, name of waveform, will be converted to ALL CAPS.
```
BenediktBurger, line 422: Better to put the type in the signature.

tests/instruments/keysight/test_keysight681xB.py
BenediktBurger, line 1: Please add the copyright notice at the top.

BenediktBurger (changes requested):
Thanks for your work.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,494 @@
...
+    current_acdc = Instrument.measurement("MEAS:CURR:ACDC?", """Measure ACDC current in amperes.""")
+    current_amplitude = Instrument.measurement(
+        "MEAS:CURR:AMPL:MAX?", """Measure peak current amplitude in amperes."""
+    )
+
+    crest_factor = Instrument.measurement(
+        "MEAS:CURR:CRESTFACTOR?", """Measure current crest factor."""
+    )
+
+    power_dc = Instrument.measurement("MEAS:POW:DC?", """Measure DC power.""")
```
BenediktBurger, line 115: I guess it is also in Watts?

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,494 @@
...
+        """Control the AC RMS amplitude in volts of the output waveform when triggered.""",
+        validator=strict_range,
+        values=[0, 1],
+        dynamic=True,
+    )
+
+    voltage_trigger_mode = Instrument.control(
+        "VOLT:MODE?",
+        "VOLT:MODE %s",
+        """Control the voltage trigger mode. Can be FIX|STEP|PULS|LIST.""",
```
BenediktBurger, line 203: I would write it more like a normal text:

```suggestion
        """Control the voltage trigger mode. Can be 'FIX', 'STEP', 'PULS', or 'LIST'.""",
```

Ticks indicate that these are literal values and normally you separate elements via comma in a text.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,494 @@
...
+        validator=strict_discrete_set,
+        values=["FIX", "STEP", "PULS", "LIST"],
+        cast=str,
+    )
+
+    pulse_count = Instrument.control(
+        "PULSE:COUNT?",
+        "PULSE:COUNT %f",
+        """Control the number of pulses when trigger mode is set to PULS.""",
```
BenediktBurger, line 212: I guess it is an integer or can you have "half pulses"? I see, that you want to use floats for large numbers (like the limit 9.9e37), but that might help a user to understand useful values. Typically we use the following phrase to indicate type, validator, and limits (see https://pymeasure.readthedocs.io/en/latest/dev/coding_standards.html#docstrings):

```suggestion
        """Control the number of pulses when trigger mode is set to PULS (int strictly from 1 to 9.9e37).""",
```

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,494 @@
...
+        """Arm the trigger system (SEQ1). Before a trigger can have effect, the trigger subsystem
+        must be armed, or 'initialized'. This method arms the trigger for a single event."""
+        self.write("INIT:SEQ1")
+
+    def arm_continuous_trigger(self, run=True):
+        """Arm the trigger system (SEQ1). Before a trigger can have effect, the trigger subsystem
+        must be armed, or 'initialized'. This method arms or disarms the trigger for a continuous
+        run.
+
+        :param run: if True, enable continuous triggering; if False, stop continuous triggering."""
```
BenediktBurger, line 261: So `arm_continuous_trigger(run=False)` disarms the trigger? That is not that evident.
I suggest two methods, one to arm and one to disarm.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,494 @@
...
+        ```
+        """
+        vset = self.voltage_setpoint
+        self.voltage_setpoint = 0
+        self.output_enabled = True
+        self.voltage_trigger_mode = "STEP"
+        self.voltage_trigger_level = vset
+        self.trigger_sync_source = "PHAS"
+        self.trigger_sync_phase = trig_phase
+        sleep(1)  # MUST dwell here for trigger to work.
```
BenediktBurger, line 300: I would add a small note in the docstring, that this methods waits a second.
Typically methods return quite fast and someone might get surprised that seemingly nothing happens.

Ok, 1 second is not that long, but it might worth it.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,494 @@
...
+        self.arm_immediate_trigger()
+        self.send_GPIB_trigger()
+
+    def output_pulse(self, v_default, v_pulse, pulse_period, npulses=1, pulse_on_time=-1):
```
BenediktBurger, line 304: Above you already used a type hint, feel free to add them everywhere:

```suggestion
    def output_pulse(self, v_default: float, v_pulse: float, pulse_period: float, npulses: int = 1, pulse_on_time: float = -1) -> None:
```

At least I guess that pulse_on_time could be float.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,494 @@
...
+
+    def output_pulse_at_phase(
+        self,
+        v_default: float,
+        v_pulse: float,
+        pulse_period: float,
+        trig_phase: float = 0.0,
+        npulses: int = 1,
+        pulse_on_time: float = -1,
+        trigger_source: str = "GPIB",
```
BenediktBurger, line 334: Just for your information, you do not have to do it:

There exists a `typing.Literal` which allows you to specify not only a type, but a certain value. For example below code will only allow certain values for the string.

```suggestion
        trigger_source: Literal["GPIB", "BUS", "EXT", "IMM"] = "GPIB",
```

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,494 @@
...
+        if trigger_source == "GPIB" or trigger_source == "BUS":
+            self.send_GPIB_trigger()
+        else:
+            self.trigger_source = trigger_source
+
+    # WAVEFORMS #
+    user_wfm_catalog = Instrument.measurement(
+        "TRACE:CATALOG?",
+        """Get the user waveform catalog.""",
+        get_process_list=lambda names: [str(name).replace('"', "") for name in names],
```
BenediktBurger, line 372: You do not need to cast to str. If below you already cast to str, any element of names is already a string!


```suggestion
        get_process_list=lambda names: [name.replace('"', "") for name in names],
```

pymeasure/instruments/keysight/keysight681xB.py
BenediktBurger: The print line is still there. Please replace it with a log statement:

```suggestion
        log.info("Retrieving user waveform catalog. This might take a minute...")
```

For that reason you initialized the log at the top.

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,494 @@
...
+        voltage is output when the waveform is selected. When queried, trace data is returned
+        as normalized values in the range of 1. Waveform data is stored in nonvolatile memory
+        and is retained when input power is removed. Up to 12 user defined waveforms may be
+        created and stored."
+
+        :param name: str, name of waveform, will be converted to ALL CAPS.
+        :param data1024: numpy array of length 1024, dtype float, all values in [0.0, 1.0].
+                         Points will be rounded to 5 digits of precision.
+        :param delete_existing: if True, any waveform with the same name will be deleted before
+                                adding the new data. If False, raises an exception if name exists.
```
BenediktBurger, line 426:
```suggestion
                                adding the new data. If False, raises an exception if a file with that name already exists.
```

pymeasure/instruments/keysight/keysight681xB.py
```diff
@@ -0,0 +1,492 @@
...
+    IRMS_MAX = 13
+    VPEAK_MAX = 425
+    IPEAK_MAX = 80
+
+    def __init__(self, adapter, name="Keysight 6813B AC Power Source/Analyzer", **kwargs):
+        super().__init__(adapter, name, **kwargs)
+        self.voltage_setpoint_values = [0, self.VRMS_MAX]
+        self.current_setpoint_values = [0, self.IRMS_MAX]
+        self.voltage_trigger_level_values = [0, self.VRMS_MAX]
+
```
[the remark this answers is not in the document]
BenediktBurger: You could put that values setting in the parent class (these three lines).
Then you could delete the init calls altogether.

Once you create a device (e.g. `inst = Keysight6813B`), it will call the super init method, which will set the values according to the variables, e.g. according to Keysight6813B.VRMS_MAX.

tests/instruments/keysight/test_keysight681xB.py
```diff
@@ -0,0 +1,439 @@
...
+    ) as inst:
+        inst.clipped_sine_setpoint_pct = 50
+
+
+def test_clipped_sine_setpoint_pct_getter():
+    with expected_protocol(
+            Keysight6812B,
+            [(b'FUNC:CSIN?', b'1.00000E+2\n')],
+    ) as inst:
+        assert inst.clipped_sine_setpoint_pct == 100.0
```
BenediktBurger, line 51: Just for your information: It would work also with the integer:

```suggestion
        assert inst.clipped_sine_setpoint_pct == 100
```

BenediktBurger (changes requested):
A few more comments, but it looks already good.

PR #1519, opened 2026-08-01
Rework unit initialization in Agilent B1500

ilev-sergey:
Introduces new `initialize_units` method that combines `initialize_all_smus`, `initialize_cmu`, `initialize_all_spgus` that are now deprecated.

Also fixes existing bug that I accidentally introduced in the initial [PR](https://github.com/pymeasure/pymeasure/pull/1344) with SPGU implementation while [reworking](https://github.com/pymeasure/pymeasure/pull/1344/changes/10b7968de7c2bc1c3eec72832db7890cb96d4dd9#diff-0a9f5185d06142806c6c4dfc46f4133c909f59f90ff3a8c8dc7292f018460e33L180-L182) SPGU class into Channel subclass. As was previously [discussed](https://github.com/pymeasure/pymeasure/pull/1344#discussion_r2301380809), in this instrument the dynamically initialized units are available by their ordinal numbers (`.smu1`, `.smu2`, etc.), while the actual slot number that is used when sending commands is present as `.id` attribute. The same logic should exist for SPGU units. I accidentally broke that while reworking and didn't notice that since in our setup SPGU is installed in slot 1, so it is accessible as `.spgu1` even with current realization.

According to the manual, 
> If the SPGUs are installed, the SPGUs must be installed in the slots from the slot number 1

So, this change should not have an effect for most cases anyway.

### DIFF
--- a/pymeasure/instruments/agilent/agilentB1500.py
+++ b/pymeasure/instruments/agilent/agilentB1500.py
@@ -154,7 +154,7 @@ class AgilentB1500(SCPIMixin, Instrument):
                     raise NotImplementedError(f"Module {module[0]} is not implemented yet!") from e
         return out
 
-    def initialize_all_units(self) -> None:
+    def initialize_units(self) -> None:
         """Initialize all supported units.
 
         Query available modules once and create a :class:`.SMU`, :class:`.SPGU`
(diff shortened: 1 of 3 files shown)


pymeasure/instruments/agilent/agilentB1500.py
```diff
@@ -1791,11 +1860,13 @@ class SPGU(Channel):
     """Provide specific methods for the SPGU module of the Agilent B1500 mainframe.
 
     :param parent: Instance of the B1500 mainframe class
-    :param channel: Channel number of the SPGU
+    :param index: Index of the SPGU
+    :param slot: Slot number of the SPGU
```
BenediktBurger, line 1864: What is the difference?

BenediktBurger (approved):
Thank you for your work.

PR #1521, opened 2026-08-16
Fix AMI430.has_persistent_switch_enabled always returning True

### LINKED ISSUE #1205: AMI 430 Power Supply "has_persistent_switch_enabled" function not working
The "has_persistent_switch_enabled" function in the AMI430 class always returns "True":

def has_persistent_switch_enabled(self):
      """ Returns a boolean if the persistent switch is enabled. """
      return bool(self.ask("PSwitch?"))

As "self.ask("PSwitch?")" returns a non-empty string ("0" or "1") "bool(self.ask("PSwitch?"))" will always return "True".

calling "bool(int(self.ask("PSwitch?")))" instead should work.

Edit: Some minor comments regarding the documentation:

"magnet.state" will not return the field but the state as float.
In line 52 this is also documented instead as magnet.status.

The units of the ramp rate/ field etc. functions depend on the unit setting in the AMI 340 (can be set to T or kG). Here the documentation is confusing as one might think that the python code will always set it in kG.

hmms:
## Summary

`AMI430.has_persistent_switch_enabled()` (`pymeasure/instruments/ami/ami430.py`) did `return bool(self.ask("PSwitch?"))`. Since `ask()` returns a string and any non-empty string is truthy in Python, this always returned `True` regardless of the actual switch state.

Fixed by parsing the reply before comparing: `return int(self.ask("PSwitch?")) == 1`, matching the same int-string-to-bool idiom already used elsewhere in the codebase (`sr830.py`, `anritsuMG3692C.py`, `ametek7270.py`).

The issue also flagged the class docstring's usage example pointing at `magnet.status` for "Returns the status of the magnet" — `status` is the inherited `*STB?` GPIB status byte, not the magnet's actual status (`magnet_status`/`state` are). Fixed the example to `magnet.magnet_status`.

Fixes #1205

## Not addressed here

The issue's third comment (docstrings hardcoding kGauss units, e.g. on `coilconst`/`target_field`/`ramp_rate_field`, when the device's unit is actually configurable to kG or T) is left open, no concrete wording was proposed and it touches several unrelated docstrings. Happy to take it on separately if useful.

## Test plan

- Added `tests/instruments/ami/test_ami430.py` (previously no coverage existed for this driver at all), using the existing `expected_protocol` pattern: covers both `True`/`False` replies for `has_persistent_switch_enabled`, plus basic coverage for `enable_persistent_switch`/`disable_persistent_switch`.
- `pytest tests/instruments/ami/` — 4/4 passed. `pytest tests/instruments/test_all_instruments.py -k ami` — 25/25 passed, no regression.
- `ruff check` and `pyright` on both changed files — clean.

### DIFF
--- a/tests/instruments/ami/test_ami430.py
+++ b/tests/instruments/ami/test_ami430.py
@@ -39,9 +39,6 @@ def test_has_persistent_switch_enabled_true():
 
 
 def test_has_persistent_switch_enabled_false():
-    # Regression test: bool(self.ask(...)) is always True for a non-empty
-    # string reply, so a "0" (switch disabled) reply was incorrectly
-    # reported as enabled. See github.com/pymeasure/pymeasure/issues/1205.
     with expected_protocol(
             AMI430,
             INIT + [(b"PSwitch?", b"0")],
--- a/tests/instruments/ami/test_ami430.py
+++ b/tests/instruments/ami/test_ami430.py
@@ -1,0 +1,65 @@
+#
+# This file is part of the PyMeasure package.
+#
+# Copyright (c) 2013-2026 PyMeasure Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


tests/instruments/ami/test_ami430.py
```diff
@@ -0,0 +1,65 @@
...
+            AMI430,
+            INIT + [(b"PSwitch?", b"1")],
+    ) as instr:
+        assert instr.has_persistent_switch_enabled() is True
+
+
+def test_has_persistent_switch_enabled_false():
+    # Regression test: bool(self.ask(...)) is always True for a non-empty
+    # string reply, so a "0" (switch disabled) reply was incorrectly
+    # reported as enabled. See github.com/pymeasure/pymeasure/issues/1205.
```
BenediktBurger: We do not need this comment. It is only useful for this PR, not for the code itself.

BenediktBurger (approved):
Thank you for fixing this issue.

If you want to, you may add yourself to the authors file.

hmms:
Thanks, Benedikt! I've addressed your comments in latest commit. Please go ahead and merge whenever you're ready.

PR #1525, opened 2026-08-20
Agilent E5270B add auto ranging

Falk3579:
- Add limited auto ranging to the current measurement. This improves the resolution for small currents if the SMU channel has a high current compliance set. The auto ranging does not go below the 1nA range to keep the measurement time at a reasonable level.
- Add auto ranging for voltage measurement. This improves the resolution for small voltages if the SMU channel has a high voltage compliance set.
- Adapt protocol test to auto ranging

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,7 @@ Changed
   ``Procedure.refresh_parameters`` is now a deprecated no-op (retained for API compatibility).
   ``UnknownProcedure`` now returns empty ``parameter_objects``/``metadata_objects`` dicts so loading an unimportable procedure no longer raises.
 - Procedure use :class:`ProcedureStatus` enum instead of status and status message dicts.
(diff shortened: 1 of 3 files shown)


BenediktBurger (approved):
Thanks for the changes.

One question: Does it have any unexpected effect in comparison to the earlier command? E.g. longer measurement time, other resolution, other range?

Falk3579:
The measurement time can increase slightly if the actual current is quite low (e.g. a few pA). With the previous version the measurement range was determined by the compliance value. This lead to low resolution in case of high compliance but low measured current.

BenediktBurger:
Do we have to warn users about that change or is the increased measurement time that low, that they won't notice the increase but get better results?

Falk3579:
I short notice to the users would be good.

BenediktBurger:
Maybe add a note to the changes.rst file in the changed section, such that user will read about it.

Falk3579:
I added a comment in the changes.rst.

BenediktBurger:
Thank you for your work!

Issue #1524, opened 2026-08-18
DHO800 buggy get_waveform method when using MAX/RAW mode


7omg:
Hi,
I noticed that when doing consecutive `get_waveform` calls with MAX or RAW mode, while the scope is in RUN state, the second `get_waveform` call will hang indefinetely. 

Due to RAW/MAX mode requiring the scope to be in STOP state, the method automatically puts the scope in STOP state and remembers the scope's original state (RUN in this case). However when the original state is restored to RUN the internal buffer is cleared(?) and the next `get_waveform` call comes too fast that the scope can update its buffer in between.

```python
import logging

from pymeasure.instruments.rigol.rigol_dho_base import DHOBase

log = logging.getLogger(__name__)
logging.basicConfig(level=logging.DEBUG)
logging.getLogger("pyvisa").setLevel(logging.WARNING)

scope = DHOBase("USB0::6833::1101::DHO8A272806674::0::INSTR")
scope.run()

# this get_waveform call will put scope in STOP and then again in RUN
# which clears the internal buffer
data = scope.get_waveform(channel=1, mode="MAX", fmt="WORD")

# next get_waveform calls will hang trying to read buffer contents
data = scope.get_waveform(channel=1, mode="MAX", fmt="WORD")
print(len(data), data)
```

Adding a sleep between the calls avoids the issue, but requiring users to insert an arbitrary delay is undesirable.

I thought of two ways to fix this:
1. Don't restore previous state, instead force put scope in STOP mode and display warning
2. raise RuntimeError when MAX/RAW mode is requested and scope is not in STOP mode

I think the second option is preferable because it makes the state requirement explicit and avoids silently changing the scope's acquisition state.

Please let me know which approach you prefer, and I'll make a PR.

BenediktBurger:
As written in the PR: Option 2, raising an error, is the best way.

PR #1529, opened 2026-08-23
Fix consecutive data acquisitions in MAX/RAW mode for DHO800/900

7omg:
Fixes bug described in #1524


pymeasure/instruments/rigol/rigol_dho_base.py
```diff
@@ -550,46 +550,32 @@ def get_waveform(self, channel=1, mode="NORM", fmt="BYTE"):
         chunk_size = 1000
         dtype = "H" if fmt == "WORD" else "B"
 
-        # Stop scope if needed, remember state to restore later
+        # Check status for MAX/RAW mode
         was_running = self.trigger_status != "STOP"
         if mode in ("MAX", "RAW") and was_running:
-            self.stop()
-            time.sleep(0.1)  # wait for scope to be stopped
+            raise RuntimeError(f"Scope must be in STOP state to acquire data in {mode} mode!")
```
BenediktBurger, line 575: Quite often it is good to put a value into quotes, e.g. `" in '{mode}' mode!"`, because it is not always clear what belongs to the variable (maybe even spaces) and what is "text".

In your case, you only have MAX and RAW, such that this additional measure is not necessary.

BenediktBurger (approved):
Thank you for the fixes. Yes, the solution to raise an error is the best option. Sorry for not having answered your issue yet.

If you want to, you can put your name in the authors file. Otherwise, please let me know that I can merge without your name there.

7omg:
Added my name to the authors file. Ready to merge!

7omg:
Hi @BenediktBurger, is there anything else needed from my side?

BenediktBurger:
Thank you for your work. I just had to return from vacation in order to merge.

PR #1542, opened 2026-09-03
Fix for FutureWarning in Formfactor Velox driver

Falk3579:
Change cast from 'int' to 'cast_or_str(int)'

### DIFF
--- a/pymeasure/instruments/formfactor/velox.py
+++ b/pymeasure/instruments/formfactor/velox.py
@@ -142,7 +142,7 @@ class WaferMap(Channel):
     end_reached = False
     """Get whether the end of the wafermap is reached or not (bool)."""
 
-    def step_first_die(self) -> list[int] | str:
+    def step_first_die(self) -> list[int | str]:
         """Move the chuck to the first die and clear the the binning results.
 
         :return: List of the int containing the coordinates of the first die.
@@ -158,7 +158,7 @@ class WaferMap(Channel):
                            cast=cast_or_str(int),
                            )
 
-    def step_next_die(self) -> list[int] | str:
+    def step_next_die(self) -> list[int | str]:
         """Move the chuck to the next logical die.
 
         :return: List of the int containing the coordinates of the new die.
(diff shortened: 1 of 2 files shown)


pymeasure/instruments/formfactor/velox.py
```diff
@@ -155,10 +155,10 @@ def step_first_die(self) -> list[int]:
         """
         return self.values("StepFirstDie",
                            separator=" ",
-                           cast=int,
+                           cast=cast_or_str(int),
```
BenediktBurger, line 158: What do we want to return at the end of a wafer?

With your modification it would be `["End", "of", "wafer"]`.

Wouldn't it better to fail on str and return None or raise an error? (it would break current behavior, though).

tests/instruments/formfactor/test_formfactorVelox.py
BenediktBurger, line 150: If we want to return something in case of error, we should check for it:
~~~python
with ()...:
   result = inst.wafermap.step_next_die()
assert result == expected_result
~~~

BenediktBurger (changes requested):
Thank you for fixing my typing work.

I have a question on how you would like the instrument to work, as I guess you use it.

pymeasure/instruments/formfactor/velox.py
Falk3579: Yes, I use this prober.
If the prober has reached the end of the wafermap it responses with "703: End of wafer -- StepNextDie".
`read` splits the return string at the colon in error code and response and sets `self.error_code` and `self.error_message` if the error code is <>0.
In my workflow I always check `self.error_code` before I process the response any further. So I don't care about a list of strings.
Errors different from 703 already raise a `ConnectionError` in `read`.

tests/instruments/formfactor/test_formfactorVelox.py
Falk3579: I added tests for `self.error_code` and `self.error_message`

BenediktBurger:
Thank you for your work!

PR #1543, opened 2026-09-07
Fix AQ6370 ethernet authentication case sensitivity

niklasschulz99:
# Description

`AQ6370Series.authenticate_ethernet()` compared the instrument's handshake responses (`AUTHENTICATE CRAM-MD5.` and `READY`) with a case-sensitive, exact assert. This fails on hardware whose firmware replies in lower case: an AQ6370B (firmware 01.01) answers `ready`, so the login actually succeeds but the assert raises and aborts the connection.

Found while connecting to a real AQ6370B over Ethernet.

# Changes

- Compare the responses case-insensitively (`.strip().upper()`), so lower-case firmware replies are accepted.
- Replace the two assert statements with an explicit `raise ConnectionError(...)` — asserts are stripped under python -O and are not appropriate for library control flow; the exception also reports the unexpected response.
- Add protocol tests covering the lower-case handshake and the failure case.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,10 @@ Changed
 - Procedure use :class:`ProcedureStatus` enum instead of status and status message dicts.
 - Added auto ranging to Agilent E5270B ``voltage`` and ``current``. It improves measurement resolution but might slightly increase the measurement time.
 
+Instruments
+-----------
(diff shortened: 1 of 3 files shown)


BenediktBurger:
Thank you very much.

PR #1544, opened 2026-09-07
Fix AQ6370 wavelength_start upper bound typo

niklasschulz99:
### Description

`AQ6370Series.wavelength_start` had an upper validator bound written as `1700 - 9`, which evaluates to **1691** (metres) instead of `1700e-9`. The limit was therefore effectively unbounded for any real wavelength, so out-of-range start wavelengths were never rejected.

### Changes

- Fix the `wavelength_start` upper bound to `1700e-9` (manual: START WL **50–1700 nm** for AQ6370C/D).
- Correct the docstring, which wrongly stated `2250e-9` — that is the **STOP** range (`wavelength_stop`: 600–2250 nm), not the start range.
- Add a regression test asserting that a start wavelength above 1700 nm now raises `ValueError`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,10 @@ Changed
 - Procedure use :class:`ProcedureStatus` enum instead of status and status message dicts.
 - Added auto ranging to Agilent E5270B ``voltage`` and ``current``. It improves measurement resolution but might slightly increase the measurement time.
 
+Instruments
+-----------
+- Fix Yokogawa AQ6370 ``wavelength_start`` upper bound, which was 1691 m due to a ``1700 - 9`` typo (now ``1700e-9``).
+
 Version 0.16.0 (2026-05-20)
 ===========================
 
--- a/tests/instruments/yokogawa/test_aq6370d.py
+++ b/tests/instruments/yokogawa/test_aq6370d.py
@@ -178,6 +178,14 @@ def test_wavelength_start_setter():
         inst.wavelength_start = 8e-07
 
 
+def test_wavelength_start_upper_bound():
+    # Regression: the upper bound was 1691 m (typo "1700 - 9"), an absurdly large
+    # limit that never rejected anything. Per the manual the start wavelength is
+    # 50 to 1700 nm, so a value above 1700 nm must now be rejected.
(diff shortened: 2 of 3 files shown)


tests/instruments/yokogawa/test_aq6370d.py
```diff
@@ -178,6 +178,14 @@ def test_wavelength_start_setter():
         inst.wavelength_start = 8e-07
 
 
+def test_wavelength_start_upper_bound():
+    # Regression: the upper bound was 1691 m (typo "1700 - 9"), an absurdly large
+    # limit that never rejected anything. Per the manual the start wavelength is
+    # 50 to 1700 nm, so a value above 1700 nm must now be rejected.
```
BenediktBurger: That comment was only useful for the PR, afterwards, it is not necessary to tell people about earlier typos in a comment.

Please remove the comment.

CHANGES.rst
```diff
@@ -31,6 +31,10 @@ Changed
 - Procedure use :class:`ProcedureStatus` enum instead of status and status message dicts.
 - Added auto ranging to Agilent E5270B ``voltage`` and ``current``. It improves measurement resolution but might slightly increase the measurement time.
 
+Instruments
+-----------
+- Fix Yokogawa AQ6370 ``wavelength_start`` upper bound, which was 1691 m due to a ``1700 - 9`` typo (now ``1700e-9``).
```
BenediktBurger: This will lead to a merge conflict now.

Typically, the maintainer create the obvious messages from the PR titles.
We only add those fixes manually during PRs for things which are easy to overlook or are quite important (deprecations etc.).

BenediktBurger (approved):

tests/instruments/yokogawa/test_aq6370d.py
niklasschulz99: Good point - removed the comment, kept the test. Thanks!

CHANGES.rst
niklasschulz99: Understood - removed the CHANGES.rst entry, which also resolves the merge conflict. Thanks for the context!

PR #1545, opened 2026-09-07
Add SRS DC205 precision DC voltage source

niklasschulz99:
### Description

Adds a driver for the Stanford Research Systems DC205 precision DC voltage source (`srs/dc205.py`).

### Highlights

- `voltage` uses a dynamic validator that follows `voltage_range` (±1.01/±10.1/±101 V). The DC205 does not auto-range, so an out-of-range voltage raises `ValueError` until the range is changed.
- Full scan configuration (range, begin/end, time, shape, cycle, display) plus `scan_state` and `arm_scan()` / `disarm_scan()` / `start_scan()` (`*TRG`).
- Numeric response mode is forced on connection (`TOKN 0`) so the mapped properties parse reliably.
- 20 protocol tests; docs and CHANGES added.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,10 @@ Changed
 - Procedure use :class:`ProcedureStatus` enum instead of status and status message dicts.
 - Added auto ranging to Agilent E5270B ``voltage`` and ``current``. It improves measurement resolution but might slightly increase the measurement time.
 
+Instruments
+-----------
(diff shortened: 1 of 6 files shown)


niklasschulz99:
Closed/reopened to re-trigger CI
The failures were transient setup-micromamba download errors (504 / binary-not-found), unrelated to this PR, and I can't re-run the jobs as a fork contributor.

PR #1550, opened 2026-09-12
Add strict_discrete_case_insensitive_set validator

### LINKED ISSUE #1364: Suggestion: Case Insensitive Validators
These mixed case strings are hard to memorize.
I know, the manual states such a string as you can use the short or long form.
However, they are more difficult to read/write.

What does the device return?
What does the device accept? -> I guess case insensitive.

What do you think about a `strict_discrete_case_insensitive_set` validator, which does not validate the case?

_Originally posted by @BenediktBurger in https://github.com/pymeasure/pymeasure/pull/1325#discussion_r2351811741_

blkmanba1:
Closes #1364

### Motivation

As suggested in #1364, values such as `NORMal` or `AVERage` are hard to read and write, while the instruments themselves accept those command strings case-insensitively.
This adds a `strict_discrete_case_insensitive_set` validator that accepts any spelling and sends the spelling documented in `values`.

### Note on the return value

The validator returns the matched element of `values` rather than the value passed in, so that devices receive the documented spelling.
This is also required for `map_values=True` to keep working: the setter looks the validated value up in `values` (`values.index(val)` for sequences, `values[val]` for mappings), which fails for a differently-cased spelling.

### Validation

- `pytest tests/instruments/test_validators.py` - 9 passed
- `pytest tests/instruments/test_common_base.py` - 111 passed
- `pytest` (full suite) - 9286 passed, 3457 skipped, 1 xfailed
- `ruff check`, `flake8`, `pyright` - clean
- `make html` and `make doctest` with `-W` - pass; the 3 pre-existing `ni.virtualbench`/`nicaiu` warnings are identical on unmodified `master`

### Changes

- `pymeasure/instruments/validators.py`: new validator
- `tests/instruments/test_validators.py`: new test
- `docs/dev/adding_instruments/properties.rst`: short description and example


pymeasure/instruments/validators.py
```diff
@@ -104,6 +104,30 @@ def strict_discrete_set(value: T, values: Iterable[T] | dict[T, Any]) -> T:
         raise ValueError(f'Value of {value} is not in the discrete set {values}')
 
 
+def strict_discrete_case_insensitive_set(value: T, values: Iterable[T] | dict[T, Any]) -> T:
+    """Validate a value against a discrete set, ignoring the case of strings.
+
+    Returns the element of ``values`` that matches ``value``.
```
BenediktBurger:
```suggestion
    Returns the element of ``values`` that matches ``value`` case insensitively.
```

If they match exactly, you could return any one, the clue is the insensitivity, right?

tests/instruments/test_validators.py
```diff
@@ -65,6 +66,17 @@ def test_strict_discrete_set():
         strict_discrete_set(20, range(10))
 
 
+def test_strict_discrete_case_insensitive_set():
+    values = ["NORMal", "AVERage"]
+    assert strict_discrete_case_insensitive_set("NORMal", values) == "NORMal"
+    assert strict_discrete_case_insensitive_set("normal", values) == "NORMal"
+    assert strict_discrete_case_insensitive_set("AvErAgE", values) == "AVERage"
+    assert strict_discrete_case_insensitive_set("ON", {"ON": 1, "OFF": 0}) == "ON"
```
BenediktBurger: Ok, maybe here also a mixed case?

